PR #196, opened 2020-11-12
Copy logs into the OS clipboard for mobile

bakulf:
This patch covers this scenario: I see the logs and I want to share them. I cannot copy them. Now there is a button to do it.

A couple of comments:
1. the UI is ugly :)
2. maybe we should give a feedback when the user clicks on 'copy'.


lesleyjanenorton (approved):
R++

PR #231, opened 2020-11-17
Fix #135 - Improve toggle a11y

### LINKED ISSUE #135: Improve VPNToggle's accessibility.
`VPNToggle` accessible naming seems to be a bit special case. And this issue is 2 part:

* Looks like Qt has a bug with supporting accessibility name and description. Description is not exposed as supplementary information for an accessible object and instead overrides accessible name. So right now the name of the toggle button that is exposed to screen reader is just its description. This means we can not use name and descriptions properties as intended. The best way to mitigate that (though not ideal) is to combine name and description as part of the name (similar to https://github.com/mozilla-mobile/mozilla-vpn-client/blob/master/src/ui/settings/ViewLanguage.qml#L63-L66)

* It seems like the best way to be consistent with what's visible (e.g. text and description above the toggle) and the name/description for the toggle as well as avoid duplication is to add 2 property variables to toggle `describedby` and `labelledby` (similar to how ARIA allows for that on the WEB). The value for those properties would be the references to `logoTitle` and `logoSubtitle`. Then we would build `Accessible.name` for the toggle as follows:
```
//% "%1 %2"
Accessible.name: qsTrId("vpn.toggle.accessibleName")
    .arg(labelledby.text)
    .arg(describedby.text)
```

I believe, this should also simplify (allow for removal of) property changes on state change inside the toggle that are related to accessible name, description.

lesleyjanenorton:
Add tooltip to toggle and improve accessible name. - Fix #135

### DIFF
--- a/src/ui/components/VPNToggle.qml
+++ b/src/ui/components/VPNToggle.qml
@@ -12,6 +12,12 @@ VPNButtonBase {
     id: toggleButton
 
     property var toggleColor: Theme.vpnToggleDisconnected
+    property var describedBy: logoTitle.text
+    property var labelledBy: logoSubtitle.text
+    property var toolTipTitle: ""
+
+
+    Accessible.name: VPNController.state === VPNController.StateOn ? describedBy + ". " + toolTipTitle : describedBy + ". " + labelledBy + "."
 
     function handleClick() {
         if (VPNController.state !== VPNController.StateOff)
@@ -66,10 +72,7 @@ VPNButtonBase {
             PropertyChanges {
                 target: toggleButton
                 //% "Turn VPN on"
-                Accessible.name: qsTrId("vpn.toggle.on")
-                //% "VPN is off"
-                Accessible.description: qsTrId("vpn.toggle.isOff")
-                toggleColor: Theme.vpnToggleDisconnected
+                toolTipTitle: qsTrId("vpn.toggle.on")
             }
 
         },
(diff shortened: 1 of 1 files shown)


src/ui/components/VPNToggle.qml
```diff
@@ -12,6 +12,12 @@ VPNButtonBase {
     id: toggleButton
 
     property var toggleColor: Theme.vpnToggleDisconnected
+    property var describedBy: logoTitle.text
+    property var labelledBy: logoSubtitle.text
+    property var toolTipTitle: ""
+
+
+    Accessible.name: VPNController.state === VPNController.StateOn ? describedBy + ". " + toolTipTitle : describedBy + ". " + labelledBy + "."
```
lesleyjanenorton, line 20: @yzen  After trying this with the default mac screen reader, I left `labelledBy` out of the `Accessible.name` when the VPN is connected because it resets itself every second and gets very noisy and difficult to understand.

bakulf (approved):

src/ui/components/VPNToggle.qml
yzen: Thanks for testing it with VO :). I tried as well and it still feels a little verbose. I think I'm gonna take back my comment about needing both label and description here as the description does not really add much value.

How about we just keep:
```
property var toolTipTitle: ""
Accessible.name: toolTipTitle
```

and avoid setting labelledBy and describedBy properties.

yzen (approved):
Thanks for PR, It looks good with comments addressed, thanks!

PR #240, opened 2020-11-17
Support Start On Boot && AlwaysOn

### LINKED ISSUE #180: "Launch VPN app on Startup"  - does not work on Android

strseb:
Would Fix #180 -
-> Adds the Ability for Supported API levels to Start the API on boot 
-> Adds Support for the Always ON Vpn android setting

Marking on wip as I neeed to test lower api levels, just tested on my pixel so far.

### DIFF
--- a/android/src/com/mozilla/vpn/VPNService.kt
+++ b/android/src/com/mozilla/vpn/VPNService.kt
@@ -4,18 +4,16 @@
 
 package com.mozilla.vpn
 
-import android.R
-import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.PendingIntent
+import android.app.Service
 import android.content.Context
 import android.content.Intent
 import android.os.Build
 import android.os.IBinder
 import android.util.Log
 import androidx.core.app.NotificationCompat
-import androidx.core.content.ContextCompat
 import com.wireguard.android.backend.Statistics
 import com.wireguard.android.backend.Tunnel
 import com.wireguard.android.backend.VpnServiceBackend
@@ -23,12 +21,18 @@ import com.wireguard.android.backend.applyConfig
 import com.wireguard.config.Config
 import com.wireguard.crypto.Key
 import com.wireguard.crypto.KeyFormatException
+import java.lang.reflect.Array.newInstance
+import java.lang.reflect.Field
+import java.lang.reflect.Method
+
 
 class VPNService : android.net.VpnService() {
     private val tag = "VPNService"
     var tunnel: Tunnel? = null
     private var mBinder: VPNServiceBinder? = null
-
+    val NOTIFICATION_CHANNEL_ID = "com.mozilla.vpnNotification"
+    val CONNECTED_NOTIFICATION_ID = 1337
+    private val mNotificationBuilder = NotificationCompat.Builder(this, NOTIFICATION_CHANNEL_ID);
     /**
      * EntryPoint for the Service, gets Called when AndroidController.cpp
      * calles bindService. Returns the [VPNServiceBinder] so QT can send Requests to it.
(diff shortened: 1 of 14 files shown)


android/src/com/mozilla/vpn/VPNService.kt
```diff
@@ -176,13 +220,14 @@ class VPNService : android.net.VpnService() {
         val intent = Intent(this, activity)
         val pendingIntent = PendingIntent.getActivity(this, 0, intent, 0)
 
-        val builder = NotificationCompat.Builder(this, NOTIFICATION_CHANNEL_ID)
+        mNotificationBuilder
```
bakulf, line 223: I would like to have the notification controlled by the app, but I guess, at this point, the app has not been executed yet, right?

src/platforms/android/androidstartatbootwatcher.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+AndroidStartAtBootWatcher::AndroidStartAtBootWatcher(bool startAtBoot)
+{
+    startAtBootChanged(startAtBoot);
+}
+
+void AndroidStartAtBootWatcher::startAtBootChanged(bool startAtBoot)
+{
+    logger.log() << "StartAtBoot changed:" << startAtBoot;
+    AndroidController::Instance()->enableStartAtBoot(startAtBoot);
```
bakulf, line 27: mmm are you sure the controller is activated at this time?

src/platforms/android/androidutils.h
```diff
@@ -17,6 +20,8 @@ class AndroidUtils final : public QObject
 public:
     static QString GetDeviceName();
 
+    static bool CanEnableStartOnBoot();
```
bakulf, line 23: lower case 'c'

bakulf (approved):

android/src/com/mozilla/vpn/VPNService.kt
strseb: Yep, i.e. connect-on-boot or always-on are enabled the service will get started from the OS without starting our qt code.

android/src/com/mozilla/vpn/VPNService.kt
```diff
@@ -133,7 +175,10 @@ class VPNService : android.net.VpnService() {
         if (fileDescriptor != null) {
             Log.v(tag, "Got file Descriptor for VPN - Try to up")
             backend.tunnelUp(tunnel, fileDescriptor, config.toWgUserspaceString())
-            this.startSticky()
+            mNotificationBuilder
+                .setContentTitle("Todo: Connected")
```
bakulf, line 179: We need to localize these strings... The best would be that the app sends the localized strings to the service as preference.

src/platforms/android/androidcontroller.cpp
```diff
@@ -215,11 +233,17 @@ bool AndroidController::VPNBinder::onTransact(int code,
     Q_UNUSED(flags);
 
     QJsonDocument doc;
-    QString logs;
+    QString buffer;
     switch (code) {
     case EVENT_INIT:
         logger.log() << "Transact: init";
-        emit m_controller->initialized(true, false, QDateTime());
+         buffer = QTextCodec::codecForMib(106)->toUnicode(data.readData());
```
bakulf, line 240: Can you make a separate function to do this step. Otherwise I'll ask you what 106 is each time.

src/platforms/android/androidcontroller.cpp
```diff
@@ -242,9 +266,9 @@ bool AndroidController::VPNBinder::onTransact(int code,
         logger.log() << "Transact: backend logs";
 
         // Note: 106 means UTF-8 encoding
-        logs = QTextCodec::codecForMib(106)->toUnicode(data.readData());
+        buffer = QTextCodec::codecForMib(106)->toUnicode(data.readData());
```
bakulf, line 269: here too

bakulf (changes requested):
All good except the AndroidController::instance.
In the latest version of the code, we delete the controllerImpl when the user completes a logout or when the server sends an authentication failure.
This is done because we need to unregister and register the current device. During this operation, we generate new private/public keys and we reinitialize a new controller.

Would it work with your new setup?

PR #263, opened 2020-11-19
Fix VPNToggle a11y regression. Click is not handled when Space/Return…

yzen:
… are pressed or a11y action is triggered.

### DIFF
--- a/src/ui/components/VPNToggle.qml
+++ b/src/ui/components/VPNToggle.qml
@@ -27,6 +27,8 @@ VPNButtonBase {
         return VPN.activate();
     }
 
+    onClicked: handleClick()
+
     // property in VPNButtonBase {}
     visualStateItem: toggle
     state: VPNController.state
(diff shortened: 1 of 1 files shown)


lesleyjanenorton (approved):
Thanks!

src/ui/components/VPNToggle.qml
```diff
@@ -27,6 +27,8 @@ VPNButtonBase {
         return VPN.activate();
     }
 
+    onClicked: handleClick()
```
bakulf, line 30: This is a problem. I removed it because `handleClick` is executed twice when the user clicks on the button.

bakulf (changes requested):

src/ui/components/VPNToggle.qml
bakulf: Can you please check if this happens with your patch?

yzen: Just tried logging inside `handleClick()` and getting only 1 log message in app output. I believe it looks good.. Is the consequence of double `handleClick` is that the toggle bounces off-on-off? If so it definitely does not happen.

PR #264, opened 2020-11-19
Simplify the Accessible.name for VPNToggle and add a missing tooltip …

yzen:
…text for StateDisconnecting


lesleyjanenorton (approved):
Thanks @yzen !

PR #272, opened 2020-11-20
Fix #246 Mobile view logs link should not be styled as external links

### LINKED ISSUE #246: ViewLogs in mobile needs to use the '->' icon

lesleyjanenorton:


src/ui/components/VPNGetHelp.qml
```diff
@@ -32,9 +32,14 @@ Item {
         model: VPNHelpModel
 
         delegate: VPNExternalLinkListItem {
+            property var isMobileViewLogsLink: (window.fullscreenRequired() && id === 0)
```
bakulf, line 35: maybe we can add a boolean flag in the model to communicate when it's an external link.

bakulf: I can take this part.

bakulf (approved):

PR #275, opened 2020-11-20
Update upload/download legends a11y labelling.

yzen:


lesleyjanenorton (approved):

src/ui/components/VPNGraphLegendMarker.qml
```diff
@@ -43,10 +43,11 @@ Row {
     }
     Accessible.focusable: true
     Accessible.role: Accessible.StaticText
-    Accessible.name: markerLabel
-    Accessible.description: markerData
-    focus: true
-    activeFocusOnTab: true
+    //% "%1: %2 %3"
```
flodolo, line 46: Please always add comments to strings with placeables, localizers have no way to tell if they need to change the formatting otherwise.

Looking at this code I have no idea what's the reference. Can someone help and possibly fix it?

PR #276, opened 2020-11-21
Fix visibility (to assistive tech) of the interactive elements under …

### LINKED ISSUE #136: Make VPNConnectionInfo accessible.
Right now connection information pane is not accessible.
There are a couple of things we can do to make it more delightful:
- internal `ChartView` must have `Accessible.name` and `Accessible.role` that describes the graph.
- Label that presents IP information must have `Accessible.name: text`, this way it will be available to screen readers as static text. It must also be in the right accessible order. Right now it seems to not be following the `Close` button when moving virtual cursor forward (I suspect it's simply because of the order of components inside `VPNConnectionInfo`).
- `VPNGraphLegendMarker` should have `Accessible.role` set to `Accessible.StaticText`. The accessible order seems to be off  as well, so we need to make sure that order of components matches the layout order (I suspect). It also must have `Accessible.name` for the same reasons as the IP label. It should be comprised of the label, value and range like this (granted the id's are added to the inner Text components):
```
//% "%1: %2 %3"
Accessible.name: qsTrId("vpn.connectionInfo.accessibleName") // or something similar
    .arg(label.text)
    .arg(value.text)
    .arg(range.text)
```

yzen:
…the connection info popup when it is visible.

Fixes #136

### DIFF
--- a/src/ui/components/VPNControllerDevice.qml
+++ b/src/ui/components/VPNControllerDevice.qml
@@ -17,7 +17,7 @@ VPNClickableRow {
 
     accessibleName: titleText + " - " + label.text
     activeFocusOnTab: true
-
+    Accessible.ignored: rowShouldBeDisabled
 
     RowLayout {
         anchors.fill: parent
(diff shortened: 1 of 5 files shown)


lesleyjanenorton (approved):

lesleyjanenorton:
Thank you @yzen !

PR #277, opened 2020-11-21
Fix #226: Make LinkButton inherit from ButtonBase

### LINKED ISSUE #226: Keyboard focus improvements/follow up.
- Focus outline for Log Out should be red (similar to delete device button)
- Focus styling for Device List s sometimes missing:
  - Go to My Devices
  - Make sure you have an item that has a delete button
  - only tab (forward) from back button to list to delete as long as needed until you get to back button again
  - tab (forward) to the list, notice that the list is not highlighted focused, but it actually is..

stanleykerr:
# Summary
This fixes the accessibility highlighting display issue with VPNLinkButton, as seen in #226.

# Explanation
Modified VPNLinkButton to inherit from VPNButtonBase and added an extra "base style" property. This keeps the logic for buttons in the base class to allow for easier management.

These design choices allow to keep the current implementation of link buttons and how they're styled without majorly re-hauling how themes work.

I do have the question of whether enter/space should be used for "clicking" the button when focused. In the VPNButtonBase class, only enter is used for actually clicking the button, while VPNLinkButton, allows for either enter or space.

### DIFF
--- a/src/ui/components/VPNLinkButton.qml
+++ b/src/ui/components/VPNLinkButton.qml
@@ -7,35 +7,30 @@ import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 import "../themes/themes.js" as Theme
 
-RoundButton {
+VPNButtonBase {
     id: root
 
     required property var labelText
     property variant fontName: Theme.fontInterFamily
-    property var uiState:Theme.uiState
+    property var baseColor: Theme.linkButton
     property var linkColor: Theme.blueButton
 
-    signal clicked()
-
     radius: 4
-    focusPolicy: Qt.StrongFocus
+
     onFocusChanged: if (focus && typeof(ensureVisible) !== "undefined") ensureVisible(root)
     horizontalPadding: Theme.hSpacing
 
-    Keys.onPressed: {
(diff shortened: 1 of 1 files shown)


lesleyjanenorton:
Looks great - thank you @stanleykerr ! 

Re: click events on space/enter - If a components is clickable and has focus, space and return key clicks should always fire `clicked()` events. There are some gotchas with the default handling of these events across various components, which is why you see the keys being handled differently in a number of places. I think in this case, space key clicks trigger `clicked()` events by default.

lesleyjanenorton (approved):

src/ui/components/VPNLinkButton.qml
```diff
@@ -69,14 +64,24 @@ RoundButton {
         }
     ]
 
-    VPNMouseArea {
+    VPNUIStates {
+        colorScheme: root.baseColor
+        itemToFocus: root
 
+        VPNFocusBorder {
+            border.color: root.linkColor.focusBorder
```
lesleyjanenorton, line 72: Tiniest of nits and doesn't need to block but let's change this to `border.color: root.linkColor.defaultColor` so that the red border around the sign out links matches the red border around the red trashbin icon buttons in the devices view.

PR #282, opened 2020-11-21
Fix console warnings in VPNToggle {} and ViewOnboarding {}

lesleyjanenorton:


src/ui/components/VPNToggle.qml
```diff
@@ -286,7 +286,7 @@ VPNButtonBase {
         anchors.fill: toggle
         hoverEnabled: (VPNController.state === VPNController.StateOn || VPNController.state === VPNController.StateOff)
         cursorShape: Qt.PointingHandCursor
-        onClicked: clicked()
+        onClicked: handleClick()
```
bakulf, line 289: The issue here is that handleClick() is called twice: the mouse-area and the button-base (line 26~). Is there a way to avoid it?

bakulf (changes requested):

src/ui/components/VPNToggle.qml
lesleyjanenorton: yep- thank you for catching! fixed in d8429e7

src/ui/components/VPNToggle.qml
```diff
@@ -286,7 +286,6 @@ VPNButtonBase {
         anchors.fill: toggle
         hoverEnabled: (VPNController.state === VPNController.StateOn || VPNController.state === VPNController.StateOff)
         cursorShape: Qt.PointingHandCursor
-        onClicked: clicked()
```
bakulf, line 289: @yzen will not be happy about it. He just added this line :) 
See: https://github.com/mozilla-mobile/mozilla-vpn-client/commit/eeeabb3b5cd672c85b2471a647c9361725baa96e

lesleyjanenorton: @yzen Do we need this line? I'm able to connect/disconnect via keyboard without it.

bakulf:
I merge this PR. Let's open an issue if this breaks something.

src/ui/components/VPNToggle.qml
yzen: Catching up on this, looks like things are accessible with the Toggle, in onboarding (I tried with keyboard and VO on mac) 🎉

PR #301, opened 2020-11-25
Add VPNLoader.qml

lesleyjanenorton:
Adds VPNLoader {} and uses it for the sign in validation and subscription confirmation states.
Placeholder pending UX decisions requested in #302 
Screen Shot 2020-11-25 at 11 42 57 AM

### DIFF
--- a/src/ui/states/StateSubscriptionValidation.qml
+++ b/src/ui/states/StateSubscriptionValidation.qml
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import QtQuick 2.5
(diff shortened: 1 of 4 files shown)


src/ui/states/StateSubscriptionValidation.qml
```diff
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import QtQuick 2.5
+import "../components"
 
-Text {
-    text: "Hello world!"
+VPNLoader {
+    //% "Confirming subscription..."
```
flodolo, line 8: We should stick to the general UX guideliens, so `…` instead of `...`.

Issue #299, opened 2020-11-25
[Android] Local network access VPN feature does not work as expected


ValentinaPC:
#### Affected versions:
- 2;

#### Affected Platforms:
- Android 10;

#### Prerequisites:
- Have the VPN app installed;
- Have a FxA account with active subscription for Mozilla VPN;
- Have a functional router;

#### Steps to reproduce:
1. Open the VPN app;
2. Tap “Get Started”;
3. Finalize authentication;
4. Go to “Settings”->Network settings->”Local network access” and check the option.
5. Turn the VPN on;
6. Using your router IP address, check to see if you can access it;

#### Expected result:
-  The router homepage is loaded.

#### Actual result:
- The page returns a timeout error.

#### Notes:
- router homepage is loaded with success if the VPN is turned off.
- this feature behaves correctly on Windows.

ValentinaPC:
Verified this on the latest cross-plat build (from 11/30) using Android 10 and the issue is still reproducing following the above steps.

ValentinaPC:
Verified this as fixed on the latest cross-plat build (from 12/02) using Android 10.
The Local network access works as expected while selected from VPN app.

PR #308, opened 2020-11-27
Do not add Catch All IP range if LAN access is enabled

strseb:
- Fix for #299 

We always add 0.0.0.0/0 to the allowed ip - so local traffic is always routed through the vpn on android.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -149,7 +149,7 @@ void Controller::activate() {
   const Device* device = vpn->deviceModel()->currentDevice();
 
   const QList<IPAddressRange> allowedIPAddressRanges =
-      getAllowedIPAddressRanges();
+      getAllowedIPAddressRanges(server);
 
   Q_ASSERT(m_impl);
   m_impl->activate(server, device, vpn->keys(), allowedIPAddressRanges,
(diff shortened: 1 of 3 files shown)


src/controller.cpp
```diff
@@ -426,17 +426,10 @@ void Controller::captivePortalDetected() {
   deactivate();
 }
 
-QList<IPAddressRange> Controller::getAllowedIPAddressRanges() {
+QList<IPAddressRange> Controller::getAllowedIPAddressRanges(Server server) {
```
bakulf, line 429: do not duplicate the Server class. Do: `const Server& server`

src/controller.h
```diff
@@ -101,7 +101,7 @@ class Controller final : public QObject {
 
   bool processNextStep();
 
-  QList<IPAddressRange> getAllowedIPAddressRanges();
+  QList<IPAddressRange> getAllowedIPAddressRanges(Server server);
```
bakulf, line 104: here too, of course.

tests/unit/moccontroller.cpp
```diff
@@ -41,7 +41,8 @@ void Controller::statusUpdated(const QString&, uint64_t, uint64_t) {}
 
 void Controller::captivePortalDetected() {}
 
-QList<IPAddressRange> Controller::getAllowedIPAddressRanges() {
+QList<IPAddressRange> Controller::getAllowedIPAddressRanges(Server server) {
```
bakulf, line 44: here too.

bakulf (changes requested):

PR #309, opened 2020-11-27
Let's use a timer to animate the rings

bakulf:


lesleyjanenorton (approved):

src/ui/components/VPNAnimatedRings.qml
```diff
@@ -88,7 +90,13 @@ Rectangle {
 
             if (drawingRing3)
                 updateRing3();
+        }
 
+        Timer {
+            interval: 40
```
lesleyjanenorton, line 96: I think we may want to decrease this to 20

PR #315, opened 2020-11-29
Fix #90 - Add jobs argument for Qt compile script

### LINKED ISSUE #90: Use all the cores!
In scripts/qt5_compile, there is a `-j8` and in README a `-j4`. I am used to using `-j29` as I have 28 cores on my machine. Update README and provide an input to qt_compile script.

stanleykerr:
This allows the build script to be used as
```shell
./scripts/qt5_compile.sh -j 10 qt qt # before mandatory vars
./scripts/qt5_compile.sh qt qt -j 10 # after mandatory vars

# or with full name
./scripts/qt5_compile.sh --jobs 10 qt qt # before mandatory vars
./scripts/qt5_compile.sh qt qt --jobs 10 # after mandatory vars
```

### DIFF
--- a/scripts/qt5_compile.sh
+++ b/scripts/qt5_compile.sh
@@ -6,14 +6,42 @@
 
 . $(dirname $0)/commons.sh
 
+POSITIONAL=()
+JOBS=8
+
+helpFunction() {
+  print G "Usage:"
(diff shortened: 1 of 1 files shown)


bakulf:
Nice!

PR #319, opened 2020-11-30
Add GitHub action to check pull requests for l10n issues

flodolo:
What the script does is exporting strings to .ts, then:
- Check for misused character, i.e. apostrophes, quotes, ellipsis
- Check for missing comments when string has variables
- Check for empty strings

It support a list of exceptions per category via JSON file.


flodolo:
You can see the same code running [Checks](https://github.com/flodolo/mozilla-vpn-client/pull/2/checks) in this PR on my fork.

I've tried extending this to run also clang-format, but that requires more work and understanding of c++, so leaving that as a WIP for someone else to pick up later.

PR #331, opened 2020-12-01
Add clang-format action

### LINKED ISSUE #325: Run clang-format on pull requests
Filing mostly to document what I've tried on top of the l10n linter, in case someone wants to pick it up.

## clang-format GitHub Action

There's an https://github.com/marketplace/actions/clang-format-lint. I've tried running it, and it has way too many failures (about 4k lines).
https://github.com/flodolo/mozilla-vpn-client/pull/4/checks

## Use external Python wrapper.

I've found a [Python wrapper](https://github.com/Sarcasm/run-clang-format), which I had to modify slightly, because it doesn't seem to run correctly with `./` (`fnmatch` fails).

This seems better, but I still have a diff on a bunch of files
https://github.com/flodolo/mozilla-vpn-client/pull/3/checks

```
/src/logger.h
/src/loghandler.cpp
/src/mozillavpn.cpp
/src/platforms/macos/macosmenubar.cpp
/src/platforms/android/androidutils.cpp
/src/platforms/android/androidstartatbootwatcher.h
/src/platforms/windows/windowspingsendworker.cpp
/src/platforms/windows/windowscryptosettings.cpp
/src/platforms/windows/windowspingsendworker.h
```

## Other

As far as I understand:
- `clang-format` doesn't provide an easy way to check an entire folder (unless you pipe it to `find`).
- `.clang-format-ignore` is not natively supported.

But there might be still easier ways to get the same result without a complex Python script on top.

flodolo:
Fixes #325

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,3 +1,3 @@
 ./3rdparty/*
 ./android/modules/wireguard/*
-./src/hacl-star/.*
+./src/hacl-star/*
--- a/.github/workflows/l10n_lint.yaml
+++ b/.github/workflows/l10n_lint.yaml
@@ -1,4 +1,4 @@
-name: Check for l10n issues
+name: Linters (clang, l10n) 
 on:
   # Triggers the workflow on pull request events but only for the main branch
   pull_request:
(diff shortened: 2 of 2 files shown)


flodolo:
Checks were executed and passed 🎉

PR #332, opened 2020-12-01
Disable the animation in the chart - closes #269

### LINKED ISSUE #269: Traffic chart UX improvements
On iOS, the traffic chart animates rather unconventionally. It looks like an inchworm crawling. What I would have expected is a that the chart shifts to the left without changing it's "form" on the vertical axis.

RPReplayUNITO-UNDERSCORE!Final1605881919 2020-11-20 15!UNITO-UNDERSCORE!27!UNITO-UNDERSCORE!06!

Additionally, it's unclear what the numbers show.

* The IP address at the top, is that my real IP or the VPN IP?
* The download and upload "speeds", what do they show? I expected it's related to my current traffic, somehow in sync with the chart on the very right corner. But as you can see in the video above, they remain unchanged regardless of the traffic.

So, overall I find this screen rather confusing at the moment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-81)

bakulf:
Do you think this is enough?

### DIFF
--- a/src/ui/components/VPNConnectionInfo.qml
+++ b/src/ui/components/VPNConnectionInfo.qml
@@ -55,7 +55,7 @@ Popup {
             margins.bottom: 0
             margins.left: 0
             margins.right: 0
-            animationOptions: ChartView.SeriesAnimations
+            animationOptions: ChartView.NoAnimation
 
 
             ValueAxis {


lesleyjanenorton (approved):
R++ No more worm 🚫🐛

lesleyjanenorton:
I think it's definitely an improvement, easier to make sense of, and enough for now. I think we could still improve it in the future though.

PR #338, opened 2020-12-01
Accessibility paper cuts

lesleyjanenorton:


src/ui/components/VPNControllerView.qml
```diff
@@ -551,6 +551,8 @@ Rectangle {
         lineHeight: Theme.controllerInterLineHeight
         width: box.width - Theme.windowMargin * 3
         Accessible.ignored: connectionInfoVisible
+        Accessible.name: logoSubtitle.text
```
yzen, line 554: I don't think `Accessible.name` and `Accessible.role` is necessary if we use the text from the subtitle as a description for the title. I would only repalce:
```
-        Accessible.ignored: connectionInfoVisible
+        Accessible.ignored: true
```
and lets add `Accessible.description: logoSubtitle.text`  around L542

yzen (changes requested):
A quick change for title/subtitle and should be good to go.

yzen (approved):

PR #339, opened 2020-12-01
Add Google Play Referrer to the URI

### LINKED ISSUE #260: Add install attribution to mobile VPN clients
Attribution will be used to understand if acquisition efforts are successful, so that the marketing team can focus their time where new customers are finding value from the Mozilla VPN.

We have attribution today on the Windows client since the acquisition goes through the website. This is not totally the case for the mobile clients due to the app stores. 

On the v1.0 Android we included a library that passes through the Google Play store referral, but I am not sure if that is included in the new v2.0 cross platform clients.

See previous commit on adding the library to Android:

https://github.com/mozilla-mobile/guardian-vpn-android/commit/c02f0fa6e4c762df2c68b7e713136c5a4ea91a45

From previous investigations, it doesn't look like there is a similar on iOS.

strseb:
Closes #260

### DIFF
--- a/android/src/com/mozilla/vpn/VPNWebView.java
+++ b/android/src/com/mozilla/vpn/VPNWebView.java
@@ -8,12 +8,18 @@ import android.app.Activity;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.net.Uri;
+import android.os.RemoteException;
 import android.webkit.URLUtil;
 import android.webkit.WebSettings;
 import android.webkit.WebSettings.PluginState;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import android.util.Log;
+
+import com.android.installreferrer.api.InstallReferrerClient;
+import com.android.installreferrer.api.InstallReferrerStateListener;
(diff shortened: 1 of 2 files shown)


android/src/com/mozilla/vpn/VPNWebView.java
```diff
@@ -96,10 +105,35 @@ public void setUrl(final String url)
...
+        // Try to Get a Referrer and then Load the URL with it
+        m_referrer.startConnection(new InstallReferrerStateListener() {
+            @Override
+            public void onInstallReferrerSetupFinished(int responseCode) {
+                String referrerValue ="";
+                if( responseCode == InstallReferrerClient.InstallReferrerResponse.OK){
+                    try {
+                        ReferrerDetails response =  m_referrer.getInstallReferrer();
+                        referrerValue = "&" + response.getInstallReferrer();
+                        Log.v(TAG, "Recived - referrer: " + referrerString);
```
bakulf, line 117: Received

bakulf (approved):

PR #363, opened 2020-12-08
Update automation to QT 5.15.2

flodolo:
Just in time for the release of QT 6.

Also renaming the linter workflow, since it does more than just l10n linting.


bakulf:
Nice!

PR #372, opened 2020-12-09
Remove stackview.pop on requiredUpdate

### LINKED ISSUE #361: “Sign out” button from Update screen is not working [production]
#### Affected versions:
- last Android production build;

#### Affected Platforms:
- Android 10;

#### Prerequisites:
- Have the VPN app installed;
- Have a production FxA account with subscription;

#### Steps to reproduce:
1. Open the VPN app;
2. Tap “Get Started”;
3. Finalize authentication;
4. From the update screen, tap “Sign out” option;

#### Expected result:
-  User is signed out.

#### Actual result:
- Nothing happens;
ezgif com-video-to-gif(8)

strseb:
In case of the required update stackview was undefinded
fixes #361

### DIFF
--- a/src/ui/views/ViewUpdate.qml
+++ b/src/ui/views/ViewUpdate.qml
@@ -195,7 +195,6 @@ VPNFlickable {
             anchors.horizontalCenter: undefined
             Layout.alignment: Qt.AlignHCenter
             onClicked: {
-                mainStackView.pop(StackView.Immediate);
                 VPNController.logout();
             }
         }
--- a/src/ui/views/ViewUpdate.qml
+++ b/src/ui/views/ViewUpdate.qml
@@ -195,7 +195,7 @@ VPNFlickable {
             anchors.horizontalCenter: undefined
             Layout.alignment: Qt.AlignHCenter
             onClicked: {
-                stackview.pop(StackView.Immediate);
+                mainStackView.pop(StackView.Immediate);
                 VPNController.logout();
             }
         }


src/ui/views/ViewUpdate.qml
```diff
@@ -195,7 +195,7 @@ VPNFlickable {
             anchors.horizontalCenter: undefined
             Layout.alignment: Qt.AlignHCenter
             onClicked: {
-                stackview.pop(StackView.Immediate);
```
bakulf, line 198: why this? in VPNAlert we call `stackview.push()`

bakulf (changes requested):

src/ui/views/ViewUpdate.qml
strseb: Yes, for the "update Recommended" case when the user clicks on the notification. 
If i understand main.qml right , we're pushing this to the VPNStackView#mainStackView if we do `MozillaVPN::instance()->controller()->updateRequired();` right? 
```
main.qml:136
                State {
                    name: VPN.StateUpdateRequired

                    PropertyChanges {
                        target: loader
                        source: "states/StateUpdateRequired.qml"
                    }},
```

bakulf: I think you can simply remove this line. We show the 'logout' button only when the update is required:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/ui/views/ViewUpdate.qml#L46-L70

and that happens if the ViewUpdate.qml is loaded from the main.qml directly:
main.qml -> state: VPN.StateUpdateRequired -> states/SateUpdateRequired.qml -> views/ViewUpdate.qml

so there is no `stackView` involved at all.

PR #387, opened 2020-12-11
Force use of mullvad dns on android 

strseb:
Just noticed mullvad/check said we're not using the mullvad dns on android.

### DIFF
--- a/android/src/com/mozilla/vpn/VPNServiceBinder.kt
+++ b/android/src/com/mozilla/vpn/VPNServiceBinder.kt
@@ -246,7 +246,7 @@ class VPNServiceBinder(service: VPNService) : Binder() {
         ifaceBuilder.parsePrivateKey(privateKey)
         ifaceBuilder.addAddress(InetNetwork.parse(jDevice.getString("ipv4Address")))
         ifaceBuilder.addAddress(InetNetwork.parse(jDevice.getString("ipv6Address")))
-        
+        ifaceBuilder.addDnsServer(InetNetwork.parse(jServer.getString("ipv4Gateway")).address)
         val jExcludedApplication = obj.getJSONArray("excludedApps");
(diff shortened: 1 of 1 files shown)


android/src/com/mozilla/vpn/VPNServiceBinder.kt
```diff
@@ -246,7 +253,10 @@ class VPNServiceBinder(service: VPNService) : Binder() {
         ifaceBuilder.parsePrivateKey(privateKey)
         ifaceBuilder.addAddress(InetNetwork.parse(jDevice.getString("ipv4Address")))
         ifaceBuilder.addAddress(InetNetwork.parse(jDevice.getString("ipv6Address")))
-        
+
+        ifaceBuilder.addDnsServer(DNS.mullvadVpnInternal)
```
bakulf, line 257: Don't do that. Instead use: jServer.getString("ipv4Gateway"). That corresponds to Server::Ipv4Gateway()

android/src/com/mozilla/vpn/VPNServiceBinder.kt
```diff
@@ -246,7 +253,10 @@ class VPNServiceBinder(service: VPNService) : Binder() {
...
+        ifaceBuilder.addDnsServer(DNS.mullvadPublic);
```
bakulf, line 258: 1 DNS is enough. We set it elsewhere too.
Btw, is this a bug in the existing android build too?

bakulf (changes requested):

android/src/com/mozilla/vpn/VPNServiceBinder.kt
strseb: no we do ` addDnsServer(InetNetwork.parse(serverInfo.server.ipv4Gateway).address)` there

PR #389, opened 2020-12-11
WIP: Detect too old WebViews

strseb:
This is an idea i came up to handle #344. 
Even though i think it's unlikely that a user on a real device that has the play store will encounter this, a fallback is always good.
So what i try to do here is to detect old webviews and then use the DesktopAuth. 
In that case we would open the users default browser instead of a webview, which hopefully is way newer. 
Currently the user would need to navigate back to the app by hand, but still this is way better then not beeing able to login in at all.

### DIFF
--- a/src/platforms/android/androidauthenticationlistener.cpp
+++ b/src/platforms/android/androidauthenticationlistener.cpp
@@ -8,6 +8,11 @@
 #include "mozillavpn.h"
 #include "platforms/android/androidutils.h"
 
+#include <QAndroidJniObject>
+#include <QtAndroid>
+#include <jni.h>
+#include "tasks/authenticate/desktopauthenticationlistener.h"
+
 namespace {
 Logger logger(LOG_ANDROID, "AndroidAuthenticationListener");
 }
@@ -15,8 +20,7 @@ Logger logger(LOG_ANDROID, "AndroidAuthenticationListener");
 AndroidAuthenticationListener::AndroidAuthenticationListener(QObject* parent)
     : AuthenticationListener(parent) {
   MVPN_COUNT_CTOR(AndroidAuthenticationListener);
-
-  logger.log() << "Android authentication listener";
+  logger.log() << "Android authentication litener";
 }
 
 AndroidAuthenticationListener::~AndroidAuthenticationListener() {
@@ -31,5 +35,21 @@ void AndroidAuthenticationListener::start(MozillaVPN* vpn, QUrl& url,
 
   url.setQuery(query);
 
-  AndroidUtils::instance()->startAuthentication(this, url);
+  QAndroidJniObject activity = QtAndroid::androidActivity();
+  jboolean supported = QAndroidJniObject::callStaticMethod<jboolean>(
+      "com/mozilla/vpn/PackageManagerHelper", "isWebViewSupported",
+      "(Landroid/content/Context;)Z", activity.object());
+  if (supported) {
+    AndroidUtils::instance()->startAuthentication(this, url);
+    return;
+  }
+  m_legacyAuth = new DesktopAuthenticationListener(this);
+  m_legacyAuth->start(vpn, url, query);
+
+  connect(m_legacyAuth, &AuthenticationListener::completed,
+          [this](const QString& code) { emit this->completed(code); });
+  connect(m_legacyAuth, &AuthenticationListener::failed,
+          [this](const ErrorHandler::ErrorType error) {
+            emit this->failed(error);
+          });
 }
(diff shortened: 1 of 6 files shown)


src/platforms/android/androidauthenticationlistener.cpp
```diff
@@ -8,15 +8,19 @@
 #include "mozillavpn.h"
 #include "platforms/android/androidutils.h"
 
+#include <QAndroidJniObject>
+#include <QtAndroid>
+#include <jni.h>
+#include "tasks/authenticate/desktopauthenticationlistener.h"
```
bakulf, line 14: move this to the rest of the local headers

src/platforms/android/androidauthenticationlistener.cpp
```diff
@@ -31,5 +35,21 @@ void AndroidAuthenticationListener::start(MozillaVPN* vpn, QUrl& url,
...
-  AndroidUtils::instance()->startAuthentication(this, url);
+  QAndroidJniObject activity = QtAndroid::androidActivity();
+  jboolean supported = QAndroidJniObject::callStaticMethod<jboolean>(
+      "com/mozilla/vpn/PackageManagerHelper", "isWebViewSupported",
+      "(Landroid/content/Context;)Z", activity.object());
+  if (supported) {
+    AndroidUtils::instance()->startAuthentication(this, url);
+    return;
+  }
+  m_legacyAuth = new DesktopAuthenticationListener(this);
```
bakulf, line 46: Well, this doesn't fix the issue, right? We are just moving the problem elsewhere.
Of course, if the user has another default browser, the operation _could_ work.

src/platforms/android/androidauthenticationlistener.cpp
```diff
@@ -31,5 +35,21 @@ void AndroidAuthenticationListener::start(MozillaVPN* vpn, QUrl& url,
...
+  m_legacyAuth->start(vpn, url, query);
+
+  connect(m_legacyAuth, &AuthenticationListener::completed,
```
bakulf, line 49:
```
connect(m_legacyAuth, &AuthenticationListener::completed, this, &AndroidAuthenticationListener::completed);
connect(m_legacyAuth, &AuthenticationListener::failed, this, &AndroidAuthenticationListener::failed);
```

src/platforms/android/androidauthenticationlistener.cpp
bakulf, line 46: No need to store m_legacyAuth in the object. Just do:

```
DesktopAuhenticationListener* legacyAuth = new DesktopAuthenticationListener(this);
connect...

File: src/platforms/android/androidauthenticationlistener.h [hunk 4]
```diff
@@ -17,6 +18,9 @@ class AndroidAuthenticationListener final : public AuthenticationListener {
   ~AndroidAuthenticationListener();
 
   void start(MozillaVPN* vpn, QUrl& url, QUrlQuery& query) override;
+
+ private:
+  DesktopAuthenticationListener* m_legacyAuth;
```
bakulf, line 23: Remove this

src/platforms/android/androidauthenticationlistener.h
```diff
@@ -6,6 +6,7 @@
 #define ANDROIDAUTHENTICATIONLISTENER_H
 
 #include "authenticationlistener.h"
+#include "tasks/authenticate/desktopauthenticationlistener.h"
```
bakulf, line 9: Remove this.

bakulf (changes requested):

src/platforms/android/androidauthenticationlistener.cpp
strseb: Indeed, if the default browser is too old this wont work. But the user (and support) can help themselves here as they can just switch default browsers.
- Note the old client does not have any checks of this sort, it's just uses the installed firefox( if available ) or google chrome.

src/platforms/android/androidauthenticationlistener.cpp
```diff
@@ -15,8 +20,7 @@ Logger logger(LOG_ANDROID, "AndroidAuthenticationListener");
 AndroidAuthenticationListener::AndroidAuthenticationListener(QObject* parent)
     : AuthenticationListener(parent) {
   MVPN_COUNT_CTOR(AndroidAuthenticationListener);
-
-  logger.log() << "Android authentication listener";
+  logger.log() << "Android authentication litener";
```
bakulf, line 23: typo

bakulf (approved):

PR #397, opened 2020-12-16
Improve AppPermission Functionality

strseb:
I've tried to make the AppPermission View to function more of what we currently have in the current android app. Currently we just show a plain list of all apps. 
I've added: 
- Split list ( Protected / Unprotected)
- Made Lists Expandable/Collapsible
- Add a unblock/block all button

Let me know what you think :) 

Screenshot_20201216-164825

### DIFF
--- a/src/ui/components/VPNExpandableAppList.qml
+++ b/src/ui/components/VPNExpandableAppList.qml
@@ -1,0 +1,140 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+import QtQuick 2.5
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import QtGraphicalEffects 1.15
+import Mozilla.VPN 1.0
+
+VPNClickableRow {
+    id: appRow
+
+    property var listVisible : true
+    property var listModel: undefined
+    property var header: ""
+    property var description: ""
+
+    state: listVisible ? "list-visible" : "list-hidden"
+
+    Keys.onReleased: if (event.key === Qt.Key_Space) handleKeyClick()
+    handleMouseClick: function() { listVisible = !listVisible; }
+    handleKeyClick: function() { listVisible = !listVisible; }
+    clip: true
+
+
+    Behavior on y {
+        PropertyAnimation {
+            duration: 200
+        }
+    }
+
+    accessibleName: name
+
+    states: [
+        State {
+            name: "list-hidden"
+
+            PropertyChanges {
+                target: appRow
+                height: appRow.height
+            }
+
+            PropertyChanges {
+                target: applist
+                opacity: 0
+            }
+            PropertyChanges {
+                target: toggleArrow
+                rotation: -90
+
+            }
+
+        },
+        State {
+            name: "list-visible"
+
+            PropertyChanges {
+                target: appRow
+                height: appRow.height +applist.height
+            }
+
+            PropertyChanges {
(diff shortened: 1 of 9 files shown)


lesleyjanenorton:
Nice!! Will dig into this first thing in the morning but looks great so far!

src/ui/components/VPNExpandableAppList.qml
```diff
@@ -0,0 +1,143 @@
...
+    ]
+
+
+    RowLayout {
+        id: appRowHeader
+        anchors.top: parent.top
+        anchors.left: parent.left
+        width: appRow.width
+        spacing: 0
+        VPNServerListToggle {
```
lesleyjanenorton, line 70: Getting a `serverCountry is not defined` error here from the `serverCountry` reference on line 41 of VPNServerListToggle.qml

src/ui/components/VPNExpandableAppList.qml
```diff
@@ -0,0 +1,143 @@
...
+            id: toggleArrow
+            Layout.leftMargin: 15 - enabledAppRow.anchors.leftMargin
```
lesleyjanenorton, line 72: Should this just be `appRow.anchors...` ?

src/ui/components/VPNExpandableAppList.qml
```diff
@@ -0,0 +1,143 @@
...
+
+            RowLayout {
+                Layout.alignment: Qt.AlignLeft
+                Layout.fillWidth: true
+
+                VPNInterLabel {
+                    id: label
+                    text: header
+                    Accessible.role: Accessible.Heading
+                    color: Theme.fontColorDark
```
lesleyjanenorton, line 88: Needs `import "../themes/themes.js" as Theme` on line 11

src/ui/components/VPNExpandableAppList.qml
```diff
@@ -0,0 +1,143 @@
...
+        delegate: VPNCheckBoxRow {
+            labelText: appName
+            subLabelText: appID
+            isChecked: appIsEnabled
+            isEnabled: appRow.isEnabled
+            showDivider:false
+            onClicked: VPNAppPermissions.flip(appID)
+            visible: true
+            width: applist.width
+            anchors.left: applist.left
```
lesleyjanenorton, line 137: `qrc:/ui/components/VPNExpandableAppList.qml:128:19: QML VPNCheckBoxRow: Cannot anchor to an item that isn't a parent or sibling. qrc:/ui/components/VPNExpandableAppList.qml: 128`

A strange ListView gotcha: the parent these delegates is actually `applist.contentItem`, not `applist` itself.  This just needs to be `anchors.left: parent.left`.

src/ui/components/VPNExpandableAppList.qml
```diff
@@ -0,0 +1,143 @@
...
+
+        delegate: VPNCheckBoxRow {
+            labelText: appName
+            subLabelText: appID
+            isChecked: appIsEnabled
+            isEnabled: appRow.isEnabled
+            showDivider:false
+            onClicked: VPNAppPermissions.flip(appID)
+            visible: true
+            width: applist.width
```
lesleyjanenorton, line 136: should this be `width: applist.width - leftMargin`  ?

lesleyjanenorton (commented):
Just a few console warnings but overall this looks good! 


One idea, doesn't need to hold this up: Right now it looks like it's still possible to open and close empty lists. Do we want to disable opening/closing of empty lists and hide the list toggles ?

flodolo:
What happens to the layout if "Unprotect all" is extremely long? I expect that to be the common case, since I doubt many language will have the equivalent of un+protect.

You should be able to test with Italian.

PR #403, opened 2020-12-23
Fix localization comments for speed strings

flodolo:


flodolo:
Looks like there are c-lang formatting issues, not sure if I should fix them.

bakulf:
Thanks!

PR #422, opened 2021-01-05
Fix #116 -Improved scrolling to current server city

### LINKED ISSUE #116: Show the selected city when opening the server list
A few issues here:
1. Right now, when I open the list of servers, I don't see the selected city by default. Would be great to open the country and show the 'blue' selected city immediately.
2. If the city is not visible because we need to scroll, the app should scroll by default to the right place.

@lesleyjanenorton thoughts?

lesleyjanenorton:


src/ui/views/ViewServers.qml
```diff
@@ -68,12 +68,20 @@ Item {
...
-                    if (citiesList.length >= 2) {
-                        const delegateSpacing = Theme.rowHeight * 2
-                        serverList.contentY += delegateSpacing + (selectedCityIndex * delegateSpacing);
-                    }
+                    const currentCountryDistanceFromTop = (serverList.currentItem.mapToItem(null, 0, 0).y);
+                    const listVerticalCenter = serverList.height / 2
+
+                    const citySublist = delegateModel.items.get(idx).model.cities;
+                    const currentCityIdx = citySublist.indexOf(VPNCurrentServer.city);
+                    const cityItemHeight = Theme.rowHeight;
```
bakulf, line 76: This const is used just once. Maybe you can use Theme.rowHeight directly.

bakulf (approved):

PR #436, opened 2021-01-08
Public key to identify the current device instead of the name

bakulf:
Let's use the public key to identify the current device instead of the name,


strseb (approved):
Nice!

PR #440, opened 2021-01-11
Pass Tunnel State changes to QT

### LINKED ISSUE #439: Android: Disabling VPN in System Settings does not disable VPN in app
Steps
- Launch the app and sign in with a valid account
- Start/Stop any VPN (for permissions, etc)
- Go to Android Settings -> Network & Internet -> VPN -> Mozilla VPN
- Enable Always On VPN 
- The app now shows connecting / connected
- Go back and disable Always On VPN

Expected
- The VPN should be disconnected and also the UI should show this.

Actual
- The VPN gets disconnected but the app ui still shows "connected"

strseb:
Fixes #439
It might be that the OS kills our vpn connection, e.g. when always-on vpn gets set to false. This was not propagated to the front end.

### DIFF
--- a/android/src/com/mozilla/vpn/VPNTunnel.kt
+++ b/android/src/com/mozilla/vpn/VPNTunnel.kt
@@ -5,17 +5,21 @@
 package com.mozilla.vpn
 
 import com.wireguard.android.backend.Tunnel
-
+import org.mozilla.firefox.vpn.VPNService
 class VPNTunnel: Tunnel{
     val mName:String;
-    constructor(name:String){
+    val mService:VPNService;
+
+    constructor(name:String, s: VPNService ){
         this.mName = name;
+        this.mService =s;
     }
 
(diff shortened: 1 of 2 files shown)


android/src/com/mozilla/vpn/VPNTunnel.kt
```diff
@@ -5,17 +5,21 @@
...
-
+import org.mozilla.firefox.vpn.VPNService
 class VPNTunnel: Tunnel{
     val mName:String;
-    constructor(name:String){
+    val mService:VPNService;
+
+    constructor(name:String, s: VPNService ){
         this.mName = name;
+        this.mService =s;
```
bakulf, line 15: All good, but fix the indentation. Do we have a clang-format for kt?

bakulf (commented):
All good, but do not merge it yet.

PR #441, opened 2021-01-11
Submodule updates

birdsarah:


bakulf:
We do update modules at the latest tag. The current HEAD of those repos is not always compatible with MozillaVPN.

So far we use:
- https://github.com/WireGuard/wireguard-tools/tree/v1.0.20200827
- https://github.com/WireGuard/wireguard-go/tree/v0.0.20201118
- https://github.com/WireGuard/wireguard-apple/tree/0.0.20200127-17

PR #451, opened 2021-01-13
Use our own copy of wg-quick script

birdsarah:
@bakulf as you suggested, this PR is to be using our own copy of [linux.bash](https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash). After this PR is in we can slowly start chipping away at the functionality till we can remove it altogether.

There's a lot about this that probably isn't idiomatic C++ so welcome lots of code review. Notes:
* I have left a couple of "TODO" comments that are questions.
* Should we change the copyright header on the linux.bash?

### DIFF
--- a/src/platforms/linux/daemon/dbusservice.cpp
+++ b/src/platforms/linux/daemon/dbusservice.cpp
@@ -13,6 +13,7 @@
 #include <QCoreApplication>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QFile>
 
 #if defined(__cplusplus)
 extern "C" {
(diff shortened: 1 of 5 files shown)


src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +131,68 @@ QString DBusService::getLogs() {
...
+    config.m_ipv6Enabled);
+
+  if (confFile == WgQuickProcess::FAIL) {
+    return false;
+  }
+
+  // Make our copy of the bash script
+
+  // Read in bash file and write to tmp so can execute it.
+  // TODO: This seems cumbersome, but I didn't know a better way to do it
```
bakulf, line 155: Are you doing this because the script is included as a resource? What about if we install it as it is?

Here is an idea:
- in src.pro, only for Linux, we can add:
```
helperFile.path = /usr/share/mozillavpn
helperFile.files = ../linux/extra/helper.sh
INSTALLS += helperFile
```
- WgQuickProcess uses `/usr/share/mozillavpn/helper.sh` instead of `wg-quick`.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +131,68 @@ QString DBusService::getLogs() {
...
-      op, config.m_privateKey, config.m_deviceIpv4Address,
-      config.m_deviceIpv6Address, config.m_serverIpv4Gateway,
-      config.m_serverIpv6Gateway, config.m_serverPublicKey,
-      config.m_serverIpv4AddrIn, config.m_serverIpv6AddrIn,
-      config.m_allowedIPAddressRanges.join(", "), config.m_serverPort,
-      config.m_ipv6Enabled);
+
+  // Make the config file
+
+  QTemporaryDir tmpDir;
```
bakulf, line 137:
```
if (!tmpDir.isValid()) {
  logger.log() << "Something";
  return false;
}
```

src/wgquickprocess.cpp
```diff
@@ -12,40 +12,41 @@
...
-#  error Unsupported platform
-#endif
 }
 
 }  // namespace
 
-// static
-bool WgQuickProcess::run(
-    Daemon::Op op, const QString& privateKey, const QString& deviceIpv4Address,
+QString const WgQuickProcess::FAIL = "FAIL";
```
bakulf, line 32: In general, as a convention, you can use an empty string too.
In this way, the caller of writeWgConfigFile can do:

```
QString fileName = WgQuickProcess::writeConfigFile(...);
if (fileName.isEmpty()) {
  logger.log() << "Something went wrong";
}
```

but this the FAIL const approach works too!

bakulf (changes requested):
I suspect we can simplify this PR, having the script as an installed resource (see the comments).
This PR is on top of the submodule code, right? For the next time, do you mind using separate branches? Thanks!

PR #455, opened 2021-01-13
Android

bakulf:


strseb (approved):
Thats awesome!
Checked; apks all have the right package name and are pointing to the right server.

PR #462, opened 2021-01-13
let Gradle use Version.pri for Android Version

### LINKED ISSUE #448: Keep android/bundle.gradle and version.pri in sync
So far, the version number is written in 2 separate files: `android/bundle.grandle` for android builds, `version.pri` for anything else. We need to keep these 2 files in sync with a script to avoid mistakes.

strseb:
Closes #448

### DIFF
--- a/scripts/android_package.sh
+++ b/scripts/android_package.sh
@@ -117,10 +117,10 @@ print Y "Importing translation files..."
 python3 scripts/importLanguages.py $([[ "$PROD" ]] && echo "-p" || echo "") || die "Failed to import languages"
 
 printn Y "Computing the version... "
-SHORTVERSION=$(cat version.pri | grep VERSION | grep defined | cut -d= -f2 | tr -d \ )
+export SHORTVERSION=$(cat version.pri | grep VERSION | grep defined | cut -d= -f2 | tr -d \ ) # Export so gradle can pick it up
+export VERSIONCODE=$(echo $SHORTVERSION| tr -d .)"0"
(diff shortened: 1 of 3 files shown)


scripts/android_package.sh
```diff
@@ -151,7 +151,6 @@ if [[ "$RELEASE" ]]; then
     CONFIG-=debug_and_release \
     CONFIG+=release \
     $PRODMODE \
-    ANDROID_ABIS="armeabi-v7a x86 arm64-v8a" \
```
strseb, line 154: While i was at it, we have those defined in the qt project file, no need to have them here.

scripts/android_package.sh
```diff
@@ -117,10 +117,10 @@ print Y "Importing translation files..."
 python3 scripts/importLanguages.py $([[ "$PROD" ]] && echo "-p" || echo "") || die "Failed to import languages"
 
 printn Y "Computing the version... "
-SHORTVERSION=$(cat version.pri | grep VERSION | grep defined | cut -d= -f2 | tr -d \ )
+export SHORTVERSION=$(cat version.pri | grep VERSION | grep defined | cut -d= -f2 | tr -d \ ) # Export so gradle can pick it up
+export VERSIONCODE=$(echo $SHORTVERSION| tr -d .)"0"
```
bakulf, line 121: This is OK if you add a comment in version.pri saying something like:
"version numbers must be <= 9!" or a single-digit number... 

otherwise, version=2.0.12 will generate a "20120" and that is so wrong.

bakulf (changes requested):

PR #463, opened 2021-01-14
Use our own copy of wg-quick script on linux

birdsarah:
* Re-organizes wgquickprocess slightly for future convenience
* Slight consistency clean-up in src.pro


src/src.pro
```diff
@@ -319,18 +319,22 @@ else:linux:!android {
...
     dbus_conf.path = $${USRPATH}/share/dbus-1/system.d/
+    dbus_conf.files = platforms/linux/daemon/org.mozilla.vpn.conf
     INSTALLS += dbus_conf
 
-    dbus_service.files = platforms/linux/daemon/org.mozilla.vpn.dbus.service
     dbus_service.path = $${USRPATH}/share/dbus-1/system-services
+    dbus_service.files = platforms/linux/daemon/org.mozilla.vpn.dbus.service
     INSTALLS += dbus_service
 
+    wg_helper.path = $${USRPATH}/share/
```
bakulf, line 334: $${USERPATH}/share/mozillavpn ? Usually applicaiton files are stored into a folder.

src/wgquickprocess.cpp
```diff
@@ -95,30 +110,55 @@ bool WgQuickProcess::run(
...
-    return false;
   }
-
   QDir dir(tmpDir.path());
   QFile file(dir.filePath(QString("%1.conf").arg(WG_INTERFACE)));
+
   if (!file.open(QIODevice::ReadWrite)) {
     qWarning("Unable to create a file in the temporary folder");
-    return false;
+    return "";
```
bakulf, line 121: QString()

src/wgquickprocess.cpp
```diff
@@ -95,30 +110,55 @@ bool WgQuickProcess::run(
...
   }
 
   qint64 written = file.write(content);
+
   if (written != content.length()) {
     qWarning("Unable to write the whole configuration file");
+    return "";
```
bakulf, line 128: QString()

src/wgquickprocess.cpp
```diff
@@ -20,10 +20,13 @@ Logger logger(
     "WgQuickProcess");
 
 QString scriptPath() {
+  QDir appPath(QCoreApplication::applicationDirPath());
```
bakulf, line 23: What about if in `src.pro` we add:
```
DEFINES += MVPN_DATA_PATH=\\\"$${USRPATH}/share/mozillavpn\\\"
```

And here we do:
```
QDir appPath(MVPN_DATA_PATH);
return appPath.filePath("helper.sh");
```

src/wgquickprocess.cpp
```diff
@@ -35,18 +38,18 @@ QString scriptPath() {
 
 }  // namespace
 
-// static
-bool WgQuickProcess::run(
-    Daemon::Op op, const QString& privateKey, const QString& deviceIpv4Address,
+bool WgQuickProcess::validateWgArgs(
```
bakulf, line 41: I love this method! Because we can extend it (not now, but in a separate PR)!
For instance, we can replace `VALIDATE` with a proper validation such as:

- keys must have a precise length
- ipv4 addresses must have a precise format
- ipv6 addresses too
- etc

src/wgquickprocess.cpp
```diff
@@ -95,30 +110,55 @@ bool WgQuickProcess::run(
...
+  if (!validateWgArgs(privateKey, deviceIpv4Address, deviceIpv6Address,
+                      serverIpv4Gateway, serverIpv6Gateway, serverPublicKey,
+                      serverIpv4AddrIn, serverIpv6AddrIn,
+                      allowedIPAddressRanges, serverPort, ipv6Enabled)) {
+    qWarning("Config arguments are not valid.");
     return false;
   }
 
-  file.close();
+  QTemporaryDir tmpDir;
```
bakulf, line 149:
```
if (!tmpDir.isValid()) {
  message here
  return false;
}
```

src/wgquickprocess.h
```diff
@@ -8,13 +8,27 @@
 #include "daemon.h"
 
 #include <QObject>
+#include <QTemporaryDir>
```
bakulf, line 11: Just do a forwarding declaration:

```
class QTemporaryDir;
```

src/wgquickprocess.cpp
```diff
@@ -95,30 +110,55 @@ bool WgQuickProcess::run(
   content.append(
       QString("\nAllowedIPs = %1\n").arg(allowedIPAddressRanges).toUtf8());
 
-  QTemporaryDir tmpDir;
   if (!tmpDir.isValid()) {
```
bakulf, line 113: move this next to the `QTemporaryDir` creation

src/wgquickprocess.cpp
```diff
@@ -95,30 +110,55 @@ bool WgQuickProcess::run(
...
     qWarning("Cannot create a temporary directory");
```
bakulf, line 114: `return QString()` ?

bakulf (changes requested):

src/src.pro
birdsarah: Fixed in 8cd35c2

src/wgquickprocess.cpp
birdsarah: Yes! I started adding tests, but then decided not to get ahead of myself as we may not need a validation method like this when we are calling different methods / classes to handle different wg pieces.

birdsarah:
@bakulf. Thanks for the review. I've made the changes. All made perfect sense apart from "forwarding declaration." This is not a concept I'm familiar with. I'll ping you on slack to learn more.

src/wgquickprocess.cpp
birdsarah: Given that this helper file will be going away shortly I don't think we need to.

bakulf (approved):

src/wgquickprocess.cpp
birdsarah: As I was doing other work, I noticed the parseConfig method in the daemon here: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/daemon/daemon.cpp#L90. It seems to me like we don't need this additional validation given that config is validated there and we pass in everything from config. Any thoughts?

src/platforms/ios/iaphandler.mm
```diff
@@ -381,17 +381,17 @@ - (void)request:(SKRequest*)request didFailWithError:(NSError*)error {
 
   connect(request, &NetworkRequest::requestCompleted,
           [this, ids](QNetworkReply*, const QByteArray&) {
-    logger.log() << "Purchase request completed";
-    SettingsHolder::instance()->addSubscriptionTransactions(ids);
+            logger.log() << "Purchase request completed";
```
birdsarah, line 384: Had a gross merge and this ended up being included.

src/platforms/wasm/wasmwindowcontroller.cpp
```diff
@@ -21,7 +21,7 @@ namespace {
 Logger logger(LOG_MAIN, "WasmWindowController");
 
 WasmWindowController* s_instance = nullptr;
-}
+}  // namespace
```
birdsarah, line 24: Again, gross merge meant this got added.

src/wgquickprocess.cpp
```diff
@@ -22,10 +22,14 @@ Logger logger(
...
 
 QString scriptPath() {
+  QDir appPath(QCoreApplication::applicationDirPath());
 #if defined(MVPN_LINUX)
-  return "wg-quick";
+  appPath.cdUp();
+  appPath.cd("share");
+  appPath.cd("mozillavpn");
+  return appPath.filePath("wghelper.sh");
+  ;
```
birdsarah, line 31: this shouldn't be here

birdsarah:
@bakulf merge whenever you like. We can pick up the thread about validateArgs later. Not sure if we always do it, but would definitely suggest a squash merge as this is a bit of a messy PR.

src/wgquickprocess.cpp
birdsarah: Fixed in 1acbb8f

birdsarah:
@bakulf - this can be merged, I updated the linux path location as you suggested.

src/platforms/ios/iaphandler.mm
bakulf: Can you revert these changes?

src/wgquickprocess.cpp
```diff
@@ -37,18 +38,18 @@ QString scriptPath() {
...
-    const QString& configFile, const QString& privateKey,
-    const QString& deviceIpv4Address, const QString& deviceIpv6Address,
-    const QString& serverIpv4Gateway, const QString& serverIpv6Gateway,
-    const QString& serverPublicKey, const QString& serverIpv4AddrIn,
-    const QString& serverIpv6AddrIn, const QString& allowedIPAddressRanges,
-    int serverPort, bool ipv6Enabled) {
-  Q_UNUSED(serverIpv6AddrIn);
+bool WgQuickProcess::validateWgArgs(
+    const QString& privateKey, const QString& deviceIpv4Address,
+    const QString& deviceIpv6Address, const QString& serverIpv4Gateway,
```
bakulf, line 43: This change would break windows build. On windows, we call this: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/platforms/windows/daemon/windowsdaemon.cpp#L418
...passing the file name. The file is overwritten each time.

bakulf (changes requested):
There is an issue with windows. Can you take a look?

PR #466, opened 2021-01-14
Support QT 5.14

bakulf:


src/ui/components/VPNServerCountry.qml
```diff
@@ -4,8 +4,8 @@
 
 import QtQuick 2.0
 import QtQuick 2.5
```
lesleyjanenorton, line 6: Not sure it matters, but I don't think we need both QtQuick imports?

lesleyjanenorton (approved):

strseb (approved):

PR #467, opened 2021-01-14
MacOS builds from github actions

bakulf:
The result of this action is a non-signed production/staging build.

### DIFF
--- a/.github/workflows/macos-build.yaml
+++ b/.github/workflows/macos-build.yaml
@@ -1,0 +1,171 @@
+name: MacOS build
+
+on:
+  push:
+    branches:
+     - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  macos-staging:
+    runs-on: macos-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Checkout submodules
+        shell: bash
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+      - name: Install Qt
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-build.yaml
```diff
@@ -0,0 +1,171 @@
...
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+      - name: Install Qt
+        shell: bash
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git clone https://github.com/bakulf/qt_static_macos
```
bhearsum, line 30: is there a reason we're pulling this from a your repo instead of an authoritative source?

.github/workflows/macos-build.yaml
```diff
@@ -0,0 +1,171 @@
...
+            "MozillaVPN.xcodeproj" \
+            "$SHORTVERSION" \
+            "$FULLVERSION" \
+            macos
+
+      - name: Compiling
+        shell: bash
+        run: xcodebuild build CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -project MozillaVPN.xcodeproj
+
+      - name: Upload APK
```
bhearsum, line 86: typo, this should not be `apk` :)

bhearsum (approved):
This is mostly a rubberstamp, but the bits I'm most familiar with seem OK. A couple minor things below. Up to you whether or not you want to change the QT stuff.

.github/workflows/macos-build.yaml
bakulf: Because we need a static version of the QT compiled. So far, nobody has exposed the right build.

bhearsum: There isn't a tarball we can pull from somewhere? Or are those binaries in your repo?

Issue #445, opened 2021-01-12
Server list animation issues when VPN is on


lesleyjanenorton:
Details to follow...

PR #477, opened 2021-01-18
Stop the animated-circles when not visible or navigating

strseb:
Hello! About #445 - I've had the idea that we might freeze the ring-animation when we navigate to another page and wait a bit when navigating back.
So devices that are already struggling with the ring animation (Like my pixel) have GPU time for the navigation animations. That makes it way smother to navigate to languages/devices while connected
Let me know what you think :)


lesleyjanenorton (approved):
Looks great! Thank you!

PR #503, opened 2021-01-25
Add linked OpenSSL

strseb:
We also need to specify that we want openssl linked. `By default, an SSL-enabled Qt library dynamically loads any installed OpenSSL library at run-time.`
This is probably why ssl only works in your git shell. With this flag, explorer and cmd also run fine :)


bakulf:
Great! I merged this PR into my windows one.

Issue #380, opened 2020-12-10
let's remove the 'restore-subscription' feature


bakulf:
The IAP view is shown after the authentication and only if the current user doesn't have a subscription.
Because of this, there are really needs to 'restore' a previous subscription because that 'restore' is done at a server level.

If what I'm writing here is right, we can probably drop this feature.

chrismore:
@bakulf are you saying we don't need the "restore" link or functionality in the app at all? The restore function was to be used for customers who get a new iOS device, already have a subscription, and need the app store activate the VPN client since Apple will know if they have recurring subscription.

@rlr @lesleyjanenorton what are we missing here?

bakulf:
if you have a new iOS device, you install the app and you need to complete the authentication. During the authentication the server detects that the user is already subscribed and the app shows the the main view instead of the 'subscription-needed' one.

chrismore:
@ValentinaPC can you find any use case or edge case on where the "restore subscription" is required to be able to activate an iOS VPN client? We are wonder if this feature really provides value to the customer. thanks!

ValentinaPC:
@chrismore Nothing comes into my mind, unfortunately. I guess this is also due to the fact that I am not an iPhone user. Maybe we should look towards other producers that have this type of option and how they chosen to handle it. 
Also, there might be a real case scenario, that QA is not aware off.

ValentinaPC:
Verified as fixed on the last VPN build (2.0.4 - 2.202103121417) while using iOS 14.02.
The "Restore purchases" link is no longer displayed in subscription page.

timmattison:
Can you add this feature back? Me and at least one other person can't get our subscriptions working because this feature is missing. We are just in a redirect loop that shows that we are subscribed and that we need to subscribe again.

https://support.mozilla.org/en-US/questions/1344382#answer-1438944

![IMG_7427](https://user-images.githubusercontent.com/300258/131229675-0879dfb9-4af1-4018-8ab4-7b8805038650.PNG)
![IMG_7424](https://user-images.githubusercontent.com/300258/131229680-384825dd-129d-49a9-8c05-0cccda550813.PNG)

timmattison:
Actually, there is a workaround for this. Cancel the subscription, then try to buy it again in the app. After you buy it again it will "restore purchases" and it will set up your next payment for when your canceled subscription ends.

PR #508, opened 2021-01-25
Remove "Restore purchases" UI from ViewSubscriptionNeeded.qml

lesleyjanenorton:
Fixes the frontend bits of #380


bakulf:
The code is ok, but we need to remove the C++ part too. It means that we need to remove:
1. iaphandler.h -> restoreSubscription
2. iaphandler.mm -> restoreSubscription

bakulf (approved):
I removed the c++ part. We can land this code, I guess.

PR #510, opened 2021-01-26
Server-switching without restarting wireguard - linux

bakulf:


birdsarah:
@bakulf, can you provide more context on the problem. This is for changing endpoints e.g. US Dallas to Sweden Stockholm? And before we'd just do a full disconnect and reconnect? And this is better because, it's faster?

bakulf:
@birdsarah this PR allows the user to switch the endpoint server without stopping and restarting the tunnel. When supported (Linux, mac, windows - maybe android), we can talk directly with the wireguard service and pass a new list of peers. This is better because the VPN is always active during the whole operation.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
       op, config.m_privateKey, config.m_deviceIpv4Address,
       config.m_deviceIpv6Address, config.m_serverIpv4Gateway,
       config.m_serverIpv6Gateway, config.m_serverPublicKey,
       config.m_serverIpv4AddrIn, config.m_serverIpv6AddrIn,
-      config.m_allowedIPAddressRanges.join(", "), config.m_serverPort,
-      config.m_ipv6Enabled);
+      addresses.join(", "), config.m_serverPort, config.m_ipv6Enabled);
+}
+
+static inline bool parseEndpoint(struct sockaddr* endpoint,
```
birdsarah, line 152: This method exists to verify that an endpoint is valid, correct.

1) Why not make this a general helper method that all platforms can use at a higher level?
2) Why do we need to do this at all - we are not trusting the endpoint that we get from mullvad api?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+bool DBusService::switchServer(const Config& config) {
+  logger.log() << "Switching server";
+
+  wg_device* device = static_cast<wg_device*>(calloc(1, sizeof(*device)));
+  if (!device) {
+    logger.log() << "Allocation failure";
+    return false;
+  }
+
+  auto guard = qScopeGuard([&] { wg_free_device(device); });
```
birdsarah, line 222: checking my understanding: this exists so that `device` gets cleaned up if something goes wrong?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+
+  strncpy(device->name, WG_INTERFACE, IFNAMSIZ - 1);
```
birdsarah, line 224: c++ question - we can't assign `WG_INTERFACE` to `-> name` because (a) it'll fail (b) it might change (c) other?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+  device->name[IFNAMSIZ - 1] = '\0';
```
birdsarah, line 225: then we have to manually add a null termination character? this seems like we could screw it up? or is this normal c++ life?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+
+  wg_peer* peer = static_cast<wg_peer*>(calloc(1, sizeof(*peer)));
+  if (!peer) {
+    logger.log() << "Allocation failure";
+    return false;
+  }
+
+  peer->flags =
+      (wg_peer_flags)(WGPEER_HAS_PUBLIC_KEY | WGPEER_REPLACE_ALLOWEDIPS);
```
birdsarah, line 234: set up the flags

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+
+  device->first_peer = device->last_peer = peer;
```
birdsarah, line 236: set up just one peer

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+  device->flags = WGDEVICE_REPLACE_PEERS;
+
+  if (!parseEndpoint(&peer->endpoint.addr, config.m_serverIpv4AddrIn,
+                     config.m_serverPort)) {
+    return false;
+  }
+
+  wg_key_from_base64(peer->public_key, config.m_serverPublicKey.toLocal8Bit());
```
birdsarah, line 244: it's probably just me but feels like a loverhead to just know that the public key needs to be converted `toLocal8Bit` in order to be read `from_base64` can we make this more obvious somehow? or is this a reasonable thing to expect people to know. sorry for not knowing, just trying to calibrate.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+        static_cast<wg_allowedip*>(calloc(1, sizeof(*allowedip)));
+    if (!allowedip) {
+      logger.log() << "Allocation failure";
+      return false;
+    }
+
+    if (!peer->first_allowedip) {
+      peer->first_allowedip = allowedip;
+    } else {
+      peer->last_allowedip->next_allowedip = allowedip;
```
birdsarah, line 257: i don't fully understand this syntax.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+    }
+
+    peer->last_allowedip = allowedip;
+    allowedip->cidr = ip.range();
```
birdsarah, line 261: c++ question - any reason to put this call after assignment to the peer->allowed_ips instead of before?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+
+    bool ok = false;
```
birdsarah, line 263: why are we checking if the ip address is ok after setting the peer allowed ips?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+      return false;
+    }
+
+    if (!ok) {
+      logger.log() << "Invalid IP address:" << ip.ipAddress();
+      return false;
+    }
+  }
+
+  if (wg_set_device(device) != 0) {
```
birdsarah, line 283: just checking my understanding. after updating the peer ips, we are now setting device we have created using `wg_set_device`. 

i'm a little surprised this works, so just to confirm my understanding.

you allocated a new device, you gave it the name of our existing device, you set a new set of peers, but no other config. then set device just updated the peers because we had allocated the correct name and it could see it exists already.

birdsarah (commented):
I'm not sure I would call this a review. I have a bunch of questions and understanding checking. Sorry to not be more help.

src/daemon.cpp
```diff
@@ -27,6 +27,20 @@ Daemon::~Daemon() { MVPN_COUNT_DTOR(Daemon); }
 bool Daemon::activate(const Config& config) {
```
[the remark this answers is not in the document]
bakulf: The Daemon is shared between macos, windows and linux. Each daemon behaves at the same way and it has the following functionalities:

1. activate/deactivate the VPN
2. retrieve the current status
3. retrieve the logs

With this patch we are adding a new feature: server-switching. Not all the daemon impls are able to do this operation in 1 single step (yet). If they are not able to (`supportServerSwitching()`), calling `activate` when connected, ends up in a `deactivate` + `activate`.

We have different daemon implementation because each platform has a different way to run it.
- windows: Windows Service
- macos: unix socket
- linux: dbus

For each daemon impl, we have a different controller. I managed to unify windows and mac in a single impl: `LocalSocketController`. Linux has its own `LinuxController`. Maybe we can rewrite `LinuxController` on top of `LocalSocketController`, but this is not related with this patch.

To answer your question, `activate` can be called at any time. Usually they are called because of a user interaction.

src/daemon.cpp
```diff
@@ -27,6 +27,20 @@ Daemon::~Daemon() { MVPN_COUNT_DTOR(Daemon); }
...
   m_lastConfig = config;
 
+  if (m_connected) {
+    if (supportServerSwitching(config)) {
+      logger.log() << "Already connected. Server switching supported.";
+      if (!switchServer(config)) {
+        return false;
+      }
+
+      emit connected();
```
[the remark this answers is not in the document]
bakulf: Right! let me write a comment here. The idea is that, at the end of any connection operation (first connection or server switching) we need to inform the UI via a `connected` signal.

src/daemon.cpp
```diff
@@ -159,10 +172,10 @@ bool Daemon::parseConfig(const QJsonObject& obj, Config& config) {
         continue;
       }
 
-      allowedIPAddressRanges.append(
-          QString("%1/%2").arg(address.toString()).arg(range.toInt(0)));
+      config.m_allowedIPAddressRanges.append(IPAddressRange(
```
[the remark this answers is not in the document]
bakulf: No, it's related. In order to support the linux server-switching, we need to have the allowIPAddressRange in an "object" form instead of "strings". So, I had to change `m_allowedIPAddressRanges` from being `QStringList` to `QList<IPAddressRange>`.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: It's not about verifying the endpoint, but this is about the creation of a `sockaddr`.
If you see, the WG API requires a `sockaddr` pointer in the `wg_peer` struct. This method "converts" a string (QString) + port into a `sockaddr`.

I don't know if we can make it cross-platform for the following reasons:
1. on windows `sockaddr` doesn't exist, and if it exists, is not what we should use. There is for sure a windows API...
2. the `wg_peer` exists only on linux. For mac and linux the communication works differently.

Let's see what we can "steal" from this implementation when working on macos.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: Correct! This is one le the coolest feature of object-oriented languages such as C++. It's called RAII (https://en.cppreference.com/w/cpp/language/raii). The idea is:

1. This class will go out of scope when this method ends.
2. Its DTOR will run and `[&] { wg_free_device(device); }` will be executed at that time.
3. the memory is released correctly.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: You mean: `device->name = WG_INTERFACE`

That doesn't work because `name` is an array of `char`: https://github.com/WireGuard/wireguard-tools/blob/master/contrib/embeddable-wg-library/wireguard.h#L72

In C++, you cannot assign values to an array in this way. You must use a method (or do it by your own). `strncpy` is a standard C function and it's code is something like this (pseudo code):

```
void strncpy(char* s1, const char *s2, size_t s) {
  for (int i = 0; s1 && s1[i] && s2 && s2[i] && i < s; ++i) {
    s1[i] = s2[i]; 
  }
}
```

src/platforms/linux/daemon/dbusservice.cpp
bakulf: This is a C thing. A `\0` is the string terminator. This is the only thing that guarantees the ending of a string reading (if it's not done differently). I could use C++/QT method to hide all of this, but...

src/platforms/linux/daemon/dbusservice.cpp
bakulf: Yes. This is the list of operations I want to do in this PR. I added the flags based on the wg implementation.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: This is a simple concatenated list algorithm in C.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: This is a good point. In reality, I call `toLocal8bit()` because it's the simplest way to obtain a `const char*` from a QString. Another approach is to have a QByteArray and use `constData()`.
I'll add a comment here too.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: This is another linked list impl in C. Here is the idea:

1. this is the struct (here it's called wg_allowedip: https://github.com/WireGuard/wireguard-tools/blob/master/contrib/embeddable-wg-library/wireguard.h#L25)
```
struct List {
  int value;
  struct List* next;
};
```

2. You want to add items easily to this list. So you keep 2 points: first element and the last element of the list:
```
struct Data {
  struct List* first;
  struct List* last;
};
struct Data data;
```

3. Let's create a new item:
```
struct List* item = (struct List*)malloc(sizeof(struct List));
item.next = nullptr;
item.value = 42;
```

3. Adding the first element is easy because the first element is also the last one:

```
data.first = item;
data.last = item;
```

4. Let's create a new item:

```
struct List* item = (struct List*)malloc(sizeof(struct List));
item.next = nullptr;
item.value = 24;
```

5. Adding this is a bit more tricky because I need to add the element at the end of the list, so:
```
data.last->next = item;
data.last = item;
```

Now the last element is item, and the first element->next points to the last one.
Having a list in this way is a good approach because it allows you do iterate between elements easily:

```
for (List* item = data.first; item; item = item->next) {
  std::cout << "The value is: " << item->value << std::end;
}
```

src/platforms/linux/daemon/dbusservice.cpp
bakulf: None. We can change the order.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: What will deallocate the allowedip data struct if something goes wrong?
We have the qScopeGuard RAII to release the device. So, if I add the allowed struct to the device immediately, it will be released together with the device.

Let's say I do the verification and then I add it to the device. If the verification fails, I need to release the allowedip struct (calling free, or a wg function). But if I add and then verify, I don't need to call free (wg_free_device will do it for me).

src/platforms/linux/daemon/dbusservice.cpp
bakulf: This is correct. We are updating the 'settings' of an existing device. The way WG does this is to call `wg_set_device` with a "partially-completed" device data struct. The flags are used to say which part of the device struct has been completed.

If you have time, take a look at the `wg set` command implementation:
1. https://github.com/WireGuard/wireguard-tools/blob/master/src/set.c#L25-L29
2. https://github.com/WireGuard/wireguard-tools/blob/master/src/config.c

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+    logger.log() << "Failed to set the new peer";
+    return false;
+  }
+
+  return true;
+}
+
+bool DBusService::supportServerSwitching(const Config& config) const {
```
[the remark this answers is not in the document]
bakulf: Right. The idea is this: by default the daemon class doesn't support server-switching (in this way we can do 1 implementation at the time - let's start with linux).
If a single implementation supports server-switching, it has to implement this method and return `true`.
For linux, we support server-switching only if the current settings and the new one have a few attributes in commons: device addresses, server gateway addresses and the private key. If the device key has changed, for instance, we need to disconnect and reconnect - this means, no server-switching allowed.

src/platforms/linux/linuxcontroller.cpp
```diff
@@ -82,9 +82,8 @@ void LinuxController::activate(
 void LinuxController::deactivate(Reason reason) {
   logger.log() << "LinuxController deactivated";
 
-  if (reason != ReasonNone) {
-    logger.log()
-        << "No disconnect for quick server switching or connection checks";
+  if (reason == ReasonSwitching) {
+    logger.log() << "No disconnect for quick server switching";
```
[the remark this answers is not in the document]
bakulf: here we are in the 'controller' world: the UI component. The controller talks with the daemon. The controller is relatively stupid. For each activation/server-switching it calls deactivate + activate.  The `Reason` enum tells why we are doing this operation.

For server-switching we don't need to do the whole `deactivate` because the linux daemon is able to do server-switching (this PR). So, no deactivate() is needed for linux.

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: No need. Just learning.

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: I still don't understand. If we need to verify, why not verify right at the start of the switchServer function. If the addresses aren't valid, we don't proceed. Is it to save a for loop?

birdsarah: Never mind we figured it out offline.

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: Thanks for the context. Perhaps the method name could hint more at this.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -130,11 +136,162 @@ QString DBusService::getLogs() {
...
+      (resolved->ai_family == AF_INET6 &&
+       resolved->ai_addrlen == sizeof(struct sockaddr_in6))) {
+    memcpy(endpoint, resolved->ai_addr, resolved->ai_addrlen);
+    freeaddrinfo(resolved);
+    return true;
+  }
+
+  logger.log() << "Invalid endpoint" << address;
+  freeaddrinfo(resolved);
+  return false;
```
birdsarah, line 210: A style question: The ending of this method stood out because it duplicates the call to freeaddrinfo and we end on the unhappy path....as opposed to maybe returning a "success" variable....although I can see why it is what it is. 

There's a lot of methods like this, that change states but return true/false. Maybe it's a little bit of a functional programming hangover, or a wariness of "side-effects."

I'm just throwing these thoughts out there are discussion points in case you have input / thoughts.

birdsarah:
@bakulf as you know, I'm working on the wghelper replacement, and it just occurred to me: When are allowed IPs anything other than "AllowedIPs = 0.0.0.0/0, ::0/0". If never, perhaps this PR can be smaller. I mean I know we'd probably want to support something more in the future - not sure what your take is on adding what's needed vs future proofing.

birdsarah:
@bakulf - why is this PR open if you've already committed the code to main?

bakulf:
> @bakulf - why is this PR open if you've already committed the code to main? e.g. https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/platforms/linux/daemon/dbusservice.cpp#L154

I did a mistake when I landed the windows branch, which was built on top of this PR. I still need this code to be reviewed because nobody else read it expect you. The windows PR was reviewed by @strseb but not this part.

src/daemon.cpp
```diff
@@ -27,6 +27,20 @@ Daemon::~Daemon() { MVPN_COUNT_DTOR(Daemon); }
...
+        return false;
+      }
+
+      emit connected();
+      return true;
+    }
+
+    logger.log() << "Already connected. Server switching not supported.";
+  }
+
```
birdsarah, line 43: While this change is understandable, I think the whole method is now pretty confusing.

```
bool Daemon::activate(const Config& config) {
  m_lastConfig = config;

  if (m_connected) {
    if (supportServerSwitching(config)) {
      logger.log() << "Already connected. Server switching supported.";
      if (!switchServer(config)) {
        return false;
      }

      m_connectionDate = QDateTime::currentDateTime();
      emit connected();
      return true;
    }

    logger.log() << "Already connected. Server switching not supported.";
  }

  if (m_connected) {
    if (!deactivate(false)) {
      return false;
    }

    Q_ASSERT(!m_connected);
  }

  m_connected = run(Up, m_lastConfig);
  m_connectionDate = QDateTime::currentDateTime();

  logger.log() << "Status:" << m_connected;

  if (m_connected) {
    emit connected();
  }

  return m_connected;
}
```

sources of confusion for me:
* if m_connected is checked 3 times. between the first and second time it's not clear what would have caused m_connected to change
* there are two times that the connected signal can get emitted - they are both inside checks for m_connected
* if we're setting a member variable, why are we also returning it? does this add confusion about how to use the method and are we ever actually using that return value?

If i knew we had good tests to test all the possible paths and failure modes here I'd feel better, but I don't think we do. (happy to be corrected)

question for my c++ learning:
* the passed in config is set to the variable m_lastConfig. the call to supportServerSwitching then uses config. why not use m_lastConfig, which is passed around elsewhere in the method and why pass it around at all when we're calling functions on the class?

bakulf: - I agree with you, this method needs to be refactored. 
- about storing m_lastConfig at the beginning of this method, that is a bug. We should not store it before calling supportServerSwitch(). 
- about passing m_lastConfig, we can remove it from run()/serverSwitch(), yes. Follow up?

src/platforms/linux/daemon/dbusservice.cpp
bakulf: Yes, I see your point. We could expose a sockaddr struct or nullptr. But then we need to copy its content into the peer->endpoint.addr...

src/daemon.cpp
birdsarah: > Follow up?
Definitely should I open an issue?

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: Thanks. This is very helpful. 

Just for laughs: what's cooler RAII or garbage collection?

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: Thank you.

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: but....? 

genuinely curious - it's fine if the answer is "I didn't (have time)" - I just don't know

src/daemon.cpp
bakulf: Yes, please! In the meantime, I merge the latest commit and I close this PR. is it ok?

PR #426, opened 2021-01-06
Fix small UX issues in AppPermission

strseb:
Hello! I've changed a few small things in the AppPermission that came up:

- The List is no longer child of VPNClickableRow
-> Otherwise if you fat finger in the list anything other than the checkbox, the list will close (which is probably not good)

- The Action Button (unprotect all etc) is hidden if the list is empty
-> Since it doesn't do anything if the list is empty we might also just hide it

- Disable Word Wrapping for the App PackageID
-> Some apps have very long package names, with wrapping the list element would grow and look janky :(, I've disabled word wrapping - its i guess not that important for the user, unlike the App Name. 

let me know what you think :) 

New Screen attached : 
Screenshot (06 01

PR #514, opened 2021-01-27
Word wrap by default for VPNCheckBoxRow

bakulf:
On windows, without this change, the network settings will be wrapped in this way:

image

I could change the wrapping for the settings only, but why wrapAnywhere? @lesleyjanenorton let me know what you think.

### DIFF
--- a/src/ui/components/VPNCheckBoxRow.qml
+++ b/src/ui/components/VPNCheckBoxRow.qml
@@ -19,7 +19,7 @@ RowLayout {
     property bool showDivider: true
     property var iconURL: ""
     property var leftMargin: 18
-    property var subLabelWrapMode: Text.WrapAnywhere
+    property var subLabelWrapMode: Text.WordWrap
 
     signal clicked()
     spacing: 0


src/ui/components/VPNCheckBoxRow.qml
```diff
@@ -19,7 +19,7 @@ RowLayout {
     property bool showDivider: true
     property var iconURL: ""
     property var leftMargin: 18
-    property var subLabelWrapMode: Text.WrapAnywhere
+    property var subLabelWrapMode: Text.WorldWrap
```
lesleyjanenorton, line 22: Nice catch- thank you!! Just needs to be `Text.WordWrap` instead of `.WorldWrap`. It was added in #426 but should have been `Text.WordWrap`.

lesleyjanenorton (approved):
Nice catch- thank you!! Just needs to be `Text.WordWrap` instead of `.WorldWrap`. It was added in #426 but should have been `Text.WordWrap`.

PR #520, opened 2021-01-28
Reset controller on failure

strseb:
Nothing fancy actually. I realized that `m_controller.initialize();` actually already does all the magic we need. The rest ist just calling this on the backend-error and fix failures that happen on android if we do that actually 🙈

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-786)


src/controller.cpp
```diff
@@ -62,7 +62,10 @@ Controller::Controller() {
           &Controller::connectionFailed);
 }
 
-Controller::~Controller() { MVPN_COUNT_DTOR(Controller); }
+Controller::~Controller() {
+  MVPN_COUNT_DTOR(Controller);
+  m_impl.reset(nullptr);
```
bakulf, line 67: Why is this needed?

src/platforms/android/androidcontroller.cpp
```diff
@@ -6,6 +6,8 @@
 #include "ipaddressrange.h"
 #include "leakdetector.h"
 #include "logger.h"
+#include "mozillavpn.h"
+#include "errorhandler.h"
```
bakulf, line 10: alphabetic order

src/mozillavpn.cpp
```diff
@@ -721,6 +721,12 @@ void MozillaVPN::errorHandle(ErrorHandler::ErrorType error) {
 
     case ErrorHandler::BackendServiceError:
       alert = BackendServiceErrorAlert;
+      if (m_userAuthenticated) {
+        deleteTasks();
+        // Deletes the controller-impl,Resets the status
+        // and spins up a new controller-impl
+        m_private->m_controller.initialize();
```
bakulf, line 728: OK, but what about all the other controller implementations?
We need to see if all of them work when calling initialize() again.

bakulf (changes requested):

birdsarah:
@strseb - close this?

PR #524, opened 2021-01-29
Add Windows-Service Manager to Controll the Deamon

strseb:
This adds the ability for LocalSocketController to assert that the deamon is in the running state while initiating and also allowes to boot it by itself if needed.


src/localsocketcontroller.cpp
```diff
@@ -45,7 +56,6 @@ LocalSocketController::~LocalSocketController() {
 
 void LocalSocketController::errorOccurred(
     QLocalSocket::LocalSocketError error) {
-  logger.log() << "Error occurred:" << error;
```
strseb, line 48: oops. that should not be removed 🙈

PR #530, opened 2021-02-01
Boiling down wg helper script.

birdsarah:
Depends on #463 being merged.


birdsarah:
@bakulf, once #530 is merged, we can open this PR against `main` and merge. I think this is a useful intermediary step as it makes it clear what the portions of the wg-quick script are that we're actually using.

birdsarah:
@bakulf

PR #532, opened 2021-02-01
Windows SCM support

strseb:
Adds a small class, that enables us to ensure the windows deamon is in the running state while init-ing the controller :)


src/localsocketcontroller.cpp
```diff
@@ -67,6 +77,13 @@ void LocalSocketController::initialize(const Device* device, const Keys* keys) {
 
 #ifdef MVPN_WINDOWS
   QString path = "\\\\.\\pipe\\mozillavpn";
+  if(!m_serviceManager.isRunning()){
+    logger.log() << "Deamon is not running, wait until deamon is ready";
+    // We will retry once the deamon is ready
+    m_serviceManager.startService();
```
bakulf, line 83: 1. startService is confusing because the controller cannot start the service. Probably we need a better name.
2. Should we simply communicate the failure and wait for the user to click the 'restore' button? Asking this because, let's say that the service is unable to start (a crash at boot, or anything similar), the UI will be kind-of frozen. The user in the initializing state cannot do any useful operation with the VPN.
thoughts?

bakulf (changes requested):

src/localsocketcontroller.cpp
strseb: 1. Maybe startDeamon then? 
2. I see the edge case, but i dont think !isRunning() is inheritly a failure, given that we're in initialize(). Maybe a serviceStartFailure Signal, which we then escalate to the user with the restore button?

strseb: i added more errorHandles() so in case we're failing this, we can handle it.

src/localsocketcontroller.h
```diff
@@ -33,6 +37,9 @@ class LocalSocketController final : public ControllerImpl {
 
   void cleanupBackendLogs() override;
 
+ public slots:
```
bakulf, line 40: remove `public slots`. Passing a function pointer doesn't require the `slot` keyword.

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
+#include "logger.h"
```
bakulf, line 1: copyleft/right header

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "mozillavpn.h"
+#include <Windows.h>
+#include "Windows.h"
+#include "Winsvc.h"
+#include "windowsservicemanager.h"
```
bakulf, line 6: this should go on top

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
+#include "logger.h"
+#include "mozillavpn.h"
+#include <Windows.h>
+#include "Windows.h"
```
bakulf, line 4: Why 2 "Windows.h" ? Remove the second one with "

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "Winsvc.h"
```
bakulf, line 5: <>

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "windowsservicemanager.h"
+#include "windowscommons.h"
+#include <QTimer>
```
bakulf, line 8: Sort the header.

src/localsocketcontroller.cpp
```diff
@@ -67,6 +76,13 @@ void LocalSocketController::initialize(const Device* device, const Keys* keys) {
 
 #ifdef MVPN_WINDOWS
   QString path = "\\\\.\\pipe\\mozillavpn";
+  if (!m_serviceManager.isRunning()) {
+    logger.log() << "Deamon is not running, wait until deamon is ready";
+    // We will retry once the deamon is ready
+    m_serviceManager.startService();
```
bakulf, line 82: What about if it's not initialized?

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+WindowsServiceManager::~WindowsServiceManager() {
+  if (m_service != NULL) {
+    CloseServiceHandle(m_service);
+  }
+  if (m_serviceManager != NULL) {
+    CloseServiceHandle(m_serviceManager);
+  }
+}
+
+void WindowsServiceManager::startPolling(DWORD goal_state, int max_wait_sec) {
```
bakulf, line 53: goalState and maxWaitSec

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+
+  // Starts the service if execute rights are present
+  // Starts to poll for serviceStarted
+  bool startService();
+
+  // Stops the service if execute rights are present.
+  // Starts to poll for serviceStopped
+  bool stopService();
+
+ public slots:
```
bakulf, line 36: remove `public slots`

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef WINDOWSSERVICEMANAGER
```
bakulf, line 5: _H ?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+#define WINDOWSSERVICEMANAGER
+#include <Windows.h>
+#include "Windows.h"
```
bakulf, line 8: Why 2 Windows.h ?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef WINDOWSSERVICEMANAGER
+#define WINDOWSSERVICEMANAGER
```
bakulf, line 6: empty line

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+#include <Windows.h>
+#include "Windows.h"
+#include "Winsvc.h"
```
bakulf, line 9: <>

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+#include <QTimer>
+#include <QObject>
+
+/**
+ * @brief The WindowsServiceManager provides controll over the MozillaVPNBroker
+ * service via SCM
+ */
+class WindowsServiceManager : public QObject {
```
bakulf, line 17: final

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(WindowsServiceManager)
+
+ public:
+  WindowsServiceManager();
+  ~WindowsServiceManager();
+
+  // true if the Service is running
+  bool isRunning() { return getStatus().dwCurrentState == SERVICE_RUNNING; };
```
bakulf, line 26: `bool isRunning() const ` ?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+  // Gets Emitted after the Service moved From SERVICE_START_PENDING to
+  // SERVICE_RUNNING
+  void serviceStarted();
+  void serviceStopped();
+
+ private:
+  // Returns the State of the Process:
+  // See
+  // SERVICE_STOPPED,SERVICE_STOP_PENDING,SERVICE_START_PENDING,SERVICE_RUNNING
+  SERVICE_STATUS_PROCESS getStatus();
```
bakulf, line 49: `getStatus() const ` ?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+  bool m_has_access = false;
+  LPCWSTR m_serviceName = L"MozillaVPNBroker";
+  SC_HANDLE m_serviceManager;
+  SC_HANDLE m_service;  // Service handle with r/w priv.
+  DWORD m_state_target;
+  int m_currentWaitTime;
+  int m_maxWaitTime;
+  QTimer m_timer;
+
+  void startPolling(DWORD goal_state, int maxS);
```
bakulf, line 59: move it next to `getStatus`.

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+                    SC_MANAGER_QUERY_LOCK_STATUS | STANDARD_RIGHTS_READ;
+  m_serviceManager = OpenSCManager(NULL,  // local computer
+                                   NULL,  // servicesActive database
+                                   scm_rights);
+  err = GetLastError();
+  if (err != NULL) {
+    logger.log() << " OpenSCManager failed code: " << err;
+    return;
+  }
+  logger.log() << "OpenSCManager access given - " << err;
```
bakulf, line 26: well, err is null here :)

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+WindowsServiceManager::WindowsServiceManager() {
+  DWORD err = NULL;
+  auto scm_rights = SC_MANAGER_CONNECT | SC_MANAGER_ENUMERATE_SERVICE |
+                    SC_MANAGER_QUERY_LOCK_STATUS | STANDARD_RIGHTS_READ;
+  m_serviceManager = OpenSCManager(NULL,  // local computer
+                                   NULL,  // servicesActive database
+                                   scm_rights);
+  err = GetLastError();
+  if (err != NULL) {
+    logger.log() << " OpenSCManager failed code: " << err;
```
bakulf, line 23: Can we call WindowsCommon::windowLog ?

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+    emit serviceStopped();
+    return;
+  }
+}
+
+SERVICE_STATUS_PROCESS WindowsServiceManager::getStatus() {
+  SERVICE_STATUS_PROCESS serviceStatus;
+  if (!m_has_access) {
+    logger.log() << "Need read access to get service state";
+    return serviceStatus;
```
bakulf, line 93: what's the default value of it?

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,151 @@
...
+  if (state != SERVICE_STOPPED && state != SERVICE_STOP_PENDING) {
+    logger.log() << ("Service start not possible, as its running");
+    emit serviceStarted();
+    return true;
+  }
+  if (!m_has_access) {
+    logger.log() << "Need execute access to start service";
+    // Still start polling - windows should start the service
+    m_state_target = SERVICE_RUNNING;
+    pollStatus();
```
bakulf, line 115: Should we do nothing in case we do not have access?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+  // SERVICE_RUNNING
+  void serviceStarted();
+  void serviceStopped();
+
+ private:
+  // Returns the State of the Process:
+  // See
+  // SERVICE_STOPPED,SERVICE_STOP_PENDING,SERVICE_START_PENDING,SERVICE_RUNNING
+  SERVICE_STATUS_PROCESS getStatus();
+  bool m_has_access = false;
```
bakulf, line 50: m_hasAccess

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,62 @@
...
+  LPCWSTR m_serviceName = L"MozillaVPNBroker";
+  SC_HANDLE m_serviceManager;
+  SC_HANDLE m_service;  // Service handle with r/w priv.
+  DWORD m_state_target;
```
bakulf, line 54: m_stateTarget

bakulf (changes requested):
Cool!

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,154 @@
...
+    emit serviceStopped();
+    return;
+  }
+}
+
+void WindowsServiceManager::startPolling(DWORD goalState, int maxWaitSec) {
+  m_goalState = goalState;
+  m_maxWaitTime = maxWaitSec;
+  m_currentWaitTime = 0;
+  m_timer.start(1000);
```
bakulf, line 91: Use a constexpr and reduce the timer to 500ms ?

src/platforms/windows/windowsservicemanager.cpp
```diff
@@ -0,0 +1,154 @@
...
+  if (ok) {
+    logger.log() << ("Service start requested");
+    startPolling(SERVICE_RUNNING, 10);
+  } else {
+    WindowsCommons::windowsLog("StartService failed");
+  }
+  return ok;
+}
+
+bool WindowsServiceManager::stopService() {
```
bakulf, line 136: We don't actually call `stopService`. Can we remove this method?

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,61 @@
...
+  // true if the Service is running
+  bool isRunning() const { return getStatus() == SERVICE_RUNNING; };
+
+  // Starts the service if execute rights are present
+  // Starts to poll for serviceStarted
+  bool startService();
+
+  // Stops the service if execute rights are present.
+  // Starts to poll for serviceStopped
+  bool stopService();
```
bakulf, line 35: No need to this.

src/platforms/windows/windowsservicemanager.h
```diff
@@ -0,0 +1,61 @@
...
+  void pollStatus();
+
+ signals:
+  // Gets Emitted after the Service moved From SERVICE_START_PENDING to
+  // SERVICE_RUNNING
+  void serviceStarted();
+  void serviceStopped();
```
bakulf, line 42: remove the signal too.

windows/installer/MozillaVPN.wxs
```diff
@@ -91,7 +91,9 @@
 					Start="auto"
 					Arguments="windowsdaemon"
 					ErrorControl="normal"
-					Vital="yes" />
+					Vital="yes">
+					 <util:PermissionEx User="Users" ServiceQueryStatus="yes"  ServiceStart="yes" />
```
bakulf, line 95: Does it mean that each app can stop/start the service? This is what I don't like too much...
is there anything better we can do?

bakulf (changes requested):
I know that we go back and forward with this PR.. I like it a lot, but there is still something I don't understand entirely:
- we do not need to stop the service. Can we go first with a simple "start-only" implementation?
- can we set the permission _only_ to the MozillaVPN binary?

strseb:
> can we set the permission only to the MozillaVPN binary?
Not in the current state no, but in theory yes! We would need to get an SID that is only mozillavpn able to optain, the only way i know of to get an SID for a process is to put it into an [appcontainer] (https://www.malwaretech.com/2015/09/advanced-desktop-application-sandboxing.html). (That would be amazing! but also thats undocumented)

windows/installer/MozillaVPN.wxs
strseb: This only means, checkstate && start but other user apps cant stop it :)

bakulf (approved):

PR #533, opened 2021-02-02
No multiple instances on windows and on linux

bakulf:


src/platforms/windows/windowseventlistener.cpp
```diff
@@ -0,0 +1,95 @@
...
+constexpr const char* UI_PIPE = "\\\\.\\pipe\\mozillavpn.ui";
+
+namespace {
+Logger logger(LOG_WINDOWS, "WindowsEventListener");
+}
+
+WindowsEventListener::WindowsEventListener() {
+  logger.log() << "Windows event listener created";
+
+  m_server.setSocketOptions(QLocalServer::WorldAccessOption);
```
strseb, line 24: Do we need WorldAccessOption? Would this make the pipe exposed via smb to the local network?

bakulf: Good catch! I rebased this patch on top of main. Closing and reopening

strseb (approved):

PR #541, opened 2021-02-03
Android isolate qt

strseb:
There were 3 things that caused us to close down unexpectedly:
We were starting the service as a bind-first service ( so it got closed once all binders are disconnected)
We were starting the service with context of qt -> so once the qt context got closed the service aswell
Qt just kills the whole process when closing -> so the service was killed aswell. 

So in android controller we're now starting the service intent-first with the app context. ( so unbinding does not affect lifecycle)
And we're putting QT in it's own process called QtOnlyProcess. This means i had to create a small dummy-vpn service that is able to prompt the vpn permission for us, since can't use jni cross process.

### DIFF
--- a/android/src/com/mozilla/vpn/VPNPermissionHelper.kt
+++ b/android/src/com/mozilla/vpn/VPNPermissionHelper.kt
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+package org.mozilla.firefox.vpn
+
+import android.content.Context
+import android.content.Intent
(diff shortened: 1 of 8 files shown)


android/src/com/mozilla/vpn/VPNPermissionHelper.kt
```diff
@@ -0,0 +1,39 @@
...
+            val appC = c.applicationContext;
+            appC.startService(Intent(appC,VPNPermissionHelper::class.java))
+        }
+    }
+
+
+    /**
+     * Fetches the Global QTAndroidActivity and calls startActivityForResult with the given intent
+     * Is used to request the VPN-Permission, if not given.
+     * Actually Implemented in src/platforms/android/AndroidController.cpp
```
bakulf, line 36: s/I/i/

bakulf (approved):

PR #546, opened 2021-02-04
WIP - Start replacing wghelper script

birdsarah:
WIP to replace wghelper script - or start the process.


src/platforms/linux/daemon/wghelper.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+bool WireguardHelper::interface_exists() {
+  // Also confirms it is wireguard
+  QStringList wg_devices = WireguardHelper::current_wireguard_devices();
+  return wg_devices.contains(WG_INTERFACE);
+};
+
+QStringList WireguardHelper::current_wireguard_devices() {
+  // wg_list_device_names returns first\0second\0third\0forth\0last\0\0
+  char* wg_devices_raw = wg_list_device_names();
```
bakulf, line 37: use `wg_for_each_device_name`. It's part of the embedded lib.

What about something like this:

```
char* deviceNames = wg_list_device_names();
size_t len = 0;
QStringList devices;
wg_for_each_device_name(deviceNames, deviceName, len) {
  devices.append(deviceName);
 }
free(deviceNames);
return devices;
```

src/platforms/linux/daemon/wghelper.cpp
```diff
@@ -0,0 +1,52 @@
...
+  while (!(wg_devices_raw[len] == '\0' && wg_devices_raw[len - 1] == '\0')) {
+    if (wg_devices_raw[len] == '\0') {
+      wg_devices.append(device);
+      device = {};
+    } else {
+      device.append(wg_devices_raw[len]);
+    }
+    ++len;
+  }
+  return wg_devices;
```
bakulf, line 51: you need to call `free(wg_devices_raw);` otherwise, you leak the strings.

src/platforms/linux/daemon/wghelper.cpp
```diff
@@ -0,0 +1,52 @@
...
+Logger logger(LOG_LINUX, "WireguardHelper");
+}
+
+bool WireguardHelper::interface_exists() {
+  // Also confirms it is wireguard
+  QStringList wg_devices = WireguardHelper::current_wireguard_devices();
+  return wg_devices.contains(WG_INTERFACE);
+};
+
+QStringList WireguardHelper::current_wireguard_devices() {
```
bakulf, line 35: 1. Add a `// static` to know it's a static method.
2. currentWireguardDevices()

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -31,6 +32,7 @@ extern "C" {
 #endif
 
 namespace {
+WireguardHelper wg_helper;
```
bakulf, line 35: This class contains only static methods. You don't  need `wg_helper`. When you need to call a method, just do: `WireguardHelper::name_of_the_method()`

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -138,6 +140,10 @@ QString DBusService::getLogs() {
 }
 
 bool DBusService::run(Op op, const Config& config) {
+  logger.log() << "Interface exists:" << wg_helper.interface_exists();
```
bakulf, line 143: `WireguardHelper::interfaceExists()`

src/platforms/linux/daemon/wghelper.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "../../3rdparty/wireguard-tools/contrib/embeddable-wg-library/wireguard.h"
+#if defined(__cplusplus)
+}
+#endif
+
+namespace {
+Logger logger(LOG_LINUX, "WireguardHelper");
+}
+
+bool WireguardHelper::interface_exists() {
```
bakulf, line 29: We are following the QT C++ style (which is the firefox one too). Call the method: `interfaceExists()`

src/platforms/linux/daemon/wghelper.cpp
```diff
@@ -0,0 +1,52 @@
...
+  // Also confirms it is wireguard
+  QStringList wg_devices = WireguardHelper::current_wireguard_devices();
```
bakulf, line 31: Also the variables :) -> wgDevices - or simply `devices`

src/platforms/linux/daemon/wghelper.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef WIREGUARDHELPER_H
+#define WIREGUARDHELPER_H
+
+#include <QStringList>
+
+class WireguardHelper final {
+ public:
+  static bool interface_exists();
+  static QStringList current_wireguard_devices();
+  static bool add_if();
```
bakulf, line 14: This doesn't exist yet, right?

bakulf (changes requested):
In general, I like this approach, but I have a comment to make. If we want to write this code, thinking in a cross-platform way, we can say that each platform has a set of tasks to complete:

1. bring up the interface
2. configure the routing
3. configure the DNS

And, of course, during the shutdown, the opposite tasks. So, I wonder if we can continue in this way, and then generalize it. Similarly to what we have done for the 'Daemon' class. For instance, MacOS has the same operations to complete (with different commands because darwin is a BSD and not a Linux OS, but the tasks are the same).

birdsarah:
@bakulf - this is a WIP PR. It's me pushing things and having a record of them so I can ask you questions when I need to. I'm not in anyway proposing that this is the final code or code structure. I'm just trying to make progress.

birdsarah:
https://git.zx2c4.com/wireguard-tools/tree/src/config.c

birdsarah:
> @bakulf - this is a WIP PR. It's me pushing things and having a record of them so I can ask you questions when I need to. I'm not in anyway proposing that this is the final code or code structure. I'm just trying to make progress.

Apologies if WIP PRs are not the way of doing things in this repo. I should have checked to avoid confusion and time wasting.

PR #550, opened 2021-02-05
Wait for the Correct IP to be fetched

### LINKED ISSUE #549: "Connection Information" screen lists real IP
#### Affected versions:
- v. 2.0.3 last build; 

#### Affected Platforms:
- Windows 10;
- Android 10;

#### Prerequisites:
-  Have a FxA account with active subscription for Mozilla VPN;

#### Steps to reproduce:
1. Open VPN app;
2. Turn ON the VPN;
3. Select a random region;
4. Go to Connection Information screen and observe the IP;

#### Expected result:
- The IP of the selected region is listed;

#### Actual result:
- The real IP is displayed;

#### Notes:
- ipinfo.ro lists the correct IP (the one of the selected region)

**- Android** 
-- selected region IP is displayed after 20-30 seconds from location switch;
-- turning off-on the VPN, without changing the location, also triggers a different IP (maybe this is expected);

strseb:
Prevent to fetch the ip if we're not connected so the real ip is not shown while we try to get the vpn-ed ip. 
Assert that the fetched ip has to be reported in the same country as we are connected to, otherwise retry. 
Oh and change the "unknown" string to "loading".. as we're probably loading when the user sees this. :) 

Closes #549

### DIFF
--- a/src/connectiondataholder.cpp
+++ b/src/connectiondataholder.cpp
@@ -244,7 +244,6 @@ void ConnectionDataHolder::updateIpAddress() {
                 // connected server we may retry only once.
                 logger.log() << "Reported ip not in the right country, retry!";
                 TimerSingleShot::create(this, 3000, [this]() {
-                  m_updatingIpAddress = false;
                   updateIpAddress();
                 });
               }
--- a/src/connectiondataholder.cpp
+++ b/src/connectiondataholder.cpp
@@ -8,6 +8,7 @@
 #include "logger.h"
 #include "mozillavpn.h"
 #include "networkrequest.h"
+#include "timersingleshot.h"
 
 #include <QJsonDocument>
 #include <QJsonObject>
@@ -19,10 +20,10 @@ namespace {
 Logger logger(LOG_NETWORKING, "ConnectionDataHolder");
 }
 
-//% "Unknown"
-//: This refers to the current IP address, i.e. "IP: Unknown".
+//% "Loading"
+//: This refers to the current IP address, i.e. "IP: Loading".
 ConnectionDataHolder::ConnectionDataHolder()
-    : m_ipAddress(qtTrId("vpn.connectionInfo.unknown")) {
+    : m_ipAddress(qtTrId("vpn.connectionInfo.loading")) {
   MVPN_COUNT_CTOR(ConnectionDataHolder);
 
   connect(&m_ipAddressTimer, &QTimer::timeout, [this]() { updateIpAddress(); });
@@ -104,6 +105,7 @@ void ConnectionDataHolder::activate(const QVariant& a_txSeries,
                                     const QVariant& a_axisX,
                                     const QVariant& a_axisY) {
   logger.log() << "Activated";
+  updateIpAddress();
 
   QtCharts::QSplineSeries* txSeries =
       qobject_cast<QtCharts::QSplineSeries*>(a_txSeries.value<QObject*>());
@@ -200,6 +202,12 @@ void ConnectionDataHolder::reset() {
 
 void ConnectionDataHolder::updateIpAddress() {
   logger.log() << "Updating IP address";
+  auto state = MozillaVPN::instance()->controller()->state();
+  // Only start the check if we're actually connected/connecting
+  if (state != Controller::StateOn && state != Controller::StateConfirming) {
+    logger.log() << "Skip Updating IP address, not connected";
+    return;
+  }
 
   if (m_updatingIpAddress) {
     return;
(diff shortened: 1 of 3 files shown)


src/connectiondataholder.cpp
```diff
@@ -224,14 +232,23 @@ void ConnectionDataHolder::updateIpAddress() {
...
             QJsonDocument json = QJsonDocument::fromJson(data);
             if (json.isObject()) {
               QJsonObject obj = json.object();
 
+              QJsonValue country = obj.take("country");
+              if (country.toString().toLower() !=
+                  MozillaVPN::instance()->currentServer()->countryCode()) {
+                logger.log() << "Reported ip not in the right country, retry!";
+                m_updatingIpAddress = false;
+                updateIpAddress();
```
bakulf, line 244: if for any reason the server goes out of sync, this method generates a loop of updateIpAddress(). Do we care about checking the countryCode ?

bakulf (changes requested):
All good! Just a question about the loop.

src/connectiondataholder.cpp
strseb: Sadly thats essential. Even after confirming we might encounter a stray request that did not yet use the vpn, for whatever reason. I would rather add an exit condition then removing this :)

bakulf: We spoke about this on slack. This code has to changed. Maybe a simple timer can help.

src/connectiondataholder.cpp
```diff
@@ -224,14 +233,29 @@ void ConnectionDataHolder::updateIpAddress() {
   connect(request, &NetworkRequest::requestCompleted,
           [this](QNetworkReply*, const QByteArray& data) {
             logger.log() << "IP address request completed";
-
             QJsonDocument json = QJsonDocument::fromJson(data);
             if (json.isObject()) {
               QJsonObject obj = json.object();
 
+              QJsonValue country = obj.take("country");
+              if (!m_retryIssued &&
```
bakulf, line 241: you need to reset m_retryIssued. Or maybe you can simply get rid of this variable in this way:
```
if (m_checkStatusTimer.isActive() && country.toString().toLower() !=
                      MozillaVPN::instance()->currentServer()->countryCode()) {
```

src/connectiondataholder.cpp
```diff
@@ -224,14 +233,29 @@ void ConnectionDataHolder::updateIpAddress() {
...
+                  country.toString().toLower() !=
+                      MozillaVPN::instance()->currentServer()->countryCode()) {
+                // In case the country-we're reported in does not match the
+                // connected server we may retry only once.
+                logger.log() << "Reported ip not in the right country, retry!";
+                TimerSingleShot::create(this, 3000, [this]() {
+                  m_updatingIpAddress = false;
+                  updateIpAddress();
+                });
+                m_retryIssued = true;
```
bakulf, line 251: What about if we show the IP even if it's the wrong one? We will update it in a few seconds...

bakulf (changes requested):

src/connectiondataholder.cpp
strseb: good point

src/connectiondataholder.cpp
strseb: I'm resetting this in ConnectionDataHolder::reset()  - wouldn't m_checkStatusTimer make it a bit racy? In case e.g. the first request takes longer to complete and the user navigates to the page?

bakulf: m_checkStatusTimer is activated when the info page is shown and it's stopped when the user leaves it. I see the following scenarios:
- the IP check is faster than the navigation and good -> all good!
- the IP check is faster than the navigation but with a failure -> in 3 seconds we try again
- - the IP check is slower than the navigation (at the first shot or the following one) -> when completed, we evaluate the next move.

What do you think?

bakulf (changes requested):

src/connectiondataholder.cpp
```diff
@@ -224,14 +232,27 @@ void ConnectionDataHolder::updateIpAddress() {
...
 
+              QJsonValue country = obj.take("country");
+              if (m_checkStatusTimer.isActive() &&
+                  country.toString().toLower() !=
+                      MozillaVPN::instance()->currentServer()->countryCode()) {
+                // In case the country-we're reported in does not match the
+                // connected server we may retry only once.
+                logger.log() << "Reported ip not in the right country, retry!";
+                TimerSingleShot::create(this, 3000, [this]() {
+                  m_updatingIpAddress = false;
```
bakulf, line 247: This line can be removed. If there is another check in progress, we abort the new one.

bakulf (changes requested):
All good! There is only 1 remaining thing: fix the unit tests :)

bakulf:
Thanks!

PR #557, opened 2021-02-08
Geoip check for balrog updates

bakulf:


strseb (approved):

src/ui/components/VPNSystemAlert.qml
```diff
@@ -107,6 +104,20 @@ VPNAlert {
...
             }
 
+        },
+        State {
+            name: VPN.GeoIpRestrictionAlert
+
+            PropertyChanges {
+                target: alertBox
+                alertType: "geoip-restriction"
+                //% "Operation not available where you are located"
```
flodolo, line 114: I can't figure out what this means, we'll need a localization comment.

Can you explain to me, possibly rephrasing it, what "located" means in this context? When is a user located?

PR #562, opened 2021-02-08
Add BuildID

### LINKED ISSUE #511: Let's add the VersionCode or CommitID to the info page.
Since we now have several android builds with 2.0.3 flying around it's kinda hard to check if we're testing the right build.

strseb:
This adds info field for the BuildID if it's available :) 
closes #511 
Screenshot (08 02

### DIFF
--- a/scripts/android_package.sh
+++ b/scripts/android_package.sh
@@ -147,6 +147,7 @@ if [[ "$RELEASE" ]]; then
   printn Y "Use release config"
   $QTPATH/android/bin/qmake -spec android-clang \
     VERSION=$SHORTVERSION \
+    BUILD_ID=$VERSIONCODE \
     CONFIG+=qtquickcompiler \
     CONFIG-=debug \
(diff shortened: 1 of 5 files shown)


Steg17 (approved):

lesleyjanenorton:
Sorry for the lag here - this looks great but or some reason when I try to build locally I get this error: 
```
error: use of undeclared identifier 'BUILD_ID'
In file included from ../../../mv/src/platforms/android/androiddatamigration.cpp:8:
../../../mv/src/mozillavpn.h:174:54: error: use of undeclared identifier 'BUILD_ID'
  const QString buildNumber() const { return QString(BUILD_ID); }
                                                     ^

bakulf:
Can we land this PR? I think it's nice to have. To fix the compilation issue, just add `DEFINES += BUILD_ID=\\\"1234\\\"` in `tests/unit/unit.pro`

PR #581, opened 2021-02-11
Captive portal implementation for windows (unsecured network alerts for windows, linux, macOS, wasm)

bakulf:
The PR does:
1. UI for the settings/notifications (captive portal and unsecured network alerts)
2. Captive portal detection for windows - implemented as in v1.0
3. Unsecured network alert for windows
4. Unsecured network alert for Linux (via DBus)
5. Unsecured network alert for MacOS
6. Unsecured network alert for Wasm

If this PR is rebased on top of the IP filtering one, it implements captive-portal for Linux and macos too.


bakulf:
@birdsarah do you mind reviewing the network watcher code for linux? it uses DBus to retrieve the list of wifi devices and to detect unsecured networks. Feel free to review all, but in particular, take a look at:

- networkwatcher.{h,cpp}
- networkwatcherimpl.h
- platforms/linux/networkwatcher.{h,cpp}

birdsarah:
will do review today.

src/platforms/linux/linuxnetworkwatcher.h
```diff
@@ -0,0 +1,37 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef LINUXNETWORKWATCHER_H
+#define LINUXNETWORKWATCHER_H
+
+#include "networkwatcherimpl.h"
+
+#include <QMap>
+#include <QVariant>
```
birdsarah, line 11: Question about includes here for my own learning: Why, in this case, do we not do the forward include (? forgotten the exact terminology), that you suggested I use somewhere else?

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include <QtDBus/QtDBus>
+
+// https://developer.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMDeviceType
+#ifndef NM_DEVICE_TYPE_WIFI
+#  define NM_DEVICE_TYPE_WIFI 2
+#endif
+
+// https://developer.gnome.org/NetworkManager/stable/nm-dbus-types.html#NM80211ApFlags
+#ifndef NM_802_11_AP_SEC_NONE
```
birdsarah, line 18: Lets add some comments of a human readable description - "Wifi network has no security", "Wifi network has WEP," ....

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+#ifndef NM_802_11_AP_SEC_PAIR_WEP104
+#  define NM_802_11_AP_SEC_PAIR_WEP104 0x00000001
+#endif
+
+namespace {
+Logger logger(LOG_LINUX, "LinuxNetworkWatcher");
+}
+
+static inline bool checkSecurityFlags(int securityFlags) {
```
birdsarah, line 34: It may well be my intexperience with bitwise operation, or just my inexperience in general, but I find a method like this very hard to parse. It may be that expanding it out to make it more verbose is unacceptably inefficient. If so, some tests that spell out the various tests that are being tested would be great.

src/platforms/linux/linuxnetworkwatcher.h
bakulf: You can use forwarding declaration only in case of pointers and references. In this scenario, I cannot use references or pointers for how dbus slots/signals work.

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+LinuxNetworkWatcher::~LinuxNetworkWatcher() {
+  MVPN_COUNT_DTOR(LinuxNetworkWatcher);
+}
+
+void LinuxNetworkWatcher::initialize() {
+  logger.log() << "initialize";
+
+  // Let's wait a few seconds to allow the UI to be fully loaded and shown.
+  // This is not strickly needed, but it's better from a user-experience.
```
birdsarah, line 53:
```suggestion
  // This is not strictly needed, but it's better for user experience because <please fill in the because>
```

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+  TimerSingleShot::create(this, 2000, [this]() {
+    logger.log()
+        << "Retrieving the list of wifi network devices from NetworkManager";
+
+    QDBusInterface nm(
+        "org.freedesktop.NetworkManager", "/org/freedesktop/NetworkManager",
```
birdsarah, line 59: What about declaring these at the top along with the wifi flags. Keeps all the dbus constants at the same place.

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+        "org.freedesktop.NetworkManager", QDBusConnection::systemBus());
```
birdsarah, line 60: Do we want to gate this stuff at the top or somewhere else to check `org.freedesktop.NetworkManager` is available. I've been thinking about the same thing for my work.

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+    if (!nm.isValid()) {
+      logger.log()
+          << "Failed to connect to the network manager via system dbus";
+      return;
+    }
+
+    QDBusMessage msg = nm.call("GetDevices");
```
birdsarah, line 67: This is awesome.

src/platforms/linux/linuxnetworkwatcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+    QDBusArgument arg = msg.arguments().at(0).value<QDBusArgument>();
```
birdsarah, line 68: This makes me sad. Not that your fault. But there's a lot of knowledge baked into knowing where the arguments have arrived too. Not sure what the typical way of handling this is. Maybe just a link to docs.

src/platforms/linux/linuxnetworkwatcher.cpp
bakulf: is it good enought to call: `nm.isValid()`  ?

src/platforms/linux/linuxnetworkwatcher.cpp
bakulf: yeah, I agree with you. I wrote this code reading the official documentation: https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html - let's add the link.

src/platforms/linux/linuxnetworkwatcher.h
birdsarah: Thank you!

src/networkwatcher.cpp
```diff
@@ -0,0 +1,141 @@
...
+  } else {
+    m_impl->stop();
+  }
+}
+
+void NetworkWatcher::unsecuredNetwork(const QString& networkName,
+                                      const QString& networkId) {
+  logger.log() << "Unsecured network:" << networkName << "id:" << networkId;
+
+#ifndef UNIT_TEST
```
birdsarah, line 98: Is this UNIT_TEST like unit testing?

birdsarah (approved):
I finished going through the files you asked me to. Sorry yesterday just was a scrabble for water and sanity.

I'm pleasantly surprised at how nice this is (at least from the linux side).

src/networkwatcher.cpp
bakulf: yes it is. For unit tests this part of the code is not needed.

birdsarah: Fascinating. No rush, just more learning questions. (1) Is it normal to mark up main code with test notes like this? (2) Are there typical patterns where this happens?

bakulf: Our unit-tests are here: https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/tests/unit - We test only part of the codebase (yet) and, in order to do it, we compile part of the MozillaVPN code, together with the unit-test code.  When we want to disable/enable/change some detail, we use `#ifdef UNIT_TEST`.
We should reduce that at the minimum to have a better test coverage but there are a few exceptions:
1. MozillaVPN and ConnectionDataHolder - we use UNIT_TEST to have access to private class members:  exposing them would be a bad object-oriented logic encapsulation, but it's acceptable (and good for testing!) in unit-tests
2. Constants - we use UNIT_TEST to have faster operations such as: how often check new releases, how often check the IP addresses, etc etc. We have actually 3 values for most of the constants: production build, debug build, testing. See: contents.h
3. Logger - we use UNIT_TEST to compile and run android components even when not on android
4. VersionAPI, PingSender - we use UNIT_TEST to compile the Dummy ping sender or to check a generic update endpoint. Otherwise the unit-test would be platform specific.
5. Settings - we use UNIT_TEST because we don't want to write settings for real. The setting saving part is based on QSettings and I hope it's already tested by the QT framework

Here we do something similar to 4.

lesleyjanenorton (approved):
Inspector is updated! QML looks good to me and everything seems go be working locally.

src/ui/settings/ViewNetworkSettings.qml
```diff
@@ -69,7 +69,7 @@ VPNFlickable {
         anchors.top: localNetwork.visible ? localNetwork.bottom : ipv6.bottom
         visible: !vpnFlickable.vpnIsOff
 
-        //% "VPN must be off to edit network settings"
+        //% "VPN must be off to edit settings"
```
flodolo, line 72: For the future, please use new IDs when making these changes. If you don't, you'll have strings in localized files that don't match the en-US source ([more info on the general process here](https://mozilla-l10n.github.io/documentation/localization/making_string_changes.html)).

The infra in the l10n repository will invalidate these translations on the next merge, but in the meantime you're in a limbo.

src/systemtrayhandler.cpp
```diff
@@ -165,13 +169,54 @@ void SystemTrayHandler::updateContextMenu() {
...
+}
+
+void SystemTrayHandler::captivePortalBlockNotificationRequired() {
+  logger.log() << "Captive portal block notification shown";
+
+  //% "Guest Wi-Fi portal blocked"
+  QString title = qtTrId("vpn.systray.captivePortalBlock.title");
+
+  //% "The guest Wi-Fi network you’re connected to requires action. Click to"
+  //% "turn off VPN to see the portal."
```
flodolo, line 197: You're missing a space between to and turn, resulting in 

```
<source>The guest Wi-Fi network you’re connected to requires action. Click toturn off VPN to see the portal.</source>
```

Issue #521, opened 2021-01-28
Offer the Local Network Access feature on all platforms


chrismore:
The new local network access feature is part of the new v2.0 code base, but it is not exposed and implemented on all platforms. We should make it available on Windows, Mac, Android, iOS, and Linux. This issue will track the work to make it available on all platforms. 

We have issue #295 for the Linux implementation.

PR #585, opened 2021-02-15
Filtering out the captive portal endpoint

### LINKED ISSUE #295: Local Network Access supported on Linux

bakulf:
On top of the IP filtering for the captive portal, this PR closes #295 and #521, implementing the local access network for all the platforms (Linux, windows, macos, ios, for sure. Android to be checked)

### DIFF
--- a/src/ipaddress.h
+++ b/src/ipaddress.h
@@ -10,6 +10,8 @@
 class IPAddress final {
  public:
   static IPAddress create(const QString& ip);
+  static QList<IPAddress> excludes(const QList<IPAddress>& sourceList,
+                                   const QList<IPAddress>& excludeList);
 
   IPAddress();
   IPAddress(const IPAddress& other);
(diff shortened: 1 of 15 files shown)


src/ipaddress.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef IPADDRESS_H
+#define IPADDRESS_H
+
+#include <QHostAddress>
+
+class IPAddress final {
+ public:
+  static IPAddress create(const QString& ip);
+  static QList<IPAddress> excludes(const QList<IPAddress>& sourceList,
```
valenting, line 13: `exclude` seems a little better to me than `excludes`. I would expect `excludes` to be the opposite of `contains`

bakulf: I did a couple of important changes:
- fix the allowed ipv4 lists - now all works as expected
- I renamed `excludes` to `excludeAddresses`
- I kept `contains()` and `overlaps()` because they are methods of the object, used internally and they follow the QT name convention. Ex: `QString::contains()`

strseb (approved):

PR #594, opened 2021-02-17
Remove WASM header for non-wasm builds

bakulf:


bakulf:
I already merged this PR because the builds were broken. If this code is not good, we can fix it and land the code again.

PR #601, opened 2021-02-19
Fix missing spaces in captive portal messages

flodolo:


flodolo:
See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/581#pullrequestreview-593919033

PR #613, opened 2021-02-23
Use our own copy of wg-quick

birdsarah:
https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash

### DIFF
--- a/src/wgquickprocess.cpp
+++ b/src/wgquickprocess.cpp
@@ -23,6 +23,8 @@ Logger logger(
 
 QString scriptPath() {
 #if defined(MVPN_LINUX)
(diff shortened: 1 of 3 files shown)


src/wgquickprocess.cpp
```diff
@@ -23,6 +23,8 @@ Logger logger(
 
 QString scriptPath() {
 #if defined(MVPN_LINUX)
+  QDir appPath(MVPN_DATA_PATH);
+  return appPath.filePath("helper.sh");
   return "wg-quick";
```
bakulf, line 28: Remove this extra return

bakulf (approved):
Good! Apply the comment, and let's merge it! Thanks

PR #616, opened 2021-02-23
Network requests improvements

bakulf:
This PR does 2 things:
1. it fixes the authentication failure, letting the `NetworkRequest::replyFinished()` to propagate the error
2. it removes the sharing of `QNetworkReply` to the rest of the codebase. It was done for balrog and used for 2 methods: `rawHeader` and `statusCode`. Now these 2 methods are part of the `NetworkRequest` class

### DIFF
--- a/src/update/balrog.cpp
+++ b/src/update/balrog.cpp
@@ -381,15 +381,13 @@ bool Balrog::processData(const QByteArray& data) {
             });
 
     connect(request, &NetworkRequest::requestHeaderReceived,
-            [this](QNetworkReply* reply) {
+            [this](NetworkRequest* request) {
               Q_ASSERT(reply);
               logger.log() << "Request header received";
 
               // We want to proceed only if the status code is 200. The request
               // will be aborted, but the signal emitted.
-              QVariant statusCode =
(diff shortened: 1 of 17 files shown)


src/update/balrog.cpp
```diff
@@ -374,32 +370,29 @@ bool Balrog::processData(const QByteArray& data) {
...
+            [this](NetworkRequest* request) {
+              Q_ASSERT(request);
               logger.log() << "Request header received";
 
               // We want to proceed only if the status code is 200. The request
               // will be aborted, but the signal emitted.
-              QVariant statusCode =
-                  reply->attribute(QNetworkRequest::HttpStatusCodeAttribute);
-              if (statusCode.isValid() && statusCode.toInt() == 200) {
+              if (request->statusCode() == 200) {
```
birdsarah, line 385: this is nice :)

birdsarah (commented):
I'm afraid I can't comment on implications or big picture thoughts due to my ongoing unfamiliarity.

What I can say is this seems great. We're cleaning out code and the new code that's added is much cleaner and easier to understand.

PR #615, opened 2021-02-23
Strip down helper.sh

birdsarah:
Naive removal of all the parts that we're not using.

PR #620, opened 2021-02-24
First wgutil functions.

birdsarah:
Merge #615 and #619 first.

### DIFF
--- a/src/daemon/wgutils.cpp
+++ b/src/daemon/wgutils.cpp
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <arpa/inet.h>
+
+#include <QString>
+
+#include "daemon.h"
+#include "logger.h"
+#include "loghandler.h"
+#include "wgutils.h"
+#include "wgquickprocess.h"  // source of WG_INTERFACE
+
+// Import wireguard C library
+#if defined(__cplusplus)
+extern "C" {
+#endif
+#include "../../3rdparty/wireguard-tools/contrib/embeddable-wg-library/wireguard.h"
+#if defined(__cplusplus)
+}
+#endif
+// End import wireguard
+
+namespace {
+Logger logger(LOG_MAIN, "WireguardUtils");
+}
+
+/*
+ * HELPERS
+ */
+
+char* iface_name() { return QString(WG_INTERFACE).toLocal8Bit().data(); }
+
+/*
+ * END HELPERS
+ */
+
+/*
+ * PUBLIC METHODS
+ */
+
+// static
+bool WireguardUtils::interfaceExists() {
+  // Also confirms it is wireguard.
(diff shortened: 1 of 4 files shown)


src/daemon/wgutils.cpp
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <arpa/inet.h>
```
bakulf, line 4: We have the includes following these guidelines:

1. first the internal headers
2. after an empty line, the external headers
3. each block (internal and external headers) alphabetic sorted
4. if there is a corresponding header for the cpp class, that goes on top. In this case, the first internal header is wgutils.h (even if not alphabetic sorted)

So, in this case the order would be:

```
#include "wgutils.h"
#include "daemon.h"
#include "logger.h"
#include "loghandler.h"
#include "wgquickprocess.h"  // source of WG_INTERFACE

#include <arpa/inet.h>
#include <QString>
```

Then, QString is not required, I suspect.

src/daemon/wgutils.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace {
+Logger logger(LOG_MAIN, "WireguardUtils");
+}
+
+/*
+ * HELPERS
+ */
+
+char* iface_name() { return QString(WG_INTERFACE).toLocal8Bit().data(); }
```
bakulf, line 32: make this static (or move it into the namespace).

bakulf: Why this function? I mean, you are exposing WG_INTERFACE as it is.Can you use WG_INTERFACE directly?

src/daemon/wgutils.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+// static
+bool WireguardUtils::interfaceExists() {
+  // Also confirms it is wireguard.
+  return WireguardUtils::currentInterfaces().contains(iface_name());
+};
+
+// static
+QStringList WireguardUtils::currentInterfaces() {
+  char *deviceNames, *deviceName;
```
bakulf, line 50: in C++, the definition of the variables is closed to the usage:
`char *deviceNames = wg_list_device_names();`

Can you check if `wg_for_each_device_name` is able to deal with a nullptr as deviceNames? If not, you should add:

`if (!deviceNames) { return devices; }`

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -9,6 +9,7 @@
 #include "loghandler.h"
 #include "polkithelper.h"
 #include "wgquickprocess.h"
+#include "daemon/wgutils.h"
```
bakulf, line 12: alphabetic order.

bakulf (changes requested):
All good! just rebase it and fix details.

birdsarah:
> All good! just rebase it and fix details.

This was a draft PR. Wasn't intended for review or merge. This is the second time this has happened. Can we get clear about dev processes?

PR #628, opened 2021-02-25
Screen capture through the inspector

bakulf:


lesleyjanenorton (approved):
Looks good to me! I'm still running into a few issues locally but I think it's just something with my build/machine. I will get the page for viewing these images set up tomorrow!

PR #632, opened 2021-02-26
Add wghelper to daemon implement first pieces for linux

birdsarah:
This is first PR that sets up the general architecture of how adding all the pieces that will allow us to remove helper scripts is going to go. This includes moving the Daemon::Config to wgutils InterfaceConfig. Although the immediate use for that isn't clear in this PR. Future methods in wgutils will need the Daemon::Config and we can end up with a circular import if we're not careful.


src/daemon/daemon.cpp
```diff
@@ -82,6 +83,13 @@ bool Daemon::activate(const Config& config) {
     return activate(config);
   }
 
+  if (supportWGUtils()) {
```
bakulf, line 86: You can merge these 2 if-stmt: 
```
if (supportWGUtils() && wgutils()->interfaceExists()) { ...
```

src/daemon/daemon.cpp
```diff
@@ -222,8 +230,15 @@ bool Daemon::deactivate(bool emitSignals) {
     return true;
   }
 
-  m_connected = false;
+  if (supportWGUtils()) {
```
bakulf, line 233: Same here: `if (a && b)`

src/daemon/config.h
```diff
@@ -0,0 +1,25 @@
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef INTERFACECONFIG_H
```
bakulf, line 6: This file should be called 'interfaceconfig.h' or here we should replace 'interfaceConfig' with 'Config'

src/daemon/wgutils.h
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef WIREGUARDUTILS_H
```
bakulf, line 5: Here too: WireguardUtils vs WgUtils

src/daemon/wgutils.h
```diff
@@ -0,0 +1,26 @@
...
+#define WIREGUARDUTILS_H
+
+#include <QCoreApplication>
+
+constexpr const char* WG_INTERFACE = "moz0";
+
+class WireguardUtils : public QObject {
```
bakulf, line 12: Usually, the filename matches with the name of the class

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -40,6 +41,13 @@ DBusService::DBusService(QObject* parent) : Daemon(parent) {
 
 DBusService::~DBusService() { MVPN_COUNT_DTOR(DBusService); }
 
+WireguardUtilsLinux* DBusService::wgutils() {
+  if (!m_wgutils) {
+    m_wgutils = new WireguardUtilsLinux();
```
bakulf, line 46: This is a bug: here you are allocating an object but you do not deallocate it.
The simplest solution is to set a parent to WireguardUtilsLinux and pass it to QObject. See this: https://doc.qt.io/qt-5/qobject.html#QObject "The destructor of a parent object destroys all child objects."
\\

bakulf: If you add MVPN_COUNT_CTOR/DTOR in the wgutils, you will see that in debug builds the leak is reported.

src/platforms/linux/daemon/wgutilslinux.cpp
```diff
@@ -0,0 +1,40 @@
...
+bool WireguardUtilsLinux::interfaceExists() {
+  // As currentInterfaces only gets wireguard interfaces, this method
+  // also confirms an interface as being a wireguard interface.
+  return currentInterfaces().contains(WG_INTERFACE);
+};
+
+QStringList WireguardUtilsLinux::currentInterfaces() {
+  char* deviceName;
+  size_t len;
+  QStringList devices;
```
bakulf, line 30: Move these next to the first use. Line 35.

bakulf (changes requested):

src/daemon/config.h
birdsarah: What do you think of the name InterfaceConfig? It seems right, but I wanted to check. If InterfaceConfig is good, I'll rename the file.

src/daemon/wgutils.h
birdsarah: got it.

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: Thank you. I'll fix this. Sorry I still have things to learn to get this ingrained in my head.

birdsarah: Resolved in 9761274

birdsarah:
@bakulf this is ready for a re-review.

I have not implemented `release` as we started talking about in slack. I believe it's not necessary in this PR as this PR doesn't do anything that needs to be un-done. It simply provides the checks and sets up the initial architecture.

src/daemon/interfaceconfig.h
```diff
@@ -0,0 +1,25 @@
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
bakulf, line 2: No extra line here.

src/daemon/interfaceconfig.h
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef INTERFACECONFIG_H
+#define INTERFACECONFIG_H
+
+#include "ipaddressrange.h"
```
bakulf, line 9: I would include <QList> and <QString>

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -40,6 +40,14 @@ DBusService::DBusService(QObject* parent) : Daemon(parent) {
...
+    m_wgutils->setParent(this);
```
bakulf, line 46: This works too, but why not passing the argument in the CTOR?

src/platforms/linux/daemon/dbusservice.h
```diff
@@ -34,16 +35,19 @@ class DBusService final : public Daemon {
...
-  bool run(Op op, const Config& config) override;
+  bool run(Op op, const InterfaceConfig& config) override;
 
-  bool supportServerSwitching(const Config& config) const override;
+  bool supportServerSwitching(const InterfaceConfig& config) const override;
+  bool switchServer(const InterfaceConfig& config) override;
 
-  bool switchServer(const Config& config) override;
+  bool supportWGUtils() const override { return true; }
+  WireguardUtilsLinux* wgutils() override;
```
bakulf, line 44: This should return a `WireguardUtils` and not a `WireguardUtilsLinux`.

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "wireguardutilslinux.h"
```
bakulf, line 4: Empty line between the header and the first include

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -0,0 +1,51 @@
...
+#if defined(__cplusplus)
+}
+#endif
+// End import wireguard
+
+namespace {
+Logger logger(LOG_LINUX, "WireguardUtilsLinux");
+}
+
+WireguardUtilsLinux::WireguardUtilsLinux() : WireguardUtils() {
```
bakulf, line 22: No need to add `: WireguardUtils()`. It's implicit.

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+bool WireguardUtilsLinux::interfaceExists() {
+  // As currentInterfaces only gets wireguard interfaces, this method
+  // also confirms an interface as being a wireguard interface.
+  return currentInterfaces().contains(WG_INTERFACE);
+};
+
+QStringList WireguardUtilsLinux::currentInterfaces() {
+  char* deviceNames = wg_list_device_names();
+  QStringList devices;
```
bakulf, line 40: Move this to line 44. After the if-stmt.

src/platforms/linux/daemon/dbusservice.cpp
bakulf: The issue with this approach is that we can forget to call setParent().

bakulf (changes requested):
All good! Just some layout tuning. I would like to see the parenting of the WireguardUtils in the CTOR instead of via setParent(), because everywhere else in the code we use this approach.

The idea is that:
1. `WireguardUtils` needs a CTOR like this: `explicit WireguardUtils(QObject* parent) : QObject(parent) {}`
2. `WireguardUtilsLinux` will have: `explicit WireguardUtilsLinux(QObject* parent) : WireguardUtils(parent) {}`
3. Just to avoid DTOR issues, add a `virtual ~WireguardUtils() = default;` See: https://www.geeksforgeeks.org/virtual-destructor/

src/platforms/linux/daemon/dbusservice.h
birdsarah: Fine. I had it like this as in next PR I had added a WireguardUtilsLinux specific method

src/platforms/linux/daemon/wireguardutilslinux.cpp
birdsarah: And L42 should be QStringList() ?

src/daemon/interfaceconfig.h
birdsarah: and?

src/daemon/interfaceconfig.h
birdsarah: wierd

src/platforms/linux/daemon/dbusservice.cpp
birdsarah: I agree can forget to call setParent. In the other case couldn't I forget to call delete? Just parenting to QObject (via WireguardUtils) doesn't ensure object is destroyed (or maybe I got something wrong when I tried this before). Or are you thinking I make the parent Daemon in WireguardUtils.

birdsarah: Or is the idea that when I call `new WireguardUtilsLinux`, I call it with `new WireguardUtilsLinux(this)` where this is dbusservice and that way I'm forced to call use a `this` so I won't forget.

bakulf (approved):

PR #633, opened 2021-03-01
Implement core wg pieces in WireguardUtilsLinux

birdsarah:
Note: This PR is chained off #632.

This PR implements almost all the wireguard calls from `linux/daemon/helper.sh`.

The remaining wg commands are:
* `wg set "$INTERFACE" fwmark $table` - we need to get `$table` from interacting with `ip route` so that will be done later.
* `wg show` - these are all just picking up properties from the interface as the script runs standalone, we won't need to do this when we've implemented all the pieces as we will have that data in config.

Open question: I do not believe the constraints for "supportServerSwitching" need to be so strict now that the serverSwitch configures a whole new peer. We could change that in this PR or open a new issue.


src/daemon/daemon.cpp
```diff
@@ -88,6 +88,14 @@ bool Daemon::activate(const InterfaceConfig& config) {
       qWarning("Wireguard interface `%s` already exists.", WG_INTERFACE);
       return false;
     }
+    // add_if
+    if (!wgutils()->addInterface()) {
+      return false;
+    }
+    // set conf
+    if (!wgutils()->configureInterface(config)) {
```
bakulf, line 96: should we remove the interface if the configuration fails?

src/daemon/daemon.cpp
```diff
@@ -241,6 +249,12 @@ bool Daemon::deactivate(bool emitSignals) {
   m_connected = false;
   bool status = run(Down, m_lastConfig);
 
+  if (supportWGUtils()) {
```
bakulf, line 252: `if (supportWgUtils() && !wgUtils()->deleteInterface()) `

src/platforms/linux/daemon/wgutilslinux.cpp
```diff
@@ -24,6 +33,82 @@ bool WireguardUtilsLinux::interfaceExists() {
...
+  return true;
+}
+
+WireguardUtils::peerBytes WireguardUtilsLinux::getThroughputForInterface() {
+  uint64_t txBytes = 0;
+  uint64_t rxBytes = 0;
+  wg_device* device = nullptr;
+  wg_peer* peer;
+  peerBytes pb;
+  wg_get_device(&device, WG_INTERFACE);
```
bakulf, line 100: Let's check the return value here. If it returns a failure, let's return a 0,0.

bakulf (changes requested):
This looks great! We should probably remove the interface at the startup of the daemon in case something went wrong during a previous execution (a crash maybe)

src/daemon/daemon.cpp
birdsarah: YES! And the bash script does this with the "trap" and I forgot. Thanks.

birdsarah:
I have made the changes.

> This looks great! We should probably remove the interface at the startup of the daemon in case something went wrong during a previous execution (a crash maybe)

I have implemented this in commit: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/633/commits/36aea6aa0ffb261f529cba13a10c4c32b2b22dc7.  I am doing this in dbus not in parent daemon as other platforms did not have this implemented yet. I was thinking I could move this logic and the requirement for a `removeInterfaceIfExists` up to parent daemon, which makes sense to me as an aspiration even if other platforms haven't implemented yet.

birdsarah:
This PR is still off #632. I'll update once #632 is merged.

src/daemon/daemon.cpp
birdsarah: I'm no longer sure about this change. In the `if(m_connected)` block above, if something goes wrong we call deactivate. deactivate attempts to delete the interface. What about continuing to use that pattern?

bakulf (changes requested):
I'm not sure about the `deactivate` when something goes wrong. Can you tell me something more?

src/daemon/wireguardutils.h
```diff
@@ -0,0 +1,55 @@
...
+class WireguardUtils : public QObject {
+ public:
+  struct peerBytes {
+    double txBytes, rxBytes;
+  };
+
+  explicit WireguardUtils(QObject* parent) : QObject(parent){};
+  virtual ~WireguardUtils() = default;
+
+  virtual bool interfaceExists() {
```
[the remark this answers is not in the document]
birdsarah: I don't think that works, because Daemon.cpp calls the methods under if guards, but they're not available on MacOS so MacOS builds fail.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -52,20 +41,14 @@ void DBusService::setAdaptor(DbusAdaptor* adaptor) {
   m_adaptor = adaptor;
 }
 
-bool DBusService::checkInterface() {
-  logger.log() << "Checking interface";
-
-  wg_device* device = nullptr;
-  if (wg_get_device(&device, WG_INTERFACE) == 0) {
+bool DBusService::removeInterfaceIfExists() {
+  if (wgutils()->interfaceExists()) {
```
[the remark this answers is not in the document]
birdsarah: Like: https://www.geeksforgeeks.org/multiple-inheritance-in-c/

Seems nice from that code snippet. I'll give it a try.

src/daemon/daemon.cpp
birdsarah: You're correct. We call `deactivate` on L77 (https://github.com/mozilla-mobile/mozilla-vpn-client/pull/633/files#diff-d93d28344d209ab944d30f1b5438025d11fdcf98a15cb00588e0c219f46d08fbR77) if connected and don't support server switching. I agree that's not the same as `activate` failing. But `activate` could try to call `deactivate` if it fails. I'm just thinking about where do we want clean-up logic to be and whether it's going to end up being duplicated.

I'm fine leaving this as it is for now and then we can review it later if we start to get a lot of similar looking code in activate and deactivate.

src/platforms/linux/daemon/dbusservice.cpp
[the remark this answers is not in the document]
birdsarah: One reason not to do this is that the original code mirrors the code in daemon.cpp which keeps everything looking the same, which is nice.

birdsarah:
Superceded by #666 which includes all corrections from review.

PR #635, opened 2021-03-01
Breakout ip utils and implement for linux

birdsarah:
Note: This PR is chained off #633.

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -97,6 +97,11 @@ bool Daemon::activate(const InterfaceConfig& config) {
       return false;
     }
   }
+  if (supportIPUtils()) {
+    if (!iputils()->addInterfaceIPs(config)) {
+      return false;
+    }
+  }
 
   m_lastConfig = config;
   m_connected = run(Up, m_lastConfig);
--- a/src/platforms/linux/daemon/iputilslinux.cpp
+++ b/src/platforms/linux/daemon/iputilslinux.cpp
@@ -1,0 +1,103 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "iputilslinux.h"
+#include "daemon/wgutils.h"
+#include "logger.h"
+
+#include <arpa/inet.h>
+#include <net/if.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include <QHostAddress>
+
+namespace {
+Logger logger(LOG_LINUX, "IPUtilsLinux");
+}
+
+bool IPUtilsLinux::addInterfaceIPs(const InterfaceConfig& config) {
+  // TODO - These are linux specific
+  if (!addIP4AddressToDevice(config)) {
+    return false;
+  }
+  if (config.m_ipv6Enabled) {
+    if (!addIP6AddressToDevice(config)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool IPUtilsLinux::addIP4AddressToDevice(const InterfaceConfig& config) {
+  struct ifreq ifr;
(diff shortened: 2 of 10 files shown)


src/daemon/daemon.cpp
```diff
@@ -98,6 +98,14 @@ bool Daemon::activate(const InterfaceConfig& config) {
       return false;
     }
   }
+  if (supportIPUtils()) {
+    if (!iputils()->addInterfaceIPs(config)) {
```
birdsarah, line 102: The clean-up for this is just to delete the interface. That is done with wgutils. I think it's okay to not try and clean this up if it fails. We return false which means we know that activation failed. Thoughts?

birdsarah: We could change the whole structure of activate to try removing the interface if activation fails and if removing the interface fails, we could bail out altogether with a qFatal?

birdsarah: In an ideal world I'm thinking that wgutils and iputls can be added seperately and we don't need to have one to have the other. But maybe that's not a thing to really worry about.

birdsarah:
This PR is ready for review @bakulf. It's another chained PR. Hopefully we can start merging the predecessors. 

Once you've had a look, I'd like to review this code with you on a call. It brings up a number of questions for me to help my learning about C++. e.g. When do things need to be or not need to be freed? (I can read things like https://www.geeksforgeeks.org/g-fact-30/ but I'm not necessarily fully internalizing it.)

src/platforms/linux/daemon/iputilslinux.cpp
```diff
@@ -0,0 +1,137 @@
...
+    if (!addIP6AddressToDevice(config)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool IPUtilsLinux::setMTUAndUp() {
+  // Create socket file descriptor to perform the ioctl operations on
+  int sockfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
```
bakulf, line 46:
```
if (sockfd < 0) {
  logger.log() << "something here";
  return false;
}
```

src/platforms/linux/daemon/iputilslinux.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  // Get the device address to add to interface
+  QPair<QHostAddress, int> parsedAddr =
+      QHostAddress::parseSubnet(config.m_deviceIpv4Address);
+  QByteArray _deviceAddr = parsedAddr.first.toString().toLocal8Bit();
+  char* deviceAddr = _deviceAddr.data();
+  inet_pton(AF_INET, deviceAddr, &ifrAddr->sin_addr);
+
+  // Create IPv4 socket to perform the ioctl operations on
+  int sockfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
```
bakulf, line 88: this can fail.

src/platforms/linux/daemon/iputilslinux.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  // Get the device address to add to ifr6 interface
+  QPair<QHostAddress, int> parsedAddr =
+      QHostAddress::parseSubnet(config.m_deviceIpv6Address);
+  QByteArray _deviceAddr = parsedAddr.first.toString().toLocal8Bit();
+  char* deviceAddr = _deviceAddr.data();
+  inet_pton(AF_INET6, deviceAddr, &ifr6.addr);
+
+  // Create IPv6 socket to perform the ioctl operations on
+  int sockfd = socket(AF_INET6, SOCK_DGRAM, IPPROTO_IP);
```
bakulf, line 114: here too

bakulf (approved):

birdsarah:
As soon as #666 is merged. I will rebase this PR, add the correction to check for socket creation failure `if (sockfd < 0) {` and merge.

birdsarah:
Closing this PR with a clean PR off #666.

PR #639, opened 2021-03-03
Keep en-lang file in tree and remove generator comments

strseb:
Helloo! We've been talking about how it's hard to keep track of what strings already exists in the project and where they are defined. So my idea was, instead of having one qml/c++ file that has all the strings to generate the en.ts once and keep it in tree. Also i've removed all the obsolete comments now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ xcode.xconfig
 .gradle/
 
 *.ts
+!translations/mozillavpn_en.ts
 tests/unit/tests
 build
 Debug/
--- a/scripts/importLanguages.py
+++ b/scripts/importLanguages.py
@@ -85,10 +85,6 @@ with open('translations/translations.pri', 'w') as pri_file:
 print('Updated translations.pri')
 
 # Step 3
-# Generate new ts files
-os.system('lupdate src/src.pro')
-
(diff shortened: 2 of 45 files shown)


flodolo:
Sorry, this is not a change that we can do at this point, it should have been done months ago.

The generated XLIFF file is going to have a completely different structure, which means wel lose all existing translations, and any history or pending suggestions in Pontoon. And we're talking about 3500 translations.

flodolo:
> Sorry, this is not a change that we can do at this point, it should have been done months ago.

Correction: it can be done, if the resulting XLIFF file has the same structure of the current output.

This is currently creating one `file` element, and put all the strings in it. The existing XLIFF file has different files based on the location element in the ts file.

flodolo:
If you want to test this further, just keeping the first location for each string (even without line numbers) might work. 

```
    <message id="vpn.main.back">
        <location filename="../src/platforms/android/androidauthenticationview.qml" />
        <source>Back</source>
        <extracomment>Go back</extracomment>
    </message>
```

Then new strings could probably not have a location, and end up in a file with empty filename in .XLIFF.

As said, the wanted result here is having the same XLIFF file as the current code.

PR #645, opened 2021-03-04
Capitalize the language names

bakulf:


src/localizer.cpp
```diff
@@ -16,6 +16,17 @@
...
+struct StaticLanguage {
+  QString m_name;
+  QString m_localizedName;
+};
+
+// Some languages do not have the right localized/non-localized names in the QT
+// framework (and some are missing entirely). This static map is the fallback
+// when this happens.
+QMap<QString, StaticLanguage> s_languageMap{
+    {"co", StaticLanguage{"Corso", ""}}};
```
flodolo, line 29: This should be "Corsu"

flodolo (approved):
With the name fixed ;-)

PR #674, opened 2021-03-15
MacOS DNS recovering after crash

bakulf:
This is a temporary fix for the next release until we have a better daemon implementation. It turned out that macOS leaks VPN DNSs randomly. I suspect there is a crash or a restart of the daemon. When this happens, the user is unable to navigate because the DNSs are wrongly set.
This is a temporary fix that stores the DNSs in a Settings and remove them at startup if found.

If somebody has time to work on a better fix, I'm happier :)

### DIFF
--- a/macos/daemon/helper.sh
+++ b/macos/daemon/helper.sh
@@ -216,7 +216,7 @@ collect_new_service_dns() {
 		[[ $service == "*"* ]] && service="${service:1}"
 		found_services="$found_services;$service"
 		array_contains "$SERVICE_DNS" "$service" && continue
-		get_response="$(cmd networksetup -getdnsservers "$service")"
+		get_response="$(cmd networksetup -getdnsservers "$service" |tr '\n' '-')"
 		[[ $get_response == *" "* ]] && get_response="Empty"
 		[[ -n $get_response ]] && SERVICE_DNS="$SERVICE_DNS;$service=$get_response"
 		get_response="$(cmd networksetup -getsearchdomains "$service")"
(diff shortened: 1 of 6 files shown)


src/platforms/macos/daemon/macosdaemon.cpp
```diff
@@ -14,9 +14,14 @@
 #include <QJsonValue>
 #include <QLocalSocket>
 #include <QProcess>
+#include <QSettings>
 #include <QTextStream>
 #include <QtGlobal>
 
+constexpr const char* LATEST_IPV4GATEWAY = "latestServerIpv4Gateway";
```
birdsarah, line 21: why isn't this a QString?

macos/daemon/helper.sh
```diff
@@ -216,7 +216,7 @@ collect_new_service_dns() {
 		[[ $service == "*"* ]] && service="${service:1}"
 		found_services="$found_services;$service"
 		array_contains "$SERVICE_DNS" "$service" && continue
-		get_response="$(cmd networksetup -getdnsservers "$service")"
+		get_response="$(cmd networksetup -getdnsservers "$service" |tr '\n' '-')"
```
birdsarah, line 219: these changes seem unrelated to `cmd_recover`, and my bash isn't good enough to just know what they're doing. Can you elaborate?

src/platforms/macos/daemon/macosdaemonserver.cpp
```diff
@@ -41,6 +41,8 @@ int MacOSDaemonServer::run(QStringList& tokens) {
 
   MacOSDaemon daemon;
 
+  daemon.maybeRecover();
```
birdsarah, line 44: I don't love the name `maybeRecover`. You're doing a `cleanUpIfNecessary`. Not that that's much better. `maybeCleanup`? It's just not really a recovery so much as a reset or fresh start.

birdsarah (approved):
just a couple of small comments/questions. but all makes sense to me. don't have a mac to test, but presumably you have.

src/platforms/macos/daemon/macosdaemon.cpp
bakulf: No big reasons. These strings are used in `QSettings::setValue` (https://doc.qt.io/qt-5/qsettings.html#setValue) as first argument and QString() will be created from `const char*`.

macos/daemon/helper.sh
bakulf: Right! But this fixes a related DNS issue. The issue was that, if the device, before enabling the VPN, has 2 DNS entries, we were treated them as "Empty" because of this: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/macos/daemon/helper.sh#L223
Let's say that this is related bug fix.

PR #678, opened 2021-03-15
WIP: Android captive portal

strseb:
This sadly is kinda independent from our current captive portal code, as it needs to run even while the app is inactive. 
So what we do, we 
- register a network-change listener
- Invoke detectPortal() when we have a change on a wifi connection
-  detectPortal() does the same as the crossplattform captiveportalImpl ... it's just on the background-service
- In case of a portal we create an actionable-notification.
- If the user clicks it we will open the app and auto-deactivate the vpn.


strseb:
Converting to Draft as this lacks basic functionality we have in the client now: 
-302 Check. 
-Retry in case of routing failures

PR #690, opened 2021-03-16
Replace logcat reading with a log-file

### LINKED ISSUE #685: “View Log” is empty after WI-FI network switch
#### Affected versions:
- v. 2.0.4 (16155440) last build; 

#### Affected Platforms:
- Android 11;

#### Prerequisites:
- Be connected to a wifi network;

#### Steps to reproduce:
1. Install a new copy of VPN app;
2. Go to Settings - Get Help;
3. Tap View Log
4. Observe content;
5. Switch to another wifi network;
6. Repeat 2-4 STRs;
7. Optional - switch to mobile data;

#### Expected result:
- The View Log page has info.

#### Actual result:
- The View Log page is empty;

#### Notes
- you must use a network that will trigger VPN No Signal state or if possible switch to mobile data.

strseb:
Closes #685

Logcat has a few issues. If the call fails and the callback is never done we'll just show a white page. 
Also logcat does not keep logs as long as we want to, so i want to replace it with a drop-in replacement that also stores everything to disk aswell as printing to logcat. - Just like the qt version. 
This should be more robust as we can just catch i/o errors :)

### DIFF
--- a/android/src/com/mozilla/vpn/VPNLogger.kt
+++ b/android/src/com/mozilla/vpn/VPNLogger.kt
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.vpn
+
+import android.content.Context
+import java.io.File
+import java.time.LocalDateTime
+import android.util.Log as nativeLog
+
+/*
+ * Drop in replacement for android.util.Log 
+ * Also stores a copy of all logs in tmp/mozilla_deamon_logs.txt
+*/
+class Log {
+    private var file: File
+    private constructor(context: Context) {
+        val tempDIR = context.cacheDir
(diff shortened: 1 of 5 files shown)


android/src/com/mozilla/vpn/VPNLogger.kt
```diff
@@ -0,0 +1,59 @@
...
+    private var file: File
+    private constructor(context: Context) {
+        val tempDIR = context.cacheDir
+        file = File(tempDIR, "mozilla_deamon_logs.txt")
+    }
+
+    companion object {
+        var instance: Log? = null
+        fun init(ctx: Context) {
+            if (instance == null) {
```
bakulf, line 26: question: is "instance" a static thing? because if yes, we cannot Log.init(a), Log.init(b), right?

bakulf (approved):

android/src/com/mozilla/vpn/VPNLogger.kt
strseb: Yes its static -so init(a) would do init create the instance and init(b) would do nothing.

PR #715, opened 2021-03-19
Suspend Connection Health when the app is inactive

strseb:
Stop ConnectionHealth when the app goes into an inactive state and resume once the app is back up - so we're not getting a "no connection" after every resume :)


bakulf:
Nice! Thanks!

PR #719, opened 2021-03-19
Linux settings for the PPA (staging vs prod)

bakulf:


strseb (approved):
Looks reasonable. Not an expert in ppa files tho.

PR #723, opened 2021-03-19
Add a dummy key for android-debug builds

### LINKED ISSUE #718: CI: Add a Github-Secret keystore for debug-signing
In every CI run,since we have no keystore, the SDK generates new random debug-signing-keys and uses them to sign our debug.apk's. This makes upgrading between stage-ci builds impossible since the keys are different, which means stage testers need to uninstall, download, install each apk. 
it would be great to have a generic ~/.android/debug.keystore that is constant. we might also just keep it in tree, if that is okay, so devs can do builds that are at interchangable with the ci builds aswell :)

strseb:
I would like to have our Android-Debug builds to use the same keystore. Currently debug builds are using a self generated keystore, for each compiling device / ci job. This makes it impossible to upgrade debug-builds from diffrent devs or ci builds. Since the certs don't match then. 
This key is not used to sign anything meaningful and also can't be used for anything else then debug builds.
So we might aswell just have it in tree, i think. 
Closes #718


strseb:
In case we don't want this to be public for everybody i would still like a variation of this for the CI. 
Somebody with Github Actions Secret access could:
-> Generate a new key with some secret password
-> change the in build.gradle the key password to `System.getenv("SOME_GITHUB_CI_ENV")`

So we have at least a CI->CI upgrade path :)

bakulf (approved):

PR #730, opened 2021-03-22
Fix random timout banner

### LINKED ISSUE #495: “No internet connection. Try again!” red banner appears after a few seconds from turning on the VPN
#### Affected versions:
- 2.0.3 beta-debug build; 

#### Affected Platforms:
- All Android;

#### Prerequisites:
- Have the VPN app installed;

#### Steps to reproduce:
1. Open the VPN app;
2. Tap to turn ON the VPN;
3. Wait about 10-15 seconds;

#### Expected result:
- Nothing strange happens.

#### Actual result:
- The “No internet connection. Try again!” red banner is displayed.

#### Notes:
- while the banner is being displayed, the VPN is still ON.

https://user-images.githubusercontent.com/15685960/105497416-79676a80-5cc7-11eb-9b3c-d294d65452cd.mp4

strseb:
Closes #495 

Issue here is: 
We have a task that starts an http-request.  (e.g. account&Servers)
We connect vpn.
The Request times out, as expected.  
We show the no connection banner.

My idea: Add a new connection-error for things that might be caused by vpn and are okay to ignore if the vpn is connected.

### DIFF
--- a/src/errorhandler.cpp
+++ b/src/errorhandler.cpp
@@ -38,7 +38,7 @@ ErrorHandler::ErrorType ErrorHandler::toErrorType(
     case QNetworkReply::HostNotFoundError:
       [[fallthrough]];
     case QNetworkReply::TimeoutError:
-      [[fallthrough]];
+      return VPNDependentConnectionError;
     case QNetworkReply::UnknownNetworkError:
       // On mac, this means: no internet
       // On Android check if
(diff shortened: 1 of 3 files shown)


src/errorhandler.cpp
```diff
@@ -38,7 +38,7 @@ ErrorHandler::ErrorType ErrorHandler::toErrorType(
     case QNetworkReply::HostNotFoundError:
       [[fallthrough]];
     case QNetworkReply::TimeoutError:
```
bakulf, line 40: This is wrong. Host-Not-Found should not be treated at "VPNDepedentConnectionError' right?

bakulf (changes requested):

src/errorhandler.cpp
strseb: Right, that was not supposed to be changed.

PR #736, opened 2021-03-23
Captive-portal detection - support for redirect

bakulf:


src/networkrequest.cpp
```diff
@@ -205,7 +205,7 @@ NetworkRequest* NetworkRequest::createForCaptivePortalDetection(
     QObject* parent, const QUrl& url, const QByteArray& host) {
   Q_ASSERT(parent);
 
-  NetworkRequest* r = new NetworkRequest(parent, 200);
+  NetworkRequest* r = new NetworkRequest(parent, 0);
```
strseb, line 208: What about 40X-50X response codes, are those still then handles as an error cases?

src/captiveportal/captiveportalrequest.cpp
```diff
@@ -76,20 +76,28 @@ void CaptivePortalRequest::createRequest(const QUrl& url) {
...
 
             --m_pendingRequests;
             m_completed = true;
             deleteLater();
 
+            // Usually, captive-portal pages do a redirect to an internal page.
+            if (request->statusCode() != 200) {
+              logger.log() << "Captive portal detected. Expected 200, received:"
+                           << request->statusCode();
+              emit completed(true);
```
strseb, line 90: Maybe we should also check that there really is something to redirect to e.g
```
!request->rawHeader("Location").isEmpty();
```

bakulf: What do you want to do with the location?

strseb (approved):

PR #738, opened 2021-03-23
Captive portal queue

strseb:
Here is my idea for the retry: 
1: Put all captive portal requests in a queue. 
2: After a small delay, fire the first request in the queue
 2.2: If we get a result, we pass that on and exit. 
 2.3: On error, put it back in the queue and go to 2.

If we can't confirm that there is *no* portal in 30s e.g like my telecom hotspot that just blocks detectportal.firefox.com, show a notification that a portal might affect connectivity.

### DIFF
--- a/src/captiveportal/captiveportalrequest.cpp
+++ b/src/captiveportal/captiveportalrequest.cpp
@@ -76,12 +76,20 @@ void CaptivePortalRequest::createRequest(const QUrl& url) {
           });
 
   connect(request, &NetworkRequest::requestCompleted,
-          [this](const QByteArray& data) {
+          [this, request](const QByteArray& data) {
             logger.log() << "Captive portal request completed:" << data;
 
             --m_pendingRequests;
             m_completed = true;
             deleteLater();
+
+            // Usually, captive-portal pages do a redirect to an internal page.
+            if (request->statusCode() != 200) {
+              logger.log() << "Captive portal detected. Expected 200, received:"
+                           << request->statusCode();
+              emit completed(true);
+              return;
+            }
 
             if (QString(data).trimmed() == CAPTIVEPORTAL_REQUEST_CONTENT) {
               logger.log() << "No captive portal!";
@@ -89,7 +97,7 @@ void CaptivePortalRequest::createRequest(const QUrl& url) {
               return;
             }
 
-            logger.log() << "Captive portal detected!";
+            logger.log() << "Captive portal detected. Content does not match.";
             emit completed(true);
           });
 }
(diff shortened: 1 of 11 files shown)


src/captiveportal/captiveportalrequest.cpp
```diff
@@ -61,46 +62,48 @@ void CaptivePortalRequest::run() {
 }
 
 void CaptivePortalRequest::createRequest(const QUrl& url) {
+  if (m_completed) {
```
bakulf, line 65: We are single-thread. This cannot really happen. Am I right?

src/captiveportal/captiveportalrequest.cpp
```diff
@@ -61,46 +62,48 @@ void CaptivePortalRequest::run() {
...
   logger.log() << "request:" << url.toString();
 
-  ++m_pendingRequests;
-
   NetworkRequest* request = NetworkRequest::createForCaptivePortalDetection(
       this, url, CAPTIVEPORTAL_HOST);
 
   connect(request, &NetworkRequest::requestFailed,
-          [this](QNetworkReply::NetworkError error, const QByteArray&) {
+          [this, url](QNetworkReply::NetworkError error, const QByteArray&) {
```
bakulf, line 74: You are not using 'url'.

src/captiveportal/captiveportalrequest.h
```diff
@@ -19,15 +22,16 @@ class CaptivePortalRequest final : public QObject {
...
  private:
   void createRequest(const QUrl& url);
-  void maybeComplete();
+  void nextStep();
+  void onResult(CaptivePortalResult portalDetected);
 
  private:
-  uint32_t m_pendingRequests = 0;
   bool m_completed = false;
+  QQueue<QUrl> m_requestQueue = QQueue<QUrl>();
```
bakulf, line 34: Remove this.

src/networkrequest.cpp
```diff
@@ -205,7 +205,7 @@ NetworkRequest* NetworkRequest::createForCaptivePortalDetection(
     QObject* parent, const QUrl& url, const QByteArray& host) {
   Q_ASSERT(parent);
 
-  NetworkRequest* r = new NetworkRequest(parent, 200);
+  NetworkRequest* r = new NetworkRequest(parent, 0);
```
bakulf, line 208: Do you have to rebase the tree?

bakulf (changes requested):

strseb:
I've updated the pr to adress code-related the stuff. @bakulf did we have a decision so far on the whole CaptivePortalResult::PortalPossible: state? Should i remove it ?

PR #739, opened 2021-03-24
Use main in import strings workflow

flodolo:


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "i18n"]
 	path = i18n
 	url = https://github.com/mozilla-l10n/mozilla-vpn-client-l10n.git
-	branch= master
+	branch= main
```
flodolo, line 4: Technically, I think we could remove this if we just want the default branch?

flodolo:
Note: unfortunately I can't really test automation from a fork :-(

.gitmodules
```diff
@@ -1,7 +1,6 @@
 [submodule "i18n"]
 	path = i18n
 	url = https://github.com/mozilla-l10n/mozilla-vpn-client-l10n.git
-	branch= master
```
bakulf, line 4: I would add `branch = main` because, by default, git uses "master".

bakulf (approved):

flodolo:
Should we try merging and check if automation works as expected? Alternatively, I can push to a branch in this repo and verify it there.

flodolo:
Seems to work https://github.com/mozilla-mobile/mozilla-vpn-client/pull/742

PR #740, opened 2021-03-24
Add comments and fix tooltip for the language switcher toggle

flodolo:


src/ui/settings/ViewLanguage.qml
```diff
@@ -120,7 +125,8 @@ Item {
                     objectName: "settingsSystemLanguageToggle"
                     toolTipTitle: {
                         if (checked) {
-                           //% "Disable to select different language"
+                           //% "Disable to select a different language"
```
flodolo, line 128: Not sure if it's just me, but "Disable to select different language" sounds a bit off.

PR #744, opened 2021-03-24
Separate thread for linux-network-watcher

bakulf:
The reason why this is needed is that a couple of QT dbus methods do I/O on main-thread and, changing network, it takes a few seconds in which the UI is frozen.


birdsarah (approved):
Looks good. Nice and clean. 

Only thing is that I don't yet understand where/how the new m_thread gets created, but I'm guessing that it's Qt taking care of it behind the scenes.

bakulf:
Right. The m_thread is created when started. Thanks for the review.

PR #747, opened 2021-03-24
Revert the system-language code

bakulf:
This adds extra complexity for the QA team, but it has bugs on windows.
Let's re-include this code after the next release.


lesleyjanenorton (approved):

src/ui/settings/ViewLanguage.qml
```diff
@@ -62,110 +53,13 @@ Item {
...
-                    id: labelWrapper
-                    spacing: 4
-                    Layout.maximumWidth: parent.width - useSystemLanguageToggle.width - 16
-
-                    states: [
-                        State {
-                            when: useSystemLanguageToggle.checked
-                            PropertyChanges {
-                                target: labelDescription
-                                //% "Mozilla VPN will use your system’s default language."
```
flodolo, line 87: Is there any chance to just leave the strings and comments in the file, so we don't lose translations and are ready when this get merged back?

bakulf: Yes, done. Take a look again please.

flodolo (approved):

PR #755, opened 2021-03-25
es_AR vs es_MX

bakulf:


src/localizer.cpp
```diff
@@ -26,7 +26,10 @@ struct StaticLanguage {
 // framework (and some are missing entirely). This static map is the fallback
 // when this happens.
 QMap<QString, StaticLanguage> s_languageMap{
-    {"co", StaticLanguage{"Corsu", ""}}};
+    {"co", StaticLanguage{"Corsu", ""}},
+    {"es_AR", StaticLanguage{"Spanish (Argentina)", "Español, Argentina"}},
```
flodolo, line 30: Is there a reason to go with the comma, instead of `Español (Argentina)`?

I don't mind, as long as we are consistent, at least among Spanish variants.

flodolo (approved):

src/localizer.cpp
bakulf: no valid reasons.

PR #760, opened 2021-03-29
Language

### LINKED ISSUE #754: VPN Languages are not alphabetically ordered
**Affected versions:**
- v. 2.1.0;

**Affected platforms**
- Windows 10

**Prerequisites:**
- Have the VPN app installed;

**Steps to reproduce**
1. Open the VPN 
2. Go to Settings -> Language.
3. Check how the available languages are displayed.

**Expected results**
- The languages are alphabetically ordered.

**Actual Results**
- The available languages are randomly displayed.

Screenshot_21

bakulf:
- Closes #754
- better english language names

### DIFF
--- a/src/localizer.cpp
+++ b/src/localizer.cpp
@@ -29,6 +29,8 @@ QMap<QString, StaticLanguage> s_languageMap{
     {"co", StaticLanguage{"Corsu", ""}},
     {"es_AR", StaticLanguage{"Spanish (Argentina)", "Español, Argentina"}},
     {"es_MX", StaticLanguage{"Spanish (Mexico)", "Español, México"}},
+    {"en_GB", StaticLanguage{"English (United Kingdom)", ""}},
+    {"en_CA", StaticLanguage{"English (Canada)", ""}},
 };
 
 }  // namespace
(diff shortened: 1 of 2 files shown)


src/localizer.cpp
```diff
@@ -185,15 +195,29 @@ QVariant Localizer::data(const QModelIndex& index, int role) const {
...
+  for (const Language& language : m_languages) {
+    languages.append(language.m_code);
+  }
+
+  return languages;
+}
+
+bool Localizer::languageSort(const Localizer::Language& a,
+                             const Localizer::Language& b) {
+  return a.m_localizedName < b.m_localizedName;
```
bakulf, line 222: I would like to use QCollator but it doesn't work on Linux as it should.
If this doesn't work, I can investigate why we have a crash when using QCollator

src/localizer.cpp
```diff
@@ -29,7 +29,10 @@ QMap<QString, StaticLanguage> s_languageMap{
     {"co", StaticLanguage{"Corsu", ""}},
     {"es_AR", StaticLanguage{"Spanish (Argentina)", "Español, Argentina"}},
     {"es_MX", StaticLanguage{"Spanish (Mexico)", "Español, México"}},
+    {"en_GB", StaticLanguage{"English (United Kingdom)", ""}},
```
flodolo, line 32: Maybe just UK, to avoid the language getting too long?

flodolo (approved):

src/localizer.cpp
bakulf: We have space enough because the language is in a separate line.
image

PR #790, opened 2021-04-07
Remove 32bit and x86 support for windows

bakulf:


strseb (approved):
🎉

strseb:
Small side note:
This would mean we would also drop windows-on-arm support until w10 Build 21277 is released, as this adds x64->arm emulation.

PR #817, opened 2021-04-08
 FVP-02 WP5: Harden android

### LINKED ISSUE #803: FVP-02-008 WP5: Android app allows backups of application data
The allowBackup property in the AndroidManifest.xml file specifies if the data pertinent
to the apps can be backed up.2

Without setting the android:allowBackup flag to false, the
backup feature is enabled by default. In case an attacker is able to send adb commands
to user-phones, they could get access to all of the stored data from the protected data
folders, inclusive of the VPN configuration data.
Affected File:
`android/AndroidManifest.xml`
As this feature does not require a rooted phone, disallowing backups completely should
be considered. Due to the fact that an absence of the flag will set it to true by default, it is
recommended to explicitly set the allowBackup flag to false within the application tag.

### LINKED ISSUE #804: FVP-02-009 WP5: Secure flag missing on views for Android app
During the assessment of the Android app, the discovery was made that the
FLAG_SECURE security flag is not deployed to protect views that display sensitive
content. By applying the flag for Android views, the app’s windows can no longer be
manually “screenshotted”. Additionally, the items would be excluded from automatic
screenshots or screen-recordings, which ultimately prevents screen-data from leakage
to alternative apps.
Particularly for the implemented views displaying sensitive data, e.g. during the login
process, adding this flag is important. An attacker would otherwise be able to steal
sensitive data, such as login credentials or personal information, doing it after it has
been displayed. Hence, this information could be stolen from the services via a malicious
app.
It is recommended to add the FLAG_SECURE within the WindowManager responsible
for handling views like the WebView. The flag can be set via WindowManager.
LayoutParams, i.e. as FLAG_SECURE within the function of setFlags(). As for additional
information on how to prevent this type of attacks, please refer to the OWASP Mobile
Security Testing Guide

### LINKED ISSUE #805: FVP-02-010 WP5: Android app supports insecure v1 signature
The discovery was made that the provided Android staging and production builds are
signed with an insecure v1 APK signature. Using the insecure v1 signature makes the
app prone to the known Janus4

vulnerability on devices running Android < 7. The
problem lets attackers smuggle malicious code into the APK without breaking the
signature. At the time of writing, the app supports a minimum SDK of 21 (Android 5),
which only uses the v1 signature and is, hence, vulnerable to this attack.
The existence of this flaw means that attackers could trick users into installing a
malicious attacker-controlled APK which matches the v1 APK signature of the Mozilla
VPN Android application. As a result, a transparent update would be possible without
warnings appearing in Android, effectively taking over the existing application and all of
its data. It is recommended to increase the minimum supported SDK level to at least 24
(Android 7) to ensure that this known vulnerability cannot be exploited on devices
running older Android versions. In addition, the production builds should only be shipped
with v2 and v3 APK signatures.

### LINKED ISSUE #808: FVP-02-012 WP5: Unencrypted shared preferences
During the assessment of the Android app, the discovery was made that the application
does not always consistently use the encrypted shared preference feature provided by
the Android SDK. This may lead to an information disclosure in case a local attacker is
able to get root access to the phone or the data is obtainable via backups (see FVP-02-
008). Sensitive information stored within the shared_prefs data folder in plain-text, such
as user VPN IPs and private keys, could be revealed.

It is advised to use the provided wrapper class called EncryptedSharedPreferences to
encrypt sensitive data stored within the shared_prefs folder, so as to make the
application more robust against the illustrated attacks. The wrapper class uses the
Android Keystore for handling the master key and is used to encrypt/decrypt all other
keysets. For more information, please refer to the official Android guide on storing data
more securely. Additionally, it is also advised to store VPN configuration data via
encrypted shared preferences, which is actually also written to the vpn.moz file in plain-
text.

### LINKED ISSUE #809: FVP-02-013 WP5: Android app exposes sensitive data to system logs
It was found that the Android app makes frequent use of logging features to be able to
monitor events. However, this can be considered a bad practice, especially in production
environments where tokens and codes of Mozilla VPN users might be accessible by
third-parties.
In case the device is connected to the computer with debugging enabled via USB, an
attacker may be able to get access to the logs via adb logcat. From there, extraction of
user-tokens may be achievable. Note that apps with system privileges are able to
access logs directly on rooted devices.

strseb:
** Dont Merge yet. - As this would include dropping support of android <7.  **
-->This is required for the Encrypted Shared prefs and #805 - to force using the v2 APK signature. 

This should address all WP5 related issues.

Disallow Backups of Data (2ca5182)
Closes #803

Encypt Shared Prefs (830231c)
Closes #808

Make the Web-View a secured View (10c90ae)
Closes #804

Stop Posting all Log messages to Logcat on Release Builds (e3b3eba)
Closes #809

Closes #805 - As we require android 7 now.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -20,6 +20,7 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
     <supports-screens android:largeScreens="true" android:normalScreens="true" android:anyDensity="true" android:smallScreens="true" />
     <application
         android:hardwareAccelerated="true"
+        android:allowBackup="false"
         android:name="org.qtproject.qt5.android.bindings.QtApplication"
         android:label="Mozilla VPN"
         android:extractNativeLibs="true"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,6 +50,8 @@ dependencies {
     implementation 'androidx.core:core-ktx:1.1.0'
     implementation 'com.android.installreferrer:installreferrer:1.1'
     implementation 'com.wireguard.android:tunnel:1.0.20200927'
+    implementation "androidx.security:security-crypto:1.1.0-alpha03"
+    implementation "androidx.security:security-identity-credential:1.0.0-alpha02"
     coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:1.0.10"
 }
 
@@ -122,7 +124,7 @@ android {
 
     defaultConfig {
         resConfig "en"
-        minSdkVersion 21
+        minSdkVersion 24
         targetSdkVersion 30
         versionCode  System.getenv("VERSIONCODE").toInteger()
         versionName  System.getenv("SHORTVERSION")
--- a/android/src/org/mozilla/firefox/vpn/BootReceiver.kt
+++ b/android/src/org/mozilla/firefox/vpn/BootReceiver.kt
@@ -18,7 +18,7 @@ class BootReceiver : BroadcastReceiver() {
             Log.i(TAG, "This device does not support start on boot - exit")
             return
         }
-        val prefs = context.getSharedPreferences("org.mozilla.firefox.vpn.prefrences", Context.MODE_PRIVATE)
+        val prefs = Prefs.get(context)
         val startOnBoot = prefs.getBoolean("startOnBoot", false)
         if (!startOnBoot) {
             Log.i(TAG, "This device did not enable start on boot - exit")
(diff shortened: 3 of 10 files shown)


android/src/com/mozilla/vpn/VPNLogger.kt
```diff
@@ -33,14 +34,17 @@ class Log {
         }
         fun i(tag: String, message: String) {
             instance?.write("[info] - ($tag) - $message")
+            if (BuildConfig.DEBUG) { return; }
```
bakulf, line 37: is it not the opposite? `if (!BuildConfig.DEBUG)`

bakulf (approved):

PR #820, opened 2021-04-09
Surveys

bakulf:
This PR includes:
- survey model
- unit tests
- the triggering survey logic

We still need to:
- define the strings for the notification
- populate the server component with the list of real surveys (so far there are just 2 surveys for testing)


strseb:
Code looks good but I have 2 things here: 

This is currently using SystemTrayNotificationHandler/QSystemTrayIcon to push the notification, which would make this only work on Desktop right? 
If so, maybe we should extend the cross plattform NotificationHandler interface with the ability to callback if a certain notification is called and then put it ontop of that so android/ios also get some of the survey fun :) 

Or! how about making this a "update available" level notification in app? Just worried because as you've said there are people already complaining about the switching system-notification 🙈

strseb (approved):

bakulf:
There is something wrong with the new UI. The survey disappears after a few seconds.

src/ui/components/VPNAlerts.qml
```diff
@@ -0,0 +1,47 @@
...
+        alertLinkText: qsTrId("vpn.updates.updateNow")
+        isLayout: true
+    }
+
+    VPNAlert {
+        id: surveyAlert
+
+        isLayout: true
+        alertType: "survey"
+        //% "TBD"
```
flodolo, line 38: Please don't do this? There's a pipeline attached to this project, and now I have "TBD" strings queued for merge and translation?
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/133/files

PR #841, opened 2021-04-13
Do not commit i18n, instead pull on every build

strseb:
As discussed on slack, lets not have to manually keep that fresh, instead pull incoming language changes on every build


bakulf:
Have you checked the github action workflows too?

bakulf (approved):

strseb:
good catch!

Issue #846, opened 2021-04-13
windows client using too many gpu resources


ListeningGarden:
**platform**: windows 10 pro 20h2 19042.867
**vpn client version**: 2.1.0 (2.202103250849)

**observed behavior**: when client window is active, it is utilizing up to 20% of my gpu (rtx 3070, so no small amount!) even if the window is not in focus.

![Taskmgr_XI270lgCXm](https://user-images.githubusercontent.com/36242661/114620540-ac434e00-9c79-11eb-9b43-07fa0149176b.png)

when closing the app to tray or closing the window altogether, usage drops completely:

![Taskmgr_RY3IgKZPMa](https://user-images.githubusercontent.com/36242661/114620667-ce3cd080-9c79-11eb-8994-3718a41e1090.png)

**expected behavior**: mozilla vpn should not be consuming nearly a quarter of a brand new computer's resources - it should be completely invisible to the end user's experience, not taxing the system as if it were a 3d game. 

log included in case it helps narrow down the cause!

[mozillavpn.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6306801/mozillavpn.txt)

bakulf:
I suspect it's the ring animation. The last time I checked, disabling the animation, the CPU consumption became reasonable. @lesleyjanenorton do you have time to see if we can optimize the animation ring and if that is the main source of the CPU usage?

ValentinaPC:
Verified this on Windows 10 x64 while suing the VPN client v. 2.3.0 (2.202105131131).
Seems like whenever VPN app is visible, no matter if the VPN is ON or OFF, the gpu usage increases.
If the client is not on visible (minimized or in tray), everything is ok.
More details in the screenshots below.
![gpu usage](https://user-images.githubusercontent.com/15685960/118259393-c3b06980-b4b9-11eb-8623-be7844054b24.png)

Reopening.

ValentinaPC:
Verified this as fixed on Windows 10 using the VPN build 2.3.0 (2.202105241958).
While the VPN is in background, the used GPU resources are optimized. Fixed on the above build also.

PR #848, opened 2021-04-14
Only allow ring animation if the app is visible

strseb:
Touching  #846 - Whilst we need to get this more performant, we can definitely limit the cpu usage when the window is minimized. 
We could also remove Qt.ApplicationInactive condition, so we stop when the window gets out of focus. 
... But's might also look wierdish having the frozen animation in a - non focused window when it's visible.


lesleyjanenorton (approved):
Thanks, looks great! 🎉

ValentinaPC:
Ring animation works if the VPN app is visible, but not in focus.
image

PR #853, opened 2021-04-15
Make testAuthentication.js work with non-en languages

lesleyjanenorton:


bakulf:
Why just this test? Plus in theory, tests should always run in English.

lesleyjanenorton:
I didn't realize! I'll close this.

PR #857, opened 2021-04-16
Reduce the server logging for long long strings

bakulf:


mbirghan:
One check fails. It seems like it does not have anything to do with your patch but we should look into it before merging.

PR #861, opened 2021-04-16
Bump Current Main Branch version to 2.3

strseb:
I'ts certainly not 2.1 anymore :)


bakulf (approved):
Can you do the same for the releases/2.2 branch? 2.2 of course.

PR #862, opened 2021-04-16
Android: Add Custom Back Button Handling

### LINKED ISSUE #622: Android Crash: Type "abort"
This type of crash is ~60% of all our reported crashes. 
This happens only on Huawei devices, apparently when they try to exit the app with the back button.

There is an open QT bug for this, currently open: https://bugreports.qt.io/browse/QTBUG-85449
Also there is a diy patch in the bug, maybe we should also compile qt/android and include this. 
Stacktrace:

```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 0 >>> org.mozilla.firefox.vpn <<<

backtrace:
  #00  pc 000000000006f06c  /apex/com.android.runtime/lib64/bionic/libc.so (abort+160)
  #00  pc 00000000000500fc  /system/lib64/libc++.so (abort_message+232)
  #00  pc 0000000000050218  /system/lib64/libc++.so (demangling_terminate_handler()+44)
  #00  pc 00000000000646c4  /system/lib64/libc++.so (std::__terminate(void (*)())+12)
  #00  pc 000000000006466c  /system/lib64/libc++.so (std::terminate()+52)
  #00  pc 00000000000bb150  /system/lib64/libc++.so (std::__1::thread::~thread()+20)
  #00  pc 00000000000d0f68  /apex/com.android.runtime/lib64/bionic/libc.so (__cxa_finalize+212)
  #00  pc 00000000000cc950  /apex/com.android.runtime/lib64/bionic/libc.so (exit+24)
  #00  pc 0000000000034534  /data/app/org.mozilla.firefox.vpn-18k5up_u3M-aF41KJvRmsg==/lib/arm64/libplugins_platforms_qtforandroid_arm64-v8a.so (startQtApplication(_JNIEnv*, _jclass*)) (SourceCode: androidjnimain.cpp)
```

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-498)

strseb:
The Problem with #622 is that qt tries to turn itself down and failing to do so. 
My idea is to just not send the back command to qt, and handle it ourselves. 
Currently i just pass it to our back handler, and if it's not handled i move the activity back in the stack. 
This makes it inactive (like pressing home) instead of trying a teardown. 

If that does not work, we can also try to trigger a clean teardown here from our side.

Waiting for confirmation if this works as i have no such device to reproduce :( - closes #622


src/mozillavpn.cpp
```diff
@@ -196,6 +197,9 @@ void MozillaVPN::initialize() {
     AndroidDataMigration::migrate();
     settingsHolder->setNativeAndroidDataMigrated(true);
   }
+
+  // Register natvies of the vpn activity
```
bakulf, line 201: natives?

src/platforms/android/androidvpnactivity.h
```diff
@@ -0,0 +1,17 @@
+#ifndef ANDROIDVPNACTIVITY_H
```
bakulf, line 1: header

bakulf (changes requested):

strseb:
Marking no longer WIP: 
I was able to test this on a Remote huawei device - pressing back button now puts the app into a sleep state instead of trying to exit.

bakulf (approved):

PR #869, opened 2021-04-19
Let Ring-Animation Paint to framebuffer

strseb:


src/ui/components/VPNAnimatedRings.qml
```diff
@@ -64,6 +66,19 @@ Rectangle {
         property var ringXCenter: parent.width / 2
         property var ringYCenter: animatedRingsWrapper.yCenter
 
+        renderStrategy: Canvas.Threaded
+        renderTarget: Canvas.FramebufferObject
```
strseb, line 70: Turns out: QT actually for every paint it put's this into a bitmap and copies it over to gpu manually. That's pretty slow. Since we have a gl context on every plattform, we can just enable the _FAST_ mode

src/ui/components/VPNAnimatedRings.qml
```diff
@@ -64,6 +66,19 @@ Rectangle {
...
+
+        // Finds the Minimum Box that we need to repaint and marks this
+        // for painting
+        function makeDirty(){
```
strseb, line 74: 2nd thing i did, mostly for the background consumption: 
We can tell QT which part of the canvas needs to be updated, so it can check itself if a "dirty" part is visible and scheudle painting / or stop. 
I tested this on my macbook -> now gpu usage drops to <1% when i drag firefox over the window :)

lesleyjanenorton (approved):

PR #897, opened 2021-04-27
Glean

bakulf:
This is the first integration experiment. I just want to know if you like the following approach:

1. include glean.js (this is taken from the current release. No changes applied)
2. expose the metrics/pings using glean_parser
3. expose the samples to C++
4. trigger the first events


brizental:
> create the pings/metrics manually: glean_parser doesn't work for QT, it seems. See the comment.

Indeed :( We have a bug open for that, [for reference](https://bugzilla.mozilla.org/show_bug.cgi?id=1706252). I plan to work on it until the end of this week.

glean/glean.js
```diff
@@ -0,0 +1,2 @@
...
+//# sourceMappingURL=glean.js.map
```
brizental, line 2: Hm, I wonder if Qt picks up on the source maps. Do you have any idea @bakulf ? If it doesn I can include the source map file in the release as well. That could be useful for debugging.

glean/load.js
```diff
@@ -0,0 +1,29 @@
...
+
+const _glean = RealGlean.Glean.default;
+
+// Initialization.
+_glean.initialize('MozillaVPN', true, {debug: {logPings: true}});
+
+// TODO: I would like to use glean_parser, but the output is meant to be used
+// by webextensions only (the `import DatetimeMetricType from
+// "@mozilla/glean/webext/private/metrics/datetime"` would not work here). Is
+// OK to create the list of metrics/pings manually in this way?
```
brizental, line 20: Yes, for now this is way. I suppose you saw this already, but this is what we did in our own Qt [sample](https://github.com/mozilla/glean.js/blob/main/samples/qt-qml-app/pings.js) as well.

brizental (commented):
Big fan of the approach of having a single metric to validate that all works as expected before going ahead with more complex data collection 💯  

On the integration itself: right now the `appStarted` metric is not sent in any pings. I see it contais `sendInPings: ['custom']`, but no custom ping is defined. Is this the intended behaviour?

glean/glean.js
bakulf: QT does not support source maps (yet?).

brizental: Might as well remove the source maps comment from this file in future releases of Glean.js either way :)

src/ui/main.qml
```diff
@@ -65,6 +66,23 @@ Window {
...
+        Glean.initialize(VPN.versionString, {
+            post(url, body, headers) {
+                return new Promise((resolve, reject) => {
+                    const xhr = new XMLHttpRequest();
+                    xhr.open("POST", url);
+
+                    for (const header in headers) {
+                      xhr.setRequestHeader(header, headers[header]);
+                    }
+                    xhr.onloadend = () => {
```
brizental, line 78: Is `onloadend` fired when there is a timeout or when network is not available? Note that, this function should always resolve and return _some_ status to Glean.

glean/load.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+    'millisecond');
+
+// public functions
+
+let _initialized = false;
+function initialize(appVersion, uploader) {
+  if (!_initialized) {
+    _initialized = true;
+    _glean.initialize('MozillaVPN', true, {
```
brizental, line 43: It's not ideal not pass a hard coded value for the `uploadEnabled` status, but rather the user's preference on whether or not telemetry should be enabled.

src/constants.h
```diff
@@ -48,6 +48,9 @@ CONSTEXPR(uint32_t, CAPTIVEPORTAL_REQUEST_TIMEOUT_MSEC, 10000, 4000, 0)
 // How fast the animated icon should move
 CONSTEXPR(uint32_t, STATUSICON_ANIMATION_MSEC, 200, 200, 0)
 
+// How often glean pings are sent
+CONSTEXPR(uint32_t, GLEAN_TIMEOUT_MSEC, 3600000, 1000, 0)
```
brizental, line 52: May I ask why this scheduling was chosen? I am wondering since the metric collected right now is the `appStarted` metric, scheduling can be simplified a bit and only instead send the ping once everytime the app is started. 

Note that the ping is only sent when there are metrics to send (because it is declared with `sendIfEmpty: false`), since `appStarted` is only set on start, subsequent pings will not be sent either way :)

brizental (changes requested):
A few things:

1. It is necessary to get data review for the `ping` ping, the `appStarted` metric and for adding Glean.js to the application, before this is merged :) For more info on  the data-review process, check out: https://wiki.mozilla.org/Data_Collection. For an example of a data-review request for adding Glean.js, check out: https://github.com/mozilla-rally/rally-core-addon/pull/505#issuecomment-815804980

2. It is also necessary to [file a data engineering bug](https://bugzilla.mozilla.org/enter_bug.cgi?assigned_to=nobody%40mozilla.org&bug_ignored=0&bug_severity=--&bug_status=NEW&bug_type=task&cf_fx_iteration=---&cf_fx_points=---&comment=Application%20friendly%20name%3A%20my_app_name%0D%0AApplication%20ID%3A%20org.mozilla.my_app_id%0D%0ADescription%3A%20Brief%20description%20of%20your%20application%0D%0AGit%20Repository%20URL%3A%20https%3A%2F%2Fgithub.com%2Fmozilla%2Fmy_app_name%0D%0ALocations%20of%20%60metrics.yaml%60%20files%20%28can%20be%20many%29%3A%0D%0A%20%20-%20src%2Fmetrics.yaml%0D%0ALocations%20of%20%60pings.yaml%60%20files%20%28can%20be%20many%29%3A%0D%0A%20-%20src%2Fpings.yaml%0D%0ADependencies%2A%3A%0D%0A%20-%20org.mozilla.components%3Aservice-glean%0D%0ARetention%20Days%3A%20N%2A%2A%0D%0A%0D%0A%0D%0A%2A%20Dependencies%20can%20be%20found%20%5Bin%20the%20Glean%20repositories%5D%28https%3A%2F%2Fprobeinfo.telemetry.mozilla.org%2Fglean%2Frepositories%29.%20Each%20dependency%20must%20be%20listed%20explicitly.%20For%20example%2C%20the%20default%20Glean%20probes%20will%20only%20be%20included%20if%20glean%20itself%20is%20a%20dependency.%0D%0A%0D%0A%2A%2A%20Number%20of%20days%20that%20raw%20data%20will%20be%20retained.%20A%20good%20default%20is%20180.%0D%0A%0D%0AIf%20you%20need%20new%20dependencies%2C%20please%20file%20new%20bugs%20for%20them%2C%20separately%20from%20this%20one.%20For%20any%20questions%2C%20ask%20in%20the%20%23glean%20channel.&component=General&contenttypemethod=list&contenttypeselection=text%2Fplain&defined_groups=1&filed_via=standard_form&flag_type-4=X&flag_type-607=X&flag_type-800=X&flag_type-803=X&flag_type-936=X&form_name=enter_bug&maketemplate=Remember%20values%20as%20bookmarkable%20template&needinfo_from=fbertsch%40mozilla.com%2C%20&op_sys=Unspecified&priority=--&product=Data%20Platform%20and%20Tools&rep_platform=Unspecified&short_desc=Enable%20new%20Glean%20App%20my_app_name&target_milestone=---&version=unspecified) to enable the `MozillaVPN` id on the Glean Pipeline :)

3. Even though glean_parser does not provide Qt templates yet, I'd encourage already creating a .yaml files with the ping and metrics being included here. Once the VPN's application id is enabled, these files will be read and the VPN's metrics and pings will be automatically documented in the [Glean Dictionary](https://dictionary.protosaur.dev/). Alternatively, it is also possible to use glean_parser to generate Markdown documentation of these metrics and pings based on the .yaml files.

All of this is also laid out in the [Adding Glean to your project](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html#glean-integration-checklist) checklist on the Glean book.

src/ui/main.qml
bakulf: onloadend is called when the XHR completes the task, with a failure or with a success.

brizental: Yes, but timeout [means there the task was never completed](https://xhr.spec.whatwg.org/#events).

bakulf: Does glean.js check the status code to be == 20x?

glean/bin/create_header.py
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os.path
```
Dexterp37, line 6: This is an interesting one :-D 

The _canonical_ way to do this would be to extend the glean-parser like [FOG does](https://searchfox.org/mozilla-central/source/toolkit/components/glean/build_scripts/glean_parser_ext/cpp.py). However, for the sake of expediency and since this is a very small piece of code, I'm ok with keeping it as it is. Know that this will break when we'll change the metrics.yaml format (we keep evolving that format, and that's why we provide the glean-parser). Maybe you can add a comment to the top, mentioning that if this file gets more complicated or breaks frequently the canonical approach should be re evaluated.

glean/generated/gleansample.h
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Dexterp37, line 1: Our team recommends not committing the generated code, when possible, as it hides some issues and might be problematic to debug if Glean changes and this is not changed: would you consider removing this?

glean/generated/pings.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
```
Dexterp37, line 1: Our team recommends not committing the generated code, when possible, as it hides some issues and might be problematic to debug if Glean changes and this is not changed: would you consider removing this?

glean/generated/sample.js
```diff
@@ -0,0 +1,128 @@
+"use strict";
```
Dexterp37, line 1: Our team recommends not committing the generated code, when possible, as it hides some issues and might be problematic to debug if Glean changes and this is not changed: would you consider removing this?

src/constants.h
Dexterp37: @bakulf I'm curious about this, too :-)

glean/load.js
```diff
@@ -0,0 +1,34 @@
...
+    // TODO: this should be removed
+    debug: {logPings: true}
+  });
+}
+
+function sendPing() {
+  Pings.ping.submit();
+}
+
+function setUploadEnabled(enabled) {
```
Dexterp37, line 32: Is this ever called? Can users toggle telemetry collection on/off during the application lifecycle?

src/ui/main.qml
Dexterp37: > 
> 
> Does glean.js check the status code to be == 20x?

Glean's upload logic needs to be aware of the status code because it will make decisions based on that, especially whether or not to retry upload or to remove the stored ping. However, our Qt platform implementation has no storage so this should not make too much of a difference.

I believe the uploader implementation should be provided by Glean.js (we do that for other platforms, too). For the sake of unblocking you, I believe it's fine for it to be as it is. If you want a reference, here's the current [webext uploader](https://github.com/mozilla/glean.js/blob/main/glean/src/platform/webext/uploader.ts).

Dexterp37 (commented):
This looks really good overall. I left a few comments here and there.

src/ui/main.qml
Dexterp37: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1709371

src/ui/main.qml
```diff
@@ -65,6 +66,27 @@ Window {
         if (VPN.startMinimized)
             this.showMinimized();
 
+        Glean.glean.initialize('MozillaVPN', VPNSettings.gleanEnabled, {
+          appBuild: `MozillaVPN/${VPN.versionString}`,
```
Dexterp37, line 70:
```suggestion
          appBuild: `MozillaVPN/${VPN.versionString}`,
          appDisplayVersion: VPN.versionString,
```

glean/pings.yaml
```diff
@@ -0,0 +1,22 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/2-0-0
+
+ping:
```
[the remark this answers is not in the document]
Dexterp37: Well, you can call it `main` :-)

glean/metrics.yaml
```diff
@@ -0,0 +1,168 @@
...
+# the `glean_parser` PyPI package.
+
+# Metrics in this file may make use of SDK reserved ping names. See
+# https://mozilla.github.io/glean/book/dev/core/internal/reserved-ping-names.html
+# for additional information.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+sample:
```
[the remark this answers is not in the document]
Dexterp37: Ah I see, ok

Dexterp37:
I think the Glean.js usage looks good here. My only concerns are about metric definition (as highlighted above) but, since this is only a sample for testing, that's fine. Looks good overall! Feel free to ping me if you want me to have another look, I'll gladly do.

PR #900, opened 2021-04-27
iOS switch servers without disconnecting

mbirghan:


src/platforms/ios/ioscontroller.swift
```diff
@@ -227,11 +204,22 @@ public class IOSControllerImpl : NSObject {
 
                 Logger.global?.log(message: "Loading the tunnel succeeded")
 
-                // If we were switching server, now it's time to consider the operation completed.
-                switchingServer = false
-
                 do {
-                    try (self.tunnel!.connection as? NETunnelProviderSession)?.startTunnel()
+                    if (reason == 1) {
```
bakulf, line 208: Maybe write a comment about the 'reason == 1'.

src/platforms/ios/iostunnel.swift
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: MIT
+// Copyright © 2018-2020 WireGuard LLC. All Rights Reserved.
+
+import Foundation
+import NetworkExtension
+import os
+
+class PacketTunnelProvider: NEPacketTunnelProvider {
```
bakulf, line 8: All good! I wonder if we can keep the PacketTunnelProvider implementation and "extend" it. 

For instance we could add to the Info.plist these:
```
<key>NSExtension</key>
        <dict>
                <key>NSExtensionPointIdentifier</key>
                <string>com.apple.networkextension.packet-tunnel</string>
                <key>NSExtensionPrincipalClass</key>
                <string>$(PRODUCT_MODULE_NAME).MozillaVPNPacketTunnelProvider</string>
        </dict>
```

Then implement the MozillaVPNPacketTunnelProvider in this way:

```
class MozillaVPNPacketTunnelProvider : PacketTunnelProvider {
 ...
  override func handleAppMessage(_ messageData: Data, completionHandler: ((Data?) -> Void)? = nil) {
  ...
```

bakulf (changes requested):
Let me use `request changes`. Tell me if what I suggest makes sense or not. If not, just ask me to review the PR again and I give an r+. Thanks!

src/platforms/ios/iostunnel.swift
mbirghan: That was my first instinct as well. I tried that but since the `adapter` property is private the subclass cannot access it and therefore would not work.

src/platforms/ios/ioscontroller.swift
mbirghan: Sure I will add one.

src/platforms/ios/iostunnel.swift
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: MIT
+// Copyright © 2018-2020 WireGuard LLC. All Rights Reserved.
```
bakulf, line 2: We should ask if we can/have to change the header. I check this. It's not a blocker.

bakulf (approved):

PR #902, opened 2021-04-27
User Set-Local DNS

### LINKED ISSUE #864: Consider supporting setting a custom dns server
Something i've grabbed on reddit, users where asking if we supported setting a custom dns server instead of the vpn-internal one. I think mostly to have support for their PI-Hole in the local network :) 

# Feature Description

There might be cases where a user withes to use a Local DNS - e.g to enforce filtering rules (Corperate-DNS, AdBlocking, Parental Filter) 
The User should be able to:
-> Go to the Settings
-> Select to not use our DNS
-> Enter their own dns server
--> When connecting, sites like mullad.net/check should then show that the entered DNS is used. 

# Dependecies

- [ ] #920 Need UX to support new DNS setting
- [ ] #902 Engineer custom DNS Option

strseb:
This enables a user to Override the DNS Server used to a Local DNS, in case they want to use a PiHole or similar

Keeping this as Draft rn: 
Todo:
- [x] Android Support
- [x] Dummy Support
- [?] Linux Support (not tested yet)
- [x] Windows Support
- [x] IOS Support
- [?] Mac Support (not tested yet)

Closes #864


src/controller.cpp
```diff
@@ -187,9 +187,20 @@ void Controller::activateInternal() {
     vpnDisabledApps = settingsHolder->vpnDisabledApps();
   }
 
+  // Use the Gateway as DNS Server
+  // If the user as entered a valid dns, use that instead
+  QHostAddress dns = QHostAddress(server.ipv4Gateway());
+  if (FeatureList::instance()->userDNSSupported() &&
+      !settingsHolder->useGatewayDNS() &&
+      settingsHolder->userDNS().size() > 0 &&
```
bakulf, line 195: `isEmpty()`

src/src.pro
```diff
@@ -766,7 +766,7 @@ else {
 RESOURCES += $$PWD/../translations/servers.qrc
 
 exists($$PWD/../translations/translations.pri) {
-    include($$PWD/../translations/translations.pri)
+  #  include($$PWD/../translations/translations.pri)
```
bakulf, line 769: ?

src/settingsholder.cpp
```diff
@@ -305,3 +316,24 @@ void SettingsHolder::addVpnDisabledApp(const QString& appID) {
...
+
+  logger.log() << "is null " << address.isNull();
+
+  if (address.isNull()) {
+    return false;
+  }
+  /* Currently we need to limit this to LAN-DNS
+   * (at least on windows) since the killswitch makes
+   * sure that no dns traffic may happen to outside of lan
+   */
```
bakulf, line 332: Can you tell me more about this? Why cannot we have an external DNS?

src/ui/components/VPNTextInput.qml
```diff
@@ -0,0 +1,65 @@
+import QtQuick 2.0
```
bakulf, line 1: mozilla license header.

src/ui/settings/ViewNetworkSettings.qml
```diff
@@ -23,58 +23,122 @@ VPNFlickable {
...
+            //: Associated to a group of settings that require the VPN to be disconnected to change
+            errorMessage: qsTrId("vpn.settings.vpnMustBeOff")
+        }
+
     }
 
+
+
+
+
```
bakulf, line 141: a few empty lines :)

bakulf (changes requested):
Tell me more why we need to use lan-only DNS servers.

src/settingsholder.cpp
strseb: The problem here seems to be the WFP rules, wg-go is setting. So Even if we exclude <some-ip>, dns traffic to it is blocked. If wanted we can easily override it.

strseb:
@bakulf - marked this as WIP for now, i've updated the pr to include the 3 mullvad dns servers as an option to choose or a user defined dns

src/ui/components/VPNTextInput.qml
bakulf: it's still missing, right?

bakulf (changes requested):

bakulf (approved):

bakulf:
But fix the linux compilation and all the other failures.

PR #903, opened 2021-04-27
Release notes for v2.2.0 and v2.1.0 - closes #891

### LINKED ISSUE #891: Write a release-note file and store it in the repo

bakulf:
Can you please take a look at these 2 release notes? If there is anything to change/add/remove, let's do it.

### DIFF
--- a/docs/releasenotes/v2.1.0.md
+++ b/docs/releasenotes/v2.1.0.md
@@ -1,0 +1,7 @@
+# v2.1.0 Release Notes
+
+ * VPN Client for windows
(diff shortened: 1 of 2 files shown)


birdsarah (approved):
Seems like a good start to me. As I haven't been involved in much client development I don't have a great perspective on whether we've missed something here.

birdsarah:
Should we enforce squash merges moving forward to help with diffing between versions?

bakulf:
> Should we enforce squash merges moving forward to help with diffing between versions?

"enforce squash merges" is already enabled. We should disable the other kinds of merging.

Issue #798, opened 2021-04-07
FVP-02-003 General: Balrog incorrectly verifies certificate chain


bakulf:
It was found that Balrog does not properly verify the certificate chain, permitting rogue
root certificates and their fellowship to pass. The attacker-controlled leaf certificate holds
a public key that will be used to verify the update used in Windows and macOS.
This signifies the risk of state-funded attackers who are in charge of a trusted certificate
authority being able to perform Man-in-the-Middle attacks on the TLS connection
initiated by Mozilla VPN to receive updates. Attackers can now replace the binary code
of the update with malicious malware bypassing the Balrog mechanism that intends to
detect those attacks.

bakulf:
@birdsarah can I assign this bug to you?

birdsarah:
Closed via #993

PR #914, opened 2021-04-29
Simplify balrog implementation

### LINKED ISSUE #797: FVP-02-002 WP1: Balrog does not verify certificate chain on macOS
It was found that Balrog does not verify the whole certificate chain on macOS. This
allows attackers to supply a self-signed leaf certificate, effectively indicating a bypass of
Balrog. This could be abused by state-funded attackers who are in charge of a trusted
valid certificate authority. They could perform a Man-in-the-Middle attack and replace the
binary code provided by the Mozilla VPN update with malicious malware.

birdsarah:
Fixes #797, #798 

Instead of rolling our own, we use the mozilla go package autograph (https://github.com/mozilla-services/autograph/). 

This allows for simplification of both the api.go code (handled by autograph except for logger code) and the update/balrog.cpp. In particular as we can just past the x5u url to the autograph library we save network request code as well as a lot of parsing and handling.

In addition, we also resolve the FVP-02-002 and FVP-02-003.

Note, we have verified that balrog algorithm is p384ecdsa which is supported by autograph module.


src/update/balrog.cpp
```diff
@@ -54,8 +54,9 @@ constexpr const char* BALROG_MACOS_UA = "Darwin_x86";
 #  error Platform not supported yet
 #endif
 
-constexpr const char* BALROG_CERT_SUBJECT_CN =
-    "aus.content-signature.mozilla.org";
+// TODO - Do we need to do this cert subject verification?
+//constexpr const char* BALROG_CERT_SUBJECT_CN =
+//    "aus.content-signature.mozilla.org";
```
birdsarah, line 59: @bakulf - do we need to do the cert subject verification. Code was:

```
  const QSslCertificate& leaf = list.constFirst();
  logger.log() << "Validating cert subject";
  QStringList cnList = leaf.subjectInfo("CN");
  if (cnList.isEmpty() || cnList[0] != BALROG_CERT_SUBJECT_CN) {
    logger.log() << "Invalid CN:" << cnList;
    return false;
  }
```

If yes, I suggest we also hand this work off to the balrog go module and I can write a small helper function to take care of this, which we can pass the `BALROG_CERT_SUBJECT_CN` to keep the balrog go module very agnostic.

If no, I'll push another small commit to remove these commented codes.

balrog/api.go
```diff
@@ -13,220 +13,44 @@ package main
...
+	"unsafe"
+
+	"github.com/mozilla-services/autograph/signer/contentsignaturepki"
 )
 
 var loggerFunc unsafe.Pointer
 
 type CLogger struct {
-  level C.int
+	level C.int
```
birdsarah, line 26: Note these formatting changes are the result of running `go fmt`

balrog/go.sum
```diff
@@ -0,0 +1,382 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
birdsarah, line 1: I'm wondering if all these dependencies are a problem.

Many are not, to my reading of the code, needed by any of the code paths we use, but are part of other functions of the "autograph" module which includes a variety of features.

So, if we really didn't want to use the autograph module as packaged, we could probably do something with git submodules and then copying just the pieces of code we need (the contentsignaturepki directory and maybe one other). Or just manually copy in the files, but with submodules we could more easily stay up todate. But this would definitely be a little gross.

balrog/go.sum
```diff
@@ -0,0 +1,382 @@
...
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.2.1/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
+github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
+github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/ugorji/go v0.0.0-20180112141927-9831f2c3ac10 h1:4zp+5ElNBLy5qmaDFrbVDolQSOtPmquw+W6EMNEpi+k=
+github.com/ugorji/go v0.0.0-20180112141927-9831f2c3ac10/go.mod h1:hnLbHMwcvSihnDhEfx2/BzKp2xb0Y+ErdfYcrs9tkJQ=
+github.com/youtube/vitess v2.1.1+incompatible h1:SE+P7DNX/jw5RHFs5CHRhZQjq402EJFCD33JhzQMdDw=
+github.com/youtube/vitess v2.1.1+incompatible/go.mod h1:hpMim5/30F1r+0P8GGtB29d0gWHr0IZ5unS+CG0zMx8=
```
bakulf, line 226: Weird dependencies...

balrog/go.sum
bakulf: If those dependencies are included in the final dll/so, probably it's better to reduce them to the minimum. In this way, we reduce the size of the final package.
I also wonder if there are issues with the licenses.

How long would it take to replace the "autograph" module with a smaller package? Via submodule probably, as you suggested?

bakulf (approved):
The code looks perfect. Thanks! Let's talk about the dependencies.

birdsarah:
Note from @g-k - double check that the code path we're using from autograph does verify the whole signature chain.

src/update/balrog.cpp
birdsarah: autograph doesn't implement as its expectation is a closed loop. It verifies the chain all matches, but doesn't do a check like this on the leaf.

birdsarah:
Closing in favor of the new plan with minimal dependencies.

PR #950, opened 2021-05-01
Sort countries and cities by localized names

bakulf:
This PR contains the following commits (squashed):
1. rename ServerData::country() to ServerData::countryName()
2. rename ServerData::city() to ServerData::cityName()
3. sort countries/cities when the language changes
4. expose `localizedName` in the ServerCountryModel
5. expose `cities` as an array of QVariant<QStringList> composed by name and localizedName
6. update the tests

### DIFF
--- a/src/ui/components/VPNServerCountry.qml
+++ b/src/ui/components/VPNServerCountry.qml
@@ -36,7 +36,7 @@ VPNClickableRow {
 
     activeFocusOnTab: true
 
-    accessibleName: VPNLocalizer.translateServerCountry(code, name)
+    accessibleName: localizedName
     Keys.onDownPressed: repeater.itemAt(index + 1) ? repeater.itemAt(index + 1).forceActiveFocus() : repeater.itemAt(0).forceActiveFocus()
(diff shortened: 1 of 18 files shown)


src/ui/components/VPNServerCountry.qml
```diff
@@ -36,7 +36,7 @@ VPNClickableRow {
 
     activeFocusOnTab: true
 
-    accessibleName: VPNLocalizer.translateServerCountry(code, name)
+    accessibleName: localizedName
```
mbirghan, line 39: nit: Maybe call it `localizedCountryName` to make it more obvious what it is

mbirghan (approved):
Looks awesome! 
Maybe just one little naming change for clarity.

src/ui/components/VPNServerCountry.qml
bakulf: This string comes from the `ServerCountryModel` and it matches the `name` attribute.
I like the idea of renaming it, but then I would rename each `name` attribute of each model.
I'll do it in a separate PR.

PR #960, opened 2021-05-04
Fix #801: Potential race condition in ping sender

### LINKED ISSUE #801: FVP-02-006 WP3: Race condition in Ping Sender could expose gateway IP
It was found that Mozilla VPN was prone to a race condition vulnerability in the Ping
Sender that frequently delivers ICMP packets to the internal IP address of the
WireGuard gateway. Shortly after turning the VPN off, those ICMP packets are at risk of
being sent outside of the WireGuard tunnel and might reveal which gateway IP was
used. Since this event is very rare and unreliable, whilst information leakage is
additionally scarce, this issue is of purely informational nature.

oskirby:
There is a potential race condition when sending ICMP pings for connection health monitoring where closing the tunnel could result
in exposure of the gateway IP address (see issue #801). By binding to the source address when sending a ping, we should prevent this from occurring.

So far I have tested this on Linux and Windows. I've updated the platform code for Android, iOS and Mac while I was at it, but I haven't been able to test any of them yet.

### DIFF
--- a/src/connectioncheck.cpp
+++ b/src/connectioncheck.cpp
@@ -29,8 +29,8 @@ void ConnectionCheck::start() {
   logger.log() << "Starting a connection check";
 
   MozillaVPN::instance()->controller()->getStatus(
-      [this](const QString& serverIpv4Gateway, uint64_t txBytes,
-             uint64_t rxBytes) {
+      [this](const QString& serverIpv4Gateway, const QString& deviceIpv4Address,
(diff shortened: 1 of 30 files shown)


bakulf (approved):
All good! Thanks! Can we test it on android and IOS before landing it?

oskirby:
I should be able to give this a try on Android, I just haven't set up the build environment to do so yet. For iOS I don't have any devices to test it on.

Issue #871, opened 2021-04-19
Android: Background-Service Shutdown on Swipe Kill 


strseb:
Affected Device: 
OS Android version 10:
-> Samsung Galaxy S9, model SM-G960U 1.1

strseb:
Failed to reproduce this normally. I tried using Devices from Browserstack. 
An S9 with android 8
An S20 ultra with android 10

I was only able to reproduce this on the Android 10 when enabling the powersaving modes - But afaik the customer did exempt the app from that?

strseb:
@ValentinaPC , do you possibly have a device to reproduce this on ? :)

ValentinaPC:
@strseb I've tried to reproduce this on:
- Samsung Galaxy S10 (SM-G973F/DS) with Android 11
- Google Pixel IV with Android 10
and I was not able to duplicate the issue.
On both devices the VPN client (2.2) behaves as expected after swipe-kill.

ValentinaPC:
Verified this with our available resources (see above comment), and we were not able to duplicate the issue.
We will mark this as "verified".

PR #969, opened 2021-05-05
Android wg go

### LINKED ISSUE #867: Android device shows connected while app shows not connected
Note that I hide the VPN notification in case that has anything to do with this.

strseb:
This PR removes the Use of the 'com.wireguard.android:tunnel' module. 

This still uses the `com.wireguard.config*` files as they are super convenient have nice type-checks, but we can remove this in the next refactor step and generate that all in the qt-process, for now i did want to touch the least possible parts on the service to achieve this. 

So Why: 
The module provides the "goBackend" for us -> this is continent but it spins up a 2nd private service. 
This one loads wireguard and holds the vpn-tun. So we're having 2 services now to keep alive for full functionality. 
Given that we're seeing on phones either the ipc service #867 or vpn service #871 die, we should simply not have 2.

Thus i want to compile wg-go ourselves and use it first class, as we do on all the other daemons :)  

One note:
i copied the com/wireguard/config* namespace over because if we were to import this via gradle, we also get the wireguard-build of wg-go included, which we dont need anymore :) 

closes #867


bakulf (approved):
Can you add the wireguard files to the .clang-format-ignore?

PR #973, opened 2021-05-05
Lower C++ language version to C++14

abpostelnicu:
In order to be able to build on platforms that don't have fully support for C++17, C++14 should be considered the minimum required.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -11,7 +11,7 @@ QT += quick
 QT += widgets
 QT += charts
 
-CONFIG += c++1z
+CONFIG += c++14
 
 TEMPLATE  = app
 
--- a/tests/unit/unit.pro
+++ b/tests/unit/unit.pro
@@ -17,6 +17,11 @@ DEFINES += UNIT_TEST
 TEMPLATE = app
 TARGET = tests
 
+wasm {
+    CONFIG += c++1z
+    QMAKE_CXXFLAGS *= -Werror
+}
+
 INCLUDEPATH += \
             . \
             ../../src \


bakulf (changes requested):
This breaks wasm. Can you investigate why? We can also have 2 different compiler versions, one for the builds, and one for wasm.

abpostelnicu:
@bakulf do you think this is related with my change, I see this warning that is transformed into an error:
```
In file included from /opt/5.15.0/wasm_32/include/QtNetwork/QNetworkReply:1:
266
In file included from /opt/5.15.0/wasm_32/include/QtNetwork/qnetworkreply.h:44:
267
In file included from /opt/5.15.0/wasm_32/include/QtCore/QIODevice:1:
268
In file included from /opt/5.15.0/wasm_32/include/QtCore/qiodevice.h:45:
269
In file included from /opt/5.15.0/wasm_32/include/QtCore/qobject.h:47:
270
/opt/5.15.0/wasm_32/include/QtCore/qstring.h:1229:22: warning: definition of implicit copy constructor for 'QCharRef' is deprecated because it has a user-declared copy assignment operator [-Wdeprecated-copy]
271
    inline QCharRef &operator=(const QCharRef &c) { return operator=(QChar(c)); }
```

bakulf:
So, no, those are not your errors and I still don't understand why changing the clang creates an issue with wasm (which doesn't use clang!). The issue seems to be in how we treat warning with C++14. Can you please try the following changes:

1. in test/unit/unit.pro add:
```
wasm {
    CONFIG += c++1z
   QMAKE_CXXFLAGS *= -Werror
}
```

abpostelnicu:
> So, no, those are not your errors and I still don't understand why changing the clang creates an issue with wasm (which doesn't use clang!). The issue seems to be in how we treat warning with C++14. Can you please try the following changes:
> 
> 1. in test/unit/unit.pro add:
> 
> ```
> wasm {
>     CONFIG += c++1z
>    QMAKE_CXXFLAGS *= -Werror
> }
> ```

We could just remove the warning that causes issues Wno-…

bakulf (approved):

Issue #977, opened 2021-05-06
Back-compability for the new `ping` race condition PR


bakulf:
Recently we landed an improvement in the way we send ICMP packages to test the network stability. But this is not compatible with the existing mozillavpn daemon/service. In case the user updates the client without restarting the daemon (normal on Linux, for instance), the client crashes:

`Fatal: ASSERT: "obj.contains("deviceIpv4Address")" in file platforms/linux/linuxcontroller.cpp, line 151 (linuxcontroller.cpp:151)`

I wonder if we can support an empty/not-set `deviceIpv4Address` entry.

@oskirby, thoughts?

oskirby:
Oops, I should have checked this usecase. For the socket based platforms (everything except windows) we just have to skip the call to `bind()` which should be a trivial change.

Windows, on the other hand, needs to use a different API call if the source address is unset.

I can get started on a fix right away.

oskirby:
As an additional improvement, we should also make sure to restart the dbus service when re-installing the VPN client package. This should be doable using the dh_systemd helpers in the Debian control files.

ValentinaPC:
@bakulf Is there a way for QA to verify this? Thanks!

PR #986, opened 2021-05-06
Add systemd service file and kill old versions on upgrade

oskirby:
When launching the D-Bus service as a bus-activatable service we can start the daemon but we don't have a good way to stop it
during a software update. Adding a systemd service file allows the OS to manage that service when interacting with the package
manager.

However, when upgrading from 2.2.0 and earlier the systemd service file doesn't exist. For these cases we need a preinst hook to kill off any old versions that might be running.

Together, these two changes should ensure that the user can't be left running a version of the daemon that's older than the UI after performing an upgrade via the package manager.

See: mozilla-mobile/mozilla-vpn-client#977

### DIFF
--- a/linux/debian/mozillavpn.service
+++ b/linux/debian/mozillavpn.service
@@ -1,0 +1,7 @@
+[Unit]
+Description=MozillaVPN D-Bus service
+
+[Service]
+Type=dbus
+BusName=org.mozilla.vpn.dbus
+ExecStart=/usr/bin/mozillavpn linuxdaemon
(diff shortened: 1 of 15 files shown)


linux/debian/mozillavpn.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
```
bakulf, line 1: Where should this file be installed? And what does install it? Is the debian package to take care of it?

bakulf (approved):

PR #990, opened 2021-05-07
Telemetry Policy view

bakulf:


lesleyjanenorton (approved):
This is great. Is it expected that the user will see the prompt each time they sign in?

Issue #842, opened 2021-04-13
Linux: Support Split tunnel


strseb:
We should be able to support this using C-groups :)

ValentinaPC:
@oskirby @strseb can we have some mockups for this? It had a different layout, and now, has changed. We need to know if this will be the final design, so that we can update our Test Cases. Thank you in advance!

ValentinaPC:
Verified this on 2.4.0 (2.202107191711) VPN version on Linux Ubuntu 20.04.
The App Permission feature was introduced with success and works as expected. Some minor issues were logged separately.
Marking this as verified.

PR #996, opened 2021-05-09
WIP: Support per-app VPN on Linux

oskirby:
I feel like this is getting close to completion, but there are still a few lingering bugs. In the meantime it would be good to get some more eyes on this feature in case there are changes needed.
Closes: #842

To use this feature, you will have to start the firewall service manually before it can track and apply firewall rules to applications (eg: `mozillavpn linuxappfirewall`).

This breaks the problem of split tunnelling into the following modules/classes:
- `PidTracker` is a module that hooks the kernel's `fork()` and `exit()` system calls to build up a mapping the process tree.
- `AppTracker` listens on the D-Bus for application launch events from the GTK framework (eg: Gnome, LXDE, Unity).
- `StartupNotifyWatcher` listens for application launch events on the primary X display using the startup notification protocol (KDE, Xfce).
- `FirewallService` is a D-Bus service that moves processes into and out of some cgroups that we setup for split tunnelling.
- `LinuxAppImageProvider` takes `.desktop` files, and searches the `XDG_DATA_DIRS` for the corresponding icons.
- `LinuxAppListProvider` searches the `XDG_DATA_DIRS` for `.desktop` files describing the user's installed applications.
- `Netfilter` is a Go module for accessing the nftables firewall.

Some known bugs, and areas for improvement:
 - [x]  ~~When stopping Chrome, the `PidTracker`  seems to miss an `exit()` call, causing zombie application entries. I think I need to do refcounting on the kernel threads to properly resolve this.~~
 - [x] ~~Moving applications into and out of the cgroups multiple times causes connections to get dropped. I think this is related to conntrack getting confused and I need to refine the netfilter rules a bit to play better with the conntrack module.~~
 - [ ] It might make more sense to merge the daemon and firewall services together.
 - [ ] Write a systemd service file to start the firewall at boot.
 - [ ] Most applications on Ubuntu/Debian still share the a DNS resolver on localhost, so we can still leak location data from DNS.


bakulf (approved):

oskirby:
I am going to close this PR in favour of a new one (#1133) rebased upon the removal of the Linux helper script (2f291e3) and after some refactoring to combine the firewall and D-Bus services.

PR #1000, opened 2021-05-10
Silent server switching

mbirghan:
This patch includes:
- adding a field to `MozillaVPN` to track the currently active server 
- adding two new states (`StateSilentSwitching`, `StateConfirmingSilentSwitch`) which mirror functionality of `StateConnecting` and `StateConfirming` without notifying the UI
- implement a silent switch function to switch servers without notifying the UI
- switching servers when the connection is unstable or there is no connection


bakulf (changes requested):
The code looks good but I wonder if we can do something different. What scares me most is the introduction of 2 new States, not exposed to the UI, and because of this, we could have blank views.

Currently the Controller is a complex state machine with a bit of spaghetti code to go from state A to state B. I wonder if we can do something better. Here is a raw idea. Tell me if you like it and if this can improve the current code:

1. for MozillaVPN we have `Task` objects to do operations. We can do something similar for the `Controller`.
2. When we start an operation, we schedule a `Task` or something similar.
3. At the end of the task, we check if there is a pending one and we execute it.
4. In this way we can get rid of the `NextStep` enum.
5. After that, the silent-switch can be implemented as a invisible task to be scheduled when in `StateOn`, without having the state.

What do you think?

src/controller.cpp
```diff
@@ -192,6 +194,60 @@ void Controller::activateInternal() {
...
+    if (iterator->publicKey() == vpn->serverPublicKey()) {
+      servers.erase(iterator);
+      break;
+    }
+    ++iterator;
+  }
+
+  Server server = Server::weightChooser(servers);
+  Q_ASSERT(server.initialized());
+
```
bakulf, line 229: Maybe check if the current publicKey is != the new one.

src/controller.cpp
```diff
@@ -241,29 +303,37 @@ void Controller::connected() {
...
     logger.log() << "Invalid confirmation received";
     return;
   }
 
   m_connectionRetry = 0;
   emit connectionRetryChanged();
 
-  setState(StateOn);
-  emit timeChanged();
+  if (m_state != StateOn) {
```
bakulf, line 314:
```
if (m_state == StateOn) {
  emit silentSwitchinDone();
  return;
}
```

src/controller.cpp
```diff
@@ -241,29 +303,37 @@ void Controller::connected() {
...
+  } else {
+    emit silentSwitchDone();
+  }
 }
 
 void Controller::connectionFailed() {
   logger.log() << "Connection failed!";
 
-  if (m_state != StateConfirming) {
+  if (m_state == StateOn) {
```
bakulf, line 332: Move this block after line 336.

bakulf: Plus, if we are unable to connect, we are in a bad scenario. Maybe we should call `deactivate()`. Or have a connectionRetry logic. What do you think?

src/mozillavpn.cpp
```diff
@@ -1106,6 +1111,13 @@ void MozillaVPN::deactivate() {
   scheduleTask(new TaskControllerAction(TaskControllerAction::eDeactivate));
 }
 
+void MozillaVPN::silentSwitch() {
+  logger.log() << "VPN tunnel silent server switch";
+
+  deleteTasks();
```
bakulf, line 1117: Why deleting all the tasks?

bakulf (changes requested):

src/mozillavpn.cpp
mbirghan: I was worried that multiple switch tasks might be in the queue. After a successful switch we do not want to switch again if not necessary.
If multiple pings in a row indicate an unstable connection the task will be added multiple times. Or am I missing something?

src/controller.cpp
mbirghan: Since I removed the currently active server previously this should not happen. But I will add an assertion.

src/controller.cpp
mbirghan: I did not call `deactivate()` to use the connectionRetry logic from the function. 
If calls `m_impl->deactivate(ControllerImpl::ReasonConfirming)` which triggers the retry. 

What we could do that would improve the logic is that after unsuccessful retries we silently switch servers again. This however would be a little bit more complex but would ensure that we find the best server.

src/controller.cpp
```diff
@@ -192,6 +194,61 @@ void Controller::activateInternal() {
...
+
+  MozillaVPN* vpn = MozillaVPN::instance();
+  Q_ASSERT(vpn);
+
+  QList<Server> servers = vpn->servers();
+  Q_ASSERT(!servers.isEmpty());
+
+  if (servers.length() <= 1) {
+    logger.log()
+        << "Cannot siltent switch servers because there is only one available";
```
bakulf, line 213: siltent -> silent

bakulf (approved):

oskirby (approved):

PR #1002, opened 2021-05-10
Breakout ip utils and implement for linux

birdsarah:
This PR is a corrected version of #635 that is branched off #666.

https://github.com/mozilla-mobile/mozilla-vpn-client/commit/877205fd81a2444c22db913490d41bf5e4adf0e5 fixes code review from #635.


src/platforms/linux/daemon/iputilslinux.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+  auto guard = qScopeGuard([&] { close(sockfd); });
+
+  // Get the index of named ifr and link with ifr6
+  struct ifreq ifr;
+  strncpy(ifr.ifr_name, WG_INTERFACE, IFNAMSIZ);
+  ifr.ifr_addr.sa_family = AF_INET6;
+  int ret = ioctl(sockfd, SIOGIFINDEX, &ifr);
+  if (ret) {
+    logger.log() << "Failed to get ifrindex. Return code: " << ret;
```
oskirby, line 135: There's a typo in the error message here: ifrindex should be ifindex

oskirby:
As a high-level comment, what's the distinction between `WireguardUtils` and `IPUtils`? It seems like the two classes area dealing with the same things: turning an `InterfaceConfig` into a running wireguard network interface. Might be worth merging the two classes.

birdsarah:
> As a high-level comment, what's the distinction between `WireguardUtils` and `IPUtils`? It seems like the two classes area dealing with the same things: turning an `InterfaceConfig` into a running wireguard network interface. Might be worth merging the two classes.

I can explain my thinking, that doesn't mean that we don't want to merge, but to give you my thought process. Doing this work for linux is just the first step, we want to get rid of the macosx script too. The different classes hold the different command line calls - wireguardutils is calls that were done to `wg`, iputils to `ip`, etc. In my head I thought that maybe different command line tools would potentially be shared, or not, across platforms. So the mac helper script calls `wg` but not `ip`. But as I haven't looked at what it will actually take on mac this may have been one of those trying to optimize too soon moments.

birdsarah:
I forgot as it's been a while, @bakulf and I also discussed this code structure. I'm going to merge for now, but as we get rid of the MacOS script we can keep making sure this isn't unnecessarily complicatedl

birdsarah:
The other reason that this structure may be useful in the short term is it lets us implement piece by piece.

PR #1007, opened 2021-05-11
Localize MacOS installer

bakulf:


strseb (approved):

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,42 @@ bool dockClickHandler(id self, SEL cmd, ...) {
 void MacOSUtils::showDockIcon() {
   [NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
 }
+
+void MacOSUtils::installerStrings() {
+  //% "Mozilla VPN for MacOS"
```
flodolo, line 94: nit: macOS

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,42 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+  //% "A previous installation of Mozilla VPN exists at /Applications/Mozilla VPN.app. This"
+  //% "installer will remove the previous installation prior to installing. Please back up any data"
+  //% "before proceeding."
+  qtTrId("macosinstaller.previous_build.message");
+
+  //% "You will now be guided through the installation steps for the Mozilla VPN. Thank you for"
+  //% "choosing your VPN from the trusted pioneer of internet privacy."
+  qtTrId("macosinstaller.welcome.message1");
+
+  //% "Click \"Continue\" to continue the setup."
```
flodolo, line 115: We should use typographical quotes here, instead of escaping

```
   //% "Click “Continue” to continue the setup."
```

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,42 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+  qtTrId("macosinstaller.welcome.message2");
+
+  //% "Success!"
+  qtTrId("macosinstaller.conclusion.title");
+
+  //% "The Mozilla VPN is successfully installed. Go to your Applications folder to open up the VPN
+  // and start taking control of your online privacy."
+  qtTrId("macosinstaller.conclusion.message1");
+
+  //% "Trouble with this install?"
```
flodolo, line 125: Should it be installation here?

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+        for child in node:
+            if child.tag != nodeName:
+                continue
+            if child.text == None:
+                return ''
+            return child.text
+    return ''
+
+# Generate the translation folder for a particular locale. `nodeTag` can be
+# `source` (for english only) or `translation` (for anything else)
```
flodolo, line 36: nit: English only

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+def translate(root, locale, nodeTag):
+  folder = os.path.join('macos', 'pkg', 'Resources', locale + '.lproj')
+  os.mkdir(folder)
+
+  localizableFile = os.path.join(folder, 'Localizable.strings')
+  f = open(localizableFile, 'w')
```
flodolo, line 42: You can use with and avoid `close()` at the end.

```
  with open(localizableFile, 'w') as f:
      for key in keys:
          f.write(f"'{key.replace('macosinstaller.', '').replace('.', '_')}' = '{getTranslation(root, key, nodeTag)}';\n")
```

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+  for key in keys:
+      f.write(f"'{key.replace('macosinstaller.', '').replace('.', '_')}' = '{getTranslation(root, key, nodeTag)}';\n")
```
flodolo, line 44: Looking at https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html

They seem to use double quotes as delimiter? Independently from that, I assume you need to escape the delimiter character in the translation.

I also think it would be more readable if you transform the key in a separate variable.

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+  f.write('<br />\n')
+  f.write(f'<p style="margin: 5px;">{getTranslation(root, "macosinstaller.conclusion.message2", nodeTag)}\n');
+  f.write(f'<a rel="noopener noreferrer" href="https://support.mozilla.org/products/firefox-private-network-vpn?utm_source=mozilla-vpn&utm_medium=mozilla-vpn-installer&utm_campaign=mac-installer" style="color: #0a84ff;">{getTranslation(root, "macosinstaller.conclusion.message3", nodeTag)}</a></p>\n')
+  f.write('</body>\n</html>')
+  f.close()
+
+# Let's check if the `macos/pkg/Resources` folder exists.
+resourceFolder = os.path.join('macos', 'pkg', 'Resources')
+if not os.path.isdir(resourceFolder):
+    print(f"Folder '{resourceFolder}' should exist!")
```
flodolo, line 74: You can exit directly with `exit(f"Folder '{resourceFolder}' should exist!")`.

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+# corresponding `Resources` folder.
+for translation in os.listdir('translations'):
+    translationFile = os.path.join('translations', translation)
+    if not os.path.isfile(translationFile):
+        continue
+
+    if not translation.startswith('mozillavpn_'):
+        continue
+
+    tree = ET.parse(translationFile)
```
flodolo, line 104: You probably want the rest of this code in a try…except?

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+        if not fullTranslated:
+            break
+    if not fullTranslated:
+        continue
+
+    locale = translation.split(".")[0]
+    locale = locale.split("mozillavpn_")[1]
+    translate(root, locale, 'translation')
+
+# Finally, the english translation.
```
flodolo, line 119: nit: English

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,127 @@
...
+            if child.tag != nodeName:
+                continue
+            if child.text == None:
+                return ''
+            return child.text
+    return ''
+
+# Generate the translation folder for a particular locale. `nodeTag` can be
+# `source` (for english only) or `translation` (for anything else)
+def translate(root, locale, nodeTag):
```
flodolo, line 37: You could avoid having `nodeTag` at all in the function, and just have at the beginning

```
noteTag = "source" if locale == "en" else "translation"
```

Right now the logic is spread a bit all over the place.

flodolo (changes requested):
Mostly nits, but there's definitely one issue (escaping delimiter characters when building the .strings file).

Could you run the Python file through `black` to make the syntax a bit more consistent?

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,42 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+  //% "Unable to install"
+  qtTrId("macosinstaller.unsupported_version.title");
+
+  //% "Mozilla VPN requires Mac OS X 10.6 or later."
+  qtTrId("macosinstaller.unsupported_version.message");
+
+  //% "Previous Installation Detected"
+  qtTrId("macosinstaller.previous_build.title");
+
+  //% "A previous installation of Mozilla VPN exists at /Applications/Mozilla VPN.app. This"
```
flodolo, line 106: You need a space at the end of these sentence that continue on a new line. 

This is how the string currently looks like when exported

```
    <message id="macosinstaller.previous_build.message">
        <source>A previous installation of Mozilla VPN exists at /Applications/Mozilla VPN.app. Thisinstaller will remove the previous installation prior to installing. Please back up any databefore proceeding.</source>
    </message>
```

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,42 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+  //% "choosing your VPN from the trusted pioneer of internet privacy."
+  qtTrId("macosinstaller.welcome.message1");
+
+  //% "Click “Continue” to continue the setup."
+  qtTrId("macosinstaller.welcome.message2");
+
+  //% "Success!"
+  qtTrId("macosinstaller.conclusion.title");
+
+  //% "The Mozilla VPN is successfully installed. Go to your Applications folder to open up the VPN
```
flodolo, line 121: You're missing quotes here, and this cuts the second part of the string off in the export.

scripts/import_macos_pkg_resources.py
```diff
@@ -0,0 +1,133 @@
...
+
+    fullTranslated = True
+    for key in keys:
+        fullTranslated = getTranslation(root, key, 'translation') != ''
+        if not fullTranslated:
+            break
+    if not fullTranslated:
+        continue
+
+    translate(root, locale, 'translation')
```
flodolo, line 120: This still has 3 parameters

flodolo (changes requested):

src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,44 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+  //% "Unable to install"
+  qtTrId("macosinstaller.unsupported_version.title");
+
+  //% "Mozilla VPN requires Mac OS X 10.6 or later."
+  qtTrId("macosinstaller.unsupported_version.message");
+
+  //% "Previous Installation Detected"
+  qtTrId("macosinstaller.previous_build.title");
+
+  //% "A previous installation of Mozilla VPN exists at /Applications/Mozilla "
```
flodolo, line 106: Please fix this, now you have a space in the middle of the file path.

flodolo (approved):

src/platforms/macos/macosutils.mm
bakulf: No this is right. The path is `/Applications/Mozilla VPN.app`

flodolo: Uh, completely missed that 🤦🏼

PR #1009, opened 2021-05-12
Add glean events

lesleyjanenorton:


glean/metrics.yaml
```diff
@@ -14,13 +14,31 @@
...
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1708637
+    data_reviews:
+      - TBD
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - amarchesini@mozilla.com
+    expires: never
+
```
bakulf, line 33: Extra space here and elsewhere in this file.

src/ui/components/VPNConnectionStability.qml
```diff
@@ -72,6 +73,21 @@ RowLayout {
         }
     ]
 
+    transitions: [
```
bakulf, line 76: It could be that this qml file is not rendered (for instance, if we have the settings opened). What about if we trigger this event from C++?
Instead of this block of code, you can add here: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/connectionhealth.cpp#L68

```
if (stability == Unstable) {
  emit MozillaVPN::instance()->triggerGleanSample(GleanSample::connectionHealthUnstable);
} else if (stability == NoSignal) {
  emit MozillaVPN::instance()->triggerGleanSample(GleanSample::connectionHealthNoSignal);
}
```

At the beginning of the same file, here: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/connectionhealth.cpp#L6 (alphabetic order) add:

```
#include "gleansample.h"
```

src/ui/states/StateSubscriptionBlocked.qml
```diff
@@ -33,6 +33,8 @@ VPNStackView {
             buttonOnClick: stackview.handleButtonClick,
             signOffLinkVisible: true,
             getHelpLinkVisible: false
-        }
-    )
+            }
+        );
+        Glean.sample.SubscriptionBlockedViewed.record();
```
bakulf, line 38: `s` vs `S`. It should be `subscriptionBlockedViewed`

src/ui/views/ViewErrorFullScreen.qml
```diff
@@ -121,7 +122,10 @@ VPNFlickable {
                 anchors.horizontalCenter: undefined
                 anchors.bottom: undefined
                 anchors.bottomMargin: undefined
-                onClicked: stackview.push(getHelpComponent)
+                onClicked: {
+                    Glean.sample.GetHelpClickedFullScreenError.record();
```
bakulf, line 126: `getHelpClickedFullScreenError` ?

src/ui/views/ViewInitialize.qml
```diff
@@ -15,7 +17,10 @@ Item {
         objectName: "getHelpLink"
 
         labelText: qsTrId("vpn.main.getHelp")
-        onClicked: stackview.push(getHelpComponent)
+        onClicked: {
+            Glean.sample.GetHelpClickedViewInitialize.record();
```
bakulf, line 21: same here.

bakulf (changes requested):

bakulf (approved):

bakulf (approved):

PR #1016, opened 2021-05-12
Debounce Captive Portal Notification

strseb:
2 Bugs here to solve. 

On desktop when the window loses focus, we suspend ping-send worker. This thus we set stability to "stable".
This means when the user connects to a cp-wifi while the app is in the background, we would not start detection. 

Also i limit the amound of Captive Portal requests to only one after we're no longer stable. 
Reason here beeing, connection health might move from stable->instable->noSignal causing 2 cp detections and  2 notifications.
We would need to go to "stable" once again to trigger a new one.

### DIFF
--- a/src/captiveportal/captiveportaldetection.cpp
+++ b/src/captiveportal/captiveportaldetection.cpp
@@ -13,7 +13,6 @@
 #include "mozillavpn.h"
 #include "settingsholder.h"
 
-
 namespace {
 Logger logger(LOG_CAPTIVEPORTAL, "CaptivePortalDetection");
 }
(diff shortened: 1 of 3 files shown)


src/captiveportal/captiveportaldetection.cpp
```diff
@@ -45,6 +45,12 @@ void CaptivePortalDetection::stateChanged() {
       vpn->controller()->state() != Controller::StateConfirming) {
     logger.log() << "No captive portal detection required";
     m_impl.reset();
+    // Since we now reached a stable state, on the next time we have an instablity check for portal again.
+    m_shouldRun= true;
```
bakulf, line 49: apply clang-format

src/captiveportal/captiveportaldetection.cpp
```diff
@@ -103,6 +109,7 @@ void CaptivePortalDetection::detectionCompleted(CaptivePortalResult detected) {
   logger.log() << "Detection completed:" << detected;
 
   m_impl.reset();
+  m_shouldRun = false;
```
bakulf, line 112: Should we do it only for NoPortal/PortalDetected?

bakulf (approved):

Issue #593, opened 2021-02-17
Support IPv6 captive portal detection 


bakulf:
This is not supported yet because we don't have the IPv6 filtering for `IPAddress` yet.

PR #1018, opened 2021-05-12
Remove Unresolved CaptivePortal Todo code 

strseb:
Since we have #593 still open, we have no code to filter a certain v6 adress out. 
So currently instead we add "[cp-address-v6]/0" to the allowed ipv6 adress ranges, which is not correct.
So i think we should remove this todo for now :)


bakulf (approved):

bakulf:
Can you fix the linter issue?

PR #1022, opened 2021-05-13
Final death of linux helper script

birdsarah:
@oskirby and I will work on the final death of the linux helper script.


birdsarah:
@oskirby there's some code reorg I'd like to do:

1) I don't think the go module belongs under src/platforms/linux but rather at the outer level.
2) I'd like to follow the existing pattern for now and to move these new functions into their own module FirewallUtils for now. As per my comments on #1002, I *think* it's a useful structure and when it turns out not to be I'm happy to eat my words and do the clean-up.
3 Have you already reviewed the dependencies for this go module to make sure they're all MPL compatible?

I'm not suggesting you have to do all this work, we can split it, just checking in with you on it.

oskirby:
@birdsarah 
1. Yeah, we can move the Go module outside of the `src/` tree.
2. I find that splitting each step into its own module gets messy when there are platform-specific details that need to be shared between the modules (for example, the firewall mark needs to be set by `wg_set_device()` and then also needs to be provided to the firewall). My preference would be to keep all the "interface setup" stuff in `WireguardUtils`. We can still do incremental changes from `helper.sh` into `WireguardUtils` by providing do-nothing implementations for the methods in `WireguardUtils` and then call the helper script at the end to perform any steps that aren't implemented in `WireguardUtils` yet.
3. I have not reviewed the licenses of the Go dependencies for compatibility with the MPL.

oskirby:
The netfilter module depends on the following modules, all of which are compatible with the MPL:
 - github.com/google/go-cmp: BSD 3-clause License
 - github.com/google/nftables: Apache License 2.0
 - github.com/jsimonetti/rtnetlink: MIT License
 - github.com/koneu/natend: Unlicense/Public Domain
 - github.com/vishvananda/netns: Apache License 2.0
 - golang.org/x/crypto: BSD 3-clause License
 - golang.org/x/net: BSD 3-clause License
 - golang.org/x/text: BSD 3-clause License

src/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -0,0 +1,193 @@
...
+  m_linkDomains.clear();
+
+  /* Revert the VPN interface's DNS configuration */
+  if (m_ifindex > 0) {
+    QList<QVariant> argumentList = {QVariant::fromValue(m_ifindex)};
+    QDBusPendingReply<> reply = m_resolver->asyncCallWithArgumentList(
+        QStringLiteral("RevertLink"), argumentList);
+
+    QDBusPendingCallWatcher* watcher = new QDBusPendingCallWatcher(reply, this);
+    QObject::connect(watcher, SIGNAL(finished(QDBusPendingCallWatcher*)), this,
```
bakulf, line 66: This method is called during the shutdown. Wondering if the use of `this` here can be an issue.

bakulf: Same before when using `this` as the parent of QDbusPendingCallWatcher.

src/platforms/linux/daemon/dnsutilslinux.h
```diff
@@ -0,0 +1,39 @@
...
+
+ public:
+  DnsUtilsLinux(QObject* parent);
+  ~DnsUtilsLinux();
+  bool updateResolvers(const QString& ifname,
+                       const QList<QHostAddress>& resolvers) override;
+  bool restoreResolvers() override;
+
+ private:
+  int m_ifindex = 0;
```
bakulf, line 25: move these at the end of the file.

src/platforms/linux/daemon/dnsutilslinux.h
```diff
@@ -0,0 +1,39 @@
...
+#define DNSUTILSLINUX_H
+
+#include "daemon/dnsutils.h"
+#include "dbustypeslinux.h"
+
+#include <QDBusInterface>
+#include <QDBusPendingCallWatcher>
+
+class DnsUtilsLinux final : public DnsUtils {
+  Q_OBJECT
```
bakulf, line 15: `Q_DISABLE_COPY_MOVE(DnsUtilsLinux)`

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -83,17 +135,62 @@ bool WireguardUtilsLinux::configureInterface(const InterfaceConfig& config) {
...
+  }
+  uint32_t blockClassId = getCgroupClass(VPN_BLOCK_CGROUP);
+  if (blockClassId != 0) {
+    NetfilterBlockCgroup(blockClassId);
+  }
+  if (config.m_ipv6Enabled) {
+    GoString goIpv6Address = {.p = qPrintable(config.m_deviceIpv6Address),
+                              .n = config.m_deviceIpv6Address.length()};
+    if (config.m_deviceIpv6Address.contains('/')) {
+      goIpv6Address.n = config.m_deviceIpv6Address.indexOf('/');
```
bakulf, line 175: Double string search here. It's OK but we can also do;
```
int slashPos = config.m_deviceIpv6Address.indexOf('/');
if (slashPos != -1) goIpv6Address.n = slashPos;
```

bakulf (approved):

src/platforms/linux/daemon/dnsutilslinux.cpp
oskirby: I believe that `this` is still valid within the scope of a destructor, so I wouldn't expect this to cause a crash. However, I'm unsure whether the order of operations would garbage collect the connections and the `QObject` hierarchy when this occurs during shutdown. I'll rewrite the destructor to try and make this a bit more explicit.

PR #1026, opened 2021-05-13
Fix #780 - Confusing sign out language

### LINKED ISSUE #780: Device removed message displayed upon signing out
When I signed out of Mozilla VPN, the app showed a message saying "Signed out and device removed". This message is confusing. Which "device" is be removed? And removed from what?

Found by: cpeterson
Platform: Windows

lesleyjanenorton:
- Changes the sign out alert to green
- Tweaks the sign out message

@flodolo Do you think this requires adding a new string? It is a one word change and, without knowing how this has been handled in other locales, I'm wondering if it makes sense to keep the existing string ID and update the English string only.

Screen Shot 2021-05-13 at 4 30 08 PM

### DIFF
--- a/src/ui/components/VPNSystemAlert.qml
+++ b/src/ui/components/VPNSystemAlert.qml
@@ -5,11 +5,13 @@
 import QtQuick 2.0
 import QtQuick.Controls 2.5
 import Mozilla.VPN 1.0
+import "../themes/themes.js" as Theme
 
 VPNAlert {
     id: alertBox
 
     state: VPN.alert
+
     states: [
         State {
             name: VPN.NoAlert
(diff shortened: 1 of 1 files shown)


flodolo:
The automation for VPN will invalidate that string anyway, even if you don't change xthe string ID.

For any other project, I'd require a new ID for this type of change. If it's confusing for en-US, it's likely confusing for other languages as well.

flodolo:
(btw, see [this blog post](https://blog.mozilla.org/l10n/2021/05/01/mozilla-vpn-client-a-localization-tale/) for a description of how the automation is set up)

lesleyjanenorton:
thank you @flodolo

bakulf (approved):

Issue #1033, opened 2021-05-14
macOS installer strings are not exported to .ts


flodolo:
I originally filed the [issue](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/issues/136) in the l10n repository, but at this point this sounds like a code issue that should be investigated here.

`lupdate` on Linux is not exporting the `macosinstaller.*` strings, this is breaking the automation to expose new strings for localization. It's broken in automation (Ubuntu 20.04 with qt 5.15.2), and it's broken for me in a VM with 5.12.8.

It is also visible in the linter automation in this repository.

```
Updating 'vpn/translations/mozillavpn_en.ts'...
    Found 181 source text(s) (181 new and 0 already existing)
```

Running the script on macOS generates 192 strings

```
Updating 'translations/mozillavpn_en.ts'...
    Found 192 source text(s) (192 new and 0 already existing)
```

Note that this is breaking automation, and it's something that should be figured as soon as possible. Any new strings will create pull requests that remove all the installer strings.

Since you're using the .ts file to import strings for the installer, I'd expect the work in #1007 to be broken as well? Or maybe that works, because strings are imported from the localized XLIFF, and they're currently available in those.

flodolo:
I think I've found where the problem is.

[This page](https://doc.qt.io/qt-5/linguist-programmers.html) says that

> The lupdate tool extracts user interface strings from your application. It reads the application .pro file to identify which source files contain text to be translated. This means your source files must be listed in the SOURCES or HEADERS entry in the .pro file, or in resource files listed in the RESOURCE entry. **If your files are not listed, the text in them will not be found.**

Looking at src.pro, I see

```
    SOURCES += \
            platforms/macos/macosmenubar.cpp \
            platforms/macos/macospingsendworker.cpp \
            platforms/macos/macosstartatbootwatcher.cpp \
            systemtraynotificationhandler.cpp \
            tasks/authenticate/desktopauthenticationlistener.cpp

    OBJECTIVE_SOURCES += \
            platforms/macos/macoscryptosettings.mm \
            platforms/macos/macosnetworkwatcher.mm \
            platforms/macos/macosutils.mm
```

If I copy `platforms/macos/macosutils.mm` to SOURCES, then `lupdate` exports 192 strings instead of 181 also on Linux.

I guess on macOS OBJECTIVE_SOURCES is considered, on Linux is not.

[...]

PR #1036, opened 2021-05-17
Move the macos localization string in a cross-platform file - closes …

bakulf:
…#1033


flodolo (approved):
I see 192 strings in the Linter, we should be good
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1036/checks?check_run_id=2599532698

I'll have to do some work to move strings around, since they move from one file to another.

Issue #988, opened 2021-05-07
Surveys per platform


bakulf:
Currently, the app fetches a list of the surveys from the guardian website REST API. We should extend the REST API to expose the list of platforms for each survey. In this way, we can have surveys for IOS, for Android and so on.

The API is this: `/api/v1/vpn/surveys` and the response is:

```
[{
  "id":"unique_id"
  "url":"http://vpn.mozilla.org",
  "trigger_time":4
  },{
  "id":"unique_id2",
  "url":"http://vpn.mozilla.org",
  "trigger_time":10
}]
```

I would suggest to add a `platforms` attribute for each survey object. The value can be an array with the list of platform for that survey. Something like:

```
[{
  "id":"unique_id"
  "url":"http://vpn.mozilla.org",
  "trigger_time":4
  // This is for each platform.
  },{
  "id":"unique_id2",
  "url":"http://vpn.mozilla.org",
  "trigger_time":10
 "platforms": [ "windows", "ios" ],
}]
```

ValentinaPC:
Is there a way for QA to verify this? Thanks!

PR #1043, opened 2021-05-17
Add optional filter for surveys by platform

oskirby:
This allows surveys to include a `platforms` array, which lists the platforms on which a survey should be trigger-able, if the array is absent, or empty, then the survey should apply to all platforms.

Closes: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/988

### DIFF
--- a/src/models/survey.cpp
+++ b/src/models/survey.cpp
@@ -91,7 +91,7 @@ bool Survey::isTriggerable() const {
     return false;
   }
 
-  if ((m_platforms.count() > 0) &&
+  if (!m_platforms.isEmpty() &&
       !m_platforms.contains(Constants::PLATFORM_NAME)) {
     return false;
   }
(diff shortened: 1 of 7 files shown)


src/models/survey.cpp
```diff
@@ -74,6 +91,11 @@ bool Survey::isTriggerable() const {
     return false;
   }
 
+  if ((m_platforms.count() > 0) &&
```
bakulf, line 94:
```
if (!m_platfrorms.isEmpty() ...
```

src/mozillavpn.cpp
```diff
@@ -421,17 +421,7 @@ void MozillaVPN::openLink(LinkType linkType) {
...
-#elif defined(MVPN_MACOS)
-      url.append("macos");
-#elif defined(MVPN_IOS)
-      url.append("ios");
-#elif defined(MVPN_ANDROID)
-      url.append("android");
-#else
-      url.append("dummy");
-#endif
+      url.append(Constants::PLATFORM_NAME);
```
bakulf, line 424: great! Thanks for this!

bakulf (approved):

PR #1058, opened 2021-05-18
Linux daemon should output logs to /var/log

oskirby:
The linux daemon is started as a D-Bus service by systemd, in which case it will not have a valid $HOME directory and it would be more appropriate to send out logs to /var/log.

Closes: mozilla-mobile/mozilla-vpn-client/issues/1057

### DIFF
--- a/src/loghandler.cpp
+++ b/src/loghandler.cpp
@@ -25,6 +25,8 @@ constexpr const char* LOG_FILENAME = "mozillavpn.txt";
 
 namespace {
 QMutex s_mutex;
+QString s_location =
+    QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
 LogHandler* s_instance = nullptr;
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


src/loghandler.cpp
```diff
@@ -233,26 +222,49 @@ void LogHandler::writeLogs(QTextStream& out) {
...
+    QFile file(logFileName);
     file.remove();
   }
 
   s_instance->openLogFile(lock);
 }
 
+// static
+void LogHandler::setLocation(const QString& path) {
+  s_location = path;
```
bakulf, line 242: Super racy, but, I would suggest to add `QMutexLocker lock(&s_mutex);`

src/loghandler.cpp
```diff
@@ -233,26 +222,49 @@ void LogHandler::writeLogs(QTextStream& out) {
...
+
+  if (s_instance && s_instance->m_logFile) {
+    cleanupLogs();
```
bakulf, line 245: Maybe pass `lock` here and have 2 `cleanupLogs()` methods, 1 with the proofOfLock and 1 without it.

bakulf (changes requested):

src/loghandler.cpp
oskirby: Good point, I'll revise the PR with locking added.

bakulf (approved):

PR #1065, opened 2021-05-19
Experiment: function Tests Matrix-Run 

strseb:
Experiment: instead of running all tests in one loop and generate a huge logfile (imho), 
maybe we can have a job automatically for every test so it's faster to spot what is failing and why.


bakulf:
This is wonderful!

Issue #896, opened 2021-04-26
combine staging and production builds


bhearsum:
Currently, we hardcode the VPN server at build time. This means we have separate builds for `staging` and `production`. This has a couple of downsides:
1) We produce double the amount of builds needed
2) We're QA'ing different bits than we ship.

Elsewhere, (eg: Desktop Firefox and mobile versions) we allow things like this to be changed at runtime, or by configs, which eliminates both of the issues above. See https://mozilla.github.io/ecosystem-platform/docs/process/using-the-staging-environment#ios for an example of what we do on iOS.

PR #1076, opened 2021-05-20
Add hidden developer options and merge prod/staging builds

oskirby:
This adds a place to put developer features without getting in the way of the normal UX. The immediate needs I was trying to address are for feature flags so that we can hide features that are not fully tested, and to add a UI to switch between staging and production (see: #896)

I decided to put this hidden menu in the Get Help page so that it can be accessed before the sign-in flow. Enabling this menu is similar to Android's developer options menu, and is achieved by tapping or clicking on the `Get Help` text at the top of the Get Help menu 6 times within 10 seconds. The menu can be hidden again using the `Show Developer Developer Options` checkbox.

~~So far, this is just UI, the `Use Staging Servers` and `Enable Untested Features` checkboxes don't do anything other than storing their state in the settings.~~

### Get Help with Developer Options Enabled
mozilla-vpn-dev-enabled
### Developer Options Menu
mozilla-vpn-dev-menu


oskirby:
I have merged in some of the patches from #910 into this pull request in order to connect the UI and settings up to the underlying mechanisms to switch between production and staging. Unlike most settings, this will require a restart of the VPN to take effect so a warning will be displayed whenever the setting disagrees with the effective mode.

To keep the functional tests working, I've also added an option to the CLI to force the UI into staging mode, regardless of whatever the stored settings happen to be.

oskirby:
I am closing this in favor of a slightly newer PR #1327

ValentinaPC:
@oskirby did we drop the untested features option? Thanks!

PR #1091, opened 2021-05-23
Consider just .ts files for the macos installer localization

bakulf:


scripts/import_macos_pkg_resources.py
```diff
@@ -94,7 +95,7 @@ def translate(root, locale):
     if not os.path.isfile(translationFile):
         continue
 
-    if not translation.startswith('mozillavpn_'):
+    if not translation.startswith('mozillavpn_') or not translation.endswith('.ts'):
```
flodolo, line 98: I'm curious, what other `mozillavpn_*` files are in there besides .ts files? From my tests I haven't seen any.

bakulf: Compiling the app, the translation files are converted in .qm files. Same name, different extension. I just need to ignore them with that line.

flodolo (approved):

PR #1092, opened 2021-05-23
Use DNT/GPC to inform the server when the user doesn't want to be tracked - closes #1089

### LINKED ISSUE #1089: Consider implementing "Do not track" header
The next step, after the "telemetry policy view", is to communicate the policy in the HTTP requests using the DnT header.

bakulf:
Of course, this is just the client-side. Then we need extra work to parse the header and follow what the user asks.


birdsarah (approved):
tests?

bakulf:
So far, we are unable to test/inspect network requests. Our unit-tests have a mock network request implementation.

PR #1094, opened 2021-05-24
IP Lookup in ipv4 and ipv6 (if supported)

bakulf:


src/connectiondataholder.cpp
```diff
@@ -216,50 +217,44 @@ void ConnectionDataHolder::updateIpAddress() {
...
+          return;
+        }
+
+        logger.log() << "IP address request completed";
+        if (m_checkStatusTimer.isActive() &&
+            country != MozillaVPN::instance()->currentServer()->countryCode()) {
+          // In case the country-we're reported in does not match the
+          // connected server we may retry only once.
+          logger.log() << "Reported ip not in the right country, retry!";
+          TimerSingleShot::create(this, 3000, [this]() { updateIpAddress(); });
```
oskirby, line 237: I don't see the mechanism here that would prevent this from looping forever every 3000ms in the event that the country code never matches.

src/ipfinder.cpp
```diff
@@ -0,0 +1,163 @@
...
+  if (hostInfo.error() != QHostInfo::NoError) {
+    logger.log() << "Unable to perform a DNS Lookup:" << hostInfo.errorString();
+    emit completed(QString(), QString(), QString());
+    deleteLater();
+    return;
+  }
+
+  bool ipv6Enabled = SettingsHolder::instance()->ipv6Enabled();
+
+  for (const QHostAddress& address : hostInfo.addresses()) {
```
oskirby, line 58: This isn't necessarily an issue because we control the DNS records for the API, but it's not uncommon to get a long list of IP addresses back when resolving a hostname, but we only really need a single IPv4 and IPv6 address to complete the IP check.

oskirby (approved):
looks good to me

src/connectiondataholder.cpp
bakulf: This is a good point. But I haven't touched this part. Can we work on it as a follow up?

PR #1107, opened 2021-05-25
Glean pings sent more often (20 minutes)

bakulf:
In prod we used to send pings once per hour. But because we don't store data, probably it's better to speed up the ping sending. @brizental what do you think about this change?


brizental (approved):
The change looks good.

Probably we will have a better idea of what is the best ping scheduling once we start collecting data. Let's keep this open for change once that happens :)

PR #1111, opened 2021-05-26
Fix functional tests

bakulf:


bakulf:
This doesn't work yet because of the device limit. We need to rotate the accounts in chunks.

bakulf:
Ok... it's not a proper fix, but at least it's better than before.

strseb (approved):

PR #1122, opened 2021-05-28
Build fixes and RPM packaging for Fedora 33+

oskirby:
After a bit of experimenting, I've been able to build and run the VPN client on Fedora 33 and 34, and I've written a spec file that should enable us to build binary RPM packages out of a git checkout by doing:

```
sudo dnf builddep linux/mozillavpn.spec
rpmbuild -bb linux/mozillavpn.spec
```

Some users have noted problems with SELinux, but I wasn't able to reproduce them in my virtual machine, it just seemed to work for me, even with `getenforce` reporting that SELinux is enabled.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -440,7 +440,7 @@ else:linux:!android {
     INSTALLS += dbus_service
 
     systemd_service.files = ../linux/debian/mozillavpn.service
-    systemd_service.path = /lib/systemd/system
+    systemd_service.path = /usr/lib/systemd/system
     INSTALLS += systemd_service
 
     CONFIG += link_pkgconfig
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ curl -L https://download.qt.io/archive/qt/5.15/5.15.1/single/qt-everywhere-src-5
 tar xvf qt-everywhere-src-5.15.1.tar.xz
 mv qt-everywhere-src-5.15.1 qt
 sudo apt build-dep qt5-default
-sudo apt install libxcb-xinerama0-dev
+sudo apt install clang llvm golang
+sudo apt install libxcb-xinerama0-dev libxcb-util-dev
 bash scripts/qt5_compile.sh qt qt
 ```
 
@@ -60,7 +61,7 @@ git submodule update
 # glean
 ./scripts/generate_glean.py
 # translations
-python scripts/importLanguages.py
+./scripts/importLanguages.py
 ```
 
 #### Build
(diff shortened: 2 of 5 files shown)


src/src.pro
```diff
@@ -435,7 +435,7 @@ else:linux:!android {
     INSTALLS += dbus_service
 
     systemd_service.files = ../linux/debian/mozillavpn.service
-    systemd_service.path = /lib/systemd/system
+    systemd_service.path = /usr/lib/systemd/system
```
bakulf, line 438: Do we need to include this change in 2.3?

bakulf (approved):

src/src.pro
oskirby: This install target only exists in the qmake project to satisfy users who want to install from sources, and on Debian-based distributions it makes no difference since /usr/lib/systemd and /lib/systemd are hardlinked to the same place.

bakulf:
Can you rebase this patch? Then merge it directly. Thank you!
Do you think it makes sense to write a few lines to describe how to create a RPM package?

oskirby:
Rebasing this seems to result in a a segfault deep in QtQuick and OpenGL when running on Fedora due to commit 1d4dee17d5b2060012c49d3ee09bea8386c852a3 so I think that there is something still not quite right the ring animation handling.

oskirby:
I think my comment blaming canvas change in #1113 for the crash is incorrect, however, it still seems to occur on Fedora despite being fixed on Ubuntu.

oskirby:
[mozillavpn-crash-log.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6600269/mozillavpn-crash-log.txt)

BrentonPoke:
Does this work for anyone? I get an error when trying to build it in 34.
```
make[1]: Entering directory '/home/brentonpoke/rpmbuild/BUILD/tests/unit'
make[1]: Nothing to be done for 'first'.
make[1]: Leaving directory '/home/brentonpoke/rpmbuild/BUILD/tests/unit'
make[1]: Entering directory '/home/brentonpoke/rpmbuild/BUILD/src'
lupdate /home/brentonpoke/Documents/mozilla-vpn-client/src -no-obsolete -ts /home/brentonpoke/Documents/mozilla-vpn-client/src/../translations/mozillavpn_en.ts
make[1]: lupdate: No such file or directory
make[1]: *** [Makefile:1089: ts] Error 127
make[1]: Leaving directory '/home/brentonpoke/rpmbuild/BUILD/src'
make: *** [Makefile:48: sub-src-make_first] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.gQo43X (%build)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.gQo43X (%build)
```

PR #1124, opened 2021-05-28
Support for hirsute ubuntu

bakulf:


linux/debian/control.prod.hirsute
```diff
@@ -0,0 +1,102 @@
...
+               flex,
+               libqt5charts5-dev (>=5.15.2-2),
+               libqt5networkauth5-dev (>=5.15.2-2),
+               qtbase5-dev (>=5.15.2+dfsg-5),
+               qtbase5-dev-tools (>=5.15.2+dfsg-5),
+               qtdeclarative5-dev (>=5.15.2+dfsg-5),
+               qtdeclarative5-dev-tools (>=5.15.2+dfsg-5),
+               qt5-qmake-bin (>=5.15.2+dfsg-5),
+               qttools5-dev-tools (>=5.15.2-5),
+               libxcb-render0-dev,
```
oskirby, line 17: I would have expected that building against the Ubuntu-provided versions of Qt, we shouldn't need to explicitly call out all of the X11 stuff as Build-Depends since we don't actually link to any X11 stuff in our code directly.

oskirby (approved):

PR #1133, opened 2021-06-01
Support per-app VPN on Linux

oskirby:
This is a rebase of PR  #996 to merge the firewall and daemon together into the same D-Bus servce and to incorporate the changes resulting from the removal of the Linux helper script in commit 2f291e35db90ada46fbbe95c3c0d891364ed3a23

This change still lacks a solution to the shared DNS cache/resolvers provided by systemd, so there still exists a potential for the VPN to be detected by an application excluded from the VPN, but I think that should be addressed as a separate issue.

### DIFF
--- a/src/platforms/linux/linuxapplistprovider.h
+++ b/src/platforms/linux/linuxapplistprovider.h
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef LINUXAPPLISTPROVIDER_H
+#define LINUXAPPLISTPROVIDER_H
+
+#include <applistprovider.h>
+#include <QObject>
+#include <QProcess>
+
+class LinuxAppListProvider : public AppListProvider {
+  Q_OBJECT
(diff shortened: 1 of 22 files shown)


src/platforms/linux/linuxapplistprovider.h
```diff
@@ -0,0 +1,23 @@
...
+#define LINUXAPPLISTPROVIDER_H
+
+#include <applistprovider.h>
+#include <QObject>
+#include <QProcess>
+
+class LinuxAppListProvider : public AppListProvider {
+  Q_OBJECT
+ public:
+  LinuxAppListProvider(QObject* parent);
```
bakulf, line 15: explicit

src/platforms/linux/linuxapplistprovider.h
```diff
@@ -0,0 +1,23 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef LINUXAPPLISTPROVIDER_H
+#define LINUXAPPLISTPROVIDER_H
+
+#include <applistprovider.h>
+#include <QObject>
+#include <QProcess>
+
+class LinuxAppListProvider : public AppListProvider {
```
bakulf, line 12: final

bakulf (approved):

PR #1135, opened 2021-06-01
keeping VPN active after cmd+w

### LINKED ISSUE #1086: [macOS] cmd+w completely quits the app
Running the last staging app on macOS (`2.3.0 (2.202105211512)`): closing the window with my mouse keeps the app running, but pressing cmd+W completely quits, while it should only close the window.

mbirghan:
Closes #1086

### DIFF
--- a/src/platforms/macos/macosmenubar.cpp
+++ b/src/platforms/macos/macosmenubar.cpp
@@ -6,6 +6,8 @@
 #include "leakdetector.h"
 #include "logger.h"
 #include "mozillavpn.h"
+#include "qmlengineholder.h"
+#include "platforms/macos/macosutils.h"
 
 #include <QAction>
 #include <QMenu>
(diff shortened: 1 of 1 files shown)


src/platforms/macos/macosmenubar.cpp
```diff
@@ -63,7 +65,10 @@ void MacOSMenuBar::initialize() {
   m_preferencesAction->setMenuRole(QAction::PreferencesRole);
   m_preferencesAction->setVisible(vpn->state() == MozillaVPN::StateMain);
 
-  m_closeAction = fileMenu->addAction("", vpn->controller(), &Controller::quit);
```
bakulf, line 66: Should we keep this too or not?

bakulf (approved):

src/platforms/macos/macosmenubar.cpp
mbirghan: We have a separate quit action on line 50 (52 after the patch) which does the same thing.
So we do not need to keep it.

PR #1148, opened 2021-06-04
Fix the wasm compilation

bakulf:


mbirghan (approved):
Looks good!

Thanks for catching that mistake.

PR #1151, opened 2021-06-04
Fix OpenGL framebuffer crash on Linux

oskirby:
In recent builds of the VPN client, I have been experiencing intermittent crashes when loading the main view. Backtraces show that a null `this` pointer is making its way into the OpenGL context during the Qt canvas creation and things go badly from there. I first encountered this on my Fedora/RPM pull request (mozilla-mobile/mozilla-vpn-client/pull/1122) but I think users have been experiencing this in the 2.3.0 release (mozilla-mobile/mozilla-vpn-client/issues/1150)


bakulf:
Good catch!

Issue #888, opened 2021-04-22
Use `qmldir` to import QML components


bakulf:
The use of qmldir is documented here: https://doc.qt.io/qt-5/qtqml-modules-qmldir.html
When this is done, we can use a similar approach to load `glean.js`.

data-sync-user:
➤ Andrea Marchesini commented:

No QA needed.

PR #1161, opened 2021-06-09
Update glean to v0.15 and force glean_parser to v3.5

oskirby:


src/ui/components/VPNConnectionStability.qml
```diff
@@ -6,7 +6,9 @@ import QtQuick 2.5
 import QtQuick.Layouts 1.14
 import Mozilla.VPN 1.0
 import "../themes/themes.js" as Theme
-import "/glean/load.js" as Glean
+
+import org.mozilla.Glean 0.15
+import generated 0.15
```
bakulf, line 11: Would be great to have a better name. Do you think is it possible?

bakulf (approved):

src/ui/components/VPNConnectionStability.qml
oskirby: I agree actually, I was thinking of renaming this to 'telemetry' but as a first cut I just wanted to keep it as close as possible to the Glean example app where they call it 'generated'

oskirby:
Would this also close #888 or is there more to the qmldir issue than fixing up the imports for `glean.js`?

PR #1163, opened 2021-06-09
Virtual price for IAP

bakulf:
The UI is... simply not done! What this PR does is the following:

- fetch the list of products from the guardian-website
- expose the products via a IAP model (sorted by type)
- trigger the subscription using the product ID

I haven't spent time on the UI because the UX has not been defined yet.

### DIFF
--- a/src/models/devicemodel.h
+++ b/src/models/devicemodel.h
@@ -22,7 +22,7 @@ class DeviceModel final : public QAbstractListModel {
   DeviceModel();
   ~DeviceModel();
 
-  enum ServerCountryRoles {
+  enum ModelRoles {
     NameRole = Qt::UserRole + 1,
     PublicKeyRole,
     CurrentOneRole,
--- a/src/platforms/ios/iaphandler.mm
+++ b/src/platforms/ios/iaphandler.mm
@@ -283,6 +283,7 @@ void IAPHandler::registerProduct(const QJsonValue& value) {
   product.m_name = obj["id"].toString();
   product.m_price = Constants::SUBSCRIPTION_CURRENCY_VALUE_USD;
   product.m_type = productTypeToEnum(obj["type"].toString());
+  product.m_featuredProduct = obj["featured_product"].toBool();
   if (product.m_type == ProductUnknown) {
     logger.log() << "Unknown product type:" << obj["type"].toString();
     return;
@@ -468,6 +469,7 @@ QHash<int, QByteArray> IAPHandler::roleNames() const {
   roles[ProductIdentifierRole] = "productIdentifier";
   roles[ProductPriceRole] = "productPrice";
   roles[ProductTypeRole] = "productType";
+  roles[ProductFeaturedRole] = "productFeatured";
   return roles;
 }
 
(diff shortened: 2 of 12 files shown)


src/models/devicemodel.h
```diff
@@ -22,7 +22,7 @@ class DeviceModel final : public QAbstractListModel {
   DeviceModel();
   ~DeviceModel();
 
-  enum ServerCountryRoles {
+  enum ModelRoles {
```
mbirghan, line 25: Why did you rename this?

src/platforms/ios/iaphandler.mm
```diff
@@ -309,12 +371,24 @@ - (void)request:(SKRequest*)request didFailWithError:(NSError*)error {
...
+  Q_ASSERT(m_pendingProducts.contains(productIdentifier));
+  Product& productData = m_pendingProducts[productIdentifier];
+  productData.m_price = priceValue;
+  productData.m_product = product;
+
+  m_registeredProducts.append(productData);
+  m_pendingProducts.remove(productIdentifier);
+
+  if (m_pendingProducts.isEmpty()) {
+    logger.log() << "All the products has been registered";
```
mbirghan, line 388: In `unknownProductRegistered` we add `(with failures)`. 
Can we be sure that we did not have any failures in a different product when logging this?

mbirghan (approved):
I did not look at the UI. Everything else looks good!

src/models/devicemodel.h
bakulf: because "ServerCountryRoles" is not a good name for a "DeviceModel".

src/platforms/ios/iaphandler.mm
bakulf: We do have logs when the registration fails. And also here a bit. Let's land this. We can improve the logging in a follow-up.

src/ui/components/VPNSubscriptionOption.qml
```diff
@@ -0,0 +1,186 @@
...
+                }
+
+                // TODO (maybe) - Do we want to add the subscription duration in months to the model?
+                property var subscriptionDuration: getSubscriptionDuration(productType)
+
+                //% "Monthly plan"
+                property string productSingleMonth: qsTrId("vpn.subscription.monthlyPlan")
+
+                //: %1 is replaced by the subscription duration in months. %2 is replaced by the total subscription cost.
+                //% "%1-month plan: %2"
```
flodolo, line 148: This won't work for most locales that have complex plural forms (imagine having to decline _month_ differently for 3 and 6). I'm going to suggest to use `Multiple months (%1): %2`, but it's far from great, and we should look into a proper fix if possible (i.e. changing the sentence)

@bakulf 
Imagine translating this into Italian, where the form "x-month" doesn't make any sense. It would be "Plan for x months". Now imagine having to translate it for 1 and 1+.

src/ui/components/VPNSubscriptionOption.qml
```diff
@@ -0,0 +1,186 @@
...
+
+            Layout.alignment: Qt.AlignTop
+            spacing: 0
+
+            VPNInterLabel {
+                // TODO: Compute savings and send to string
+                property var productSavings: "10%"
+
+                //: Appears on the in-app purchase view beside a subscription plan. "%1" is replaced by the percentage amount saved when selecting that plan.
+                //% "Save %1"
```
flodolo, line 171: This should have been `Save %1%`, with only the number saved in the variable. Here you're assuming to know how to correctly format 10% for the current locale. Is that the case?

If the latter is true, that would be great. Note that you'd have the same problem with currency formatting.

PR #1167, opened 2021-06-09
In app feedback

mbirghan:
This PR creates the backend code for sending the feedback to the guardian website. It includes:
- creation of the network request
- creation of the task to send the feedback message
- introducing a function that can be called to create and send the feedback

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -671,10 +671,10 @@ void MozillaVPN::submitFeedback(const QString& feedbackText, const qint8 rating,
     Q_ASSERT(buffer);
 
     delete out;
+    // buffer is getting copied by TaskSendFeedback so we can delete it afterwards
     scheduleTask(new TaskSendFeedback(feedbackText, *buffer, rating, category));
     delete buffer;
   });
-
 
   return;
 }
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -660,23 +660,48 @@ void MozillaVPN::removeDevice(const QString& deviceName) {
   }));
 }
 
-void MozillaVPN::submitFeedback(const QString& feedbackText, const qint8 rating, const QString& category) {
-  logger.log() << "Submit Feedback";
+void MozillaVPN::createFeedbackTask(const QString& feedbackText, const qint8 rating,
+      const QString& category, std::function<void(Task*)>&& a_finalizeCallback) {
+
+  logger.log() << "Create feedback task";
+
+  std::function<void(Task*)> finalizeCallback = std::move(a_finalizeCallback);
 
   QString* buffer = new QString();
   QTextStream* out = new QTextStream(buffer);
 
-  serializeLogs(out, [this, out, buffer, feedbackText, rating, category]{
+  serializeLogs(out, [finalizeCallback, out, buffer, feedbackText, rating, category]{
     Q_ASSERT(out);
     Q_ASSERT(buffer);
 
     delete out;
     // buffer is getting copied by TaskSendFeedback so we can delete it afterwards
-    scheduleTask(new TaskSendFeedback(feedbackText, *buffer, rating, category));
+    finalizeCallback(new TaskSendFeedback(feedbackText, *buffer, rating, category));
     delete buffer;
   });
 
   return;
+}
+
+void MozillaVPN::submitFeedback(const QString& feedbackText, const qint8 rating, const QString& category) {
+  logger.log() << "Submit Feedback";
+
+  createFeedbackTask(feedbackText, rating, category, [this](Task* task) { scheduleTask(task); });
+}
+
+void MozillaVPN::createFeedbackToResend(const QString& feedbackText, const qint8 rating, const QString& category) {
+  logger.log() << "Scheduling feedback to be resend";
+
+  createFeedbackTask(feedbackText, rating, category, [this](Task* task) { m_resendFeedbackTasks.append(task); });
+}
+
+void MozillaVPN::resendFeedback() {
+  logger.log() << "Resending feedback";
+
+  m_tasks.append(m_resendFeedbackTasks);
+  m_resendFeedbackTasks.clear();
+
+  maybeRunTask();
 }
 
 void MozillaVPN::accountChecked(const QByteArray& json) {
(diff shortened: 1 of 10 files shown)


src/mozillavpn.cpp
```diff
@@ -659,6 +660,25 @@ void MozillaVPN::removeDevice(const QString& deviceName) {
...
+  QString* buffer = new QString();
+  QTextStream* out = new QTextStream(buffer);
+
+  serializeLogs(out, [this, out, buffer, feedbackText, rating, category]{
+    Q_ASSERT(out);
+    Q_ASSERT(buffer);
+
+    delete out;
+    scheduleTask(new TaskSendFeedback(feedbackText, *buffer, rating, category));
+    delete buffer;
```
bakulf, line 675: Why a task? You could send it directly, I guess. But, ignoring this part, please write a comment saying that the content of the "buffer" is copied by the TaskSendFeedback CTOR.

src/mozillavpn.cpp
```diff
@@ -659,6 +660,25 @@ void MozillaVPN::removeDevice(const QString& deviceName) {
...
+  });
+
+
+  return;
```
bakulf, line 679: remove this :)

src/networkmanager.cpp
```diff
@@ -45,12 +62,7 @@ QByteArray NetworkManager::userAgent() {
 
     // System data
     out << "sys:";
-#ifdef MVPN_WASM
-    out << "WASM";
-#else
-    out << QSysInfo::productType().toLocal8Bit() << " "
-        << QSysInfo::productVersion().toLocal8Bit();
-#endif
+    out << NetworkManager::osVersion();
```
bakulf, line 65: `out << "sys:" << NetworkManager::osVersion();`

src/networkrequest.cpp
```diff
@@ -294,6 +295,32 @@ NetworkRequest* NetworkRequest::createForHeartbeat(QObject* parent) {
   return r;
 }
 
+NetworkRequest* NetworkRequest::createForFeedback(QObject* parent, const QString& feedbackText, const QString& logs, const qint8 rating, const QString& category) {
+  NetworkRequest* r = new NetworkRequest(parent, 201);
+
+  // QUrl url(Constants::API_URL);
+  QUrl url("http://localhost:3000");
```
bakulf, line 302: :) no localhost hehe

src/networkrequest.cpp
```diff
@@ -294,6 +295,32 @@ NetworkRequest* NetworkRequest::createForHeartbeat(QObject* parent) {
...
+  url.setPath("/api/v1/vpn/feedback");
+  r->m_request.setUrl(url);
+
+  r->m_request.setHeader(QNetworkRequest::ContentTypeHeader,
```
bakulf, line 306: What about the authentication header?

src/networkrequest.cpp
```diff
@@ -294,6 +295,32 @@ NetworkRequest* NetworkRequest::createForHeartbeat(QObject* parent) {
...
+                         "application/json");
+
+  QJsonObject obj;
+  obj.insert("feedbackText", QJsonValue(feedbackText));
```
bakulf, line 310: I suspect you can drop the `QJsonValue` here.

src/tasks/sendfeedback/tasksendfeedback.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "models/user.h"
+#include "mozillavpn.h"
+#include "networkrequest.h"
+
+namespace {
+Logger logger(LOG_MAIN, "TaskSendFeedback");
+}
+
+TaskSendFeedback::TaskSendFeedback(const QString& feedbackText, const QString& logs, const qint8 rating, const QString& category)
+    : Task("TaskSendFeedback"), m_feedbackText(feedbackText.left(1000)), m_logs(logs.right(100000)), m_rating(rating), m_category(category) {
```
bakulf, line 18: Can we have:
```
constexpr uint32_t FEEDBACK_MESSAGE_MAX_LENGTH = 1000;
constexpr uint32_t FEEDBACK_LOG_MAX_LENGTH = 100000;
```

src/tasks/sendfeedback/tasksendfeedback.cpp
```diff
@@ -0,0 +1,42 @@
...
+void TaskSendFeedback::run(MozillaVPN* vpn) {
+  logger.log() << "Sending the feedback";
+
+  NetworkRequest* request =
+      NetworkRequest::createForFeedback(this, m_feedbackText, m_logs, m_rating, m_category);
+
+  connect(request, &NetworkRequest::requestFailed,
+          [this, vpn](QNetworkReply::NetworkError error, const QByteArray&) {
+            logger.log() << "Failed to send feedback" << error;
+            vpn->errorHandle(ErrorHandler::toErrorType(error));
```
bakulf, line 33: No error handling here.

bakulf (changes requested):

src/mozillavpn.cpp
mbirghan: I saw that a lot of other neworkrequest are using tasks so I wanted to use the same pattern.

src/tasks/sendfeedback/tasksendfeedback.cpp
mbirghan: The question is what type of error handling do we want to have?

mbirghan: I don't think we should display to the user that there was an error but we could try to resend it later. 
What do you think?

src/mozillavpn.cpp
bakulf: Sounds good to me. Just write a comment about the lifetime of buffer. Something like:
`TaskSendFeedback CTOR copies the content of buffer. We can delete it immediately  after.` Or something better.

src/tasks/sendfeedback/tasksendfeedback.cpp
bakulf: ok! makes sense!

mbirghan:
I added my idea for resending the feedback after it failed.

bakulf:
I rebased this patch on top of the UI work. I close this PR but I keep all the original commits (squashed) and the authorship.

PR #1171, opened 2021-06-10
Same init path for all service start methods

strseb:
.If the Service has a cold start (e.g. app off and user enabled always-on) and is booted with onStartCommand, we did not load wg-go thus crash.


ValentinaPC:
Verified this on the last VPN Android build. Following the steps above, the crash is no longer reproducible.

PR #1178, opened 2021-06-11
Fix the strings for the connection notifications

bakulf:


src/ui/settings/ViewNotifications.qml
```diff
@@ -116,7 +116,7 @@ VPNFlickable {
 
             //% "Connection Change Notification"
             labelText: qsTrId("vpn.settings.notification.connectionChange")
-            //% "Get notified when the Connection Status Changed"
+            //% "Get notified when the connection status changed"
```
flodolo, line 119: Re-reading this: shouldn't it be "changes"?

flodolo (approved):

PR #1189, opened 2021-06-12
Fix SIGABRT on linux during the shutdown - closes #1187

### LINKED ISSUE #1187: SIGABRT when closing from Connection Information screen
@ValentinaPC noted while testing v2.3.1 for #1150 that a crash was observed when terminating the VPN Client from the Connection Information view, and I have been able to reproduce the issue using the UI. A quick look at the stack trace suggests that this is due to heap corruption, and the application exits with a message of `corrupted size vs. prev_size
Aborted (core dumped)`

Steps to reproduce:
1. Log into the VPN and enable the `Launch VPN app on Startup` option
2. Quit the VPN app using the `Quit Mozilla VPN` option from the system tray menu.
3. Launch the VPN client
4. Wait for the connection to start, and then open the connection information view.
5. goto step 2.

Eventually, this will result in an `SIGABRT` when exiting the application. A coredump generated from the shows that the fault occurs from all over the place in Qt, but it always ends in `malloc.c` as follows:

```
Core was generated by `./src/mozillavpn'.
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7fa4f09ae6c0 (LWP 32267))]
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007fa4f15fd859 in __GI_abort () at abort.c:79
#2  0x00007fa4f16683ee in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7fa4f1792285 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007fa4f167047c in malloc_printerr (str=str@entry=0x7fa4f17942a8 "corrupted size vs. prev_size in fastbins") at malloc.c:5347
#4  0x00007fa4f1670c4c in malloc_consolidate (av=av@entry=0x7fa4f17c3b80 <main_arena>) at malloc.c:4493
#5  0x00007fa4f1672160 in _int_free (av=0x7fa4f17c3b80 <main_arena>, p=0x4048f50, have_lock=<optimized out>) at malloc.c:4400
#6  0x00000000012d6ae0 in QV4::IdentifierTable::~IdentifierTable() ()
#7  0x00000000012c7b98 in QV4::ExecutionEngine::~ExecutionEngine() ()
#8  0x00000000012be7de in QJSEngine::~QJSEngine() ()
#9  0x00000000013d1d86 in QQmlEngine::~QQmlEngine() ()
#10 0x00000000004dd836 in QmlEngineHolder::~QmlEngineHolder() (this=0x7ffec8ec8f30) at qmlengineholder.cpp:29
#11 0x0000000000465bed in CommandUI::run(QStringList&)::$_0::operator()() const (this=0x7ffec8ec9270) at commands/commandui.cpp:425
#12 0x0000000000464911 in std::_Function_handler<int (), CommandUI::run(QStringList&)::$_0>::_M_invoke(std::_Any_data const&) (__functor=...)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/std_function.h:285
#13 0x0000000000450a67 in std::function<int ()>::operator()() const (this=0x7ffec8ec9270) at /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/std_function.h:688
#14 0x0000000000451063 in Command::runQmlApp(std::function<int ()>&&) (this=0x377eb90, a_callback=...) at command.cpp:142
#15 0x00000000004646a7 in CommandUI::run(QStringList&) (this=0x377eb90, tokens=...) at commands/commandui.cpp:85
#16 0x0000000000453b24 in CommandLineParser::parse(int, char**) (this=0x7ffec8ec9598, argc=1, argv=0x7ffec8ec96b8) at commandlineparser.cpp:85
#17 0x00000000004a2b0f in main(int, char**) (argc=1, argv=0x7ffec8ec96b8) at main.cpp:15
```

bakulf:
There are two issues here:
- SignalHandler receives signal notifications on random threads. This is fixed
  using QMetaObject::invokeMethod to emit signals.
- QmlEngineHolder must be created before MozillaVPN to avoid memory corruption

### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -123,6 +123,11 @@ int CommandUI::run(QStringList& tokens) {
       }
     }
 
+    // This object _must_ live longer than MozillaVPN to avoid shutdown crashes.
+    QmlEngineHolder engineHolder;
+    QQmlApplicationEngine* engine = QmlEngineHolder::instance()->engine();
+    engine->addImportPath("qrc:///glean");
+
     MozillaVPN vpn;
     vpn.setStartMinimized(minimizedOption.m_set);
 
@@ -151,10 +156,6 @@ int CommandUI::run(QStringList& tokens) {
     // Font loader
     FontLoader::loadFonts();
 
-    // Create the QML engine and expose a few internal objects.
-    QmlEngineHolder engineHolder;
-    QQmlApplicationEngine* engine = QmlEngineHolder::instance()->engine();
-    engine->addImportPath("qrc:///glean");
     vpn.initialize();
 
 #ifdef MVPN_MACOS
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -31,6 +31,7 @@ CREATE_LOG_OP_REF(uint64_t);
 CREATE_LOG_OP_REF(const char*);
 CREATE_LOG_OP_REF(const QString&);
 CREATE_LOG_OP_REF(const QByteArray&);
+CREATE_LOG_OP_REF(void*);
 
 #undef CREATE_LOG_OP_REF
 
--- a/src/logger.h
+++ b/src/logger.h
@@ -56,6 +56,7 @@ class Logger {
     Log& operator<<(const QStringList& t);
     Log& operator<<(const QByteArray& t);
     Log& operator<<(QTextStreamFunction t);
+    Log& operator<<(void* t);
 
    private:
     Logger* m_logger;


src/signalhandler.cpp
```diff
@@ -37,7 +39,11 @@ SignalHandler::SignalHandler() {
 }
 
 void SignalHandler::saHandler(int signal) {
-  logger.log() << "Signal" << signal;
   Q_ASSERT(self);
-  emit self->quitRequested();
+  logger.log() << "Signal" << signal << "from thread"
```
oskirby, line 43: I can't find a good explanation whether QMetaObject::invokeMethod() is async-safe with regards to POSIX signals so I am concerned that there is still a race condition here, and I have also encountered deadlocks in the logger during signal handling. I would recommend an approach closer to what's seen in mozilla-mobile/mozilla-vpn-client/pull/1190/

oskirby:
This seems to have fixed the memory corruption issue.

oskirby (approved):

src/signalhandler.cpp
bakulf: Right! I revert this part and land the rest.

PR #1193, opened 2021-06-12
Add python script to run inspector commands

oskirby:
I found this tool to be handy when trying to reproduce infrequent crashed by automating GUI actions with shell scripting. For example, to click the connection information button: `./scripts/inspector.py click connectionInfoButton`


bakulf (approved):

bakulf:
If you have extra time, we could try to add readline and have a kind-of shell app.

PR #1197, opened 2021-06-14
Experiment: Github CI to build hirsute packages

oskirby:
This is a test to try and get automated Debian packages built for 21.04 hirsute by using `pbuilder` to create clean build environments for distributions which may differ from the base image provided by Github.


oskirby:
The test seems to work! Although, only for distributions which can be built without relying on a PPA for the Qt libraries, so for the meantime this means just Ubuntu 21.04/hirsute.

.gitignore
```diff
@@ -24,6 +24,7 @@ mozillavpn-daemon
 translations/*.qm
 translations/*.pri
 linux/netfilter/netfilter.[ah]
+linux/netfilter/vendor/
```
bakulf, line 27: There are issues with the content of this PR. This is unrelated, right?

bakulf (changes requested):

.gitignore
oskirby: This line in the `.gitignore` file is intended to be there. When we download the Go module's dependencies for offline building using the `go mod vendor`, it will download them all into this directory. We can treat it as generated content, so it shouldn't be committed to the repository.

bakulf (approved):

PR #1207, opened 2021-06-17
✨Experiment✨ Extend custom DNS with AD & Tracking Protection

strseb:
This is based on #902 

With our Code for custom dns in place, we can extend the feature to offer ad-blocking and/or tracking blocking dns servers. 
Which i think would be nice :) 

Unbenannt

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-796)

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -206,14 +206,9 @@ void Controller::activateInternal() {
 
   // Use the Gateway as DNS Server
   // If the user as entered a valid dns, use that instead
-  QHostAddress dns = QHostAddress(server.ipv4Gateway());
-  if (FeatureCustomDNS::instance()->isSupported() &&
-      !settingsHolder->useGatewayDNS() &&
-      settingsHolder->userDNS().size() > 0 &&
-      settingsHolder->validateUserDNS(settingsHolder->userDNS())) {
-    dns = QHostAddress(settingsHolder->userDNS());
-    logger.debug() << "User DNS Set" << dns.toString();
-  }
+  QHostAddress dns =
+      QHostAddress(SettingsHolder::instance()->getDNS(server.ipv4Gateway()));
+  logger.debug() << "DNS Set" << dns.toString();
 
   Q_ASSERT(m_impl);
   m_impl->activate(serverList, device, vpn->keys(),
@@ -282,14 +277,7 @@ bool Controller::silentSwitchServers() {
     serverList.append(hop);
   }
 
-  QHostAddress dns = QHostAddress(server.ipv4Gateway());
-  if (FeatureCustomDNS::instance()->isSupported() &&
-      !settingsHolder->useGatewayDNS() &&
-      settingsHolder->userDNS().size() > 0 &&
-      settingsHolder->validateUserDNS(settingsHolder->userDNS())) {
-    dns = QHostAddress(settingsHolder->userDNS());
-    logger.debug() << "User DNS Set" << dns.toString();
-  }
+  QHostAddress dns = QHostAddress(settingsHolder->getDNS(server.ipv4Gateway()));
 
   Q_ASSERT(m_impl);
   m_impl->activate(serverList, device, vpn->keys(),
@@ -690,11 +678,10 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
     }
   }
   if (shouldExcludeDns()) {
+    auto dns = SettingsHolder::instance()->getDNS(server.ipv4Gateway());
     // Filter out the Custom DNS Server, if the User has one.
-    logger.debug() << "Filtering out the DNS address"
-                   << SettingsHolder::instance()->userDNS();
-    excludeIPv4s.append(
-        IPAddress::create(SettingsHolder::instance()->userDNS()));
+    logger.debug() << "Filtering out the DNS address" << dns;
+    excludeIPv4s.append(IPAddress::create(dns));
   }
 
   QList<IPAddressRange> list;
@@ -742,7 +729,8 @@ bool Controller::shouldExcludeDns() {
   if (!FeatureCustomDNS::instance()->isSupported()) {
     return false;
   }
-  if (settings->useGatewayDNS()) {
+  // Only a Custom DNS might require to be routed outside of the VPN-Tunnel
+  if (settings->dnsProvider() != SettingsHolder::DnsProvider::Custom) {
     return false;
   }
   auto dns = settings->userDNS();
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -18,6 +18,7 @@
 #include "mozillavpn.h"
 #include "rfc/rfc1918.h"
 #include "rfc/rfc4193.h"
+#include "rfc/rfc4291.h"
 #include "rfc/rfc5735.h"
 #include "serveri18n.h"
 #include "settingsholder.h"
@@ -679,7 +680,7 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
   }
   if (shouldExcludeDns()) {
     auto dns = SettingsHolder::instance()->getDNS(server.ipv4Gateway());
-    // Filter out the Custom DNS Server, if the User has one.
+    // Filter out the Custom DNS Server, if the user has set one.
     logger.debug() << "Filtering out the DNS address" << dns;
     excludeIPv4s.append(IPAddress::create(dns));
   }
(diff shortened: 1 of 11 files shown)


src/ui/settings/ViewProtections.qml
```diff
@@ -0,0 +1,192 @@
+
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
bakulf, line 3: Remove these 2 empty lines

src/ui/settings/ViewSettingsMenu.qml
```diff
@@ -132,6 +132,15 @@ VPNFlickable {
             onClicked: settingsStackView.push("../settings/ViewNotifications.qml")
             visible: VPNFeatureList.captivePortalNotificationSupported || VPNFeatureList.unsecuredNetworkNotificationSupported || VPNFeatureList.notificationControlSupported
         }
+        VPNSettingsItem {
```
bakulf, line 135: I don't review the UI because we need to apply the Peter's UX when ready.

bakulf (changes requested):

src/platforms/windows/windowsnetworkwatcher.cpp
```diff
@@ -40,7 +40,7 @@ void WindowsNetworkWatcher::initialize() {
 
   DWORD prefNotifSource;
   if (WlanRegisterNotification(m_wlanHandle, WLAN_NOTIFICATION_SOURCE_MSM,
-                               true /* ignore duplicates */, wlanCallback, this,
+                               true /* ignore duplicates */, (WLAN_NOTIFICATION_CALLBACK) wlanCallback, this,
```
strseb, line 43: This compile error popped out of nowhere. 
msvc was not able to cast wlanCallback automatically?

strseb:
@bakulf , @lesleyjanenorton  - I've updated the pr, and pulled new UI from #1530 into it. Only thing left are the strings.

src/settingsholder.h
```diff
@@ -47,13 +42,26 @@ class SettingsHolder final : public QObject {
...
   ~SettingsHolder();
 
   static SettingsHolder* instance();
 
+  enum DnsProvider {
+    Custom = 4,
+    BlockTracking = 3,
+    BlockAds = 2,
+    BlockAll = 1,
+    Gateway = 0,
```
bakulf, line 61: Can you sort by values? Gateway as the first one and so on?

src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -16,229 +16,100 @@ import telemetry 0.15
 
 
 Item {
+    property string settingId_default: "default"
```
bakulf, line 19: Why do we need this? Can we simply use the Q_ENUM entries?

bakulf (changes requested):
Fix the strings. Now the string PR is merged.

src/ui/settings/ViewAdvancedDNSSettings.qml
strseb: We can't do this, at least if i tried qml complained that we cant use script values as properties of a listmodel.

strseb: Update: we __could__ if we are okay with creating the list model dynamicly

lesleyjanenorton: We can get around this by sending only the string IDs in the ListModel and then `text: VPNl18n.tr(VPNl18n[tabLabelStringId])` to the button delegate in `VPNTabNavigation { }`. But this is working too!

src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -16,229 +16,97 @@ import telemetry 0.15
...
+
+    }
+
     VPNMenu {
         id: menu
         objectName: "settingsAdvancedDNSSettingsBackButton"
 
-        //% "Advanced DNS settings"
-        title: qsTrId("vpn.settings.advancedDNSSettings.title2")
+        title: "DNS Settings"
```
lesleyjanenorton, line 47: Needs a localized string.

lesleyjanenorton (approved):
Looks great. Left a comment with a workaround for the strings but not sure it's worth the refactor since this is working.

src/controller.cpp
```diff
@@ -685,11 +670,10 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
...
   }
   if (shouldExcludeDns()) {
+    auto dns = SettingsHolder::instance()->getDNS(server.ipv4Gateway());
     // Filter out the Custom DNS Server, if the User has one.
-    logger.debug() << "Filtering out the DNS address"
-                   << SettingsHolder::instance()->userDNS();
-    excludeIPv4s.append(
-        IPAddress::create(SettingsHolder::instance()->userDNS()));
+    logger.debug() << "Filtering out the DNS address" << dns;
+    excludeIPv4s.append(IPAddress::create(dns));
```
bakulf, line 676: Are we actually sure we want to do this? Can we do the DNS lookup via VPN?

src/controller.cpp
```diff
@@ -737,7 +721,8 @@ bool Controller::shouldExcludeDns() {
   if (!FeatureList::instance()->userDNSSupported()) {
     return false;
   }
-  if (settings->useGatewayDNS()) {
+  // Only a Custom DNS might require to be routed outside of the VPN-Tunnel
+  if (settings->dnsProvider() != SettingsHolder::DnsProvider::Custom) {
```
bakulf, line 725: Is it really needed? Can we do this only for LAN servers?

src/ui/components/VPNViewDNSSettings.qml
```diff
@@ -0,0 +1,223 @@
...
+          case settingId_adblock.toString():
+            return VPNSettings.dnsProvider === VPNSettings.BlockAds;  
+          case settingId_antiTracking.toString():
+            return VPNSettings.dnsProvider === VPNSettings.BlockTracking;  
+          case settingId_adblockAndAntiTracking.toString():
+            return VPNSettings.dnsProvider === VPNSettings.BlockAll;  
+          case settingId_local.toString():
+            return VPNSettings.dnsProvider === VPNSettings.Custom;  
+          default:
+            console.log("TODO")
```
bakulf, line 38: TODO what? :) We should log this and return ... false?

src/ui/components/VPNViewDNSSettings.qml
```diff
@@ -0,0 +1,223 @@
...
+          case settingId_adblock.toString():
+            return VPNSettings.dnsProvider = VPNSettings.BlockAds;  
+          case settingId_antiTracking.toString():
+            return VPNSettings.dnsProvider = VPNSettings.BlockTracking;  
+          case settingId_adblockAndAntiTracking.toString():
+            return VPNSettings.dnsProvider = VPNSettings.BlockAll;  
+          case settingId_local.toString():
+            return VPNSettings.dnsProvider = VPNSettings.Custom;  
+          default:
+            console.log("TODO")
```
bakulf, line 59: No TODO, log it and set it to Gateway. What do you think?

src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -16,229 +16,97 @@ import telemetry 0.15
 
 
 Item {
+    property string settingId_default: "default"
+    property string settingId_local: "local"
+    property string settingId_adblock: "adblock"
+    property string settingId_antiTracking: "antiTracking"
+    property string settingId_adblockAndAntiTracking: "adblockAndAntiTracking"
```
bakulf, line 23: So we keep them. Ok!

src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -16,229 +16,97 @@ import telemetry 0.15
...
-
-                        VPNTextBlock {
-                           //% "Resolve website domain names using a DNS in your local network"
-                           text: qsTrId("vpn.advancedDNSSettings.localDNS.resolveWebsiteDomainNames")
-                           width: parent.width
-                        }
-                    }
+                Component.onCompleted: {
+                    defaultTabListModel.append({
+                                                "settingId": "default",
```
bakulf, line 77: Can this be set to VPNSettings.Gateway ?

bakulf (approved):

bakulf (approved):

PR #1214, opened 2021-06-18
Use system Qt for Ubuntu Focal builds

oskirby:
Making use of the system Qt libraries allows us to install the VPN without needing to install a PPA (just the .deb file alone is sufficient), and would also allow us to support Debian releases too.


bakulf:
This is similar to commit db2029b16da0bdc5d7e4bfbc67202e957d1ab22c .when I wrote that PR, I had to check each view to see if all works fine. Have you done a similar check? Because we cannot rely on our tests yet, this needs to be done manually.

oskirby:
I can run the functional tests locally without issue, it only seems to be on the CI where they are broken. Though I only verified that the tests run to completion and not that the views are free of bugs.

bakulf (approved):

PR #1219, opened 2021-06-20
Make ping workers async and remove threading

oskirby:
To try and make the connection health checker a bit more resilient to packet loss, we should convert the Windows ping sender to use asynchronous calls when sending pings. This allows the ping worker thread to continue sending more pings during the timeout on the lost ping. This prevents `sendPing()` calls from backing up and triggering unnecessary server switching when they all arrive late.

Since the windows implementation was the only one that blocked anyways, it's no longer necessary to have a worker thread at all after converting them to use asynchronous operations.

### DIFF
--- a/src/networkrequest.cpp
+++ b/src/networkrequest.cpp
@@ -360,7 +360,7 @@ void NetworkRequest::replyFinished() {
   QByteArray data = m_reply->readAll();
 
   if (m_reply->error() != QNetworkReply::NoError) {
-    logger.log() << "Network error:" << m_reply->error()
+    logger.log() << "Network error:" << m_reply->errorString()
                  << "status code:" << status << "- body:" << data;
     emit requestFailed(m_reply->error(), data);
     return;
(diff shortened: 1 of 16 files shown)


bakulf:
I like this but I'm not sure we can land this code yet. There are still a few "blocking" operations on mac and Linux related to sendto(). By documentation:
```When  the  message  does  not  fit  into the send buffer of the socket, send() normally blocks, unless the socket has been placed in nonblocking I/O mode.  In nonblocking mode it would fail with the error EAGAIN or
       EWOULDBLOCK in this case.  The select(2) call may be used to determine when it is possible to send more data.```

I wonder if we can be 100% sure our buffer is < the socket buffer each time. Or if we should use a non-blocking socket, or (easier) use an async QT component to call sendto. What do you think?

oskirby:
For the Linux and MacOS senders, I don't think it's an issue since we are using raw sockets for the ICMP ping transmission. Even if it were to block due to insufficient buffer space, it should only be blocked until the network interface can actually transmit enough data to clear the buffer and even that is unlikely since we are only ever sending a single ping on this socket anyways.

The socket blocking behaviour is really only a problem for TCP, where the connection relies on a remote host to drain the socket buffer.

It would be possible to put the socket into non-blocking mode and rely on QSocketNotifer to handle the transmission too, but I think that is overkill.

bakulf:
Good! Thanks for answering.

PR #1222, opened 2021-06-22
Use the public key to identify the device instead of the name - closes #1013

### LINKED ISSUE #1013: Twin-Devices Are Handled Like One Device
"I have registered two LGE LM-G900 smartphones for Mozilla VPN. The app was successfully installed on both devices. However, it seems that the devices in the login directory are not correctly recognised and differentiated. On the smartphones only 3 of 5 devices are displayed, but there should be four.
I can never activate the two smartphones at the same time. "

bakulf:


oskirby (approved):
I think the change looks good to resolve the linked issue and should allow multiple devices to exist with the same name.

However, if a user does have two devices with the same name, and they need to remove one of them, it would be nice if there was a way to tell them apart before removal (for example, display the public key on the popup).

bakulf:
Good point! I'll ask @jess-cook03 and @peterbenvenuto to follow your suggestion.

PR #1230, opened 2021-06-23
Revert "Use system Qt for Ubuntu Focal builds (#1214)"

oskirby:
After further testing, it seems that there is still some breakage when building with Qt 5.12 for Ubuntu/Focal. It would be better to revert back to requiring Qt 5.14 for the main branch until the issues can be resolved. The issues identified on Qt 5.12 include:
 - QML Connections class has a different syntax for passing arguments to slots.
 - Memory corruption during teardown remains unresolved.
 - In-App Feedback form appears to be broken.

This reverts commit 4b644a699c04e2ae310d7850e4fda7b1e6129767.


bakulf (approved):
I didn't check the PR entirely. I assume it has been generated reverting the previous commit.

oskirby:
That's correct, this was PR created using the `git revert` command, I haven't touched any files by hand.

PR #1231, opened 2021-06-24
Use packet loss and RTT to drive silent server switching

oskirby:
Windows builds have been somewhat unstable in the face of packet loss, where the loss of a single ICMP ping can trigger a server switch. In an attempt to make it behave a little better, we refactor the `PingHelper` class to gather statistics and use packet loss and RTT measurements to determine when a server switch is necessary.

A connection is considered to be unstable when packet loss to the VPN server exceeds 10%, or if any particular ping arrives more than 1000ms late.


src/pinghelper.cpp
```diff
@@ -31,53 +50,141 @@ void PingHelper::start(const QString& serverIpv4Gateway,
...
+  logger.log() << "Sending ping seq:" << m_sequence;
+#endif
+
+  int index = m_sequence % PING_STATS_WINDOW;
+  m_pingData[index].timestamp = QDateTime::currentMSecsSinceEpoch();
+  m_pingData[index].latency = -1;
+  m_pingData[index].sequence = m_sequence;
+  m_pingSender->sendPing(m_gateway, m_sequence);
+
+  m_sequence++;
```
bakulf, line 98: 1 line of comment saying that we do not care about the overflow of this variable because we use it as an index + a counter.

src/pinghelper.cpp
```diff
@@ -31,53 +50,141 @@ void PingHelper::start(const QString& serverIpv4Gateway,
...
+  }
 
-  while (m_pings.length() > PINGS_MAX) {
-    m_pings.at(0)->deleteLater();
-    m_pings.removeAt(0);
+  if (recvCount <= 0) {
+    return 0.0;
   }
+
+  totalMsec += recvCount / 2;
```
bakulf, line 132: I don't understand this part. I thought you want to compute the average which should be: `totalMsec / recvCount`. Why do you add this extra `recvCount /2`.
Maybe write a comment

src/platforms/dummy/dummypingsender.cpp
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
bakulf, line 1: It would be better to use a `git mv` to keep the history.

bakulf (approved):

src/platforms/dummy/dummypingsender.cpp
oskirby: I did, but I think with the name change from `xxxSendWorker` to `xxxSender` meant that too many lines had changed for git to be able to identify them as the same file.

src/pinghelper.cpp
oskirby: This is a math trick that produced the nearest-integer rounding instead of the floor when doing an integer divide.

PR #1232, opened 2021-06-24
Display server latency on connection info view

oskirby:
I threw this together while debugging PR #1231. Maybe it's something worth adding permanently to the Connection Info screen.

Screenshot from 2021-06-23 10-50-01

### DIFF
--- a/src/ui/components/VPNConnectionInfo.qml
+++ b/src/ui/components/VPNConnectionInfo.qml
@@ -12,6 +12,7 @@ import "../themes/themes.js" as Theme
 Item {
     property var rBytes: VPNConnectionData.rxBytes
     property var tBytes: VPNConnectionData.txBytes
+    property var latencyMSec: VPNConnectionHealth.latency
 
     id: connectionInfo
     anchors.fill: box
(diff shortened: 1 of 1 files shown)


bakulf:
This is a cool idea! Just a couple of comments:
- let's inform the UX folks about it
- let's check with @flodolo if we could have string issues. I wonder if there is enough space to have `ping download upload` on the same line in all the languages.

flodolo:
>     * let's check with @flodolo if we could have string issues. I wonder if there is enough space to have `ping download upload` on the same line in all the languages.

It feels quite risky. For example, the German translation for Ping is apparently _Datensatz_. If you start getting very long translations for all 3 of them, it might break.

Having them on 3 lines should take the same vertical space, but there would be plenty of horizontal space for longer translations. Alternatively, having ping on one line, and download/upload on another (with the current layout, just 2 columns)

oskirby:
Hmm yes, this will probably break badly if the translation for "Ping" is too long. Probably best to reject this PR and let the UX team come up with a better solution.

bakulf:
Can we split the UI/UX part and land all the rest?

oskirby:
The functionality for this has already landed in PR #1231, the only thing that remains un-merged are the QML changes.

birdsarah:
Closing as stale.

PR #1233, opened 2021-06-24
Merge translations when running importLanguages.py

oskirby:
Using the lconvert command to merge the TS and XLIFF files means that we can now get the engineering English strings to show up immediately in the UI without needing to wait for an automated CI run on the i18n submodule.

### DIFF
--- a/scripts/importLanguages.py
+++ b/scripts/importLanguages.py
@@ -103,8 +103,8 @@ print('Updated translations.pri')
 os.system(f"{lupdate} src/src.pro")
 
 # Step 4
-# Now import translations into the files
+# Now merge translations into the files
 for l10n_file in l10n_files:
-    os.system(f"{lconvert} -if xlf -i {l10n_file['xliff']} -o {l10n_file['ts']}")
+    os.system(f"{lconvert} -i {l10n_file['ts']} -if xlf -i {l10n_file['xliff']} -o {l10n_file['ts']}")
 
 print(f'Imported {len(l10n_files)} locales')


scripts/importLanguages.py
```diff
@@ -103,8 +103,8 @@
 os.system(f"{lupdate} src/src.pro")
 
 # Step 4
-# Now import translations into the files
+# Now merge translations into the files
 for l10n_file in l10n_files:
-    os.system(f"{lconvert} -if xlf -i {l10n_file['xliff']} -o {l10n_file['ts']}")
+    os.system(f"{lconvert} -i {l10n_file['ts']} -if xlf -i {l10n_file['xliff']} -o {l10n_file['ts']}")
```
flodolo, line 108: Could you explain what's the benefit of this change? I think I might be missing something.

The localized .ts files shouldn't exist before this command creates them from XLIFF, since we .gitignore them.

oskirby: The .ts files are created a few lines above using the `lupdate` command, but they will initially contain only the un-localized content from the C++ and QML sources. Then at line 108 we then use `lconvert` to produce the localized .ts file from the XLIFF.

However, when adding strings to the UI, they won't be present in the XLIFF, meaning that `lconvert` will overwrite the .ts file with a new one that is missing the new string. The result is that the initial strings provided by the `//%` comments get lost until such time as the XLIFF can be updated.

The fix presented here is to provide both the un-localized .ts file as well as the XLIFF as inputs to `lconvert` when generating the localized .ts that preserves the new strings found by `lupdate`

flodolo: > The .ts files are created a few lines above using the `lupdate` command, but they will initially contain only the un-localized content from the C++ and QML sources.

Ah, didn't realize it would create the .ts for all locales at that point.

The fix makes sense, I'm a bit surprised that qt doesn't fall back to English automatically if the .ts file doesn't have some of the strings.

flodolo (approved):

bakulf (approved):

PR #1234, opened 2021-06-24
Increase the wasm memory buffer

bakulf:


src/src.pro
```diff
@@ -788,6 +788,8 @@ else:wasm {
 
     CONFIG += c++1z
 
+    QMAKE_WASM_TOTAL_MEMORY=18087936
```
oskirby, line 791: That seems like an unusually specific number, where did it come from? If it's something derived from the run-time behaviour of the code then we probably want to round it up to make room for changes in the code.

bakulf: Better number + comments: 32mb. This is definitely too much, but for github actions should be good enough.

oskirby (approved):

PR #1242, opened 2021-06-25
Strings for the DNS settings

bakulf:


src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -0,0 +1,146 @@
...
+
+                    // Now bother user if the ip is invalid :)
+                    case VPNSettings.UserDNSInvalid:
+                        //% "Invalid IP address"
+                        ipInput.error = qsTrId("vpn.settings.userDNS.invalid")
+                        ipInput.valueInvalid = true;
+                        break;
+
+                    case VPNSettings.UserDNSNotIPv4:
+                        //% "We currently support only ipv4 IP addresses"
```
flodolo, line 122: As commented elsewhere, this should say `… IPv4 IP addresses`

src/ui/settings/ViewAdvancedDNSSettings.qml
```diff
@@ -0,0 +1,146 @@
...
+
+            // TODO: this should be a radiobox
+            VPNCheckBoxRow {
+                id: useLocalDNS
+                objectName: "settingMozillaDNSEnabled"
+                width: parent.width - Theme.windowMargin
+
+                //% "Use local DNS"
+                labelText: qsTrId("vpn.advancedDNSSettings.localDNS")
+                //% "Resolve websites using a DNS in your local network"
```
flodolo, line 78: I feel like "Resolve websites" is imprecise enough to trigger someone who knows what a DNS does, and to not help someone who doesn't.

lesleyjanenorton: @chrisrooneys what are your thoughts here?

chrisrooneys: @flodolo @lesleyjanenorton is "Access" technically accurate enough?

flodolo: It isn't. The accurate version would be `Resolve domain names using a DNS…`, or `Resolve website domain names using a DNS…` if we want to keep mentionining "websites".

chrisrooneys: Between those I think "Resolve website domain names using a DNS" covers both groups of users better.

lesleyjanenorton: thank you @flodolo & @chrisrooneys !

lesleyjanenorton (approved):

PR #1245, opened 2021-06-25
Generic QML model to filter other models

bakulf:


oskirby (approved):
It took me a while to understand what this is doing, but I think this is good.

bakulf:
Yeah, it's a callback-based filter model.

bakulf:
I don't land this PR because I want to know what @lesleyjanenorton thinks about it. Plus, the language view doesn't implement the final UI yet. If @lesleyjanenorton is OK with this code, I can land the C++ piece only.

lesleyjanenorton (approved):

lesleyjanenorton:
Thank you!! Working perfectly and will be so helpful for the new split tunneling search UI too. 

I've added some flourishes:

- VPNSearchBar.qml (This would benefit from some placeholder text but I didn't want to add anything since we're in string freeze)
- Updates tabbing to include the search bar
- Small refactors to vertically re-center the selected language when the view is opened

--
Better centering when the view is opened...
Screen Shot 2021-07-07 at 4 23 27 PM


Search bar without focus...
Screen Shot 2021-07-07 at 4 23 35 PM

Search bar searching...
Screen Shot 2021-07-07 at 3 03 26 PM


Search bar can't find anything...
Screen Shot 2021-07-07 at 3 03 33 PM

PR #1247, opened 2021-06-25
More testing fixes

oskirby:
This is kind of a grab bag of things that have bugged me while trying to diagnose issues in the functional tests, changes include:
 - Close selenium webdriver after authentication.
 - Expose QML/C++ singletons via inspector.
 - Add currentView property to the VPN singleton.
 - Fix potential race during authentication/reset_devices.
 - Upload logs of failed tests as artifacts.

### DIFF
--- a/src/inspector/inspectorwebsocketconnection.cpp
+++ b/src/inspector/inspectorwebsocketconnection.cpp
@@ -42,7 +42,7 @@ QString s_updateVersion;
 
 }  // namespace
 
-static QQuickItem* findObject(const QString& name) {
+static QObject* findObject(const QString& name) {
   QStringList parts = name.split("/");
   Q_ASSERT(!parts.isEmpty());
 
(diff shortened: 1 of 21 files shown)


bakulf (approved):

src/inspector/inspectorwebsocketconnection.cpp
```diff
@@ -512,6 +521,13 @@ static QList<WebSocketCommand> s_commands{
                        return obj;
                      }},
 
+    WebSocketCommand{"translate", "Translate a string", 1,
```
bakulf, line 524: This is not used, yet. Right? At least in the tests.

bakulf (approved):

src/inspector/inspectorwebsocketconnection.cpp
oskirby: No, there's nothing that uses it (yet).

I was finding that the tests would fail if my locale was set to anything other than `en`, so I had the idea that maybe we could compare a results against a translated string. But it turned out to be easier to just set `LANGUAGE=en` at the start of the test scripts.

PR #1249, opened 2021-06-26
RPM packaging in github CI

oskirby:
This is a test to see if we can build functioning RPM packages from a github workflow.


oskirby:
This seems to produce working packages for Fedora 33 and newer. In theory the .RPM files produced by this process should run on any distro that comes with Qt 5.15 or newer.

bakulf (approved):

Issue #1252, opened 2021-06-27
VPN fails to connect on Debian 10 (unable to create ping socket)


matthewhughes934:
Note: Due to #1192 I'm on Debian 10 running:

```
$ uname --kernel-release 
5.10.0-0.bpo.7-amd64
```

Instead of the default `4.19`

With `linuxdaemon` running as root, I run `mozillavpn activate` which just seems to hang for sometime, the logs at `$HOME/.local/share/mozillavpn.txt` give:

```
[27.06.2021 14:18:00.396] (main - Command) MozillaVPN 2.4.0
[27.06.2021 14:18:00.396] (main - Command) User-Agent: MozillaVPN/2.4.0 (sys:debian 10)
[27.06.2021 14:18:00.397] (main - CryptoSettings) Read the settings file
[27.06.2021 14:18:00.399] (main - SettingsHolder) Creating SettingsHolder instance
[27.06.2021 14:18:00.399] (main - Localizer) Loading language:
[27.06.2021 14:18:00.399] (main - SettingsHolder) Setting languageCode
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Product Bugs/Errors
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Network Connection/Speed
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Product Quality
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Access to service
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Compatibility
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Ease of Use
[27.06.2021 14:18:00.400] (model - FeedbackCategoryModel) Adding: Other
[27.06.2021 14:18:00.400] (main - MozillaVPN) Creating MozillaVPN singleton
[27.06.2021 14:18:00.401] (model - DeviceModel) Reading the device list from settings
[27.06.2021 14:18:00.401] (model - ServerCountryModel) Reading the server list from settings
[27.06.2021 14:18:00.410] (model - ServerData) fr France Paris
[27.06.2021 14:18:00.410] (main - MozillaVPN) Checking model initialization
[27.06.2021 14:18:00.410] (controller - Controller) Initializing the controller
[27.06.2021 14:18:00.412] (linux - DBusClient) Status via DBus
[...]

Issue #1250, opened 2021-06-27
scripts/qt5_compile.sh fails on Debian 10


matthewhughes934:
Following the instructions for building on Debian in the README I reached:

```
bash scripts/qt5_compile.sh qt qt
```

Which failed with a message:

```
WARNING: -debug-and-release is only supported on Darwin and Windows platforms.
Qt can be built in release mode with separate debug information, so
-debug-and-release is no longer necessary.

ERROR: Feature 'debug_and_release' was enabled, but the pre-condition 'config.darwin || config.win32' failed.

Check config.log for details.
Configuration error.
```

Removing `-debug-and-release` from the `./configure` call in `scripts/qt5_compile.sh` allowed Qt to build successfully.

matthewhughes934:
This should be fixed with https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1256

PR #1256, opened 2021-07-04
Misc improvements from linux build

matthewhughes934:
Some miscellaneous improvements for the build process based on my experience building on Debian which I hope other's might find helpful: 

- Log strerror linuxpingsender socket create failure

    This helped me diagnose #1252 so others might find it useful.

- Update .gitignore rules for `glean`

    With a96253e44482f11745f1d2ddd1ee35d1e90c5d63 scripts/generate_glean.py
    was updated to write under 'glean/telemetry' rather than
    'glean/generated' `.gitignore` was updated then to reflect this, but
    then was updated again with e27252cf59ea28bb16fbfcca78b6ba53abac5fd9.
    Since it appears `glean` is indeed writing under `telemetry`, restore
    this previous ignore behaviour

- Add .gitignore rule for QT compilation

    One of the suggested methods for installing QT5 in the README is to
    compile it under `qt/`, so ignore this as a convenience for anyone
    choosing to compile like this.

- Remove unspported QT config flag for Linux builds

    '-debug-and-release' is not supported on Linux (see issue #1250), so
    move it to be for MacOS builds only

### DIFF
--- a/scripts/qt5_compile.sh
+++ b/scripts/qt5_compile.sh
@@ -67,6 +67,7 @@ LINUX="
 "
 
 MACOS="
+  -debug-and-release \
   -appstore-compliant \
   -no-feature-qdbus \
   -no-speechd
@@ -89,7 +90,6 @@ print Y "Wait..."
   --recheck-all \
   -opensource \
   -confirm-license \
-  -debug-and-release \
   -static \
   -strip \
   -silent \
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ mozillavpnnp.vcxproj*
 # glean
 /glean/telemetry/*
 !/glean/telemetry/.keepme
+
+# For users compiling QT per the README
+/qt/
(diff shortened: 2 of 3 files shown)


scripts/qt5_compile.sh
```diff
@@ -67,6 +67,7 @@ LINUX="
 "
 
 MACOS="
+  -debug-and-release \
```
matthewhughes934, line 70: Is this flag necessary? From what I understand this will trigger two builds of QT (with/without extra debug configurations), this might be a bit of a shock for users just wanting a build of the VPN. Though perhaps it's targeted for internal builds.

bakulf: I prefer to keep `-release` only. Or have an extra param to pick debug/release/both.

bakulf (approved):

bakulf:
Thank you for this PR!

PR #1259, opened 2021-07-05
Expose the montly payment and the savings

bakulf:
This path does also the following:
- compute the price (total and monthly)
- one single registration operation of all the products

### DIFF
--- a/src/platforms/ios/iaphandler.h
+++ b/src/platforms/ios/iaphandler.h
@@ -21,6 +21,7 @@ class IAPHandler final : public QAbstractListModel {
     ProductYearly,
     ProductUnknown = -1
   };
+  Q_ENUM(ProductType);
 
   static IAPHandler* createInstance();
 
@@ -29,8 +30,10 @@ class IAPHandler final : public QAbstractListModel {
   enum ModelRoles {
     ProductIdentifierRole = Qt::UserRole + 1,
     ProductPriceRole,
+    ProductMonthlyPriceRole,
     ProductTypeRole,
     ProductFeaturedRole,
+    ProductSavingsRole,
   };
   Q_INVOKABLE void subscribe(const QString& productIdentifier);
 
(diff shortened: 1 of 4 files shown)


src/platforms/ios/iaphandler.mm
```diff
@@ -337,59 +356,96 @@ - (void)request:(SKRequest*)request didFailWithError:(NSError*)error {
...
-    m_productsRegistrationState = eRegistered;
-    sortProductsAndCompleteRegistration();
+    [numberFormatter release];
   }
+
+  logger.log() << "Monthly Price:" << monthlyPriceValue;
+
+  productData->m_price = priceValue;
+  productData->m_monthlyPrice = monthlyPriceValue;
+  productData->m_nonLocalizedMonthlyPrice = [monthlyPriceNS doubleValue];
```
mbirghan, line 433: Isn't `monthlyPriceNS` localized? Why do we call the variable `m_nonLocalized...`. 
Reading in `SKProduct` the `price` property is localized right?

src/platforms/ios/iaphandler.h
```diff
@@ -91,16 +97,20 @@ class IAPHandler final : public QAbstractListModel {
   struct Product {
     QString m_name;
     QString m_price;
+    QString m_monthlyPrice;
+    // This is not exposed and it's not localized. It's used to compute the
+    // saving %.
+    double m_nonLocalizedMonthlyPrice = 0;
```
mbirghan, line 103: I would not store monetary values in `double` variables. I would store this as cent amount as an `unsigned int`. Even though we only use this for calculation for now we would then have an accurate price representation which might be handy.

mbirghan (approved):
Looks good! 
Just two little things I noticed

src/platforms/ios/iaphandler.mm
bakulf: `SKProduct` price is an `NSDecimalNumber`. And it's not localized. To localize it we need to run the following code:

```
  NSNumberFormatter* numberFormatter = [[NSNumberFormatter alloc] init];
  [numberFormatter setFormatterBehavior:NSNumberFormatterBehavior10_4];
  [numberFormatter setNumberStyle:(NSNumberFormatterStyle)NSNumberFormatterCurrencyStyle];
  [numberFormatter setLocale:product.priceLocale];

  NSString* price = [numberFormatter stringFromNumber:product.price];
```

src/platforms/ios/iaphandler.h
bakulf: it must be double because often the price is 4.99. unsigned int is not enough. Note that localized monthly price is used only to compute the saving %.

Issue #1165, opened 2021-06-09
Connection speed drastically reduced while connected to VPN


jr-johnson-support:
PREREQUISITES

```
The user is signed into the VPN client, but not connected to a server.
```

STEPS TO REPRODUCE

```
Navigate to an online connection speed-test tool (e.g., fast.com, speedtest.net), and run a test.
Launch the VPN client, and connect to a server.
Re-run the speed test from step 1.
```

OBSERVED RESULTS

There is a significant drop in the user's connection speed after connecting to a VPN server, as shown in the attached screenshots. This drop is consistent with results using the official WireGuard iOS app, a screenshot for which is also attached.
EXPECTED RESULTS

Users will experience a minimal drop in connection speed after connecting to a VPN server.

```
Options
```

[PGAND-193.zip](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6624160/PGAND-193.zip)

This is a legacy issue: https://jira.mozilla.com/browse/PGAND-193
Support is still getting incoming tickets around this contact reason: 138 tickets

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-40)
┆author: Data Sync

bakulf:
@mbirghan can we try to see if #1000 would help here too?

mbirghan:
Sure we should be able to use most of it.
But to know when to switch we need to create a baseline to determine when the speed drop is "significant".

@baku do you have an idea how we would measure the baseline?

oskirby:
Packet loss and increased variability in round trip times are usually a good indication of a when a server is struggling to keep up with its traffic load. We might be able to gather these statistics from the ping workers to make a judgment of when a server ought to be replaced.

rbillings:
We'll get more info after v2.4 with the logging to better identify the issue

jr-johnson-support:
144 tickets inbounded over the last 6 months for this issue.

data-sync-user:
➤ Rebecca Billings commented:

Stale

data-sync-user:
➤ Santiago Andrigo commented:

Moving back to Cancelled

data-sync-user:
➤ Santiago Andrigo commented:

Moving back to Cancelled

PR #1261, opened 2021-07-06
Add jitter as metric for Unstable connections

mbirghan:
This PR addresses some of the ideas in #1165. 
It introduces a new metric jitter to detect unstable connections and initiate a `silentServerSwitch`.


mbirghan:
Oh I didn't rebase so just saw Owens patch. 

So this might be obsolete.

mbirghan:
For some reason there are files in the commit that I did not touch. Also the commit on here is a different one that I have on my machine. I am going to create a fresh branch.

PR #1264, opened 2021-07-06
Add jitter as a metric for unstable connections

mbirghan:
This PR addresses some of the ideas in #1165.
It introduces a new metric jitter to detect unstable connections and initiate a `silentServerSwitch`.


oskirby (commented):
The implementation of this looks fine, I just worry a little bit that 30ms of jitter might be too strict of a metric and it could lead to false positives, especially if the user is on a congested wireless network.

Has this been tested under a broad range of network conditions? I'm especially concerned about how it would perform on a cellular network.

mbirghan:
I did some tests and 30ms seemed to be a reasonable threshold. 
However I could not test in all conditions. 
A different idea would be to instead of a fixed amount use percentage of the current ping. 
@oskirby what do you think?

PR #1270, opened 2021-07-08
Use the start time of the connection instead of a timer to show time

### LINKED ISSUE #1263: VPN timer does not display the correct uptime
#### Platforms:
- Android 9, iOS;

#### Prerequisites:
- The user has a VPN account with an active subscription;
- The VPN app is installed;

#### Steps to reproduce:
1. Open the VPN app;
2. Turn on the VPN to any location;
3. Close the phone’s display and take note at what time you do this;
4. Wait for 5-10 minutes;
5. Open your phone and reopen the VPN app;
6. Check the VPN timer;

#### Expected result:
- The timer should display the correct time passed;

#### Actual result:
- The timer shows that less time has passed;

#### Notes:
- Might need a few tries as it doesn’t always reproduce;
- Kill swiping the app and reopening it fixes the timer;

mbirghan:
Closes #1263

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -135,9 +135,7 @@ void Controller::implInitialized(bool status, bool a_connected,
   // If we are connected already at startup time, we can trigger the connection
   // sequence of tasks.
   if (a_connected) {
-    resetConnectionTimer();
-    m_connectionTimerExtraSecs =
-        connectionDate.secsTo(QDateTime::currentDateTime());
+    m_connectedTimeInUTC = connectionDate.toUTC();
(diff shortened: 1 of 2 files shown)


bakulf:
This would not work because if you change the timezone, or change the time on your device, the timer/date will be broken.
Note that what you are proposing here was our first implementation. Then we replace it with the elapsed timer.

mbirghan:
I changed the time saved to UTC and only work with UTC values. This should fix the problem of changing timezones.

bakulf (approved):

PR #1274, opened 2021-07-08
Update wireguard-apple submodule

mbirghan:
Just noticed that we did not update the submodule when upgrading to the new wireguard-apple version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ The procedure to compile MozillaVPN for macOS is the following:
 3. Update the submodules:
 ```
   $ git submodule init
-  $ git submodule update --remote
+  $ git submodule update
 ```
 4. Install python3 dependencies:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,7 +191,6 @@ The IOS procedure is similar to the macOS one:
 2. Update the submodules:
 ```
   $ git submodule init
-  $ git submodule update --remote
```
birdsarah, line 194: This should be `git submodule update` (just missing the --remote). By running `git submodule update` after a pull a dev will bring their version up to that of the repo's.

birdsarah (changes requested):
All references to update `--remote` not just this one. It's bad habit for devs to all be working off different commits of submodules so we should fix everywhere.

birdsarah (approved):

Issue #1275, opened 2021-07-08
D-Bus service hangs at system shutdown on Ubuntu/Groovy


oskirby:
**VPN Version:**
- 2.3.2 (2.202107081305)

**Affected Platforms:**
- Ubuntu 20.10/Groovy

**Steps to reproduce:**
1. Install the VPN client on Ubuntu 20.10/Groovy
2. Launch and then exit the VPN client to trigger the backend service.
3. Simulate a shutdown by stopping the backend service with `systemctl stop mozillavpn`

**Expected result:**
1. The `systemctl stop mozillavpn` command should exit within a second or two, indicating that the service has terminated.

**Actual result:**
1. The `systemctl stop mozillavpn` command hangs for about 90 seconds until `systemd` gives up and issues a `SIGKILL`

We can observe that the D-Bus service did not respond to `SIGTERM` and needed to be killed more aggressively by examining the `systemd` journal:
```
osk@ubuntu-groovy:~/mozilla-vpn-client$ journalctl -efu mozillavpn
Jul 08 13:59:51 ubuntu-groovy systemd[1]: Starting MozillaVPN D-Bus service...
Jul 08 13:59:51 ubuntu-groovy systemd[1]: Started MozillaVPN D-Bus service.
Jul 08 14:00:09 ubuntu-groovy systemd[1]: Stopping MozillaVPN D-Bus service...
Jul 08 14:01:39 ubuntu-groovy systemd[1]: mozillavpn.service: State 'stop-sigterm' timed out. Killing.
Jul 08 14:01:39 ubuntu-groovy systemd[1]: mozillavpn.service: Killing process 2011 (mozillavpn) with signal SIGKILL.
Jul 08 14:01:39 ubuntu-groovy systemd[1]: mozillavpn.service: Killing process 2012 (QDBusConnection) with signal SIGKILL.
Jul 08 14:01:39 ubuntu-groovy systemd[1]: mozillavpn.service: Main process exited, code=killed, status=9/KILL
Jul 08 14:01:39 ubuntu-groovy systemd[1]: mozillavpn.service: Failed with result 'timeout'.
Jul 08 14:01:39 ubuntu-groovy systemd[1]: Stopped MozillaVPN D-Bus service.
```

PR #1276, opened 2021-07-08
Fix broken SignalHandler slot connection on Qt<5.15

oskirby:
The API to the `QSocketNotifier` changed between Qt 5.14 and 5.15, which caused the `SignalHandler` class to stop receiving signals. We can work around the problem by using functions pointer instead of trying to marshal the signal with the `SIGNAL()` macro.

Closes: #1275

### DIFF
--- a/src/signalhandler.cpp
+++ b/src/signalhandler.cpp
@@ -6,6 +6,7 @@
 #include "logger.h"
 #include "signal.h"
 
+#include <fcntl.h>
 #include <unistd.h>
 
 namespace {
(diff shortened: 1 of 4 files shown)


src/signalhandler.cpp
```diff
@@ -59,7 +60,7 @@ SignalHandler::~SignalHandler() {
 
 void SignalHandler::pipeReadReady() {
   int signal;
-  if (read(m_pipefds[0], &signal, sizeof(signal)) == sizeof(signal)) {
+  while (read(m_pipefds[0], &signal, sizeof(signal)) == sizeof(signal)) {
     logger.log() << "Signal" << signal;
     emit quitRequested();
```
bakulf, line 65: Im not sure about emitting multiple quitRequested signal. Maybe we should emit just one of them and ignore the rest.

bakulf (approved):

src/signalhandler.cpp
oskirby: If we get multiple POSIX signals queued up, we always emit multiple `quitRequested` signals, it just takes an extra trip through the Qt even loop to wake up the `QSocketNotifier` again if there is still data waiting in the pipe to be read. However, the `while` loop doesn't seem to make a difference to this bug, so I can put back to the way it was.

PR #1278, opened 2021-07-09
Fallback attempt to use language code as region for City translation

strseb:
If we have Spanish as sys language, we might get "es" (which we don't have in our json ) instead of "es_ES" (which we have), so we should just attempt as a last attempt to try "language_LANGUAGE" as a possible present translation :)


bakulf:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/translations/servers.json -> we do have es_ES and so on in our json. This mapping is done in our server-language generation script.

bakulf (approved):

PR #1282, opened 2021-07-10
Generate PPA release assets from Github CI

oskirby:
It would be nice if we could generate the releases for the Ubuntu PPA from github CI to help eliminate user error. This change does a few things:
1. We revert back to using `dpkg-buildpackage` instead of `dpkg-source` in the `linux_script.sh` so that we can produce the necessary `.changes` file for the PPA release process.
2. When the CI is started from a push to a release branch, we force the package upstream version to match the branch name. 
3. Add a step in the github CI to come up with distinct build revisions, which will come up as one of the following formats:
 - `~github<run#>` if the CI was started as a push to `main`
 - `~pr<pull_request#>` if the CI was started as a pull request.
 - `<#number of commits since branched from main>` if the CI was started as a push to a release branch.
 - `~unknown` if the CI can't figure out why it was started.
4. And finally, the distro gets added to the start of the build revision.

The use of the tilde (`~`) is a special character in version number sorting that *always* sorts before everything else, which means that, when the upstream version is otherwise the same, packages get sorted in the following order
 - Packages built from a push to `main` have the lowest priority.
 - Packages built from a pull request are sorted above `main`
 - Packages built from an unknown cause are sorted above pull requests.
 - Packages built from a release have the highest priority.

As an example, pushing the 42nd commit onto `release/1.2.3` and building for Ubuntu hirsute would generate a full package version of `1.2.3-hirsute42`

In theory this means that you should be able to generate everything you would need to release to a PPA by running: `./scripts/linux_script.sh --source --version <version number>` or just letting the CI generate it for you.

TODO List:
- [x] Add support for doing package signing.
- [ ] Figure out how to automate uploads to the PPA on each push to a release branch.
- [ ] Nightly builds!


bakulf:
Question for you:
on launchpad, we need to upload 1 single orig.tar.gz file for all the packages. For instance, we need to upload orig.tar.gz for 2.3.3 and from there generate hirsute, bionic, focal, etc. Is this supported?
Usually, this is done using the ppa_script.sh script and the `-o` option.

oskirby:
To add a little more nuance, launchpad doesn't take issue with there being multiple `.orig.tar.gz` files in the upload, it takes issue with them all having different checksums (for example, because the timestamps changed). The source bundle created by this process works with launchpad because under the hood it's creating a single `.orig.tar.gz` and then creating each of the releases off of it - so the checksums all match when uploaded.

I've been testing this script out on my PPA here: https://launchpad.net/~okirby/+archive/ubuntu/mozilla-vpn-nightly and it does successfully build all the suites when doing: `./scripts/linux_script.sh --source -k <GPG Key Fingerprint> --ppa ppa:okirby/mozilla-vpn-nightly`

Obnoxiously, because the checksums on the `.orig.tar.gz` can't change once it's been uploaded, it's incredibly easy to burn a version number by uploading something that doesn't work. So I'm presently trying to figure out how to include some revision thing into the upstream version number. I'm not happy with this solution just yet.

oskirby:
After a bit of thinking, I'm going to break this up into two separate changes:
1. Updating the `linux_script.sh` to produce all the appropriate files for doing PPA uploads without changing the versioning structure.
2. Rework the versioning structure to enable nightly builds and adding automated releases to the PPA.

PR #1302, opened 2021-07-12
Experiment: Automating PPA release candidate builds

oskirby:
This is an experiment to try generating automating releases to a Launchpad PPA on every push to a release branch.

While we're at it, we also add suffixes to the package version for CI-generated packages as follows to ensure that the uploaded artifacts are easier to distinguish once installed:
- Pull requests are suffixed with `~pr<pull request #>` 
- Pushes to a release branch is suffixed with `~rc<# of commits since main>`
- Everything else is suffixed with `~github<workflow run #>`

For this to work, we need to decide on a PPA to which these packages should be uploaded and/or published, and github needs access to a GPG signing key. For the purpose of testing this, I have been using my personal PPA at `ppa:okirby/mozilla-vpn-test` and the signing key on my personal github account, but these will need to be changed to something a little more official if we decide to accept this PR.

### DIFF
--- a/.github/workflows/ppa_automation.yml
+++ b/.github/workflows/ppa_automation.yml
@@ -1,0 +1,68 @@
+name: PPA Automated Releases
+on:
+  push:
+    branches:
+      - 'releases/**'
+
+jobs:
+  ppa-release-candidate:
+    name: PPA Releases
+    runs-on: ubuntu-latest
+    environment: PPA Automation
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Generate release version
+        shell: bash
+        env:
+          GITREF: ${{github.ref}}
+        run: |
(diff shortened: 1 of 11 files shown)


.github/workflows/ppa_automation.yml
```diff
@@ -0,0 +1,68 @@
+name: PPA Automated Releases
+on:
+  push:
+    branches:
+      - 'releases/**'
```
bakulf, line 5: also the `main` one?

bakulf (approved):
I like this a lot! I was planning to do something similar today. Thank you so much!

.github/workflows/ppa_automation.yml
oskirby: I'd like to see automated builds happening off `main` as well, but they could be a little more risky and we need to make sure they go to a different PPA from the release candidates. Otherwise we could get ourselves into trouble while if the release branches have a lower version number than main.

For example, we are releasing version 2.3.3 imminently while `main` is on 2.4.0

bakulf:
@oskirby  Can we close this PR?

Issue #1225, opened 2021-06-22
Captive portal detection fails when IPv6 is enabled and local network access disabled


oskirby:
There seems to be an issue when the VPN is on that prevents network access to the addresses excluded from the tunnel (such as the captive portal and the VPN server's public IP address). This results in a bunch of error messages in the log like this:
```
[22.06.2021 07:56:55.729] (captiveportal - CaptivePortalRequest) request: http://34.107.221.82/success.txt
[22.06.2021 07:56:55.729] (networking - NetworkRequest) Network request created
[22.06.2021 07:56:55.769] (networking - NetworkRequest) Network reply received - status: 0 - expected: 0
[22.06.2021 07:56:55.769] (networking - NetworkRequest) Network error: Permission denied status code: 0 - body:
[22.06.2021 07:56:55.769] (captiveportal - CaptivePortalRequest) Captive portal request failed: 99
[22.06.2021 07:56:55.769] (captiveportal - CaptivePortalMultiRequest) Captive portal detection: 2
[22.06.2021 07:56:55.769] (captiveportal - CaptivePortalMultiRequest) Captive portal detect failed, retry!
```

We can confirm that a problem exists by attempting to ping the captive portal and the VPN server. In my case, they were 34.107.221.82 and 89.44.10.162 respectively, both of which report a general failure.

![pingfail](https://user-images.githubusercontent.com/7256630/122950370-3f12ff80-d331-11eb-892b-5219ff9c105e.png)

[mozillavpn-captive-portal-failure.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6695502/mozillavpn-captive-portal-failure.txt)

oskirby:
I think I have narrowed the problem down a bit. The packets are being dropped by the windows firewall, which is being enabled by Wireguard whenever a default route is found in the allowedIPs (see: https://github.com/WireGuard/wireguard-windows/blob/ba4edc55c4712016921bea54dbd7c0408a69ae7b/tunnel/addressconfig.go#L177). This can occur when IPv6 is enabled and local network access is disabled.

This could possibly be explained as a wireguard bug, since an IPv6 setting is causing IPv4 traffic to get blocked... but we ought to be able to program something in the firewall to permit this traffic.

oskirby:
There are two ways that we can approach this issue:
[...]

PR #1305, opened 2021-07-13
Rework routing table for IPv6 local network access

oskirby:
For correctness, we should try to exclude multicast addresses from routing over the VPN when local network access is enabled, and we also add a workaround for #1225 by excluding multicast in general for Windows.

### DIFF
--- a/src/rfc/rfc4291.cpp
+++ b/src/rfc/rfc4291.cpp
@@ -2,24 +2,19 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#include "rfc4193.h"
+#include "rfc4291.h"
 
 // static
-QList<IPAddressRange> RFC4193::ipv6global() {
-  // Allow all addresses except:
-  //  fc00::/7 (Unique-local addresses from RFC4193)
-  //  fe80::/10 (Link-local addresses)
-  //  fec0::/10 (Site-local, but deprecated by RFC3879)
-  //  ff00::/8 (Multicast addresses)
+QList<IPAddressRange> RFC4291::ipv6unicast() {
+  // Allow all addresses except multicast (ff00::/8)
   QList<IPAddressRange> list;
-  // Allow everything except fc00::/6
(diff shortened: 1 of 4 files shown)


bakulf:
This should work great on top of the other PR I wrote. What do you think?

src/rfc/rfc4291.cpp
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "rfc4291.h"
+
+// static
+QList<IPAddressRange> RFC4291::ipv6unicast() {
+  // Allow all addresses except multicast (ff00::/8)
+  QList<IPAddressRange> list;
```
bakulf, line 10: Rebasing this to the latest `main`, this should become:
```
  return QList<IPAddress>{IPAddress::create("ff00::/8")};
```

src/rfc/rfc4193.cpp
```diff
@@ -5,25 +5,21 @@
...
 
 // static
-QList<IPAddressRange> RFC4193::ipv6() {
-  // Allow all IP's except fc00::/7
+QList<IPAddressRange> RFC4193::ipv6global() {
+  // Allow all addresses except:
+  //  fc00::/7 (Unique-local addresses from RFC4193)
+  //  fe80::/10 (Link-local addresses)
+  //  fec0::/10 (Site-local, but deprecated by RFC3879)
+  //  ff00::/8 (Multicast addresses)
```
bakulf, line 13: Rebasing this to the latest `main`, this should become:
```
  return QList<IPAddress>{
    IPAddress::create("fc00::/7"),
    IPAddress::create("fe80::/10"),
    IPAddress::create("fec0::/10"),
    IPAddress::create("ff00::/8"),
  };
```

bakulf (approved):

bakulf:
@oskirby Do we still need this PR?

oskirby:
@bakulf I don't think we need this anymore. It was mostly written as a workaround for issues in the Windows firewall and we now have a proper fix in place.

PR #1307, opened 2021-07-13
Support IPv6 in IPAddress

bakulf:
Please, this PR requires extra care. It's full of complex things (at least for me).
There are 2 new components:

* BigInt - a super simple BigInt class that allows a few math operations on top of big integers stored as a vector of uint8. The reason why uint8 is because this is how data is stored in QT for IPv6 addresses. See Q_IPV6ADDR.
* BigIntIPv6Addr - this is a wrapper around BitInt to support IPv6 addresses.

The reason why BigInt supports > 16 bytes is that some math operations are "easier" if we can use 17 bytes. For instance in the "step" computation of the subnets, where we need to do steps like: `(a + 1) >> 1` where 'a' could be a sequence of 16 bytes set to 0xFF.


src/ipaddress.cpp
```diff
@@ -3,37 +3,66 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
oskirby, line 3: I'm not seeing anything particularly wrong, but I feel as though the netmasking might be a little easier to understand if we added a static method along the lines of `BigIntIPv6Addr::fill()` or `BigIntIPv6Addr::netmask()` that just returns an IPv6 address with a netmask of the desired length.

tests/unit/testipaddress.cpp
```diff
@@ -172,6 +226,35 @@ void TestIpAddress::excludeAddresses_data() {
       << "10.0.0.0/8"
       << "0.0.0.0/5,11.0.0.0/8,12.0.0.0/6,128.0.0.0/1,16.0.0.0/4,32.0.0.0/"
          "3,64.0.0.0/2,8.0.0.0/7";
+
+  QTest::addRow("world vs ::1")
```
oskirby, line 230: This is a bit of a nightmare to make sense of. It's hard to see why this would be correct. Can it be re-ordered in such a way to show the pattern (eg: increasing prefix length).

oskirby (approved):

tests/unit/testipaddress.cpp
bakulf: in theory... it's alphabetic order :) But we can sort it by prefix length.

PR #1308, opened 2021-07-13
Move done button down

### LINKED ISSUE #1291: Move “Done” button from In App Feedback “Thank you” page to the bottom
#### VPN version:
* 2.4.0 (2.202107110548);

#### Affected Platforms:
* Windows 10;

#### Prerequisites:
* Have one FxA account with subscription;

#### Steps to reproduce:
1. Go to VPN ~~> Settings~~>Give feedback;
1. Select a rating option;
1. Fill all mandatory fields and submit your feedback;
1. Observe the “Thank you” screen;

#### Expected result:
* The screen layout is as available in the [mock-ups](https://www.figma.com/file/nPn8jQUlIrhVTqh48wm7UK/In-App-Feedback-Form?node-id=149%3A0);

#### Actual result:
* The “Done” button is in the middle of the screen;

image

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-706)

mbirghan:
Closes #1291

### DIFF
--- a/src/ui/settings/ViewGiveFeedback.qml
+++ b/src/ui/settings/ViewGiveFeedback.qml
@@ -315,8 +315,8 @@ Item {
         Item {
             ColumnLayout {
                 anchors.top: parent.top
-                anchors.topMargin: window.height * .10
                 anchors.horizontalCenter: parent.horizontalCenter
+                anchors.bottom: parent.bottom
                 width: Math.min(Theme.maxHorizontalContentWidth, parent.width - Theme.windowMargin * 4)
                 spacing: 24
                 VPNPanel {
@@ -334,6 +334,9 @@ Item {
                    Layout.fillWidth: true
                    Layout.preferredHeight: Theme.rowHeight
                    onClicked: stackview.pop(StackView.Immediate)
+                   anchors.bottom: parent.bottom
+                   anchors.horizontalCenter: parent.horizontalCenter
+                   anchors.bottomMargin: Theme.windowMargin * 4
                 }
             }
         }
--- a/src/ui/settings/ViewGiveFeedback.qml
+++ b/src/ui/settings/ViewGiveFeedback.qml
@@ -316,7 +316,13 @@ Item {
             ColumnLayout {
                 anchors.top: parent.top
                 anchors.horizontalCenter: parent.horizontalCenter
-                anchors.bottom: parent.bottom
+                Component.onCompleted: {
+                    if (window.fullscreenRequired()) {
+                        anchors.bottom = parent.bottom;
+                        return;
+                    }
+                    anchors.topMargin = window.height * .10;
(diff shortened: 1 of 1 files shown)


lesleyjanenorton:
Hey @mbirghan, first of all - this looks great! I left a comment in the issue because I think there was some confusion between the spec for desktop and the spec for mobile. Ideally these would be the same but in this case, the button moves to the bottom only on mobile. 

Desktop:
Screen Shot 2021-07-14 at 1 41 29 PM

Mobile:
Screen Shot 2021-07-14 at 1 41 12 PM

lesleyjanenorton (changes requested):
Looks great - just needs one tweak!

mbirghan:
Hey @lesleyjanenorton, thank you for your feedback. I really appreciate it. 
I updated the PR and now I think it looks like it should on mobile and desktop. It is probably not the most elegant solution and I would be happy to know what I could improve or whether I should have taken a different approach.

src/ui/settings/ViewGiveFeedback.qml
```diff
@@ -334,6 +340,15 @@ Item {
                    Layout.fillWidth: true
                    Layout.preferredHeight: Theme.rowHeight
                    onClicked: stackview.pop(StackView.Immediate)
+                   anchors.horizontalCenter: parent.horizontalCenter
+                   Component.onCompleted: {
+                    if (window.fullscreenRequired()) {
+                        anchors.bottom = parent.bottom;
```
lesleyjanenorton, line 346: Thank you @mbirghan! A few last little things that I should have mentioned (sorry!) : Because this VPNButton {} is child of a ColumnLayout {}, we need to use Layout properties, instead of anchors, to position it (more on that [here](https://doc.qt.io/qt-5/qtquicklayouts-overview.html)) which would like this:

```
Component.onCompleted: {
  if (windows.fullscreenRequired()) {
    Layout.alignment = Qt.AlignBottom;
    Layout.bottomMargin = Theme.windowMargin * 4;
  }
}


The anchors in L320-325 are fine though since ColumnLayout {} is a child of Item {} and not another Layout.

File: src/ui/settings/ViewGiveFeedback.qml [hunk 2]
```diff
@@ -334,6 +340,15 @@ Item {
                    Layout.fillWidth: true
                    Layout.preferredHeight: Theme.rowHeight
                    onClicked: stackview.pop(StackView.Immediate)
+                   anchors.horizontalCenter: parent.horizontalCenter
```
lesleyjanenorton, line 343: We should remove this line or change it to use a Layout aligner - `Layout.alignment: Qt.AlignHCenter`

lesleyjanenorton (approved):

PR #1318, opened 2021-07-14
Extra string for the "good" rating in app-feedback-form

bakulf:
Before landing this, I want to be sure it's OK for the localization team to have this extra string so late in the process.
@flodolo feel free to say "No, we cannot land this PR".


lesleyjanenorton:
Update from triage meeting yesterday: Since we're already in 'string freeze', we'll land this string after 2.4 release.

lesleyjanenorton (approved):

PR #1321, opened 2021-07-14
Add fallback to raw sockets in LinuxPingSender

oskirby:
Some Linux distros, such as Debian 11, are starting to disable ICMP ping sockets, in which case the least-intrusive way install the VPN and still be able to send pings is to use a raw socket and then acquire `CAP_NET_RAW` either by starting as root or using `setcap cap_net_raw+ep /usr/bin/mozillavpn` after installation.

### DIFF
--- a/src/platforms/linux/linuxpingsender.cpp
+++ b/src/platforms/linux/linuxpingsender.cpp
@@ -10,7 +10,9 @@
 
 #include <arpa/inet.h>
 #include <errno.h>
+#include <linux/filter.h>
 #include <netinet/in.h>
+#include <netinet/ip.h>
 #include <netinet/ip_icmp.h>
 #include <sys/socket.h>
 #include <unistd.h>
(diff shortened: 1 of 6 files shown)


src/platforms/linux/linuxpingsender.cpp
```diff
@@ -19,12 +21,52 @@ namespace {
...
+    m_ident = 0;
+    return m_socket;
+  }
+  if ((errno != EPERM) && (errno != EACCES)) {
+    return -1;
+  }
+
+  // As a fallback, create a raw socket, which requires root permissions
+  // or CAP_NET_RAW to be granted to the VPN client.
+  m_socket = socket(AF_INET, SOCK_RAW, IPPROTO_ICMP);
```
bakulf, line 38: I like this approach. And we can also have another fallback solution (the 3rd one) which is similar to what mullvad does: let's run the `ping` tool. See: https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-core/src/ping_monitor/unix.rs#L82-L113

bakulf (approved):

PR #1323, opened 2021-07-14
Allow update while connected for Windows builds

oskirby:
The windows installer stops and restarts the client and the daemon, so there is no need to disconnect while downloading the updates.

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -1238,6 +1238,8 @@ void MozillaVPN::update() {
 
   setUpdating(true);
 
+  // The windows installer will stop the client and daemon before installation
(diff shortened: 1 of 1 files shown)


bakulf (approved):

src/mozillavpn.cpp
```diff
@@ -1238,11 +1238,13 @@ void MozillaVPN::update() {
 
   setUpdating(true);
 
+#ifndef MVPN_WINDOWS
```
bakulf, line 1241: Can you write 1 line of comment to explain why we have this #ifndef ?

oskirby: Will do

PR #1332, opened 2021-07-16
Remove TextField/TextArea focus on outside click

lesleyjanenorton:
On iOS it is currently quite difficult, in some cases impossible without clicking on another button, to drop focus and close the native keyboard once you've clicked into any of the text inputs.  This lets users bail out of text editing and close the native keyboard by tapping or clicking anywhere outside of the input.


bakulf (approved):

oskirby:
This PR seems to have broken the functional tests, it seems to be preventing the test worker from being able to authenticate.

PR #1333, opened 2021-07-16
Implement wgutils for Windows and disable firewall/kill switch

oskirby:
This is the proper fix for #1225 and might also fix the firewall issues noted in the split tunneling feature. As a high-level tour of what had to be changed here:
1. Implement `WireguardUtils` for Windows so that we can break out the routing table management.
2. Bump `wireguard-windows` to version v0.3.15, which adds the ability to disable the wireguard-provided firewall and routing table.
3. To make this patch bigger, I also refactored the tunnel service management so that it all lives in one file (this is really just cleanup for multihop).


bakulf (approved):

strseb:
This is awesome!

strseb (approved):

PR #1334, opened 2021-07-16
Use QTextStream to create the wg config file

bakulf:


oskirby:
I think this was already done in #1333

bakulf:
Good!

PR #1344, opened 2021-07-17
Speed up tests by mocking the devices API

oskirby:
It would be nice if the tests would fail faster, so that we can get a better picture of when our code has broken things. To that end, I've written a Python/Flask API proxy that allows us to mock out the device list, and allows us to run all tests in parallel so long as they don't need to connect to a real VPN tunnel.

Plus a bunch of other speedups:
- Skip building unit tests so that we can get to the functional tests quicker.
- Cache `.deb` files downloaded by the build task so that we can skip the PPA.
- Trim dependencies used by the test runner jobs.

While I was mucking about, I also fixed the glean tests, which got broken by #1325


oskirby:
Re-opened as #1345 so that the test-runners can get the API credentials.

PR #1347, opened 2021-07-18
Fix Functional Tests

oskirby:
There were a collection of bugs that were breaking the functional tests, including:
- Glean tests broken due to change in ping submission (tests only).
- IPv6 captive portal detection breaks when IPv6 is unreachable.
- IPv6 excluded addresses were incorrectly parsed as IPv4 in `fromIPAddressList()`


bakulf (approved):
Nice!

PR #1349, opened 2021-07-19
Fix the MacOS resource generation

bakulf:
This script is now broken because of the macos menu localization

### DIFF
--- a/.github/l10n/check_l10n_issues.py
+++ b/.github/l10n/check_l10n_issues.py
@@ -21,7 +21,7 @@ with open(os.path.join(script_folder, "exceptions.json")) as f:
     exceptions = json.load(f)
 
 # Load the .ts file
-ts_file = os.path.join(vpn_root_folder, "translations", "mozillavpn_en.ts")
+ts_file = os.path.join(vpn_root_folder, "translations", "en", "mozillavpn_en.ts")
 tree = etree.parse(ts_file)
 root = tree.getroot()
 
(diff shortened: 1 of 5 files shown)


scripts/import_macos_pkg_resources.py
```diff
@@ -91,16 +91,16 @@ def translate(root, locale):
...
 for translation in os.listdir('translations'):
-    translationFile = os.path.join('translations', translation)
-    if not os.path.isfile(translationFile):
+    translationDir = os.path.join('translations', translation)
+    if not os.path.isdir(translationDir):
         continue
 
-    if not translation.startswith('mozillavpn_') or not translation.endswith('.ts'):
+    translationFile = os.path.join(translationDir, f"mozillavpn_{translation}.ts")
+    print(translationFile)
```
flodolo, line 99: Is this print a left over?

.github/l10n/check_l10n_issues.py
```diff
@@ -21,7 +21,7 @@
     exceptions = json.load(f)
 
 # Load the .ts file
-ts_file = os.path.join(vpn_root_folder, "translations", "mozillavpn_en.ts")
+ts_file = os.path.join(vpn_root_folder, "translations", "en", "mozillavpn_en.ts")
```
flodolo, line 24: Is it expected that the linter is now failing? Would it fix itself once merged?

`Updating 'translations/mozillavpn_en.ts'`, which means the file in `en` doesn't exist.

flodolo: I think this line should be reverted. This script extracts its own .ts file in `translations/mozillavpn_en.ts`. As far as I understand, the import script relies on `translations.pri` for extracting them in subfolders, and that's not needed here.

bakulf: Right. This file now is into an "en" folder.

flodolo (approved):

PR #1351, opened 2021-07-19
First experiment to generate strings from a YAML file

bakulf:
┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-785)


flodolo (changes requested):
I think the approach is reasonable, but I'd suggest to start with a few new strings to make sure everything works, before migrating all the others.

From a quick test, it doesn't look like the comment is exported.

While I didn't look at the code in depth, another couple of points:
- We need clear specs of what's supported in the YAML file, and what's the format. Do we support multiline strings? How?
- We need to make sure to escape strings when creating the files in `translations/generated`
- Files in `translations/generated` should be committed to the repository, and there should be a CI job that makes sure those files are up to date (i.e. developer adds them to YAML, but doesn't export strings). For that, we should split the logic in its own script, and not be part of the `importLanguages.py´ script (that's definitely not importing translations…). That would also simplify the work on the l10n side, since these files can be read, and don't need to be generated from scratch.

bakulf:
>     * We need clear specs of what's supported in the YAML file, and what's the format. Do we support multiline strings? How?

Yes, now we have comments and multiline-string support.

>     * We need to make sure to escape strings when creating the files in `translations/generated`

Done!

>     * Files in `translations/generated` should be committed to the repository, and there should be a CI job that makes sure those 

There are 2 scripts:
- generate_strings.py that does the "magic".
- importLanguages.py that converts the l18n xliff files and calls generate_strings.py

Do you still think we should commit the generated files? Note that the compilation process requires calling importLanguages.py

flodolo:
> Do you still think we should commit the generated files? Note that the compilation process requires calling importLanguages.py

What I care is being able to generate the XLIFF file on the l10n repo side without having to import the l10n repo itself.

What we currently do right now is:
1. Run `lupdate` on each branch and merge the resulting `.ts` files in one.
2. Run the Python script that convert the .ts file to XLIFF (`extract_source_strings.py`).
3. Propagate changes to all other locales via `update_other_locales.py`

If we have a separate script to generate the strings, then we could run that as part of the workflow before step 1, so we might not need to store them in tree. But I need to verify that assumption first.

scripts/generate_strings.py
```diff
@@ -0,0 +1,149 @@
...
+
+def generateStrings():
+  if not os.path.isfile('translations/strings.yaml'):
+    print('Unable to find translations/strings.yaml')
+    exit(1)
+
+  yaml_file = open("translations/strings.yaml", 'r')
+  yaml_content = yaml.safe_load(yaml_file)
+
+  def complain(stringId):
```
flodolo, line 20: That's more `stop` than `complain` ;-)

flodolo (commented):
This looks a lot better than the first version ;-)

Only a couple of requests:
- Can you pass the python scripts through black? At least the new file, so that code style is consistent.
- Can you add back some actual strings in the YAML file for testing?

scripts/generate_strings.py
```diff
@@ -0,0 +1,164 @@
...
+            )
+
+            def serialize(string):
+                ret = "\\n".join(string)
+                return ret.replace('"', '\\"')
+
+            for string in stringIds:
+                output.write(f"    //% \"{serialize(string['value'])}\"\n")
+                for comment in string["comments"]:
+                    output.write(f"    //! {comment}\n")
```
flodolo, line 157: Comments start with `//:`

flodolo (changes requested):
One error to fix on the comments generation.

I think we need a linter for `strings.yaml` at this point? Otherwise someone can commit a broken file, and the builds will break.

.github/workflows/linux.yaml
```diff
@@ -50,6 +50,8 @@ jobs:
           # Add external PPA, latest version of QT is 5.12.x for Ubuntu 20.04
           sudo add-apt-repository ppa:beineri/opt-qt-5.15.2-focal -y
           sudo apt-get install debhelper devscripts equivs -y
+          pip3 install pyhumps
```
flodolo, line 53: I think it would make more sense to have a requirements.txt, and install from that? That way, you only update dependencies once, and you can also set versions to avoid sudden breakages in CI.

Also, I think at this point you can just use `pip` and `python,` no need for the 3 suffix?

translations/strings.yaml
```diff
@@ -0,0 +1,55 @@
...
+# ```
+# QString foo =
+#    L18nStrings::instance()->tr(L18nStrings::InitialViewGetStartedButton)
+# ```
+#
+# Remember! When you change this file, you must run the
+# `./scripts/generate_strings.py` script to regenerate the string files.
+#
+
+main:
```
flodolo, line 52: Make sure to remove these before merging.

flodolo (approved):

.github/workflows/linux.yaml
bakulf: OK! Follow up for this.

Issue #1350, opened 2021-07-19
[Windows] VPN cannot be turned ON


ValentinaPC:
#### VPN version:
* 2.4.0 (2.202107170646);

#### Affected Platforms:
* Win 10;

#### Prerequisites:
* Have one FxA account without subscription;
* Have the VPN installed on your machine;

#### Steps to reproduce:
1. Login to VPN with the account from prerequisites;
1. Turn ON the app;

#### Expected result:
* VPN is turned ON with success;

#### Actual result:
* VPN cannot be turned on and internet is not working on device;

#### Notes:
* there were cases when VPN started working, but after making some changes in the “Network settings” section from app Settings, it could not be turned back ON and the results regarding internet were the same.
* Internet starts working ONLY after VPN uninstall;
[mozillavpn-2021-7-19_1.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6841646/mozillavpn-2021-7-19_1.txt)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-744)

bakulf:
@oskirby can this be a regression of your PRs?

oskirby:
I'll take a look into it

oskirby:
Looking through the logs, and based on the build date, I think this was actually fixed by #1347. I think the bug was that `fromIPAddressList()` was outputting an IPv6 address and calling it IPv4, which eventually causes the daemon to try and produce bogus routing table entries.

I do not seem to be able to reproduce the issue on builds off the `main` branch after that fix.

ValentinaPC:
Verified as fixed on VPN v. 2.4.0 (2.202107191712) build, while using Windows.
The VPN can be turned ON and works as expected. Same for the internet.
![image](https://user-images.githubusercontent.com/15685960/126291608-3895e9cc-ee92-4cf2-887e-692616999e2d.png)

PR #1356, opened 2021-07-19
Fix VPN Activation on Windows

strseb:
Small fix for #1350
-> Also a bit better error handling, if the activation fails we need to disable the TunnelService. 
->Adding to RFC4193::ipv6() exclude list makes WireguardUtilsWindows::addRoutePrefix fail with "invalid parameter" :( 
No idea why so far, but that might be not essential to have in 2.4 i guess?

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -654,8 +654,11 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
     excludeIPv4s.append(RFC1918::ipv4());
 
     if (ipv6Enabled) {
+      #ifndef MVPN_WINDOWS
+      // TODO: Does not seem to work on windows so far :(
       logger.log() << "Filtering out the local area networks (rfc 4193)";
       excludeIPv6s.append(RFC4193::ipv6());
+      #endif 
     }
   } else if (FeatureList::instance()->userDNSSupported() &&
              !SettingsHolder::instance()->useGatewayDNS() &&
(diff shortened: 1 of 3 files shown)


src/controller.cpp
```diff
@@ -654,8 +654,11 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
     excludeIPv4s.append(RFC1918::ipv4());
 
     if (ipv6Enabled) {
+      #ifndef MVPN_WINDOWS
+      // TODO: Does not seem to work on windows so far :(
```
bakulf, line 658: Can you tell us more? Why it's broken?

src/daemon/daemon.cpp
```diff
@@ -81,6 +82,16 @@ bool Daemon::activate(const InterfaceConfig& config) {
     Q_ASSERT(!m_connected);
     return activate(config);
   }
+  Q_ASSERT(!m_connected);
+  auto cleanup= qScopeGuard([&] {
```
bakulf, line 86: linter :)

bakulf (changes requested):

oskirby:
I think that the combination of `RFC4193::ipv6()` and the IPv6 support in the exclusion list is resulting in duplicate routing table entries trying to be added. In the logs from #1350 I noticed that we seem to be adding both `2000::/5` and '2800::/5` to the routing table, which both have the same leading 5 bits.

PR #1365, opened 2021-07-20
`Select all` button in the AppPermission view

bakulf:


bakulf:
It seems that we do not want this. But in case, the PR is ready.

PR #1369, opened 2021-07-20
Unique device ID to remove devices - closes #1360

### LINKED ISSUE #1360: Avoid duplicate devices
Before 2.4.x, at the end of the authentication flow, the client deleted devices with the same name as the current one to avoid duplicate devices. For example, if the current device is called "Baku's Android" and there was another "Baku's Android" device, the old one was removed.

But it turned out that there are users with multiple devices with the same name. Because of this, in 2.4.x we have removed the device deletion during the authentication. But we can do better than this!

What about if we add a `unique_id` field in the `Device` interface (guardian-website) and we populate this new field with something unique. We can use this unique id to remove the correct duplicate device.

Android: https://developer.android.com/training/articles/user-data-ids
iOS: https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor?language=objc
MacOS: https://developer.apple.com/library/archive/technotes/tn1103/_index.html
Windows: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcurrenthwprofilea (maybe)
Linux: https://man7.org/linux/man-pages/man5/machine-id.5.html :question:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-750)

bakulf:
┆Issue is synchronized with this [Jiraserver Bug](https://mozilla-hub.atlassian.net/browse/VPN-822)

### DIFF
--- a/src/models/device.cpp
+++ b/src/models/device.cpp
@@ -59,6 +59,8 @@ QString Device::currentDeviceReport() {
   return buffer;
 }
 
+QString Device::uniqueDeviceId() { return QSysInfo::machineUniqueId(); }
+
 Device::Device() { MVPN_COUNT_CTOR(Device); }
 
 Device::Device(const Device& other) {
@@ -70,6 +72,7 @@ Device& Device::operator=(const Device& other) {
   if (this == &other) return *this;
 
   m_deviceName = other.m_deviceName;
+  m_uniqueId = other.m_uniqueId;
   m_createdAt = other.m_createdAt;
   m_publicKey = other.m_publicKey;
   m_ipv4Address = other.m_ipv4Address;
@@ -91,6 +94,9 @@ bool Device::fromJson(const QJsonValue& json) {
   if (!name.isString()) {
     return false;
   }
+
+  QJsonValue uniqueId = obj.value("unique_id");
+  // No checks here.
 
   QJsonValue pubKey = obj.value("pubkey");
   if (!pubKey.isString()) {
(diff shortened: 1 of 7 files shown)


src/models/device.cpp
```diff
@@ -59,6 +59,8 @@ QString Device::currentDeviceReport() {
   return buffer;
 }
 
+QString Device::uniqueDeviceId() { return QSysInfo::machineUniqueId(); }
```
strseb, line 62: Just a note: We need a followup bug here for android as machineUniqueId is not implemented for it: 
https://bugreports.qt.io/browse/QTBUG-72684

bakulf:
can we land this PR? :)

birdsarah:
> can we land this PR? :)

Open questions from me: 
1) what about basti's comment about android?
2) why not use public key?

bakulf:
>     1. what about basti's comment about android?

For android, we need to implement something unique too. But it can be done in a follow-up.

>     2. why not use public key?

Because we don't have it. Here is the scenario:
1. As a user, I install the client on device A.
2. I remove the client from device A without a logout.
3. I reinstall the client on device A. How to know that device A1 and A2 are the same device? Right now we don't know this and we end up duplicating the device (the public key is stored in the config file which is removed...). We use to identify the device by "name" but this is not a good approach because there are users with 2 (or more) devices with the same name. 

The solution here is to use a unique ID per device.

strseb (approved):

src/models/device.cpp
birdsarah: Stupid question: what happens on android in this case then?

birdsarah: I READ THE DOCS! SHOCKING! The answer is: "If no unique ID could be determined, this function returns an empty byte array."

PR #1376, opened 2021-07-20
Multihop Tunnelling

oskirby:
This is a work in progress to get multihop tunnelling working. The approach taken here is to replace the `Server& server` argument with a `QList<Server> serverList` in the `Controller::activate()` method, platforms that then support multihop can bring up the interfaces as needed to enable a multihop tunnel. Platforms that don't support multihop should just ignore everything but the first element.

Behind the scenes, in the daemon, we add a `hopindex` argument to the `InterfaceConfig` which tells the daemon which hop out of a multihop tunnel is being configured, with zero being the exit node.

TODO List:
- [x] Implement Multihop for Linux
- [x] Implement Multihop for Windows
- [x] Implement Multihop for MacOS
- ~Better server selection algorithm for the intermediate nodes.~ moved to #1502 
- [x] Commit wireguard-go fixes somewhere.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-794)


src/controller.cpp
```diff
@@ -623,13 +641,14 @@ void Controller::statusUpdated(const QString& serverIpv4Gateway,
...
 QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
-    const Server& server) {
+    const QList<Server>& serverList) {
   logger.debug() << "Computing the allowed ip addresses";
 
   bool ipv6Enabled = SettingsHolder::instance()->ipv6Enabled();
 
   QList<IPAddress> excludeIPv4s;
   QList<IPAddress> excludeIPv6s;
+  const Server& server = serverList.last();
```
bakulf, line 651: Ok so, the first one is the exit node and the last one is the ingress node.

src/daemon/daemon.cpp
```diff
@@ -326,29 +333,51 @@ QString Daemon::logs() {
...
-  if (!supportWGUtils()) {
-    qFatal("Have you forgotten to implement switchServer?");
+bool Daemon::supportServerSwitching(const InterfaceConfig& config) const {
+  if (!m_connections.contains(config.m_hopindex)) {
     return false;
   }
+  const InterfaceConfig& current =
+      m_connections.value(config.m_hopindex).m_config;
+
+  return current.m_privateKey == config.m_privateKey &&
```
bakulf, line 343: we could implement an `operator==()` or a similar method.

windows/tunnel/interfacewatcher.go
```diff
@@ -0,0 +1,156 @@
+/* SPDX-License-Identifier: MIT
+ *
+ * Copyright (C) 2019-2021 WireGuard LLC. All Rights Reserved.
+ */
+
```
bakulf, line 5: Is this code taken from the windows wireguard repo? Is this copyright header correct?

bakulf (approved):

windows/tunnel/interfacewatcher.go
oskirby: Yes, all the code in this directory is taken from the wireguard-windows repository, which is under the MIT software license and I have simply removed a couple of functions. However, I forgot to include the copyright file that goes along with these files in order to be compliant with the MIT license.

src/daemon/daemon.cpp
oskirby: This is not quite the same as testing for equality, it's really testing that the bits that would go into the `[Interface]` section of a wg-quick config file are unchanged but allows for the peers to be different.

src/controller.cpp
oskirby: That is correct, I can add a comment to help explain this better.

bakulf (approved):

src/platforms/linux/daemon/iputilslinux.cpp
```diff
@@ -56,7 +61,10 @@ bool IPUtilsLinux::setMTUAndUp() {
   strncpy(ifr.ifr_name, WG_INTERFACE, IFNAMSIZ);
 
   // MTU
-  ifr.ifr_mtu = 1420;
+  // FIXME: We need to know how many layers deep this particular
```
bakulf, line 64: Follow up?

src/ui/settings/ViewNetworkSettings.qml
```diff
@@ -98,6 +98,26 @@ Item {
...
             }
 
+            VPNCheckBoxRow {
+                id: multihopTunnel
+                objectName: "settingMultihopTunnel"
+                visible: VPNFeatureList.multihopSupported
+                Layout.rightMargin: Theme.windowMargin
+
+                //% "Multihop tunnel"
+                labelText: qsTrId("vpn.settings.multihop")
```
flodolo, line 108: These two strings just landed outside of strings.yaml (I thought the idea was to always use that from now on?).

More important, they conflict with the terminology used in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1573 (multi-hop vs multihop)

Which one is correct?

There seems to be also differences in terms of implementation, since I can't find anything similar to these strings in the PR.

oskirby: Ah, right, I forgot about the change to use strings.yaml. Sorry about that.

For what it's worth, this setting should only be temporary until such time as the UX for multihop is completed. So, hopefully we can simply remove these strings before code freeze.

src/platforms/linux/daemon/iputilslinux.cpp
oskirby: Good catch! This is mostly a case of the comment being stale after some of the refactoring that went on in this PR. Since this one interface is shared by all hops, we have to set an MTU at least big enough for all the hops. Assuming the route to the internet has an Ethernet MTU of 1500, then the correct value to choose would be `ETH_MTU - WG_MTU_OVERHEAD`.

However, the spirit behind this comment is that each hop adds another `WG_MTU_OVERHEAD` to the packet, meaning that the final hops have a smaller effective MTU than the first hops. This isn't something that we can configure on the interface and we would have to do it in the routing table somehow.

PR #1377, opened 2021-07-21
Fix recursive rearrange in ViewNetworkSettings.qml + UX requests 

lesleyjanenorton:


src/ui/components/VPNCheckBoxAlert.qml
```diff
@@ -8,17 +8,18 @@ import Mozilla.VPN 1.0
 import "../themes/themes.js" as Theme
 
 RowLayout {
-    id: turnVPNOffAlert
     property var leftMargin:  56
+    property var alertColor: Theme.checkBoxWarning
+    property var errorMessage: "..."
```
bakulf, line 13: This is not exposed, right?

bakulf (approved):

src/ui/components/VPNCheckBoxAlert.qml
lesleyjanenorton: It's not exposed currently but I just pushed up something to remove the placeholder text so it doesn't get accidentally exposed in the future.

PR #1379, opened 2021-07-21
Expose unfinished translations only in staging builds

bakulf:


bakulf:
Because the bug was in windows and macos builds, we can release a 2.3.3 with this issue fixed. iOS, android and Linux are OK.

flodolo (approved):

Issue #1380, opened 2021-07-21
 Split tunneling is not working on Windows


ValentinaPC:
#### VPN version:
- 2.4.0 (2.202107201434)

#### Affected Platforms:
- Windows 10;

#### Prerequisites:
- Have one FxA account without subscription;
- Have the VPN installed on your machine;

#### Steps to reproduce:
1. Login to VPN with the account from prerequisites;
2. Go to Settings-> App Permissions;
3. Disable the feature;
4. Observe the list of the apps;
5. Mark an app (Firefox) to be excluded from the protection list.
6. Turn on VPN and use Firefox to check its protection;

#### Expected result:
4- All apps should be checked/unchecked (same behavior on all platforms - "all checked" is the current behavior on Linux);
6- The REAL IP is listed while using the unprotected app;

#### Actual result:
4- All apps are unchecked on Win and Android and checked on Linux;
6- Internet is not working on the unprotected app;

#### Notes:
- Not reproducing on Android/Linux;

ValentinaPC:
Verified as fixed on VPN 2.4.0 (2.202107220713) while using Windows.
The feature works as expected.

PR #1383, opened 2021-07-21
Do not rely on the metrics table to choose right adapters for split-tunnel

strseb:
As afraid, checking the metrics table does not really work except for clean vm's >:c 

I.e NordVPN put's itself into the first spot even when disabled. 

So he's what i want to do instead: 
When the deamon starts before the vpn starts, check what adapter can route to 1.1.1.1, and use that as the "Internet" Adapter.  
For the vpn adapter check the name :)

### DIFF
--- a/src/platforms/windows/daemon/windowsdaemon.cpp
+++ b/src/platforms/windows/daemon/windowsdaemon.cpp
@@ -39,6 +39,12 @@ WindowsDaemon::~WindowsDaemon() {
   MVPN_COUNT_DTOR(WindowsDaemon);
   logger.log() << "Daemon released";
 }
+void WindowsDaemon::prepareActivation(const InterfaceConfig& config) {
+  // Before Creating the interface we need to check which adapter
+  // routes to the Server endpoint
+  auto serveraddr = QHostAddress(config.m_serverIpv4AddrIn);
+  m_inetAdapterIndex = WindowsCommons::AdapterIndexTo(serveraddr);
+}
 
 bool WindowsDaemon::run(Op op, const InterfaceConfig& config) {
   bool splitTunnelEnabled = config.m_vpnDisabledApps.length() > 0;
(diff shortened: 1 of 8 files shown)


strseb:
Putting into draft while QA checks if this is the right fix :)

strseb:
Opening this up, as Valentina confirmed this closes #1380

src/platforms/windows/daemon/windowssplittunnel.cpp
```diff
@@ -24,6 +24,9 @@ Logger logger(LOG_WINDOWS, "WindowsSplitTunnel");
 }
 
 WindowsSplitTunnel::WindowsSplitTunnel(QObject* parent) : QObject(parent) {
+  // Take a snapshot what adapter is currently routing to the internet
+  // which will later recieve the splitted traffic
```
bakulf, line 28: receive?

src/platforms/windows/windowscommons.cpp
```diff
@@ -99,3 +103,33 @@ QString WindowsCommons::tunnelLogFile() {
 
   return QString();
 }
+
+// static
+int WindowsCommons::CurrentGatewayIndex() {
+  QHostAddress dst("1.1.1.1");
```
bakulf, line 109: replace this with the API endpoint.

bakulf (changes requested):

src/platforms/windows/daemon/windowsdaemon.cpp
```diff
@@ -39,6 +39,12 @@ WindowsDaemon::~WindowsDaemon() {
   MVPN_COUNT_DTOR(WindowsDaemon);
   logger.log() << "Daemon released";
 }
+void WindowsDaemon::prepareActivation(const InterfaceConfig& config) {
+  // Before Creating the interface we need to check which adapter
```
bakulf, line 43: s/C/c/

src/platforms/windows/daemon/windowsdaemon.cpp
```diff
@@ -39,6 +39,12 @@ WindowsDaemon::~WindowsDaemon() {
...
+  // routes to the Server endpoint
```
bakulf, line 44: s/S/s/

bakulf (approved):

PR #1388, opened 2021-07-21
Enable calling abortAuthentication multiple times

### LINKED ISSUE #938: Crash caused by being offline
From user feedback: Calling "Get Started" ("Erste Schritte") while the phone is offline crashes the VPN app.

platform: Android 10

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-165)

strseb:
Currently move away from the authenticating state when we abort but also when we encounter an error
so we might abort Authentication when already back into StateInitialize because also an error occured

Fixes #938 

﻿

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -374,7 +374,10 @@ void MozillaVPN::authenticate() {
 }
 
 void MozillaVPN::abortAuthentication() {
-  logger.log() << "Abort authentication";
+  if(m_state == StateInitialize){
+    return;
+  }
+  logger.log() << "Abort authentication from:" << m_state;
   Q_ASSERT(m_state == StateAuthenticating);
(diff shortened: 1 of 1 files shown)


src/mozillavpn.cpp
```diff
@@ -844,7 +847,7 @@ void MozillaVPN::errorHandle(ErrorHandler::ErrorType error) {
     } else {
       emit triggerGleanSample(GleanSample::authenticationFailure);
     }
-    setState(StateInitialize);
```
bakulf, line 847: I don't understand this change. The only result I see is that we trigger `triggerGleanSample`twice.
Can you tell me more why this fixes the android crash?

bakulf (changes requested):

PR #1396, opened 2021-07-22
Better logging - debug/warning/error

bakulf:


oskirby:
Yes please!

Also of note, I had already implemented something like this in #1327 to resolve the production vs. staging logging question (see [src/logger.cpp](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1327/files#diff-2c46dd80094c3ffd00cd309628cb1d6e5c695f69f8dafb5c40747369a5d6ded0))

src/logger.cpp
```diff
@@ -11,13 +11,17 @@ Logger::Logger(const QString& module, const QString& className)
 Logger::Logger(const QStringList& modules, const QString& className)
     : m_modules(modules), m_className(className) {}
 
-Logger::Log Logger::log() { return Log(this); }
```
oskirby, line 14: I'd also like to see a Logger::info() in between `warning()` and `debug()` for messages that we think should be printed to console even though they might not be an error (eg: login banners, major state changes and etc). I find it kind of unsettling that the console is completely silent on production builds. I prefer a little sign of life in my applications.

oskirby (approved):

PR #1413, opened 2021-07-23
Add support for browser-less login using CLI commands

oskirby:
This adds the ability to perform email and password-based login using the `login` CLI command, and does so using the firefox accounts API directly instead of needing to start a browser session. To complete such a login, run the command `mozillavpn login -p`, after which you will be prompted for a username (FxA e-mail) and password on the console.

Authentication is performed using the android flow, which seems like a bit of a hack, but it turned out to be a hard problem to extract the OAuth code from the desktop flows, which bury it somewhere in the javascript and session context.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-791)


src/tasks/authenticate/taskpasswordauth.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+namespace {
+Logger logger(LOG_MAIN, "TaskPasswordAuth");
+}
+
+TaskPasswordAuth::TaskPasswordAuth(const QString& username,
+                                   const QString& password)
+    : Task("TaskPasswordAuth"), m_username(username) {
+  MVPN_COUNT_CTOR(TaskPasswordAuth);
+
```
bakulf, line 29: In theory this should be a bit more complex than this because we should:
1. check if the user account exists (/v1/account/status + email)
2. if it doesn't exist, we should create it.

src/tasks/authenticate/taskpasswordauth.cpp
```diff
@@ -0,0 +1,246 @@
...
+void TaskPasswordAuth::authFailed(ErrorHandler::ErrorType error) {
+  logger.log() << "Authentication failed" << error;
+  MozillaVPN::instance()->errorHandle(ErrorHandler::AuthenticationError);
+  emit completed();
+}
+
+void TaskPasswordAuth::netRequestFailed(QNetworkReply::NetworkError error,
+                                        const QByteArray& data) {
+  Q_UNUSED(data);
+  authFailed(ErrorHandler::toErrorType(error));
```
bakulf, line 90: In theory here we should check if the email validation is required.
See: errno=125 + verificationEmail = email-captcha

When this happens we should send a request to: /v1/account/login/send_unblock_code then continue passing the code.

src/tasks/authenticate/taskpasswordauth.cpp
```diff
@@ -0,0 +1,246 @@
...
+  QByteArray challenge =
+      QCryptographicHash::hash(m_verifier, QCryptographicHash::Sha256);
+
+  QUrlQuery query;
+  query.addQueryItem("code_challenge", challenge.toBase64());
+  query.addQueryItem("code_challenge_method", "S256");
+  query.addQueryItem("user_agent", NetworkManager::userAgent());
+
+  // Login using the android auth flow to convince the API to supply the OAuth
+  // code as a query parameter during redirection instead of invoking a browser
```
bakulf, line 67: This seems an hack :) We should have `../login/cli`

bakulf (approved):

src/tasks/authenticate/taskpasswordauth.cpp
oskirby: This is completely a hack, I agree. I would actually recommend just making the API a little more consistent about how it returns the OAuth code (eg: if no port is specified, always set it as a query string parameter), but for the scope of this PR I wanted to work with the existing API.

src/tasks/authenticate/taskpasswordauth.cpp
oskirby: I feel that if the account doesn't exist then we should return an authentication failure from this task and inform the user about it (for example, it might have been a typo). Going directly to account creation seems inappropriate for a CLI command.

When run from the GUI, however, we could use the `NoSuchAccount` error status and use it to drive a state change to an account creation view, but that's something that could be added in another PR later.

PR #1416, opened 2021-07-23
Fix moving connection time label

### LINKED ISSUE #944: Connection time label moves
From user feedback: The connection time label moves horizontally because it's centered and the numbers have variable width (e.g. HH:MM:S0 to HH:MM:S1 will move the start left)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-168)

flozia:
As mentioned in #944 the connection time label jumps horizontally while running. This is caused by the numbers having different spacings.

The Inter font family has no monospaced variant, but it offers the OpenType feature [tabular figures (tnums)](https://rsms.me/inter/#features/tnum) — this would be a perfect use case here.  I have not come across a way to enable `tnums` in QML. With CSS the `tnums` can be activated with `font-feature-settings: "tnum";`.

For now I created a label component that sets each character individually with a fixed width.

Resolves #944

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-790)


src/ui/components/VPNControllerView.qml
```diff
@@ -618,16 +657,28 @@ Item {
             height: 16
         }
 
-        VPNInterLabel {
-            id: logoSubtitle
-            objectName: "controllerSubTitle"
-
-            lineHeight: Theme.controllerInterLineHeight
-            Accessible.ignored: true
-            width: parent.width - Theme.windowMargin
```
lesleyjanenorton, line 627: This is tricky! Unfortunately if we remove width here the `logoSubtitle` text loses its ability to wrap, which becomes problematic when viewing the app in more verbose locales (try  going to Settings -> Languages and changing the default language to German and you'll see what I mean). But if we leave it, the timer isn't visible or is way out of alignment.

I think the easiest thing to do here may actually be to treat "Secure and private • 00 : 00 : 00" string as a special case and handle it separately from the other `logoSubtitle` strings - sort of like we do for `VPNConnectionStability { }`. In that case we'd just need to:
- revert the changes `logoSubtitle` and hide it when `VPNController.StateOn`
- add a new `RowLayout { } ` that is only visible when `VPNController.StateOn` to handle the "Secure and private •" string and `VPNSemiMonoLabel { }`. Something along the lines of:
```
RowLayout {
  id: someNewId
  opacity: .8
  visible: false

  VPNInterLabel {
    objectName: "secureAndPrivateSubtitle"

    lineHeight: Theme.controllerInterLineHeight
    Accessible.ignored: true
    onPaintedHeightChanged: if (visible) col.handleMultilineText()
    onVisibleChanged: if (visible) col.handleMultilineText()
    color: Theme.white

    //% "Secure and private"
    //: This refers to the user’s internet connection.
    text: qsTrId("vpn.controller.active") + " • "
  }

  VPNSemiMonoLabel {
    id: connectionTime
    Accessible.ignored: true
  }
}

File: src/ui/components/VPNSemiMonoLabel.qml [hunk 2]
```diff
@@ -0,0 +1,49 @@
+import QtQuick 2.0
```
lesleyjanenorton, line 1: Needs disclaimer.
```

**Review by lesleyjanenorton — CHANGES_REQUESTED** [review]
Thank you @flozia ! This is a great little hack for forcing mono-spaced text! Noticed a few things when running locally and left more detailed comments/ideas on workarounds in the files.

**Review by lesleyjanenorton — COMMENTED** [review]
I just realized why we are seeing the  "Connection Unstable..." pop up simultaneously and create layout issues when it disappears. We just need to update these instances of `target: logoSubtitle` in VPNConnectionStability.qml to `target: logoSubtitleOn`  https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b53217ecb558bbba07502ec664fa307e54e50261/src/ui/components/VPNConnectionStability.qml#L47 https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b53217ecb558bbba07502ec664fa307e54e50261/src/ui/components/VPNConnectionStability.qml#L60 <br> And then remove the lines 300-303 in VPNControllerView.qml where we set the visibility to true anytime the state is VPNController.StateOn.

File: src/ui/components/VPNControllerView.qml [hunk 3]
```diff
@@ -630,6 +661,33 @@ Item {
             onVisibleChanged: if (visible) col.handleMultilineText()
         }
 
+        RowLayout {
+          id: logoSubtitleOn
+
+          anchors.horizontalCenter: parent.horizontalCenter
+          opacity: 0.8
+          visible: false
```
lesleyjanenorton, line 669: We may not need this line after making the changes to VPNConnectionStability.qml

src/ui/components/VPNControllerView.qml
```diff
@@ -274,11 +294,12 @@ Item {
...
                 target: logoSubtitle
-                //% "Secure and private"
-                //: This refers to the user’s internet connection.
-                text: qsTrId("vpn.controller.active") + "  •  " + formatTime(VPNController.time)
-                color: "#FFFFFF"
-                opacity: 0.8
+                visible: false
+            }
+
+            PropertyChanges {
```
lesleyjanenorton, line 300: This can be removed after changes to VPNConnectionStability.qml

lesleyjanenorton (commented):
_

src/ui/components/VPNControllerView.qml
```diff
@@ -274,11 +294,7 @@ Item {
 
             PropertyChanges {
                 target: logoSubtitle
-                //% "Secure and private"
-                //: This refers to the user’s internet connection.
-                text: qsTrId("vpn.controller.active") + "  •  " + formatTime(VPNController.time)
-                color: "#FFFFFF"
-                opacity: 0.8
+                visible: false
             }
```
lesleyjanenorton, line 299: My bad - I think I told you we could get rid of this little block but it looks like like we do actually need it. After that, everything is good to go. Thank you so much for all of the iterations with this!

```            
PropertyChanges {
    target: logoSubtitleOn
    visible: true
}

flozia: I added the block back in. I didn‘t notice any difference in visual behaviour of the subtitle — were there any other side effects?

lesleyjanenorton: Yeah I noticed that if I disconnected when the 'Unstable' message was visible, then reconnected, `logoSubtitleOn` was still hidden.

lesleyjanenorton (approved):
Thank you!

PR #1420, opened 2021-07-26
Introduce AuthenticationInApp state machine

bakulf:
There are a couple of things I would like to do:
1. complete the state machine with more error handling [DONE]
2. support account creation
3. move _everything_ in the AIP listener [DONE]

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-797)


bakulf:
This PR contains much more than what the title says:
- it has front-end code
- it implements sign in/up and the email verification
- error handling

Basically, what is missing is:
- password forgotten
- all the frontend implementation
- testing :)

src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -0,0 +1,592 @@
...
+                    });
+          });
+}
+
+void AuthenticationInAppListener::processErrorCode(int errorCode) {
+  AuthenticationInApp* aip = AuthenticationInApp::instance();
+  Q_ASSERT(aip);
+
+  switch (errorCode) {
+    case 101:  // Account already exists
```
oskirby, line 359: I dislike seeing naked integers hanging around in the code, I feel like these should be an `enum`.

oskirby (approved):
I think this is a great step forward, though I would suggest that some of the less common authentication states (such as password recovery and account creation) should get redirected to a browser session rather than trying to handle them in-app.

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: I agree. Would be OK-ish if I write a URL of where the numbers are defined?

oskirby: That would certainly be better

PR #1444, opened 2021-07-28
Better naming for AIP and TOTP support

bakulf:
The new flow is the following:
1. checkAccount(email) -> from here 2a or 2b
2a. setPassword(password) + signIn()
2b. setPassword(password) + signUp()
3. optional: email verification
4. optional totp verification
5. optional email verification (security reasons - unblock code)

┆Issue is synchronized with this [Jiraserver Bug](https://mozilla-hub.atlassian.net/browse/VPN-821)


src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -137,12 +134,13 @@ QByteArray AuthenticationInAppListener::generateAuthPw() const {
   return hash.result(32, "identity.mozilla.com/picl/v1/authPW");
 }
 
+void AuthenticationInAppListener::setPassword(const QString& password) {
+  m_password = password;
```
oskirby, line 138: As a nitpicky thing, I would store the authPW instead of the user's unhashed password here just to make it a little harder to grep for passwords out of a coredump. Stealing someone's authPW could compromise their FxA account, but stealing the unhashed password could compromise a wider set of services.

oskirby (approved):
Looks good, but I have only reviewed the C++ stuff. I'll defer to @lesleyjanenorton for the QML and UI components

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: Good point! done.

PR #1448, opened 2021-07-28
placeholder to make windows' compiler happy when there are no strings

bakulf:
┆Issue is synchronized with this [Jiraserver Bug](https://mozilla-hub.atlassian.net/browse/VPN-826)

### DIFF
--- a/scripts/generate_strings.py
+++ b/scripts/generate_strings.py
@@ -165,6 +165,10 @@ const char* const L18nStrings::_ids[] = {
                     output.write(f"    //: {comment}\n")
                 output.write(f"    QT_TRID_NOOP(\"{string['stringId']}\"),\n\n")
 
+            # This is done to make windows compiler happy
+            if len(stringIds) == 0:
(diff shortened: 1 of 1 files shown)


scripts/generate_strings.py
```diff
@@ -165,6 +165,10 @@ def serialize(string):
                     output.write(f"    //: {comment}\n")
                 output.write(f"    QT_TRID_NOOP(\"{string['stringId']}\"),\n\n")
 
+            # This is done to make windows compiler happy
+            if len(stringIds) == 0:
+                output.write(f"    QT_TRID_NOOP(\"vpn.dummy.ignore\"),\n\n")
```
flodolo, line 170: This is really not great, as we'll end up with this placeholder exposed for localization in the pipeline.

If there's no other way around (can we avoid generating these files if there are no strings?), I would suggest to move an actual string to `strings.yaml` instead.

flodolo (changes requested):

flodolo (approved):
WFM if it solves the issue, since this doesn't get in the pipeline

PR #1450, opened 2021-07-28
Fix Hirsute/Groovy/RPM packages by fixing python build-deps

oskirby:
Instead of needing pip, we can rely on the distribution to provide `pyyaml`, as long as we mention it as a `Build-Depends`. There isn't a corresponding package for `pyhumps`, but that one is easy enough to work around by writing our own implementation of `pascalize()`

Unfortunately, groovy is still broken by the recent addition of `QPasswordDigestor` but we aren't actually building that in CI.

┆Issue is synchronized with this [Jiraserver Bug](https://mozilla-hub.atlassian.net/browse/VPN-828)


flodolo:
I think this is more for baku to review, as I don't know anything about the rest of the build system.

The only note is that I don't think we strictly need to make the script executable?

oskirby:
Another set of eyes to make sure I didn't break something in my `pascalize()` implementation might be nice since I'm not entirely sure what kind of strings we are expecting to feed into it.

As for the executable bit, I'm a fan of making python scripts executable wherever possible, so that I don't need to type the leading `python3` when running them, but you're right, it's not necessary.

bakulf (approved):

bakulf:
Let's forget about Groovy. Launchpad PPA has removed the support for groovy. We should drop the support for Groovy too.

PR #1457, opened 2021-07-29
Fixed go link errors.

swatson555:
When linking against netfilter.a the linker would emit errors because it can't find Go's runtime object code. Linking with `-lgo` fixes theses errors.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -359,6 +359,8 @@ else:linux:!android {
     QT += networkauth
     QT += dbus
 
+    LIBS += -lgo
+
     CONFIG += c++14
 
     DEFINES += MVPN_LINUX


oskirby:
Thank you for the PR, this looks like a reasonable change.

Would you mind telling us a bit more about the distro you are trying to compile this on, and what version of Go you might be using? We don't seem to encounter this on any of our usual platforms.

swatson555:
I'm compiling this on the latest version of Fedora and they're currently using Go 1.6.

oskirby:
Unfortunately, I'm afraid that this change does not build successfully on any of the Ubuntu targets, which doesn't seem to have a `libgo.a` anywhere that I can find.

PR #1459, opened 2021-07-29
Text input states

### LINKED ISSUE #1426: Change Informational Blue colour in app to align with design system and colour palette
The blue we're currently using for focus states is aligned with firefox. 

We'd like to change the blue focus states and border to match the blue we currently have in our design system. 

This is the what the input field will look like with the design system colours: #0060DF for the thin border and #0060DF at 40% opacity for the thicker outline.

Screen Shot 2021-07-26 at 4 47 32 PM

Here's the link to the [figma file](https://www.figma.com/file/cBaBmcPisqZ0QQX5LLzpjU/Nebula-Design-System-V1?node-id=202%3A110) which has the proposed colours.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-805)

flozia:
This adds states to text inputs in order to handle the multiple states of the [Nebula Design System inputs](https://www.figma.com/file/cBaBmcPisqZ0QQX5LLzpjU/Nebula-Design-System-V1?node-id=202%3A110).

The focus of the this PR is to resolve #1426 and adjust the text input highlights.


src/ui/components/forms/VPNTextArea.qml
```diff
@@ -38,13 +35,16 @@ Item {
 
         TextArea.flickable: TextArea {
             property var maxCharacterCount: 1000
-            property bool loseFocusOnOutsidePress: true
+            property bool loseFocusOnBlur: true
```
lesleyjanenorton, line 38: We'll need to update this line too https://github.com/mozilla-mobile/mozilla-vpn-client/blob/242e86cfb0862be5f172214fca96f8227eb949b6/src/ui/main.qml#L103 This may make more sense as `forceBlurOnOutsidePress` - it's actually a hack to forcefully remove focus from the input and close the system keyboard on iOS when the user taps anywhere outside the input.

lesleyjanenorton (commented):
This is looking great! The only thing I'm noticing is this doesn't seem to be working with `VPNComboBox` which also uses `VPNInputBackground`- I think possibly because VPNInputStates is looking for `textField` which isn't available in the case of the combo box. It might be possible to do something like `property variant itemToTarget: textField` then overwrite it as necessary.

flozia:
> This is looking great! The only thing I'm noticing is this doesn't seem to be working with `VPNComboBox` which also uses `VPNInputBackground`- I think possibly because VPNInputStates is looking for `textField` which isn't available in the case of the combo box. It might be possible to do something like `property variant itemToTarget: textField` then overwrite it as necessary.

Thank you for catching this and the suggestion! This should be fixed now.

lesleyjanenorton:
That fixed it! The interaction states are visible again on VPNComboBox but I'm still seeing a few console warnings when opening the combo box that I'm not immediately sure how to triage.

flozia:
I tried to make the VPNInputStates reusable for the VPNComboBox as well, but it becomes too cumbersome and decided to add dedicated states to the ComboBox for now.

src/ui/components/forms/VPNComboBox.qml
```diff
@@ -127,4 +130,71 @@ ComboBox {
...
+                color: Color.input.hover.text
+            }
+
+            PropertyChanges {
+                target: combo.background
+                border.color: Color.input.hover.border
+                border.width: 1
+            }
+        },
+        State {
```
lesleyjanenorton, line 169: Let's move this "focused" state block to the top. Right now if you tab to combo box and you haven't already selected an option, the conditionals for "empty" will evaluate to true and the focus property changes don't get set. Otherwise this looks awesome.

lesleyjanenorton (approved):
Looks great! Thanks @flozia !

PR #1476, opened 2021-08-02
Replace `log` with `sensitive/warning/error/debug` 

mbirghan:
This PR removes all uses of `logger.log()` and replaces them.

To decide what would be an `error` and what would be a `warning` I used this heuristic:
error: everything that stops the service from behaving normally
warning: something unexpected happened but we can recover and resume normal behavior

I used `sensitive` for everything that could identify a user.

Everything else was replaced with `debug`.


bakulf (changes requested):

bakulf:
Owen likes the info() method. Owen what do you think about this PR? can you please help to review it?

src/platforms/ios/iosdatamigration.mm
```diff
@@ -117,44 +117,44 @@ void migrateKeychainData() {
...
   if (status != noErr) {
-    logger.log() << "No credentials found";
+    logger.error() << "No credentials found";
     return;
   }
 
   QByteArray data = QByteArray::fromNSData(dataNS);
 #ifdef QT_DEBUG
-  logger.log() << "Credentials:" << data;
+  logger.debug() << "Credentials:" << data;
```
oskirby, line 126: Would this be better done as `sensitive()`?

src/platforms/linux/linuxpingsender.cpp
```diff
@@ -157,22 +157,22 @@ void LinuxPingSender::rawSocketReady() {
...
-    logger.log() << "malformed IP packet:" << strerror(errno);
+    logger.error() << "malformed IP packet:" << strerror(errno);
     return;
   }
 
   // Check the ICMP packet
   struct icmphdr packet;
   if (inetChecksum(data + iphdrlen, rc - iphdrlen) != 0) {
-    logger.log() << "invalid checksum";
+    logger.error() << "invalid checksum";
```
oskirby, line 175: This should be reduced to warning() or maybe even debug(). It's an unlikely event but could occur due to network conditions.

src/platforms/windows/windowscryptosettings.cpp
```diff
@@ -82,7 +82,7 @@ bool CryptoSettings::getKey(uint8_t key[CRYPTO_SETTINGS_KEY_SIZE]) {
     return true;
   }
 
-  logger.log() << "Invalid key";
+  logger.debug() << "Invalid key";
```
oskirby, line 85: This seems like something a bit more serious than a `debug()` message. Probably worthy of a `warning()`

src/update/balrog.cpp
```diff
@@ -142,23 +142,23 @@ bool Balrog::fetchSignature(NetworkRequest* initialRequest,
...
 
-  logger.log() << "Fetching x5u URL:" << x5u;
+  logger.debug() << "Fetching x5u URL:" << x5u;
 
   NetworkRequest* x5uRequest = NetworkRequest::createForGetUrl(this, x5u, 200);
 
   connect(x5uRequest, &NetworkRequest::requestFailed,
           [this](QNetworkReply::NetworkError error, const QByteArray&) {
-            logger.log() << "Request failed" << error;
+            logger.error() << "Request failed" << error;
```
oskirby, line 155: I've been noticing that we get a lot of 404 responses when running the latest version, presumeably because there is nothing to upgrade to. Perhaps we should consider lowering the severity of this event (or carving out some exception for a 404).

oskirby (commented):
Please remove `scripts/__pycache__/generate_strings.cpython-39.pyc` it should not be included in the commit.

oskirby:
This is an enormous diff, thank you for taking it on.

oskirby (approved):

Issue #1401, opened 2021-07-22
SplitTunnel View: Implement "Apps Missing Notification"


strseb:
The Figma File describes a Notification, in case an App Goes missing:
-> i.e it was splitted but now not found 
-> it was added by "missing" and is now missing again. 

We should add that! :)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-777)

ValentinaPC:
Verified this on VPN 2.5.0 (2.202108191038) while using Windows 10.
After adding an app manually a success banner is displayed, but immediately after a second one, that mentions to add the app manually. If we repeat the steps to add it manually, both banners are displayed again.
Reopening.
![addapp1](https://user-images.githubusercontent.com/15685960/130199946-8fccdb9f-44f2-450e-ad74-19d9b4872901.gif)

strseb:
.msi is not an application. We should catch this but in the mean-time can you re-try with an .exe file?

ValentinaPC:
Yes, verified the ".exe" file also in https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1402. All good for .exe.

data-sync-user:
➤ Luminita Buruian commented:

Closing this as it has already been verified.

Thank you.

Valentina Virlics

PR #1478, opened 2021-08-02
Add alerts for Split-Tunnel 

### LINKED ISSUE #1402: Banner not displayed when adding an application to the exclusion list from App Permissions menu
#### Affected versions:
- Mozilla VPN 2.4.0 (2.202107220713);

#### Affected Platforms:
- Windows;

#### Prerequisites:
- Have an FxA account with an active subscription for Mozilla VPN;
- Have Mozilla VPN app installed on your device and in off state;
 
#### Steps to reproduce:
1. Go to Settings - App permissions;
2. Scroll down to the bottom of the screen and click on the “+Add Application” blue button;
3. Select an application that is not on the exclusion list and add it to the list;
4. Observe the screen.

#### Expected result:
- “Success! You added the app to this list.” green banner is displayed at the top of the screen. The added application is checked by default and highlighted.

#### Actual result:
- “Success! You added the app to this list.” green banner is not displayed at the top of the screen. The added application is not checked and is not highlighted. 

#### Notes:
- See the mock-ups for Split Tunneling feature: https://www.figma.com/file/k9gjzhdAisu1SHJwGX1U7j/Split-Tunneling_Settings?node-id=69%3A808.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-778)

strseb:
Hi! 
So this does 2 things: 
-> Split Logic and Styling from VPNAlert.qml
Currently the Logic what should happen is handled inside vpnalert and its colors are set by the parent, i tried to make it the opposite. So only the parent needs to know what should happen when "x" is pressed etc. 

-> Create a signal for "missing apps" added apps in appermission.cpp

Closes #1402 #1401


src/apppermission.cpp
```diff
@@ -122,22 +141,43 @@ void AppPermission::receiveAppList(const QMap<QString, QString>& applist) {
...
   endResetModel();
+
+  // In Case we removed Missing Apps during cleanup,
+  // Notify the user
+  if(removedMissingApps.count() == 0){
+    return;
+  }
+  //% "Add apps manually"
+  //: "Refers to the Apps that are Missing from the Split Tunnel list"
+  QString action = qtTrId("splittunnel.missing.action");
```
bakulf, line 169: can you use strings.yaml instead?

strseb: ✔️

src/apppermission.cpp
```diff
@@ -113,7 +115,25 @@ void AppPermission::requestApplist() {
...
+  // If that happens 
+  if (SettingsHolder::instance()->hasMissingApps()) {
+    auto missingAppList = SettingsHolder::instance()->missingApps();
+    for (const auto& appPath : missingAppList) {
+      // Check if the App Still exists, otherwise clean up.
+      if (!m_listprovider->isValidAppId(appPath)) {
+        SettingsHolder::instance()->removeMissingApp(appPath);
+        removedMissingApps.append(m_listprovider->getAppName(appPath));
+      }else{
+          applistCopy.insert(appPath, m_listprovider->getAppName(appPath));
```
bakulf, line 130: indentation

src/apppermission.cpp
```diff
@@ -122,22 +142,40 @@ void AppPermission::receiveAppList(const QMap<QString, QString>& applist) {
...
   for (auto id : keys) {
-    m_applist.append(AppDescription(id, applist[id]));
+    m_applist.append(AppDescription(id, applistCopy[id]));
   }
   std::sort(m_applist.begin(), m_applist.end());
   endResetModel();
+
+  // In Case we removed Missing Apps during cleanup,
+  // Notify the user
+  if(removedMissingApps.count() == 0){
```
bakulf, line 165: isEmpty() ?

src/ui/components/VPNAlert.qml
```diff
@@ -5,115 +5,213 @@
 import QtQuick 2.5
 import QtQuick.Controls 2.14
 import QtQuick.Layouts 1.14
-import Mozilla.VPN 1.0
 import "../themes/themes.js" as Theme
+import QtGraphicalEffects 1.0
```
bakulf, line 9: Move it above

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
...
+
+splittunnel:
+  missingAppAction:
+    value:
+      - Add apps manually
+    comment:
+      - Refers to the Apps that are Missing from the Split Tunnel list
+  missingAppOne: 
+    value:
+      - \%0 Apps missing from list
```
bakulf, line 75: "App" -> "app" ? Check with @flodolo the right string to use here.

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
...
+    comment:
+      - Refers to the Apps that are Missing from the Split Tunnel list
```
bakulf, line 77: Here as well. Most of the words should be lowercase.

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
...
+  missingAppMultiple: 
+    value:
+      - \%0 Apps missing from list.
```
bakulf, line 80: Why this % is escaped and not the others?

bakulf (approved):

bakulf (approved):

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
   useStagingTitle: Use Staging Servers
   useStagingSubtitle: Staging services are used to test features before they go live to production
   restartRequired: VPN must be restarted to take effect
+
+splittunnel:
+  missingAppAction:
```
flodolo, line 68: It would be great if we could pick a more semantic name here. Is it a button or something else?

translations/strings.yaml
flodolo: Personally I think that _apps_ should be lowercase. But that's more a question for who owns UX/content than l10n (not sure that was the question though).

With that said:
* Why the escaping? How does the string looks like once exported to .ts? 
* There should always be a comment explaining what replaces`%0`, as the linter correctly points out for `vpn.splittunnel.missingAppAddedOne`.

Having 2 strings for singular/plural is not helpful: we conveniently make it work for English and another bunch of languages, but mess with the others. We should try to find a string that works for all locales, e.g. `Apps missing from list: %0`. For the record, that's what Google suggests when it comes to plurals in WebExtensions ([reference](https://www.chromium.org/developers/design-documents/extensions/how-the-extension-system-works/i18n#TOC-Plural-form)).

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
...
+    value:
+      - \%0 Apps missing from list
+    comment:
+      - Refers to the Apps that are Missing from the Split Tunnel list
+  missingAppMultiple: 
+    value:
+      - \%0 Apps missing from list.
+    comment:
+      - Apps in the Split-Tunnel settings %1 is the amount of apps missing
+  missingAppAddedOne: Success! You added the %1 to this list.
```
flodolo, line 83: Why `%1` and not `%0`? What is `%1`? If it's the name of the app, "the NAME" doesn't make a lot of sense.

translations/strings.yaml
```diff
@@ -63,3 +63,22 @@ settingsDev:
...
+  missingAppOne: 
+    value:
+      - \%0 Apps missing from list
+    comment:
+      - Refers to the Apps that are Missing from the Split Tunnel list
+  missingAppMultiple: 
+    value:
+      - \%0 Apps missing from list.
+    comment:
+      - Apps in the Split-Tunnel settings %1 is the amount of apps missing
```
flodolo, line 82: After reading this a few times, I realized there's a missing period in the middle that makes the sentence quite impossible to understand.

flodolo (changes requested):
We should stay away from plurals, especially having 2 strings for singular/plural.

I'm also confused by the escaping (please double check the generated .ts file) and the mix of %0 and %1. For the former, if the escaping is needed, we might need extra work on the parser.

translations/strings.yaml
strseb: Good point. Let's remove the singular string. 

A line beginning with % is reserved for YAML directives, that's why it's escaped 
the exported .ts does not contain it e.g
```
<message id="vpn.splittunnel.missingAppMultiple">
        <source>%0 apps missing from list.</source>
        <oldsource>%0 Apps missing from list.</oldsource>
        <extracomment>apps in the Split-Tunnel settings. %0 is the amount of apps missing</extracomment>
    </message>
````

File: translations/strings.yaml (hunk 7 above)
Comment (strseb) [reply translations/strings.yaml]: a value beginning with % is reserved for "%TAG" and "%YAML" directive, so the parser throws if i dont escape it

File: translations/strings.yaml [hunk 11]
```diff
@@ -63,3 +63,21 @@ settingsDev:
...
+    comment:
+      - Displayed in notification as "call to action" when apps are missing from the split-tunnel settings.
+  missingAppMultiple: 
+    value:
+      - \%0 apps missing from list.
+    comment:
+      - apps in the split-tunnel settings. %0 is the amount of apps missing
+  missingAppAddedOne: 
+    value:
+      - Success! You added the %0 to this list.
```
flodolo: This sounds off to me: `Success! You added the Instagram to this list.`

Either we have app after the name, or drop the "the". I think it would make sense to also have the name between quotes, i.e. `Success! You added “%0” to this list.` (imagine the app is called "Send to"…).

translations/strings.yaml
```diff
@@ -63,3 +63,21 @@ settingsDev:
...
+  missingAppActionButton:
+    value:
+      - Add apps manually
+    comment:
+      - Displayed in a notification as "call to action" when apps are missing from the split-tunnel settings.
+  missingAppMultiple: 
+    value:
+      - \%0 apps missing from list.
+    comment:
+      - Apps listed in the split-tunnel settings. %0 is the amount of apps missing
```
flodolo, line 77: Nit: `%0 is the number of missing apps`. I would also add a bit of help for locales with complex plurals

```
- Apps listed in the split-tunnel settings. %0 is the number of missing apps.
- If necessary, this can be translated as "Apps missing from list: %0".
```

translations/strings.yaml
strseb: Good catch, it's a call-to-action on a notification. Updated the name to "missingAppActionButton" as well as the comment to reflect that

translations/strings.yaml
```diff
@@ -63,3 +63,21 @@ settingsDev:
...
+  missingAppAddedOne: 
+    value:
+      - Success! You added %0 to this list.
```
flodolo, line 80: Maybe worth putting the name between quotes? Imagine the name of the app is "Send to"…

```
- Success! You added “%0” to this list.
```

flodolo (approved):

flodolo:
TIL GitHub is awful if you're commenting in a PR while it receives updates. It seems you lost the comment, but it's still there hidden, and you realize it only after publishing the review…

PR #1488, opened 2021-08-03
Improve the Windows Killswitch implementation

### LINKED ISSUE #1370: Windows: WFP Allow rule for v6
WindowsFirewall::allowDHCPTraffic and allowTrafficTo is currently v4 only :c

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-756)

strseb:
So currently we're mimiking the Killswitch of wg-go, meaning once the `allowed ip != *` the killswitch gets disabled. Thaaats not good. 
So lets instead of block all or nothing, lets block in case ``allowed ip != *`` block all traffic to any `allowedip` unless it's from the vpn adapter. 
Also this resolves the missing v6 support todos :) 

Closes #1370


src/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -7,6 +7,8 @@
 #include "leakdetector.h"
 #include "../windowscommons.h"
 #include "../../daemon/interfaceconfig.h"
+#include "../../ipaddressrange.h"
```
oskirby, line 10: You don't need the `../../` in the include directives since `src/` is in the search path.

src/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -262,60 +266,38 @@ bool WindowsFirewall::allowTrafficForAppOnAdapter(const QString& exePath,
...
-  }
-  logger.log() << "Ruleset applied: allowDHCPTraffic";
 
-  if (!allowHyperVTraffic(MED_WEIGHT)) {
-    logger.log() << "Ruleset failed: allowHyperVTraffic";
-    disableKillSwitch();
-    return false;
-  }
-  logger.log() << "Ruleset applied: allowHyperVTraffic";
+  logger.log() << "Enableing Killswitch Using Adapter:" << vpnAdapterIndex;
```
oskirby, line 280: I think this is spelled "Enabling"

src/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -262,60 +266,38 @@ bool WindowsFirewall::allowTrafficForAppOnAdapter(const QString& exePath,
...
 
-  if (!allowTrafficForAppOnAll(getCurrentPath(), MAX_WEIGHT)) {
-    logger.log() << "Ruleset failed: allowTrafficForAppOnAll";
-    disableKillSwitch();
-    return false;
-  }
-  logger.log() << "Ruleset applied: allowTrafficForAppOnAll: MozillaVPN";
+  IPAddressRange v4CatchAll("0.0.0.0", 0, IPAddressRange::IPv4);
+  IPAddressRange v6CatchAll("::0", 0, IPAddressRange::IPv6);
+  bool hasCatchAllIP = config.m_allowedIPAddressRanges.contains(v4CatchAll) || config.m_allowedIPAddressRanges.contains(v6CatchAll);
```
oskirby, line 284: It's worth pointing out that with local network access disabled, this will result in a default route for IPv6, in which case we will call `blockAll()` regardless of whatever has been configured for IPv4. It seems like it would make more sense to always follow the `blockTrafficTo()` case, since that would also wind up including a default route.

oskirby (approved):

bakulf:
All good to me! Rebase the PR and merge it. Thanks!

PR #1490, opened 2021-08-04
Fixed Go link errors on Linux build.

swatson555:
Some Linux distros choose to configure Go differently to how it's assumed to be in the build scripts. When linking against netfilter  on the latest version of Fedora the linker would emit errors because it can't find Go's runtime object code. Linking with `-lgo` resolves these errors.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -359,6 +359,10 @@ else:linux:!android {
     QT += networkauth
     QT += dbus
 
+    system(c++ -lgo 2>&1 | grep "__go_init_main" > /dev/null) {
+        LIBS += -lgo
+    }
+
     CONFIG += c++14
 
     DEFINES += MVPN_LINUX


swatson555:
Unfortunately I don't know enough about cmake to fix this build script. My patch doesn't seem to work the way I expect. It looks like it's trying to link with `-lgo` on the Ubuntu builds when it shouldn't. I have identified the issue so if this latest patch doesn't work maybe someone else can fix this issue.

bakulf (changes requested):

bakulf:
Another option is to pass LIBS+=-lgo when running qmake. It should work. Maybe if this solution is acceptable, we can update the README.

src/src.pro
```diff
@@ -359,6 +359,10 @@ else:linux:!android {
     QT += networkauth
     QT += dbus
 
+    system(c++ -lgo 2>&1 | grep "__go_init_main" > /dev/null) {
```
bakulf, line 362: I like it, but @oskirby what do you think about this approach?

bakulf (approved):

src/src.pro
oskirby: I think this approach looks good. Maybe move this into golang.pri to keep all of the Go stuff in one place and add a comment to explain why we need this.

olerich:
This patch led to compilation problems on arch linux. 

I think the problem is that the line 
```
system(c++ -lgo 2>&1 | grep "__go_init_main" > /dev/null)
```
Is not working as intended since `c++ -lgo` produces the `__go_init_main` error for me, but `-lgo` is still added as a compiler flag. See #1811 for more details.

PR #1492, opened 2021-08-04
Add the Adjust SDK to Android

mbirghan:
I this PR I added the Adjust SDK to Android. 
Thanks to @strseb for helping :)

Parts of this PR:
- Added a `VPNApplication` to initialize adjust
- Add the Adjust SDK token as a build variable
- Add a way to call adjust functions from C++

The Adjust SDK will only be initialized and track events if the environment variable `ADJUST_SDK_TOKEN` is defined at build time. Also adjust will use the sandbox environment if the application is build for `debug` and will use the production environment when build for `release`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -49,10 +49,12 @@ apply plugin: 'kotlin-android-extensions'
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar', '*.aar'])
     implementation 'androidx.core:core-ktx:1.1.0'
-    implementation 'com.android.installreferrer:installreferrer:1.1'
+    implementation 'com.android.installreferrer:installreferrer:2.2'
     implementation 'com.android.billingclient:billing-ktx:4.0.0'
     implementation "androidx.security:security-crypto:1.1.0-alpha03"
     implementation "androidx.security:security-identity-credential:1.0.0-alpha02"
+    implementation 'com.adjust.sdk:adjust-android:4.28.2'
+    implementation 'com.google.android.gms:play-services-ads-identifier:17.0.1'
     coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:1.0.10"
 }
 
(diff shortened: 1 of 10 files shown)


android/build.gradle
```diff
@@ -144,6 +146,7 @@ android {
         targetSdkVersion 30
         versionCode  System.getenv("VERSIONCODE").toInteger()
         versionName  System.getenv("SHORTVERSION")
+        buildConfigField "String", "ADJUST_SDK_TOKEN", '"' + System.getenv("ADJUST_SDK_TOKEN") + '"'
```
strseb, line 149: I mean if anybody does not have the ADJUST_SDK_TOKEN and does a release build this feature will just silently broken. Can we maybe get an assert in the gradle-release task or in the android_package.sh that a release build requires this needs to be set? :)

android/src/org/mozilla/firefox/vpn/qt/VPNAdjustHelper.java
```diff
@@ -0,0 +1,23 @@
...
+package org.mozilla.firefox.vpn.qt;
+
+import com.adjust.sdk.Adjust;
+import com.adjust.sdk.AdjustEvent;
+
+import org.mozilla.firefox.vpn.BuildConfig;
+
+public class VPNAdjustHelper
+{
+    public VPNAdjustHelper() {}
```
strseb, line 14: Nobody creats this, lets make the contructor private :)

strseb (changes requested):

android/build.gradle
mbirghan: My idea was to make it possible to build for release without having adjust active for testing purposes. But maybe we do not need that.

strseb: we only use the release-job builds for the Playstore so they should be including everything

strseb (approved):

bakulf (approved):

bakulf:
Can you add something to the README file about the new env variable?

PR #1503, opened 2021-08-05
iOS IAPHandler re-org

birdsarah:
First PR towards Google IAP. 

This PR is mostly a no-op re-organization of iOS IAP code setting up a parent class `IAPHandler` that can be used by both `IOSIAPHandler` and `AndroidIAPHandler`.

It includes the new v3 products endpoint as part of moving `TaskIOSProducts` -> `TaskProducts`. The new endpoint returns products for all platforms and so the `addProduct` function in `IAPHandler` now includes a platform filter.

ToDo:
- [x] Baku's initial review comments
- [x] Update Products handler to filter products by platform.
- [x] Why is linting failing on an unrelated file? ~Should I just push the lint fix?~ I pushed a lint fix. I don't know why it was needed.
- [x] WASM compilation
- [x] Unit test failures

Could do:
- [ ] Setup some unit tests for IAPHandler


birdsarah:
@bakulf I think this is close to ready and would appreciate a review:
* I'm expecting that there's C++ mess-ups - lay it on me
* There's a ToDo comment because I was wondering if I could / how I could extend the `Product` struct in the IOSIAPHandler as the m_productNS is iOS specific.
* If possible can you run this on your test / dev account that has in app purchase and see if this is actually still working

bakulf (changes requested):

src/platforms/iaphandler.h
```diff
@@ -10,9 +11,8 @@
 
 class QJsonValue;
 
-class IAPHandler final : public QAbstractListModel {
+class IAPHandler : public QAbstractListModel {
   Q_OBJECT
-  Q_DISABLE_COPY_MOVE(IAPHandler)
```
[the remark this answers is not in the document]
birdsarah: I've done this. Can you tell me more about when to use or not use this macro. I read the [docs](https://doc.qt.io/qt-5/qobject.html#Q_DISABLE_COPY_MOVE) but it didn't give me enough context.

src/iaphandler.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include <QJsonValue>
+#include <QScopeGuard>
+
+#ifdef MVPN_IOS
+#  include "platforms/ios/iosiaphandler.h"
+#endif
+
+namespace {
+Logger logger(LOG_IAP, "IAPHandler");
+IAPHandler* s_instance = nullptr;
```
birdsarah, line 23: @bakulf I followed your idea below for the createInstance but I'm a little unsure about how / whether an s_instance will actually be available at runtime - and can't test :(

src/commands/commandui.cpp
```diff
@@ -351,6 +351,16 @@ int CommandUI::run(QStringList& tokens) {
     qmlRegisterType<AndroidWebView>("Mozilla.VPN", 1, 0, "VPNAndroidWebView");
 #endif
 
+    if (FeatureList::instance()->inAppPurchaseSupported()) {
```
birdsarah, line 354: I don't think this if is necessary, but I don't think it's a problem either.

src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -341,73 +341,73 @@ void AuthenticationInAppListener::finalizeSignInOrUp() {
...
-                NetworkRequest::createForGetUrl(this, redirect.toString(), 200);
-
-            connect(request, &NetworkRequest::requestFailed,
-                    [this](QNetworkReply::NetworkError error,
-                           const QByteArray& data) {
-                      logger.error()
-                          << "Failed to fetch the final redirect data" << error;
-                      processRequestFailure(error, data);
-                    });
+  connect(
```
birdsarah, line 344: I can revert the commit that made this change if necessary.

It has nothing to do with this PR but the linter was failing.

birdsarah:
@bakulf I think this PR is in good shape. See notes at top for extra details.

It would be nice to start adding tests - specifically I would have added one for the new filter in `addProduct` as it's an incremental functional change but I'd like to get this PR merged and make progress building on it with the core android work.

The addition of the DummyIAPHandler and the tweak I needed to make to the wasmnetworkrequest have made me realize that it should be possible to pop the subscribe screen on the wasm viewer if we wanted to hook that up. But that's definitely out of scope for this PR.

src/iaphandler.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+  QJsonArray products = obj["products"].toArray();
+  if (products.isEmpty()) {
+    logger.error() << "No products found";
+    return;
+  }
+
+  m_productsRegistrationState = eRegistering;
+
+  for (auto p : products) {
```
birdsarah, line 79: see comment in iosiaphandler.mm

src/platforms/ios/iosiaphandler.mm
```diff
@@ -198,88 +197,45 @@ - (void)request:(SKRequest*)request didFailWithError:(NSError*)error {
...
-
-  QJsonArray products = obj["products"].toArray();
-  if (products.isEmpty()) {
-    logger.error() << "No products found";
-    return;
-  }
-
-  m_productsRegistrationState = eRegistering;
-
-  for (const QJsonValue& value : products) {
```
birdsarah, line 273: when i moved this line to cpp file the clang compiler complained in ci (although not on my machine), so i changed it to `auto` - added a comment at relevant line above.

scripts/test_coverage.sh
```diff
@@ -10,6 +10,31 @@ REPORT_FILE=/tmp/report.html
 LANGUAGE=en
 LANG=en
```
bakulf, line 12: This is nice, but it's unrelated. Can you create a separate PR?

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: Let me fix this in a separate PR

bakulf: Done, rebase the main branch.

src/iaphandler.cpp
```diff
@@ -0,0 +1,282 @@
...
+  return IOSIAPHandler::createInstance();
+#else
+  return DummyIAPHandler::createInstance();
+#endif
+}
+
+// static
+IAPHandler* IAPHandler::instance() {
+#ifdef MVPN_IOS
+  return IOSIAPHandler::instance();
```
bakulf, line 39: What about if here you do: `return new IOSIAPHandler();` and `return new DummyIAPHandler()`. ?

src/platforms/dummy/dummyiaphandler.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "platforms/dummy/dummyiaphandler.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QCoreApplication>
+
+namespace {
+Logger logger(LOG_IAP, "DummyIAPHandler");
+DummyIAPHandler* s_instance = nullptr;
```
bakulf, line 13: This should go in the IAPHandler.cpp class.

src/platforms/dummy/dummyiaphandler.cpp
```diff
@@ -0,0 +1,46 @@
...
+}  // namespace
+
+DummyIAPHandler::DummyIAPHandler(QObject* parent) : IAPHandler(parent) {
+  MVPN_COUNT_CTOR(DummyIAPHandler);
+  Q_ASSERT(!s_instance);
```
bakulf, line 18: Move this to the parent class.

src/platforms/ios/iosiaphandler.h
```diff
@@ -0,0 +1,34 @@
...
+#define IOSIAPHANDLER_H
+
+#include "iaphandler.h"
+
+class IOSIAPHandler final : public IAPHandler {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(IOSIAPHandler)
+
+ public:
+  static IOSIAPHandler* createInstance();
```
bakulf, line 15: you can remove these 2.

src/platforms/ios/iosiaphandler.mm
```diff
@@ -198,88 +197,45 @@ - (void)request:(SKRequest*)request didFailWithError:(NSError*)error {
...
-    addProduct(value);
-  }
+// static
+IOSIAPHandler* IOSIAPHandler::createInstance() {
```
bakulf, line 225: Remove this. And use `new IOSIAPHandler()` in the parent createInstance method.

src/platforms/ios/iosiaphandler.mm
bakulf: what's the error?

bakulf (changes requested):

birdsarah:
@bakulf (not sure why I can't reply inline but)

`for (const QJsonValue& value : products) {`

> what's the error?

https://github.com/mozilla-mobile/mozilla-vpn-client/runs/3270943125#step:10:1389

src/platforms/ios/iosiaphandler.mm
birdsarah: https://github.com/mozilla-mobile/mozilla-vpn-client/runs/3270943125#step:10:1389

birdsarah:
@bakulf I backed out the unrelated changes, and merged main. 

I made the changes you requested on the CTOR/DTORs - I think this finally makes sense. Sorry for being slow on the uptake on this one.

Things seem to be passing - huzzah!

src/platforms/dummy/dummyiaphandler.cpp
```diff
@@ -0,0 +1,25 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "platforms/dummy/dummyiaphandler.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QCoreApplication>
+
+namespace {
+Logger logger(LOG_IAP, "DummyIAPHandler");
```
bakulf, line 12: You are not using this. Remove it. And remove the logger.h... or use it somewhere in the methods.

src/platforms/ios/iosiaphandler.mm
bakulf: ah! right! The reason why we don't see it in the mm file is that... we use a "different" compiler. The fix is: `for (const QJsonValue value : product) {` or without `const`.

bakulf (approved):

src/iaphandler.cpp
bakulf: This is OK! You can also write: `QJsonValue p : product`

bakulf (approved):

PR #1504, opened 2021-08-06
Add VPNTabNavigation.qml

### LINKED ISSUE #1473: Add new VPNTabbedNavigation component
To support the new UI for Multi-Hop and Advanced DNS Settings, we need to create a VPNTabbedNavigation{} component.

The component should switch views based on the selected tab button. For now, the component only needs to accommodate two tabs but it would be major extra credit to build this in such a way that it could just as easily support any number of tabs.  We can probably make use of this Qt.Quick [TabBar](https://doc.qt.io/qt-5/qml-qtquick-controls2-tabbar.html) control https://doc.qt.io/qt-5/qml-qtquick-controls2-tabbar.html.

Figma: 
https://www.figma.com/file/3ro5ivHWvcQZFbRWGQbUoS/Multi-Hop-Feature?node-id=339%3A8972
https://www.figma.com/file/If64AyDRUKkGIkxNP0LBG3/DNS-Settings?node-id=203%3A2257

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-838)

lesleyjanenorton:
This adds the VPNTabNavigation.qml component which we'll need for the new DNS and Mult-Hop UI. 

Screen Shot 2021-08-05 at 3 30 46 PM

Closes #1473

### DIFF
--- a/src/ui/components/VPNTabNavigation.qml
+++ b/src/ui/components/VPNTabNavigation.qml
@@ -79,10 +79,6 @@ Item {
         height: root.height - bar.contentHeight
         clip: true
 
-        Component.onCompleted: {
-            console.log(bar.children[1].x)
-        }
-
         onCurrentIndexChanged: PropertyAnimation {
                 target: stack
                 property: "opacity"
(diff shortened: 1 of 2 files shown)


src/ui/components/VPNTabNavigation.qml
```diff
@@ -0,0 +1,99 @@
+import QtQuick 2.0
```
bakulf, line 1: the header here

bakulf (approved):

src/ui/components/VPNTabNavigation.qml
lesleyjanenorton: done, thank you!

PR #1519, opened 2021-08-09
Rename stateError to hasError prop in the ViewServers.qml

bakulf:


flozia (approved):
Looks good — thanks for the fix!

PR #1521, opened 2021-08-09
Add a validation function for the email address

bakulf:


lesleyjanenorton (approved):
It works 🎉 ! Let's merge it! And then let's maybe think about whether we want to continuously validate the email or do it only when the user hits 'submit'.

PR #1524, opened 2021-08-10
Incremental decoder for password verification

bakulf:


src/authenticationinapp/authenticationinapp.cpp
```diff
@@ -167,3 +171,47 @@ bool AuthenticationInApp::validateEmailAddress(const QString& emailAddress) {
...
+// static
+bool AuthenticationInApp::validatePasswordLength(const QString& password) {
+  return password.length() >= PASSWORD_MIN_LENGTH;
+}
+
+bool AuthenticationInApp::validatePasswordEmail(const QString& password) {
+  Q_ASSERT(m_state == StateSignUp);
+  Q_ASSERT(m_listener);
+
+  return m_listener->emailAddress() != password;
```
oskirby, line 216: I would probably do `!m_listener->emailAddress().contains(password)` for a slightly broader check of obviously guessable passwords.

src/authenticationinapp/incrementaldecoder.cpp
```diff
@@ -0,0 +1,57 @@
...
+IncrementalDecoder::~IncrementalDecoder() {
+  MVPN_COUNT_DTOR(IncrementalDecoder);
+}
+
+IncrementalDecoder::Result IncrementalDecoder::match(QTextStream& stream,
+                                                     const QString& input) {
+  uint32_t index = 0;
+  QString prev;
+
+  while (!stream.atEnd()) {
```
oskirby, line 25: It would be nice if this search could be done a bit more efficiently (eg: binary search) since the password dictionary is sorted and kinda large. Unfortunately the encoding of the password file makes this difficult.

src/authenticationinapp/incrementaldecoder.h
```diff
@@ -0,0 +1,33 @@
...
+#define INCREMENTALDECODER_H
+
+#include <QObject>
+
+class QTextStream;
+
+// This class implements the decoder part of this
+// https://en.wikipedia.org/wiki/Incremental_encoding
+
+class IncrementalDecoder final : public QObject {
```
oskirby, line 15: It doesn't seem like this needs to be a QObject if all it's doing is wrapping up a file reader.

oskirby (approved):

src/authenticationinapp/incrementaldecoder.h
bakulf: I know, but, if I don't make it a QObject, I cannot use Q_ENUM and I cannot use the types with QTest.

PR #1530, opened 2021-08-11
DNS settings UI

lesleyjanenorton:
Screen Shot 2021-08-10 at 6 20 49 PM
Screen Shot 2021-08-10 at 6 20 57 PM

Still Needs:
- Real strings
- Backend tie-ins
  - what should happen when a setting is clicked
  - how to know if a setting is enabled
  - which tab to show when the view is opened (advanced or default). 
- Styling tweaks

Note: It would probably make sense to rename 'ViewAdvancedDNSettings.qml' but I've left as is since renaming would mean that the existing strings would have to be relocalized.

### DIFF
--- a/src/qml.qrc
+++ b/src/qml.qrc
@@ -452,5 +452,6 @@
         <file>ui/authenticationInApp/ViewAuthenticationVerificationSessionByTotpNeeded.qml</file>
         <file>ui/authenticationInApp/ViewAuthenticationFallbackInBrowser.qml</file>
         <file>ui/components/VPNTabNavigation.qml</file>
+        <file>ui/components/VPNViewDNSSettings.qml</file>
     </qresource>
 </RCC>
--- a/src/ui/components/VPNTabNavigation.qml
+++ b/src/ui/components/VPNTabNavigation.qml
@@ -14,14 +14,25 @@ Item {
     property alias tabList: tabButtons.model
     property alias stackContent: stack.children
 
+
+    Rectangle {
+        // grey divider
+        anchors.bottom: bar.bottom
+
+        anchors.left: parent.left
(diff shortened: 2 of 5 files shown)


bakulf:
This is great! But we want this for 2.6 only. For 2.5 we just want the custom DNS fields. Can we keep this in a branch?

strseb:
Niceee!
The backend for this already exists in my side branch https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1207 

In the branch exists the Setting `VPNSettings.dnsProvider` and a QEnum with the Settings values:
```
enum DnsProvider {
    Custom = 3,
    BlockTracking = 2,
    BlockAds = 1,
    BlockAll = 0,
    Gateway = -1,
  };
  Q_ENUM(DnsProvider)
```

From there i.e for the "BlockTracking" setting:  
> what should happen when a setting is clicked
```
onClicked: { VPNSettings.dnsProvider = VPNSettings.BlockTracking}
```

>how to know if a setting is enabled
```
checked/enabled: VPNSettings.dnsProvider === VPNSettings.BlockTracking
```
>which tab to show when the view is opened (advanced or default).

I would propose to do show "default" if default dns is enabled  ` VPNSettings.dnsProvider == VPNSettings.Gateway,`

lesleyjanenorton:
@strseb thank you!! I will update accordingly and ping you for review when this is closer to be being finished. 
@bakulf thank you for the heads up!! We can definitely just keep this in a branch for now. I was going off of the [spec](https://www.figma.com/file/If64AyDRUKkGIkxNP0LBG3/DNS-Settings?node-id=203%3A1773) which says 2.5.

lesleyjanenorton:
Closing since this was added in #1207

PR #1533, opened 2021-08-11
Let the controller tell the VPN-Client which features are supported at runtime

### LINKED ISSUE #1366: Cleanup: Deamon Communicate Capabilities
>What we really need is an API between the client and daemon that communicate which features are available. Such an API would clean up the Linux implementation of that quite a bit. - Owen

strseb:
I would like to add a new Type, `ControllerCapabilities` , which can hold information about what feature the controller supports, if it is available or deactivated because of an error. 
The controller has to provide this information to the Client at the Initialize stage; it can update this anytime via sending a signal to the `controllerCapabilitiesUpdated` slot. 

This gives us the opportunity to e.g. disable the split-tunnel feature if we failed to talk to the driver, and re-enable it once the issue is resolved. 

Closes #1366


src/models/controllercapabilities.cpp
```diff
@@ -0,0 +1,48 @@
+#include "controllercapabilities.h"
```
bakulf, line 1: copyright/left

src/models/controllercapabilities.h
```diff
@@ -0,0 +1,64 @@
+#ifndef CONTROLLERCAPABILITIES_H
```
bakulf, line 1: copyright/left

src/models/controllercapabilities.h
```diff
@@ -0,0 +1,64 @@
...
+class ControllerCapabilities : public QObject {
+  Q_OBJECT
+
+ public:
+  // Serialises the Obj
+  QJsonObject serialize();
+  // Takes json and populates properties
+  void read(const QJsonObject& obj);
+
+  ControllerCapabilities() {}
```
bakulf, line 19: = default

src/models/controllercapabilities.h
```diff
@@ -0,0 +1,64 @@
...
+    SupportLevel m_X = Unknown;              \
+    bool isXsupported() const { return m_X == SupportLevel::Supported; }
+
+  // Note: When adding a Capability please
+  // make sure to update the copy constructor :)
+
+#  if defined(MVPN_ANDROID)
+  // Supported by default
+  Capability(SplitTunnel, Supported);
+#  elif defined(MVPN_IOS) || defined(MVPN_OSX)
```
bakulf, line 52: MVPN_MACOS

src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -107,3 +108,22 @@ void AppTracker::gtkLaunchEvent(const QByteArray& appid, const QString& display,
...
+  if (LinuxDependencies::findCgroupPath("net_cls").isNull()) {
+    return ControllerCapabilities::SupportLevel::Unsupported;
+  }
+
+  /* Application tracking is only supported on GTK-based desktops */
+  QProcessEnvironment pe = QProcessEnvironment::systemEnvironment();
+  if (!pe.contains("XDG_CURRENT_DESKTOP")) {
+    return ControllerCapabilities::SupportLevel::Unsupported;
+  }
+  QStringList desktop = pe.value("XDG_CURRENT_DESKTOP").split(":");
```
bakulf, line 123: The daemon doesn't have knowledge of the desktop apps. This env is not set.

src/models/controllercapabilities.h
```diff
@@ -0,0 +1,64 @@
...
+    Unsupported = -2,
+  };
+  Q_ENUM(SupportLevel);
+
+#  define Capability(name, defaultValue) \
+    _cap(name, m_##name, is##name##Supported, defaultValue)
+
+#  define _cap(name, m_X, isXsupported, def) \
+    Q_PROPERTY(SupportLevel name MEMBER m_X) \
+    SupportLevel m_X = Unknown;              \
```
bakulf, line 43: I prefer a hashtable with entries, but this would work too.

bakulf (changes requested):
I like the idea, but we need to combine client and server capabilities to define the features.
The best would be to expose these capabilities to the dev-options menu too.
Basically, we could remove our `#if X #else Y` and use pref names everywhere. Similarly to what we have in firefox.

src/controllerimpl.h
```diff
@@ -90,6 +90,11 @@ class ControllerImpl : public QObject {
   void statusUpdated(const QString& serverIpv4Gateway,
                      const QString& deviceIpv4Address, uint64_t txBytes,
                      uint64_t rxBytes);
+
+  // This signal is emitted when the Controller feature set changed.
+  // For example a feature subsystem is failing, so the feature is no longer
+  // available.
+  void controllerCapabilitiesUpdated(ControllerCapabilities capabilities);
```
oskirby, line 97: I'm not sure that I see a use case for why the feature set would change at runtime. Even if something catastrophic were to occur (like the driver crashes), it ought to be rare enough that we can just report a backend failure or something similar to let the user know that something is broken.

src/models/controllercapabilities.h
```diff
@@ -0,0 +1,64 @@
...
+    Unknown = 0,
+    // Feature is supported but currently unavailable
+    // e.g: "failed to communicate with the windows-split tunnel driver"
+    Deactivated = -1,
+    // Feature is not supported by Controller / OS
+    Unsupported = -2,
+  };
+  Q_ENUM(SupportLevel);
+
+#  define Capability(name, defaultValue) \
```
oskirby, line 38: As a preference thing, I am of the opinion that all macros should be uppercase unless they are specifically trying to emulate a standard function that isn't uppercase.

Issue #1537, opened 2021-08-11
Android IAP - attach FxA UID to purchase with setObfuscatedAccountId.


birdsarah:
This will be used by FXA to verify the purchase.

We need to:

* setup guardian to return the FxA UID (not just the email address) to client
* add this to the client models/user.cpp
* use this value from the model in the play billing request with `setObfuscatedAccountId`

This work will have:

* [ ] PR on guardian-website (owner @lloan / @birdsarah) 
* [ ] PR for client (owner @birdsarah)
* [ ] PR on fxa (owner @bbangert)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-860)

birdsarah:
Update after discussing with @bbangert, we are using FxA email in `setObfuscatedAccountId` property

birdsarah:
After trying to implement I've read the docs more closely and seen that storing a plain text identifier is not an option: https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.Builder#setObfuscatedAccountId(java.lang.String)

```
Specifies an optional obfuscated string that is uniquely associated with the user's account in your app.

If you pass this value, Google Play can use it to detect irregular activity, such as many devices making purchases on the same account in a short period of time. Do not use this field to store any Personally Identifiable Information (PII) such as emails in cleartext. Attempting to store PII in this field will result in purchases being blocked. Google Play recommends that you use either encryption or a one-way hash to generate an obfuscated identifier to send to Google Play.

You can also retrieve this identifier via the Purchase object.

This identifier is limited to 64 characters.

Optional:

    To buy in-app item
    To create a new subscription
    To replace an old subscription 
```

This will have to wait until we agree on a new plan.

birdsarah:
Code that did it though is here: 312f14a81251a5071ccf1ca1b9111a944b429bf2

data-sync-user:
➤ Sarah Bird commented:

Why are you assigning me this work Pun Lam? We haven’t planned when this will be done.

data-sync-user:
➤ Sarah Bird commented:

Adrienne Davenport this is a small task that would be a good addition to improve security for 2.6.

Issue #1528, opened 2021-08-10
Android IAP Edge Case 1 - User not signed into Play Store 


birdsarah:
I get the impression that it would be unusual for a user to not be signed into the play store - especially as to get the app they need to be signed, but here goes.

Steps to reproduce (you can't really repro this as the feature is in development, this is for UX's benefit):
0. Be logged OUT of Play store on android
1. Open app on android
2. Login screen appears, sign in with an account that does not have a subscription

Expected:
3. Some kind of modal / display pops that tells the user they need to sign into Google Play

Actual:
3. The web view login screen will close and the user will just be left with the authenticating spinning wheel (because the play store cannot be queried for the subscription options if the user is signed out of the play store).

Engineering notes: 
1. Although it may be possible to create a button that opens the play store, I think just a text warning is okay as I believe it's very few users that will get into this state. They can hit "cancel" to stop the auth flow and then try again.
2. I also don't think it's worth checking for play store state before a user hits "get started" as they may already have a web subscription in which case they won't hit this edge case.

Other:
This actually might be an issue on iOS too that we haven't tested. @bakulf can you test this scenario on a test flight instance of iOS? I can't test it in prod until my subscription runs out.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-856)

data-sync-user:
➤ Jessica Cook commented:

Sharing the UX documentation here for each edge case: https://www.figma.com/file/fXfxErqwSeqQzA65jcvwuZ/IAP-Android-iOS_Variable-Pricing?node-id=153%3A6740 ( https://www.figma.com/file/fXfxErqwSeqQzA65jcvwuZ/IAP-Android-iOS_Variable-Pricing?node-id=153%3A6740|smart-link )

birdsarah:
The majority of this is done in #1541 except for a function that opens the play store.

birdsarah:
Launch play store is done in commit 315083c

birdsarah:
I used the built in error screen. Hitting "Sign Out" triggers a restart once the user has hopefully logged in. Unfortunately the "Sign Out" wording isn't ideal but didn't have another string.

I opted not to use a modal because there's state changing behind the scenes and it's messy.

<img src="https://user-images.githubusercontent.com/1796208/130343202-f3010967-1ee8-4a6b-ac74-a42364db611d.png" width=200/>

data-sync-user:
[...]

Issue #1432, opened 2021-07-27
Android IAP Client


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-741)

data-sync-user:
➤ Luminita Buruian commented:

Most of the above scenarios have been verified while running the Android IAP test suites, that can be found here ( https://testrail.stage.mozaws.net/index.php?/plans/view/47508 ) .

The scenarios that have not been validated are the ones with the Google play store purchases listed as Yes - Unacked  a purchase was made but still needs to be acknowledged, because we had no way of controlling the acknowledgement side of things.

Since the QA signoff for Android IAP has been sent, we can consider this closed. Let us know if there is anything else to consider here.

Thank you.

CC: Valentina Virlics Rebecca Billings

data-sync-user:
➤ Sarah Bird commented:

This was completed in v2.5 should not have been re-opened. cc James Dirks

data-sync-user:
➤ Sarah Bird commented:

Pun Lam Adrienne Davenport this is ridiculous. i keep putting this into done and data sync has moved it out multiple times now.

birdsarah:
Closing here too as jira sync is busted.

PR #1541, opened 2021-08-12
Android InApp Purchase

### LINKED ISSUE #961: Android: IAP Support
Might be become mandatory

https://support.google.com/googleplay/android-developer/answer/9858738

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-173)

birdsarah:
Fixes #961, #1432, #1528, #1550, #1551

The staging servers are currently configured to work with my personal dev account VPN play store, until they're hooked up to Mozilla play store builds to test this need to be signed by me. I'll update when staging servers are up.

ToDo:
* [x] Get subscription info from play store and present on subscribe screen
* [x] Manually check unknownProductRegistered handling
* [x] Have subscribe button initiate google play transaction
* [x] On payment success, get token from google and send to guardian
* [x] On payment failure, return user to subscribe screen
* [x] On `tokenValid = true`, acknowledgePurchase
* [x] On `tokenValid = false`, basic acknowledgePurchaseFailed - do better in edge case 3 (#1551)
* [x] Review ToDo's in code with....baku and jess?
* [x] Iterate through purchases to ensure that not already subscribed to a different plan. If they are, display the edge case 2 error message.

ToDo (Can be in standalone PRs):
* [x] #1528
* [x] #1550 
* [x] #1551 

Opened follow-on issues:
* [x] #1537 
* [x] #1620

Questions for:
* ~On tokenValid = true do we let user into vpn or wait for guardian to give us the new capabilities~ - we'll get capabilities from fxa.
* ~What happens if a user updates their subscription, do we handle appropriately~ - should be a no-op because FxA will handle it, have test cases for it.

### DIFF
--- a/src/platforms/android/androidutils.h
+++ b/src/platforms/android/androidutils.h
@@ -5,7 +5,7 @@
 #ifndef ANDROIDUTILS_H
 #define ANDROIDUTILS_H
 
-#include "jni.h"
+#include <jni.h>
 #include <QObject>
 #include <QString>
 #include <QUrl>
--- a/tests/auth/mocmozillavpn.cpp
+++ b/tests/auth/mocmozillavpn.cpp
@@ -150,4 +150,4 @@ void MozillaVPN::appReviewRequested() {}
 
 bool MozillaVPN::validateUserDNS(const QString&) const { return false; }
 
-void MozillaVPN::reset(bool forceInitialState) {}
+void MozillaVPN::reset(bool) {}
--- a/tests/unit/mocmozillavpn.cpp
+++ b/tests/unit/mocmozillavpn.cpp
@@ -150,4 +150,4 @@ void MozillaVPN::appReviewRequested() {}
 
 bool MozillaVPN::validateUserDNS(const QString&) const { return false; }
 
-void MozillaVPN::reset(bool forceInitialState) {}
+void MozillaVPN::reset(bool) {}


src/platforms/android/androidiaphandler.cpp
```diff
@@ -168,45 +169,140 @@ void AndroidIAPHandler::onNoPurchases(JNIEnv* env, jobject thiz) {
...
-  IAPHandler::instance()->stopSubscription();
-  IAPHandler::instance()->subscriptionFailed();
-  IAPHandler::instance()->productsRegistered();
+  IAPHandler* iap = IAPHandler::instance();
+  iap->stopSubscription();
+  emit iap->subscriptionFailed();
+}
+
+// static
+void AndroidIAPHandler::dispatchToMainThread(std::function<void()> callback) {
```
birdsarah, line 185: @strseb I copied this code from the AndroidWebView because I need to make a network request and that needs to happen in the main thread (that i wasn't in after coming back from kt code). Should this just be a generic android util. Or am I misusing this code and I don't need a timer or something else.

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,308 @@
...
+void AndroidIAPHandler::onSkuDetailsReceived(JNIEnv* env, jobject thiz,
+                                             jstring data) {
+  Q_UNUSED(thiz);
+
+  const char* buffer = env->GetStringUTFChars(data, nullptr);
+  if (!buffer) {
+    logger.error() << "onSkuDetailsReceived - failed to parse data.";
+    return;
+  }
+  QByteArray raw = QByteArray(buffer);
```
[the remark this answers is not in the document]
bakulf: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1553

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,308 @@
...
+  env->ReleaseStringUTFChars(data, buffer);
+
+  logger.debug() << "onSkuDetailsReceived - parsing raw data: " << raw;
```
bakulf, line 98: Anything "sensitive" here?

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,308 @@
...
+      "org/mozilla/firefox/vpn/InAppPurchase", "purchaseProduct",
+      "(Ljava/lang/String;Landroid/app/Activity;)V", jniString.object(),
+      appActivity.object());
+}
+
+// static
+void AndroidIAPHandler::onNoPurchases(JNIEnv* env, jobject thiz) {
+  Q_UNUSED(env)
+  Q_UNUSED(thiz);
+  // ToDo - I'm not sure when this scenario would occur
```
bakulf, line 170: Ok... this is a silly comment, but, if we use "TODO", my vim shows them highlighted ... :)

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,308 @@
...
+// static
+void AndroidIAPHandler::onPurchaseUpdated(JNIEnv* env, jobject thiz,
+                                          jstring data) {
+  Q_UNUSED(thiz);
+  const char* buffer = env->GetStringUTFChars(data, nullptr);
+  if (!buffer) {
+    logger.error() << "onPurchaseUpdated - failed to parse data.";
+    return;
+  }
+  QByteArray rawJson = QByteArray(buffer);
```
bakulf, line 205: Here too

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,308 @@
...
+            if (!json.isObject() || !json.object().contains("tokenValid")) {
+              logger.debug() << "Unexpected json returned";
+              stopSubscription();
+              emit subscriptionFailed();
+              return;
+            }
+
+            bool tokenValid = json.object()["tokenValid"].toBool();
+
+            if (tokenValid != true) {
```
bakulf, line 282: `if (!tokenValid) {`

src/platforms/android/androidiaphandler.h
```diff
@@ -0,0 +1,37 @@
...
+  static void dispatchToMainThread(std::function<void()> callback);
+
+  // Functions called via JNI
+  static void onSkuDetailsReceived(JNIEnv* env, jobject thiz, jstring data);
+  static void onNoPurchases(JNIEnv* env, jobject thiz);
+  static void onPurchaseUpdated(JNIEnv* env, jobject thiz, jstring data);
+  static void onSubscriptionFailed(JNIEnv* env, jobject thiz);
+};
+
+#endif  // DUMMYIAPHANDLER_H
```
bakulf, line 37: This is not dummy :)

bakulf (changes requested):

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -0,0 +1,293 @@
...
+        private const val TAG = "InAppPurchase"
+
+        private var instance: InAppPurchase? = null
+
+        /**
+         * TODO - Discuss / get consult from basti
+         * Because we launch IAPHandler in mozillavpn, I believe all users will be forced
+         * to have Play and be logged in irrespective of whether they need to IAP.
+         * But the dependency on billing library may have this effect anyway.
+         */
```
strseb, line 87: Given that we only publish with google play this should be fine. We already have other dependencies that break totally if the play store isn't present.

src/platforms/android/androidiaphandler.cpp
strseb: AndroidUtils is probably the right place for it now :)

strseb (approved):
awesome!

src/mozillavpn.cpp
```diff
@@ -749,6 +750,16 @@ void MozillaVPN::reset(bool forceInitialState) {
...
   m_private->m_serverData.forget();
 
+  if (FeatureList::instance()->inAppPurchaseSupported()) {
+    IAPHandler::instance()->stopSubscription();
+    // TODO
+    // Need to clear out registered products
+    // and reset initial state that's set in the
+    // header file - i'm not sure of the best way of doing this.
+    // If we don't then we can crash if e_Registering when
+    // shouldn't be.
```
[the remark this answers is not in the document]
birdsarah: Couple of things: 

1) I didn't know if you put the initialization / setting of m_productsRegistrationState in the header file for a reason and so whether we need to keep it like that.
2) If the status is still eRegistering then we aren't finished and we could be in some undefined state so probably best to start fresh than to figure out where we are? Kind of depends on the number of reasons we could hit `reset`. Your guidance is much appreciated here. Registering the products isn't that bad in the grand scheme of things I'd' say.

src/platforms/android/androidiaphandler.cpp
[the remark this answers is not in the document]
birdsarah: Thanks for this. The problem with my inexperience is that I end up copying bad patterns. The bright side it highlights opportunities?

src/platforms/android/androidiaphandler.cpp
birdsarah: no - this is just price and currency info.

src/platforms/android/androidiaphandler.cpp
birdsarah: I'll try and remember :D

src/platforms/android/androidutils.cpp
```diff
@@ -134,3 +135,71 @@ void AndroidUtils::appReviewRequested() {
...
+  timer->setSingleShot(true);
+  QObject::connect(timer, &QTimer::timeout, [=]() {
+    callback();
+    timer->deleteLater();
+  });
+  QMetaObject::invokeMethod(timer, "start", Qt::QueuedConnection);
+}
+
+// static
+QByteArray AndroidUtils::getQByteArrayFromJString(JNIEnv* env, jstring data) {
```
birdsarah, line 152: @bakulf In a now outdated comment you had asked if "raw" contained anything sensitive. 

I've now moved all those functions into util helper functions. How should I handle debug logging as the utils could be used for sensitive data in the future?

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -0,0 +1,293 @@
...
+            return
+        }
+        val billingParams = BillingFlowParams
+            .newBuilder()
+            .setSkuDetails(skuDetails)
+            // ToDo - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1537
+            // .setObfuscatedAccountId(fxaUid)
+            .build()
+        val billingResult = billingClient.launchBillingFlow(activity, billingParams)
+        val responseCode = billingResult.responseCode
```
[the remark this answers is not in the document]
birdsarah: I'm pretty sure we take care of any necessary handling in onPurchasesUpdated, but we could check and do onSubscriptionFailed to be extra careful.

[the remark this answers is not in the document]
birdsarah: https://developer.android.com/google/play/billing/integrate#launch :) you were right

birdsarah:
If you cancel a subscription through google play store then you keep access to the product through the end of the subscription period.

If you have a cancelled a subscription then it will not show up in `queryPurchasesAsync` so we cannot block, android side, showing the subscription screen. 

Source: https://developer.android.com/reference/com/android/billingclient/api/BillingClient#queryPurchasesAsync(java.lang.String,%20com.android.billingclient.api.PurchasesResponseListener)
> Only active subscriptions and non-consumed one-time purchases are returned. This method uses a cache of Google Play Store app without initiating a network request. 

As long as FxA continues to indicate that the user has a subscription through the end of the cancellation period then we will not run into trouble as we will not trigger in app billing.  We would want FxA to hold this information through the end of the cancellation period anyway for maintaining service on all devices and on `TasksAccountAndServer` check.

Have requested a testcase from Luminita to cover this.

birdsarah:
For reviewers, I wanted to highlight some state scenarios that can happen after authentication, and my intention for how the code should handle them. This is not covering when things fail mid-flow just describing the expectation for the 6 different possible initial state combinations. I have realized that I should write up some proper documentation (ladder / state diagrams) and will do so as soon as I get a moment.

After logging in, we have the FxA profile listing the user has having a subscription yes|no.
We also have the Google play store which can list purchases as:
* No - no purchases
* Yes Acked - a fully acknowledged purchase
* Yes - Un-acked - a purchase was made but still needs to be acknowledged

In short those states are:
* Google|yes-acked
* Google|yes-unacked
* Google|no
* FxA|yes
* FxA|no

And can be paired in 6 combinations. Below I list how I'm handling each one.


### Scenario A
**Start**
Google|no
FxA|no
**Action**
InApp Purchase
**End**
Google|yes
FxA|yes


### Scenario B (trivial)
**Start**
Google|no
FxA|yes
**Action**
User can log into app
**End**
Google|no
FxA|yes


### Scenario C
**Start**
Google|yes-acked
FxA|no
**Action**
Put up screen saying they're logged into wrong account
**End**
Indeterminate - user needs to log into different FxA or contact support


### Scenario D (trivial)
**Start**
Google|yes-acked
FxA|yes
**Action**
User can log into app
**End**
Google|yes-acked
FxA|yes


### Scenario E
**Start**
Google|yes-unacked
FxA|no
**Action**
Retry acknowledgement which, if purchase is valid, should shift FxA status into yes
**End**
Google|yes-acked
FxA|yes

### Scenario E
**Start**
Google|yes-unacked
FxA|yes
**Action**
Because user has access to FxA InApp Purchase flow won't be triggered and this state will remain.
This state shouldn't happen. The only way it could happen is if purchase acknowledgement failed after FxA confirmed it (network error guardian side
or google play failed to acknowledge). An un-acked purchase will be automatically cancelled by Google. On cancellation FxA will then update to show the
cancellation and the user will need to re-subscribe. This is a complicated scenario that's hard to reproduce and hard to know what the UI for it should be.
My plan is we add telemetry for this scenario and flag it to the support agents and monitor to see if it becomes a problem for users. Note that users who are 
cancelled by google will get their money back without action from us.

birdsarah:
wasm compile is failing because of new `jni.h` import in `androidutils.h`.  I don't know how to fix. (I tried a forward declaration and can't get it to work. I don't know why wasm is trying to load android utils).

birdsarah:
Help Please: I'm struggling a bit with glean events. But also it's been a long weekend and maybe it's just my eyes bleeding.

bakulf:
Unit tests load jni because we cover the Android migration component. We test how we import data from vpn v1 to v2.

src/platforms/android/androidutils.h
```diff
@@ -32,6 +33,14 @@ class AndroidUtils final : public QObject {
 
   static void appReviewRequested();
 
+  static void dispatchToMainThread(std::function<void()> callback);
+
+  static QByteArray getQByteArrayFromJString(JNIEnv* env, jstring data);
```
bakulf, line 38: About the compilation issue, it seems that `androidsharedprefs.cpp` doesn't need to include `androidutils.h`. If you remove the include, the unit-tests should compile.

bakulf: About the unit-test you also need to add `void MozillaVPN::launchPlayStore() {}` in the 2 `mocmozillavpn.cpp` files in the `tests` folder.

bakulf: Fixing the `androidsharedprefs.cpp` #include should fix the wasm compilation too.

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    override fun onQueryPurchasesResponse(
+        billingResult: BillingResult,
+        purchases: MutableList<Purchase>
+    ) {
+        if (billingResult.responseCode == BillingClient.BillingResponseCode.OK) {
+            processPurchases(purchases)
+        }
+        // TODO - Could try re-querying up to n times before declaring failure
```
bakulf, line 258: Does the documentation say to do it? If not, I guess we are fine with what you have implemented.

src/mozillavpn.h
```diff
@@ -44,17 +44,19 @@ class MozillaVPN final : public QObject {
 
  public:
   enum State {
-    StateInitialize,
```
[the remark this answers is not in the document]
bakulf: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1609

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+  if (billingResponse["code"].toInt() == -99) {
+    // The billing service was disconnected.
+    // Lets try a reset if we need a subscription.
+    // TODO - This is speculative. I put it here because I got an inappropriate
+    // launch of the "Sign in to Play Store" window. But I'm not exactly sure
+    // how to trigger a billing service disconnected and I'm not sure what the
+    // right action should be.
+    MozillaVPN* vpn = MozillaVPN::instance();
+    if (vpn->user()->subscriptionNeeded()) {
+      vpn->reset(true);
```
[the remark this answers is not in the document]
bakulf: I guess it's fine.

src/platforms/android/androidutils.cpp
bakulf: We should not log "data" in these methods. What you are doing here is perfect.

src/platforms/android/androidutils.cpp
```diff
@@ -134,3 +135,73 @@ void AndroidUtils::appReviewRequested() {
...
+
+  const char* buffer = env->GetStringUTFChars(data, nullptr);
+  if (!buffer) {
+    logger.error() << "getQJsonObjectFromJString - failed to parse data.";
+    return empty;
+  }
+  QByteArray raw(buffer);
+  env->ReleaseStringUTFChars(data, buffer);
+
+  logger.debug() << "getQJsonObjectFromJString - parsing raw data: "
```
bakulf, line 187: What about this:
1. No 'raw data' logging here.
2. if the caller of this method receives an empty object, it can log: `"fail to parse string:" << getQStringFromJString(env, data);`

src/platforms/android/androidutils.cpp
```diff
@@ -134,3 +135,73 @@ void AndroidUtils::appReviewRequested() {
...
+  QString out(buffer);
+  env->ReleaseStringUTFChars(data, buffer);
+  return out;
+}
+
+// static
+QJsonObject AndroidUtils::getQJsonObjectFromJString(JNIEnv* env, jstring data) {
+  QJsonObject empty = QJsonObject();
+
+  const char* buffer = env->GetStringUTFChars(data, nullptr);
```
bakulf, line 179: You can remove this using: `QByteArray raw(getQByteArrayFromJString(env, data));`

src/platforms/android/androidutils.cpp
```diff
@@ -134,3 +135,73 @@ void AndroidUtils::appReviewRequested() {
...
+    return QString();
+  }
+  QString out(buffer);
+  env->ReleaseStringUTFChars(data, buffer);
+  return out;
+}
+
+// static
+QJsonObject AndroidUtils::getQJsonObjectFromJString(JNIEnv* env, jstring data) {
+  QJsonObject empty = QJsonObject();
```
bakulf, line 177: You use this only as the fallback value. You can also remove this line, and when needed, simply call: `return QJsonObject();`.

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+  iap->stopSubscription();
+  emit iap->subscriptionFailed();
+}
+
+// The rest - instance methods
+
+void AndroidIAPHandler::updateProductsInfo(const QJsonArray& returnedProducts) {
+  Q_ASSERT(m_productsRegistrationState == eRegistering);
+
+  QList<QString> productsUpdated;
```
bakulf, line 237: `QStringList` (which is _exactly_ the same as `QList<Qtring>`)

src/platforms/android/androidiaphandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+  iap->stopSubscription();
+  emit iap->subscriptionFailed();
+}
+
+// static
+void AndroidIAPHandler::onSubscriptionFailed(JNIEnv* env, jobject thiz,
+                                             jstring data) {
+  Q_UNUSED(thiz);
+  QJsonObject json = AndroidUtils::getQJsonObjectFromJString(env, data);
+  logger.error() << "onSubscriptionFailed"
```
bakulf, line 225: hopefully, there are no sensitive data here and in the 3 previous methods.

bakulf (approved):
It looks good! Just a few comments.

src/featurelist.cpp
```diff
@@ -153,8 +153,7 @@ bool FeatureList::accountCreationInAppSupported() const {
 }
 
 bool FeatureList::inAppPurchaseSupported() const {
-#if defined(MVPN_IOS)
-  // Soon, android will have IAP.
+#if defined(MVPN_IOS) || defined(MVPN_ANDROID)
```
bakulf, line 156: This needs to be rebased. Apply the same change to `src/features/featureinapppurchase.h`

src/mozillavpn.cpp
```diff
@@ -698,6 +709,14 @@ void MozillaVPN::createSupportTicket(const QString& email,
   });
 }
 
+void MozillaVPN::launchPlayStore() {
```
bakulf, line 712: Wondering if this method should be included in a `#ifdef MVPN_ANDROID`. Do we need to expose it elsewhere?

src/mozillavpn.h
```diff
@@ -149,6 +151,7 @@ class MozillaVPN final : public QObject {
                                        const QString& subject,
                                        const QString& issueText,
                                        const QString& category);
+  Q_INVOKABLE void launchPlayStore();
```
bakulf, line 154: `#ifdef MVPN_ANDROID`

src/mozillavpn.h
```diff
@@ -233,7 +236,7 @@ class MozillaVPN final : public QObject {
 
   [[nodiscard]] bool setServerList(const QByteArray& serverData);
 
-  void reset(bool forceInitialState);
+  Q_INVOKABLE void reset(bool forceInitialState);
```
bakulf, line 239: why Q_INVOKABLE? I don't see where this is used.

src/networkmanager.cpp
```diff
@@ -63,6 +64,10 @@ QByteArray NetworkManager::userAgent() {
     // System data
     out << "sys:" << NetworkManager::osVersion();
 
+    if (FeatureList::instance()->inAppPurchaseSupported()) {
```
bakulf, line 67: `FeatureInAppPurchase::instance()->isSupported()`

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
birdsarah: There's no documentation for this particular case. I'll just add a logging statement in the off chance we start hitting it at least there'll be a breadcrumb.

glean/metrics.yaml
```diff
@@ -385,7 +385,43 @@ sample:
...
-      with an Apple ID that already has a subscription.
+      with an Apple / Google Play ID that already has a subscription.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1711123
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1711123#c1
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - amarchesini@mozilla.com
```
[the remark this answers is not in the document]
birdsarah: I'd like to change these all to a google group email address that we can then all choose to subscribe to - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1619

src/mozillavpn.h
birdsarah: It's used for the "we couldn't auth screen"

src/platforms/android/androidiaphandler.cpp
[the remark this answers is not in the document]
birdsarah: It works. There may be a less aggressive way for what I'm trying to achieve here but it does help the user force a restart so we can try re-acknowledging purchases.

src/platforms/android/androidiaphandler.cpp
birdsarah: No, these are all returning error codes from Android IAP. It is the Serializable `BillingResultData` that comes back.

src/platforms/android/androidutils.cpp
birdsarah: I don't think that is even necessary. If we get in the if statements for errors we log so there's no logging data just for the hell of it.

src/platforms/android/androidutils.cpp
birdsarah: It was only on the third read through that I realized you were suggesting to use my own method to help here! :facepalm:

src/mozillavpn.cpp
birdsarah: I wrapped in #ifdef.

birdsarah:
@bakulf the unit tests are compiling, but failing because of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1541/files#diff-d72ddb7294cb79350397b90bb1a42c12164d5235e1dcbf6bf859db34c595e19cR67-R69.

If I remove the if statement they pass. I don't understand enough about SimpleNetworkManager quickly and the unit test setup to know why this might cause a failure.

src/iaphandler.cpp
```diff
@@ -190,6 +194,15 @@ void IAPHandler::productsRegistrationCompleted() {
   emit productsRegistered();
 }
 
+void IAPHandler::cancelProductsRegistration() {
+  logger.debug() << "Cancel products registration";
+  beginResetModel();
+  m_products.clear();
+  m_productsRegistrationState = eNotRegistered;
+  endResetModel();
+  emit productsRegistrationFailed();
```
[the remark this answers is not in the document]
birdsarah: Done in 759bf97e

tests/auth/mocmozillavpn.cpp
```diff
@@ -146,3 +146,5 @@ void MozillaVPN::abortAuthentication() {}
 void MozillaVPN::appReviewRequested() {}
 
 bool MozillaVPN::validateUserDNS(const QString&) const { return false; }
+
+void MozillaVPN::reset(bool forceInitialState) {}
```
bakulf, line 150: remove the param name, or add `Q_UNUSED(forceInitializeState)`. Same in the other mocmozilla.cpp

bakulf:
The unit-test issue is... we need to initialize the featureList in the unit-test and we are not doing it yet (there were no needs yet).
To fix the unit-test I would do 2 things:

1. Initialize the features after creating the `QCoreApplication` object in main.cpp.
 ```
{
  // Let's initialize the feature list without having `SettingsHolder` objects around.
  SettingsHolder settingsHolder;
  FeatureList::instance()->initialize();
 }
```

2. Initialize the SettingsHolder in `TestNetworkManager::basic` after the creation of the `SimpleNetworkManager`.

PR #1542, opened 2021-08-12
Passing emailAddress to the auth URL

bakulf:


src/authenticationinapp/authenticationinapplistener.h
```diff
@@ -18,8 +18,8 @@ class AuthenticationInAppListener final : public AuthenticationListener {
   explicit AuthenticationInAppListener(QObject* parent);
   ~AuthenticationInAppListener();
 
-  void start(const QString& codeChallenge,
-             const QString& codeChallengeMethod) override;
+  void start(const QString& codeChallenge, const QString& codeChallengeMethod,
+             const QString& emailAddress) override;
```
oskirby, line 22: If the `emailAddress` is optional, it might be reasonable to give it a default value.

oskirby (approved):

PR #1543, opened 2021-08-12
Add "Excellent" string

chrisrooneys:
This is the "excellent" string needed per the latest spec 
image

### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -55,6 +55,7 @@ serversView:
 
 feedbackForm:
   chooseCategory: Choose a category
+  excellentLabel: Excellent
 
 settingsDev:
   title: Developer Options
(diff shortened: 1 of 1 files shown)


translations/strings.yaml
```diff
@@ -55,6 +55,7 @@ serversView:
 
 feedbackForm:
   chooseCategory: Choose a category
+  excellentLabel: Excellent
```
flodolo, line 58: Can we add a comment to provide more context? e.g. 

```suggestion
  excellentLabel: 
    value: Excellent
    comment: 
      - One of the answers on the feedback question about the user
      - experience with the VPN client.
```

flodolo (changes requested):

flodolo (approved):

PR #1555, opened 2021-08-14
Preparing the project for a Fedora release.

swatson555:

PR #1559, opened 2021-08-16
Revert "Preparing the project for a Fedora release. (#1555)"

bakulf:
This reverts commit 738cc4d94e8fda1ee5ba35ede22f7c8b02a60afe

The reason I want to revert this PR is that, it breaks the CI. There are issues with glean_parser, not being installed when needed.


oskirby (approved):
Thanks, I will try to find some time later today to revisit this issue and find a solution that works for building in CI and from a fresh checkout.

PR #1561, opened 2021-08-16
Fix 'flicker' on App permissions view

### LINKED ISSUE #1410: Small flicker for “Protect all apps with VPN” disabled button while accessing the App Permission screen
#### VPN version:
2.4.0 (2.202107230655)

#### Affected Platforms:
* Linux Ubuntu 20.04;

#### Prerequisites:
* Have one FxA account without subscription;
* Have the VPN installed on your machine;

#### Steps to reproduce:
1. Login to VPN with the account from prerequisites;
1. Go to Settings-> App Permissions;
1. Disable the App Permission feature;
1. Turn ON the VPN;
1. Go back to App Permissions page;
1. Observe the switch button.

#### Expected result:
* There are no display or layout issues.

#### Actual result:
* A small flicker can be seen on the button.

#### Notes:
* Not reproducing on Windows.
Captura de pantalla 2021-07-23 16UNITO-UNDERSCORE!45!UNITO-UNDERSCORE!15!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-788)
┆Reporter: Data Sync

jswinarton:
Closes #1410.

The issue was caused by a QML component that had the expandable app list sub-view set as the default state for the view. When the view was rendered, the app would see that `VPNSettings.protectSelectedApps` is `false` and fire the transition to the hidden state, which caused the screen to flicker. This solution sets the default state of the view to match the "hidden" state. (This means that the app list fades in if the view is selected and `protectSelectedApps` is true, but it's barely noticeable imo.)

### DIFF
--- a/src/ui/components/VPNExpandableAppList.qml
+++ b/src/ui/components/VPNExpandableAppList.qml
@@ -24,6 +24,12 @@ ColumnLayout {
     anchors.rightMargin: Theme.vSpacing
     spacing: Theme.vSpacing
 
+    // Ensure the inital presentation of the component is consistent with the
+    // "hidden" state. This prevents the transition from firing on the
+    // component load which appears as a "flicker".
+    opacity: 0
+    visible: false
+
     states: [
         State {
             name: "visibleAndEnabled"


lesleyjanenorton (approved):
Thanks! Looks great!

Issue #1548, opened 2021-08-12
Missing WireguardGo logs when tunnel daemon fails to start


oskirby:
When an error prevents the windows tunnel daemon from starting, we find ourselves in a position where there are no logs being reported to the backend that might explain the failure. This is caused by starting the log monitor thread only after the tunnel daemon has started successfully. The lack of visibility into the tunnel daemon and its Go code makes it very difficult to debug when things go wrong as a result.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-865)

data-sync-user:
➤ Luminita Buruian commented:

Closing this as it does not need QA.

Thank you.

PR #1563, opened 2021-08-17
Move windows ring logger into its own thread

oskirby:
If the tunnel daemon fails to start, then we can be left in the condition where we never actually started the ring logger, meaning that we might not get any logs as to the cause of the failure. To make debug easier when this occurs, we move the ring logger into its own thread so that it can run asynchronously from whatever the service startup is doing.

Closes: #1548

### DIFF
--- a/src/platforms/windows/daemon/windowstunnellogger.cpp
+++ b/src/platforms/windows/daemon/windowstunnellogger.cpp
@@ -50,8 +50,6 @@ WindowsTunnelLogger::WindowsTunnelLogger(const QString& filename,
 
 WindowsTunnelLogger::~WindowsTunnelLogger() {
   MVPN_COUNT_DTOR(WindowsTunnelLogger);
-  m_timer.stop();
-
   if (m_logdata) {
     timeout();
 
--- a/src/platforms/windows/daemon/windowstunnellogger.cpp
+++ b/src/platforms/windows/daemon/windowstunnellogger.cpp
@@ -8,8 +8,6 @@
 
 #include <QDateTime>
 
-constexpr uint32_t WINDOWS_TUNNEL_MONITOR_TIMEOUT_MSEC = 2000;
-
 /* The ring logger format used by the Wireguard DLL is as follows, assuming
  * no padding:
  *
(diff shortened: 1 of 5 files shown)


src/platforms/windows/daemon/windowstunnellogger.cpp
```diff
@@ -0,0 +1,132 @@
...
+    RINGLOG_HEADER_SIZE +
+    ((RINGLOG_MESSAGE_SIZE + RINGLOG_TIMESTAMP_SIZE) * RINGLOG_MAX_ENTRIES);
+
+namespace {
+Logger logger(LOG_WINDOWS, "tunnel.dll");
+}  // namespace
+
+WindowsTunnelLogger::WindowsTunnelLogger(const QString& filename,
+                                         QObject* parent)
+    : QObject(parent), m_logfile(filename, this), m_timer(this) {
```
bakulf, line 41: remove `this` and `this`. The m_timer can be created without passing the parent object.
It will be created and destroyed together with WIndowsTunnelLogger.

src/platforms/windows/daemon/windowstunnellogger.cpp
```diff
@@ -0,0 +1,132 @@
...
+  m_startTime = QDateTime::currentMSecsSinceEpoch() * 1000000;
+  m_logindex = -1;
+
+  connect(&m_timer, SIGNAL(timeout()), this, SLOT(timeout()));
+  m_timer.start(RINGLOG_POLL_MSEC);
+}
+
+WindowsTunnelLogger::~WindowsTunnelLogger() {
+  MVPN_COUNT_DTOR(WindowsTunnelLogger);
+  m_timer.stop();
```
bakulf, line 53: This is not needed.

src/platforms/windows/daemon/windowstunnelservice.cpp
```diff
@@ -162,56 +97,27 @@ void WindowsTunnelService::timeout() {
...
-  }
+bool WindowsTunnelService::start(const QString& configFile) {
+  logger.debug() << "Starting the tunnel service";
 
-  /* Report new messages. */
-  while (m_logindex != nextLogIndex()) {
-    processMessage(m_logindex);
-    m_logindex = (m_logindex + 1) % RINGLOG_MAX_ENTRIES;
+  m_logworker = new WindowsTunnelLogger(WindowsCommons::tunnelLogFile());
+  if (m_logworker) {
```
bakulf, line 104: `new` doesn't fail. remove this `if (m_logworker) {`

bakulf (approved):
r+ with comments. Thanks!

src/platforms/windows/daemon/windowstunnellogger.cpp
oskirby: I disagree - the reason for the use of `this` is to ensure that `m_timer` shares its thread affinity with the parent `WindowsTunnelLogger` which will get moved into a different thread shortly after the constructor returns. Otherwise, the timer will remain in the caller's thread, which is likely to block for a while as the windows services get started.

src/platforms/windows/daemon/windowstunnelservice.cpp
oskirby: Ah, the joy of virtual memory. I am far too used to treating memory allocation with suspicion :)

PR #1564, opened 2021-08-17
Preparing the project for a Fedora release.

swatson555:


bakulf:
Thanks for restoring this PR. If you see, the rpm package task fails because glean_parser is not installed.

oskirby:
Since the `glean_parser` is the last thing that's blocking us from going from git checkout to buildable sources, perhaps it might be acceptable to commit the generated glean content directly into the repository, and then maybe add some CI magic/linting to ensure that the generated content is kept up to date if we ever change the yaml files.

We can't really add a build step that runs `glean_parser`, since it's not something we can get from the official Ubuntu/Fedora repositories and at least in the case of Ubuntu/PPA, the builds are performed in containers without network access so `pip` won't work.

oskirby:
Much of this PR has now been duplicated by #1571 which was merged in commit 58900e5f9690ac89daec8174ff6fac3aa8ca2d10

birdsarah:
@swatson555 closing this as stale. But if you'd like to re-open we'd be delighted.

PR #1565, opened 2021-08-17
Rework FeatureList + Add Featurelist inspector UI

strseb:
Goals: 
Create a central model called `Feature` , which is defined by an ID and can contain metadata (like description, images and other stuff needed for our whats new component)
-> A Feature can query if it is enabled or new
-> Enabled means: "we want to ship it" && "the OS supports it" && (optional, the controller / guardian support this) 
-> New means "its not in the already-seen-list" 
-> Enabled can be overwritten by setting a flag in the feature list inspector ( seen in the dev menu)
-> Every feature is a basic QObject with signals/properties ( e.g change in the dev flags will immediately reflect in the UI)

All Features are then created & owned by FeatureList.cpp, which is now also a Qt-ListModel. 
This allows to use it in the inspector and later in the "Whats new component" using the list-filtermodel


src/featurelist.h
```diff
@@ -7,52 +7,202 @@
...
+            false,         // Can be enabled in devmode
+            true           // Check OS
+        );
+
+    QList<const Feature*> getFeatureList() const{
+        return m_featurelist;
+    }
+
+private:
+    QList<const Feature*> m_featurelist{
```
bakulf, line 194: This makes it quite complex to "add" new features. We must add an ID, a new Feature object, and add it to this list.

Can we do something similar to how the tests are registered in the unit-test? Something like:

1. in featurelist.cpp:
```
namespace {
QHash<QString, const Feature*> s_features;
}
```
2. in the CTOR of Feature:
```
Feature::Feature(const QString& id) {
  s_featureList.push(id, this);
 }
```
3. for QML we can expose things using QQmlPropertyMap (maybe)
4. For C++, we need a getter method based on strings. and to avoid errors, if the string doesn't match an existing feature, we crash the client.

src/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -278,7 +278,7 @@ bool WindowsFirewall::enableKillSwitch(int vpnAdapterIndex,
   }                                                   \
   logger.debug() << "Rule enabled:" << name << "\n";
 
-  logger.log() << "Enabling Killswitch Using Adapter:" << vpnAdapterIndex;
+  logger.debug() << "Enabling Killswitch Using Adapter:" << vpnAdapterIndex;
```
bakulf, line 281: This should be already fixed. Rebase the patch.

src/plattformsupport.cpp
```diff
@@ -0,0 +1,125 @@
+#include "plattformsupport.h"
```
bakulf, line 1: header?

src/plattformsupport.h
```diff
@@ -0,0 +1,26 @@
...
+#include <QString>
+#include <QList>
+
+// Describes what Features are Available to Each plattform
+// Only VPNFeatures that have m_checkPlattformSupport set are checked here
+class PlattformSupport
+{
+public:
+   // Checks if a Feature is Supported
+   static bool supports(const QString& featureID);
```
bakulf, line 13: What about if we move this to a single feature impl?
We could have a `features` folder with a simple `FeatureX.h` file for each feature. Each feature should say:
1. I work on this platform or not
2. I could work on this platform.. but I'm disabled because I'm not ready yet.

src/ui/resources/settings/whatsnew.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="15" viewBox="0 0 14 15" fill="none" xmlns="http://www.w3.org/2000/svg">
```
bakulf, line 1: Keep this part out. @flozia and @lesleyjanenorton are working on it.

bakulf (changes requested):

src/ui/resources/settings/whatsnew.svg
strseb: This is just the icon for the DevMenu->Dev_FeatureFlag. Any other icon prefrences ? :D

src/models/feature.h
```diff
@@ -0,0 +1,109 @@
...
+    QVariant data(int role) const;
+
+    enum ModelRoles {
+       RoleId,
+       RoleName,
+       RoleDescription,
+       RoleShortDescription,
+       RoleImagePath,
+       RoleIconPath,
+       RoleEnabled,
```
strseb, line 89: Note to me: Unused.

src/models/feature.cpp
```diff
@@ -0,0 +1,161 @@
...
+            [this]() { emit supportedChanged(); });
+  }
+  all_features.insert(m_id, this);
+
+  auto releaseVersion = VersionApi::stripMinor(aReleaseVersion);
+  auto currentVersion = VersionApi::stripMinor(APP_VERSION);
+
+  auto cmp = VersionApi::compareVersions(releaseVersion, currentVersion);
+  if (cmp == -1) {
+    // Release verion < currVersion
```
bakulf, line 45: Version

src/models/feature.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Static
+QList<Feature*> Feature::getAll(){
+    return all_features.values();
+}
+// Static
+const Feature* Feature::get(const QString& featureID){
+    if(!all_features.contains(featureID)){
+        // Todo: CRASH
```
bakulf, line 66: TODO

src/models/feature.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+    if(!m_released){
+        logger.debug() << "Not released " << m_id;
+        return false;
+    }
+    logger.debug() << "Check Support" << m_id;
+    return checkSupportCallback();
+}
+
+bool Feature::isSeen() const {
```
bakulf, line 94: Maybe we should keep this out of this pr.
What about if we land your pr without this, and we let the Whatsnew implement it. I like features as a separate thing.

src/models/feature.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return false;
+  }
+  const auto settings = SettingsHolder::instance();
+  return !settings->seenFeaturesList().contains(m_id);
+}
+
+QString Feature::Name() const {
+    return L18nStrings::instance()->tr(m_name_id);
+}
+QString Feature::Description() const {
```
bakulf, line 106: Name - > name,
Description - > description
Etc

src/models/feature.h
```diff
@@ -0,0 +1,115 @@
...
+            bool writeable = false);
+
+   public:
+    static QList<Feature*> getAll();
+    // Returns a Pointer to the Feature with id, crashes client if 
+    // feature does not exist :)
+    static const Feature* get(const QString& featureID);
+    // Unique Identifier of the Feature, used to Check
+    // Capapbilities of the Daemon/Server or if is Force-Enabled in the Dev Menu
+    const QString m_id;
```
bakulf, line 44: Make these private or protected.

src/ui/developerMenu/ViewFeatureList.qml
```diff
@@ -0,0 +1,97 @@
...
+    property string searchApps: qsTrId("vpn.protectSelectedApps.searchApps")
+
+    //% "Add Application"
+    //: Button label
+    property string addApplication: qsTrId("vpn.protectSelectedApps.addApplication")
+
+
+    VPNMenu {
+        id: menu
+        title: "Feature List"
```
bakulf, line 31: Localize?

src/ui/developerMenu/ViewFeatureList.qml
```diff
@@ -0,0 +1,97 @@
...
+import "../themes/themes.js" as Theme
+
+import org.mozilla.Glean 0.15
+import telemetry 0.15
+
+
+Item {
+    id: root
+
+    //% "Search Apps"
```
bakulf, line 20: Why this?

src/update/versionapi.cpp
```diff
@@ -152,3 +152,13 @@ int VersionApi::compareVersions(const QString& a, const QString& b) {
...
   return 0;
 }
+
+// static
+QString VersionApi::stripMinor(const QString& a) {
+  QStringList aParts;
+  for (const QString& part : a.split(".")) aParts.append(part);
+  while (aParts.length() < 3) aParts.append("0");
+  aParts.removeLast();
+  aParts.append("0");
```
bakulf, line 162: What about if there is a 1.2.3.4.5?

bakulf (changes requested):

bakulf (approved):

flodolo:
@bakulf @strseb 
Why are we making this menu localizable?
* The target (devs) will know English
* Having them translated might make things more confusing
* Most feature name don't make a lot of sense without additional context

Schermata 2021-08-25 alle 07 38 39

flodolo:
BTW, all those "In app" should probably be "In-app".

PR #1567, opened 2021-08-17
Adjust ios sdk

mbirghan:
In this PR:
- add the AdjustSdk.framework to the 3rdparty resources
- include the SDK into the build phases
- implement an adjusthelper class to call adjust functions from c++
- initialize the SDK on startup
- create a build variable to store the SDK token


src/mozillavpn.cpp
```diff
@@ -76,6 +77,8 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
 
   logger.debug() << "Creating MozillaVPN singleton";
 
+  IOSAdjustHelper::initialize();
```
bakulf, line 80: should we add a `#ifdef MVPN_IOS` ?

src/platforms/ios/iosadjusthelper.h
```diff
@@ -0,0 +1,18 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef IOSADJUSTHELPER_H
+#define IOSADJUSTHELPER_H
+
+#include <QString>
+
+class IOSAdjustHelper {
+ public:
+  IOSAdjustHelper() = default;
```
bakulf, line 12: This should not needed

src/platforms/ios/iosadjusthelper.h
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef IOSADJUSTHELPER_H
+#define IOSADJUSTHELPER_H
+
+#include <QString>
+
+class IOSAdjustHelper {
```
bakulf, line 10: final?

src/platforms/ios/iosadjusthelper.mm
```diff
@@ -0,0 +1,34 @@
...
+
+Logger logger(LOG_IOS, "IOSAdjustHelper");
+
+}  // namespace
+
+void IOSAdjustHelper::initialize() {
+
+  NSString *adjustToken = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"ADJUST_SDK_TOKEN"];
+
+  if(!adjustToken.length) {
```
bakulf, line 21: Does this mean, if adjustToken is empty? We want the opposite, right?
Plus, this is a pointer. What about if we do:

`if (adjustToken) {`

src/platforms/ios/iosadjusthelper.mm
```diff
@@ -0,0 +1,34 @@
...
+    NSString *yourAppToken = adjustToken;
```
bakulf, line 22: your? :) Can we simply pass adjustToken to ADJConfig configWithAppToken without having this extra pointer?

src/platforms/ios/iosadjusthelper.mm
```diff
@@ -0,0 +1,34 @@
...
+    NSString *environment = Constants::inProduction() ? ADJEnvironmentProduction : ADJEnvironmentSandbox;
+    ADJConfig *adjustConfig = [ADJConfig configWithAppToken:yourAppToken
+                                                environment:environment];
+    [adjustConfig setLogLevel:ADJLogLevelDebug];
+    [Adjust appDidLaunch:adjustConfig];
+  }
+}
+
+void IOSAdjustHelper::trackEvent(const QString& eventToken) {
+  ADJEvent *event = [ADJEvent eventWithEventToken:eventToken.toNSString()];
```
bakulf, line 32: Can this be called also if the token is empty?

bakulf (changes requested):
Can we use adjustSDK as a git submodule?

src/src.pro
```diff
@@ -715,7 +716,8 @@ else:ios {
             platforms/ios/ioscontroller.h \
             platforms/ios/iosdatamigration.h \
             platforms/ios/iosnotificationhandler.h \
-            platforms/ios/iosutils.h
+            platforms/ios/iosutils.h \
+            platforms/ios/iosadjusthelper.h
```
bakulf, line 720: alphabetic order

src/src.pro
```diff
@@ -704,6 +704,7 @@ else:ios {
             platforms/ios/iosglue.mm \
             platforms/ios/iosnotificationhandler.mm \
             platforms/ios/iosutils.mm \
+            platforms/ios/iosadjusthelper.mm \
```
bakulf, line 707: alphabetic order

bakulf (changes requested):

src/mozillavpn.cpp
mbirghan: Yea we should good catch!

src/platforms/ios/iosadjusthelper.mm
mbirghan: It can be called when the token is empty and it will just not track anything since Adjust is not initialized.

src/platforms/ios/iosadjusthelper.mm
mbirghan: I had it mixed up so I removed the `!`. 
Using `adjustToken.length` checks the `adjustToken` for being nil and length 0.

src/src.pro
mbirghan: I put it at the top right below `iosiaphandler.h`

src/src.pro
```diff
@@ -711,11 +712,12 @@ else:ios {
 
     OBJECTIVE_HEADERS += \
             platforms/ios/iosiaphandler.h \
+            platforms/ios/iosadjusthelper.h \
             platforms/ios/iosauthenticationlistener.h \
             platforms/ios/ioscontroller.h \
             platforms/ios/iosdatamigration.h \
             platforms/ios/iosnotificationhandler.h \
-            platforms/ios/iosutils.h
+            platforms/ios/iosutils.h \
```
bakulf, line 720: Remove this change.

bakulf (commented):
Let's see if we can remove the adjust sdk files from the repo.

3rdparty/README.md
```diff
@@ -0,0 +1,6 @@
+# 3rdparty modules documentation
+
+## AdjustSdk.framework
+Source: https://github.com/adjust/ios_sdk/releases/download/v4.29.4/AdjustSdkDynamic.framework.zip
```
bakulf, line 4: Interesting... What about if we write a script to download this package at build-time?
It can be part of the xcode_patcher.rb too. What do you think?

3rdparty/README.md
```diff
@@ -0,0 +1,6 @@
...
+Version: 4.29.4\
```
bakulf, line 5: extra slash here?

scripts/apple_compile.sh
```diff
@@ -71,6 +71,25 @@ if [[ "$OS" == "ios" ]]; then
   NETWORKEXTENSION=1
   # No web-extension for IOS
   WEBEXTENSION=
+
+  command -v wget >/dev/null 2>&1 || die "wget must be installed to download the Adjust SDK"
+  command -v unzip >/dev/null 2>&1 || die "unzip must be installed to unzip the Adjust SDK"
+  command -v sha256Sum >/dev/null 2>&1 || die "sha256sum must be installed to verify the Adjust SDK"
```
bakulf, line 77: s/sha256Sum/sha256sum/

bakulf (approved):

3rdparty/README.md
mbirghan: This is to end the line because on my preview `Documentation` would be on the same line otherwise

bakulf (approved):

PR #1586, opened 2021-08-19
Rating for ios/android

bakulf:


android/src/org/mozilla/firefox/vpn/qt/VPNAppReview.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.android.play.core.review.ReviewInfo;
+import com.google.android.play.core.review.ReviewManager;
+import com.google.android.play.core.review.ReviewManagerFactory;
+import com.google.android.play.core.tasks.Task;
+
+public class VPNAppReview {
+  private static final String TAG = "VPNAppReview";
+
+  public static void appReviewRequested(Activity activity, Context context) {
```
strseb, line 19: Does this specifically require the AppContext? 
Activity inherits from context so if not, you can just use this in `ReviewManagerFactory.create` :)

strseb (approved):

lesleyjanenorton (approved):
Looks great!

src/platforms/ios/iosutils.mm
```diff
@@ -61,3 +62,9 @@
   NSString* encodedReceipt = [receipt base64EncodedStringWithOptions:0];
   return QString::fromNSString(encodedReceipt);
 }
+
+void IOSUtils::appReviewRequested() {
+  if ([SKStoreReviewController class]) {
+    [SKStoreReviewController requestReview];
```
mbirghan, line 68: Apple does not recommend using `requestReview` on button presses as it does not always open a popup [(requestReview docs)](https://developer.apple.com/documentation/storekit/skstorereviewcontroller/2851536-requestreview)

We should rather open a link that leads to the app store like outlined [here](https://developer.apple.com/documentation/storekit/requesting_app_store_reviews) under `Manually Request a Review`

mbirghan (changes requested):
Looks good! Just one issue with the request on iOS.

mbirghan (approved):

src/platforms/ios/iosutils.mm
bakulf: Good point! We should rise up this issue with the UX team. Follow up?

PR #1588, opened 2021-08-19
Support for curl and openssl for the apple compilation

bakulf:


mbirghan (approved):
Looks good! 
Very nice improvements

PR #1590, opened 2021-08-19
add iap tracking functionality and fix android adjust environment

mbirghan:
This PR:
- fixes the android adjust environment to be according to the app environment (staging/production)
- add the iap tracking function to android

the iap function for ios is not needed as this can be called directly from the iosiaphandler without type conversions.


scripts/android_package.sh
```diff
@@ -112,9 +115,16 @@ python3 scripts/generate_glean.py || die "Failed to generate glean samples"
 printn Y "Computing the version... "
 export SHORTVERSION=$(cat version.pri | grep VERSION | grep defined | cut -d= -f2 | tr -d \ ) # Export so gradle can pick it up
 export VERSIONCODE=$(date +%s | sed 's/.\{3\}$//' )"0" #Remove the last 3 digits of the timestamp, so we only get every ~16m a new versioncode
+export ADJUST_SDK_TOKEN=$ADJUST_SDK_TOKEN # Export it even if it is not set to override system env variables
```
bakulf, line 118: `export ADJUST_SDK_TOKEN`

bakulf (approved):

PR #1597, opened 2021-08-20
Add a link to split-tunnel help page

strseb:
Note: This should not be merged until we have the right sumo article and changed the url.

Also wierd, our font seems to be failing on `we’ve`

grafik

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -449,6 +449,12 @@ void MozillaVPN::openLink(LinkType linkType) {
       url = NetworkRequest::apiBaseUrl();
       url.append("/r/vpn/subscriptionBlocked");
       break;
+    case LinkSplitTunnelHelp:
+      // TODO: This should link to a more helpful article
+      url =
+          "https://support.mozilla.org/kb/"
+          "split-tunneling-use-mozilla-vpn-specific-apps-wind";
+      break;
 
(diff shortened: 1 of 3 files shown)


lesleyjanenorton (approved):

src/mozillavpn.cpp
```diff
@@ -449,6 +449,12 @@ void MozillaVPN::openLink(LinkType linkType) {
       url = NetworkRequest::apiBaseUrl();
       url.append("/r/vpn/subscriptionBlocked");
       break;
+    case LinkSplitTunnelHelp:
+      // TODO: This should link to a more helpful article
```
bakulf, line 453: @lesleyjanenorton can you help finding the final url?

bakulf (approved):

src/mozillavpn.cpp
lesleyjanenorton: Sure!

lesleyjanenorton: Seems like `https://support.mozilla.org/en-US/kb/split-tunneling-use-mozilla-vpn-specific-apps-wind` is final but needs to be updated.

lesleyjanenorton: Running this by Brian to see how we can get something better and will follow back up.

PR #1603, opened 2021-08-20
Multi-hop UI

lesleyjanenorton:
### WIP Multi-hop UI 

Needs:
 - To use real list of Recent connections
 - To use real list of servers for Select location buttons
 - To call something when an entry location has been selected - something like `VPNController.changeEntryServer()`
 - To call something when an exit location has been selected (although maybe this will continue to be done with `VPNController.changeServer()`) 
- A few remaining UI TODOs

Main view server location button updated to show entry/exit locations
Screen Shot 2021-08-20 at 10 53 08 AM

Multi-hop tab
Screen Shot 2021-08-20 at 10 53 24 AM

Recent connections are visible at the top of the "Single-hop" server list
Screen Shot 2021-08-20 at 10 53 41 AM

Recent connections disappear when searching
Screen Shot 2021-08-20 at 11 00 00 AM

Menu title changes when in Multi-hop entry and exit server lists
[<-] Back button returns you to the main Multi-hop tab 
Screen Shot 2021-08-20 at 10 53 50 AM

Recent connections are not visible in multi-hop entry and exit server lists
[<-] Back button returns you to the main Multi-hop tab 
Screen Shot 2021-08-20 at 10 59 08 AM


src/ui/components/VPNTabNavigation.qml
```diff
@@ -62,7 +69,8 @@ Item {
                 }
 
                 contentItem: VPNBoldLabel {
-                    text: buttonLabel
+                    // Workaround since VPNl18n.tr(VPNl18n()) cannot be used as a value in a ListItem
+                    text: VPNl18n.tr(VPNl18n[tabLabelStringId])
```
lesleyjanenorton, line 73: @flozia here is the string/list hack

PR #1611, opened 2021-08-24
Move DNS-related code into a DNSHelper class

bakulf:


src/dnshelper.cpp
```diff
@@ -0,0 +1,128 @@
...
+      break;
+  }
+
+  Q_ASSERT(dnsProvider == SettingsHolder::Custom);
+
+  QString dns = settingsHolder->userDNS();
+  // User wants to use a Custom DNS, let's check that this is valid.
+  if (dns.isEmpty() || !validateUserDNS(dns)) {
+    logger.debug()
+        << "Saved Custom DNS seems invalid, defaulting to gateway dns";
```
oskirby, line 57: Overall this file is inconsistent in the capitalization of DNS, but this log message is particularly egregious. I don't have a particular preference whether we use Pascal, Camel or lowercase, but we should at least be consistent about it.

oskirby (approved):

PR #1621, opened 2021-08-25
No i18n strings for the dev-menu

bakulf:


src/ui/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -17,20 +17,22 @@ VPNFlickable {
...
-        objectName: "settingsDeveloperUnlock"
 
         anchors.top: menu.bottom
         anchors.topMargin: Theme.windowMargin
         width: parent.width - Theme.windowMargin
 
-        labelText:  VPNl18n.tr(VPNl18n.SettingsDevShowOptionTitle)
-        subLabelText: VPNl18n.tr(VPNl18n.SettingsDevShowOptionSubtitle)
+        // Do not translate this string!
+        labelText:  "Show Developer Options"
```
flodolo, line 33: I would keep all the strings in this file translatable (also considering they're already translated).

I don't expect them to change, they're clear, and it might be useful for users who accidentally enable this menu (not that it's easy).

translations/strings.yaml
```diff
@@ -64,21 +64,6 @@ feedbackForm:
...
-settingsDev:
-  title: Developer Options
-  showOptionTitle: Show Developer Options
-  showOptionSubtitle: Show or hide the developer options menu
-  useStagingTitle: Use Staging Servers
-  useStagingSubtitle: Staging services are used to test features before they go live to production
-  restartRequired: VPN must be restarted to take effect
-  featureListTitle: Feature list
-  editableFeatureList:
-    value: Editable feature list
```
flodolo, line 76: I don't think "editable" makes a lot of sense here. Maybe "Features that can be toggled" or something similar?

translations/strings.yaml
```diff
@@ -64,21 +64,6 @@ feedbackForm:
...
-    comment: Editable means that, in dev-mode we can activate/deactivate this feature
-  allFeatureList:
-    value: Feature list
-    comment: List of all the features
-
 featureNames:
```
flodolo, line 67: These are all the strings that should go.

flodolo (changes requested):
I would actually keep these strings translatable (also considering most of them are already translated).

I don't expect them to change, they're clear, and it might be useful for users who accidentally enable this menu (not that it's easy).

What I don't want exposed

src/features/featuresplittunnel.h
```diff
@@ -18,7 +17,7 @@ constexpr const char* FEATURE_SPLIT_TUNNEL = "splitTunnel";
 class FeatureSplitTunnel final : public Feature {
  public:
   FeatureSplitTunnel()
-      : Feature(FEATURE_SPLIT_TUNNEL, L18nStrings::FeatureNamesSplittunnel,
+      : Feature(FEATURE_SPLIT_TUNNEL, "split-tunnel",
```
flodolo, line 20: `Split-tunnel`?

flodolo (approved):

PR #1623, opened 2021-08-25
Better multi-hop string

bakulf:


translations/strings.yaml
```diff
@@ -93,7 +93,7 @@ featureNames:
   inAppPurchase: In app Purchase
   unauthSupport: Unauth user support form
   glean: Glean
-  multiHop: Multi Hop
+  multiHop: Multi-hop
```
flodolo, line 96: This is one of the strings that should go, but OK to land and then drop.

flodolo (approved):

PR #1624, opened 2021-08-25
Add settings to hold multihop entry server

oskirby:
This extends the `ServerData` class to hold an exit server and optional entry server, which permits the configuration of multihop tunnelling. From QML, this can be achieved by providing optional parameters to the `VPNCurrentServer.update()` or `VPNController.changeServer()` methods. When those options are omitted, only a single hop is used.

For example: `VPNController.changeServer("aq", "McMurdo Station", "ca", "North Pole")`

While we were at it, we also added a recent connections list to the `SettingsHolder` class which contains a list of up to 5 recent connections.

### DIFF
--- a/src/commands/commandselect.cpp
+++ b/src/commands/commandselect.cpp
@@ -21,11 +21,12 @@ CommandSelect::~CommandSelect() { MVPN_COUNT_DTOR(CommandSelect); }
 int CommandSelect::run(QStringList& tokens) {
   Q_ASSERT(!tokens.isEmpty());
   return runCommandLineApp([&]() {
-    if (tokens.length() != 2) {
+    if (tokens.length() < 2) {
       QTextStream stream(stdout);
-      stream << "usage: " << tokens[0] << " <server_hostname>" << Qt::endl;
+      stream << "usage: " << tokens[0] << " <server_hostname> [entry_hostname]"
+             << Qt::endl;
       stream << Qt::endl;
-      stream << "The list of <server_hostname>s can be obtained using: '"
+      stream << "The list of hostnames can be obtained using: '"
              << tokens[0].split(" ").at(0) << " servers'" << Qt::endl;
       return 1;
     }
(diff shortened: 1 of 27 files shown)


src/commands/commandselect.cpp
```diff
@@ -21,11 +21,12 @@ CommandSelect::~CommandSelect() { MVPN_COUNT_DTOR(CommandSelect); }
 int CommandSelect::run(QStringList& tokens) {
   Q_ASSERT(!tokens.isEmpty());
   return runCommandLineApp([&]() {
-    if (tokens.length() != 2) {
+    if (tokens.length() < 2) {
```
bakulf, line 24: < 2  || > 3 ?

src/controller.cpp
```diff
@@ -194,13 +194,9 @@ void Controller::activateInternal() {
...
   // node as the first element, and the entry node as the final entry.
   QList<Server> serverList = {server};
-  if (FeatureMultiHop::instance()->isSupported() &&
-      settingsHolder->multihopTunnel()) {
-    ServerData data;
-    Server hop = vpn->randomHop(data);
-    while (serverList.contains(hop)) {
-      hop = vpn->randomHop(data);
-    }
+  if (FeatureMultiHop::instance()->isSupported() && vpn->multihop()) {
```
bakulf, line 197: Maybe we can move the `isSupported()` check in the `multihop()` method.

src/models/serverdata.cpp
```diff
@@ -21,56 +21,101 @@ bool ServerData::fromSettings() {
...
+                     settingsHolder->currentServerCity(),
+                     settingsHolder->entryServerCountryCode(),
+                     settingsHolder->entryServerCity());
 
-  logger.debug() << m_countryCode << m_countryName << m_cityName;
+  logger.debug() << toString();
+  return true;
+}
+
+bool ServerData::fromString(const QString& data) {
```
bakulf, line 38: This is just for testing, right?

bakulf: Can we expose it only `#ifdef UNIT_TEST` ?

bakulf (approved):

src/controller.cpp
oskirby: I prefer to keep the stuff in the `models` directory to be just plain data containers, and to leave these kinds of conditionals in the root objects of the project. But it's not a big deal to change it either.

oskirby: Actually, it turns out that calling `isSupported()` from within a unit test results in a segmentation fault. And I only realized that after pushing. Oops.

src/models/serverdata.cpp
oskirby: I'll make a note to myself to remove this function if we don't wind up using it after merging with #1603

Issue #1680, opened 2021-08-31
Implement settings refactor


lesleyjanenorton:
**Figma Spec**
https://www.figma.com/file/iBkvpAPTSMQHeRxIrCynQi/Settings-Redesign?node-id=4%3A920

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-932)

data-sync-user:
➤ Valentina Virlics commented:

Verified this on https://mozilla-hub.atlassian.net/browse/VPN-152 ( https://mozilla-hub.atlassian.net/browse/VPN-152|smart-link ) ;

data-sync-user:
➤ Luminita Buruian commented:

Closing this as it has already been verified.

Thank you.

PR #1685, opened 2021-09-01
Add system preferences settings strings

lesleyjanenorton:
These adds strings needed for #1680.

### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -215,3 +215,7 @@ customDNSSettings:
 
 networkSettings:
   multihopTitle: Multi-hop tunnel
+
+systemPreferencesSettings:
+  systemPreferencesTitle: System preferences
+  launchVPNOnStartUpDescription: Mozilla VPN will launch when you start up your device or open the app


lesleyjanenorton:
@flodolo  Sorry for the late ask here as I realize we are past all of the freezes. We were asked to get a refactor to the settings view in that requires these strings. Is it at all feasible/realistic that these could be translated in time for release?

flodolo:
> Is it at all feasible/realistic that these could be translated in time for release?

I don't think it's realistic, mostly because of the amount of content we're asking community to translate these days is very large (we just exposed both the website and the add-on for Relay, Firefox 93 is approaching content freeze this week).

How feasible is to land the strings, and get the implementation behind some flags? Based on the number of translations done, we'll be able to chose if the flag can be flipped or not closer to release date.

translations/strings.yaml
```diff
@@ -215,3 +215,7 @@ customDNSSettings:
 
 networkSettings:
   multihopTitle: Multi-hop tunnel
+
+systemPreferencesSettings:
+  systemPreferencesTitle: System preferences
```
flodolo, line 220: Out of curiosity, why "preferences"? Firefox is moving towards "settings", unified across platforms (macOS, Linux, Windows).

lesleyjanenorton: Thank you for the feedback! I'm not sure but maybe @chrisrooneys can add context.

PR #1691, opened 2021-09-01
Removed some unused authentication codes

bakulf:


src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -513,18 +513,6 @@ void AuthenticationInAppListener::processErrorCode(int errorCode) {
       aip->requestState(AuthenticationInApp::StateStart, this);
       break;
 
-    case 129:  // Invalid phone number
```
bakulf, line 516: These 2 coes are related to SMS endpoints... not used by our implementation.

PR #1694, opened 2021-09-01
update adjust event tokens

mbirghan:
This PR updates the event tokens for the adjust sdk.

I added the needed events in the adjust dashboard. Each event is associated with an auto generated event token. 
I added a enum to store all possible event tokens and added a function to convert from event to eventToken.

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -108,6 +108,9 @@ constexpr const char* PLATFORM_NAME =
 constexpr const char* PLACEHOLDER_USER_DNS = "127.0.0.1";
 
 #if defined(MVPN_ADJUST)
+// These are the two auto-generated token from the Adjust dashboard for the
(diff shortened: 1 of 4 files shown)


bakulf (approved):

src/constants.h
```diff
@@ -107,6 +107,18 @@ constexpr const char* PLATFORM_NAME =
 
 constexpr const char* PLACEHOLDER_USER_DNS = "127.0.0.1";
 
+#if defined(MVPN_ADJUST)
+constexpr const char* ADJUST_SUBSCRIPTION_COMPLETED =
+#  if defined(MVPN_IOS)
+    "jl72xm"
```
bakulf, line 113: Write a comment about these 2 codes.

mbirghan: Done!

Issue #637, opened 2021-03-02
Android / Samsung: Fails to Confirm connection. 


strseb:
Customers are reporting problems to get a running connection on Samsung devices. 

Logs:
```
[23.02.2021 18:17:01.734] (android - AndroidController) Activation
[23.02.2021 18:17:01.734] (android - AndroidController) Prompting for VPN permission
[23.02.2021 18:17:01.969] (android - AndroidController) Transact: connected
[23.02.2021 18:17:01.969] (controller - TimerController) TimerController - Operation completed: 1 1
[23.02.2021 18:17:02.720] (networking - PingHelper) Sending a new ping. Total: 10
[23.02.2021 18:17:02.720] (networking - PingSender) PingSender send to 10.64.0.1
[23.02.2021 18:17:02.721] (controller - TimerController) TimerController - Timeout: 2
[23.02.2021 18:17:02.721] (controller - Controller) Connected from state: 3
[23.02.2021 18:17:02.721] (controller - Controller) Setting state: 3
[23.02.2021 18:17:02.721] (networking - ConnectionCheck) Starting a connection check
[23.02.2021 18:17:02.722] (controller - Controller) check status
[23.02.2021 18:17:02.722] (android - AndroidController) check status
[23.02.2021 18:17:02.722] (linux|networking - LinuxPingSendWorker) LinuxPingSendWorker - start 10.64.0.1
[23.02.2021 18:17:02.724] (linux|networking - LinuxPingSendWorker) Ping sent
[23.02.2021 18:17:02.726] (android - AndroidController) Transact:: update
[23.02.2021 18:17:02.726] (controller - Controller) Status updated
[23.02.2021 18:17:02.726] (networking - ConnectionCheck) Stopping a connection check
[23.02.2021 18:17:02.727] (networking - PingHelper) PingHelper deactivated
[23.02.2021 18:17:02.727] (networking - PingHelper) PingHelper activated for server: 10.64.0.1
[23.02.2021 18:17:03.725] (networking - PingHelper) Sending a new ping. Total: 0
[23.02.2021 18:17:03.726] (networking - PingSender) PingSender send to 10.64.0.1
[23.02.2021 18:17:03.727] (linux|networking - LinuxPingSendWorker) LinuxPingSendWorker - start 10.64.0.1
[23.02.2021 18:17:03.730] (linux|networking - LinuxPingSendWorker) Ping sent
[...]

PR #1705, opened 2021-09-02
GenericPingSender handle QProcess::ProcessError

strseb:
RE: Android-Samsung devices which fail to confirm connection: #637 
It seems on some OEM's executing ping does require root, or ping might not even be present.
We should check if the QProcess fails to start and handle it. 

The handling itself is up to debate, if we can't use ICMP ourselves and can't use the ping bin either, I would propose to just fake it for now, which is probably better than devices being falsely blocked from connecting :)

### DIFF
--- a/src/pingsender.cpp
+++ b/src/pingsender.cpp
@@ -3,11 +3,16 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "pingsender.h"
+#include "logger.h"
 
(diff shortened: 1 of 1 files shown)


bakulf (approved):
Yep, I agree. Maybe the best is to detect when this happens and stop executing the process. What do you think? Follow up?

bakulf:
Fix the compilation errors before landing the patch.

Issue #1681, opened 2021-08-31
UI vertical padding is missing in Remove Device


data-sync-user:
For some languages (German) or long device names the UI vertical padding seem to be missing;
image ( [https://user-images.githubusercontent.com/15685960/131480639-54635d99-aca7-444b-9af3-d0a08f8ae3f0.png](https://user-images.githubusercontent.com/15685960/131480639-54635d99-aca7-444b-9af3-d0a08f8ae3f0.png) )

See [https://mozilla-hub.atlassian.net/browse/VPN-114](https://mozilla-hub.atlassian.net/browse/VPN-114|smart-link)  for more info.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-933)

data-sync-user:
➤ Valentina Virlics commented:

Verified this as fixed on the last 2.5.0 (2.202109031604) VPN build, on Windows and Linux.

Postfix screenshot attached.

!padd.png|width=218,height=358!

PR #1707, opened 2021-09-02
Fix remove device popup layout

flozia:
This fixes the layout issues mentioned in #1681.

image

### DIFF
--- a/src/ui/components/VPNRemoveDevicePopup.qml
+++ b/src/ui/components/VPNRemoveDevicePopup.qml
@@ -135,6 +135,8 @@ Popup {
                 horizontalAlignment: Text.AlignHCenter
                 Layout.alignment: Qt.AlignHCenter
                 Layout.bottomMargin: Theme.vSpacing
+                Layout.leftMargin: Theme.popupMargin
+                Layout.rightMargin: Theme.popupMargin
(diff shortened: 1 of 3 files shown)


src/ui/components/VPNRemoveDevicePopup.qml
```diff
@@ -130,17 +130,25 @@ Popup {
...
-                //% "Please confirm you would like to remove\n%1."
-                text: qsTrId("vpn.devices.deviceRemovalConfirm").arg(popup.deviceName)
+                lineHeight: 22;
+                text: textString.arg(popup.deviceName)
+                wrapMode: Text.WrapAtWordBoundaryOrAnywhere
+
+                Component.onCompleted: {
+                    //: %1 is the name of the device being removed. The name is displayed on purpose on a new line.
+                    //% "Please confirm you would like to remove\n%1."
+                    textString = qsTrId("vpn.devices.deviceRemovalConfirm").replace("\n", " ")
```
flozia, line 150: I don‘t think this is optimal and the line break should be removed from the strings. Since we there is string freeze for `V2.5` this could be a temporary solution.

lesleyjanenorton (approved):
Thanks @flozia! Merge at will!

PR #1709, opened 2021-09-03
Fix xcode patcher for iOS and MacOS builds, update Readme

oskirby:
To build for iOS, we lack a script to build Qt from sources so it's only really practical to use the Qt online installer for iOS development.

We also have a problem that the `APP_ID_MACOS` and `GROUP_ID_IOS` variables don't get expanded when used by Xcode, so for code-signing to work we must set them in the `config.build_settings` dictionary for that to succeed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,11 @@ The built up will show up in `Release/Mozilla VPN.app` (relative to the root of 
 
 ### IOS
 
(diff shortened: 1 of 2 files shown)


mbirghan (approved):
Looks good!
Thank you for doing this.

bakulf (approved):

PR #1737, opened 2021-09-06
Improve the asking for review to comply with apple and google guidelines

mbirghan:
Note: I left the previous code in since we will need it in the future.


src/ui/settings/ViewGiveFeedback.qml
```diff
@@ -412,8 +412,8 @@ Item {
                 anchors.topMargin: Theme.vSpacing
                 anchors.horizontalCenter: parent.horizontalCenter
                 onClicked: {
+                    VPN.openAppStoreReviewLink();
                     feedbackStackView.push(thankYouView);
-                    VPN.appReviewRequested();
```
bakulf, line 416: This PR leaves the review code. Maybe we should remove this completely. 
Plus this needs to be discussed with Brian

mbirghan: I want to use the review code for prompts on events so I don't think we should remove it.

I will talk to Brian and we can land this in combination with the prompts.

bakulf (approved):

PR #1743, opened 2021-09-07
Wrap “WhatsNew” overview title

flozia:
Fixes longer titles not wrapping in the “WhatsNew” overview list.

| Before | After |
| --- | --- |
| text-not-wrapping | text-wrapping |


lesleyjanenorton (approved):
wrapping perfectly! One nit that is maybe not caused by this patch: do some of the icon/labels in `VPNSettingsItem` look just _slightly_ out of alignment?

flozia:
You are right — I‘m seeing this now as well. I checked this on main and this is not introduced by this PR. I’ll create a separate issue.

PR #1745, opened 2021-09-07
iOS share logs screen

mbirghan:


strseb:
As the android stuff merged you probably can remove the "Draft" :)

Issue #1686, opened 2021-09-01
The VPN app crashes when installed on an Android 8 tablet


data-sync-user:
**VPN version:**

* 2.5.0 (16303750);

**Affected Platforms:**

* Android 8 on a Pixel C tablet;

**Prerequisites:**

* Have the correct APK to install on your device
**Steps to reproduce:**
1. Install the app ou your device;
2. Open the VPN app;

**Expected result:**

* The VPN app opens up successfully on the “Get Started” screen;

**Actual result:**

* The VPN app starts to open and then it crashes, displaying a message that “Mozilla VPN has stopped”;

**Notes:**

* I have added a video of the behavior, together with the .apk

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-935)

data-sync-user:
➤ Basti (Sebastian) Streich commented:

Luminita Buruian  - Hi! Are you able to get adb logs for that crash?

data-sync-user:
➤ Luminita Buruian commented:

Basti (Sebastian) Streich Unfortunately no, because I cannot open the app to go to the Help > View log menu.

By adb logs you mean through Logcat command-line tool (found something online)? Unfortunately I have never done that in the past and I am not sure how to do it.

Thank you.

data-sync-user:
➤ Basti (Sebastian) Streich commented:

I’m sorry, yes i would need the logs from adb logcat

It’s a bit tricky to setup here is a good artice to get adb installed https://www.xda-developers.com/install-adb-windows-macos-linux/ ( https://www.xda-developers.com/install-adb-windows-macos-linux/|smart-link ) 

Once that is set-up you can enter adb logcat to print the devices logs. It would be awesome if you could send me that output 🙂

data-sync-user:
➤ Luminita Buruian commented:

Thank you for the article, it was extremely helpful 🙂 .

I have managed to get the logs, I tried to only get the ones after triggering the app and the crash. I hope they are what you need. Please see them attached.

[^mozilla VPN crash logs.txt]

Please let me know if this helps and you need anything else.

CC: Madalin Cotetiu Valentina Virlics

data-sync-user:
➤ Luminita Buruian commented:

Hello,

I have checked the latest build from main, containing the Init IAP handler lazy to handle onCreateRace #1702 but the VPN app still crashes on the Android 8 Pixel C tablet.

[...]

PR #1751, opened 2021-09-07
Do not create an IAP Handler Obj

strseb:
Crash reason for https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1686
```
QAndroidJniObject javaClass(CLASSNAME);
```
This line creates an instance of CLASSNAME. But as this does not have a default constructor it is null; jni causes a crash on ```env->GetObjectClass(javaClass.object<jobject>()); ```

### DIFF
--- a/src/platforms/android/androidiaphandler.cpp
+++ b/src/platforms/android/androidiaphandler.cpp
@@ -71,12 +71,14 @@ void AndroidIAPHandler::maybeInit() {
         {"onSkuDetailsReceived", "(Ljava/lang/String;)V",
          reinterpret_cast<void*>(onSkuDetailsReceived)},
     };
-    QAndroidJniObject javaClass(CLASSNAME);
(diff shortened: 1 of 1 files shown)


birdsarah (approved):
Seems good.

If we error out here though, what are the downstream consequences?

In most cases we won't need an IAPHandler so obviously want to fail gracefully. If we do need IAP, will they need to restart? Is this just something that could happen occasionally or was it repeatable on your device?

strseb:
This is imho super rare. QA only has one device where this happens (but there 100%). 
Now currently if we would really run into this, the client would crash on the first IAP callback, as they are not defined and we dont have a global jni error handler.

birdsarah:
Totally agree on the fix. I just wanted context on whether we should do further handling. I don't have enough android experience to have a feel for probabilities. Thanks.

bakulf (approved):

PR #1752, opened 2021-09-07
Add a Link for the Inspector in the Dev Menu

strseb:
Small nit: I never can seem to remember the right port. Can we have a link in the Dev-Settings for it? :)


src/ui/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -72,9 +71,27 @@ VPNFlickable {
         imageRightSrc: "../resources/chevron.svg"
         onClicked: stackview.push("../developerMenu/ViewFeatureList.qml")
     }
+    VPNExternalLinkListItem {
+        id:inspectorLink
+        visible: stagingServer.isChecked && !restartRequired.visible
```
oskirby, line 76: I think the intention here is that we only want the link to be visible when the inspector is actually running, which we can get directly with `VPN.stagingMode`. Making use of `restartRequired.visible` leads to a bit of a convoluted dependency graph.

oskirby (approved):

PR #1758, opened 2021-09-07
First experiments with Qt6

bakulf:


src/platforms/android/androidappimageprovider.cpp
```diff
@@ -19,8 +18,16 @@ Logger logger(LOG_CONTROLLER, "AndroidAppImageProvider");
 AndroidAppImageProvider::AndroidAppImageProvider(QObject* parent)
     : QQuickImageProvider(QQuickImageProvider::Image,
                           QQmlImageProviderBase::ForceAsynchronousImageLoading),
-      QObject(parent) {
+#if QT_VERSION < 0x060000
```
oskirby, line 21: This is okay for as long as Qt6 is in beta, but once its properly released, it would be acceptable to assume a specific version of Qt for any source files that are Android-specific and to drop support for Qt5 in order to clean this up.

scripts/android_package.sh
```diff
@@ -75,16 +75,15 @@ if ! [ -d "src" ] || ! [ -d "linux" ]; then
   die "This script must be executed at the root of the repository."
 fi
 
-if ! [ -d "$QTPATH/android/bin/" ]; then
+if [ -d "$QTPATH/android/bin/" ]; then
+  QTPATH=$QTPATH/android
```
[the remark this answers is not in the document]
bakulf: This works fine. These 2 lines are to support back compatibility.

src/platforms/android/androidappimageprovider.cpp
bakulf: This code has been moved to a separate class, but the approach has not changed: we need to treat the parent object differently between Qt5 and Qt6.

oskirby (approved):

PR #1767, opened 2021-09-08
fix android adjust tracking

mbirghan:


android/AndroidManifest.xml
```diff
@@ -12,6 +12,7 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"/>
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
```
strseb, line 15: Can you explain why more permissions? The sdk mentions we need this only 
```If you are not targeting the Google Play Store (you need it)``

File: android/AndroidManifest.xml [hunk 2]
```diff
@@ -26,6 +27,14 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
         android:extractNativeLibs="true"
         android:theme="@style/AppTheme"
         android:icon="@mipmap/vpnicon">
+        <receiver
+            android:name="com.adjust.sdk.AdjustReferrerReceiver"
+            android:permission="android.permission.INSTALL_PACKAGES"
+            android:exported="true" >
+            <intent-filter>
+                <action android:name="com.android.vending.INSTALL_REFERRER" />
```
strseb, line 35: Hasn't this been disabled last year?

strseb (changes requested):
Code looks fine buuut i have questions :D

android/AndroidManifest.xml
mbirghan: I oriented myself at an example they had but we can probably remove that.

android/AndroidManifest.xml
mbirghan: Yep I read the documentation wrong. Good catch :)

strseb (approved):

PR #1769, opened 2021-09-08
IPMonitor - a simple tool to see what is going on

bakulf:


birdsarah (changes requested):
Can you add to README or other docs just a couple of lines:
- how to use ipmonitor (e.g. do you need to use the -j option you've added for command server to get parsable input?)
- when you might want to
- clarifying that this is a dev tool that isn't packaged with the vpn client (the name is a little "i'm watching you" if it's not clear that this is a dev only tool not part of the client).

birdsarah (approved):

.gitignore
```diff
@@ -119,3 +119,7 @@ macos/pkg/Resources/*.lproj
 AdjustSdkDynamic.framework.zip
 AdjustSdkDynamic/
 AdjustSdk.framework/
+
+# ipmonitor
+tools/ipmonitor/ipmonitor
+tools/ipmonitor/ipmonitor.exe
```
oskirby, line 125: You got my hopes up for a Windows build

tools/ipmonitor/README.md
```diff
@@ -0,0 +1,98 @@
+# Mozilla VPN - IPMonitor
+
+IPMonitor is a simple libpcap app to see which packages are sent and received
```
oskirby, line 3: The correct networking terminology here is "packets" when referring to network-layer unit of data being sent and received.

tools/ipmonitor/README.md
```diff
@@ -0,0 +1,98 @@
...
+outside the VPN tunnel. This is meant to be used as a debugging/testing tool
+and it's not included in the MozillaVPN app bundle.
+
+[![asciicast](https://asciinema.org/a/HZVQMHfQew3cm0lYUsLVZ78AT.png)](https://asciinema.org/a/HZVQMHfQew3cm0lYUsLVZ78AT)
```
oskirby, line 7: Are asciinema.org links stable? If this is just a GIF or PNG then we should commit it to this repository along with the README so that it can't be taken down or leave a broken link in the future.

tools/ipmonitor/README.md
```diff
@@ -0,0 +1,98 @@
...
+Which interface do you want to use?
+```
+
+At this point you have to choose the network interface you want to monitor.
+`en0` for MacOS and a `wlp*` interface on linux could be a good starting point.
+
+The app has this simple UI:
+```
+Total: 832 - VPN: 539 - LAN: 69 - OUT: 224
+D: < - Src:   192.168.0.128 - Dest:   193.32.249.69
```
oskirby, line 81: I would recommend breaking the `D:` line into multiple lines, because the addresses and direction can change too rapidly for a user to see when there is significant traffic. The lines I would recommend are (in whatever order seems most appropriate):
- Transmitted packets via VPN
- Received packets via VPN
- Transmitted leaks (packets transmitted on LAN interface to non-VPN destination address)
- Received leaks (packets received on LAN interface from non-VPN source address)
- Transmitted packets via LAN to local destinations (optional)
- Received packets via LAN from local sources (optional)

Such a format might look like:
```
Total: 832 - VPN: 539 - LAN: 69 - OUT: 224
VPN  Loc: Netherlands, Amsterdam
VPN  Tx: 192.168.0.128 --> 193.32.249.69
VPN  Rx: 193.32.249.69 --> 192.168.0.128
LEAK Tx: 192.168.0.128 --> 8.8.8.8
LEAK Rx: 8.8.8.8       --> 192.168.0.128
```

strseb (approved):

tools/ipmonitor/main.cpp
```diff
@@ -0,0 +1,428 @@
...
+#define IP_OFFMASK 0x1fff /* mask for fragmenting bits */
+  u_char ip_ttl;          /* time to live */
+  u_char ip_p;            /* protocol */
+  u_short ip_sum;         /* checksum */
+  in_addr ip_src, ip_dst; /* source and dest address */
+#define IP_HL(ip) (((ip)->ip_vhl) & 0x0f)
+#define IP_V(ip) (((ip)->ip_vhl) >> 4)
+};
+
+class O final : public QTextStream {
```
oskirby, line 48: I dislike the single-character name for this class, could you please name it to something that reflects its purpose? `Output` or `Console` or something?

tools/ipmonitor/README.md
oskirby: If this is too complicated, then at least please have a separate line for transmit and receive.

oskirby (changes requested):

.gitignore
bakulf: In theory we can have it for windows too! Maybe after v2.5 we can update this code and implement it using the pcap for windows too.

tools/ipmonitor/README.md
bakulf: The issue here is that a single packet has a precise direction A -> B or B -> A. What you are suggesting here is a session.

oskirby (approved):

PR #1779, opened 2021-09-09
Take the Apple token from the env variables too

bakulf:


scripts/apple_compile.sh
```diff
@@ -24,9 +24,12 @@ helpFunction() {
   print N "By default, the project is compiled in release mode. Use -d or --debug for a debug build."
   print N "Use -n or --networkextension to force the network-extension component for MacOS too."
   print N ""
+  print N "If MVPN_ADJUST_TOKEN env is found, this will be used at compilation time."
```
mbirghan, line 27: Can we maybe call it `MVPN_ADJUST_TOKEN_IOS` to make clear that it is the iOS token?

mbirghan (changes requested):
Looks good!
Can we also add this option to the README.

mbirghan (approved):
Great! 
If we want we could add this as an option to the README.

Issue #1673, opened 2021-08-31
Browser navigation is not working if multi-hop is enabled


data-sync-user:
**VPN version:**

* 2.5.0 (2.202108310209);

**Affected Platforms:**

* Windows 10;

**Preconditions:**

* Have VPN installed;
* Have an account with subscription;
* Enable the multi-hop

**Steps to reproduce:**
1. Open the VPN and sign in with the account from prerequisites; 
2. Choose entry and exit location;
3. Turn the VPN on;

**Expected result:**

* The internet connection is working;

**Actual result:**

* The internet connection is broken;

**Note:**

* tried different browsers;
* zoom/skype calls work;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-930)

data-sync-user:
➤ Owen Kirby commented:

I seem to be able to reproduce this now on main, the problem appears to be related to a strange choice of DNS server. When running nslookup my Windows machine appears to be using the client’s assigned IPv6 address as the DNS resolver instead of the Mullvad server. But otherwise, IPv4 and IPv6 connectivity to the internet appears to be working fine, which would explain why zoom and skype continue to work.

data-sync-user:
➤ Owen Kirby commented:

I am fairly certain that the problem has to do with calling WindowsFirewall::enableKillSwitch() for each hop of a multihop connection, leading to some strangeness going on in the Windows firewall. We see a smoking gun in log messages like the following:

Debug: (Windows - WindowsFirewall) Requesting to allow Traffic to:   : 53

I think the fix is to split the firewall kill switch into two parts: one that sets per-interface firewall rules to be called in tandem with WireguardUtilsWindows::addInterface() and a second method to set per-peer firewall rules to be called in tandem with WireguardUtilsWindows::updatePeer()

However, it’s getting late, so I will probably have to finish this tomorrow.

data-sync-user:
➤ Rebecca Billings commented:

This was fixed here: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1791 ( https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1791|smart-link )

data-sync-user:
➤ Maria Berlinger commented:

Hello,

This is verified as fixed on Mozilla VPN 2.5.0 (2.202109111821) using Windows 10.

Issue #1526, opened 2021-08-10
Eliminate the use of WgQuickProcess


oskirby:
Presently, we make use of the `WgQuickProcess` to build a `wg-quick` compatible configuration file, which is then passed into `Daemon::run()` to handle anything that isn't provided by `WireguardUtils`, `IPUtils` or `DNSUtils`. When used to bring up the interface, this includes the user's private key in plaintext, which could be exposed to other users on the system with admin access.

After the refactoring necessary for multihop in PR #1376, this config file is now only used for:

* DNS configuration (MacOS and Windows).
* ifup and address configuration on Windows.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-854)
┆author: Data Sync

data-sync-user:
➤ Andrea Marchesini commented:

Owen Kirby do you think we can close this bug?

data-sync-user:
➤ Owen Kirby commented:

Yeah, we can close this. Even though we still have the WgQuickProcess class kicking around, the reason for removing it is no longer applicable as we don’t have to write the config file to disk anymore.

data-sync-user:
➤ Owen Kirby commented:

We no longer write the wireguard config file to disk when starting the tunnel service on Windows, so the security risk posed by this issue no longer exists.

PR #1791, opened 2021-09-09
Implement DNS utils for Windows

oskirby:
When relying on the Wireguard Go module to configure DNS, we find ourselves with a chicken-and-egg bug in Windows when multihop is enabled. The DNS server lists need to be provided when creating the interface, but that is typically done for the entry server, however, we aren't given the DNS configuration until we configure the exit server, which is configured last.

To resolve the issue, we implement the `DnsUtils` class for Windows and remove the DNS setup from the Go module, which conveniently takes us one step closer to completing #1526

Closes: #1673

### DIFF
--- a/src/platforms/windows/daemon/dnsutilswindows.cpp
+++ b/src/platforms/windows/daemon/dnsutilswindows.cpp
@@ -1,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "dnsutilswindows.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QScopeGuard>
+
+#include <windows.h>
+#include <iphlpapi.h>
+
+namespace {
+Logger logger(LOG_WINDOWS, "DnsUtilsWindows");
+}
+
+DnsUtilsWindows::DnsUtilsWindows(QObject* parent) : DnsUtils(parent) {
+  MVPN_COUNT_CTOR(DnsUtilsWindows);
+
+  logger.debug() << "DnsUtilsWindows created.";
+}
+
+DnsUtilsWindows::~DnsUtilsWindows() {
+  MVPN_COUNT_DTOR(DnsUtilsWindows);
+  restoreResolvers();
+  logger.debug() << "DnsUtilsWindows destroyed.";
+}
+
+bool DnsUtilsWindows::updateResolvers(const QString& ifname,
+                                      const QList<QHostAddress>& resolvers) {
+
+  // Lookup the interface GUID
+  NET_LUID luid;
+  GUID guid;
+  if (ConvertInterfaceAliasToLuid((wchar_t*)ifname.utf16(), &luid) != 0) {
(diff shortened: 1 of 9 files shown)


oskirby:
It seems that `SetInterfaceDnsSettings()` is only supported on sufficiently new builds of Windows 10, so this PR actually breaks the VPN for older versions of Windows. In fact, it doesn't even start the daemon correctly because the dynamic linker just explodes before running anything.

oskirby:
For windows builds 1809 and earlier, it seems that the only way to configure DNS resolvers is to use `netsh`. So, if `GetProcAddress()` fails to lookup the native win32 API to configure DNS, we can provide a fallback that uses `netsh` in a `QProcess()` to do the same.

One possible piece that's missing in this approach is the configuration of the DNS search domains. In the Wireguard Go module, this is done by editing the registry, but it's not clear if we need to do this when we are only configuring the root domain.

This fallback path was tested in a VM running windows 10 build 1703.

src/platforms/windows/daemon/dnsutilswindows.cpp
```diff
@@ -0,0 +1,179 @@
...
+  if (ConvertInterfaceAliasToLuid((wchar_t*)ifname.utf16(), &luid) != 0) {
+    logger.error() << "Failed to resolve LUID for" << ifname;
+    return false;
+  }
+  m_luid = luid.Value;
+
+  logger.debug() << "Configuring DNS for" << ifname;
+  if (m_setInterfaceDnsSettingsProcAddr == nullptr) {
+    return updateResolversNetsh(resolvers);
+  } else {
```
bakulf, line 52: No `else` after a `return`.

src/platforms/windows/daemon/dnsutilswindows.cpp
```diff
@@ -0,0 +1,179 @@
...
+  return netsh.exitCode() == 0;
+}
+
+bool DnsUtilsWindows::restoreResolvers() {
+  if (m_luid == 0) {
+    return true;
+  }
+
+  QList<QHostAddress> empty;
+  if (m_setInterfaceDnsSettingsProcAddr != nullptr) {
```
bakulf, line 174: In the previous check, the order is the opposite. it was `if (a == nullptr)...` No needs to change.

src/platforms/windows/daemon/dnsutilswindows.cpp
```diff
@@ -0,0 +1,179 @@
...
+    return updateResolversWin32(empty);
+  } else {
```
bakulf, line 176: No `else` after a `return`

bakulf (approved):

src/platforms/windows/daemon/dnsutilswindows.h
```diff
@@ -0,0 +1,35 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef DNSUTILSWINDOWS_H
+#define DNSUTILSWINDOWS_H
+
+#include "daemon/dnsutils.h"
+
+#include <QHostAddress>
+#include <QMap>
```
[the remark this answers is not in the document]
oskirby: That is a copy paste error that made it's way here from `dnsutilsmacos.h`. Sorry.

PR #1792, opened 2021-09-10
Don't use Unicode character to create space between alert strings

lesleyjanenorton:


oskirby:
Maybe this is a silly question, but why are we using unicode at all for this? Surely a good old fashion ASCII space should do the job?

bakulf (approved):

src/ui/components/VPNAlert.qml
```diff
@@ -186,7 +186,7 @@ Rectangle {
             anchors.verticalCenter: parent.verticalCenter
 
             Label {
-                property var whiteSpaceCharacter: Qt.platform.os === "windows" ? " " : "\u0001" 
+                property var whiteSpaceCharacter: Qt.platform.os === "windows" || Qt.platform.os === "android" ? " " : "\u0001"
```
bakulf, line 189: Why do we want to use 0001? That is a control simbol and not a whitespace.

bakulf (changes requested):

lesleyjanenorton:
@oskirby you would think but for whatever reason it doesn't work on apple devices

lesleyjanenorton:
@oskirby It does the job everywhere except the update banner on Apple devices. 
Screen Shot 2021-09-10 at 10 17 45 AM
Screen Shot 2021-09-10 at 10 17 33 AM

bakulf (approved):

PR #1793, opened 2021-09-10
Fix linux text scaling on CI screen

oskirby:
I think this might fix the strange text scaling problems that we've been seeing on the CI screen. The change that really made this work was the removal of `Layout.preferredWidth` which seems to encourage Linux to draw the text as small as possible, everything else was just cleanup.

vpn-ci-scaling


lesleyjanenorton (approved):
looks great! 🎉

PR #1798, opened 2021-09-10
A simple logviewer in html/js

bakulf:


tools/logviewer/main.js
```diff
@@ -0,0 +1,241 @@
...
+      this.parseLine(line);
+    }
+
+    this.createDateRange();
+    this.createModuleList();
+    this.createComponentList();
+    this.populateLogTable();
+  },
+
+  parseLine(line) {
```
oskirby, line 25: I'm noticing a lot of `return` statements in here while attempting to parse out the category and class. However, and we do often see a few log messages that don't come from our own `Logger` class and are instead generated from Qt's logging infrastructure instead. 

I'm left a little bit worried that we might be loosing some of those other log messages in this view by discarding anything that doesn't match the Category and Class syntax.

oskirby (approved):
This is really cool, I like it a lot!

It inspires me to think of some neat things we could do with this page:
- Connect to the inspector web socket and update the logs live from the VPN client and/or daemon?
- Serve this page directly from the `View Logs` link on the `Get Help` menu instead of needing to save logs to the Desktop
- Add a button to upload the logs somewhere and connect them to a support ticket.

tools/logviewer/main.js
bakulf: Good point. Follow up to support qt logs and the wireguard ones

PR #1800, opened 2021-09-10
Key rotation

bakulf:


strseb (approved):

src/settingsholder.h
```diff
@@ -131,13 +131,14 @@ class SettingsHolder final : public QObject {
...
          setConnectionChangeNotification);
   GETSET(bool, hasFeaturesTourShown, featuresTourShown, setFeaturesTourShown);
-
   GETSET(QStringList, hasMissingApps, missingApps, setMissingApps)
   GETSET(QStringList, hasSeenFeatures, seenFeatures, setSeenFeatures)
   GETSET(QStringList, hasDevModeFeatureFlags, devModeFeatureFlags,
          setDevModeFeatureFlags);
   GETSET(QStringList, hasRecentConnections, recentConnections,
          setRecentConnections);
+  GETSET(bool, hasDeviceKeyRegenerated_2_5, deviceKeyRegenerated_2_5,
```
oskirby, line 140: Could we instead do this by storing the version number under which the key current device key was generated? That way we don't have to keep adding bespoke settings every time an issue like this occurs and the question instead becomes a comparison of version numbers.

oskirby: For example: `if (settingsHolder->deviceKeyVersion() < "2.5") { /* do some stuff */ }`

oskirby (changes requested):

oskirby (approved):

PR #1806, opened 2021-09-10
Fix minor support flow issues

mbirghan:
This PR fixes two issues.
1. for authenticated users we did not send an email
2. from the thank you view we did not get back to the settings page


mbirghan:
Fixed also the string issue with `%1` on the thank you page

bakulf (approved):

PR #1807, opened 2021-09-10
More comments about strings and arguments

bakulf:


translations/strings.yaml
```diff
@@ -45,6 +45,29 @@
...
+# If a string contains arguments, add '%<number>' for each argument. For
+# instance, you can write 2 strings in these way:
+#   foo:
+#     value: You have %1 active devices
+#     comment: %1 is the number of active devices.
+#   bar:
+#     value: Your device %1 has been created at %2.
+#     comment: %1 is the device name. %2 is a date/time value.
+#
+# Please, write a comment to describe what the arguments are about.
```
flodolo, line 57: Worth mentioning that linter will fail if they don't?

flodolo (approved):

translations/strings.yaml
bakulf: we can actually implement this linter thing. Good point!

PR #1813, opened 2021-09-13
Comments are required for strings with arguments

bakulf:


flodolo (changes requested):
This is already implemented
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/.github/l10n/check_l10n_issues.py#L54

Any concern that it's not working as expected?

bakulf:
This is in the commit

PR #1823, opened 2021-09-14
MacOS: Fix squashed/overlarge status bar icon

### LINKED ISSUE #1134: System tray icon squished when on secondary display
On Mac when a display is not the active display the system tray icon gets sqashed. 

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-68)

jswinarton:
Fix a MacOS issue where the status bar icon appears horizontally squashed under the following conditions:
- Using MacOS Big Sur
- Using an external display
- The display has a device pixel ratio greater than 1

The icon also appeared much larger than other status bar icons, which is also fixed.

The underlying reason for this is [a bug in Qt 5.15](https://bugreports.qt.io/browse/QTBUG-88600) that causes the system tray icon class to not properly respect the device pixel ratio of the display. The irritating thing is that this bug has actually been fixed, but was released as part of Qt 5.15.3 which apparently is a [commercial-only patch release](https://www.qt.io/blog/commercial-lts-qt-5.15.3-released). This means we need to resort to an ugly hack, because capitalism.

The hack implementation (borrowed from the original bug report) uses an Objective-C technique called "swizzling" that makes it possible to replace the underlying implementation of an interface at runtime. The fix replaces the `setImage` method on `NSStatusBarButton` with a new version that correctly scales the image according to MacOS UI specifications.

Note that I'm fairly new to C++ / Objective-C so any feedback on code style/best practices/etc would be highly appreciated!

New appearance on secondary monitor:
Screen Shot 2021-09-15 at 1 36 53 PM

Fixes #1134.


src/platforms/macos/macosutils.mm
```diff
@@ -89,3 +89,105 @@ bool dockClickHandler(id self, SEL cmd, ...) {
...
+@end
+
+@implementation NSImageScalingHelper
++ (NSImage*)imageByScaling:(NSImage*)sourceImage size:(NSSize)targetSize {
+  NSImage* newImage = nil;
+
+  if ([sourceImage isValid]) {
+    NSSize sourceSize = [sourceImage size];
+
+    if (sourceSize.width != 0.0 || sourceSize.height != 0.0) {
```
bakulf, line 132: This should be a && instead of ||.

bakulf (approved):

src/platforms/macos/macosutils.mm
jswinarton: Thanks, fixed

lesleyjanenorton (approved):

PR #1824, opened 2021-09-15
Custom staging server

mbirghan:


src/constants.cpp
```diff
@@ -5,10 +5,18 @@
...
+
 namespace {
 bool s_inProduction = true;
+QString s_stagingServerAddress = "";
 }  // namespace
 
 bool Constants::inProduction() { return s_inProduction; }
 
-void Constants::setStaging() { s_inProduction = false; }
+QString Constants::getStagingServerAddress() { return s_stagingServerAddress; }
```
bakulf, line 17: `const QString&` .We don't need to duplicate the string each time.

src/constants.h
```diff
@@ -6,11 +6,13 @@
 #define CONSTANTS_H
 
 #include <stdint.h>
+#include <QString>
 
 namespace Constants {
 
 // Returns true if we are in a production environment.
 bool inProduction();
+QString getStagingServerAddress();
```
bakulf, line 15: `const QString&`

src/constants.h
```diff
@@ -6,11 +6,13 @@
 #define CONSTANTS_H
 
 #include <stdint.h>
+#include <QString>
```
bakulf, line 9: forward declaration:

`class QString;`

src/mozillavpn.cpp
```diff
@@ -194,6 +194,8 @@ MozillaVPN::State MozillaVPN::state() const { return m_state; }
 
 bool MozillaVPN::stagingMode() const { return !Constants::inProduction(); }
 
+QString MozillaVPN::stagingServerAddress() const { return Constants::getStagingServerAddress(); }
```
bakulf, line 197: same here. `const QString&

src/mozillavpn.h
```diff
@@ -120,6 +121,7 @@ class MozillaVPN final : public QObject {
   const QString& serverPublicKey() const { return m_serverPublicKey; }
 
   bool stagingMode() const;
+  QString stagingServerAddress() const;
```
bakulf, line 124: here as well.

src/mozillavpn.cpp
bakulf: You don't really need this. Remove it

src/ui/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -91,8 +154,9 @@ VPNFlickable {
     }
 
     VPNCheckBoxAlert {
+        id: restartRequired
         anchors.top: inspectorLink.bottom
-        visible: VPN.stagingMode != VPNSettings.stagingServer
+        visible: VPN.stagingMode != VPNSettings.stagingServer || VPN.stagingServerAddress != VPNSettings.stagingServerAddress
```
bakulf, line 159: I have a different solution. What about this:
1. `visible: false`
2.
```
Connections {
  target: VPNSettings
  function onStagingServerAddressChanged() { restartRequired.visible = true; }
  function onStagingServerChanged() { restatRequired.visible = true; }
}
```

tests/auth/mocmozillavpn.cpp
```diff
@@ -24,6 +24,8 @@ MozillaVPN::~MozillaVPN() {}
 MozillaVPN::State MozillaVPN::state() const { return StateInitialize; }
 bool MozillaVPN::stagingMode() const { return true; }
 
+QString MozillaVPN::stagingServerAddress() const { return Constants::API_STAGING_URL; }
```
bakulf, line 27: Remove this

tests/unit/mocmozillavpn.cpp
```diff
@@ -26,6 +26,8 @@ MozillaVPN::State MozillaVPN::state() const { return TestHelper::vpnState; }
 
 bool MozillaVPN::stagingMode() const { return true; }
 
+QString MozillaVPN::stagingServerAddress() const { return Constants::API_STAGING_URL; }
```
bakulf, line 29: And this :)

bakulf (changes requested):
Good! I think we can make this PR smaller. Apply the comments if you think they are OK.

src/networkrequest.cpp
```diff
@@ -92,7 +92,7 @@ QString NetworkRequest::apiBaseUrl() {
   if (pe.contains("MVPN_API_BASE_URL")) {
```
bakulf, line 92: This part will overwrite the staging server value. I would like to see:
1. remove this `MVPN_API_BASE_URL` check here and simply return: `Constants::getStagingServerAddress()`.
2. see the other comment in settingsholder.cpp

src/settingsholder.cpp
```diff
@@ -271,6 +274,10 @@ GETSETDEFAULT(SETTINGS_DEVELOPERUNLOCK_DEFAULT, bool, toBool,
 GETSETDEFAULT(SETTINGS_STAGINGSERVER_DEFAULT, bool, toBool,
               SETTINGS_STAGINGSERVER, hasStagingServer, stagingServer,
               setStagingServer, stagingServerChanged)
+GETSETDEFAULT(SETTINGS_STAGINGSERVERADDRESS_DEFAULT, QString, toString,
```
bakulf, line 277: Instead of having a default value == "", what about if you do:
```
GETSETDEFAULT(getEnvVariable("MVPN_API_BASE_URL"), ...
```

where `getEnvVariable` is something _private_ like:

```
QString Settings::getEnvVariable(const QString& name) const {
  QProcessEnvironment pe = QProcessEnvironment::systemEnvironment();
  if (pe.contains(name)) return pe.value(name);
  return QString();
}
```

bakulf (changes requested):

src/ui/developerMenu/ViewDeveloperMenu.qml
mbirghan: Your suggestion has two drawbacks: 
1. We would show the warning even if we change it back to the previous settings. 
2. If we leave the page and return it resets and we don't see the warning even if we changed something before

mbirghan:
@bakulf I addressed your comments and responded to one of your comments with drawbacks we should consider.

bakulf (approved):

bakulf:
Can you fix the linter and the unit-test.

bakulf (approved):

src/settingsholder.cpp
oskirby: Shouldn't this default to `Constants::API_STAGING_URL` when the environment variable is set? Otherwise it seems like the only way to test against our normal staging server is by typing the URL in manually after each reset of the settings.

mbirghan: This is one of the goals of this. We do not want it to be as easy as it currently is for a user to accidentally find the dev menu and then switch to the staging server. 
Copy pasting the address into this field is fairly easy and not a big problem if you know the address.
We could look into ways to set this at build time for easier development.

PR #1827, opened 2021-09-15
Fix libqt5test5 deps for Ubuntu hirsute/groovy

oskirby:
The combination of production and staging meant that we had a missing dependency on libqt5test5 for builds of the VPN client on distributions which use the Qt5 packages provided by Ubuntu (for now, this means hirsute and groovy). The packages build successfully in CI but fail to start on a new VM image with these distros.


bakulf (approved):
We need to release a 2.5.0 package. Same orig but different Debian files. Right?

oskirby:
Pushing this commit to releases/2.5.0 would change the timestamps of the source tarball, which the PPA would treat as though the sources have been modified. We have handled this by manually updating the PPA build for hirsute but reusing the `orig.tar.gz`. You can see the results of this here: https://launchpad.net/~mozillacorp/+archive/ubuntu/mozillavpn/+build/22083212

bakulf:
Yes :) this is what I meant. Thanks!

PR #1843, opened 2021-09-20
Use Constants::API_PROD as default value for Stage_server if none is given

strseb:
We should use Constants::API_STAGING_URL as Default setting value if MVPN_API_BASE_URL is not present - Otherwise auth uri might be empty when unset;  thus we get stuck on windows (we crash on android) ... also I don't want to copy our guardian uri every time I install a debug build on my phone :) 🙈

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -71,8 +71,6 @@ CONSTEXPR(uint32_t, surveyTimerMsec, 300000, 4000, 0)
   inline type functionName() { return inProduction() ? prod : beta; }
 
 constexpr const char* API_PRODUCTION_URL = "https://vpn.mozilla.org";
-constexpr const char* API_STAGING_URL =
-    "https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net";
 
 constexpr const char* LOGO_URL = ":/ui/resources/logo-dock.png";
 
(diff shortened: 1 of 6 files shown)


bakulf:
I would like to hide the 'staging' env. What about if we use the prod URL by default? And we remove API_STAGING_URL entirely?

strseb:
Sure :)

bakulf (approved):

src/settingsholder.h
```diff
@@ -159,6 +159,7 @@ class SettingsHolder final : public QObject {
   void removeEntryServer();
 
   QString getEnvVariable(const QString& name) const;
```
mbirghan, line 161: This can be removed.

src/constants.h
```diff
@@ -71,8 +71,6 @@ CONSTEXPR(uint32_t, surveyTimerMsec, 300000, 4000, 0)
   inline type functionName() { return inProduction() ? prod : beta; }
 
 constexpr const char* API_PRODUCTION_URL = "https://vpn.mozilla.org";
-constexpr const char* API_STAGING_URL =
-    "https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net";
```
mbirghan, line 75: We still use this in two places. We should remove it there too.
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/platforms/wasm/wasmnetworkrequest.cpp#L59
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/tests/unit/mocnetworkrequest.cpp#L40

mbirghan (changes requested):

mbirghan (approved):

Issue #1495, opened 2021-08-04
VPN on Android does not connect, stuck on "attempting to confirm connection"


JKingZapata:
Users on Android devices are unable to connect to Mozilla VPN version 2.4. VPN shows "attempting to confirm connection" and after a moment the switch turns off.

**Technical Information**

* OS version: Android 8.1, Android 9
* Device model: Alcatel 3T8, Moto G6, Samsung J3
* FX product Mozilla VPN 2.4

**Steps to Reproduce**

1. Sign up to a Mozilla VPN subscription
1. Download/install Mozilla VPN
1. Flip switch to connect to VPN

**Actual Results**

1. Sign up to a Mozilla VPN subscription
1. Download/install Mozilla VPN
1. Flip switch to connect to VPN
1. Switch shows "attempting to confirm connection"
1. Switch never connects and turns off after a moment

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-845)

strseb:
Probably a duplicate of  #637 :/

JKingZapata:
Attaching logs

[VPN 2.4 connection issue Android.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/6932670/VPN.2.4.connection.issue.Android.txt)

birdsarah:
Also #1265

LuminitaB:
We have checked this today on the following setup:
- Samsung Galaxy S10 Plus, with Android 9
- Mozilla VPN app v. 2.5.0 (16293690)
The VPN does not connect, it stays in the "Attempting to confirm connection" status and it throws the "No internet connection try Again" message. (See below image)

![Screenshot_20210820-104127](https://user-images.githubusercontent.com/52028275/130201655-82a573b7-2a44-4010-9dc1-bd538a9106c8.jpg)
[Mozilla VPN logs.docx](https://github.com/mozilla-mobile/mozilla-vpn-client/files/7020076/Mozilla.VPN.logs.docx)

I have also attached the logs from the app:
[Mozilla VPN logs.docx](https://github.com/mozilla-mobile/mozilla-vpn-client/files/7020081/Mozilla.VPN.logs.docx)

ValentinaPC:
I encountered the "Connecting..." issue on both Windows and Linux, and forced me to restart the PC so that I can use it again.
Here are logs from Windows.
[...]

PR #1851, opened 2021-09-21
Check Ping return code and add fallback path for ping errors

strseb:
RE: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1495

From the new logs of luminita, the ping process exits with code 2 - (an error). We should check and handle it. 
My idea here is whenever we have such a ping error once a session, let's switch to the DUMMY_PING sender, so we don't do unnecessary steps here, like opening sockets and starting processes :) .


src/pingsender.cpp
```diff
@@ -61,13 +61,22 @@ void PingSender::genericSendPing(const QStringList& args, qint16 sequence) {
 
   connect(process, qOverload<int, QProcess::ExitStatus>(&QProcess::finished),
           this, [this, sequence](int exitCode, QProcess::ExitStatus) {
-            if (exitCode == 0) emit recvPing(sequence);
+            logger.info() << "ping exit code: " << exitCode;
```
bakulf, line 64: This is going to be super spammy. Remove this line

src/pinghelper.cpp
```diff
@@ -50,17 +51,25 @@ void PingHelper::start(const QString& serverIpv4Gateway,
 
   m_gateway = serverIpv4Gateway;
   m_source = deviceIpv4Address.section('/', 0, 0);
-  m_pingSender =
+
+  if (s_has_critical_ping_error) {
+    logger.info() << "Encountered Unrecoverable ping error, using DUMMY Ping";
+    m_pingSender = new DummyPingSender(m_source, this);
```
bakulf, line 57: I like the solution, but we should go back to the normal ping sender after a while. Maybe after a few minutes?

bakulf (changes requested):

src/pinghelper.cpp
strseb: Good idea! Done :)

src/pinghelper.cpp
```diff
@@ -192,3 +202,23 @@ double PingHelper::loss() const {
...
+  // and restart the pinghelper, to replace the impl with a dummy impl
+  // which we will use for the rest of the session
+  emit pingSentAndReceived(1);  // Fake a ping response with 1ms;
+  s_has_critical_ping_error = true;
+  stop();
+  start(m_gateway, m_source);
+
+  TimerSingleShot::create(qApp, 600000, [&] {  // 10 Minutes
+    logger.debug() << "Removing ping error state";
+    s_has_critical_ping_error = false;
```
bakulf, line 222: Here you have to call stop() + star(), right? Plus, if you set, as `parent` the m_pingSender, you don't need to check if the pingSender has been deleted in the meantime.

bakulf (changes requested):

src/pinghelper.cpp
```diff
@@ -192,3 +201,25 @@ double PingHelper::loss() const {
...
+  emit pingSentAndReceived(1);  // Fake a ping response with 1ms;
+  s_has_critical_ping_error = true;
+  stop();
+  start(m_gateway, m_source);
+
+  TimerSingleShot::create(this, 600000, [&] {  // 10 Minutes
+    logger.debug() << "Removing ping error state";
+    s_has_critical_ping_error = false;
+    stop();
+    start(m_gateway, m_source);
```
bakulf, line 223: What about if the time has been stopped in the meantime?

bakulf (changes requested):

src/pinghelper.cpp
strseb: good point lets just not do that :)

bakulf (approved):

Issue #1599, opened 2021-08-20
Glean.js makes iOS to crash for too much stack recursion


bakulf:
We should fix this issue, asap! For now, let's disable glean for iOS.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-882)

data-sync-user:
➤ Nma-Sie-Anyene Davenport commented:

No action needed from PSP  team

brizental:
Hey folks, just an update here. Unfortunately I am not able to reproduce this bug in my own iPhone. 

If possible, it would be useful to check if the crash persists even when pings are sent without passing a signal from C++ to QML. Since the crash happens exactly at the time the signal is sent, maybe there is some issue with the communication between the two layers.

Also, do we have access to the JavaScript console logs at the time of the crash? That could potentially have important information about the crash. I would suggest enabling the [`logPings`](https://mozilla.github.io/glean/book/reference/debug/logPings.html) debug feature in Glean to have the most logs Glean will provide.

data-sync-user:
➤ Adrienne Davenport commented:

JR Johnson - Can you give us information regarding the frequency of this occurrence, please?

data-sync-user:
➤ JR Johnson commented:

Adrienne Davenport  I am not familiar with this issue currently. Can you give us identification criteria? If its as simple as inbound referring to IOS app crashes, then we have not seen any an impact ticket wise currently.

data-sync-user:
➤ Sarah Bird commented:

It’s actually Beatriz Rizental Machado who’s working on this. But I’m assigning it to me as I’m helping with fix verification and any other way I can, and it would be weird to have Beatriz on our sprint board.

data-sync-user:
➤ Sarah Bird commented:

Adrienne Davenport JR Johnson this is not happening in production as we disabled glean on iOS to prevent it happening in production.

PR #1854, opened 2021-09-21
[DO NOT MERGE] Bug 1727776, #1599 - Attempt at fixing Glean.js + iOS issues

brizental:
Related to: #1599

Hey folks! 

This is a "do not merge" PR because it contains an unreleased version of Glean.js. 

**tl;dr;** I was able to fix the issue for Release builds, but the issue persists in Debug builds. The current crash is caused by a bug in the Qt JS interpreter and I was not able to figure out a change in the JS code that does not cause it. Since it happens only in Debug mode, I would like to know from you if this is a good enough situation for now. I will file a bug against the Qt bug tracker in order to see if that gets fixed upstream, but that is a slow process for sure. If possible I would like for you to test it out as well to see if you also can't reproduce the bug in Release builds anymore.

There are many changes to Glean.js here -- including extremely verbose logs, those will be turned off in the real release. The [commit history](https://github.com/mozilla/glean.js/pull/744/commits) on the Glean.js PR has one by one the changes I made and why. On the description of that PR there is also a detailed description of my investigation efforts.

Regarding the VPN app changes, during my investigations I was working with the newest version of Glean.js, so I also updated Glean.js and glean_parser in here for testing. I also re-enabled Glean in all platforms, but that would probably need to be undone in order to disable it in iOS debug mode.


src/ui/main.qml
```diff
@@ -63,33 +63,11 @@ Window {
             minimumWidth = Theme.desktopAppWidth
         }
 
-        Glean.initialize('MozillaVPN', VPNSettings.gleanEnabled && VPNFeatureList.get("glean").isSupported, {
+        Glean.setLogPings(true);
+        Glean.setDebugViewTag("vpn-testing");
```
brizental, line 67: This tag is [a debugging feature](https://mozilla.github.io/glean/book/reference/debug/debugViewTag.html). Setting this means all pings will be tagged upon sending, in practice that means the pings are re-routed to Glean's debug ping viewer. You can see the ping I sent when testing this here: https://debug-ping-preview.firebaseapp.com/pings/vpn-testing

birdsarah:
EXCELLENT - I'll take a look

brizental:
Quick update here. The crash I described has been confirmed as a Qt JS interpreter bug. See: https://bugreports.qt.io/browse/QTBUG-96788

birdsarah:
Closing as this PR has served it's purpose and 0.21 of glean is released and we'll be working on integration. thanks so much @brizental .

PR #1860, opened 2021-09-22
Settings refactoring

bakulf:
This is a big refactoring of how we expose setting "properties". There is a new file (`settinglist.h`) containing a map of the properties. And this map is included in multiple ways to give different meetings to the `SETTING` macro.

### DIFF
--- a/src/settingsholder.h
+++ b/src/settingsholder.h
@@ -14,45 +14,16 @@ class SettingsHolder final : public QObject {
   Q_OBJECT
   Q_DISABLE_COPY_MOVE(SettingsHolder)
 
-  Q_PROPERTY(bool ipv6Enabled READ ipv6Enabled WRITE setIpv6Enabled NOTIFY
-                 ipv6EnabledChanged)
-  Q_PROPERTY(bool localNetworkAccess READ localNetworkAccess WRITE
-                 setLocalNetworkAccess NOTIFY localNetworkAccessChanged)
-  Q_PROPERTY(bool unsecuredNetworkAlert READ unsecuredNetworkAlert WRITE
-                 setUnsecuredNetworkAlert NOTIFY unsecuredNetworkAlertChanged)
-  Q_PROPERTY(bool captivePortalAlert READ captivePortalAlert WRITE
-                 setCaptivePortalAlert NOTIFY captivePortalAlertChanged)
-  Q_PROPERTY(bool startAtBoot READ startAtBoot WRITE setStartAtBoot NOTIFY
-                 startAtBootChanged)
-  Q_PROPERTY(bool protectSelectedApps READ protectSelectedApps WRITE
-                 setProtectSelectedApps NOTIFY protectSelectedAppsChanged)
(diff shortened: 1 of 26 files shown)


src/settingsholder.h
```diff
@@ -14,43 +14,13 @@ class SettingsHolder final : public QObject {
...
-                 gleanEnabledChanged)
-  Q_PROPERTY(
-      bool serverSwitchNotification READ serverSwitchNotification WRITE
-          setServerSwitchNotification NOTIFY serverSwitchNotificationChanged)
-  Q_PROPERTY(bool connectionChangeNotification READ connectionChangeNotification
-                 WRITE setConnectionChangeNotification NOTIFY
-                     connectionChangeNotificationChanged)
+#define SETTING(type, toType, getter, setter, ...) \
+  Q_PROPERTY(type getter READ getter WRITE setter NOTIFY settingsChanged)
+
```
strseb, line 19: I'm a bit cautious to put all properties under one signal, as this would mean to re-evaluate all bindings that exist. Which is a lot of event overhead, IMHO.   
What if we would let the macro also define the old Signals that we had? e.g:

```suggestion
#define SETTING(type, toType, getter, setter, ...) \
  Q_SIGNAL void getter##Changed(); \
  Q_PROPERTY(type getter READ getter WRITE setter NOTIFY getter##Changed)

```

src/settingslist.h
```diff
@@ -0,0 +1,441 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// 1. Define the SETTING macro
+// 2. include this file
+// 3. undefine the SETTING macro
```
strseb, line 7: I really like this !

strseb (changes requested):

src/settingsholder.h
bakulf: I tried. This does not work because Q_SIGNAL is processed by `moc` and that tool does not support C++ macros. So, the signal will not be implemented in the `moc_settingholder.cpp`.

oskirby: If we get into the world of metaprogramming, we should be able to expose the settings as dynamic properties. That might enable us to cut down on the boilerplate macros, while still retaining the ability to access them via QML. I'm not sure if that allows us to dynamically create signals though.

oskirby (changes requested):
Overall, I like this a lot. There's a huge amount of cleanup. However, I don't like the need to funnel everything through a single `settingsChanged()` signal and rely on the receiver to do the filtering since that seems to result in a lot of boilerplate in the places that need to handle signal changes.

I would really like to find a solution that allows the users of a setting to receive a signal only if their specific setting changes.

src/settingsholder.h
oskirby: If we want to go way into the weeds of dynamic `Objects`, there might be some breadcrumbs here: https://doc.qt.io/archives/qq/qq16-dynamicqobject.html

oskirby (approved):
Looks great!

PR #1870, opened 2021-09-23
WIP: experiments with adjust proxy

mbirghan:


src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,160 @@
...
+  QByteArray line = m_buffer.left(pos);
+  m_buffer.remove(0, pos + 1);
+
+  QList<QByteArray> parts = line.split(' ');
+  if (parts.length() < 2) {
+    return;
+  }
+
+  m_method.append(parts[0]);
+  m_route.append(parts[1]);
```
bakulf, line 61: Maybe the parsing of the `parts[1]` should be done using QUrl + QUrlQuery

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,160 @@
...
+    pos = header.indexOf(":");
+    if (pos == -1) {
+      continue;
+    }
+
+    QByteArray headerName = header.left(pos);
+    header.remove(0, pos + 1);
+    QByteArray headerValue = header.trimmed();
+
+    m_headers.insert(headerName, headerValue);
```
bakulf, line 95: We do need to parse the headers just to retrieve the Content-Length one. All the rest should be kept and stored as it is.
The Content-Length header is important because we have to read that amount of data from the buffer after the end of the header "\n\n" or "\r\n\r\n".

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  if (m_parametersString.isEmpty()) {
+    m_parametersString = m_buffer.trimmed();
+  }
+
+  // Parsing parameters
+  while (true) {
+    pos = m_parametersString.indexOf("&");
```
bakulf, line 104: Let's use QUrlQuery here.

src/adjustproxyconnection.h
```diff
@@ -0,0 +1,40 @@
...
+  void forwardRequest();
+
+ private:
+  QTcpSocket* m_connection;
+
+  QByteArray m_buffer;
+  QString m_method;
+  QByteArray m_route;
+  QByteArray m_parametersString;
+  QHash<QByteArray, QByteArray> m_headers;
```
bakulf, line 36: Maybe the order is important. Let's use QList instead to keep the order.

src/adjustproxyconnection.h
```diff
@@ -0,0 +1,40 @@
...
+
+ private:
+  void readData();
+
+  void filterParameters();
+
+  void forwardRequest();
+
+ private:
+  QTcpSocket* m_connection;
```
bakulf, line 30: = nullptr

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,160 @@
...
+        this, m_route, m_headers, m_parametersString);
+  } else {
+    logger.warning() << "Method not supported!";
+    return;
+  }
+
+  connect(request, &NetworkRequest::requestFailed,
+          [this](QNetworkReply::NetworkError error, const QByteArray& data) {
+            logger.error() << "Failed adjust request" << error;
+            logger.debug() << data;
```
bakulf, line 150: Here we should answer with an error... maybe?

bakulf (changes requested):
This looks super promissing! Well done!

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+        return;
+      }
+      processFirstLine();
+    case ProcessingState::FirstLineDone:
+      if (m_buffer.isEmpty()) {
+        return;
+      }
+      processHeaders();
+    case ProcessingState::HeadersDone:
+      processParameters();
```
bakulf, line 60: `[[fallthrough]];`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+  Q_ASSERT(m_connection);
+  QByteArray input = m_connection->readAll();
+  m_buffer.append(input);
+
+  switch (m_state) {
+    case ProcessingState::NotStarted:
+      if (m_buffer.isEmpty()) {
+        return;
+      }
+      processFirstLine();
```
bakulf, line 53: `[[fallthrough]];`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+    case ProcessingState::FirstLineDone:
+      if (m_buffer.isEmpty()) {
+        return;
+      }
+      processHeaders();
```
bakulf, line 58: `[[fallthrough]];`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+    case ProcessingState::HeadersDone:
+      processParameters();
+    case ProcessingState::ParametersDone:
+      break;
+    default:
+      logger.debug() << "Unknown ProcessingState: " << m_state;
```
bakulf, line 64: `Q_ASSERT(false);`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  if (m_state == ProcessingState::ParametersDone) {
+    filterParameters();
```
bakulf, line 68: Why here and not in the switch stmt?

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+  if (pos == -1) {
+    return;
+  }
+
+  QByteArray line = m_buffer.left(pos);
+  m_buffer.remove(0, pos + 1);
+
+  QList<QByteArray> parts = line.split(' ');
+  if (parts.length() < 2) {
+    m_connection->close();
```
bakulf, line 83: `logger.info() << "Invalid HTTP request";`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return;
+  }
+
+  m_method.append(parts[0]);
+  m_route = parts[1];
```
bakulf, line 88: `parts[1].trimmed()`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+    int pos = m_buffer.indexOf("\n");
+    if (pos == -1) {
+      break;
+    }
+
+    QByteArray line = m_buffer.left(pos);
+    m_buffer.remove(0, pos + 1);
+
+    QByteArray header = line.trimmed();
+    if (header == "") {
```
bakulf, line 104: `header.isEmpty()`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    m_headers.append(headerPair);
+  }
+
+  m_state = ProcessingState::HeadersDone;
+}
+
+void AdjustProxyConnection::processParameters() {
+  if (m_method == "GET") {
+    m_paramters = QUrlQuery(m_route);
```
bakulf, line 129: `s/paramters/parameters/g`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+    if (m_buffer.isEmpty()) {
+      return;
+    }
+    m_paramters = QUrlQuery(m_buffer.trimmed());
+  }
+
+  m_state = ProcessingState::ParametersDone;
+}
+
+void AdjustProxyConnection::filterParameters() {
```
bakulf, line 140: rename this method to `filterParametersAndForwardRequest()`

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+        NetworkRequest::createForAdjustForwardGet(this, m_route, m_headers);
+  } else if (m_method == "POST") {
+    request = NetworkRequest::createForAdjustForwardPost(
+        this, m_route, m_headers, m_paramters.toString().toUtf8());
+  } else {
+    logger.warning() << "Method not supported!";
+    return;
+  }
+
+  connect(
```
bakulf, line 164: Maybe we should "proxy" the response entirely. Ignoring the request failure or success.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+      request, &NetworkRequest::requestFailed,
+      [this](QNetworkReply::NetworkError, const QByteArray& data, int status) {
+        QByteArray response(HTTP_RESPONSE);
+        response.replace("<status>", QByteArray::number(status));
```
bakulf, line 168: You can obtain the status from the request: `request->statusCode()`

src/mozillavpn.cpp
```diff
@@ -90,7 +92,18 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
...
+  QObject::connect(controller(), &Controller::readyToQuit, adjustProxy,
+                   &AdjustProxy::close);
+  for (int i = 0; i < 5; i++) {
+    quint16 port = QRandomGenerator::global()->bounded(1024, 65536);
+    bool succeeded = adjustProxy->initialize(port);
+    if (succeeded) {
+      break;
+    }
+  }
+  AdjustHandler::initialize(adjustProxy->serverPort());
```
bakulf, line 105: if !succeeded, do not activate adjust

src/networkrequest.cpp
```diff
@@ -133,6 +133,55 @@ NetworkRequest* NetworkRequest::createForAuthenticationVerification(
...
+  url.append(route.query());
+  r->m_request.setUrl(url);
+
+  for (QPair header : headers) {
+    if (header.first != "Host" && header.first != "Content-Length") {
+      r->m_request.setRawHeader(header.first, header.second);
+    }
+  }
+
+  r->getRequest();
```
bakulf, line 157: I would suggest having just POST requests, with a JSON object like this:

```
{
  method: "GET",
  url: "the url",
  headers: {key, value},
  body: null_or_string
}
```

src/networkrequest.h
```diff
@@ -30,6 +30,15 @@ class NetworkRequest final : public QObject {
       QObject* parent, const QString& pkceCodeSuccess,
       const QString& pkceCodeVerifier);
 
+  static NetworkRequest* createForAdjustForwardGet(
```
bakulf, line 33: I would unify these 2 in 1 method.

src/networkrequest.h
```diff
@@ -147,7 +156,8 @@ class NetworkRequest final : public QObject {
 
  signals:
   void requestHeaderReceived(NetworkRequest* request);
-  void requestFailed(QNetworkReply::NetworkError error, const QByteArray& data);
+  void requestFailed(QNetworkReply::NetworkError error, const QByteArray& data,
+                     int status);
```
bakulf, line 160: No need this. status can be obtain calling `statusCode()`

bakulf (changes requested):

src/adjustproxyconnection.cpp
mbirghan: For that, we would need to have access to the `m_reply` of the `NetworkRequest` which is private.

src/mozillavpn.cpp
mbirghan: We check in the `initialize` function if the proxy is listening.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "qmlengineholder.h"
+#include "networkrequest.h"
+
+#include <QFile>
+#include <QHostAddress>
+#include <QTcpSocket>
+#include <QUrl>
+#include <QUrlQuery>
+
+constexpr const char* HTTP_RESPONSE =
```
bakulf, line 18: This could be a QString() with %1 and %2.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+void AdjustProxyConnection::readData() {
+  Q_ASSERT(m_connection);
+  QByteArray input = m_connection->readAll();
+  m_buffer.append(input);
+
+  switch (m_state) {
+    case ProcessingState::NotStarted:
+      if (m_buffer.isEmpty()) {
```
bakulf, line 50: you could remove this check. Or move it in the processFirstLine()

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+        return;
+      }
+      processFirstLine();
+      [[fallthrough]];
+    case ProcessingState::FirstLineDone:
+      if (m_buffer.isEmpty()) {
```
bakulf, line 56: Move this check into processHeaders()

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+  m_route = parts[1].trimmed();
+
+  m_state = ProcessingState::FirstLineDone;
+}
+
+void AdjustProxyConnection::processHeaders() {
+  while (true) {
+    int pos = m_buffer.indexOf("\n");
+    if (pos == -1) {
+      break;
```
bakulf, line 99: This should be a return. We don't want to set status = `HeadersDone` if we have an unfinished line.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  m_state = ProcessingState::HeadersDone;
+}
+
+void AdjustProxyConnection::processParameters() {
+  if (m_method == "GET") {
+    m_parameters = QUrlQuery(m_route);
+  } else {
+    if (m_buffer.isEmpty()) {
```
bakulf, line 133: This method is not correct. I could POST data to `/foo?a=b`.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+    header.remove(0, pos + 1);
+    QByteArray headerValue = header.trimmed();
+
+    QPair<QString, QString> headerPair;
+    headerPair.first = QString(headerName);
+    headerPair.second = QString(headerValue);
+
+    m_headers.append(headerPair);
+  }
+
```
bakulf, line 125: Here we need to parse the headers to retrieve the `content-length` one to see the length of the body.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+            m_connection->write(response);
+            m_connection->close();
+          });
+
+  connect(request, &NetworkRequest::requestCompleted,
+          [this](const QByteArray& data) {
+            logger.debug() << "Adjust request succeeded";
+            logger.debug() << data;
+            QByteArray response(HTTP_RESPONSE);
+            response.replace("<status>", "200");
```
bakulf, line 191: `QByteArray::number(request->statusCode())`. There are many successful status code. Anything in the range 200 <> 299 is considered OK.

src/networkrequest.cpp
```diff
@@ -133,6 +133,41 @@ NetworkRequest* NetworkRequest::createForAuthenticationVerification(
...
+    }
+  }
+
+  QJsonObject obj;
+  obj.insert("method", method);
+  obj.insert("path", route);
+  obj.insert("headers", headersObj);
+  obj.insert("parameters", parameters);
+
+  QJsonDocument json;
```
bakulf, line 164: `QJsonDocument json(obj);`

bakulf (changes requested):

src/networkrequest.cpp
```diff
@@ -133,6 +133,41 @@ NetworkRequest* NetworkRequest::createForAuthenticationVerification(
...
+  r->m_request.setHeader(QNetworkRequest::ContentTypeHeader,
+                         "application/json");
+
+  QUrl url(apiBaseUrl());
+  url.setPath("/api/v1/vpn/adjust");
+  r->m_request.setUrl(url);
+
+  QJsonObject headersObj;
+  for (QPair<QString, QString> header : headers) {
+    if (header.first != "Host" && header.first != "Content-Length") {
```
bakulf, line 153: Compare the headers using a case-insensitive algorithm:
```
  if (QString::compare(header.first, "host", Qt::CaseInsensitive) != 0 && QString::compare(header.first, "content-length", Qt::CaseInsensitive)) {
  ..
}
```

But probably, the best approach is to skip these headers when parsing them,

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  m_state = ProcessingState::ParametersDone;
+}
+
+void AdjustProxyConnection::filterParametersAndForwardRequest() {
+  QList<QPair<QString, QString>> newParameters;
+
+  for (QPair<QString, QString> parameter : m_parameters.queryItems()) {
+    if (parameter.first == "adid" || parameter.first == "app_token" ||
```
bakulf, line 146: Here, I would suggest something different. Let's create 2 QList<> of params (allowed, denied). Anything else should be kept in a list, written in the log file, and shared with the guardian-website.

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+  for (QPair<QString, QString> parameter : m_parameters.queryItems()) {
+    if (allowList.contains(parameter.first)) {
+      newParameters.append(
+          QPair<QString, QString>(parameter.first, parameter.second));
+    } else if (denyList.contains(parameter.first)) {
+      newParameters.append(
+          QPair<QString, QString>(parameter.first, denyList[parameter.first]));
+    } else {
+      unknownParameters.append(
```
bakulf, line 223: Let's still add it to the newParameters and let's store the parameter.first in the unknownParameters. The unknownParameters is just for reporting.

src/adjustproxyconnection.cpp
bakulf: we can expose it :)

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include <QHostAddress>
+#include <QTcpSocket>
+#include <QUrl>
+#include <QUrlQuery>
+
+const QString HTTP_RESPONSE =
+    "HTTP/1.1 %1\nContent-Type: "
+    "application/json\n\n%2\n";
+
+const QList<QString> allowList = {"adid",
```
bakulf, line 24: Often the compilers do miracles, but doing this you are doing 2 operations: 1. CTOR without args, 2. assignment. 
You can do it in 1 step using the CTOR: `const QString("HTTP/1.1 ...")`

bakulf: same here for the QList. YOu can do:

const QStringList allowList{"a", "b", "c"...};

src/adjustproxyconnection.cpp
mbirghan: Yes, we can expose it. But then we would have to write the logic for all events for the reply because we could not use the `NetworkRequest`  implementation.
Since we don't proxy the request in guardian but create a new request with the same body and status I think this is not necessary.

bakulf (approved):

src/adjustproxyconnection.cpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    m_headers.append(headerPair);
+  }
+
+  m_state = ProcessingState::HeadersDone;
+}
+
+void AdjustProxyConnection::processParameters() {
+  uint32_t bodyLength = m_buffer.trimmed().length();
```
bakulf, line 193: cast it to `int` to fix the android build.

birdsarah:
Don't forget to take "WIP" out of PR titles - for posterity's sake.

Issue #1872, opened 2021-09-23
Mullvad proxy servers unreachable when local network access is enabled


data-sync-user:
When attempting to use the Mullvad SOCKS5 proxy servers and local network access at the same time, we find that the proxy servers are unreachable (they don't respond to ICMP pings, and nmap reports no open ports). This is because the IPv4 address range used by the proxy servers falls within the 10.0.0.0/8 range, and is therefore routed out the user's LAN instead of being forwarded into the Mullvad network.

To fix this, we should ensure that an appropriate routing prefix for the Mullvad proxy servers is added when activating the VPN. IIRC, i think this prefix range is 10.124.0.0/23.

See: https://mullvad.net/en/help/socks5-proxy/ for more information on the proxy configuration.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1041)

data-sync-user:
➤ Luminita Buruian commented:

Owen Kirby is there something the QA team needs to do on this issues or should we just close this?

Thank you.

data-sync-user:
➤ Owen Kirby commented:

The steps to reproduce this issue should go something like this.

Steps to Reproduce:

  1. Enable Local Network Access in the Network settings
  2. Activate the VPN and verify that you are connected.
  3. Open any browser of your choice, and configure it to use a SOCKS v5 proxy server of au3-wg.socks5.mullvad.net:1080 
  4. Check for leaks at https://mullvad.net/en/check/ ( https://mullvad.net/en/check/ )

Expected Outcome:

  * Mullvad leak check reports no leaks, and displays a location of Melbourne, Australia via server au3-wireguard

Actual Behaviour (before this fix):

  * Browser hangs, eventually reporting an error that the internet and/or proxy server is not available.

data-sync-user:
➤ Madalin Cotetiu commented:

Owen Kirby, Verified as fixed on Mozilla VPN 2.6.0 (2.202109242130), while using MacOS Big Sur 11.6. The Mullvad leak check reports no leaks, and displays a location of Melbourne. The navigation on the internet is possible. See the attached screenshot. 

!image_2021_09_27T14_03_40_863Z.png|width=2888,height=1434!

Thank you.

PR #1873, opened 2021-09-23
Allow access to Mullvad proxy network

oskirby:
In order to access the internal Mullvad network of proxies servers, we need to add them explicitly to the allowed IP ranges since the internal Mullvad network may overlap with the 10.0.0.0/8 private address range.

The private IPv4 address range of the Mullvad network isn't explicitly described anywhere, so the size of this prefix range as 20-bits is just a guess. At present there are 427 Wireguard servers in the network and the 3rd digit is always zero or one, so we could probably get away with a /23 prefix but that leaves very little room for growth if Mullvad adds more servers.

Closes: #1872

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -42,6 +42,12 @@ constexpr const uint32_t TIMER_MSEC = 1000;
 // X connection retries.
 constexpr const int CONNECTION_MAX_RETRY = 9;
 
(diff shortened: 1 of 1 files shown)


bakulf (approved):
Can we read the IPs from the server list?

oskirby:
Unfortunately, I don't think we don't include these proxy addresses in the server list. It seems like it would be really handy to do so.

PR #1880, opened 2021-09-25
Small UX/UI improvements

bakulf:


lesleyjanenorton (approved):
This is great! My only immediate nit is that there is a bit of strangeness around width and scrolling along the x-axis after the viewer is populated but we can fix this later!

PR #1881, opened 2021-09-26
Experiment with modules

bakulf:
We want to extend the VPN app in a modular application. "Modular" means a lot of things. Right now, as a prototype, I want to introduce the concept of static modules (no dynamic loading). The code requires a big refactoring, but, it's a starting point.

In this PR you will see:
1. renaming of `MozillaVPN.{cpp,h}` to `Core.{cpp.h}`.
2. Remove the `Core*` to the `Task::run()` method.
(these 2 commits can land at any time)
3. Introduce a `ModuleModel` object and a `Model` class, connected to features. Right now a Model has a name, a qmlWidget and a qmlView. We can do more, but we need UX first.


src/modules/modulevpn.cpp
```diff
@@ -0,0 +1,12 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "modulevpn.h"
+#include "features/featuremodulevpn.h"
+
+ModuleVPN::ModuleVPN()
+    : Module("vpn", FEATURE_MODULE_VPN, "qrc:/ui/modules/VPN.qml",
+             "qrc:/ui/modules/VPNWidget.qml") {}
+
```
bakulf, line 11: It would be great to move the "VPN" components here:
- Controller
- CaptivePortal
- AppPermissions
- etc

But we can do it step by step.

src/command.cpp
```diff
@@ -89,7 +89,7 @@ int Command::runCommandLineApp(std::function<int()>&& a_callback) {
   FeatureList::instance()->initialize();
 
   qInstallMessageHandler(LogHandler::messageQTHandler);
-  logger.info() << "MozillaVPN" << APP_VERSION;
+  logger.info() << "Core" << APP_VERSION;
```
oskirby, line 92: I think this should stay as "MozillaVPN" as a name for our product, or at least until we come up with a name for this expanded product concept that's more than just a VPN.

src/command.cpp
```diff
@@ -117,7 +117,7 @@ int Command::runGuiApp(std::function<int()>&& a_callback) {
 
   qInstallMessageHandler(LogHandler::messageQTHandler);
 
-  logger.info() << "MozillaVPN" << APP_VERSION;
+  logger.info() << "Core" << APP_VERSION;
```
oskirby, line 120: This should stay as "MozillaVPN" for the same reasons as listed above.

src/features/featuremodulesend.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "models/feature.h"
+
+constexpr const char* FEATURE_MODULE_SEND = "moduleSend";
+
+class FeatureModuleSend : public Feature {
+ public:
+  FeatureModuleSend()
+      : Feature(FEATURE_MODULE_SEND, "Module: Send",
+                false,               // Is Major Feature
```
oskirby, line 16: Shouldn't this be a major feature?

src/features/featuremodulevpn.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "models/feature.h"
+
+constexpr const char* FEATURE_MODULE_VPN = "moduleVPN";
+
+class FeatureModuleVPN : public Feature {
+ public:
+  FeatureModuleVPN()
+      : Feature(FEATURE_MODULE_VPN, "Module: VPN",
+                false,               // Is Major Feature
```
oskirby, line 16: This seems like it should be a major feature

oskirby (commented):
I think this looks fine for the most part. It seems like a work-in-progress. I get the impression that there are more commits to come with the Send module

birdsarah:
Closing as stale.

PR #1893, opened 2021-09-28
Fix the compilation errors on windows

bakulf:


src/platforms/windows/daemon/windowssplittunnel.cpp
```diff
@@ -8,15 +8,13 @@
...
 #define PSAPI_VERSION 2
 #include <Windows.h>
 #include <psapi.h>
-#include <fwpmu.h>
-#include <initguid.h>
-#include <iphlpapi.h>
+
 #include <QCoreApplication>
 #include <QFileInfo>
+#include <QNetworkInterface>
```
bakulf, line 17: Please, check if there are implications in the removal of these 3 includes. It turned out that windot11.h is not compatible with the new msvc preprocessor.

strseb: Seems to be fine to remove it :)

strseb (approved):

PR #1922, opened 2021-09-30
Adding permissions needed for Android 12

mbirghan:
Starting early next year Google will require new permissions to read Google Advertiser ID which is required by the Adjust SDK. 

Documentation:
https://help.adjust.com/en/article/get-started-android-sdk#add-permissions
https://developers.google.com/android/reference/com/google/android/gms/ads/identifier/AdvertisingIdClient.Info#public-string-getid


bakulf (approved):

mbirghan:
Ok, I looked into this a bit more because the idea of asking users for permission to use the Advertiser ID is a bit unsettling. 
Even though Adjust is saying that it is needed I found out that if this permission is not set the only thing that happens is that the Advertiser ID is just a string of zeros. 
We should ask Adjust if this is needed for attribution and if not remove it. I will also research if Android offers anything similar to SKAdNetwork. 
In any case, I think we should make someone from the product team is aware of it.

PR #1927, opened 2021-09-30
Import Adjust and make it work with HTTP

mbirghan:
In order to be able to connect to our localhost HTTP server, we need to change the Adjust SDK code to work with HTTP. 
I imported the Adjust SDK Android SDK into the android src folder and replaced the HTTPS parts with HTTP. 
Also I added `android:usesCleartextTraffic="true"` so that HTTP connections are permitted.


mbirghan:
Close this PR because I created another for the submodule part

PR #1929, opened 2021-09-30
Implementation the server-extra model

bakulf:


oskirby (commented):
This seems fine, but I don't like how we are starting to have a fairly complicated way to lookup servers that is now different between the `Server` and `ServerExtras`. We've got a list of countries, each with a list of cities, each with a list of servers, and then separately we now have map of extras that are indexed by the server public key.

It's almost like what we need is an sqlite database to look stuff up from (eg: SELECT * FROM servers WHERE country == us).

Instead of creating a `ServerExtra` model, I would have preferred to see us adding some extra fields to the `Server` model (with sensible defaults), and populating those fields from an extra task.

oskirby (approved):

bakulf:
I applied your comments. Can you take a look again?

strseb (approved):

PR #1934, opened 2021-10-01
Input messages component

flozia:
This add a new component `VPNInputMessages` to handle the display of input messages.


src/ui/components.qrc
```diff
@@ -3,6 +3,8 @@
         <file>components/forms/qmldir</file>
         <file>components/forms/VPNComboBox.qml</file>
         <file>components/forms/VPNInputBackground.qml</file>
+        <file>components/forms/VPNInputMessage.qml</file>
```
lesleyjanenorton, line 6: `VPNContextualAlert` & `VPNContextualAlerts` maybe? I think the naming here might be confusing since that these don't always appear in inputs.

src/ui/components/forms/VPNInputMessage.qml
```diff
@@ -0,0 +1,82 @@
...
+                    from: 0
+                    to: 1
+                    duration: 100
+                }
+            }
+        }
+    ]
+    Layout.fillWidth: true
+
+    VPNIcon {
```
lesleyjanenorton, line 57: Let's do something similar to what we do in `VPNCheckboxAlert` so that the icon stays vertically centered with the first line of text in cases where there are multiple lines of text.

src/ui/components/forms/VPNInputMessages.qml
```diff
@@ -0,0 +1,52 @@
...
+ColumnLayout {
+    id: messagesContainer
+    property var messages
+
+    anchors.left: parent.left
+    anchors.right: parent.right
+    anchors.topMargin: Theme.listSpacing
+    spacing: Theme.listSpacing / 4
+
+    DelegateChooser {
```
lesleyjanenorton, line 20: Nice!

src/ui/components/forms/VPNInputMessages.qml
```diff
@@ -0,0 +1,52 @@
...
+        id: messagesChooser
+        role: "type"
+
+        DelegateChoice {
+            roleValue: "info"
```
lesleyjanenorton, line 25: Does this get used anywhere?

lesleyjanenorton (commented):
Just a few questions/comments! Let's also update the languages view to to use this too.

src/ui/components/forms/VPNInputMessages.qml
flozia: Currently this is not used anywhere. If I’m not mistaken I saw use for it in a WIP version of the [in-app sign in/auth design](https://www.figma.com/file/gSRPHqjzqjlMA9rXDc5SqH/VPN_InApp_SignIn%2FAuth?node-id=1%3A708) — I’ll remove it for now.

src/ui/components/forms/VPNInputMessages.qml
flozia: :)

src/ui/components/forms/VPNInputMessages.qml
flozia: This message type is currently being used. If I’m not mistaken I have seen use for the `info` type in an earlier version of the [in-app auth spec](https://www.figma.com/file/gSRPHqjzqjlMA9rXDc5SqH/VPN_InApp_SignIn%2FAuth?node-id=1%3A708). I’ll remove this for now.

lesleyjanenorton (approved):

PR #1936, opened 2021-10-01
Improve Adjust initialization for QA

mbirghan:


src/mozillavpn.cpp
```diff
@@ -102,7 +102,9 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
       break;
     }
   }
-  AdjustHandler::initialize(adjustProxy->serverPort());
+  connect(this, &MozillaVPN::stateChanged, [adjustProxy]() {
```
bakulf, line 105: OK! 1 extra step, can we move the port creation inside the AdjustHandler::initialize?

bakulf (changes requested):

src/adjusthandler.cpp
```diff
@@ -14,21 +15,43 @@
...
-  if (!AdjustProxy::instance()->isListening()) {
+namespace {
+bool s_initialized = false;
+}  // namespace
+
+void AdjustHandler::maybeInitialize() {
+  if (s_initialized) {
+    return;
+  }
+
```
bakulf, line 28: Move `m_initialized = true` here.

bakulf: In this way, if the proxy doesn't start, we do not try each time we change state.

src/adjustproxyconnection.cpp
```diff
@@ -238,6 +238,19 @@ void AdjustProxyConnection::filterParametersAndForwardRequest() {
...
+
+  for (QPair<QString, QString> header : m_headers) {
+    headersString.append(header.first);
+    headersString.append(": ");
+    headersString.append(header.second);
+    headersString.append(", ");
+  }
+
+  logger.debug() << "Sending Adjust request with: " << m_method << ", "
+                 << headersString << m_queryParameters.toString() << ", "
```
bakulf, line 251: `logger.sensitive()` ?

src/adjustproxyconnection.cpp
```diff
@@ -238,6 +238,19 @@ void AdjustProxyConnection::filterParametersAndForwardRequest() {
...
+                 << m_bodyParameters.toString();
```
bakulf, line 252: here too

src/mozillavpn.cpp
```diff
@@ -92,17 +91,8 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
   logger.debug() << "Creating MozillaVPN singleton";
 
 #ifdef MVPN_ADJUST
-  AdjustProxy* adjustProxy = new AdjustProxy(qApp);
```
bakulf, line 95: Remove the `adjustproxy.h` include

bakulf (approved):
r+ with comments! Nice!

PR #1950, opened 2021-10-05
Support GDPR for Adjust

bakulf:
We don't know yet if we have to support GDPR. But in case we need it, this is a PR. I have to test the android part.


mbirghan (approved):
Looks good!
Just one small question. If I understand that correctly we don't turn adjust back on if the User reenables the gleanTelemetry right? Is this intended?

bakulf:
Yeah, because adjust is "unable" to reinitialize itself.

PR #1956, opened 2021-10-05
Add logging to the adjust proxy to debug adjust issue

mbirghan:
@bakulf Do you have any other ideas where we should add logs?

### DIFF
--- a/src/adjust/adjustproxy.cpp
+++ b/src/adjust/adjustproxy.cpp
@@ -47,6 +47,8 @@ bool AdjustProxy::initialize(quint16 port) {
 }
 
 void AdjustProxy::newConnectionReceived() {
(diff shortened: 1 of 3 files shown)


src/adjust/adjustproxy.cpp
```diff
@@ -47,6 +47,8 @@ bool AdjustProxy::initialize(quint16 port) {
 }
 
 void AdjustProxy::newConnectionReceived() {
+  logger.debug() << "New Adjust Proxy connection recieved";
```
bakulf, line 50: `received`

bakulf (approved):
All good, but it must be rebased.

bakulf:
Ah! Can you also cover the GDPR part?

PR #1976, opened 2021-10-07
Speed up the device removal - VPN-1094

bakulf:


bakulf:
It's not just a guardian thing. Often, we are slow because we have pending tasks.

Issue #1963, opened 2021-10-06
Win 11 Devices are displayed as Windows 10


data-sync-user:
**VPN version:**
2.5.0;

**Affected Platforms:**
Win 11, Version: 21H2, OS build: 22000.194;

**Prerequisites:**
Have one FxA account without subscription for Mozilla VPN;

**Steps to reproduce:**
1.Install the VPN from the prod website on a win 11 machine;
2. Log in with an account that has a subscription;
3. Go to the my devices menu in the VPN appl

**Expected result:**
The current device is displayed as running windows 11;

**Actual result:**
The current device is displayed as running windows 10;

!mydevices.png|thumbnail!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1107)

data-sync-user:
➤ Madalin Cotetiu commented:

Verified as fixed on Mozilla VPN 2.6.0 (2.202110112011) - STAGING, while using Windows 11 Pro v, 21H2. The current device is displayed as running windows 11.

See the attached screenshot. 

!Screenshot 2021-10-12 102258.png|width=214,height=330!

PR #1982, opened 2021-10-07
Fix Version Detection on Windows 11

strseb:
This is a dirty hack - 
If there is a better api to get that info from lets use that instead :) 

#closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1963


bakulf:
It seems that as soon as windows is officially released, the APIs should work as expected. Can you file an issue to track that?

PR #2002, opened 2021-10-07
WebExt: Add command to request status, report authentication state.

oskirby:
Since the VPN client status is only reported upon a state change, it's possible that a Firefox addon might never receive the status if it is launched after the client has started. So we add another command to the web extension to request it.

While we're at it, we also report the user's authentication state too.


oskirby:
And to throw one extra bit of metadata, I've also added the currently selected location to the web extension.

PR #2003, opened 2021-10-07
Move the telemetry policy view after the authentication

bakulf:


birdsarah:
I do not like this plan at all.

Getting data on how many clients are failing to authenticate will be extremely valuable. This will cut off key data on our funnel.

birdsarah:
Perhaps relevant: my glean PR (#1911) updates the glean initialization and telemetryPolicyShow workflow slightly.

PR #2008, opened 2021-10-10
Remove random lines to make the app work with qt6

bakulf:


bakulf:
Of course, this PR has not been written to land. It's just a list of broken functionalities in QT6.

Issue #2019, opened 2021-10-12
The Mobile and Desktop - Customer Benchmark Surveys are not collecting any responses.


data-sync-user:
The VPN surveys for benchmarking and NPS scores are not collecting responses.

We have 2 surveys that should be triggered within the VPN client after a certain period of time.
1. Desktop triggered 14 days after install (https://survey.alchemer.com/s3/6332395/Mozilla-VPN-Desktop-Survey)
2. Mobile triggered 28 days after install (https://survey.alchemer.com/s3/6353593/Mozilla-VPN-Mobile-Survey). 

To date, there have only been 1 response to each survey (total 2):
1. Desktop on Aug 11 at 5:34 am EST
2. Mobile on Sep 23 at 3:18 am EST
The responses look legitimate. 

We should expect to have more responses by now. Perhaps the survey trigger is not happening.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1157)

data-sync-user:
➤ Basti (Sebastian) Streich commented:

Jessica Cook - It seems there are currently 2 bugs in the client that stop the surveys from being shown, one stopping mobile clients from fetching the list of surveys and another miscalculating the time spend in the app so far, so it’s very unlikely any client has reported itself to be eligible. (i’m now more surprised that we even have 2 responses 🙂) 

I have an open pr, so this should go out in 2.6.

PR #2025, opened 2021-10-13
Fix Issues in Surveyhandling

strseb:
Fixing 2 issues: 
-> The Periodic timer for fetching surveys is 1h - it's super unlikely that on mobile the app will ever be alive this long, so it never gets fetched, moving this to init for now .
-> We're comparing elapsed minutes with seconds -  So we would need to wait a lot for users to be shown this prompt :) 

Also while at it, fixing the settings report not printing string-lists like "consumed-surveys" which confused me quite a bit. 

#closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2019


bakulf:
Can you force the surveys as already shown?

PR #2031, opened 2021-10-14
Qt6 Graphical Effects Compatibility

flozia:
This adds `qrc` files to handle QML compatibility between Qt5 and Qt6. Depending on the Qt Version we can load custom components in order to bridge dropped or changed features.


flozia:
From what I can tell the [Qt5Compat.GraphicalEffects](https://doc-snapshots.qt.io/qt6-dev/graphicaleffects5.html) module does not have a version number. I also could not find a module specific version number in the [qt5compat source](https://github.com/qt/qt5compat).

This might be related to this [Qt Creator bug ("package import requires a version number" when using versionless QML import)](https://bugreports.qt.io/browse/QTCREATORBUG-24533) which seems to be resolved in `Qt Creator 4.13.2`. In `Qt Creator 5.0.1` I’m not seeing the error.

PR #2036, opened 2021-10-14
Clean up functional tests

birdsarah:
Improvements to the functional test code:
* isolation per test
* cleaner tests (setup and teardown is not intermingled with tests)

Does the following:
* basic housekeeping - a package.json, README 
* changes how vpn gets started per test, this removes the llvm reporting, but I discussed with baku ahead of time it's rarely used, there's nothing enforced, and could be accessed by individual devs if needed in the future.
* makes each "it" a test assertion we care about
* no shared state between tests:
  * restart the app
  * clean settings per functional test
* makes testScreenCapture a test suite that is skipped currently as we've deactivated it.

Does not:
* Work on localizeServers, which is not run as a functional test, but used elsewhere.

### DIFF
--- a/tests/functional/testSettings.js
+++ b/tests/functional/testSettings.js
@@ -254,6 +254,7 @@ describe('Settings', function() {
 
 
   it('Checking the get help', async () => {
+    console.log('  Checkpoint a');
     await vpn.waitForElement('settingsGetHelp');
     await vpn.waitForElementProperty('settingsGetHelp', 'visible', 'true');
 
@@ -261,6 +262,7 @@ describe('Settings', function() {
         'settingsView', 'contentY', 'i',
         parseInt(await vpn.getElementProperty('settingsGetHelp', 'y')));
     await vpn.wait();
+    console.log('  Checkpoint b');
 
     await vpn.clickOnElement('settingsGetHelp');
     await vpn.wait();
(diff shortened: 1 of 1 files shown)


strseb:
> tests in one test run that's parallelized....but I have a feeling that won't work as we hardcode the inspector port and the tests need the inspector.

Just a random thought. What if we were instead of using a local binary - were using the wasm client? We could have a webdriver-tab for each client instance, and let the wasm client expose the inspector i/o via some functions. In this way, the test could also controll the lifetime of the client :)

birdsarah:
> Just a random thought. What if we were instead of using a local binary - were using the wasm client? We could have a webdriver-tab for each client instance, and let the wasm client expose the inspector i/o via some functions. 

That could be cool. But it also feels like we're getting further away from testing whether our real clients work and closer to just testing Qt, which I don't seem the point of.....it may be that we want to split these tests into tests that test ui logic, which seem like excellent wasm tests....and tests that test actual platform / client related behavior....and maybe even delete some of the tests that feel like we're just testing copy or Qt. I learned how to write qml unit tests the other day. Maybe we just to qml unit tests for some things too.

> In this way, the test could also control the lifetime of the client :)

This PR goes a way towards that. The client startup / teardown is in a beforeEach / afterEach and there's no reason why every test needs to use it.

birdsarah:
Addressed all your comments @oskirby, added a robustness improvement, fixed the test that was disabled and had a ToDo.

PR #2051, opened 2021-10-15
Bound properties before onComplete

mkestler-rtp:
For Qt to set up the correct behaviors for the window buttons, the max size properties have to be set before the window is created.  They were previously set in onComplete.  This not only disables the minimize button, but corrects the double click behavior on the title bar, and makes the system menu items have the correct state.


lesleyjanenorton:
Looks great and makes sense but one thought/question: If I'm remembering correctly, we moved these into `onComplete` because setting maximum/minimum heights earlier seemed to be contributing to a bug where the dimensions of the app became irreversibly contorted on Windows when moving the app from a device's display to an external display either by mouse-drag or key command (something like Shift + Left/Right Arrow"). If we can just confirm that this doesn't cause a regression there then let's merge!

PR #2070, opened 2021-10-20
Decouple authentication and subscription

mbirghan:


birdsarah:
Thanks for the comment. Merging. I do worry that if someone adds to something at the end of this method but forgets to also add it up here we're in trouble. @bakulf maybe we talk about it in our 1:1.

PR #2080, opened 2021-10-21
Re-enable screencapture tests

birdsarah:
Use new functional test paradigm and run screen capture tests.


lesleyjanenorton:
Chiming in way late but I think there is still a use case for the screen capture tests - it's traditionally been a great way to scan multiple locales quickly for edge case UI breakages that aren't caught elsewhere.

birdsarah:
@lesleyjanenorton just going through old issues. I think that your use case is VERY important and would be better handled by qml tests that measure if there's been an overrun instead of relying on someone to trawl through hundreds of screenshots that will only grow each time we add new features or a language. I'm thinking about opening an epic that (a) removes these tests and (b) puts an ask out there for a new testing framework.

PR #2152, opened 2021-10-27
Banner Fixes

mkestler-mozilla:
The "No Internet" banner was fixed by ignoring NoConnectionError in MozillaVPN::errorHandle when we are in the State::StateConfiming state.  The routes being updated, so we really can't rely on network failures being a true indicator.
The intermittent "Unable To Connects" were caused by the periodic tasks that are scheduled.  I removed error reporting from the two tasks that did report errors on the periodic cycle.  The do not know the state of the app, so it is hard to accurately report from them.  We can look at cleaning some of the overall state up in another issue perhaps.


src/mozillavpn.cpp
```diff
@@ -104,11 +104,14 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
   connect(&m_alertTimer, &QTimer::timeout, [this]() { setAlert(NoAlert); });
 
   connect(&m_periodicOperationsTimer, &QTimer::timeout, [this]() {
-    scheduleTask(new TaskAccountAndServers());
-    scheduleTask(new TaskCaptivePortalLookup());
-    scheduleTask(new TaskHeartbeat());
-    scheduleTask(new TaskSurveyData());
-    scheduleTask(new TaskGetFeatureList());
+    if (controller()->state() != Controller::State::StateConnecting &&
```
bakulf, line 107: In theory this should not be needed. The tasks can be schedule at any time. They do not run during these states because we should have a TaskControllerAction to prevent it.

bakulf:
A task is `completed` when it emits the `completed` signal.  So, a task can have all the async callbacks needed, and then, it must `emit completed`. In the meantime, `MozillaVPN::m_running_task` should point to it. If this is not the case, we have a bug elsewhere. Can you take a look? Thanks!

bakulf (changes requested):
See my latest comment. I suspect the bug is elsewhere. If you want, let's talk on Monday! I would like to have a better understanding of what is going on. Thank you for digging into it!

src/tasks/controlleraction/taskcontrolleraction.h
```diff
@@ -31,10 +32,10 @@ class TaskControllerAction final : public Task {
  private slots:
   void stateChanged();
   void silentSwitchDone();
-
+  void checkStatus();
  private:
   const TaskAction m_action;
-
+  Controller::State m_lastState;
```
bakulf, line 38: ` = Controller::StateOff`

bakulf (approved):

PR #2166, opened 2021-10-30
Fix the unauth support screen

mbirghan:
The way we can open the unauth support screen is on the "Get started" screen which is shown if we are in the "StateInitialize".

### DIFF
--- a/src/ui/views/ViewContactUs.qml
+++ b/src/ui/views/ViewContactUs.qml
@@ -17,6 +17,8 @@ Item {
     // This property is used to cache the emailAddress between the sub-views.
     property string emailAddress: ""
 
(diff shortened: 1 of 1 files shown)


src/ui/views/ViewContactUs.qml
```diff
@@ -102,7 +102,7 @@ Item {
                     ColumnLayout {
                         Layout.fillHeight: true
                         spacing: 24
-                        visible: !VPN.userAuthenticated
+                        visible: !VPN.userAuthenticated || VPN.StateInitialize
```
bakulf, line 105: You want VPN.state === VPN.StateInitialize

bakulf (changes requested):

bakulf (approved):

PR #2171, opened 2021-11-01
Push buttons down on "Leave review" screen

mbirghan:
The buttons on the "Leave review" screen were too high. 

I used a similar way to do this we have on the "Get started" screen.


lesleyjanenorton (approved):
looks good @mbirghan !

PR #2186, opened 2021-11-03
In case guardian does not expose multi-hop/socks-name props, move on - closes VPN-1267

bakulf:


strseb (approved):
Should we in that case maybe disable the related feature flags?

PR #2191, opened 2021-11-03
Taskscheduler

bakulf:


src/taskscheduler.cpp
```diff
@@ -0,0 +1,85 @@
...
+void TaskScheduler::deleteTasks() { maybeCreate()->deleteTasksInternal(); }
+
+// static
+TaskScheduler* TaskScheduler::maybeCreate() {
+  if (!s_taskScheduler) {
+    s_taskScheduler = new TaskScheduler(MozillaVPN::instance());
+  }
+  return s_taskScheduler;
+}
+
```
mkestler-mozilla, line 33: You can use the new C++ magic static here.  It is also thread safe (for free).  Here is a little article on it:
http://www.nuonsoft.com/blog/2017/08/10/implementing-a-thread-safe-singleton-with-c11-using-magic-statics/comment-page-1/

mkestler-mozilla (approved):

PR #2200, opened 2021-11-04
Fix the URLs for a few image resources

bakulf:


flozia (approved):
Thanks for catching these!

PR #2205, opened 2021-11-04
Extend the adjustSDK package filtering

bakulf:


mbirghan (changes requested):
Looks good but can you add the the new fields added to `allowList` to `adjust.md` as well?

mbirghan (approved):
Great!

PR #2210, opened 2021-11-05
Glean upgrade

birdsarah:
Includes glean bug fix that we were experiencing https://bugzilla.mozilla.org/show_bug.cgi?id=1738118


birdsarah:
The taskcluster clang fail is an out of memory error.

oskirby (approved):

bakulf (approved):

Issue #2177, opened 2021-11-02
release note for v2.6


data-sync-user:
We need a release note file under `docs/releasenotes` for 2.6.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1261)

PR #2211, opened 2021-11-05
Add changelog for v2.6.0

oskirby:
Closes: #2177

### DIFF
--- a/docs/releasenotes/v2.6.md
+++ b/docs/releasenotes/v2.6.md
@@ -1,0 +1,16 @@
+# v2.6 Release Notes
+
+## New Features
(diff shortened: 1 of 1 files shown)


birdsarah (approved):
Approving this.

I was looking at the changelog though and we've had 210 commits between 2.5.1 and 2.6.0. 

I was thinking that a brief summary like this isn't capturing our efforts well. Perhaps we should work on a process where we add to a running changelog as we commit PRs (that the dev puts in a note about during PR review) and then we can consolidate it at the end of a release cycle.

oskirby:
I completely agree, this changelog sucks and it feels like I'm not doing justice to the work that went into this release, but the commit history isn't easy to parse either. I would very much be in favour of a process where we update the notes as we go.

PR #2222, opened 2021-11-08
Temporary fix to reduce the "unstability/connectivity" alert for 2.6

bakulf:
See VPN-1296


oskirby (approved):
I think this looks good, but we should keep a JIRA ticket open for a more dynamic fix Ina future release

mkestler-mozilla (approved):

PR #2224, opened 2021-11-08
Fix string casing in QML files

### LINKED ISSUE #1858: Multiple areas in app where sentence case is not used
Sentence case should be used everywhere within the app, I've found multiple places where it isn't being used and I know one of them is a regression as we've fixed it before. 

The affected screens: 

* App permissions (search placeholder and button)
* Get help (Help centre menu item)
* Give feedback (Dropdown placeholder) 

!Group 1.png|thumbnail!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1033)

lesleyjanenorton:
Fixes #1858

### DIFF
--- a/src/ui/settings/ViewAppPermissions.qml
+++ b/src/ui/settings/ViewAppPermissions.qml
@@ -17,11 +17,11 @@ import telemetry 0.24
 Item {
     id: root
 
-    //% "Search Apps"
+    //% "Search apps"
     //: Search bar placeholder text
     property string searchApps: qsTrId("vpn.protectSelectedApps.searchApps")
 
-    //% "Add Application"
+    //% "Add application"
     //: Button label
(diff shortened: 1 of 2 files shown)


flodolo (approved):

flodolo:
Interestingly enough, these changes don't show up in the export to the l10n repository. I assume because there's a conflict between the branches, and the order of branches/import is not significant :-| 

I [manually fixed](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/8cc9a81ffbf21e4504528b01be2eecf5cac4118c) en-CA and en-GB, but at some point these changes will show up when we remove the old branches. We need to make sure those strings still show up localized for all languages on `main` though.

PR #2230, opened 2021-11-09
Fix CaptivePortal Detection on Windows

strseb:
Problem: We now use the "critical Ping Error" on Windows, meaning ConnectionHealth might deliver useless data. Thus, we will never get a "bad" connection → so we will never check captive portal for we are in this state. 
My idea was to hook Into the Network Watcher. So every time the user changes their access point and are connected to the VPN, we will start a CP check. 

If this approach is okay, I would update the Linux/macOS network watchers as well in a follow-up pr. Maybe we might even be able to untie captive-portal from relying on the ping at all :)

### DIFF
--- a/src/captiveportal/captiveportalrequest.cpp
+++ b/src/captiveportal/captiveportalrequest.cpp
@@ -57,13 +57,14 @@ void CaptivePortalRequest::createRequest(const QUrl& url) {
   NetworkRequest* request = NetworkRequest::createForCaptivePortalDetection(
       this, url, CAPTIVEPORTAL_HOST);
 
-  
-  connect(request,&NetworkRequest::requestRedirected,[this](NetworkRequest* request, const QUrl& url){
-    // In Case the Captive Portal request Redirects, we 100% have one. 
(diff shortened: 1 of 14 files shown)


src/captiveportal/captiveportalrequest.cpp
```diff
@@ -57,6 +57,13 @@ void CaptivePortalRequest::createRequest(const QUrl& url) {
   NetworkRequest* request = NetworkRequest::createForCaptivePortalDetection(
       this, url, CAPTIVEPORTAL_HOST);
 
+  
+  connect(request,&NetworkRequest::requestRedirected,[this](NetworkRequest* request, const QUrl& url){
+    // In Case the Captive Portal request Redirects, we 100% have one. 
+    logger.info() << "Portal Detected -> Redirect to " <<url.toString();
+    request->abort();
+    onResult(PortalDetected);
+  });
```
strseb, line 66: This popped up while testing, when we have a redirect to something that requires dns it is not available (as we might use the tunneled one) - the request will fail with "hostnotfound". The intent to redirect is all we need :)

bakulf (approved):

PR #2176, opened 2021-11-02
Do not delete input characters automatically when limit is exceeded

flozia:
Resolves ~~#2174~~ #2158

PR #2243, opened 2021-11-10
Fix: Contact form character limit

### LINKED ISSUE #2158: Inserting extra characters after the form reached its limit deletes the characters at the end
**Affected VPN version:**

* 2.6.0~rc2 (2.6.0~rc2-focal1);

**Affected Platforms:**

* All;

**Preconditions:**

* Have Mozilla VPN installed;

**Steps to reproduce:**
1. Open the VPN;
2. Access the “Get help” option;
3. Access the “Contact us”;
4. Write 1000 characters in the “Describe issue” textbox;
5. Move the cursor from the end anywhere else and try to type something;

**Expected result:**

* You cannot type because the max character limit is reached;

**Actual result:**

* You can write and the characters at the end are deleted;

**Additional notes:**

* This reproduces for authenticated users too;
* See the attached screen record;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1250)

flozia:
The default QML TextArea component does not provide a default prop for a character limit. In PR #2176 I accidentally only removed our previous approach to remove overflowing character dynamically. This previous method of limiting the maximum input removed characters from the end of the existing string.

This PR removes newly added characters that exceed the maximum input length for VPNTexArea instead of removing characters from the end of the string.

Resolves #2158


lesleyjanenorton (approved):

src/ui/components/forms/VPNTextArea.qml
```diff
@@ -76,6 +78,28 @@ Item {
...
+
+                // Remove overflowing chars in case text input
+                // exceeds the maximum length
+                if (exceedsCharLimit) {
+                    // Remember cursor position
+                    const prevCursorPosition = textArea.cursorPosition - textInputLength;
+                    // Remove overflowing chars
+                    const trimmedString = removeCharsInRange(text, prevCursorPosition, prevCursorPosition + textInputLength);
+                    textArea.text = trimmedString;
+                    // Restore cursor previous cursor position
```
flozia, line 99:
```suggestion
                    // Restore previous cursor position
```

src/ui/components/forms/VPNTextArea.qml
```diff
@@ -34,7 +34,8 @@ Item {
         }
 
         TextArea.flickable: TextArea {
-            property var maxCharacterCount: 1000
+            property int maxCharacterCount: 1000
+            property int textLength: textArea.text.length
```
bakulf, line 38: why a property here for this? You can have this as a variable in the function.

src/ui/components/forms/VPNTextArea.qml
```diff
@@ -76,6 +78,28 @@ Item {
                 id: textAreaStates
                 itemToTarget: textArea
             }
+
+            function removeCharsInRange(string, startIndex, endIndex) {
+                return string.slice(0, startIndex) + string.slice(endIndex);
+            }
+
+            function handleOnTextChanged(text) {
```
bakulf, line 86: what is 'text' here? It's the new added text or the entire one in the textarea?

flozia: It’s the entire text in the textarea after the text changed.

src/ui/components/forms/VPNTextArea.qml
flozia: Storing the text length in a property here lets us compare the previous text with the updated text in `handleOnTextChanged`. Doing this in the function would mean that `textLength` and `updatedTextLength` always have the same value.

src/ui/components/forms/VPNTextArea.qml
```diff
@@ -76,6 +78,28 @@ Item {
...
+                const updatedTextLength = text.length;
+                const textInputLength = updatedTextLength - textLength;
+                const exceedsCharLimit = textLength + textInputLength > textArea.maxCharacterCount;
+
+                // Remove most recently added characters in case
+                // the text input exceeds the maximum length
+                if (exceedsCharLimit) {
```
bakulf, line 93: I think we can do this without having `textLength`. Let me try:

```
if (text.length > textArea.maxCharacterCount) {
    // The cursor is reset when `text` changes. Let's store the old value, normalized if needed.
    const cursorPos = textArea.cursorPosition > textArea.maxCharacterCount ? textArea.maxCharacterCount : textArea.cursorPosition;
    textArea.text = text.slice(0, textArea.maxCharacterCount);
    textArea.cursorPosition = cursorPos;
}
```

This should work.

bakulf (changes requested):

src/ui/components/forms/VPNTextArea.qml
```diff
@@ -76,6 +78,28 @@ Item {
...
             }
+
+            function removeCharsInRange(string, startIndex, endIndex) {
+                return string.slice(0, startIndex) + string.slice(endIndex);
+            }
+
+            function handleOnTextChanged(text) {
+                const updatedTextLength = text.length;
+                const textInputLength = updatedTextLength - textLength;
+                const exceedsCharLimit = textLength + textInputLength > textArea.maxCharacterCount;
```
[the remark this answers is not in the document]
flozia: Yes it is.

birdsarah:
@flozia I have good news for this PR :D

The qml tests PR is now merged. The logic that you have in this PR seems like it would be an excellent candidate for some unit tests. There are some examples under `/tests/qml/tst_**.qml` that you can pull from, but I'm also happy to help coach you through it. 

Maybe we even want to do a big live coding session with @lesleyjanenorton so that we can work through together what these tests would look like and take the opportunity to learn about how to do qml tests along the way. The tests you'd add for this PR would certainly be novel and not like tests I've already written.

bakulf (approved):

bakulf:
Now it makes sense!

PR #2247, opened 2021-11-10
Build qt with sqlite for windows

mkestler-mozilla:
This was only affecting local builds.  Will not cause issues with releases or CI builds.


bakulf:
typo in the title of the PR.

bakulf (approved):
This is already landed, right?

mkestler-mozilla:
Mistakenly thought the debug and release flag had not landed.  It was the verbose flag that triggered the delta.  Closing, we don't want verbose on as the default.

PR #2249, opened 2021-11-10
Add an "Check for Updates Button"

strseb:
This PR exposes Release Manager to QML and adds a manual option to check for updates :) 

Mozilla_VPN_2021-11-10_19-54-45_Trim_SparkVideo


lesleyjanenorton (approved):
Nice! This is great! I think this is good to go as soon as the conflict in strings.yaml is resolved.
One tiny nit that definitely doesn't need to block: `updateStatusUpToDate.svg` looks pixelated. Removing the `anchors.fill: all` on L303 of VPNAboutUs.qml seems to fix the pixelation but then goofs up the sizing. If there is an quick/easy fix, awesome. If not, we can do it in a follow up.

src/constants.h
```diff
@@ -38,7 +38,7 @@ constexpr int RECENT_CONNECTIONS_MAX_COUNT = 5;
 #endif
 
 // Let's refresh the IP address any 10 minutes (in milliseconds).
-CONSTEXPR(uint32_t, ipAddressTimerMsec, 600000, 10000, 0)
+CONSTEXPR(uint32_t, ipAddressTimerMsec, 600000, 300000, 0)
```
bakulf, line 41: unrelated change :)

src/constants.h
```diff
@@ -47,23 +47,23 @@ CONSTEXPR(uint32_t, checkStatusTimerMsec, 1000, 1000, 0)
...
 CONSTEXPR(uint32_t, statusIconAnimationMsec, 200, 200, 0)
 
 // How often glean pings are sent
-CONSTEXPR(uint32_t, gleanTimeoutMsec, 1200000, 4000, 0)
+CONSTEXPR(uint32_t, gleanTimeoutMsec, 1200000, 600000, 0)
 
 // How often we check the surveys to be executed (no network requests are done
 // for this check)
-CONSTEXPR(uint32_t, surveyTimerMsec, 300000, 4000, 0)
+CONSTEXPR(uint32_t, surveyTimerMsec, 300000, 150000, 0)
```
bakulf, line 66: Revert all these changes. Or tell us more why you want to include them

src/mozillavpn.h
```diff
@@ -126,6 +126,7 @@ class MozillaVPN final : public QObject {
   Q_PROPERTY(bool debugMode READ debugMode CONSTANT)
   Q_PROPERTY(QString currentView READ currentView WRITE setCurrentView NOTIFY
                  currentViewChanged)
+  Q_PROPERTY(QObject* releaseMonitor READ releaseMonitor CONSTANT)
```
bakulf, line 129: Don't do this. Expose the releaseMonitor as VPNReleaseMonitor in `commandui.cpp`

src/releasemonitor.cpp
```diff
@@ -39,7 +39,10 @@ void ReleaseMonitor::runInternal() {
           &ReleaseMonitor::updateRequired);
   connect(updater, &Updater::updateRecommended, this,
           &ReleaseMonitor::updateRecommended);
-  connect(updater, &QObject::destroyed, this, &ReleaseMonitor::releaseChecked);
+  connect(updater, &QObject::destroyed, [this]() {
+    emit updateCheckResult(false);
```
bakulf, line 43: in this way, when the object is released, we go back to `updateCheckResult = false` and we will open the no-update popup :/

bakulf: I guess you need to change the `Updater` class and expose a boolean flag `updateRequiredOrRecommended()` to be called here.

src/releasemonitor.h
```diff
@@ -16,13 +16,16 @@ class ReleaseMonitor final : public QObject {
...
 
-  void runSoon();
+  Q_INVOKABLE void runSoon();
 
   void update();
 
  signals:
   // for testing
   void releaseChecked();
+  // Is fired once balrog was checked
```
bakulf, line 26: balrog or verify API

bakulf (changes requested):
We need a test to check the type of "popup" we show.

src/constants.h
strseb: Oh no this was not supposed to be there 🙈 I'll remove it before merging

src/constants.h
```diff
@@ -71,6 +71,8 @@ CONSTEXPR(uint32_t, surveyTimerMsec, 300000, 4000, 0)
   inline type functionName() { return inProduction() ? prod : beta; }
 
 constexpr const char* API_PRODUCTION_URL = "https://vpn.mozilla.org";
+constexpr const char* API_STAGING_URL =
+    "https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net";
```
bakulf, line 75: This is unrelated :)

src/releasemonitor.cpp
```diff
@@ -39,7 +39,15 @@ void ReleaseMonitor::runInternal() {
...
   connect(updater, &Updater::updateRecommended, this,
           &ReleaseMonitor::updateRecommended);
-  connect(updater, &QObject::destroyed, this, &ReleaseMonitor::releaseChecked);
+  connect(updater, &Updater::updateRequired, this,
+          &ReleaseMonitor::updateRequiredOrRecommended);
+  connect(updater, &Updater::updateRecommended, this,
+          &ReleaseMonitor::updateRequiredOrRecommended);
+  connect(updater, &Updater::updateNotAvailable, this,
+          &ReleaseMonitor::updateNotAvailable);
+  connect(updater, &QObject::destroyed, [this]() {
```
bakulf, line 48: This can be as it was before: `connect(updater, &QObject::destroyed, this, &ReleaseMonitor::releaseChecked);` or it does not work?

src/ui/components/VPNAboutUs.qml
```diff
@@ -161,4 +160,207 @@ Item {
...
+            sourceSize.height: Theme.iconSize * 1.5
+            sourceSize.width: Theme.iconSize * 1.5
+            visible: true
+            z:6
+
+            SequentialAnimation {
+                id: updateButtonImageAnimation
+                running: false
+                PropertyAnimation {
+                               target: updateButtonImage
```
bakulf, line 197: indentation

src/ui/components/VPNList.qml
```diff
@@ -11,7 +11,7 @@ ListView {
     id: list
 
     property var listName
-
+    height:contentHeight
```
bakulf, line 14: is this related? Does it change any other views?

src/update/balrog.cpp
```diff
@@ -300,6 +300,8 @@ bool Balrog::processData(const QByteArray& data) {
               // will be aborted, but the signal emitted.
               if (request->statusCode() == 200) {
                 emit updateRecommended();
+              } else {
+                emit updateNotAvailable();
```
bakulf, line 304: If you see, `Balrog::processData` has many reasons to do not call this signal: it has many `return false`. Same for `VersionApi`. I would suggest a different approach.

1. a method to know the status of the updater:
```
Class Updater ... {
  bool recommendedOrRequired() const { return m_recommendedOrRequired; }
  private:
  m_recommendedOrRequired = false;
}
```

2. let's use signals to set that boolean:
```
Updater::Updater() {
  connect(this, &updateRecommended, [this] { m_recommendedOrRequired = true; });
  connect(this, &updateRequired, [this] { m_recommendedOrRequired = true; });
 ...
```

3. in `ReleaseMonitor` do:
```
connect(updater, &QObject::destroyed, [this, updater]() {
  if (!updater->recommendedOrRequired()) emit notAvailable(); });
```

Thoughts?

src/update/updater.h
```diff
@@ -22,6 +22,7 @@ class Updater : public QObject {
  signals:
   void updateRequired();
   void updateRecommended();
+  void updateNotAvailable();
```
bakulf, line 25: No need for this.

src/update/versionapi.cpp
```diff
@@ -39,6 +39,7 @@ void VersionApi::start() {
             logger.debug() << "Request completed";
 
             if (!processData(data)) {
+              emit updateNotAvailable();
```
bakulf, line 42: remove this too.

src/update/versionapi.cpp
```diff
@@ -116,6 +117,7 @@ bool VersionApi::processData(const QByteArray& data) {
     logger.debug() << "Update recommended.";
     emit updateRecommended();
   }
+  emit updateNotAvailable();
```
bakulf, line 120: And this too.

bakulf (changes requested):

src/ui/components/VPNList.qml
strseb: The list is just used in about-us. We need to limit the hight to it's content children otherwise the list grows to fill the parent.

src/update/balrog.cpp
strseb: Good point!

src/ui/components/VPNAboutUs.qml
```diff
@@ -161,4 +160,207 @@ Item {
...
+                    duration: 2000
+                    loops: Animation.Infinite
+                }
+            }
+        }
+    }
+    Connections {
+        target: VPNReleaseMonitor
+        function onUpdateRequiredOrRecommended() {
+            if(!listenForUpdateEvents){return;}
```
bakulf, line 210: indentation

src/ui/components/VPNAboutUs.qml
```diff
@@ -161,4 +160,207 @@ Item {
...
+            updateButtonImageAnimation.stop();
+            updateAvailablePopup.open()
+            listenForUpdateEvents=false;
+        }
+        function onUpdateNotAvailable() {
+            if(!listenForUpdateEvents){return;}
```
bakulf, line 216: indentation

bakulf (approved):

flodolo:
Any idea why the diff for strings.yaml involves the whole file? Was the line-ending changed?

Sadly, this also killed blame in the file :-(

translations/strings.yaml
```diff
@@ -1,288 +1,297 @@
...
+    value: General Notifications
+    comment: A notification channel; so the user can subscribe to "General Notifications"
+
+updateButton:
+  checkForUpdateButtonText: Check for Updates
+  titleNoUpdate: You are up to date
+  descriptionNoUpdate: There is no new update available at this time
+  actionNoUpdate: Okay
+  titleOnUpdate: Updates available
+  descriptionOnUpdate: Click to update to the latest version of the application
```
flodolo, line 292: @chrisrooneys 
Noticed this while translating. Does this show up only on desktop clients? If not, "click" might not be the best choice.

Issue #2248, opened 2021-11-10
Glean local storage broken on Ubuntu/21.10


oskirby:
Just reviewing the recent changes to glean for the 2.6.0 release, and it seems that we are getting errors on some Ubuntu versions as follows:

```
[10.11.2021 10:37:23.675] Debug: (main - MozillaVPN) main window loaded
[10.11.2021 10:37:23.676] Debug: (main - MozillaVPN) Initializing Glean
[10.11.2021 10:37:23.676] Debug: Initializing glean with channel set to: production (main.qml:301, onInitializeGlean)
[10.11.2021 10:37:23.676] Debug: (Glean.undefined) Error while attempting to access LocalStorage.
 TypeError: Cannot call method 'openDatabaseSync' of undefined (glean.lib.js:1, a)
[10.11.2021 10:37:23.678] Debug: (Glean.undefined) Error while attempting to access LocalStorage.
 TypeError: Cannot call method 'openDatabaseSync' of undefined (glean.lib.js:1, a)
[10.11.2021 10:37:23.678] Debug: (Glean.undefined) Error while attempting to access LocalStorage.
 TypeError: Cannot call method 'openDatabaseSync' of undefined (glean.lib.js:1, a)
[10.11.2021 10:37:23.678] Debug: (Glean.undefined) Error while attempting to access LocalStorage.
 TypeError: Cannot read property 'openDatabaseSync' of undefined (glean.lib.js:1, a)
[10.11.2021 10:37:23.678] Debug: (Glean.undefined) Error while attempting to access LocalStorage.
 TypeError: Cannot read property 'openDatabaseSync' of undefined (glean.lib.js:1, a)
```

This was reproduced by installing the `mozillavpn` package from the [Testing PPA](https://launchpad.net/~okirby/+archive/ubuntu/mozilla-vpn-testing) and I suspect that the problem may be due to a missing package dependency.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1329)

birdsarah:
sqlite also most likely :(

oskirby:
Looks like `qml-module-qtquick-localstorage` was the culprit, but I will give this a test on the other Ubuntu variants first before assembling a PR

PR #2250, opened 2021-11-10
Add qml-module-qtquick-localstorage deps for Ubuntu

oskirby:
For Ubuntu releases where we can make use of the system-provided Qt libraries, we need to add a runtime dependency on the `qml-module-qtquick-localstorage` package in order for the Glean telemetry to be functional.

Closes: #2248


birdsarah (approved):
lgtm - verified this includes sqlite - https://packages.ubuntu.com/focal/qml-module-qtquick-localstorage

PR #2251, opened 2021-11-10
Added a settling period to VPN connect/disconnect

mkestler-mozilla:
I would have made this a state, but since the controller is tied to the UI, it made things difficult.  This works as well and is probably simpler.  Basic flag and timer when we connect and disconnect to ignore errors while the routes and interface become stable.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -39,6 +39,8 @@
 #  include "platforms/dummy/dummycontroller.h"
 #endif
 
+constexpr auto SETTLE_TIMEOUT = 3000;
+
 constexpr const uint32_t TIMER_MSEC = 1000;
 
 // X connection retries.
@@ -367,6 +369,9 @@ void Controller::connectionConfirmed() {
   }
 
   setState(StateOn);
+
+  startUnsettledPeriod();
+
   emit timeChanged();
 
   if (m_nextStep != None) {
@@ -403,6 +408,8 @@ void Controller::connectionFailed() {
   m_impl->deactivate(ControllerImpl::ReasonConfirming);
 }
 
+bool Controller::isUnsettled() { return !m_settled; }
+
 void Controller::disconnected() {
   logger.debug() << "Disconnected from state:" << m_state;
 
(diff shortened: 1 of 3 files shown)


bakulf:
I like this solution! But I have a different implementation to suggest. Let's see if it makes sense:
- if we have tasks for everything, all the network requests start from tasks (let's say that this is true for now, or we can work to make it happen) - we can enforce it with assertions.
- we can extend the ControllerAction task to have a "settle" timeout before calling "completed".

The reason why I suggest this is that with your approach we "hide" errors. I would like to see if we can avoid them instead.
What do you think?

mkestler-mozilla:
> 
> 
> I like this solution! But I have a different implementation to suggest. Let's see if it makes sense:
> 
>     * if we have tasks for everything, all the network requests start from tasks (let's say that this is true for now, or we can work to make it happen) - we can enforce it with assertions.
> 
>     * we can extend the ControllerAction task to have a "settle" timeout before calling "completed".
> 
> 
> The reason why I suggest this is that with your approach we "hide" errors. I would like to see if we can avoid them instead. What do you think?

I don't see this as hiding the errors.  I think it is a more defensive approach to the API.  We can go through and have all the tasks check for the unsettled flag now, but as things are added or changed we may not catch new tasks not respecting the API.  I see it as a guarantee that the unsettled period is honored.  Does that seem reasonable?
I don't see any issue moving the timer to the controller action.

oskirby:
Just my two cents.

I think that this is not the right approach in the long run, and we need to do a better job of managing the state transitions from `Off` -> `Connecting` -> `Connected`. Ideally, we should only transition from `Connecting` to `Connected` once the network has stabilized and therefore we shouldn't need a settling time when we can just look at the state to determine if a network error should be considered transient or fatal.

Likewise, a similar argument could be made for the `Connected` -> `Disconnecting` -> `Off` transitions.

mkestler-mozilla:
> 
> 
> Just my two cents.
> 
> I think that this is not the right approach in the long run, and we need to do a better job of managing the state transitions from `Off` -> `Connecting` -> `Connected`. Ideally, we should only transition from `Connecting` to `Connected` once the network has stabilized and therefore we shouldn't need a settling time when we can just look at the state to determine if a network error should be considered transient or fatal.
> 
> Likewise, a similar argument could be made for the `Connected` -> `Disconnecting` -> `Off` transitions.

PR #2253, opened 2021-11-11
Add QML tests

### LINKED ISSUE #1262: Update Glean.js to v0.21.0
Hey folks, the new release of Glean.js is out ([v0.16.0](https://github.com/mozilla/glean.js/releases/tag/v0.16.0)) and has many big changes regarding QML. Here are the highlights and suggestions on how to adress them:

#### 1. Glean.js now provides a default `httpClient` for QML
This means it is not mandatory to add a custom `httpClient`, as [the VPN app has right now](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/20db046b7de8f2613fffe26c2d4f2e4902fb2f49/src/ui/main.qml#L70-L90). The provided default client [uses XHR]() just like the custom one provided by the app right now, so I suggest you simply drop the custom one and use the default.

#### 2. Glean.js will now gzip pings upon submission
This only requires action in case you decide to continue using a custom uploader / want to implement another one.

Implementing the gzipping required that we change the signature of the `Uploader.post` function to accept both a `String` and `UInt8Array`. That means any custom uploader implementation will need to be updated to support the enhanced signature. Note that for XHR specifically it is necessary to change the request depending on the type we are sending over the wire. See [default uploader implementation](https://github.com/mozilla/glean.js/blob/main/glean/src/platform/qt/uploader.ts#L52-L56) for specifics.

#### 3. Glean.js now has persistent storage for QML
Prior to this version Glean.js did not persist metrics and pings throughout sessions, which meant data not sent in a single session were lost and most importantly meant that every new run generated a new client id.

That is not the case from v0.16.0 onwards. This changes dramatically the shape of the data you will get from now on. An approach here is, after update, to bisect data based on the [`telemetry_sdk_build` metric](https://mozilla.github.io/glean/book/user/pings/index.html?highlight=telemetry_sdk#the-client_info-section), which contains a string with the Glean version a ping was built on. Another approach, if possible, is to wipe all data coming from versions prior to v0.16.0, since it will be nearly impossible to correlate data from before and after this version.

Anyways, for this one I cannot really recommend what would be the best approach. I am available for discussion and questions whenever necessary :grinning:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-687)

birdsarah:
Fixes #1262 

Goals:
* add infrastructure for writing qml tests
* add glean tests


birdsarah:
After adding the settings holder, the tests are failing spectacularly as follows:

```
********* Start testing of qml_tests *********
Config: Using QtTest library 5.15.2, Qt 5.15.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 5.3.1 20160406 (Red Hat 5.3.1-6)), ubuntu 21.04
PASS   : qml_tests::VPNAboutUsTests::initTestCase()
PASS   : qml_tests::VPNAboutUsTests::test_releaseVersion()
PASS   : qml_tests::VPNAboutUsTests::cleanupTestCase()
QFATAL : qml_tests::UnknownTestFunc() ASSERT: "s_instance" in file ../../src/settingsholder.cpp, line 37
FAIL!  : qml_tests::UnknownTestFunc() Received a fatal error.
   Loc: [Unknown file(0)]
Totals: 3 passed, 1 failed, 0 skipped, 0 blacklisted, 50ms
********* Finished testing of qml_tests *********
[1]    2095575 IOT instruction (core dumped)  ./tests/qml/qml_tests
```

I'm sure it's something really obvious, but I can't see it. I thought https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2253/files#diff-2a13687937c4188e42bee7f3f9c4941c2faa0335238788aad32651bdc9af480dR24 would be enough to instantiate the settings holder, but apparently s_instance isn't set :(

bakulf:
The SettingsHolder is not initialized correctly. You have added it to the `void qmlEngineAvailable(QQmlEngine* engine)`, but the tests do not run into that method. So, the `SettingsHolder` object is created and "destroyed" when it runs out of scope. I suggest doing this: add `SettingsHolder` as a member variable. In this way it will exist for the whole execution.

birdsarah:
@bakulf or @brizental would appreciate a consult. This is coming together nicely but I'm stuck on hopefully a final hurdle. 

See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2253/files#diff-c0d60f7414a96dd6cae2fd18f3c1fe709918e3e299c6872e18c83e1c6827f847R55

This is failing with the error 

```
FAIL!  : qml_tests::MainTestsGleanEvents::test_onRecordGleanEventRecordsAppropriateSample() Uncaught exception: Sample is not defined
   Loc: [/home/sbird/Dev/mozilla/mozilla-vpn-client/tests/qml/tst_mainWindowGleanEvents.qml(55)]
```

I don't understand. I thought that if `Glean` was in scope, which it is, then `Sample` would be in scope. But as you can see it isn't. 

I ran with `QML_IMPORT_TRACE=1`; it's long but here's some pertinant lines

```
DEBUG : qml_tests::UnknownTestFunc() QQmlImports(qrc:/ui/main.qml)::importExtension: loaded ":/glean/telemetry/qmldir"
QDEBUG : qml_tests::UnknownTestFunc() QQmlImports(qrc:/glean/telemetry/pings.js)::addLibraryImport: "org.mozilla.Glean" 0.24 as "Glean"
QDEBUG : qml_tests::UnknownTestFunc() QQmlImports(qrc:/glean/telemetry/pings.js)::importExtension: loaded ":/glean/org/mozilla/Glean/qmldir"
QDEBUG : qml_tests::UnknownTestFunc() QQmlImports(qrc:/glean/telemetry/sample.js)::addLibraryImport: "org.mozilla.Glean" 0.24 as "Glean"
QDEBUG : qml_tests::UnknownTestFunc() QQmlImports(qrc:/glean/telemetry/sample.js)::importExtension: loaded ":/glean/org/mozilla/Glean/qmldir"
```

which also make me think that Sample should be there.

@bakulf maybe it's a test setup issue?
@brizental when does Sample get instantiated? Do I need to wait for it?

bakulf:
> @bakulf maybe it's a test setup issue? @brizental when does Sample get instantiated? Do I need to wait for it?

The Samples are generated by `glean_parser` when the `scripts/generate_glean.py` script runs. They are stored in the `glean/telemetry` folder. The Glean team did an amazing thing generating the `qmldir` file too. So, what is missing in your test is 1 line only: `import telemetry 0.24`.

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "../../src/mozillavpn.h"
```
bakulf, line 4: empty line between 3 and 4?

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "../../src/settingsholder.h"
```
bakulf, line 5: You have `INCLUDEPATH += ../../src` in `qml.pro`. You don't need `../../`.

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+TestHelper* TestHelper::instance() {
+  static TestHelper* s_instance = new TestHelper();
+  return s_instance;
+}
+
+bool TestHelper::mainWindowLoadedCalled() {
+  // I don't understand why I need to call instance()
+  // here, would appreciate a c++ lesson and/or correction
+  // of my code.
+  return instance()->m_mainWindowLoadedCalled;
```
bakulf, line 23: You have spot a bug in your code. You don't need to use `instance()->..` here. This method should be:
```
bool TestHelper::mainWindowLoadedCalled() const { return m_mainWindowLoadedCalled; }
```
(note the `const`).

The reason why you need `instance()` is the following:

1. `TestHelper` is called the first time in the main, when this line runs: `QUICK_TEST_MAIN_WITH_SETUP(qml_tests, TestHelper)`. Let's call this object "A"
2. This object is created but it's not set into `s_instance` because `::instance()` is not called.
3. Later on, in your code, you run `MozillaVPN::mainWindowLoaded`
4. ... in here you call `TestHelper::instance`
5. ... and here you run `new TestHelper`, creating a new instance of `TestHelper`. Let's call this "B".
6. `MozillaVPN::mainWindowLoaded` called `setMainWindowLoadedCalled` using "B".
7. But when `TestHelper::mainWindowLoadedCalled` is executed by the QML engine, "A" is used. And "A" returns false.

The fix I suggest is the following:
1. The `TestHelper` CTOR should set the `s_instance`: (ignore the indentation please)
```
namespace { TestHelper* s_instance = nullptr; }
TestHelper::TestHelper() {
  Q_ASSERT(s_instance == nullptr); // In this way we are 100% sure there is only 1 instance of TestHelper around.
  s_instance = this;
  // .. all the rest of the code
}
```

2. Unset the `s_instance` in the DTOR:
```
TestHelper::~TestHelper() {
  Q_ASSERT(s_instance == this); // If it's not us, what is this?
  s_instance = nullptr;
}
```

3. a simple `::instance()` static method:
```
TestHelper* TestHelper::instance() {
  Q_ASSERT(s_instance); // We _must_ have an object around.
  return s_instance;
}
```

(Without my silly comments)

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void TestHelper::setMainWindowLoadedCalled(bool val) {
+  m_mainWindowLoadedCalled = val;
+}
+
+void TestHelper::triggerInitializeGlean() {
```
bakulf, line 30: check if this and the following can be `const`.

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+void TestHelper::triggerAboutToQuit() {
+  emit MozillaVPN::instance()->aboutToQuit();
+}
+
+// The next two functions duplicate mozillavpn.h logic, but I'm not
+// sure how else to generate an "expected" value for tst_mainWindowGlean.qml
+// as it needs to be dynamic based on the platform people are running tests on.
+// At the least this should fail give someone pause for thought if they change
+// the mozillavpn.h logic.
+QString TestHelper::osVersion() { return QSysInfo::productVersion(); }
```
bakulf, line 51: this can be `const` too.

bakulf: Can you use MozillaVPN as a singleton and write this as:
`QString TestHelper::osVersion() const { return MozillaVPN::instance()->osVersion(); }` ?

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+QString TestHelper::architecture() {
+  return QSysInfo::currentCpuArchitecture();
```
bakulf, line 53: Same here.

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void TestHelper::qmlEngineAvailable(QQmlEngine* engine) {
+  engine->addImportPath("qrc:///compat");
+  engine->addImportPath("qrc:///components");
+  engine->addImportPath("qrc:///glean");
+  engine->addImportPath("qrc:///themes");
+  engine->addImportPath("qrc:///");
```
bakulf, line 61: do we need this too?

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+      [this](QQmlEngine*, QJSEngine*) -> QObject* {
+        QObject* obj = this;
+        QQmlEngine::setObjectOwnership(obj, QQmlEngine::CppOwnership);
+        return obj;
+      });
+
+  qmlRegisterSingletonType<MozillaVPN>(
+      "Mozilla.VPN", 1, 0, "VPNCloseEventHandler",
+      [this](QQmlEngine*, QJSEngine*) -> QObject* {
+        QObject* obj = &this->closeEventHandler;
```
bakulf, line 96: `QObject* obj = &closeEventHandler; `

tests/qml/helper.cpp
```diff
@@ -0,0 +1,108 @@
...
+        QQmlEngine::setObjectOwnership(obj, QQmlEngine::CppOwnership);
+        return obj;
+      });
+
+  qmlRegisterSingletonType<MozillaVPN>(
+      "Mozilla.VPN", 1, 0, "VPNWhatsNewModel",
+      [this](QQmlEngine*, QJSEngine*) -> QObject* {
+        QObject* obj = this->whatsNewModel();
```
bakulf, line 104: Why here a method and for `closeEventHandler` a variable member? This can be: `QObject* obj = m_whatsNewModel;`

tests/qml/helper.h
```diff
@@ -0,0 +1,44 @@
...
+  static TestHelper* instance();
+  Q_INVOKABLE void triggerAboutToQuit();
+  Q_INVOKABLE void triggerInitializeGlean();
+  Q_INVOKABLE void triggerRecordGleanEvent(const QString& event);
+  Q_INVOKABLE void triggerSetGleanSourceTags(const QStringList& tags);
+  Q_INVOKABLE QString osVersion();
+  Q_INVOKABLE QString architecture();
+  Q_PROPERTY(bool mainWindowLoadedCalled READ mainWindowLoadedCalled)
+
+  CloseEventHandler closeEventHandler;
```
bakulf, line 29: Why this here and not as `whatsNewModel`? I'm OK with this, it's just style inconsistency :)

tests/qml/main.cpp
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "helper.h"
+#include <QtQuickTest>
+#include "../../src/settingsholder.h"
+
+SettingsHolder settingsHolder;
```
bakulf, line 8: Can this be moved into `TestHelper` as a member var?

tests/qml/tst_mainWindowGleanEvents.qml
```diff
@@ -0,0 +1,64 @@
...
+
+            return {
+                resolved: promiseResolved,
+                data: promiseData
+            }
+        }
+
+        function test_onRecordGleanEventRecordsAppropriateSample() {
+            TestHelper.triggerRecordGleanEvent("authenticationAborted")
+            const { data: authenticationAbortedData } = awaitOn(Sample.authenticationAborted.testGetValue());
```
brizental, line 55: The `test*` APIs can only be called when Glean is in "test mode". Glean is put in test mode after you call the `Glean.testResetGlean` API. It is important that this API is called before the app code calls `Glean.initialize`, this makes for a not so nice code but I have a bug to try to make it nicer. This is how I do that on the QML sample tests https://github.com/mozilla/glean.js/blob/main/samples/qt/src/Tests/tst_maintests.qml#L76-L85 :)

src/ui/main/qmldir
```diff
@@ -0,0 +1,2 @@
+module main
+VPN 0.1 VPN.qml
```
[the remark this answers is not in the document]
birdsarah: This file doesn't exist any more, only in an earlier commit, I'm not sure why it's showing up for you recently.

tests/qml/helper.cpp
birdsarah: ok. 

should make a linter for this?. 

i'll clean up everywhere.

tests/qml/helper.cpp
birdsarah: Otherwise I couldn't access `main.qml` which I'm testing in glean tests. Took me ages to figure this out. But I may have been wrong, so will double check.

tests/qml/helper.cpp
birdsarah: thank you :D

tests/qml/helper.cpp
birdsarah: along the way I got confused about how to create things in order only after initialization steps had been done. I now know and can make this consistent.

tests/qml/helper.h
birdsarah: I totally agree about the inconsistency. Will fix.

tests/qml/main.cpp
birdsarah: NO! Everything breaks. 

It *may* be that now I've finally got everything it works but I don't think so.

tests/qml/helper.cpp
birdsarah: Discussed with baku; we're sticking with magic statics. 

We don't need to use the macro to run the tests, which means we don't need a public constructor, so everything will work!

tests/qml/tst_mainWindowGleanEvents.qml
birdsarah: Thank you for taking the time to explain this!

tests/qml/helper.cpp
birdsarah: I don't think so because the MozillaVPN::instance() is the mocmozillavpn.cpp implementation, no?

If you know of a way of me being able to use the real mozillavpn instance that would be awesome.

tests/qml/helper.cpp
birdsarah: If I don't have this line, then I get the following failure 

```
QWARN  : qml_tests::UnknownTestFunc() file:///home/sbird/Dev/mozilla/mozilla-vpn-client/tests/qml/tst_mainWindow.qml:7:1: module "ui" is not installed 
     import ui 0.1 
     ^
```

tests/qml/main.cpp
birdsarah: it works now :D

tests/qml/helper.cpp
```diff
@@ -0,0 +1,111 @@
...
+void TestHelper::triggerAboutToQuit() const {
+  emit MozillaVPN::instance()->aboutToQuit();
+}
+
+// The next two functions duplicate mozillavpn.h logic, but I'm not
+// sure how else to generate an "expected" value for tst_mainWindowGlean.qml
+// as it needs to be dynamic based on the platform people are running tests on.
+// At the least this should fail give someone pause for thought if they change
+// the mozillavpn.h logic.
+QString TestHelper::osVersion() const { return QSysInfo::productVersion(); }
```
birdsarah, line 54: @bakulf github might swallow our conversation because i fixed a different problem and I'd like to understand.

You wrote:
> Can you use MozillaVPN as a singleton and write this as: `QString TestHelper::osVersion() const { return MozillaVPN::instance()->osVersion(); }` ?

I don't think so because the MozillaVPN::instance() is the mocmozillavpn.cpp implementation, no?

If you know of a way of me being able to use the real mozillavpn instance that would be awesome.

tests/qml/tst_mainWindowGleanMocks.qml
```diff
@@ -0,0 +1,128 @@
...
+            TestHelper.triggerInitializeGlean()
+            compare(spyUploadEnabledInitialize, false)
+
+            VPNSettings.gleanEnabled = true
+            TestHelper.triggerInitializeGlean()
+            compare(spyUploadEnabledInitialize, true)
+        }
+
+        /*
+         * TODO - The following are qml tests that I don't know how to write
```
birdsarah, line 98: I have added an issue for this TODO https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2296 so it doesn't get lost, and I've added the glean tests to the integration test cases and have a meeting with QA to review the mechanism they need to follow.

tests/qml/helper.cpp
birdsarah: it's because the `qmldir` that defines the "ui" module is at the top of the `qrc:` resource tree i think

birdsarah:
@bakulf this is ready for review. thanks for all the feedback so far. i'm stoked with how it came out.

tests/qml/helper.cpp
bakulf: you don't need to use the "real" mozillavpn instance. There is only 1 instance, which is yours in this QML app.
But if you see, your mocmozillavpn.cpp implements the class definition from 'mozillavpn.h', and the `osVersion` implementation comes from this header file.

If you are unable to replace this with:
`QString TestHelper::osVersion() const { return MozillaVPN::instance()->osVersion(); }`

we have a bug somewhere else.

bakulf (approved):

bakulf (approved):

PR #2255, opened 2021-11-11
No unknown params for adjust SDK

bakulf:


src/adjust/adjustfiltering.cpp
```diff
@@ -146,7 +160,6 @@ QUrlQuery AdjustFiltering::filterParameters(QUrlQuery& parameters,
       continue;
     }
 
-    newParameters.addQueryItem(parameter.first, parameter.second);
```
mbirghan, line 149: Do we just want to remove them completely or replace them with dummy values?

mbirghan (changes requested):
Just one question otherwise awesome!

src/adjust/adjustfiltering.cpp
bakulf: we do not have dummy values... Better to remove them than sending invalid values.

mbirghan: Ok sounds good!

mbirghan (approved):

PR #2262, opened 2021-11-11
Pass android IAP errors to glean

strseb:
Changes in this pr: 
-Make generate_glean.py also emit all glean-metric keys into a kotlin enum. 
```
enum class GleanEvent(val key: String) {
	 androidHelloWorld("androidHelloWorld"),
	 iapNoSkuDetails("iapNoSkuDetails"),
	 iapLaunchBillingflowFailed("iapLaunchBillingflowFailed"),
````
- Add a JNI api to pass glean events to the qt-app. 
- Try to catch all errors and have a glean event for it. 

Todo: 
- [x] ktlint
- [x] Get the confirmation from @birdsarah that we have every error captured we need 
- [x] Request a data review 

Open Data Review Request: https://bugzilla.mozilla.org/show_bug.cgi?id=1746457

### CODE REVIEW

File: src/platforms/android/androidutils.cpp [hunk 1]
```diff
@@ -218,4 +218,17 @@ QByteArray AndroidUtils::DeviceId() {
...
+JNIEXPORT void JNICALL Java_org_mozilla_firefox_vpn_qt_VPNUtils_sendGleanEvent(
+    JNIEnv* env, jobject VPNUtils, jstring event) {
+  Q_UNUSED(VPNUtils);
+  const char* buffer = env->GetStringUTFChars(event, nullptr);
+  if (!buffer) {
+    return;
+  }
+  QString eventString(buffer);
+  logger.info() << "Glean Event via JNI:" << eventString;
+  emit MozillaVPN::instance()->recordGleanEvent(eventString);
```
birdsarah, line 232: As you see here we are actaully recording the glean event. Sending a ping happens later and seperately. I think it would be better if all the android code also called it `recordGleanEvent` so that we don't get confused about the two concepts.

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -306,10 +317,12 @@ class InAppPurchase private constructor(ctx: Context) :
     fun processPurchases(purchases: MutableList<Purchase>?) {
         if (purchases == null) {
             Log.d(TAG, "onPurchasesUpdated: null purchase list")
+            VPNUtils.sendGleanEvent(GleanEvent.iapPurchaseUpdateFailed)
             return
         }
         for (purchase in purchases) {
             if (purchase.purchaseState != Purchase.PurchaseState.PURCHASED) {
+                VPNUtils.sendGleanEvent(GleanEvent.iapUnexpectedNonPurchased)
```
birdsarah, line 325: Can we call this `GleanEvent.iapUnexpectedPurchaseState`

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -306,10 +317,12 @@ class InAppPurchase private constructor(ctx: Context) :
     fun processPurchases(purchases: MutableList<Purchase>?) {
         if (purchases == null) {
             Log.d(TAG, "onPurchasesUpdated: null purchase list")
+            VPNUtils.sendGleanEvent(GleanEvent.iapPurchaseUpdateFailed)
```
birdsarah, line 320: I do not believe this is a failure state. I think the processPurchases call back can get fired sometimes even when there's nothing to do and it's just a "log and move on" type situation rather than an error (which is why I did a `Log.d`) but let's check the docs.

birdsarah: I'm okay with keeping it though as it is useful data as to whether this happens a lot. If we do keep it; I'd like it to be a separate event and not share a name so we can distinguish the error states.

strseb:
@birdsarah  - Awesome! thanks for the feedback. 
 - I've now prefixed all events with iapG (for google). 
 - Renamed the events that did not match the callback
 - Added an extra event for the "null" list state.

src/platforms/android/androidutils.cpp
```diff
@@ -218,4 +218,17 @@ QByteArray AndroidUtils::DeviceId() {
 void AndroidUtils::openNotificationSettings() {
   QAndroidJniObject::callStaticMethod<void>(
       "org/mozilla/firefox/vpn/qt/VPNUtils", "openNotificationSettings", "()V");
+}
+
+JNIEXPORT void JNICALL Java_org_mozilla_firefox_vpn_qt_VPNUtils_sendGleanEvent(
```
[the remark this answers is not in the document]
strseb: I just learned about this. Yes it is! The  `jniexport`  macro auto defines the function directly after our mozilla-vpn lib is loaded into the process.

strseb:
This has been data-reviwed, thus we can land whenever the code was reviewed aswell

mbirghan (approved):

PR #2268, opened 2021-11-15
Search bar placeholder and error message on language view

### LINKED ISSUE #1859: Language settings search bar should have placeholder and error message.
The language settings screen is missing a search bar placeholder that should say "*Search languages*"

The search bar should also have an error message that should say "*No results. Try a different search*" to match the search bar in the server list.

Here's what the language screen currently looks like: 

!Group 2.png|thumbnail!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1034)

flozia:
The search bar in the language settings did not have a placeholder and error message. Error messages are now part of the search bar component and a placeholder has been added.

Resolves #1859


lesleyjanenorton (approved):
Looks good to me! Added @flodolo as a reviewer since we're adding a string.

translations/strings.yaml
```diff
@@ -76,6 +76,9 @@ serversView:
   searchPlaceholder: Search countries
   searchNoResultsLabel: No results. Try a different search
 
+languageView:
+  searchPlaceholder: Search languages
```
flodolo, line 80: I know it might seem obvious, but English is confusing. Could we add a comment that this is for a field searching **through** languages?

flozia: Sure — makes sense. Added a comment to the string.

flodolo (approved):

PR #2269, opened 2021-11-15
Add PPA automation for tagged releases

oskirby:
This should also trigger the PPA automation workflow when pushing a release tag to github, and will ensure that the build-revision mangling is removed when doing so.

Some food for though: Would it be worthwhile uploading the signed source bundle (or any other build artifact) as a release asset on github?

### DIFF
--- a/.github/workflows/ppa-automation.yaml
+++ b/.github/workflows/ppa-automation.yaml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - 'releases/**'
+    tags:
+      - 'v[0-9]+.[0-9a-z]+.[0-9a-z]+'
   schedule:
     - cron: '0 5 * * *'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ppa-automation.yaml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - 'releases/**'
+    tags:
+      - 'v[0-9]+.[0-9a-z]+.[0-9a-z]+'
```
birdsarah, line 7: why do we need the letters in the `[]` blocks?

.github/workflows/ppa-automation.yaml
```diff
@@ -26,6 +28,9 @@ jobs:
           if [[ "$GITREF" =~ ^refs/heads/releases/([0-9][^/]*) ]]; then
             BUILDREV="~rc$(git rev-list --count --first-parent origin/main..HEAD)"
             VERSION=$(echo $GITREF | cut -d'/' -f4)
+          elif [[ "$GITREF" =~ ^refs/tags/v([0-9a-z.]+) ]]; then
```
birdsarah, line 31: should this match the tag regex?

birdsarah (approved):
some questions, but i don't see problems with merging this. it just means that stuff gets uploaded to owen's ppa more often (if i'm understanding correctly)

.github/workflows/ppa-automation.yaml
oskirby: I didn't want to exclude the possibility of release numbering weirdness, like `2.6.1a` or something, but we've never actually tagged such a release.

.github/workflows/ppa-automation.yaml
oskirby: I suppose I was just being lazy, since we should only get into this path when the on-push hook is run anyways, and we're using the regex here to pick the version number out of the tag.

oskirby:
This should also allow us to generate the source bundle for our official Linux releases in CI instead of requiring manual steps.

bakulf (approved):

PR #2271, opened 2021-11-16
Update `VPNSettings.startAtBoot` setting strings

lesleyjanenorton:
New strings in response to feedback that the current strings for the `VPNSettings.startAtBoot` setting are confusing. 

Screen Shot 2021-11-15 at 9 17 44 PM

[Figma spec](https://www.figma.com/file/iBkvpAPTSMQHeRxIrCynQi/Settings-Redesign?node-id=99%3A1749)


flodolo:
I'm still unclear on this function, and the title doesn't really clear it up.

1) If checked, will the app start at boot AND connect?
2) Are there separate settings to a) start the app at boot, and b) connect when the app starts?

lesleyjanenorton:
cc @chrisrooneys

chrisrooneys:
@flodolo 

If checked, will the app start at boot AND connect?
_Yes_
Are there separate settings to a) start the app at boot, and b) connect when the app starts?
_No_

For the title, we worked through it a few times to try to keep it short and focused on the main thing (the connection). It seemed like saying "open app and connect" might be too long.

flodolo:
Thanks for the explanation. Then I would suggest to add a comment that "startup" refers to the device's startup, not the app.

lesleyjanenorton:
Closing this in favor of #2418

Issue #2183, opened 2021-11-02
Android 12 tells me that "Mozilla VPN pasted from your clipboard", for no obvious reason


dholbert:
Preconditions: Have a phone running Android 12, with Mozilla VPN installed but not currently running. (I've seen this when it was already running, too, but it's more predictable if it's not.)

STR:
1. Copy a URL to your clipboard.
2. Open the Mozilla VPN app.

ACTUAL RESULTS:
Immediately, I get a notification "Mozilla VPN pasted from your clipboard".

EXPECTED RESULTS:
No such notification. There's no reason Mozilla VPN should be inspecting my clipboard (particularly without any interaction on my part besides opening the app).

Screenshot:
![image](https://user-images.githubusercontent.com/426803/139958465-4e1a68bf-c822-45a4-a9b0-fc462510f2b0.png)

See related Fenix bug: https://github.com/mozilla-mobile/fenix/issues/22271
...though this notification is more understandable on Fenix, because over there, this is usually associated with some "fill link from clipboard" UI there, which means users can understand why they're seeing the notification.  Whereas in Mozilla VPN, there's no clear reason for the app to be reading the user's clipboard (particularly at startup before any user interaction).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1266)

dholbert:
Phone model: Pixel 4a
Mozilla VPN App version: 2.5.1

strseb:
Hey! Are you running by chance a pre-release build of android 12? I was able to experience it on my pixel 4 on preview too, but it stopped happening after the latest update. 
Also, we have no code accessing the clipboard, which just makes this extra weird 🙈

dholbert:
I'm not running a pre-release build, no. I just installed the regular Android 12 release when it became availabl.

dholbert:
I just received a ~24mb "security update" to Android 12 (so now System Update shows me as being at version `Android security update: November 5, 2021`.  I retested and confirmed that I can still reproduce.

@strseb: two important factors which impact reproducibility:
(1) Do you have a URL that you've *just* copied to your clipboard?  (needs to be recently copied, or at least new-to-the-clipboard since the last time Mozilla VPN was running)
(2) Are you starting Mozilla VPN "fresh" (if it was previously running, you need to swipe it offscreen in the app-switcher to kill it)?

[...]

PR #2280, opened 2021-11-16
Android: Disable qt's read-access to the clipboard

strseb:
We don't use the clipboard on android 10>, so no need for QT to read it, as it might creep out people :) 
closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2183

### DIFF
--- a/android/src/org/mozilla/firefox/vpn/qt/VPNActivity.java
+++ b/android/src/org/mozilla/firefox/vpn/qt/VPNActivity.java
@@ -4,8 +4,11 @@
 
 package org.mozilla.firefox.vpn.qt;
 
+import android.os.Build;
 import android.os.Bundle;
 import android.view.KeyEvent;
+
+import org.qtproject.qt5.android.bindings.QtActivity;
 
 public class VPNActivity extends org.qtproject.qt5.android.bindings.QtActivity {
   @Override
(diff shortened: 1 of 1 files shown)


mbirghan (approved):
Great!

android/src/org/mozilla/firefox/vpn/qt/VPNActivity.java
```diff
@@ -39,6 +42,19 @@ public void onBackPressed() {
     }
   }
 
+  @Override
+  public Object getSystemService (String name){
+
+    if(Build.VERSION.SDK_INT >= 29 && name.equals("clipboard")){
+      // QT will always attempt to read the clipboard if content is there.
+      // since we have no use of the clipboard in android 10+
+      // we _can_  return null
```
bakulf, line 51: This seems a hack. Can you tell me more about why this is needed?

strseb: I mean ... it is! 
During app startup, but before the mvpn_app startup qt will query the clipboard service, and if that is present it will register a  [ClipBoard ChangedListener](https://developer.android.com/reference/android/content/ClipboardManager.OnPrimaryClipChangedListener) - that will always read the clipboard and the new contents to qt. 
This all happens in the android-qt glue code, so it's unrelated whether we use it on our qt-side or not. 

This is problematic, as android 12 will now notify users on clipboard reads, so every time the app goes foreground with clipboard content present, the warning will show. I think forking/re-writing the glue code is a bit out of scope since qt6 is coming.  So I figured we might just remove the access of the service, which means we would skip that whole initialization flow.
https://code.qt.io/cgit/qt/qtbase.git/tree/src/android/jar/src/org/qtproject/qt5/android/QtNative.java?h=5.9#n594

bakulf (approved):

PR #2281, opened 2021-11-16
Status indicator states for unstable and lost connection

flozia:
Adds additional states to the status indicator for when the VPN is on but the connection is unstable or lost.

**Design**
[Figma spec](https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=1%3A1631)

**Screenshots**
| VPN on & unstable | VPN on & no signal |
|---|---|
|unstableOn|noSignalOn|

### DIFF
--- a/src/ui/components/VPNMainImage.qml
+++ b/src/ui/components/VPNMainImage.qml
@@ -27,13 +27,14 @@ Rectangle {
             }
             PropertyChanges {
                 target: insetCircle
-                color: Color.green
+                color: Color.success.default
             }
             PropertyChanges {
                 target: insetIcon
                 source: "qrc:/ui/resources/shield-on.svg"
                 opacity: 1
             }
+
         },
         State {
             name: VPNController.StateConfirming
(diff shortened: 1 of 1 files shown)


src/ui/components/VPNMainImage.qml
```diff
@@ -257,7 +314,6 @@ Rectangle {
         y: 5
         antialiasing: true
         smooth: true
-        color: showVPNOnIcon ? "#3FE1B0" : "#FF4F5E"
```
lesleyjanenorton, line 260: I think removing this line means we'll need to set the value here for each state. This is what I'm seeing locally when starting up the app.
Screen Shot 2021-11-16 at 2 03 53 PM

flozia: Your are totally right — I missed setting the value in the default state.

lesleyjanenorton (approved):
Nice!

Issue #1585, opened 2021-08-19
Multi-Hop: Delay hop startup based on peer handshake completion


data-sync-user:
At present, we attempt to start a multihop connection asynchronously by starting all of the peers at once and then hope that retries in the handshake will magically sort everything out for us, but unfortunately this can sometimes result in a connection failure when the first hop has a large latency (it seems to get unstable at 175ms or more).

To resolve the problem, we really need to defer the start of the second hop until the handshake for the first hop has completed, but we lack a good API to communicate this status from the daemon to the client.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-879)

Issue #1527, opened 2021-08-10
Support excluding addresses in the daemon/controller API


oskirby:
When we start the VPN, there are a collection of IP Addresses that we wish to remain accessible without using the VPN (Wireguard server, Captive portal check, and IP address check). This is accomplished by computing the set of routing prefixes that cover everything on the internet except for these excluded addresses. Unfortunately this can rapidly become an overwhelming list of prefixes.

Some platforms could do this in the daemon much more efficiently. For example:

* Linux could make use of the routing policy table to route these addresses via the main table instead of the VPN table.
* Windows/MacOS could monitor the routing table and dynamically create host routes for the excluded addresses which track the default route.

As an added benefit, anything that monitors the routing table could also set the MTU appropriately instead of assuming an ethernet-like MTU of 1500 bytes, which may lead to degraded network performance if the MTU happens to be smaller.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-855)

PR #2291, opened 2021-11-17
Delay multihop peer startup based on handshake completion

oskirby:
This attempts to make the activation and server switch of a multihop connection a bit more reliable by waiting for each peer to come up before starting the next one. This also enables us to get feedback on the individual servers being used and makes the procedure a bit more reliable.

In the process of testing this, it was discovered that we have a race condition when switching entry servers. This occurs because the routes to the entry server are set up by carefully excluding them from the allowed IPs of the exit server, but during a server switch we find ourselves in the situation where we are switching the entry server while still using the old exit server's allowed IPs. As a resolution to this problem, we introduce route monitoring to the desktop platforms, and use it to bind the entry server's IP address to the best non-VPN interface for reaching the internet.

This route monitoring is extremely close to what was envisioned for #1527 but I would like to address that issue in a follow up PR that also cleans up the hacks used to detect entry vs. exit servers in the daemon.

Closes: #1585

### DIFF
--- a/src/platforms/linux/daemon/wireguardutilslinux.cpp
+++ b/src/platforms/linux/daemon/wireguardutilslinux.cpp
@@ -34,6 +34,8 @@ extern "C" {
 #endif
 // End import wireguard
 
+constexpr uint16_t WG_KEEPALIVE_PERIOD = 60;
+
 /* Packets sent outside the VPN need to be marked for the routing policy
  * to direct them appropriately. The value of the mark and the table ID
  * aren't important, so long as they are unique.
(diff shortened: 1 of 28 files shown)


src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -562,6 +582,55 @@ bool WireguardUtilsLinux::rtmSendRule(int action, int flags, int addrfamily) {
   return true;
 }
 
+bool WireguardUtilsLinux::rtmSendExclude(int action, int flags,
+                                         const QHostAddress& address) {
+  constexpr size_t fib_max_size =
+      sizeof(struct fib_rule_hdr) + RTA_SPACE(sizeof(struct in6_addr));
+
+  char buf[NLMSG_SPACE(fib_max_size)];
```
mkestler-mozilla, line 590: If you use an initializer, you can avoid the memsets later.  ie char buf[...]={0};

mkestler-mozilla (approved):
I'm ok if you don't fix my comments.  They are just suggestions, not functional issues.

bakulf (approved):

PR #2298, opened 2021-11-18
Nit fix followup to #2281

lesleyjanenorton:
Followup to #2281 - this tweaks a few transition animations so that icon and background color appear to change simultaneously instead of in rapid succession.


flozia (approved):
Thank you for adding the transition! 🙂

PR #2299, opened 2021-11-18
An helper script + MacOS CI to compile VPN with Qt6

bakulf:


oskirby (approved):
I think this looks good, but if anything I would say that we should just bite the bullet and switch to Qt6 exclusively at some point (not necessarily now, but soon). Because the MacOS build is statically linked there is no need to retain support for Qt5

bakulf:
Right! I agree. But we are not ready yet. The build crashes on shutdown, images are missing. It's good to have CI because we can play with packages. This is just a first step.

PR #2307, opened 2021-11-18
Use qt-ios repo from mozilla-mobile

bakulf:


mbirghan (changes requested):
Since I just changed the names of the files we will also need to change x* to gt5*

mbirghan (approved):
Great!
After this lands we can cherry pick it to a releases branch.

PR #2311, opened 2021-11-18
Vpn 890 last cleanup of errant connection messages

mkestler-mozilla:


bakulf (approved):
Marcus and I have discussed this PR. This is a temporary solution before a refactoring of the Controller and the network manager.

PR #2343, opened 2021-11-22
Modernize Singletons

mkestler-mozilla:
I've converted most of our singletons to use magic statics, fixed ctors and prevent pointer deletion.  There are still some that can't currently be fixed due to the fact that they are subclassed (bad practice).  We can address those in the future.


src/apppermission.cpp
```diff
@@ -48,18 +45,11 @@ AppPermission::AppPermission(QObject* parent) : QAbstractListModel(parent) {
...
-}
+AppPermission::~AppPermission() { MVPN_COUNT_DTOR(AppPermission); }
 
 AppPermission* AppPermission::instance() {
-  if (s_instance == nullptr) {
-    new AppPermission(qApp);
-  }
-  Q_ASSERT(s_instance);
-  return s_instance;
+  static auto instance = new AppPermission();
```
bakulf, line 51: I like it a lot! But how to avoid MVPN_COUNT_DTOR() to do the right math to detect leaks?
Without the parent, the singleton will not be destroyed when we run the check, am I right?

mkestler-mozilla: I'd suggest we remove the checks for singletons then.  The magic static is guaranteed to be destroyed on static cleanup after main has completed.  A leak really isn't possible.

src/singleton.h
```diff
@@ -0,0 +1,27 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef SINGLETON_H
+#define SINGLETON_H
+
+#include <Qobject>
+
+template <typename T>
+class Singleton {
```
bakulf, line 11: do you use it?

mkestler-mozilla: No.  I need to pull it.  QObject does not behave well with templates.

src/qmlengineholder.h
```diff
@@ -30,10 +27,13 @@ class QmlEngineHolder final : public NetworkManager {
...
   void hideWindow();
 
+  void operator delete(void*){};
+
  protected:
   void clearCacheInternal() override;
 
  private:
+  QmlEngineHolder() = default;
   QQmlApplicationEngine m_engine;
```
bakulf, line 37: this object will never be deleted.

bakulf (commented):
This is definitely a big cleanup but there is something that I would like to discuss:
- this PR changes how objects are deleted. Most of the singletons are not deleted and because of this, there is not a proper teardown of QT objects (QSettings, QmlEngine, ...).
- In the current impl, we are sure that nothing lives more than a `qApp`. In your PR, this looks OK too because the app is mainly single-thread, but there is nothing to prevent it. 

I wonder if we can combine this with the QT-parenting solution. Is there a reason why this PR doesn't follow this approach? I tried to search for QT+singleton and there are almost 0 "official" results, just some forums.

mkestler-mozilla:
> This is definitely a big cleanup but there is something that I would like to discuss:
> 
>     * this PR changes how objects are deleted. Most of the singletons are not deleted and because of this, there is not a proper teardown of QT objects (QSettings, QmlEngine, ...).
> 
>     * In the current impl, we are sure that nothing lives more than a `qApp`. In your PR, this looks OK too because the app is mainly single-thread, but there is nothing to prevent it.
> 
> 
> I wonder if we can combine this with the QT-parenting solution. Is there a reason why this PR doesn't follow this approach? I tried to search for QT+singleton and there are almost 0 "official" results, just some forums.

The singletons are destroyed in the static denit phase after main exits.  That should not involve the delete operator.
The only way to protect from someone doing a delete Singleton::instance(); call is to remove the operator.

That said, I prefer the straight up Meyer pattern that returns a reference to the instance.  This removes the need to remove the delete operator.  I used the pointer approach because the code base likes to pass around raw pointers.

As for mixing this in to the QObject cleanup hierarchy, that is a non-starter.  You can't allow something to control the lifecycle of the singleton other than the singleton itself.  It fundamentally breaks the pattern and would force us to allow some holes in protecting the instance.

Another thing I'd like to point out is that we have way too many singletons.  That causes a lot of problems in the tests, and will bring all sorts of heartburn if we go multithreaded.  In the past I have abandoned singletons all together in favor of a ServiceLocator pattern.  In that, there is one singleton that manages a set of "service" objects that can be located by name.  You can set up parent child relationships in them to control startup and shutdown order.  You can give the services a nice clean startup() stop() semantic to activate and deactivate.  It also allows you to swap the service based on platform easily, which is what we try to do with some of the bad singletons we have that extend a base singleton (see NetworkManager and is subclasses).  They illustrate another fundamental problem with singletons.

src/authenticationinapp/authenticationinapp.cpp
```diff
@@ -16,28 +16,12 @@ constexpr int PASSWORD_MIN_LENGTH = 8;
...
-AuthenticationInApp* AuthenticationInApp::instance() {
-  if (!s_instance) {
-    new AuthenticationInApp(qApp);
-  }
-  Q_ASSERT(s_instance);
-  return s_instance;
-}
-
-AuthenticationInApp::AuthenticationInApp(QObject* parent) : QObject(parent) {
-  MVPN_COUNT_CTOR(AuthenticationInApp);
```
bakulf, line 32: you can remove 'leakdetector.h' everywhere you remove the MVPN_COUNT_CTOR/DTOR macros.

src/localizer.h
```diff
@@ -15,6 +15,8 @@ class Localizer final : public QAbstractListModel {
   Q_OBJECT
   Q_DISABLE_COPY_MOVE(Localizer)
 
+  void operator delete(void*){};
```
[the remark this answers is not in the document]
bakulf: You can remove this, right?

tests/qml/Makefile.Debug
```diff
@@ -0,0 +1,6355 @@
+#############################################################################
```
bakulf, line 1: This file is generated. It should not be included here.

tests/qml/Makefile.Release
```diff
@@ -0,0 +1,6352 @@
+#############################################################################
```
bakulf, line 1: This file too. Remove it.

bakulf (changes requested):
Nice! 3 main comments:
1. Can you please remove the Makefile.* files?
2. Do we still need to "disable" the delete?
3. We should remove the 'leakdetector.h' when the *_CTOR macro is removed.

src/loghandler.cpp
```diff
@@ -50,55 +51,47 @@ LogLevel qtTypeToLogLevel(QtMsgType type) {
 
 // static
 LogHandler* LogHandler::instance() {
-  MutexLocker lock(&s_mutex);
-  return maybeCreate(lock);
+  LogLevel minLogLevel = Debug;  // TODO: in prod, we should log >= warning
```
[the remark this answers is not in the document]
mkestler-mozilla: Moved it to the constructor.

src/platforms/windows/daemon/windowsfirewall.h
```diff
@@ -32,8 +32,8 @@ class WindowsFirewall final : public QObject {
   bool disableKillSwitch();
 
  private:
-  WindowsFirewall(QObject* parent);
-  HANDLE m_sessionHandle;
+  WindowsFirewall();
+  HANDLE m_sessionHandle = INVALID_HANDLE_VALUE;
```
bakulf, line 36: nice!

bakulf (approved):

PR #2345, opened 2021-11-23
try macos build

mbirghan:


bakulf (approved):

bakulf:
This cannot be an alternative to MacOS. The real alternative is task-cluster.

PR #2348, opened 2021-11-23
Move address exclusion support into the daemon

oskirby:
This cleans up some of the lingering hacks in the multihop peer startup pull request (#2291), while at the same time allowing the daemon to manage the list of excluded addresses, which it may be better able to do using platform-specific techniques (eg: routing policy rules or by listening for routing table updates).

Platform implementation status:
- [x] Windows address exclusion support
- [x] MacOS address exclusion support
- [X] Linux address exclusion support
- [x] Verify that iOS is not broken
- [x] Verify that Android is not broken
- [x] Repair unit tests

Closes: #1527

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -765,18 +765,6 @@ QStringList Controller::getExcludedAddresses(const QList<Server>& serverList) {
     list.append(dns);
   }
 
-  // It's not clear if this is really needed for the Android platform, but
-  // the Controller on the desktop platforms sorts this out using the server
-  // hop list, and iOS doesn't need to explicitly exclude the entry servers.
-#ifdef MVPN_ANDROID
-  const Server& entryServer = serverList.last();
-  // filtering out the ingress server's public IPv4 and IPv6 addresses.
-  logger.debug() << "Exclude the entry server:" << entryServer.ipv4AddrIn();
-  list.append(entryServer.ipv4AddrIn());
-  logger.debug() << "Exclude the entry server:" << entryServer.ipv6AddrIn();
-  list.append(entryServer.ipv6AddrIn());
-#endif
-
   return list;
 }
 
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -231,7 +231,8 @@ void Controller::activateInternal() {
 
   Q_ASSERT(m_impl);
   m_impl->activate(serverList, device, vpn->keys(),
-                   getAllowedIPAddressRanges(serverList), vpnDisabledApps, dns,
+                   getAllowedIPAddressRanges(serverList),
+                   getExcludedAddresses(serverList), vpnDisabledApps, dns,
                    stateToReason(m_state));
 }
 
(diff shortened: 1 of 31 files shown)


src/controller.cpp
```diff
@@ -752,6 +729,48 @@ QList<IPAddressRange> Controller::getAllowedIPAddressRanges(
...
+  }
+
+  // Filter out the Custom DNS Server, if the user has set one.
+  if (DNSHelper::shouldExcludeDNS()) {
+    auto dns = DNSHelper::getDNS(serverList.last().ipv4Gateway());
+    logger.debug() << "Filtering out the DNS address:" << dns;
+    list.append(dns);
+  }
+
+  // It's not clear if this is really needed for the Android platform, but
```
strseb, line 759: Should also not be required for android, as far as i remember

oskirby: Awesome thanks!

src/adjust/adjustfiltering.cpp
```diff
@@ -25,9 +25,9 @@ const QVector<QString> allowList{
     "os_name",
     "os_version",
     "package_name",
+    "reference_tag",
```
bakulf, line 28: already fixed.

src/platforms/macos/daemon/macosroutemonitor.cpp
```diff
@@ -105,7 +105,7 @@ void MacosRouteMonitor::handleRtmUpdate(const struct rt_msghdr* rtm,
 
   // We expect all useful routes to contain a destination, netmask and gateway.
   if (!(rtm->rtm_addrs & RTA_DST) || !(rtm->rtm_addrs & RTA_GATEWAY) ||
-      !(rtm->rtm_addrs & RTA_NETMASK)) {
+      !(rtm->rtm_addrs & RTA_NETMASK) || (addrlist.count() < 3)) {
```
bakulf, line 108: why?

bakulf (approved):

src/platforms/macos/daemon/macosroutemonitor.cpp
oskirby: For reasons that I don't quite understand, I have seen a few occurrences of routing messages from the kernel which set more flags in `rtm_addrs` than were included as attributes. This would seem to be a violation of the routing socket protocol and it leads to an assert later on when indexing into the `addrlist`. The extra check here ensures that we simply ignore those messages.

src/platforms/macos/daemon/macosroutemonitor.cpp
```diff
@@ -268,14 +259,13 @@ void MacosRouteMonitor::rtmAppendAddr(struct rt_msghdr* rtm, size_t maxlen,
...
 bool MacosRouteMonitor::rtmSendRoute(int action, const QHostAddress& prefix,
                                      unsigned int plen, unsigned int ifindex,
-                                     const struct sockaddr* gateway) {
+                                     const void* gateway) {
   constexpr size_t rtm_max_size = sizeof(struct rt_msghdr) +
                                   sizeof(struct sockaddr_in6) * 2 +
                                   sizeof(struct sockaddr_storage);
-  char buf[rtm_max_size];
-  struct rt_msghdr* rtm = (struct rt_msghdr*)buf;
+  char buf[rtm_max_size] = {0};
```
bakulf, line 266: Nice

bakulf (approved):

PR #2356, opened 2021-11-24
Let's pass 'Task' as parent object to network requests

bakulf:


src/platforms/android/androidauthenticationlistener.cpp
```diff
@@ -27,9 +27,12 @@ AndroidAuthenticationListener::~AndroidAuthenticationListener() {
   MVPN_COUNT_DTOR(AndroidAuthenticationListener);
 }
 
-void AndroidAuthenticationListener::start(const QString& codeChallenge,
+void AndroidAuthenticationListener::start(Task* task,
+                                          const QString& codeChallenge,
                                           const QString& codeChallengeMethod,
                                           const QString& emailAddress) {
+  Q_UNUSED(task);
```
oskirby, line 34: It looks like we are using `task` down on line 52.

src/update/balrog.cpp
```diff
@@ -88,12 +88,14 @@ QString Balrog::userAgent() {
 #endif
 }
 
-void Balrog::start() {
+void Balrog::start(Task* task) {
+  Q_UNUSED(task);
```
oskirby, line 92: Another unnecessary `Q_UNUSED()`

oskirby (approved):

PR #2360, opened 2021-11-25
Fix sorting of adjust allowList leading to debug assert.

oskirby:
The `allowList` in the Adjust filtering module wasn't correctly sorted, which was triggering an assert when compiled with `MVPN_DEBUG`, and possibly means that some datum weren't being filtered correctly due to the binary search algorithm expecting a sorted list.


mbirghan:
Wow that is not good. 
@bakulf Would it make sense to exchange QVector with QMap or something similar to prevent those issues in the future?

mbirghan (approved):

bakulf (approved):

PR #2372, opened 2021-11-30
Combine IPAddressRange and IPAddress

oskirby:
It has always seemed a bit odd to me that we have both the `IPAddress` and `IPAddressRange` classes, which seem to do exactly the same thing: wrap a `QHostAddress` together with a netmask/prefix length. This PR attempts to eliminate the duplication by combining the classes together.

To keep the memory usage down, we also refactor the `IPAddress` class to generate the various masks and broadcast addresses dynamically from the prefix length, which also simplifies the excluded address generation.


bakulf (approved):
Nice cleanup

PR #2373, opened 2021-12-01
Make adjust fields changable through the feature list

mbirghan:


bakulf:
I like it but they are not persistent right? Are we sure that the feature list are fetched before activating adjust? Can we enforce this?

mbirghan:
No they are not persisitent that is a good point. 
I will create a Task to initialize Adjust which we can call after `TaskGetFeatureList`. That will force the update before we send any adjust packages.

src/adjust/adjustfiltering.cpp
```diff
@@ -3,158 +3,91 @@
...
 
-    const MirrorParam* mirrorParam = std::lower_bound(
-        mirrorList.begin(), mirrorList.end(), parameter.first,
-        [](const MirrorParam& data, const QString& value) -> bool {
-          return data.m_paramName < value;
-        });
-    if (mirrorParam && mirrorParam != mirrorList.end() &&
-        mirrorParam->m_paramName == parameter.first) {
+    if (mirrorList.contains(parameter.first)) {
+      const MirrorParam mirrorParam = mirrorList.value(parameter.first);
```
bakulf, line 75: I guess this can be: `const MirrorParam& mirrorParam = ...`

src/adjust/adjustfiltering.cpp
```diff
@@ -3,158 +3,91 @@
...
       bool found = false;
       for (const QPair<QString, QString>& otherParameter :
            parameters.queryItems()) {
-        if (mirrorParam->m_mirrorParamName == otherParameter.first) {
+        if (mirrorParam.m_mirrorParamName == otherParameter.first) {
```
bakulf, line 79: Move the `if` outside the `for`. In this way, you run the `for`-loop only if the `mirrorParam.m_mirrorParamName` is allowed.

src/tasks/initializeadjust/taskinitializeadjust.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+TaskInitializeAdjust::TaskInitializeAdjust() : Task("TaskInitializeAdjust") {
+  MVPN_COUNT_CTOR(TaskInitializeAdjust);
+}
+
+TaskInitializeAdjust::~TaskInitializeAdjust() {
+  MVPN_COUNT_DTOR(TaskInitializeAdjust);
+}
+
+void TaskInitializeAdjust::run() { AdjustHandler::initialize(); }
```
bakulf, line 23: You don't need a task to run 1 single function. Instead run this:

`TaskScheduler::scheduleTask(new TaskFunction([]{ AdjustHandler::initialize(); }));`

bakulf (approved):

PR #2374, opened 2021-12-01
TaskGroup

bakulf:
The application is definitely faster with this PR.


oskirby (approved):
This was a surprisingly small patch for such a large performance improvement.

PR #1973, opened 2021-10-06
Captive Portal Additional Messaging

strseb:
This adds the new Captive Portal Messaging flow- see https://www.figma.com/file/OQpyNjmowo3BIE23WGnOCM/Captive-Portal---Additional-Messaging?node-id=14%3A3347

**Changelist:**
-> Adds a 2nd CaptivePortal Monitor, that is active when the vpn is disconnected. 
-> before activation the controller checks if a portal is detected currently, if so aborts activation and pushes ViewCaptivePortalInfo onto stackview.
-> User can open the Portal from there. 
->User can also re-attempt a connection

**Todo**
- [x] Put strings into yaml
- [x] Write UI Test

Mozilla_VPN_2021-10-06_17-59-22_Trim_SparkVideo

PR #2378, opened 2021-12-01
Add Captive Portal Scenario 2

strseb:
Adds Scenario 2 of [this figma](https://www.figma.com/file/OQpyNjmowo3BIE23WGnOCM/Captive-Portal---Additional-Messaging?node-id=14%3A1460)

-> This opens the captive-portal alert on detection (when client is connected)
-> clicking the button deactivates the vpn and starts the cp-background monitor :) 

Marking as blocked as this is based on pr #1973


lesleyjanenorton (approved):
QML looks good. Is this still blocked? Added @flodolo to review the strings.

translations/strings.yaml
```diff
@@ -244,9 +244,18 @@ customDNSSettings:
 networkSettings:
   multihopTitle: Multi-hop tunnel
 
+
+captivePortalAlert:
+  title: Captive portal detected!
+  header: A captive portal is when the guest Wi-Fi you are using requires you to log in to get internet access. 
+  actionPreActivation: You can turn on your VPN once you have gained internet access by logging in to the portal. 
+  actionPostActivation: You must turn off your VPN in order to see the portal, and turn it back on once you have logged in. 
+  buttonTextPreActivation: Okay
```
flodolo, line 253: Note: I've seen complains from people about this form vs OK (at least one user on Bugzilla). Not a localization problem, just worth thinking about it ;-)

translations/strings.yaml
```diff
@@ -244,9 +244,18 @@ customDNSSettings:
...
+  buttonTextPostActivation: Turn off VPN
+  
 notifications:
   captivePortalBlockTitle: Guest Wi-Fi portal blocked
-  captivePortalBlockMessage: The guest Wi-Fi network you’re connected to requires action. Click here to turn off VPN to see the portal.
+  captivePortalBlockMessage2: The guest Wi-Fi network you’re connected to requires action. Click to turn off VPN to see the portal.
```
flodolo, line 258: I mentioned this before: is click the right word for these? This codebase support desktop AND mobile devices.

flodolo (approved):
It would be great to clarify the "click" bit before merging. The "Okay" it's just a heads-up, I don't have an opinion about it ;-)

translations/strings.yaml
lesleyjanenorton: @chrisrooneys

PR #2410, opened 2021-12-05
Removed the use of the Q_INIT_RESOURCE macro

bakulf:
And the user of CONFIG+=ordered - not needed in the `.pri` projects


mkestler-mozilla (approved):
Yes, since they are no longer static libs, these are not needed.

PR #2416, opened 2021-12-06
QML tests for VPNTextArea

flozia:
This PR adds QML test cases for the VPNTextArea component.


tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,88 @@
...
+        function test_exceedsCharLimitMultiple() {
+            resetTextArea();
+            fillTextInput();
+
+            const expected = vpnTextAreaTest.textInput.text;
+            const inputString = "abc";
+            insertMultipleChars(inputString);
+
+            const actual = vpnTextAreaTest.textInput.text;
+            verify(expected === actual, `text was ${actual} not ${expected}.`);
```
birdsarah, line 84: When I first read this, I expected it to pass, but for the wrong reason. I thought that you'd copied by reference and so actual was a reference to expected. I see that this isn't the case and the test works correctly.

BUT I still think this would be clearer as 

const expected = "a".repeat(1000)

Where you're being explicit about the expected outcome and the actual is what you read off the property.

birdsarah (changes requested):
Great job. The code is very readable. I think we can have a few improvements:
* I've made some comments inline; in particular to remove two of the tests you have
* You haven't added tests to check the cursor position is in the right place (i can delete the line `textArea.cursorPosition = prevCursorPosition;` and all the tests still pass).
* I was surprised when I ran these locally that he text that you're adding is going at the beginning of the text; it made me realize that I hadn't understood all of what your function is doing - i think a couple of tests about the role of cursorPosition might help me understand better the function by reading the tests.

tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,88 @@
...
+
+            const expected = "a";
+            insertSingleChar(expected);
+
+            const actual = vpnTextAreaTest.textInput.text;
+            verify(expected === actual, `text was ${actual} not ${expected}.`);
+        }
+
+        // Insert multiple chars within limit
+        function test_insertMultipleChars() {
```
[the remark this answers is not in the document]
flozia: Makes total sense — I updated the function names. I’ll make sure to pass the wisdom along and would like to adopt the term “being SUPER Sarah about this …” with you permission when it comes to test function names.

tests/qml/tst_VPNTextArea.qml
flozia: You are right. It’s very easy to misunderstand what is happening here at first glance.

tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,83 @@
+import QtQuick 2.3
```
bakulf, line 1: copyright/left mozilla header?

tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        function insertMultipleChars(chars) {
+            [...chars].forEach(character => {
+                insertSingleChar(character);
+            });
+        }
+
+        function fillTextInput() {
+            vpnTextAreaTest.textInput.text = "a".repeat(vpnTextAreaTest.textInput.maxCharacterCount);
```
bakulf, line 34: It would be nice to have tests to support multi-bytes unicode chars.  Interesting reading: https://hsivonen.fi/string-length/

tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        function test_insertingMultipleCharsSmallerThanLimitDoesNotCauseTruncationOfText() {
```
bakulf, line 37: wow, long name! But I'm OK with Sarah here.

bakulf (approved):

tests/qml/tst_VPNTextArea.qml
birdsarah: yes - please add the standard mozilla header - copy from any other file

```
 ```

File: tests/qml/tst_VPNTextArea.qml [hunk 6]
```diff
@@ -0,0 +1,83 @@
...
+            const text = "abc";
+            insertMultipleChars(text);
+
+            const expected = text.length;
+            const actual = vpnTextAreaTest.textInput.cursorPosition;
+
+            verify(expected === actual, `Cursor position was ${actual} not ${expected}.`);
+        }
+
+        function test_resetToInitialCursorPositionAfterCharIsRemovedFromText() {
```
birdsarah, line 70: just checking my understanding - the cursor is at position 1000 and will stay there and text will keep not being added.

birdsarah (approved):
thanks for addressing my concerns. approved but need to make baku's change - at least the license file.

tests/qml/tst_VPNTextArea.qml
flozia: Yes — that is what’s intended to happen.

bakulf (approved):

tests/qml/tst_VPNTextArea.qml
```diff
@@ -0,0 +1,98 @@
...
+
+        function test_textLengthMatchesUnicodesLenghtInJsAndQml() {
+            resetTextArea();
+
+            const expected = "🙋🏾".repeat(vpnTextAreaTest.textInput.maxCharacterCount);
+            vpnTextAreaTest.textInput.text = expected;
+
+            const actual = vpnTextAreaTest.textInput.text;
+
+            verify(expected.length === actual.length, `text was ${actual} not ${expected}.`);
```
flozia, line 94:
```suggestion
            verify(expected.length === actual.length, `text length was ${actual} not ${expected}.`);
```

lesleyjanenorton (approved):
Nice! Just one question/comment

nebula/ui/components/forms/VPNTextArea.qml
```diff
@@ -11,6 +11,7 @@ import components 0.1
 
 Item {
     property alias placeholderText: formattedPlaceholderText.text
+    property alias textInput: textArea
```
lesleyjanenorton, line 14: @flozia does this need to be `textArea.someProperty` ?

flozia: @lesleyjanenorton This makes it possible to access all the properties and methods of `textInput`. That is why we can do `textInput.text`, `textInput.maxCharacterCount` and `textInput.forceActiveFocus()` from the tests for example.

lesleyjanenorton: nice!

lesleyjanenorton (approved):

PR #2420, opened 2021-12-06
Server unavailable modal

flozia:
This PR adds the ”Server unavailable” modal and provides the functionalities to directly go to the ”Servers view” from every view within `StateMain`. Going to the ”Servers view” through clicking ”Choose a new location” and then clicking the ”menu back button” should always direct a user to `ViewMain`. Launching the modal will be implemented in a [separate issue](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-1242&search=858).

**Screenshot**
image

**Design**
[Figma spec](https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=207%3A6450)

**ToDos**
- [x] Create modal component
- [x] Handle “Choose a new location”


src/mozillavpn.cpp
```diff
@@ -1387,6 +1387,15 @@ void MozillaVPN::refreshDevices() {
...
 
+void MozillaVPN::goToServersView() {
+  logger.debug() << "Go to servers view";
+
+  if (m_state == StateMain) {
+    emit clearCurrentViewStack();
+    emit showServersView();
+  }
+}
+
```
flozia, line 1398: This feels a bit hacky, but I’m not sure what other options we have.

lesleyjanenorton: It's hacky but I think other options would require a big refactor of all of the competing StackViews, which are also hacky.

lesleyjanenorton (approved):

src/mozillavpn.cpp
bakulf: Can you do it in js?

flozia: @bakulf Pretty sure there is a way to do this using only JS, but I think that would involve passing around properties and/or callbacks in many components. The reason why I thought emitting the signals here is that we can listen to them easily since the `VPNServerUnavailablePopup` can be shown while being in any view within `StateMain`. Do you think we can leave the `signals` here (or move them in case we find a better place as part of [Add a generic error to handle when a server location is down VPN-858](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-858)) or is it worth the time to explore alternatives? My feeling is that until we get a chance to clean up the `StackViews` this might be the least invasive solution.

Issue #2422, opened 2021-12-07
No data measurement for Upload/Download speed in CI screen


data-sync-user:
**VPN version:**

* 2.7.0 (2.202112061827);

**Prerequisites:**

* Have and account with a subscription for Mozilla VPN;
* Have the VPN app installed;

**Steps to reproduce:**
1. Log in with the account from prerequisites;
2. Turn ON the VPN;
3. See Connection information screen;

**Expected result:**

* There are measurement units for both Download and Upload speeds;

**Actual result:**

* No measurement units displayed.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1432)

data-sync-user:
➤ Valentina Virlics commented:

All platforms have no units in CI screen.

data-sync-user:
➤ Luminita Buruian commented:

Verified this on Mozilla VPN, and the measurement units for both Download and Upload speeds are present in the CI screen.

  * 2.7.0 (2.202112131208) on Windows 10
  * 2.7.0~github3338 (2.7.0~github3338-focal1) on Ubuntu 20.04;
  * 2.7.0 (2.202112130413) on iOS

We cannot check this on MacOS, because of another issue (https://mozilla-hub.atlassian.net/browse/VPN-1435 ( https://mozilla-hub.atlassian.net/browse/VPN-1435|smart-link ) )

The measurement units are displayed above the title and the actual measurement is displayed under it. Also, when opening the CI screen we briefly see bit/s and then it turns into kbit/s.  Owen Kirby Are these behaviors intended? 

See below video:

!2021-12-13_15h09_51.mp4|width=412,height=600!

Thank you.

CC: Madalin Cotetiu

data-sync-user:
➤ Valentina Virlics commented:

Verified this as fixed on Mac OS Catalina 10.15 also, while using VPN 2.7.0 (2.202112170217).

The measurement units for both Download and Upload speeds are present in the CI screen.

Closing this.

PR #2439, opened 2021-12-10
Fix missing units in connection info screen

oskirby:
This was due to a missing QML import, leading to error messages in the `VPNGraphLegendMarker.qml` file as follows:
```
qrc:/nebula/components/VPNGraphLegendMarker.qml:19: ReferenceError: VPNl18n is not defined (VPNGraphLegendMarker.qml:19)
```

Closes: #2422

### DIFF
--- a/nebula/ui/components/VPNGraphLegendMarker.qml
+++ b/nebula/ui/components/VPNGraphLegendMarker.qml
@@ -4,6 +4,7 @@
 
 import QtQuick 2.5
 
+import Mozilla.VPN 1.0
 import themes 0.1
 
 Row {


flodolo:
Not sure why I'm flagged in this PR, but it doesn't look right?

oskirby:
> Not sure why I'm flagged in this PR, but it doesn't look right?
> 
> EDIT: never mind, I assume the import is enough to fix the issue. Still, I shouldn't be really flagged into code reviews, only strings ;-)

Oops! Sorry @flodolo I meant to tag @flozia in this code review but I guess I got you by mistake. That was my error.

flozia (approved):
Oh, looking into this probably should have been assigned to me. Thank you so much for fixing the issue!

PR #2446, opened 2021-12-13
Add Android/debug to tc

strseb:
As first step. Lets move the android-debug (qt5) job to taskcluster

### DIFF
--- a/scripts/android_package.sh
+++ b/scripts/android_package.sh
@@ -87,9 +87,6 @@ print Y "Checking Enviroment"
 if ! [ -d "src" ] || ! [ -d "linux" ]; then
   die "This script must be executed at the root of the repository."
 fi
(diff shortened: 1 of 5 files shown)


scripts/android_package.sh
```diff
@@ -87,9 +87,6 @@ print Y "Checking Enviroment"
 if ! [ -d "src" ] || ! [ -d "linux" ]; then
   die "This script must be executed at the root of the repository."
 fi
-if [ -z "${JAVA_HOME}" ]; then
```
bakulf, line 90: Why?

bakulf (approved):

scripts/android_package.sh
strseb: Check is not required :)

Issue #2425, opened 2021-12-07
VPN cannot be turned ON on MacOS


data-sync-user:
**VPN version:**

* 2.7.0 (2.202112061823);

**Affected Platforms**:

* MacOS; 

**Prerequisites**:

* Have a FxA account with an active subscription for Mozilla VPN;
* Have Mozilla VPN installed and open;
* Be signed in;

**Steps to reproduce**:
1. Turn on the VPN;
2. Observe the behaviour;

**Expected result**:

* The VPN is successfully turned on;

**Actual result**:

* The VPN cannot connect. The toggle switches to Disconnecting and OFF;

**Notes**:

* Reproducible on Production and Staging servers;
* Not reproducible on other platforms; 
* This is a blocker issue, because the VPN cannot be turned ON on MacOS;
* See the attached video;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1435)

data-sync-user:
➤ Luminita Buruian commented:

Owen Kirby Sure thing.

I have reproduced this on Mozilla VPN 2.7.0 (2.202112091222), Qt5 build. Please find the logs attached.

Let me know if you need anything else.

Thanks.

CC: Madalin Cotetiu Valentina Virlics 

[^mozillavpn-2021-12-9.txt]

data-sync-user:
➤ Owen Kirby commented:

I have been able to reproduce the issue by installing the MacOS client built from Github CI, but when built locally it appears to just work. The root cause of the problem here seems to be that the wireguard-go binary has not been correctly signed, and this leads to the tunnel failing to start and exiting with a crash. The smoking gun in the logs appears to be: Warning: (macos - WireguardUtilsMacos) Tunnel process encountered an error: QProcess::Crashed

Checking out the MacOS console also shows that the wireguard-go binary is failing to start due to a code-signing problem.

data-sync-user:
➤ Andrea Marchesini commented:

Owen, can you tell me more? What do you see when you run:
`codesign --verify --verbose /Applications/Mozilla\ VPN.app/Contents/Resources/utils/wireguard-go`

oskirby:
Here are my findings:
```
okirby@okirby-pllywp mozilla-vpn-client % codesign --verify --verbose /Applications/Mozilla\ VPN.app/Contents/Resources/utils/wireguard-go 
[...]

PR #2455, opened 2021-12-15
A try to fix the signature of the macos package

bakulf:


oskirby (approved):
I gave this a quick test on my Mac laptop, and it appears to have resolved connection the issue reported in #2425

PR #2486, opened 2021-12-21
Add Lottie animations playground

flozia:
A playground for our testing and our own awesome Lottie component (#2494) that @bakulf implemented and animations themselves. It lives within the `Developer Options` and implements only the functionalities we need for our current use case, but can be easily extended.

| Playground (Developer Options) | Playground (no selection) | Playground (playing) |
|---|---|---|
|lottie-developer-menu|lottie-playground-no-selection|lottie-playground-playing|


nebula/nebula.cpp
```diff
@@ -6,4 +6,7 @@
 
 constexpr auto QRC_ROOT = "qrc:///nebula/";
 
-void Nebula::Initialize(QQmlEngine* engine) { engine->addImportPath(QRC_ROOT); }
+void Nebula::Initialize(QQmlEngine* engine) {
+  qputenv("QML_XHR_ALLOW_FILE_READ", QByteArray("1"));
```
bakulf, line 10: This is horrible :/ We need to get rid of it.

nebula/ui/3rdparty/lottie/LottieAnimation.qml
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        // NOTE QML LoggingCategory {} has its internal QLoggingCategory created in
+        // componentCompleted(). There seems to be a situation where this console.log
+        // is executed before the LoggingCategory {} object above has completed.
+        //console.log(d.log, "Fetching source from", url);
+
+        var xhr = new XMLHttpRequest()
+        // FIXME allow asynchronous
+        xhr.open("GET", url, false /*synchronous*/);
```
bakulf, line 454: It's not good to load resources synchronously - this is ok because they are local resources, but in theory, they can be remote too.

nebula/ui/3rdparty/lottie/src/lottie_shim.js
```diff
@@ -0,0 +1,160 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Shims to simulate a browser environment
+var navigator;
+var window;
+var document;
```
bakulf, line 24: This implementation seems to be a big hack to make Lottie work. It seems a bit fragile, but I imagine that, if everything is correctly tested, we can accept it.

nebula/ui/3rdparty/lottie/src/lottie.min.js
```diff
@@ -0,0 +1,15 @@
+(typeof navigator !== "undefined") && (function(root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(function() {
```
bakulf: This is under MIT license. We need to add it to the LICENSE.md file.

nebula/ui/3rdparty/lottie/src/lottie_shim.js
```diff
@@ -0,0 +1,160 @@
...
+            return elements;
+        },
+        getElementsByClassName: function (className) {
+            var elements = [];
+            // find certain tags?
+            return elements;
+        }
+    };
+
+    // First check whether there is a non-minified version for debugging purposes
```
bakulf, line 138: This part can be removed. In particular, Qt.include() is deprecated and we should use modules instead.

bakulf (commented):
Let me think a bit more about this PR.

bakulf:
Can you rebase this PR ? Or recreate it, i could be faster. Thanks!

bakulf (approved):

Issue #2513, opened 2022-01-04
QT compilation fails during new MacOS development environment setup


vvalceanu:
Setting up a new MacOS development environment, as per README.md.
On master.
MacOS Big Sur 11.6.
Using qt-everywhere-src-5.15.1.tar.xz.

Running ``bash scripts/qt5_compile.sh `pwd`/qt qt``
yields an error:
`./qiosurfacegraphicsbuffer.h:54:32: error: unknown type name 'CGColorSpaceRef'; did you mean 'QColorSpace'?`

generated by not including a Mac OS header that was previously indirectly included. A bit more info [here](https://code.qt.io/cgit/qt/qtbase.git/commit/src/plugins/platforms/cocoa?id=dece6f5840463ae2ddf927d65eb1b3680e34a547).

Pull request with fix on the way.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1510)

data-sync-user:
➤ Andrea Marchesini commented:

I updated the README.md with a description about Qt6.

PR #2514, opened 2022-01-04
Patch QT header to fix inclusion error when compiling QT from scratch…

vvalceanu:
… on MacOS 11.6 and above.

See this issue for context: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2513


mbirghan (approved):
This looks great!
I would just like to have @bakulf review this too.

bakulf:
I'm going to close this PR because we are using QT6 and it seems to work fine on macOS 11.6.
We still need to update the README but it will be done next week.

PR #2516, opened 2022-01-04
Disable selection of unavailable servers

flozia:
Servers that are currently marked as “unavailable” should not be selectable and show an indicator next to them.

**Note**
~~Merging this PR has to wait until we landed PR #2537.~~ Created issue #2556 to use the actual server count after we merged previously mentioned PR.

**Screenshots**
| Before | After |
|---|---|
| image | image |

**Design**
[Figma spec](https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=207%3A6587)

**ToDo**
- [x]  Add disabled state to radio select
- [x]  Add availability indicator to server list
- [x]  Disable cities that have no available servers ([VPN-1240](https://mozilla-hub.atlassian.net/browse/VPN-1240))

**Linked issues**
Resolves [VPN-1506](https://mozilla-hub.atlassian.net/browse/VPN-1506)


nebula/ui/components/VPNServerCountry.qml
```diff
@@ -166,7 +166,10 @@ VPNClickableRow {
             delegate: VPNRadioDelegate {
                 property string _cityName: modelData[0]
                 property string _localizedCityName: modelData[1]
+                property string _activeServerCount: modelData[2]
                 property string _countryCode: code
+                property bool isAvailable: _activeServerCount > 0
```
flozia, line 171: @oskirby I just would like to check if I understand what is happening in PR https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2537 ([this part in particular](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2537/files#diff-4a9d949202ba311f8e12ee3251f43dd2cbe5e05405f18699628652fa20ece217R160-R172)) correctly: We can say an `activeServerCount` of zero means that all of the servers in a city are _timed out_ so the city is currently “unavailable”?

oskirby: That is correct. The `activeServerCount` is the number of servers that not been blocked due to a failed handshake. When the count reaches zero, there are no servers remaining that the VPN client can use for this location.

nebula/ui/components/VPNServerCountry.qml
```diff
@@ -178,24 +181,46 @@ VPNClickableRow {
...
-                    serversTabs[currentServer.whichHop] = [del._countryCode,  del._cityName, del._localizedCityName] // [countryCode, cityName, localizedCityName]
-                    multiHopStackView.pop()
+                    serversTabs[currentServer.whichHop] = [del._countryCode,  del._cityName, del._localizedCityName]; // [countryCode, cityName, localizedCityName]
+                    multiHopStackView.pop();
                 }
                 height: 54
                 checked: del._countryCode === focusScope.currentServer.countryCode &&  del._cityName === focusScope.currentServer.cityName
                 isHoverable: cityListVisible
-                enabled: cityListVisible
+                enabled: del.isAvailable
```
lesleyjanenorton, line 199: should this be cityListVisible && del.isAvailable? I vaguely remember an issue we had a long time ago where the cities could be clicked when the list was closed.

lesleyjanenorton: I think we should also restrict `isHoverable` when `!del.isAvailable` unless we are already getting that for free when the button is disabled. I think that ideally there would be no UI changes `onPressed`, `onMouseEnter`, etc when the city is unavailable/disabled.

lesleyjanenorton (approved):
This looks really good - just a few questions/comments.

nebula/ui/components/VPNServerCountry.qml
flozia: Excellent point! Thank you for catching this potential regression — I added `cityListVisible && del.isAvailable`.

Regarding the UI changes: The `VPNMouseArea` within the `VPNRadioDelegate` now gets disabled as well and should prevent any UI state updates.

PR #2520, opened 2022-01-04
Remove Obsolete Android-CI tasks

strseb:
Given that we now have both unsigned-debug builds for Qt5/ Qt6 on task cluster - I think there is no need to keep the github equivalents. 
In the same step we can then restrict the creation of signed builds to main & release* branches, which i think is neat :) 
https://treeherder.mozilla.org/jobs?repo=mozilla-vpn-client


oskirby (approved):

bakulf (approved):

strseb:
@bakulf  - Updated this pr. This now removed all android-gh ci, as it's now all in taskcluster/treeherder. 
I think i we would need you to remove the merge requirement for the now removed tasks before i can merge it :)

PR #2526, opened 2022-01-05
No 'BETA' mullvad servers in prod

bakulf:


strseb (approved):
LGTM But i think this tests needs to be addressed:
`FAIL!  : TestModels::serverCountryModelFromJson(good with one city) Compared values are not the same`

oskirby:
Wouldn't it make more sense to do this on guardian?

bakulf:
> more sense to do this on guardian?

It's more or less the same. If we do it in the guardian, we need to pass a flag to communicate when the client runs in prod or staging mode. We need to change 2 components: the client and the guardian website. This is less code.

strseb (approved):

Issue #2509, opened 2022-01-03
Re-Enable Lottie Support for Windows


strseb:
The Project currently does not build on windows, when lottie is enabled 
Its seems qmake generates an invalid vcxproj.

The CI mistakenly does not catch this and reports a success even though the logs say: 
```
Build started 1/3/2022 2:59:00 PM.
Project "D:\a\mozilla-vpn-client\mozilla-vpn-client\MozillaVPN.vcxproj" on node 1 (Clean target(s)).
D:\a\mozilla-vpn-client\mozilla-vpn-client\MozillaVPN.vcxproj(2888,5): error MSB4035: The required attribute "Include" is empty or missing from the element <None>.
Done Building Project "D:\a\mozilla-vpn-client\mozilla-vpn-client\MozillaVPN.vcxproj" (Clean target(s)) -- FAILED.

Build FAILED.

"D:\a\mozilla-vpn-client\mozilla-vpn-client\MozillaVPN.vcxproj" (Clean target) (1) ->
  D:\a\mozilla-vpn-client\mozilla-vpn-client\MozillaVPN.vcxproj(2888,5): error MSB4035: The required attribute "Include" is empty or missing from the element <None>.

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.05
Failed to clean up the project
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1502)

data-sync-user:
➤ Valentina Virlics commented:

Andrea Marchesini Brian Yee Sebastian Streich This as a blocker for QA team in regards of testing the v2.7 Mozilla VPN last build, on Windows. Would be great if this gets fixed by tomorrow.

Thank you!

data-sync-user:
➤ Valentina Virlics commented:

Thanks for the fix! Please add the fix in the main or release 2.7 branch as well.

data-sync-user:
➤ Andrea Marchesini commented:

Fix in VPN-1544

PR #2530, opened 2022-01-06
Disable Lottie Support for Windows

strseb:
Apparently the Lottie.pri is breaking windows builds and the CI did not catch it. 
Something about those 2 lines is making qmake generate a broken vcxproj but i dont know enough currently about how a qml import name would create broken vcxproj include-groups🤷 - See: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2509
```
QML_IMPORT_NAME = vpn.mozilla.lottie
QML_IMPORT_MAJOR_VERSION = 1

```
Since the code path is not relevant for 2.7, let's disable lottie on win so QA can get testable builds for now. 
Also let's have the CI error when there is no final msi.

### DIFF
--- a/.github/workflows/windows-build.yaml
+++ b/.github/workflows/windows-build.yaml
@@ -82,9 +82,10 @@ jobs:
         with:
             name: build
             path: windows/installer/x64/MozillaVPN.msi
-
+            if-no-files-found: error 
       - name: Upload unsigned app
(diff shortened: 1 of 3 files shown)


mkestler-mozilla (approved):

bakulf:
@strseb can you file an issue to enable Lottie again ?

Issue #1532, opened 2021-08-11
Add a generic error to handle when a server location is down


mariaberlinger:
#### VPN version:
* 2.5.0;

#### Affected Platforms:
* All;

#### Preconditions:
* Have VPN installed;
* Have an account with subscription;

#### Steps to reproduce:
1. Open the VPN and sign in with the account from prerequisites; 
1. Enable the multi-hop feature;
1. Turn the VPN on;

#### Expected result:
* If one of the servers is down a specific error message is displayed;

#### Actual result:
* The connection status jumps to “Connecting…” state or “No signal”;

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-858)

data-sync-user:
➤ Lesley Norton commented:

Link to Figma spec for this work:

https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=206%3A5300 ( https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=206%3A5300|smart-link )

data-sync-user:
➤ Valentina Virlics commented:

All dependent issues that have details or a way to be checked, were marked as qa-verified. 

Please add a specific label or a comment for the rest. Thanks!

data-sync-user:
➤ Valentina Virlics commented:

All dependent issues have a qa-not-actionable or a qa-verified label.

Closing this.

PR #2537, opened 2022-01-07
Detect handshake failure and report unreachable servers

oskirby:
This PR attempts to detect the failure of a connection handshake by setting a watchdog timer in the `Controller` class, and refactoring the platform controllers to manage the activation of each hop individually. Combining these two changes allows the `Controller` to detect when a particular server has failed.

Once a server has been deemed to be unreachable, it is deprecated by setting a cooldown timeout in the `Server` model, during which time that server will ignored during server selection. If there are no servers available when attempting to activate the VPN, then a signal will be emitted indicating that the location is down.

Some possible room for improvement:
- By integrating the cooldown timeouts with the `Server` model, the cooldowns are reset whenever updating the server list. This should be fine in production when the update rate is relatively infrequent but could make this difficult to test in production.
- We don't limit the number of connection attempts before giving up, so it might take a very long time for some locations to report unreachability if there is a large number of servers.

See: #1532


src/mozillavpn.cpp
```diff
@@ -127,6 +127,10 @@ MozillaVPN::MozillaVPN() : m_private(new Private()) {
             setState(StateBackendFailure);
           });
 
+  // TODO: This should trigger a server unavailable modal.
```
flozia, line 130: Created PR https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2538 for launching the modal.

src/mozillavpn.h
```diff
@@ -439,6 +443,7 @@ class MozillaVPN final : public QObject {
   QTimer m_alertTimer;
   QTimer m_periodicOperationsTimer;
   QTimer m_gleanTimer;
+  QTimer m_serverCooldownTimer;
```
mbirghan, line 446: Where do we use this?

src/mozillavpn.h
```diff
@@ -362,6 +363,9 @@ class MozillaVPN final : public QObject {
 
   void maybeRegenerateDeviceKey();
 
+  QList<Server> filterServerList(const QList<Server>& servers) const;
+  void checkServerCooldownTimeouts();
```
mbirghan, line 367: Where do we use this?

mbirghan (changes requested):

src/mozillavpn.h
oskirby: Thank you! I forgot to remove that while refactoring.

src/mozillavpn.h
oskirby: Thank you! I forgot to remove that while refactoring.

flozia (approved):

src/controllerimpl.h
```diff
@@ -46,8 +46,8 @@ class ControllerImpl : public QObject {
   // this method is called, the VPN client is in "connecting" state.  This
   // state terminates when the "connected" (or the "disconnected") signal is
   // received.
-  virtual void activate(const QList<Server>& serverList, const Device* device,
-                        const Keys* keys,
+  virtual void activate(const Server& server, const Device* device,
```
strseb, line 49: NIT: Since we're gonna use every field of HopConnection, maybe we can simplify this to `activate(Hop,Device,Keys)`?

src/platforms/linux/linuxcontroller.cpp
```diff
@@ -168,6 +120,7 @@ void LinuxController::operationCompleted(QDBusPendingCallWatcher* call) {
   emit disconnected();
 }
 
+#if 0
```
strseb, line 123: Should this be removed?

strseb (approved):
I have left a few comments, feel free to address them (or not :) ) 

I have a more general question here. It seems we're expecting the Controller to emit `connected` when the handshake is complete right (and timeout if not)?     
Because that behavior when we emit connected is AFAIK totally not uniform here. 
- on android emit connected after calling `wgup(conf, tun-handle) ` so if wg-go accepts the conf and handle we're connected here. 
- on windows i __AFAIK__ "activated" means: "tunnel-deamon is running and neither wintun, split-tunnel or wg-go had an error" 

Android definitely and windows AFAIK are not really checking for handshake errors from wg-go. 

If my assumption is true, i think this is still fine but we should open follow-ups to make sure all controllers have the same "idea" of what which event means :)

src/controllerimpl.h
oskirby: I like this, a lot. Thanks for the suggestion.

src/platforms/linux/linuxcontroller.cpp
oskirby: Yes, it should. I must have forgotten to clean this up after it was no longer necessary.

oskirby:
To address your comments @strseb 

You are correct that the idea is for the `connected` signal to be emitted only once the Wireguard handshake has been completed, and this is mostly important for the desktop platforms where things get complicated by multihop. For the desktop platforms, this is now generated by the `Daemon` class, which starts a timer to check for the completion of the handshake by inspecting the connection status (this change originated in PR #2291).

For mobile platforms, which don't support multiple nested connections, the timing is less important because we only ever have a single connection to worry about. In these cases, as I understand it, the `connected` signal will be generated as soon as `wgup` (or the iOS equivalent) is completed and we are simply falling back to the status quo method of relying on the `ConnectionHealth` class to detect a bad connection.

However, thanks for bringing that up, because it seems that the `ConnectionHealth` won't set a cooldown if it deems a server switch to be necessary, so it's unlikely that this feature works as expected on mobile.

mbirghan (approved):

PR #2540, opened 2022-01-07
The first version of a unit-test script on mac

bakulf:


lottie/tests/qml/qml.pro
```diff
@@ -7,7 +7,6 @@ QT += qml quick
 TEMPLATE = app
 TARGET = tst_lottie
 
-CONFIG += debug
```
bakulf, line 10: keep this

lottie/tests/unit/unit.pro
```diff
@@ -28,8 +28,6 @@ SOURCES += \
     testnavigator.cpp \
     testwindow.cpp
 
-CONFIG += debug
```
bakulf, line 31: keep this

scripts/scope_only_change.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
bakulf, line 1: header here with copyright/left.

bakulf (commented):
All good! But this is mac specific. Can we have something similar to linux or other platforms?
And maybe move files in subfolders to cleanup the `scripts` folder?

PR #2542, opened 2022-01-07
Fix mobile onboarding swipe

### LINKED ISSUE #2529: [mobile] Wrong slide displayed when swiping right on the first slide from onboarding screen
**VPN version:**

* 2.7.0 (16414640);

**Tested Platforms**:

* Android;

**Prerequisites**:

* Have Mozilla VPN app installed;

**Steps to reproduce**:
1. Open Mozilla VPN app;
2. Swipe right on the onboarding screen;
3. Observe the behavior;

**Expected result**:

* Forth slide is displayed;

**Actual result**:

* Second slide is displayed;

**Notes:**

* The second slide is displayed for both right and left swipe while in the first onboarding slide.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1518)

flozia:
Fixes #2529


src/ui/views/ViewMobileOnboarding.qml
```diff
@@ -122,18 +122,62 @@ VPNFlickable {
...
+                            } else if (swipeDirection > 0) {
+                                goToNextSlide();
+                            } else {
+                                goToNextSlide();
+                            }
+                        }
+
+                        onPositionChanged: {
+                            const parsedMouseX = parseInt(mouseX);
+                            if (previousMouseX > parsedMouseX) {
```
bakulf, line 170: This makes the area extremely reactive. We need to test what happens with small clicks. If it's too much, maybe we should have a bit of tolerance

lesleyjanenorton (commented):
Tiniest nit: It would be nice to maintain the 'swipe' effect on the image during transitions triggered by click, or even just synchronize the fade out of the image so that it matches the text. The image seems to blink out a bit quickly right now when clicking through the slides.

lesleyjanenorton (approved):

PR #2543, opened 2022-01-07
Crash reporter windows

mkestler-mozilla:
Windows error reporting using WER.


src/crashreporter/crashui/main.qml
```diff
@@ -0,0 +1,114 @@
...
+
+    width: fullscreenRequired() ? Screen.width : VPNTheme.theme.desktopAppWidth;
+    height: fullscreenRequired() ? Screen.height : VPNTheme.theme.desktopAppHeight;
+
+    //These need to be bound before onComplete so that the window buttons, menus and title bar double click behave properly
+    maximumWidth: fullscreenRequired() ? Screen.width : VPNTheme.theme.desktopAppWidth;
+    maximumHeight: fullscreenRequired() ? Screen.height : VPNTheme.theme.desktopAppHeight;
+
+    //% "Mozilla Crash Reporter"
+    title: qsTrId("vpn.crashreporter.mainTitle")
```
lesleyjanenorton, line 35: Let's add this string to strings.yaml and call it from there

src/crashreporter/crashui.cpp
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "crashui.h"
```
oskirby, line 5: Typically, we try to organize the `#include` directives into three groups as follows:
1. Project-local includes (with the `.h` for this `.cpp` always coming first).
2. Qt includes
3. System global includes

And within each group the includes should be sorted alphabetically (unless there's a technical reason otherwise).

src/crashreporter/crashuploader.cpp
```diff
@@ -0,0 +1,145 @@
...
+  urlStr = Constants::CRASH_STAGING_URL;
+#endif  // MVPN_DEBUG
+
+  logger.debug() << "Uploading to : " << urlStr;
+  QUrl url(Constants::CRASH_PRODUCTION_URL);
+  QNetworkRequest request(url);
+  request.setHeader(QNetworkRequest::UserAgentHeader, "mozillaVPN");
+  // Qt changes the content type automatically to multipart/mixed.  Socorro
+  // doesn't support that, so we have to set it ourselves and create a custom
+  // bounday.
```
oskirby, line 102: Minor typo, the comment probably should say `boundary`

src/crashreporter/crashuploader.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  QHttpPart versionPart;
+  versionPart.setHeader(QNetworkRequest::ContentDispositionHeader,
+                        "form-data; name=\"Version\"");
+  versionPart.setBody(APP_VERSION);
+  multipart->append(namePart);
+  multipart->append(formPart);
+  multipart->append(versionPart);
+
+  auto urlStr = Constants::inProduction() ? Constants::CRASH_STAGING_URL
```
oskirby, line 90: I'm a little uncertain about the use of `Constants::inProduction()` here, for two reasons:
1. It's not clear to me that the crash reporter tool will be able to load the settings necessary to determine if the user was connected to staging or not (and it will just default to production).
2. The builds aren't different between staging and production (it's just a setting). So I don't think there's any value in uploading the crash reports to different locations.

src/main.cpp
```diff
@@ -26,6 +28,16 @@ int main(int argc, char* argv[]) {
   }
 #  endif
 
+#endif
```
oskirby, line 31: I would have preferred for the crash reporter to be a separate executable, just in case the crash occurs earlier than this point. But I understand that may also be quite a big change, so I'm okay with leaving this as-is.

Assuming we keep everything as a monolithic executable, we should update the `--help` text to mention this option.

src/crashreporter/crashreporter.pri
```diff
@@ -0,0 +1,55 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+QT += network
+QT += quick
+QT += widgets
+QT += charts
```
oskirby, line 8: What are `charts` used for?

src/crashreporter/crashui/main.qml
```diff
@@ -0,0 +1,111 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
oskirby, line 1: As a minor wishlist item, if the size of the dump files are large, or the internet connection is slow, it may be appreciated to include a progress bar or something to indicate that the upload is in progress. Some visible change of state after clicking submit would also be helpful for the user.

(I have not actually seen this in action, please disregard my comment I'm interpreting the QML wrong).

oskirby (approved):
Looks pretty good, just a couple of minor cleanup items and I think we can merge it.

I've added some notes about bigger wishlist kinda stuff that's too big for this release as well.

src/crashreporter/crashui/main.qml
```diff
@@ -85,22 +85,19 @@ Window {
...
+              nchors.horizontalCenter: parent.horizontalCenter
               horizontalAlignment: Text.AlignHCenter
               text: qsTrId("vpn.crashreporter.description")
           }
 
           VPNButton {
             text: qsTrId("vpn.crashreporter.sendButtonLabel");
-            anchors.left: parent.left
-            anchors.right: parent.right
+            nchors.horizontalCenter: parent.horizontalCenter
```
lesleyjanenorton, line 95: `anchors.horizontalCenter`

lesleyjanenorton: Also on L88

src/main.cpp
mkestler-mozilla: I would have preferred a 2nd executable as well, but we couldn't due to settings and strings.  I'm not sure that help should display this as it really isn't something a user should do, and we wouldn't want it to become CLI API somehow.

src/crashreporter/crashuploader.cpp
mkestler-mozilla: I made this change at Baku's request.  The way it should work is that is the user has set staging in the client, then we use staging for the dump, unless it is a debug build which should always use staging.  The idea is that dev mode crashes don't pollute the crash reports, and we also have some means of testing against a non-production crash collector.  If inProduction doesn't do this, let me know.  All settings should be the same as the client, as this is the same exe.

src/crashreporter/crashui/main.qml
mkestler-mozilla: UX didn't want to do anything like this, just a fire and forget sort of model.  I did mention the same thing in design talks.

src/crashreporter/crashuploader.cpp
oskirby: Okay, then I'll defer to baku's judgment on this one.

PR #2557, opened 2022-01-11
Use server count to disable selection of servers

flozia:
As soon as we landed #2537 we can disable servers in the server list view in case they are currently not available.


oskirby (approved):
Looks good!

lesleyjanenorton (approved):

lesleyjanenorton:
Closing in favor of #2569 - which is exactly the same but with a cleaned up commit history and strangely passes the functional tests that I could not get to pass here.

PR #2531, opened 2022-01-06
Server unvailable system notification

flozia:
Launch a system notification when all servers in a city have been exhausted.

**Screenshots**
| System notification | Notification preferences |
| --- | --- |
| server-unavailable-notification | server-unavailable-preferences |

**Design**
[Figma spec](https://www.figma.com/file/ziJ5tGCqhfst4svjtKJlbr/2.7--Feature-Enhancements?node-id=357%3A7648)

**ToDo**
- [x] Add _Sever unavailable notification_ feature
- [x] Add option to for _System preferences_
- [x] Add final strings for system notification
- [x] Notification preference option position and strings

**Linked issues**
- Resolves: #2297

PR #2560, opened 2022-01-11
Trigger server unavailable notification

### LINKED ISSUE #2555: Trigger server unavailable system notification
Trigger the server unavailable system notification that was introduced by PR #2531 when all servers in a city have been exhausted with `NotificationHandler::instance()->serverUnavailableNotification()`. The best place to do this is probably in the `serverUnavailable` method introduced by PR #2531 in `src/controller.cpp`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1529)

flozia:
Trigger the server unavailable system notification that was introduced by PR #2531.

Resolves #2555


src/controller.cpp
```diff
@@ -588,6 +588,7 @@ void Controller::serverUnavailable() {
 
   if ((m_state == StateOn) || (m_state == StateSwitching) ||
       (m_state == StateConnecting)) {
+    NotificationHandler::instance()->serverUnavailableNotification();
```
oskirby, line 591: This should probably be done in the `MozillaVPN` class where we connect to the `readyToServerUnavailable` signal to keep the UI behavior separated from the controller. See the `connect()` call at src/mozillavpn.cpp:130

oskirby (changes requested):

oskirby (approved):

PR #2562, opened 2022-01-12
Correctly run QML tests

birdsarah:
* Put test runs in separate steps so that if each test step fails.
* Correctly return an exit code from qml tests.


birdsarah:
The fact that these failed is good! They should have been failing. Baku's PR, recently merged into main, fixes the failing QML test which I'll now merge into this PR.

birdsarah:
Done another merge from main with the glean upgrade in. This'll be a true run of the glean tests.

lesleyjanenorton (approved):

PR #2563, opened 2022-01-12
Fix crash reporter after updated theme api from #2499

birdsarah:


birdsarah:
@mkestler-mozilla requesting your review. I merged #2499 PR from baku to fix the QML tests without updating his branch to main. 

This caused everything to fail as your PR had been merged in the meantime which used the old API. I've taken my best stab at fixing the mess I made but definitely need your review.

mkestler-mozilla (approved):

PR #2564, opened 2022-01-12
Add server unavailable to inspector

flozia:
Adds the ability to timeout all servers of a city through the Mozilla VPN Inspector using `force_server_unavailable {countryCode} {cityCode}`.


strseb (approved):

src/ui/main.qml
```diff
@@ -441,6 +441,13 @@ Window {
         id: serverUnavailablePopup
     }
 
+    Connections {
+        target: VPNController
+        function onReadyToServerUnavailable() {
+            serverUnavailablePopup.open();
+        }
+    }
```
flozia, line 449: I think this part was lost during a rebase with master. Also happy to create a separate PR to add this line back in.

src/controller.cpp
```diff
@@ -445,6 +445,16 @@ void Controller::handshakeTimeout() {
   activateInternal();
 }
 
+void Controller::setCooldownForAllServersInACity(const QString& countryCode,
+                                                 const QString& cityCode) {
+  logger.debug() << "Set cooldown for all servers in a city";
```
strseb, line 450: imho, we should have a `Q_ASSERT(!Constants::inProduction())` - so everyone knows this is for debugging / testing

strseb (approved):

src/controller.cpp
flozia: That is a good point! Unfortunately we’ll need to be able to use the method in production as well so QA can test the feature — I’ll add a comment for now if that is OK.

oskirby (approved):
Looks good!

PR #2574, opened 2022-01-13
✅: Taskcluster android signing

strseb:
Well taskgraph seems to generate a valid task-set for signing, so let's try!


strseb:
Hey @ahal - I was wondering if we could review this together.  This is now generating the task-graph in the right way ... i think. 

-  Added test parameters for main/release-branch-push and pr 
- We're having a new type build/android-release, this should only be run on main/release push
- New Kind Sign: Should also run only on main/release branches. 
- The release-build should get production signing and the debug build dep-sign. 
 
I am having a `releases/test` branch to test it out but fails for missing permissions.: `project:mozillavpn:releng:signing:format:autograph_apk`

-> https://firefox-ci-tc.services.mozilla.com/tasks/S_26QEx0TMmkJboHkvejNw

But i wanted to check in with you if that looks okay before i once again start asking for more permissions :D

ahal:
Hey @strseb, yes happy to go through this with you. My schedule is pretty busy today so not anticipating having time until tomorrow though.

strseb:
Now this is ready for review! 
Changelist
-> Added right scope so release-builds are getting the token access
-> Fixed x64 release-build
-> Any release build will now be signed.

Example job: 
https://firefox-ci-tc.services.mozilla.com/tasks/UyrZt6jHQQKdjXyHE6AxsA#artifacts

scripts/windows_build.sh
```diff
@@ -1,21 +0,0 @@
-#!/bin/sh
```
mkestler-mozilla, line 1: Was this intentionally deleted?

strseb: oops! - Not intentionally, no.

mkestler-mozilla (approved):

taskcluster/ci/sign/kind.yml
```diff
@@ -0,0 +1,35 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
ahal, line 1: I'd call the kind `signing` since that's the name we use in all the other taskgraph repos that have it. Will make it easier for folks to grep for and the like.

taskcluster/ci/sign/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+        kind: build
+        platform:
+            by-build-type:
+                android-x64/release: android/x86_64
+                android-x86/release: android/x86
+                android-arm64/release: android/arm64-v8a
+                android-armv7/release: android/armv7
+                default: android/all
+                
+        tier: 3
```
ahal, line 35: You'll want this to be tier 1, otherwise it won't show up in Treeherder. You're probably thinking of levels, where "Level 3" is more important than "Level 1". But Treeherder tiers are the opposite (Tier 1 is the most important)... I know it's confusing :(

taskcluster/mozillavpn_taskgraph/transforms/signing.py
```diff
@@ -13,12 +13,17 @@
...
+    "android-arm64/release",
+    "android-armv7/release"
 ]
 
 SIGNING_BUILD_TYPES = PRODUCTION_SIGNING_BUILD_TYPES + [
-    "android-debug",
+# Note: it appears we don't have infra for debug sign
+# contact releng if you need it :) 
+#    "android-debug",
+#
```
ahal, line 26: Yeah, we only found one set of credentials to use. I think if we want to sign debug things we'll either need to re-use some keys from another project, or ask SecOps to generate a new set for us. I'll add a JIRA ticket to our board to track it (but likely won't get prioritized until someone pushes for it).

ahal (changes requested):
This lgtm, though I'll admit to not being an expert on signing best practices..

I do have a couple minor issues for you to address.

taskcluster/ci/sign/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+
+group-by: build-type
+
+job-template:
+    description: Sign MozillaVPN for Android
+    worker: {} # Will be configured by signing:transforms
+    signing-format: autograph_apk
+    run-on-tasks-for: [github-push]
+    treeherder:
+        job-symbol: S
```
ahal, line 25: Typically the symbol for these is `Bs`. So in treeherder you'd see `B` as the build and `Bs` as the signed build.

taskcluster/ci/sign/kind.yml
strseb: Ah, yes that was exactly what i was thinking :D - Thanks for clearing that up!

taskcluster/mozillavpn_taskgraph/transforms/signing.py
strseb: No need to take action for this - i just left it as a breadcrumb in case someone later wants to add it. I don't think we really have a use-case for debug signing rn :D

taskcluster/ci/signing/kind.yml
```diff
@@ -22,7 +22,7 @@ job-template:
     signing-format: autograph_apk
     run-on-tasks-for: [github-push]
     treeherder:
-        job-symbol: S
+        job-symbol: BS
```
ahal, line 25: nit: typically the s is lower case.. but I feel like maybe I'm being too picky at this point :)

ahal (approved):

bakulf (approved):

PR #2597, opened 2022-01-17
Hide slide indicator for single tour features

flozia:
In case we only show one feature in the _What’s new_ panel there is a single page indicator that might be confusing. Let’s only show it if there is more than one feature to discover.

| Before | After |
| --- | --- |
| single-feature-slide-indicator-before | single-feature-slide-indicator-after |


lesleyjanenorton (approved):
🤩

PR #2610, opened 2022-01-19
Prevent VPN from reconnecting to a location that is currently “unavailable”

### LINKED ISSUE #2583: VPN is blocked in “Disconnecting…” state if a user retries to connect to an unavailable server
**VPN version:**

* 2.7.0 (2.202201140229);

**Platforms:**

* Windows 10;

**Prerequisites:**

* Have and account with a subscription for Mozilla VPN;
* Have the VPN app installed and turned ON and connected to Austria-Vienna city;

**Steps to reproduce:**
1.  Run the command: “force_server_unavailable au vie” in Inspector
2. Dismiss the server unavailable modal and the error screen;
3. Try to turn ON the VPN;

**Expected results:**

* The server unavailability message is displayed again (maybe?);

**Actual results:**

* The VPN is blocked in the Disconnecting state.

**Notes:**

* Signing out fixes the issue;
* Same happens for multi-hop;
* The issue is "qa-major" because the user is blocked in using the app;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1554)

flozia:
When a user tries to connect to an unavailable server the VPN is stuck in `StateDisconnecting`. This PR prevents the VPN from attempting to connect to a city which has all of their servers set to a cooldown period.

Resolves #2583

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -179,11 +179,19 @@ bool Controller::activate() {
   }
 
   if (m_state == StateOff) {
+    logger.debug() << "Test - try to turn on" << m_state;
+
     if (m_portalDetected) {
       emit activationBlockedForCaptivePortal();
       m_portalDetected = false;
       return true;
     }
+
+    if (hasCooldownForAllServersInACity()) {
+      logger.debug() << "Test - hasCooldownForAllServersInACity" << m_state;
+      return true;
+    }
+
     setState(StateConnecting);
   }
 
@@ -455,6 +463,12 @@ void Controller::setCooldownForAllServersInACity(const QString& countryCode,
   Q_ASSERT(vpn);
 
   vpn->setCooldownForAllServersInACity(countryCode, cityCode);
+}
+
+bool Controller::hasCooldownForAllServersInACity() {
+  logger.debug() << "Has cooldown for all servers in a city";
+
+  return true;
 }
 
 bool Controller::isUnsettled() { return !m_settled; }
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -469,9 +469,18 @@ bool Controller::hasCooldownForAllServers() {
   MozillaVPN* vpn = MozillaVPN::instance();
   Q_ASSERT(vpn);
 
-  return vpn->hasCooldownForAllServersInACity(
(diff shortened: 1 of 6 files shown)


strseb (approved):

src/controller.cpp
```diff
@@ -457,6 +463,17 @@ void Controller::setCooldownForAllServersInACity(const QString& countryCode,
   vpn->setCooldownForAllServersInACity(countryCode, cityCode);
 }
 
+bool Controller::hasCooldownForAllServers() {
```
oskirby, line 466: I'm concerned that this doesn't perform the same check for the entry server, which may lead to a similar bug occurring in the multihop case.

src/controller.cpp
```diff
@@ -184,6 +184,12 @@ bool Controller::activate() {
       m_portalDetected = false;
       return true;
     }
+
+    if (hasCooldownForAllServers()) {
```
oskirby, line 188: I think the root cause of the issue here is more to do with the fact that calling `Controller::disconnect()` doesn't necessarily lead to a disconnected signal if we fail before instructing the platform to connect. This doesn't seem like it entirely fixes the problem, and we may have similar edge cases in server switching where this could occur.

oskirby (commented):
I don't think this is the correct solution to the problem, ultimately this check is expected to occur in `Controller::activateInternal()` when ensuring that the server list is non-empty, and the implementation of `Controller::disconnect()` must take into account the fact that the platform may not have been instructed to connect yet.

As-is, this might be good enough to get us through the 2.7 release, in which case I would probably want to fix it either in `Controller::disconnect()` or in the platform layers.

src/controller.cpp
flozia: Added a check for multi-hop.

flozia:
@oskirby I added a check to handle multi-hop in `Controller::hasCooldownForAllServers()`. As discussed I created issue #2615 so we’ll make sure to refactor this not so great fix. Do you feel comfortable if I go ahead and merge this PR or should we try to carve out more time too land a more adequate fix?

oskirby:
@flozia I think this is good enough to merge for 2.7, and I'll look into a revisiting this for 2.8.

oskirby (approved):

Issue #2590, opened 2022-01-17
Localization issues for VPN 2.7


data-sync-user:
**_A place for all the localization issues found during the testing activities for VPN 2.7._**

1. **[mobile] “Sign up” and “Already a subscriber?” buttons are not translated in German**

**Prerequisites:**

* Have Mozilla VPN installed on your device;
* Have the app language set to German;

**VPN version:**

* 2.7.0 (16423770)

**Affected Platforms:**

* Android;

**Steps to reproduce:**
1. Open the Mozilla VPN app on your device;
2. Go through the slides from on the onboarding screen;

**Expected result:**

* The information on the onboarding screen is fully translated;

**Actual result:**

* The “Sign up” and “Already a subscriber?” buttons are not translated;

**Note:**

* see the attached screenshot;

2. **“Captive portal detected!” screen  isn’t translated in Spanish**

**VPN version:**

* 2.7.0 (2.202201140229);

**Affected Platforms:**

* Windows 10;

**Preconditions:**

* Have Mozilla VPN app installed and opened;
* Be signed in while on Stage server;
* Have the VPN language set to Spain;
* "Guest Wi-fi portal alert" option is checked;
* VPN is OFF

**Steps to reproduce:**

1. Turn ON the VPN;
1. Navigate to Settings - Get help - Developer Options - Open Inspector;
1. Enter the following command in Inspector: force_captive_portal_detection;
1. Observe the “Captive Portal detected!” screen;

**Expected result:**

* The “Captive Portal detected!” screen is translated in Spanish;

**Actual result:**

* The “Captive Portal detected!” screen is in English;

**Note:**

* see the attached screenshot;

3. **[Mobile] The message under the Data collection and use option in not translated**

**Prerequisites:**

* Have Mozilla VPN installed on your device;
* Have the app language set to something other than English;

**VPN version:**

* 2.7.0 (16425470)

**Affected Platforms:**

* Android;

**Steps to reproduce:**

1. Open the Mozilla VPN app on your device;
1. Go to Settings > system Preferences and take a look at the Data collection and use option ;

**Expected result:**

* The information under the Data collection and use option is translated;

**Actual result:**

* The information under the Data collection and use option is NOT translated;

**Note:**

* see the attached screenshot;

4. [*mobile] Server Unavailable screen is not translated*

**Prerequisites:**

* Have Mozilla VPN installed on your device;
* Have the app language set to German, Spanish, French;

**VPN version:**

[...]

PR #2632, opened 2022-01-24
Android localization issues

flozia:
It seems like we did not fetch the newest translation files for Android builds anymore. This should resolve the open localization issues that are related to #2590.


lesleyjanenorton (approved):
🤩🥳

Issue #2621, opened 2022-01-20
Add option to force port 53/DNS for Wireguard traffic


data-sync-user:
A common reason why users may be unable to connect successfully to a Wireguard server, is because the UDP traffic generated by the VPN appears “suspicious” and may be blocked at their internet gateway.

To help users remedy this situation, we would like to add an option to force their Wireguard traffic to make use of UDP port 53/DNS. Because DNS is such a ubiquitous protocol, it is usually permitted by firewall devices and may allow a user to workaround an overly aggressive firewall.

Acceptance Criteria:

* An option is added to the network settings to force the use of port 53/DNS.
* When this option is enabled, all the user's Wireguard traffic is sent with a destination port of 53.
* Glean telemetry is added to report the frequency at which this setting is used.
* (optional) a warning is displayed on mobile that this option does not work on multihop.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1592)

data-sync-user:
➤ Luminita Buruian commented:

Owen Kirby Could you please let us know what the QA needs to do on this one, if anything?

Thank you.

CC: Madalin Cotetiu Valentina Virlics

data-sync-user:
➤ Owen Kirby commented:

I think that a minimum, we require QA to ensure that the VPN appears to connect normally both with this new option enabled and disabled.

To verify that the feature is working as expected, we would need to use a network analysis tool such as wireshark to see that the destination port for Wireguard traffic is set to port 53 when this feature is enabled.

data-sync-user:
➤ Owen Kirby commented:

Luminita Buruian That’s perfect! Your test methodology with Wireshark is correct, and the feature appears to be working exactly as expected.

The malformed packet warning is to be expected when this option is enabled, because Wireshark is trying to parse Wireguard traffic as though it were DNS.

data-sync-user:
➤ Luminita Buruian commented:

Owen Kirby Thank you for confirming.

Closing the issues as verified as fixed.

data-sync-user:
➤ Luminita Buruian commented:

Owen Kirby We need you help with a couple more clarifications about what the check regarding this new option, so that we can prepare out tests accordingly:

  1. Do we need to somehow check Glean telemetry to see if an event is added to report the frequency at which this setting is used? 
If so, what exactly do we need to look for?
[...]

PR #2634, opened 2022-01-24
Add option to tunnel VPN traffic on port 53/DNS

oskirby:
This option allows the user to force their VPN traffic to use UDP port 53/DNS, which may do a better job of evading aggressive network firewalls. This option can be supported generically in all cases except mobile platforms when combined with multihop.

tunnel-port53

Some room for improvement: 
- [x] Wordsmithing around how this option is displayed in the settings menu.
- [ ] Display a warning that this option has no effect when combined with mutlihop on mobile devices.

See: #2621

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -239,6 +239,11 @@ void Controller::activateInternal() {
   if (!FeatureMultiHop::instance()->isSupported() || !vpn->multihop()) {
     exitHop.m_excludedAddresses.append(exitHop.m_server.ipv4AddrIn());
     exitHop.m_excludedAddresses.append(exitHop.m_server.ipv6AddrIn());
+
+    // If requested, force the use of port 53/DNS.
+    if (settingsHolder->tunnelPort53()) {
+      exitHop.m_server.forcePort(53);
+    }
   }
   // For controllers that support multiple hops, create a queue of connections.
   // The entry server should start first, followed by the exit server.
@@ -249,6 +254,10 @@ void Controller::activateInternal() {
       logger.error() << "Empty entry server list in state" << m_state;
       serverUnavailable();
       return;
+    }
+    // If requested, force the use of port 53/DNS.
+    if (settingsHolder->tunnelPort53()) {
+      hop.m_server.forcePort(53);
     }
 
     hop.m_hopindex = 1;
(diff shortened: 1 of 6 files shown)


translations/strings.yaml
```diff
@@ -310,6 +310,8 @@ settings:
     value: Connect VPN on startup
     comment: Startup refers to a device's startup, not the VPN client's.
   startAtBootSubtitle: Mozilla VPN will launch and connect when you start up your device
+  tunnelPort53: Tunnel VPN using port 53/DNS
```
flodolo, line 313: Question for copy: we already have the details in the description, do we need to repeat the port number here? `Tunnel VPN through DNS port` seems clear enough.

From a localization perspective, if we end up using this form, I would only ask to add a comment that "tunnel" is a verb, as in "direct traffic through".

src/controller.cpp
```diff
@@ -267,6 +276,9 @@ void Controller::activateInternal() {
       serverUnavailable();
       return;
     }
+    // NOTE: For platforms without multihop support, we cannot emulate multihop
+    // and use port 53 at the same time. If the user has selected both options
+    // then let's choose multihop.
```
strseb, line 281: Should we maybe have a VPNCheckBoxAlert similar to the "vpn needs to be off" below the setting checkbox when multihop-mobile is active? Something like "this setting is ignored when multihop is active?"

translations/strings.yaml
oskirby: Good suggestions, thanks

src/models/server.cpp
```diff
@@ -191,27 +194,17 @@ uint32_t Server::choosePort() const {
...
-    for (const QPair<uint32_t, uint32_t>& range : m_portRanges) {
-      if (r <= (range.second - range.first)) {
-        port = r + range.first;
-        break;
-      }
-      r -= (range.second - range.first + 1);
-    }
-    Q_ASSERT(port != 0);
+  // Pick a port at random.
+  quint32 r = QRandomGenerator::global()->generate() % length;
```
mkestler-mozilla, line 198: Could this end up generating a port 53, or cause the resulting range to be 53?  Is that still an issue with Mullvad blocking it?

oskirby: Mullvad no longer blocks port 53, and we rely on that behaviour to support custom DNS servers.

mkestler-mozilla (approved):

flodolo (approved):

PR #2638, opened 2022-01-25
basti/fixAndroidQt6StackView

strseb:
This is a wierd one, sorry. 

On Qt6/Android, when moving to either StateMainView/StateDeviceLimit, the page would be blank. No qml error given. 
I've digged into it using the debugger. It seems when `initialItem` was _supposed_ to be handled as a qrc-URL it will be parsed into ... a float (?). So ie. doing `stackview.onCompleted:push(initialItem)` will correctly point out that there is a type mismatch. 
Thus the view will be blank. 

From the [qml docs](https://doc.qt.io/qt-5/qml-qtquick-controls2-stackview.html#initialItem-prop): 
> Specifying an initial item is equivalent to:
> Component.onCompleted: stackView.push(myInitialItem)

This however this still works with qrc-urls! So I changed all usages of this to use this style. Also i've added an error message if that case is popping up in our code again 🙈 

 
Sidenote: that qmljsdebugger stuff is totally optional (i can remove it np)  but that piece helped me so much in debugging this, maybe we should keep it :D

### DIFF
--- a/nebula/ui/components/VPNStackView.qml
+++ b/nebula/ui/components/VPNStackView.qml
@@ -10,11 +10,19 @@ import Mozilla.VPN 1.0
 
 StackView {
     id: stackView
-    Component.onCompleted: VPNCloseEventHandler.addStackView(stackView)
 
     onCurrentItemChanged: {
         var objString = currentItem.toString().split("(")[0];
         VPN.currentView = objString.split("_QML")[0];
+    }
+
+    Component.onCompleted: function(){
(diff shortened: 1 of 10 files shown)


strseb:
With this we now can navigate to any page, still looks wonky tho. 
image

flozia (approved):
Cool that you found a way to work around the issues with `initialItem` for Android. I think enabling `qml_debug` is helpful.

lesleyjanenorton (approved):
Nice! let's keep the debugger.

PR #2646, opened 2022-01-26
Add custom password input

### LINKED ISSUE #1895: Password input field
We need to build an input for passwords or extend our existing text input component. This component will be used for the _In-App Authentication_ flow. The design is not final: [Figma](https://www.figma.com/file/gSRPHqjzqjlMA9rXDc5SqH/VPN_InApp_SignIn-Auth?node-id=135%3A5929).

The component should:
- mask the characters
- contain a button for toggling characters in plain text
- show error and validation states

image

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1054)

flozia:
This extends the VPNTextInput component with a variant for handling password inputs.

Passing the prop `isPassword: true` to VPNTextInput enables a “password mode”. The visibility of the character typed into the input can be toggled with the icon button on the right.

| Characters masked | Characters plain text |
| --- | --- |
| image | image |

Resolves #1895


lesleyjanenorton (approved):
🥳

PR #2651, opened 2022-01-28
Fix ring animation for Qt6 iOS builds

flozia:
This PR should re-add support for the ring animation shader when building for iOS with Qt6.

Requiring GLSL `310 es` for the ring animation shader should be enough for supporting the necessary functions. I went for `440` initially because otherwise the builds failed. What I’ve missed previously is that when compiling for OpenGL `ES 3.0 ` — `qsb` the tool used for baking the shader — is throwing a compilation error when `#version` is not specified in the first line of the file.


nebula/ui/resources/shaders/src/animatedRings.frag
```diff
@@ -1,8 +1,9 @@
+#version 310 es
```
mbirghan, line 1: This should be moved below the comment.

mbirghan (approved):

nebula/ui/resources/shaders/src/animatedRings.frag
mbirghan: Ok then add a  comment so we know why we did it that way

PR #2652, opened 2022-01-28
Show masked images when using Qt6

flozia:
`OpacityMask` as `layer.effect` is not working as expected so that images do not show when compiling with Qt6.

### DIFF
--- a/nebula/ui/components/VPNPanel.qml
+++ b/nebula/ui/components/VPNPanel.qml
@@ -13,7 +13,6 @@ Item {
     property alias logoTitle: logoTitle.text
     property alias logoSubtitle: logoSubtitle.text
     property var logoSize: 76
-    property var maskImage: false
     property var isSettingsView: false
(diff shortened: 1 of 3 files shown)


nebula/ui/components/VPNPanel.qml
```diff
@@ -47,19 +46,6 @@ Item {
...
-                    id: mask
-
-                    anchors.fill: parent
-                    radius: logoSize / 2
-                    visible: false
-                }
-
-                layer.effect: VPNOpacityMask {
-                    maskSource: maskImage ? mask : undefined
-                }
```
flozia, line 62: If I’m not mistaken this mask is supposed to only be used for `VPNAvatar`. We already have a mask for the avatar in that component, but it did not work as expected. Fixed the mask there so this one should be redundant.

lesleyjanenorton (approved):

PR #2653, opened 2022-01-30
Fix Android Qt6 button focus styling

flozia:
Fixes the focused state style for buttons using the `VPNClickableRow` component.


strseb (approved):
Nice wfm! Screenshot_20220201-133024

PR #2675, opened 2022-02-04
GithubCI: Let our UnitTests run on Qt5 & qt6

strseb:
We should start testing for qt6 :)


bakulf (approved):

bakulf:
This does not work yet because qt6 has changed the default redirect policy: https://doc.qt.io/qt-6/network-changes-qt6.html#redirect-policies

bakulf:
Plus the new redirect policy approach seems buggy.

bakulf (approved):

PR #2698, opened 2022-02-09
Fix the qt6 HTTP redirection bug - Fixes VPN-1700

bakulf:


src/networkrequest.cpp
```diff
@@ -777,6 +777,34 @@ void NetworkRequest::replyFinished() {
...
+      logger.debug() << "Final URL:" << url.toString();
+#endif
+
+      m_request = QNetworkRequest(url);
+      m_request.setHeader(QNetworkRequest::ContentTypeHeader,
+                          "application/json");
+      m_request.setAttribute(QNetworkRequest::RedirectPolicyAttribute,
+                             QNetworkRequest::NoLessSafeRedirectPolicy);
+
+      m_reply = nullptr;
```
oskirby, line 801: I'm concerned that this may cause a memory leak of the failed `QNetworkReply` now that `NetworkRequest::maybeDeleteLater()` will do nothing if `m_reply` is null.

bakulf: We do not leak it because at line 868 we call `m_reply->setParent(this);`

oskirby (approved):

PR #2702, opened 2022-02-09
Replace QVariant::type() with QVariant::typeId()

bakulf:


bakulf:
qt5 doesn't seem happy about this change.

PR #2707, opened 2022-02-09
Openssl script update

mkestler-mozilla:
Disable engines in openssl so that we do not load them at runtime.
The -DOPENSSL_NO_ENGINES switch disables all use of engines in openssl.  It will still load an openssl.cnf file, but not load any engines from it.

### DIFF
--- a/scripts/openssl_compile.bat
+++ b/scripts/openssl_compile.bat
@@ -65,7 +65,7 @@ IF %ERRORLEVEL% NEQ 0 (
 )
 
 mkdir %BUILDDIR%
-xcopy /E /I "C:\Program Files\Common\SSL\" C:\MozillaVPNBuild\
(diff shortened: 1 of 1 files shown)


scripts/openssl_compile.bat
```diff
@@ -38,7 +40,7 @@ ECHO Compiling openssl...
 
 pushd %1
 
-perl Configure VC-WIN64A --release --prefix=%BUILDDIR% --openssldir=%BUILDDIR%\SSL no-tests
+perl Configure VC-WIN64A --release --prefix=%BUILDDIR% --openssldir=%BUILDDIR% no-tests -DOPENSSL_NO_ENGINE
```
bakulf, line 43: Can we use some default openssldir? such as C:\Program Files\OpenSSL-Win64 ? or something similar to make it even harder setting a custom config file?

bakulf (approved):

PR #2724, opened 2022-02-11
Add support for jammy ubuntu release

bakulf:


oskirby (changes requested):
For completeness, we should also update `scripts/linux/script.sh` to accept `jammy` as a valid release. In particular, take a look at the `case` statement at line 180.

oskirby (approved):

PR #2726, opened 2022-02-11
Add standalone password input component

flozia:
This PR adds a standalone password input component and some QML test for toggling the password visibility.


nebula/ui/components/forms/VPNTextField.qml
```diff
@@ -38,7 +36,7 @@ TextField {
     leftPadding: VPNTheme.theme.windowMargin
     rightPadding: VPNTheme.theme.windowMargin
```
lesleyjanenorton, line 37: Tiniest/non-blocking nit : let's increase `rightPadding` to something that takes the eye icon into account. Currently I think it's possible for there to be an overlap between the characters and the icon if the password is very long.

lesleyjanenorton (approved):
NICE! Have only one nit but I'll file a follow-up bug.

nebula/ui/components/forms/VPNTextField.qml
flozia: Good point — increased `rightPadding`!

PR #2753, opened 2022-02-14
Add strings for connection info screen

flozia:
New strings needed for the new connection info screen redesign according to the [Figma spec](https://www.figma.com/file/3V7aahpQ096Qsp2L3SCT6h/Connection-Screen%2FGraph-Redesign?node-id=239%3A2311).


translations/strings.yaml
```diff
@@ -337,23 +337,55 @@ telemetryPolicyView:
...
   question: Allow Mozilla VPN to send technical, interaction, and campaign and referral data to Mozilla?
 
+genericError:
+  unexpected: Unexpected error
 
+# Connection Info Screen
 connectionInfo:
   labelBitps:
-    value: bit/s
+    value: bits
```
flodolo, line 346: This applies to VPN as well: https://mozilla-l10n.github.io/documentation/localization/making_string_changes.html

You need new IDs for this type of changes.

translations/strings.yaml
```diff
@@ -337,23 +337,55 @@ telemetryPolicyView:
...
+    value: Tbits
     comment: Terabits per second
+  labelPing:
+    value: Ping
+    comment: Latency of the connection
+  unitPing:
+    value: ms
+    commnent: milliseconds
+  loadingIndicatorLabel:
+    value: Testing speed
```
flodolo, line 367: Why is there no ellipsis?

translations/strings.yaml
```diff
@@ -337,23 +337,55 @@ telemetryPolicyView:
...
+    comment: Shown to the user while waiting for the connection speed test results
+  errorMessage: Looks like there was an unexpected error while trying to connect. Please try again or check your connection.
+  # Speed list header for medium and fast speeds
+  listHeaderDefault:
+    value: At your current speed, here's what your device is optimized for
```
flodolo, line 372: Use typographic apostrophes `’`. This is also pointed out by the test failure (linter)

translations/strings.yaml
```diff
@@ -337,23 +337,55 @@ telemetryPolicyView:
...
+    comment: Headline for a list of things that the current connection is capable of.
+  # Speed list header for low speeds
+  listHeaderSlow:
+    value: It looks like the connection to this server location is a bit slow, you can try
```
flodolo, line 376: Why is this missing the closing colon?

translations/strings.yaml
```diff
@@ -337,23 +337,55 @@ telemetryPolicyView:
...
+    comment: Headline for a list of suggestions for troubleshooting the current connection.
+  # Bullets for slow speeds
+  slowBulletOne: Switching server locations
```
flodolo, line 379: Please add comments explaining how these strings are used. This is a bullet item in the troubleshooting section, the rests are bullet items for low and high speed test results.

flodolo (changes requested):
Why are there punctuations changes compared to Figma?

translations/strings.yaml
flozia: I thought it might make sense to add the ellipsis where the string is used so it’s not replaced by three dots by mistake? Happy to add it to the string.

translations/strings.yaml
flozia: Added the apostrophes. I was planning to add the colon when using the string. If it adds necessary context for translations I’m happy to add it.

translations/strings.yaml
flozia: I was planning to add the colon when using the string. If it adds necessary context for translations I’m happy to add it.

translations/strings.yaml
flozia: Oh — I guess there is the possibility that different locales use a different typographic sign? I’ll just add it here.

flodolo: [Never hard-code](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#dont-hardcode-characters) punctuation signs. That's up to locales to choose what to use and how.

translations/strings.yaml
flodolo: As above, never hard-code punctuation sign. For example, French needs a non breaking space before it.

translations/strings.yaml
flodolo: Example: [Japanese](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/ja/mozillavpn.xliff#L140).

flodolo:
Not exactly sure why, but the YAML parser doesn't seem to be happy (cc @bakulf)

> yaml.scanner.ScannerError: mapping values are not allowed here
>  in "/home/runner/work/mozilla-vpn-client/mozilla-vpn-client/translations/strings.yaml", line 372, column 75

It doesn't like the closing colon?

flodolo:
> It doesn't like the closing colon?

I think you need to include the value in quotes 🤯 

```
  listHeaderDefault:
    value: "At your current speed, here’s what your device is optimized for:"
    comment: Headline for a list of things that the current connection is capable of.
```

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/5f2dbfc0c70eb9fcdf54633ee316be68e59be0ac/translations/strings.yaml#L296

lesleyjanenorton (approved):

PR #2757, opened 2022-02-15
Tutorial engine

bakulf:
# Introduction
This is the engine of the tutorial system for the VPN client. The code works, but we need to add the final UI (not in this PR).

# What this PR contains
- unit-tests for the tutorial parser
- a tool to generate tutorial files (tools/tutorialgen)
- a tool to test tutorial files (tools/tutorialplayer)
- the parser of the tutorial files (src/tutorial.cpp)
- the QML object to start/stop tutorials (src/tutorial.cpp)
- the inspector hooks
- an example about how to show tooltip QML elements

# What this PR does not contain
- the final UI of the tooltips
- the guide view
- the tutorial

# How to use it
A few steps for the use of this code.

## Step 1: record a tutorial
The easiest solution is to use the `tutorialgen.js` app. Enter in `tools/tutorialgen` folder and run `npm install` to install the dependencies. Then, start the recording of a tutorial running the command: `node ./tutorialgen.js <filename>`.
The tutorial system works only if it starts from the main VPN view (after the authentication).

Use `help` to see the available commands. The main command is `tooltip` which wants 1 argument: a string id. Pick one from the l18nstring.yaml file. For instance, run the following steps:

1. `tooltip SettingsTunnelPort53`, and pick an item (the settings menu button).
2. Open the system view.
3. Run `tooltip ServerUnavailableModalBodyText` and pick the system preference menu item.
4. Open the system preferences
5. Continue like this until you have complete your tutorial.
6. Run `done`.

Your file is now ready to be used.

## Step 2: check the file
Your tutorial file should be a sequence of `TOOLTIP` commands. The first parameter is the string ID, the second one is the name of the QML object.

## Step 3: test the tutorial
The easiest way to test a tutorial is to use `tools/tutorialplayer`. Enter in this folder, run `npm install` for the dependencies. Then, run: `node tutorialplayer.js <filename>`. You should see the client running, step by step your tutorial.

## Step 4: the integration
This part is not done in this PR. It requires UI/UX work, but all the technical pieces should be available from this piece of code. See how the tutorials as triggered in `main.qml`.

### DIFF
--- a/src/inspector/inspectorutils.cpp
+++ b/src/inspector/inspectorutils.cpp
@@ -1,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "inspectorutils.h"
+#include "qmlengineholder.h"
+
+#include <QQuickItem>
+
+// static
+QObject* InspectorUtils::findObject(const QString& name) {
+  QStringList parts = name.split("/");
+  Q_ASSERT(!parts.isEmpty());
+
+  QQuickItem* parent = nullptr;
+  QQmlApplicationEngine* engine = QmlEngineHolder::instance()->engine();
+  for (QObject* rootObject : engine->rootObjects()) {
+    if (!rootObject) {
+      continue;
+    }
+
+    parent = rootObject->findChild<QQuickItem*>(parts[0]);
+    if (parent) {
+      break;
+    }
+  }
+
+  if (!parent) {
+    if (parts.length() == 1) {
+      int id = qmlTypeId("Mozilla.VPN", 1, 0, qPrintable(parts[0]));
+      return engine->singletonInstance<QObject*>(id);
+    }
+    return parent;
+  }
+
+  for (int i = 1; i < parts.length(); ++i) {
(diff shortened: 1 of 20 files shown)


src/inspector/inspectorutils.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "inspectorutils.h"
+#include "qmlengineholder.h"
+
+#include <QQuickItem>
+
+// static
+QObject* InspectorUtils::findObject(const QString& name) {
```
bakulf, line 11: This is the function from the inspectorwebsocketconnection, moved in a separate file

lesleyjanenorton (approved):
🤩

bakulf:
I would like to land this code when the frontend part is ready. No rush.
In particular, i wonder if the positioning is doable with the current `QRect`.

bakulf:
Lesley, when you have time, (for 2.9), do you mind to try this PR and check the tooltip system?

oskirby (approved):
Looks good. As a food-for-thought kind of thing, we might want to capture the escape key as a mechanism to cancel the tutorial in the event that the user started it by accident and wants a quick out without finishing it.

bakulf:
We do have the button to quit the tutorial at any time.

Issue #2776, opened 2022-02-16
Investigate if currentDevice() in Controller::activate might be realistically nullable.


strseb:
-> vpn->deviceModel()->currentDevice(vpn->keys()); is nullable. 

-> Q: What would happen if TaskAccount new devicemodel does not include the current used keys?
-> How would we get there. I think it's the crash cause for VPN-1747
maybe we should check it before deref. 

``` i.e 
  if (device == nullptr) { resetAPP()});
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1759)

bakulf:
I managed to reproduce it. Yes, this is totally possible. We need to add an intermediate TaskFunction when re-generating the device key to check if all is OK. if not, reset() is a good option. We must check that we do not have other tasks after the reset.

PR #2786, opened 2022-02-16
Handle currentDevice == nullptr on activation

strseb:
Closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2776

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -292,7 +292,8 @@ void Controller::activateNext() {
   MozillaVPN* vpn = MozillaVPN::instance();
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
   if (device == nullptr) {
-    vpn->errorHandle(ErrorHandler::NoCurrentDeviceError);
+    vpn->errorHandle(ErrorHandler::AuthenticationError);
+    vpn->reset(false);
     return;
   }
   const HopConnection& hop = m_activationQueue.first();
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -291,6 +291,10 @@ void Controller::activateInternal() {
 void Controller::activateNext() {
   MozillaVPN* vpn = MozillaVPN::instance();
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
+  if (device == nullptr) {
+    vpn->errorHandle(ErrorHandler::NoCurrentDeviceError);
+    return;
+  }
   const HopConnection& hop = m_activationQueue.first();
 
   logger.debug() << "Activating peer" << hop.m_server.publicKey();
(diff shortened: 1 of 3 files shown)


src/controller.cpp
```diff
@@ -291,6 +291,10 @@ void Controller::activateInternal() {
 void Controller::activateNext() {
   MozillaVPN* vpn = MozillaVPN::instance();
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
+  if (device == nullptr) {
+    vpn->errorHandle(ErrorHandler::NoCurrentDeviceError);
```
bakulf, line 295: Instead of adding a new error here, what about adding:
```
TaskScheduler::scheduleTask(new TaskFunction([this]() {
    if (!modelsInitialized()) {
      logger.error() << "Failed to complete the key regeneration";
      errorHandle(ErrorHandler::RemoteServiceError);
      setUserState(UserNotAuthenticated);
      return;
    }
});
```

here: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/mozillavpn.cpp#L1680

src/mozillavpn.cpp
```diff
@@ -981,6 +981,14 @@ void MozillaVPN::errorHandle(ErrorHandler::ErrorType error) {
     case ErrorHandler::UnrecoverableError:
       alert = UnrecoverableErrorAlert;
       break;
+    case ErrorHandler::NoCurrentDeviceError:
+      // We unexpectedly have no Device to use
+      // for the connection, something has gone wrong.
+      // So let's logout the user so we can recover.
+      TaskScheduler::deleteTasks();
+      reset(false);
+      alert = UnrecoverableErrorAlert;
```
bakulf, line 990: Actually, we can recover from this. A new authentication!

bakulf (changes requested):

src/mozillavpn.cpp
bakulf: Actually, we can recover from this. A new authentication!

src/controller.cpp
strseb: We can do that! But i still think we should check that it's not a `nullptr` aynway.

src/controller.cpp
```diff
@@ -291,6 +291,11 @@ void Controller::activateInternal() {
 void Controller::activateNext() {
   MozillaVPN* vpn = MozillaVPN::instance();
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
+  if (device == nullptr) {
```
bakulf, line 294: `MOZ_ASSERT(device);`
But this is OK as well.

bakulf (approved):

PR #2828, opened 2022-02-21
Create ios-build.yaml

mbirghan:


.github/workflows/ios-build.yaml
```diff
@@ -0,0 +1,256 @@
+name: MacOS
```
bakulf, line 1: 1. Can you rename it to avoid conflicts with the current MacOS github action?
2. headers :)

PR #2837, opened 2022-02-22
✨Quicksettings Integration for android 

strseb:
Small thing i wrote while waiting for taskcluster jobs 😅
Adds a tile-service that can interact with the vpn-Service.
The tile service is registered with the OS, thus adding a quick-settings option to control "on/off" on the VPN. 

https://user-images.githubusercontent.com/9611612/155134283-5b0507e0-047b-4602-bb94-b802d0357405.mp4

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
@@ -15,7 +15,7 @@ class VPNServiceBinder(service: VPNService) : Binder() {
 
     private val mService = service
     private val tag = "VPNServiceBinder"
-    private var mListener: IBinder? = null
+    private var mListener: IBinder? = null // TODO: This now might be more then one!
     private var mResumeConfig: JSONObject? = null
 
     /**
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
@@ -213,14 +213,15 @@ class VPNServiceBinder(service: VPNService) : Binder() {
             try {
                 it.transact(code, data, Parcel.obtain(), 0)
             } catch (e: DeadObjectException) {
+                // The binder is not alive, so we can remove it
+                // from the listeners list, if present.
+                mListeners.remove(it)
             }
             return
         }
         val deadBinders = ArrayList<IBinder>()
         mListeners.forEach {
             if (it.isBinderAlive) {
-                val data = Parcel.obtain()
-                data.writeByteArray(payload?.toByteArray(charset("UTF-8")))
                 try {
                     it.transact(code, data, Parcel.obtain(), 0)
                 } catch (e: DeadObjectException) {
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
@@ -120,6 +120,7 @@ class VPNServiceBinder(service: VPNService) : Binder() {
                 val obj = JSONObject()
                 obj.put("connected", mService.isUp)
                 obj.put("time", mService.connectionTime)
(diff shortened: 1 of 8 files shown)


strseb:
@albionx - I just found this PR - it bit rotted a bit already. 
In general what do you think, should we close or try to have it ? :)

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNTileService.kt
```diff
@@ -0,0 +1,137 @@
...
+        super.onClick()
+    }
+
+    private fun updateTile() {
+        qsTile.apply {
+            when (mState) {
+                State.Error -> {
+                    state = Tile.STATE_UNAVAILABLE
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                        stateDescription = "Error CommuniCating with service"
```
brizental, line 106:
```suggestion
                        stateDescription = "Error communicating with service"
```

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNTileService.kt
```diff
@@ -0,0 +1,137 @@
...
+                    state = Tile.STATE_INACTIVE
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                        subtitle = "Disconnected"
+                    }
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                        stateDescription = "Not Connected with VPN"
+                    }
+                }
+            }
+        }
```
brizental, line 134: That is some annoying version checking... Does it error without it? Android should just ignore the unknown stuff...

brizental (changes requested):
This is a great new feature, in my opinion. :100: 

iOS has a similar thing, do we have a ticket to implement it there? If not, would you mind filing a follow up?

The code looks good to me, I left small nits. The main issue I'd say is the addressing of error cases. Maybe something we can bring UX on to look at and address as a follow up too.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNTileService.kt
strseb: Removed some redudancy :) 
Not checking not an option as ```stateDescription = "Not Connected with VPN"``` will be translated into ```qsTile.setDescription(<>)``` and calling nonexistent functions is nothing the jvm likes >:c

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
```diff
@@ -88,6 +90,16 @@ class VPNServiceBinder(service: VPNService) : Binder() {
...
+            ACTIONS.reactivate -> {
+                // [data] is empty
+                // Activate the tunnel with the last config
+                try {
+                    this.mService.reconnect()
+                } catch (e: Exception) {
+                    Log.e(tag, "An Error occurred while enabling the VPN: ${e.localizedMessage}")
+                }
+                return true
+            }
```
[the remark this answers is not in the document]
strseb: Good callout! I've added the option for the tile-server to accept activation errors, so it should move into a "deactivated & disabled " state.

strseb:
> iOS has a similar thing, do we have a ticket to implement it there? If not, would you mind filing a follow up? 

We have had an issue about that, but i think the verdict was that there is no option for iOS to hook us into that surface.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNTileService.kt
brizental: Boo JVM!

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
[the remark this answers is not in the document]
brizental: Would it be better to actually move into an error state so the user knows why it's not turning on even as the yclick to do so? No need to do that on this PR if it makes sense.

brizental (approved):

PR #2839, opened 2022-02-22
Bridge in rust

bakulf:
# Intro
This is a new implementation of mozillavpnnp in rust.

# Why
Running the unit-tests for the mozillavpnnp component on windows I found several bugs:
1. the STDIN reading/writing was totally wrong. We were using stdin to synchronize the socket reading; some messages were not received; etc..
2. the TCP handling was wrong too! The connection was sync (blocking all the rest); big windows timeouts

I tried to cleanup the mess and the code was a collection of `#ifdef MVPN_WINDOWS...`. Better to rewrite it.

# Next step
1. a proper review by rust experts
2. integration with the compilation scripts

### DIFF
--- a/extension/bridge/Cargo.toml
+++ b/extension/bridge/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "mozillavpnnp"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+mio = { version = "0.8.0", features = ["os-poll", "net"] }
+chrome_native_messaging = "0.2.0"
+serde_json = "1.0.79"
--- a/extension/bridge/src/main.rs
+++ b/extension/bridge/src/main.rs
@@ -1,0 +1,125 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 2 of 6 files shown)


extension/bridge/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+                                    }
+                                }
+                            }
+                            _ => unreachable!(),
+                        }
+                    }
+                }
+            }
+            _ => {
+                eprintln!("No connection to the VPN client");
```
oskirby, line 121: Wouldn't this get kind of spammy every time the user loads their browser without the VPN client running?

extension/bridge/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "mozillavpnnp"
+version = "0.1.0"
```
oskirby, line 3: This is probably fine for testing, but we should find some way for this to track the version of the main VPN client app.

oskirby (commented):
Very cool. I think this is a great improvement. However, we need to update the README and build instructions to mention that at Rust toolchain is now required when building the web extension.

Likewise, the Linux packaging needs updates too to mention Rust as a Build-Depends.

bakulf:
This is an intermediate step. I want to enable platforms one at a time in separate PR. Would it work for you?

oskirby (approved):

PR #2841, opened 2022-02-22
Add benchmarks for network connection

flozia:
**Notes**
We’ll need to make a few improvements as a follow up:
- #2977
- #2978
- [Add benchmark for upload (VPN-1892)](https://mozilla-hub.atlassian.net/browse/VPN-1892)

**TODO**
- [x] Add benchmark for download speed
- [x] Add benchmark for ping latency
- [x] Use localized strings
- [x] Increase benchmarking duration
- [x] Write documentation
- [x] Use self-hosted file for download benchmark
- [x] Add error view 

**Screenshots**
|VPN off [1]|VPN on [2]|Benchmarks running [3]|Benchmarks ready [4]|Benchmark error [5]|
|---|---|---|---|---|
|vpn-off|vpn-on|vpn-benchmarking|vpn-benchmarked|vpn-connection-error|


src/connectionbenchmark/connectionbenchmarkdownload.cpp
```diff
@@ -0,0 +1,130 @@
...
+      downloader->bytesReceived() / downloadDuration * 1000;
+
+  m_bytesPerSecond += bytesPerSecond;
+  m_numOfFilesReceived += 1;
+
+  if (m_numOfFilesReceived == m_numOfFilesTotal) {
+    logger.debug() << "Download speed" << m_bytesPerSecond;
+    m_fileDownloaderList.clear();
+
+    setState(StateReady);
```
bakulf, line 85: In this way we need to wait for the whole download before showing a result.
What bout if we do things differently:

1. Each FileDownloader is in charge of fetching a single file (already like this!)
2. Downloader queries each file downloader once per second to fetch the number of bytes received since the last request. (via a QTimer).
3. We update the UI immediately with the data we have.

src/connectionbenchmark/connectionbenchmarkdownload.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+void ConnectionBenchmarkDownload::start() {
+  logger.debug() << "Start benchmark";
+
+  ConnectionBenchmarkDownload::populateUrlList();
+
+  // Stop speedtest when max runtime is reached
+  m_timer = new QTimer();
```
bakulf, line 95: This will be leaked! Make it as part of the class: `QTimer m_timer;`

src/connectionbenchmark/connectionbenchmarkitem.h
```diff
@@ -0,0 +1,39 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef CONNECTIONBENCHMARKITEM_H
+#define CONNECTIONBENCHMARKITEM_H
+
+#include "qmlengineholder.h"
+
+#include <QObject>
+
+class ConnectionBenchmarkItem : public QObject {
```
bakulf, line 12: I think we don't need this class. What about if we have a model able to expose just properties such as:
- ping
- download
- update (in future)

In this way, we do not need to have a list, but a simple QObject.

src/connectionbenchmark/connectionbenchmarkmodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+  m_benchmarks.append(benchmarkItem);
+  endInsertRows();
+}
+
+void ConnectionBenchmarkModel::runNextBenchmark() {
+  if (m_state == StatePing) {
+    logger.debug() << "Run ping benchmark";
+
+    // TODO: Add localized string
+    addResult(new ConnectionBenchmarkItem(
```
bakulf, line 71: Set a parent to this object.

src/connectionbenchmark/connectionbenchmarkmodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+        "ping", "Ping", "qrc:/nebula/resources/connection-green.svg", 15));
```
bakulf, line 72: Let's keep this into the front-end code. No nebula, strings, etc.. here.

src/connectionbenchmark/connectionbenchmarkmodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+
```
bakulf, line 73: We do have ping time values.

src/connectionbenchmark/connectionbenchmarkmodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+    setState(StateDownload);
+    runNextBenchmark();
+
+  } else if (m_state == StateDownload) {
+    logger.debug() << "Run download benchmark";
```
bakulf, line 78: I like that we have states because we check items one by one. We should expose that download value is not ready yet.

src/connectionbenchmark/filedownloader.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "filedownloader.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QUrl>
+
+FileDownloader::FileDownloader(QUrl fileUrl, QObject *parent) : QObject(parent) {
```
bakulf, line 11: We should call it ResourceDownloader

src/connectionbenchmark/filedownloader.cpp
```diff
@@ -0,0 +1,45 @@
...
+  MVPN_COUNT_CTOR(FileDownloader);
+
+  QNetworkRequest request(fileUrl);
+  m_networkReply = m_networkAccessManager.get(request);
```
bakulf, line 15: wrap it in a Task and use NetworkRequest

src/connectionbenchmark/filedownloader.h
```diff
@@ -0,0 +1,44 @@
...
+#define FILEDOWNLOADER_H
+
+#include <QObject>
+#include <QByteArray>
+#include <QNetworkAccessManager>
+#include <QNetworkRequest>
+#include <QNetworkReply>
+#include <QUrl>
+
+class FileDownloader : public QObject {
```
bakulf, line 15: final

src/connectionbenchmark/filedownloader.h
```diff
@@ -0,0 +1,44 @@
...
+  Q_OBJECT
+
+  public:
+    explicit FileDownloader(QUrl fileUrl, QObject *parent = 0);
```
bakulf, line 19: no explicit here. And we always want a parent.

src/tasks/downloadresource/taskdownloadresource.cpp
```diff
@@ -0,0 +1,66 @@
...
+  emit completed(data, false);
+}
+
+void TaskDownloadResource::onFailed(QNetworkReply::NetworkError error,
+                                    const QByteArray& data) {
+  logger.debug() << "Download failed";
+
+  bool hasError = error != QNetworkReply::NoError;
+
+  emit completed(data, hasError);
```
bakulf, line 65: This is the reason why nothing works. At the end of the task, you must emit `completed` of the `Task` class.
Here you have your `completed` too and because of this, the `TaskScheduler` doesn't know that this task has finished the job.

src/tasks/downloadresource/taskdownloadresource.h
```diff
@@ -0,0 +1,41 @@
...
+
+ public:
+  TaskDownloadResource(const QUrl& fileUrl);
+  ~TaskDownloadResource();
+
+  void run() override;
+  void stop();
+
+ private slots:
+  void onProgress(qint64 bytesReceived, qint64 bytesTotal);
```
bakulf, line 28: this is a qml syntax. In the rest of the codebase we have a different name convention. This is fine too.

bakulf (changes requested):

src/connectionbenchmark/benchmarkdownloadtask.h
```diff
@@ -0,0 +1,46 @@
...
+  void taskProgressed(qint64 bytesReceived, qint64 bytesTotal);
+
+ signals:
+  void finished(quint64 bytesPerSecond, bool hasUnexpectedError);
+  void progressed(qint64 bytesReceived);
+
+ private:
+  NetworkRequest* m_request = nullptr;
+  QElapsedTimer m_elapsedTimer;
+  QUrl m_fileUrl = QUrl(
```
bakulf, line 40: This can be a `constexpr const char * DOWNLOAD_URL = "...";` in the C++ file.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  TaskScheduler::scheduleTask(m_downloadBenchmarkTask);
+
+  setState(StateRunning);
+}
+
+void ConnectionBenchmark::stop() {
+  logger.debug() << "Stop benchmark";
+
+  m_downloadBenchmarkTask->stop();
```
bakulf, line 65: This can be null if for some funny reasons the JS layer calls `stop()` before `start()`. Add an `if (state == StateRunning) { Q_ASSERT(m_downloadBenchmarkTask); ... `

bakulf: nullify the task pointer here.

src/connectionbenchmark/benchmarkdownloadtask.cpp
```diff
@@ -0,0 +1,69 @@
...
+          &BenchmarkDownloadTask::handleTaskFinished);
+
+  m_elapsedTimer.start();
+}
+
+void BenchmarkDownloadTask::stop() {
+  logger.debug() << "Stop download";
+  Q_ASSERT(m_request);
+
+  m_request->abort();
```
bakulf, line 46: 1. m_request can be null here if the task has not been scheduled yet.
2. we do not have a way to remove a task from the queue yet, so, you probably need a state: "active, inactive". if `stop()` is called before `run()` you must do something.

bakulf (changes requested):

src/connectionbenchmark/connectionbenchmarkdownload.cpp
flozia: That’s a good idea!

docs/connection-info-test.md
```diff
@@ -0,0 +1,16 @@
+# VPN Connection info test
+
```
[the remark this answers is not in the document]
flozia: Sure! I’ll write more before this PR is ready.

src/connectionbenchmark/connectionbenchmark.cpp
flozia: Added the checks here.

src/connectionbenchmark/benchmarkdownloadtask.cpp
flozia: Added `StateCancelled` to cover this case.

docs/connection-info-test.md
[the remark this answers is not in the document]
flozia: Added some documentation on what the connection test is about.

src/connectionbenchmark/benchmarkdownloadtask.h
```diff
@@ -0,0 +1,53 @@
...
+  void stop();
+
+  enum State { StateActive, StateInactive, StateCancelled };
+
+ signals:
+  void aborted();
+  void finished(quint64 bytesPerSecond, bool hasUnexpectedError);
+  void progressed(qint64 bytesReceived);
+
+ private slots:
```
bakulf, line 35: "slot" is not needed.

src/connectionbenchmark/benchmarkdownloadtask.h
```diff
@@ -0,0 +1,53 @@
...
+  void setState(State state);
+
+ private:
+  State m_state = StateInactive;
+
+  NetworkRequest* m_request = nullptr;
+  quint64 m_bytesPerSecond = 0;
+
+  QElapsedTimer m_elapsedTimer;
+  QString m_fileUrl;
```
bakulf, line 50: `const QString`

src/connectionbenchmark/benchmarkpingtask.h
```diff
@@ -0,0 +1,46 @@
...
+  void run() override;
+  void stop();
+
+  enum State { StateActive, StateInactive, StateCancelled };
+
+ signals:
+  void aborted();
+  void finished(quint16 pingLatency);
+
+ private slots:
```
bakulf, line 32: this is not needed.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  if (m_state != StateError) {
+    setState(StateInitial);
+  }
+}
+
+void ConnectionBenchmark::downloadBenchmarked(quint64 bytesPerSecond,
+                                              bool hasUnexpectedError) {
+  logger.debug() << "Benchmarked dowload" << bytesPerSecond;
```
bakulf, line 114: download

bakulf (changes requested):
Really really good! Question:
do you think it makes sense to have 1 class `BenchmarkTask` with the canceling/stopping logic in it? To avoid duplicate code in the 2 tasks. This `BenchmarkTask` inherits from `Task` directly.

lesleyjanenorton (commented):
This is looking great! Noticed one thing when running locally: It's currently possible to open the settings menu during `StateRunning` and you then get into a weird state after backing out of settings back to the main view.

nebula/ui/components/VPNConnectionInfoScreen.qml
```diff
@@ -8,7 +8,6 @@ import QtQuick.Layouts 1.14
 import Mozilla.VPN 1.0
 
 Rectangle {
-    property bool isLoading: false
```
bakulf, line 11: In theory, nebula should contain only reusable components. I cannot imagine this to be reused elsewhere.

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,119 @@
...
+    emit completed();
+  });
+
+  if (m_benchmarkType == BenchmarkType::Download) {
+    runBenchmarkDownload();
+  } else if (m_benchmarkType == BenchmarkType::Ping) {
+    runBenchmarkPing();
+  }
+
+  QTimer::singleShot(m_maxRuntime, this, &BenchmarkTask::stop);
```
bakulf, line 56: Can you use `TimerSingleShot` instead?

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  setState(StateActive);
+  m_elapsedTimer.start();
+
+  connect(this, &BenchmarkTask::finished, [this]() {
+    setState(StateInactive);
+    emit completed();
+  });
+
```
bakulf, line 49: In order to have a better separation of classes, Add this protected method: `virtual void runInternal() = 0;`. Each implementation of `BenchmarkTask` will implement it.

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,119 @@
...
+    } else if (m_benchmarkType == BenchmarkType::Ping) {
+      finishedPingBenchmark();
+    }
+
+  } else {
+    setState(StateCancelled);
+  }
+}
+
+void BenchmarkTask::runBenchmarkDownload() {
```
bakulf, line 75: this will be the `runInternal` of the download task.

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  connect(m_downloadRequest, &NetworkRequest::requestCompleted, this,
+          [&](const QByteArray& data) {
+            finishedDownloadBenchmark(QNetworkReply::NoError, data);
+          });
+  connect(m_downloadRequest, &NetworkRequest::requestFailed, this,
+          &BenchmarkTask::finishedDownloadBenchmark);
+}
+
+void BenchmarkTask::runBenchmarkPing() {
```
bakulf, line 89: this will be the `runInternal` of the ping task.

src/connectionbenchmark/benchmarktype.h
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BENCHMARKTYPE_H
+#define BENCHMARKTYPE_H
+
+enum class BenchmarkType { Ping, Download };
```
bakulf, line 8: you don't need this.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,137 @@
...
+    handleStabilityChange();
+  }
+
+  connect(controller, &Controller::stateChanged, this,
+          &ConnectionBenchmark::stop);
+  connect(m_connectionHealth, &ConnectionHealth::stabilityChanged, this,
+          &ConnectionBenchmark::handleStabilityChange);
+
+  // Create ping benchmark
+  BenchmarkTask* pingBenchmark = new BenchmarkTask(
```
bakulf, line 72: New `BenchmarkDownloadTask`

bakulf (changes requested):

src/constants.h
```diff
@@ -31,6 +31,17 @@ constexpr int RECENT_CONNECTIONS_MAX_COUNT = 5;
 // Cooldown period for unresponsive servers
 constexpr uint32_t SERVER_UNRESPONSIVE_COOLDOWN_SEC = 300;
 
+// Number of msecs for max runtime of the connection benchmarks.
+constexpr uint32_t BENCHMARK_MAX_DURATION_PING = 5000;
+constexpr uint32_t BENCHMARK_MAX_DURATION_DOWNLOAD = 15000;
+// TODO: Decide on thresholds for connection speeds
+constexpr uint32_t BENCHMARK_THRESHOLD_SPEED_FAST = 3125000;    // 25 Megabit
+constexpr uint32_t BENCHMARK_THRESHOLD_SPEED_MEDIUM = 1250000;  // 10 Megabit
+// TODO: Replace with self-hosted file
```
oskirby, line 40: If we're looking for a resource that's self-hosted, I think we can pull some static resources off of `archive.mozilla.org` which seems to be geographically well distributed thanks to Cloudlflare. This may require some testing and evaluation.

src/connectionbenchmark/benchmarktask.cpp
flozia: Done!

nebula/ui/components/VPNConnectionInfoScreen.qml
flozia: Right, I’ll move the QML components.

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+void BenchmarkTask::run() {
+  logger.debug() << "Run benchmark";
+
+  if (m_state == StateCancelled) {
+    emit completed();
+    return;
+  }
+
+  runInternal();
```
bakulf, line 37: I would move this at the end of this method.

bakulf: In this way, if runInternal checks the state, it will be `StateActive`

src/connectionbenchmark/benchmarktask.h
```diff
@@ -0,0 +1,46 @@
...
+#include <QElapsedTimer>
+#include <QObject>
+
+class BenchmarkTask : public Task {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(BenchmarkTask)
+
+ public:
+  BenchmarkTask(const uint32_t& maxExecutionTime);
+  ~BenchmarkTask();
```
bakulf, line 19: `virtual`. Non `final` classes must have a `virtual` DTOR.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+BenchmarkTaskDownload::~BenchmarkTaskDownload() {
+  MVPN_COUNT_DTOR(BenchmarkTaskDownload);
+}
+
+void BenchmarkTaskDownload::runInternal() {
+  logger.debug() << "Run download benchmark";
+
+  connect(this, &BenchmarkTask::stateChanged, this,
```
bakulf, line 30: I like this approach! Let's move it to the "extreme"! I would do this:

1. remove `runInternal` entirelly
2. move this `connect` in the CTOR of the `BenchmarkTaskDownload`

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -0,0 +1,67 @@
...
+    m_request->abort();
+    m_request = nullptr;
+  }
+}
+
+void BenchmarkTaskDownload::downloadReady(QNetworkReply::NetworkError error,
+                                          const QByteArray& data) {
+  logger.debug() << "Download ready" << error;
+
+  if (this->state() == BenchmarkTask::StateActive) {
```
bakulf, line 55: remove `this->`

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -0,0 +1,67 @@
...
+    this->stop();
```
bakulf, line 56: Why calling stop() This will trigger line 46 ad the request will be aborted.

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,37 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BENCHMARKTASKDOWNLOAD_H
+#define BENCHMARKTASKDOWNLOAD_H
+
+#include "benchmarktask.h"
+#include "networkrequest.h"
+
+#include <QByteArray>
```
bakulf, line 11: remove this

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BENCHMARKTASKDOWNLOAD_H
+#define BENCHMARKTASKDOWNLOAD_H
+
+#include "benchmarktask.h"
+#include "networkrequest.h"
```
bakulf, line 9: remove this. Use a forward declaration: `class NetworkRequest;` just before `class BenchmarkTaskDownload...`

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <QByteArray>
+#include <QNetworkReply>
+#include <QObject>
+#include <QString>
```
bakulf, line 14: remove this

src/connectionbenchmark/benchmarktask.h
```diff
@@ -0,0 +1,46 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BENCHMARKTASK_H
+#define BENCHMARKTASK_H
+
+#include "task.h"
+
+#include <QElapsedTimer>
+#include <QObject>
```
bakulf, line 11: remove this. Task.h already includes it.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  quint64 bytesPerSecond = data.size() / this->executionTime() * 1000;
```
bakulf, line 59: remove `this->`

src/connectionbenchmark/benchmarktaskdownload.cpp
bakulf: I guess you are doing it to set `m_executionTime`. If yes, you don't need it.
What about if you remove `m_executionTime` and you change `executionTime()` to return ` m_elapsedTimer.elapsed();` ?

src/connectionbenchmark/benchmarktaskping.cpp
```diff
@@ -0,0 +1,51 @@
...
+    : BenchmarkTask(Constants::BENCHMARK_MAX_DURATION_PING) {
+  MVPN_COUNT_CTOR(BenchmarkTaskPing);
+}
+
+BenchmarkTaskPing::~BenchmarkTaskPing() { MVPN_COUNT_DTOR(BenchmarkTaskPing); }
+
+void BenchmarkTaskPing::runInternal() {
+  logger.debug() << "Run ping benchmark";
+
+  connect(this, &BenchmarkTask::stateChanged, this,
```
bakulf, line 30: same comment here.

src/connectionbenchmark/benchmarktaskping.h
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BENCHMARKTASKPING_H
+#define BENCHMARKTASKPING_H
+
+#include "benchmarktask.h"
+#include "networkrequest.h"
```
bakulf, line 9: remove this

src/connectionbenchmark/benchmarktaskping.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <QObject>
```
bakulf, line 11: and this

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef BENCHMARKTASKDOWNLOAD_H
+#define BENCHMARKTASKDOWNLOAD_H
+
+#include "benchmarktask.h"
+#include "networkrequest.h"
+
+#include <QByteArray>
+#include <QNetworkReply>
+#include <QObject>
```
bakulf, line 13: and this

src/connectionbenchmark/benchmarktaskping.cpp
```diff
@@ -0,0 +1,51 @@
...
+    connect(MozillaVPN::instance()->connectionHealth(),
+            &ConnectionHealth::pingChanged, this, [&] {
+              logger.debug() << "Ping changed";
+
+              m_pingLatencyAcc +=
+                  MozillaVPN::instance()->connectionHealth()->latency();
+              m_numOfPingSamples++;
+            });
+  } else if (state == BenchmarkTask::StateInactive) {
+    quint64 m_pingLatency = (int)(m_pingLatencyAcc / m_numOfPingSamples + 0.5);
```
bakulf, line 47: 3 things:
1. why 0.5?
2. can you write it in this way: `(m_pingLatencyAcc / m_numOfPingSamples) + 0.5` - same result, just easier to read
3. `static_cast<int>(...)`

bakulf: plus... why int if you want a quin64?

src/connectionbenchmark/benchmarktaskping.h
```diff
@@ -0,0 +1,33 @@
...
+
+class BenchmarkTaskPing : public BenchmarkTask {
```
bakulf, line 13: final

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,37 @@
...
+#include <QString>
+
+class BenchmarkTaskDownload : public BenchmarkTask {
```
bakulf, line 16: final

bakulf (changes requested):

src/connectionbenchmark/benchmarktaskdownload.cpp
flozia: Yes, that was my intention — I changed it!

src/connectionbenchmark/benchmarktaskping.cpp
flozia: Casting `int` always rounds the value to the floor and this makes sure it is ”rounding” the result as I would expect it to if the decimal point is `>= 0.5`. So that for example: `1.8` → `2` but `1.3` → `1`. This might be not necessary if we follow Owens suggestion to directly take the ping value.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void ConnectionBenchmark::stop() {
+  logger.debug() << "Stop benchmark";
+
+  if (m_state == StateRunning || m_state == StateError) {
+    Q_ASSERT(!m_benchmarkTasks.isEmpty());
+
+    for (BenchmarkTask* benchmark : m_benchmarkTasks) {
+      benchmark->stop();
+    }
```
bakulf, line 99: `m_benchmarkTasks.clear();`

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+  connect(m_connectionHealth, &ConnectionHealth::stabilityChanged, this,
+          &ConnectionBenchmark::handleStabilityChange);
+
+  // Create ping benchmark
+  m_pingBenchmarkTask = new BenchmarkTaskPing();
+  connect(m_pingBenchmarkTask, &BenchmarkTaskPing::finished, this,
+          &ConnectionBenchmark::pingBenchmarked);
+  m_benchmarkTasks.append(m_pingBenchmarkTask);
+  TaskScheduler::scheduleTask(m_pingBenchmarkTask);
+
```
bakulf, line 79: You need to remove the task from `m_benchmarkTasks` when it's completed.
I would do this:

`connect(pingTask, &Task::completed, this, [this, pingTask]() { m_benchmarkTasks.remove(pingTask); });`

src/constants.h
flozia: That’s perfect — thanks!

src/connectionbenchmark/benchmarktask.h
```diff
@@ -0,0 +1,41 @@
...
+#include "task.h"
+
+#include <QElapsedTimer>
+
+class BenchmarkTask : public Task {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(BenchmarkTask)
+
+ public:
+  BenchmarkTask(uint32_t maxExecutionTime);
```
bakulf, line 17: `explicit BenchmarkTask(uint32_t maxExecutionTime);`

src/connectionbenchmark/benchmarktaskdownload.h
```diff
@@ -0,0 +1,34 @@
...
+#include <QNetworkReply>
+
+class NetworkRequest;
+
+class BenchmarkTaskDownload final : public BenchmarkTask {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(BenchmarkTaskDownload)
+
+ public:
+  BenchmarkTaskDownload(const QString& fileUrl);
```
bakulf, line 19: `explicit BenchmarkTaskDownload(const QString& fileUrl);`

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,80 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "connectionbenchmarkdownload.h"
+#include "connectionbenchmark.h"
```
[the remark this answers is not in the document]
bakulf: It's still not the first :)

src/connectionbenchmark/benchmarktaskping.cpp
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "benchmarktask.h"
+#include "benchmarktaskping.h"
```
bakulf, line 6: this goes first.

src/connectionbenchmark/benchmarktaskping.cpp
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "benchmarktask.h"
```
bakulf, line 5: remove this

src/connectionbenchmark/connectionbenchmark.h
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef CONNECTIONBENCHMARK_H
+#define CONNECTIONBENCHMARK_H
+
+#include "benchmarktaskdownload.h"
```
bakulf, line 8: remove this and the next one

src/connectionbenchmark/connectionbenchmark.h
```diff
@@ -0,0 +1,77 @@
...
+#include "benchmarktaskping.h"
+#include "connectionhealth.h"
```
bakulf, line 10: remove this and use a forward declaration.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+    Q_ASSERT(!m_benchmarkTasks.isEmpty());
+
+    for (BenchmarkTask* benchmark : m_benchmarkTasks) {
+      benchmark->stop();
+    }
+
+    m_benchmarkTasks.clear();
+  };
+
+  if (m_state != StateError) {
```
bakulf, line 108: Also in case of an error, you probably want to reset the state. What do you think?

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  m_ping = pingLatency;
+  emit pingChanged();
+}
+
+void ConnectionBenchmark::handleStabilityChange() {
+  logger.debug() << "Handle stability change";
+
+  if (m_connectionHealth->stability() == ConnectionHealth::NoSignal) {
+    setState(StateError);
```
bakulf, line 139: ah! ok! so... in case there is "no-signal". how do we recover from it?

bakulf (changes requested):

src/connectionbenchmark/benchmarktask.cpp
```diff
@@ -0,0 +1,119 @@
...
+            finishedDownloadBenchmark(QNetworkReply::NoError, data);
+          });
+  connect(m_downloadRequest, &NetworkRequest::requestFailed, this,
+          &BenchmarkTask::finishedDownloadBenchmark);
+}
+
+void BenchmarkTask::runBenchmarkPing() {
+  logger.debug() << "Run ping benchmark";
+
+  connect(MozillaVPN::instance()->connectionHealth(),
```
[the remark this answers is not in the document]
bakulf: Sounds good to me! But keep the benchmarktask and when it runs, expose this value.

src/connectionbenchmark/connectionbenchmark.cpp
flozia: If there is an error we’ll show an error screen ([Figma spec](https://www.figma.com/file/3V7aahpQ096Qsp2L3SCT6h/Connection-Screen%2FGraph-Redesign?node-id=504%3A4005)) that I’ll still need to be added.

src/connectionbenchmark/connectionbenchmark.cpp
flozia: If there is an error we’ll show an error screen ([Figma spec](https://www.figma.com/file/3V7aahpQ096Qsp2L3SCT6h/Connection-Screen%2FGraph-Redesign?node-id=504%3A4005)) that I’ll still need to be added and it would be helpful to keep the error state until we actively decide to dismiss it.

src/connectionbenchmark/connectionbenchmark.cpp
flozia: Right, there is no way to recover currently. I’ll add it!

flozia: I added a `reset` method that can be used to abort the benchmark or recover from an error from the frontend.

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -0,0 +1,149 @@
...
+    m_benchmarkTasks.clear();
+  };
+}
+
+void ConnectionBenchmark::reset() {
+  logger.debug() << "Reset benchmarks";
+
+  stop();
+
+  m_connectionHealth = nullptr;
```
bakulf, line 113: This is a singleton. Actually you can remove m_connectionHealth.

src/connectionbenchmark/connectionbenchmark.h
```diff
@@ -0,0 +1,78 @@
...
+
+ private:
+  void handleStabilityChange();
+  void setConnectionSpeed();
+  void setState(State state);
+  void stop();
+
+ private:
+  QList<BenchmarkTask*> m_benchmarkTasks;
+  ConnectionHealth* m_connectionHealth = nullptr;
```
bakulf, line 69: remove this. You can always have access to it using `MozillaVPN::instance()->connectionHealth()`

bakulf (approved):
With the latest comment applied. Well done!

src/connectionbenchmark/benchmarktask.cpp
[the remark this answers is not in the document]
flozia: @oskirby I’m now only taking a single ping, but the task is still using the `pingReceived` signal from `ConnectionHealth` instead of directly returning `MozillaVPN::instance()->connectionHealth()->latency()`. The idea is that in case the speed test is started immediately after turning the VPN on we’re taking at least another sample. Does that make sense or is this unnecessary?

https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2841/files#diff-345be1b07b39c3f4ddeaaa2e98cc6e7147ca1f668c6228a4ad2629160bfdeb8cR28-R37

docs/connection-benchmarks.md
```diff
@@ -0,0 +1,55 @@
...
+## States
+
+- `StateInitial`: Benchmarking did not start or was stopped without any errors.
+- `StateRunning`: The current connection is being benchmarked.
+- `StateReady`: All benchmarks were run successfully and the results are available.
+- `StateError`: While benchmarking an unexpected error occurred.
+
+## User flow
+
+```mermaid
```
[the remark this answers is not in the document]
flozia: Good catch! Opening the settings menu should not be possible anymore.

flozia:
> This is looking great! Noticed one thing when running locally: It's currently possible to open the settings menu during `StateRunning` and you then get into a weird state after backing out of settings back to the main view.

Good catch @lesleyjanenorton! Opening the settings menu should not be possible anymore.

lesleyjanenorton (approved):
Nice work!! This is awesome!

bakulf (approved):

PR #2793, opened 2022-02-17
New states for VPNAuthInApp

bakulf:

PR #2843, opened 2022-02-23
In-App Auth button loaders and disabled style

flozia:
This PR does two things:
1. Uses the new VPNAuthInApp states that were introduced by PR #2793 to support transitions in the In-App Auth flow through showing button loaders.
2. Fixes `VPNButton`’s disabled state and style


lesleyjanenorton (approved):
Looks great, thank you!

Issue #2846, opened 2022-02-23
macOS client requires Firefox as default browser


pronouncedJerry:
For the past week my VPN client on macOS 12.3 has not been connecting. I reached out to Support for help and I am a bit baffled by the response to my issue so I would like to confirm with the development team that this is an official stance by Mozilla:

> Could you please make sure that you login via the Firefox browser in order to set up the VPN successfully? It seems like other users encountered the same problem on MacOS and it worked after setting Firefox as their default browser when logging in.

Firefox is now required to be the default browser on macOS in order for the VPN client to work?

Since I've been unable to use this service I pay for and kindly request that this new requirement be made clear to new customers so they don't fall into the same trap I did. Otherwise, it seems like a prevalent problem (and top priority) that should be fixed, no?

oskirby:
I agree that this should be fixed. Our support team will often request trying other browsers as a troubleshooting step, but we should expect login and authentication be supported on all the major browsers. Would you be able to share the browser, and its version, that you were trying to use? If you're willing to provide logs from the VPN client after your login attempt, that would be appreciated too.

pronouncedJerry:
Thank you for the quick response oskirby.

Safari version 15.4 (build 17613.1.14.1.7)

[mozillavpn-2022-2-23_2.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/8128651/mozillavpn-2022-2-23_2.txt) hopefully I removed all of my PII

oskirby:
Alright, I have been able to reproduce the issue on Safari 15.2 (17612.3.6.1.6). It seems the issue here is that Safari insists on forcing the use of HTTPS when performing the OAuth redirect to localhost. For some technical info as to why, and some really gross workarounds, see: [here](https://stackoverflow.com/questions/46394682/safari-keeps-forcing-https-on-localhost).

pronouncedJerry:
The issue referenced on SO originated 4 years ago. The issue with the vpn client start a week ago, though.

birdsarah:
@oskirby as mentioned earlier, we use [deeplinks for some mobile cases on guardian](https://github.com/mozilla-services/guardian-website/blob/releases/1.21/server/lib/auth.ts#L60) (private repo link). This documentation from apple is making me hopeful we could implement for mac: https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app but i'm not sure if it implies delivery through the mac store.

[...]

PR #2852, opened 2022-02-24
Add MacosAuthenticationListener to support deep linking

oskirby:
Recently MacOS users with Safari as their default browser have been unable to authenticate due to Safari's HSTS policy forcing the use of HTTPS when accessing localhost. As a workaround to this, we may need to implement our own custom URL handler in order to pass the OAuth auth code from Safari back to the VPN client.

This PR extends the `DesktopAuthenticationListener` class for MacOS to include support for receiving the OAuth code via a deep link of the form `mozilla-vpn://login/success?code=XXXXX` which should provide an alternate path for Safari users to complete authentication. For this to resolve the issue completely, we will need a corresponding update to the guardian website to utilize such a deep link.

See: #2846

### DIFF
--- a/src/platforms/macos/macosauthenticationlistener.cpp
+++ b/src/platforms/macos/macosauthenticationlistener.cpp
@@ -30,9 +30,10 @@ MacosAuthenticationListener::~MacosAuthenticationListener() {
   qApp->removeEventFilter(this);
 }
 
-void MacosAuthenticationListener::start(Task* task, const QString& codeChallenge,
-                                       const QString& codeChallengeMethod,
-                                       const QString& emailAddress) {
+void MacosAuthenticationListener::start(Task* task,
+                                        const QString& codeChallenge,
+                                        const QString& codeChallengeMethod,
+                                        const QString& emailAddress) {
   logger.debug() << "MacosAuthenticationListener started";
 
   Q_UNUSED(task);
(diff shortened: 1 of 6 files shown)


src/platforms/macos/macosauthenticationlistener.cpp
```diff
@@ -0,0 +1,57 @@
...
+bool MacosAuthenticationListener::eventFilter(QObject* obj, QEvent* event) {
+  if (event->type() != QEvent::FileOpen) {
+    return QObject::eventFilter(obj, event);
+  }
+
+  QFileOpenEvent* ev = static_cast<QFileOpenEvent*>(event);
+  QUrl url = ev->url();
+  if ((url.scheme() != "mozilla-vpn") || (url.authority() != "login")) {
+    logger.warning() << "Received unknown URL:" << url.toString();
+    return false;
```
bakulf, line 42: `return QObject::eventFilter(obj, event);`

src/platforms/macos/macosauthenticationlistener.h
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef MACOSAUTHENTICATIONLISTENER_H
+#define MACOSAUTHENTICATIONLISTENER_H
+
+#include "desktopauthenticationlistener.h"
+
+#include <QEvent>
```
bakulf, line 10: class QEvent;

bakulf (approved):

PR #2854, opened 2022-02-25
Use the real platform code for in-app auth

bakulf:
When I wrote the first implementation, "android" was the only platform able to expose all the data needed to start.
Now it seems that it works correctly. Sarah can you help me to check if this is correct?


birdsarah:
I'm a bit confused by this PR; but I'm also confused about the before code.

What the code was doing before was saying:
* if we're doing login via guardian then make a URL <guardian>/api/v2/vpn/login/<platform>
* if we're doing login via in app auth, make a URL <guardian>/api/v2/vpn/login/android

now it's saying:
* if we're doing login, make a URL <guardian>/api/v2/vpn/login/<platform>

Why are we making a guardian authentication url for in app auth in the first place?

birdsarah:
@bakulf lets keep the conversation going in slack. i'm wondering if we still have some issues for the success piece of the puzzle.

birdsarah (approved):
Approving because even though we may want to change things to make things cleaner; this code unblocks testing on all platforms.

PR #2873, opened 2022-02-28
Enable macos tasckluster

mbirghan:


scripts/macosx_build.sh
```diff
@@ -1,14 +1,44 @@
 #!/bin/bash
 
+echo `pwd`
+
 # Dependencies
 python3 scripts/tooltool.py --url http://taskcluster/tooltool.mozilla-releng.net fetch -m macos/qt-dependencies.tt.manifest
 export PATH="`pwd`/qt/bin:$PATH"
-# TODO: internal pypi mirror, if it doesn't get used by default?
-pip3 install glean_parser==3.5
+export PATH="/Users/task_163336061065616/Library/Python/3.6/bin:/Users/task_163336061065616/.gem/ruby/2.6.0/bin:$PATH"
```
strseb, line 8: I don't think `task_163336061065616` id is a stable path. Once this runs on a level-3 queue (or another mac ), i would expect this to be different.
I guess you could use `${realpath ~Library/Python/3.6/bin}:$PATH`?

strseb (changes requested):

taskcluster/scripts/build/macosx_build.sh
```diff
@@ -0,0 +1,77 @@
...
+python3 scripts/tooltool.py --url http://taskcluster/tooltool.mozilla-releng.net fetch -m macos/qt-dependencies.tt.manifest
+export PATH="`pwd`/qt/bin:$PATH"
+export PATH="$HOME/Library/Python/3.6/bin:$HOME/.gem/ruby/2.6.0/bin:$PATH"
+
+# install xcodeproj which is needed by xcode_patcher.rb
+# use --user-install for permissions
+gem install xcodeproj --user-install
+
+# install rust
+curl https://sh.rustup.rs -sSf | sh -s -- -y
```
strseb, line 13: not a fan of piping of piping stuff into the terminal 😅 
Since we have homebrew, could we maybe instead do:
```brew install rust```

File: taskcluster/scripts/build/macosx_build.sh [hunk 3]
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
+# Dependencies
+python3 scripts/tooltool.py --url http://taskcluster/tooltool.mozilla-releng.net fetch -m macos/qt-dependencies.tt.manifest
```
strseb, line 4: Are you using this? It seems you pull qt from `https://github.com/mozilla-mobile/qt_static_macos` later? Or does it contain something different?

strseb (approved):

taskcluster/scripts/build/macosx_build.sh
```diff
@@ -0,0 +1,77 @@
...
+make -f MozillaVPN.xcodeproj/qt_makeqmake.mak
+make -f MozillaVPN.xcodeproj/qt_preprocess.mak
+xcodebuild build CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -project MozillaVPN.xcodeproj
+
+# Package
+python3 ./scripts/macos/import_pkg_resources.py
+
+export BUILD="/builds/worker/artifacts/"
+# Full path of the app in a temp folder
+mkdir -p $BUILD
```
strseb, line 69: Maybee we should zip the .app and /resources

Currently your task is producing 139 artifacts :D

PR #2892, opened 2022-03-01
Start the debugging service ater the other instances check

bakulf:


strseb (commented):
I don't think this will work. From the docs of "startTcpDebugServer":
"Enables debugging for QML engines created after calling this function."

Therefore i think this needs to be stay before 
```QmlEngineHolder engineHolder; ```

bakulf:
Thanks! I have a new version of the same PR.

strseb (approved):

PR #2924, opened 2022-03-02
Add Crash Test to developer options

mkestler-mozilla:
Crash button added.  Only works in staging and requires 5 clicks so shouldn't be hit accidentally.

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -1690,5 +1690,5 @@ void MozillaVPN::hardResetAndQuit() {
 void MozillaVPN::crashTest() {
   char* text = new char[100];
   delete[] text;
-  delete text;
+  delete[] text;
 }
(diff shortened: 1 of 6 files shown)


src/mozillavpn.cpp
```diff
@@ -1686,3 +1686,9 @@ void MozillaVPN::hardResetAndQuit() {
   hardReset();
   quit();
 }
+
+void MozillaVPN::crashTest() {
+  char* text = new char[100];
```
bakulf, line 1691: Put it behind #ifdef MVPN_DEBUG

bakulf (commented):
Or behind a feature. What do you think?

src/mozillavpn.cpp
mkestler-mozilla: This is for QA to test.  We don't have debug builds for them.  I'm not sure what putting it behind a feature means.

bakulf (approved):

PR #2929, opened 2022-03-03
Auth tests for windows - Fixes VPN-1802

bakulf:


scripts/tests/unit_tests.sh
```diff
@@ -17,6 +17,10 @@ if ! [ -d "src" ] || ! [ -d "tests" ]; then
   die "This script must be executed at the root of the repository."
 fi
 
+if ! [[ "$MVPN_OATHTOOL" ]]; then
```
oskirby, line 20: It would be nice if we could detect when `oathtool` is in the `PATH` so that we don't have to set environment variables when running tests manually (or this could be set in the `common.sh` scripts.

oskirby (approved):

scripts/tests/unit_tests.sh
bakulf: follow up!

PR #2934, opened 2022-03-03
Use std++17 for windows

bakulf:


mkestler-mozilla:
This is already covered by the CONFIG += c++1z in src.pro.

bakulf:
oh! good catch! It's actually only for tests!

bakulf:
PR updated.

mkestler-mozilla (approved):

PR #2939, opened 2022-03-04
RELENG-820: Create android push apk tasks

gabrielBusta:
Creates tasks to upload signed android apks to the Google Play store.

### DIFF
--- a/taskcluster/mozillavpn_taskgraph/transforms/push_apk.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/push_apk.py
@@ -1,0 +1,20 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Apply some defaults and minor modifications to the jobs defined in the build kind.
+"""
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import resolve_keyed_by
+
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def add_push_apk_worker_config(config, tasks):
+    for task in tasks:
+        yield task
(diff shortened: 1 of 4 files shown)


taskcluster/mozillavpn_taskgraph/worker_types.py
```diff
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from voluptuous import Any, Required, Optional
-
+from six import text_type
```
ahal, line 6: I'm assuming you copy/pasted this from somewhere, but there's no need to support Py2 anymore, so let's `s/text_type/str` and not depend on `six` here :)

gabrielBusta:
@ahal the vpn apk worker is up

mkestler-mozilla (approved):

taskcluster/mozillavpn_taskgraph/transforms/push_apk.py
```diff
@@ -0,0 +1,20 @@
...
+from taskgraph.util.schema import resolve_keyed_by
+
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def add_push_apk_worker_config(config, tasks):
+    for task in tasks:
+        yield task
```
ahal, line 20: This whole file is a no-op :). If we don't need to do anything in the transforms, let's remove this file until we need it.

ahal (approved):
Thanks, lgtm!

gabrielBusta:
@ahal not sure what is wrong with the [install apple certificate step of the required `MacOS / MacOS packages` check](https://github.com/mozilla-mobile/mozilla-vpn-client/runs/5603670444?check_suite_focus=true#step:8:32)

ahal:
I'm not familiar with the Github Actions side of CI, but I think these might be failing because you opened up a PR based on a fork rather than on a branch in this repo. We could maybe ask to get you write access here so you can make branches as well.. Or we could ask someone with admin permissions to override the failing checks and land anyway.

Either way I don't think changes to the `/taskcluster` directory can possibly affect the Github Actions CI.

PR #2952, opened 2022-03-08
Reduce the number of log entries in android

bakulf:


strseb (approved):

strseb:
NIT: While this one is probably only obvious whats happening for me, we should have a message for all possible cases, so the log-line contains better info :)

PR #2957, opened 2022-03-09
Glean events for the auth-in-app - fixes VPN-1882

bakulf:
This PR requires a data-review too. But first, can you tell me if you are OK with these events?

### DIFF
--- a/src/authenticationinapp/authenticationinapp.cpp
+++ b/src/authenticationinapp/authenticationinapp.cpp
@@ -8,8 +8,11 @@
 #include "leakdetector.h"
 #include "incrementaldecoder.h"
 
+#include "../../glean/telemetry/gleansample.h"
+
 #include <QCoreApplication>
 #include <QFile>
+#include <QMetaEnum>
 #include <QRegularExpression>
 
 constexpr int PASSWORD_MIN_LENGTH = 8;
@@ -43,6 +46,10 @@ AuthenticationInApp::~AuthenticationInApp() {
 void AuthenticationInApp::setState(State state) {
   m_state = state;
   emit stateChanged();
+
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+      GleanSample::authenticationInappStep,
+      {{"state", QVariant::fromValue(state).toString()}});
 }
 
 void AuthenticationInApp::registerListener(
--- a/src/authenticationinapp/authenticationinapplistener.cpp
+++ b/src/authenticationinapp/authenticationinapplistener.cpp
@@ -12,6 +12,8 @@
 #include "mozillavpn.h"
 #include "networkrequest.h"
 
+#include "../../glean/telemetry/gleansample.h"
+
 #include <QCoreApplication>
 #include <QJsonDocument>
 #include <QJsonObject>
@@ -625,7 +627,8 @@ void AuthenticationInAppListener::finalizeSignInOrUp() {
       });
 }
 
-void AuthenticationInAppListener::processErrorCode(int errorCode) {
+void AuthenticationInAppListener::processErrorCode(
+    int errorCode, const QString& errorString, const QString& errorMessage) {
   AuthenticationInApp* aip = AuthenticationInApp::instance();
   Q_ASSERT(aip);
 
@@ -826,6 +829,11 @@ void AuthenticationInAppListener::processErrorCode(int errorCode) {
     case 998:  // An internal validation check failed.
       [[fallthrough]];
     default:
+      emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+          GleanSample::authenticationInappError,
+          {{"errno", QString::number(errorCode)},
+           {"error", errorString},
+           {"message", errorMessage}});
       logger.error() << "Unsupported error code:" << errorCode;
       break;
   }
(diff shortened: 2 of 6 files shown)


src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -754,6 +756,9 @@ void AuthenticationInAppListener::processErrorCode(int errorCode) {
     case 998:  // An internal validation check failed.
       [[fallthrough]];
     default:
+      emit MozillaVPN::instance()->recordGleanEventWithStringType(
+          GleanSample::authenticationInappError, "errno",
+          QString::number(errorCode));
```
birdsarah, line 761: You've gone to all the trouble of typing out the error names. I'm wondering if it'll be easier for future analysis if you pass a string that concatenates the error code and description?

src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -775,6 +780,8 @@ void AuthenticationInAppListener::processRequestFailure(
         return;
       }
 
+      emit MozillaVPN::instance()->recordGleanEventWithStringType(
+          GleanSample::authenticationInappError, "errno", "125");
```
birdsarah, line 784: To prevent future confusion, I'd like to see this handled slightly differently because we're actually only capturing a subset of 125 errors here. This is "errno 125 + not email-captcha"

src/ui/main.qml
```diff
@@ -344,6 +344,11 @@ Window {
             Sample[sample].record();
         }
 
+        function onRecordGleanEventWithStringType(sample, type, value) {
+            console.debug("recording Glean event with type", type, "and value", value);
+            Sample[sample].record({[type]: value});
+        }
```
birdsarah, line 350: I'm wondering if we can make this more generic to handle any number of "extra_keys" e.g. 

```
        Sample.onboardingCtaClick.record({
                                              "panel_id": currentPanelValues._panelId,
                                              "panel_idx": swipeView.currentIndex.toString(),
                                              "panel_cta": ctaObjectName
                                          });
```

And this particular addition happens to just be 1. This would help with my ask for errno and description.

Also, today I learned about computed keys.

src/authenticationinapp/authenticationinapp.cpp
```diff
@@ -43,6 +46,13 @@ AuthenticationInApp::~AuthenticationInApp() {
 void AuthenticationInApp::setState(State state) {
   m_state = state;
   emit stateChanged();
+
+  const QMetaObject* meta = qt_getEnumMetaObject(state);
+  QMetaEnum me =
+      meta->enumerator(meta->indexOfEnumerator(qt_getEnumName(state)));
```
birdsarah, line 52: this feels very heavy.

making sure i understand m_state is a number?

I have a recollection of doing work previously with things like 

`QString theBig = QVariant::fromValue(ModelApple::Big).toString();`

so that our debug messages didn't print out authentication state 0, 1, 2 but something human readable.

Can we do something nicer here.

birdsarah (changes requested):
added some notes, let me know what you think

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: What about if we do not have them all?
The errno should be enough because we can fetch the description for the API documentation https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/api.md  - if a new error code is added, our code does not break.

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: I like this! we can have errno + extra desc if available.

src/authenticationinapp/authenticationinapp.cpp
birdsarah: > I agree with "nicer". I disagree with "heavy".

I meant nicer :D This code has now gone away which is even better :D

bakulf:
> This PR now adds the event "authentication_inapp_step" in the metrics.yaml but doesn't add any code to record it. Either add the event or add delete the definition.

```

      GleanSample::authenticationInappStep,
      {{"state", QVariant::fromValue(state).toString()}});
```

It's done here.

src/authenticationinapp/authenticationinapp.cpp
```diff
@@ -43,6 +46,10 @@ AuthenticationInApp::~AuthenticationInApp() {
 void AuthenticationInApp::setState(State state) {
   m_state = state;
   emit stateChanged();
+
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+      GleanSample::authenticationInappStep,
+      {{"state", QVariant::fromValue(state).toString()}});
```
birdsarah, line 52: :D

birdsarah (approved):
it's sooooo clean now I missed it :D

birdsarah:
Where is the data review? 

I approved this PR with the understanding that it said "this needs data review too" and I assumed you would update the bugs link before merging.

bakulf:
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2990

PR #2960, opened 2022-03-09
Expose iaa:true in the user-agent when authentication-in-app is on - fixes VPN-1885

bakulf:


birdsarah (approved):
Nice.

Can you take a look at my comments on https://github.com/mozilla-services/guardian-website/pull/1481. I thought we were only turning in app authentication on for mobile. But maybe I've gotten confused.

bakulf:
Yes, we turn on auth-in-app only on mobile, but this is controlled by a "Feature". In this way, if we want to test it on desktop, we can simply turn on the feature. The feature is active, by default, on mobile only: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/features/featureinappauth.h#L27-L33

PR #2961, opened 2022-03-09
RELENG-804 - Refactor signing tasks

hneiva:
Mac build/signing isn't ready yet, this will refactor the signing logic to be more generic, and make it easy to add mac signing (and others) when needed.

### DIFF
--- a/taskcluster/ci/config.yml
+++ b/taskcluster/ci/config.yml
@@ -56,5 +56,11 @@ workers:
                 by-level:
                     "3": mozillavpn-3-signing
                     default: mozillavpn-t-signing
-scriptworker:
-    scope-prefix: project:mozilla-vpn-client
+        # mac-signing:
+        #     provisioner: scriptworker-prov-v1
+        #     implementation: scriptworker-signing
+        #     os: macosx
+        #     worker-type:
+        #         by-level:
+        #             "3": mozillavpn-3-signing-mac
+        #             default: mozillavpn-t-signing-mac
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -17,7 +17,7 @@ primary-dependency: build
 group-by: build-type
 
 only-for-build-types:
-    - macosx/opt
+    # - macosx/opt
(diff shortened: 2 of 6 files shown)


taskcluster/ci/signing/kind.yml
```diff
@@ -16,20 +16,29 @@ primary-dependency: build
 
 group-by: build-type
 
+only-for-build-types:
+    - macosx/opt
```
ahal, line 20: I think this will cause the `mac` task to exist as soon as the build lands. I think it would be better to exclude this so we can explicitly enable it at some point after the build.

ahal (changes requested):

taskcluster/ci/config.yml
```diff
@@ -56,5 +56,11 @@ workers:
                 by-level:
                     "3": mozillavpn-3-signing
                     default: mozillavpn-t-signing
-scriptworker:
```
gabrielBusta, line 59: Did u mean to delete these two lines at the end of the file?

hneiva: Yeah, they are not used anywhere. Probably a copy/paste from somewhere.

taskcluster/ci/signing/kind.yml
```diff
@@ -16,20 +16,29 @@ primary-dependency: build
...
+    - android-x86/release
+    - android-arm64/release
+    - android-armv7/release
+
 job-template:
-    description: Sign MozillaVPN for Android
+    description: Sign MozillaVPN
     worker: {} # Will be configured by signing:transforms
-    signing-format: autograph_apk
-    run-on-tasks-for: [github-push]
```
ahal, line 23: Actually, I think this needs to be an empty list because otherwise it will default to "all":
https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/target_tasks.py#l44

ahal (changes requested):

ahal (approved):
Lgtm!

gabrielBusta (approved):

strseb (approved):

PR #2971, opened 2022-03-10
DNS ping fallback when ICMP sockets are unavailable

oskirby:
For some Linux and Android platforms, we have encountered cases where ICMP sockets are not available without root permissions, and this can lead to a scenario when we fall back to the dummy ping sender and lose the ability to test the network connectivity. We now have a better workaround in in the form of the `DnsPingSender` which uses DNS queries, and shouldn't require any elevated permissions.

To test this on Linux, we can deny ICMP socket creation with: `echo 0 0 > /proc/sys/net/ipv4/ping_group_range`
After this change, we can observe in the logs that the client still produces realistic ping measurements.

### DIFF
--- a/src/pinghelper.cpp
+++ b/src/pinghelper.cpp
@@ -7,19 +7,8 @@
 #include "leakdetector.h"
 #include "logger.h"
 #include "pingsender.h"
+#include "pingsenderfactory.h"
 #include "timersingleshot.h"
-
-#if defined(MVPN_LINUX) || defined(MVPN_ANDROID)
-#  include "platforms/linux/linuxpingsender.h"
-#elif defined(MVPN_MACOS) || defined(MVPN_IOS)
-#  include "platforms/macos/macospingsender.h"
-#elif defined(MVPN_WINDOWS)
-#  include "platforms/windows/windowspingsender.h"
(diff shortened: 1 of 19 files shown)


src/pinghelper.cpp
```diff
@@ -53,23 +53,28 @@ void PingHelper::start(const QString& serverIpv4Gateway,
...
+      new LinuxPingSender(m_source, this);
 #elif defined(MVPN_MACOS) || defined(MVPN_IOS)
-        new MacOSPingSender(m_source, this);
+      new MacOSPingSender(m_source, this);
 #elif defined(MVPN_WINDOWS)
-        new WindowsPingSender(m_source, this);
+      new WindowsPingSender(m_source, this);
 #else
-        new DummyPingSender(m_source, this);
+      new DummyPingSender(m_source, this);
```
mkestler-mozilla, line 64: Would you mind pulling these out in to a factory class?  I'm trying to clean up some of the uglier ifdefs for object creation.  You can look at ImageProviderFactory for what I did.
I'll approve anyway, but if you have time that would be great.

mkestler-mozilla (approved):

src/pinghelper.cpp
oskirby: Will do!

PR #2984, opened 2022-03-12
Optimize the order of the password validation

bakulf:


lesleyjanenorton (approved):

bakulf:
I also added a bit of cache for the common-password file content.

Issue #2977, opened 2022-03-11
Discard data from connection download benchmark


flozia:
Instead of using `NetworkRequest` which is storing the downloaded data in memory we should replace it with `QSslSocket` or find another way to not store the downloaded data in memory.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1896)

PR #2988, opened 2022-03-12
Discard data on download progress

flozia:
The `BenchmarkTaskDownload` requires us to download data and the current implementation is storing the data in memory. Since we do not care about the data itself but only the size a solution could be to use `QSslSocket` instead of `NetworkRequest` (issue #2977).

This an experiment to see if we could keep using `NetworkRequest` with a few adjustments to the current implementation.


src/networkrequest.cpp
```diff
@@ -959,7 +964,13 @@ void NetworkRequest::handleReply(QNetworkReply* reply) {
   connect(m_reply, &QNetworkReply::finished, this,
           &NetworkRequest::maybeDeleteLater);
   connect(m_reply, &QNetworkReply::downloadProgress, this,
-          &NetworkRequest::requestUpdated);
+          [&](qint64 bytesReceived, qint64 bytesTotal) {
+            if (m_discardData) {
+              m_reply->readAll();
```
flozia, line 969: If I’m understanding correctly this would prevent us from storing downloaded data in memory. Is this a feasible approach?

flozia, line 969: If I’m understanding correctly this would prevent us from storing downloaded data in memory. Is this a feasible approach?

bakulf (changes requested):
I like this solution but i think we can simplify it a lot. What about if we introduce a new signal as you did, but passing the network relay object together with the bytes received and total. The slot, in the benchmark class, can call readAll().

Would it work?

bakulf:
Or even better you can use `sender()->replay()`

flozia:
> I like this solution but i think we can simplify it a lot. What about if we introduce a new signal as you did, but passing the network relay object together with the bytes received and total. The slot, in the benchmark class, can call readAll().
> 
> Would it work?

That works and is a lot cleaner — yes! I wasn’t sure how to get it working with `sender()->replay()` though.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -44,11 +46,26 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
 }
 
+void BenchmarkTaskDownload::downloadProgressed(qint64 bytesReceived,
+                                               qint64 bytesTotal,
+                                               QNetworkReply* reply) {
+  logger.debug() << "Handle progressed:" << bytesReceived << "(received)"
+                 << bytesTotal << "(total)";
+
+  if (bytesReceived != m_bytesReceived) {
+    m_bytesReceived = bytesReceived;
```
bakulf, line 56: Don't you have to sum these 2 values?

bakulf (changes requested):

bakulf (approved):

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -44,11 +46,24 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
+void BenchmarkTaskDownload::downloadProgressed(qint64 bytesReceived,
+                                               qint64 bytesTotal,
+                                               QNetworkReply* reply) {
+  logger.debug() << "Handle progressed:" << bytesReceived << "(received)"
+                 << bytesTotal << "(total)";
+
+  m_bytesReceived = bytesReceived;
+
+  // discard downloaded data
+  reply->readAll();
```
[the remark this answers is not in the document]
flozia: Done.

PR #2992, opened 2022-03-13
Fix red cursor in VPNTextField on Android

### LINKED ISSUE #2901: Red and misaligned cursor in text fields
**VPN version:**

* 2.7.1 (16455930);

**Platforms:**

* Android 10 - Google Pixel 4
* Android 10 - Huawei P30 - ELE-29;

**Prerequisites:**

* Have no FxA account with active subscription for Mozilla VPN;

**Steps to reproduce:**

* Open the VPN app;
* Tap “Sign up”;
* Tap in the email address fields;
* Observe the behavior;

**Expected results:**

* There are no display or layout issues. Cursor is black and aligned.

**Actual results**

* The cursor is red and not centered with the default text.

**Notes:**

* Reproducing in all text fields on Android 10.
* Reproducing partially on MacOS QT6 builds also.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1817)

lesleyjanenorton:
This fixes the red cursor on Android and the rest of #2901 (misaligned text was fixed previously in #2973). This is ready for code review but needs to be run by UX before merging.

Screen Shot 2022-03-10 at 6 23 23 PM

Fixes #2901


nebula/ui/components/forms/VPNTextField.qml
```diff
@@ -36,6 +36,26 @@ TextField {
...
 
+    cursorDelegate: Rectangle {
+        // force cursor color and blink
+        id: cursor
+        visible: parent.activeFocus && !parent.readOnly && parent.selectionStart === parent.selectionEnd
+        color: textField.placeholderTextColor
+        width: 1
+
+        Timer {
+            // This is duped from CursorDelegate.qml to preserve cursor blinking which is overwritten
```
bakulf, line 47: Why do we need this? Is it a bug in qt?

lesleyjanenorton: Yes, the cursor blink disappears entirely after customizing cursorDelegate.

bakulf: do you know if there is already a qt-bug? If no, let's file one. Then we can write the URL of the bug here.

bakulf (approved):

nebula/ui/components/forms/VPNTextField.qml
bakulf: We can land this and we can write the bug ID as a follow-up. What do you think?

PR #2999, opened 2022-03-14
Remove qt5 from xcode cloud

mbirghan:
This PR changes the Xcode Cloud workflow to use Qt6 for both iOS and MacOS.

For iOS we switch from using `qt_static_ios` to using `aqt` to install qt6. 
Note that since qt6 for iOS references the macos packages witch we have to install first. 
Also, since the installation on iOS using `aqt` references `clang_64` instead of `macos` we rename the macos installation.

### DIFF
--- a/ci_scripts/ci_post_clone.sh
+++ b/ci_scripts/ci_post_clone.sh
@@ -25,18 +25,20 @@ then
   export PATH=`pwd`/qt_static_macos/qt6/bin:$PATH
 else
   # generate qt_ios
-  git clone https://github.com/mozilla-mobile/qt_ios
-  cd qt_ios
-  cat qt6* > qt_static.tar.gz
-  tar xf qt_static.tar.gz
-  sudo mkdir /opt/6.2.3
-  sudo mv ios /opt/6.2.3
-  sudo mv macos /opt/6.2.3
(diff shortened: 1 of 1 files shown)


bakulf (approved):

ci_scripts/ci_post_clone.sh
```diff
@@ -12,32 +12,29 @@ cd /Volumes/workspace/repository
...
+  cat qt6* > qt_static.tar.gz
   tar xf qt_static.tar.gz
   cd ..
-  export QT_MACOS_BIN=`pwd`/qt_static_macos/qt/bin
-  export PATH=`pwd`/qt_static_macos/qt/bin:$PATH
+  export QT_MACOS_BIN=`pwd`/qt_static_macos/qt6/bin
+  export PATH=`pwd`/qt_static_macos/qt6/bin:$PATH
 else
-  # generate qt_ios
-  git clone https://github.com/mozilla-mobile/qt_ios
```
bakulf, line 28: Can we remove the qt_ios repo?

mbirghan: Currently we are still using it for the github ci but I can work on that as well.

PR #3006, opened 2022-03-14
Update strings.yaml

mkestler-mozilla:
Adding strings for the opt-in option of crash reporting.

### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -317,7 +317,9 @@ settings:
     value: Tunnel VPN through port 53/DNS
     comment: Tunnel is used as a verb, as in to direct traffic through
   tunnelPort53Description: Use port 53/DNS for Wireguard traffic, which may help evade firewalls
-
+  crashOptInDescription:
+    value: Allow Mozilla VPN to automatically send crash reports to help fix unexpected crashes
+    comment: This is the description of a checkbox option to allow the app to send crash reports without a user prompt
 
 aboutUs:
   licenses: Licenses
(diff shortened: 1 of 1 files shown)


translations/strings.yaml
```diff
@@ -422,6 +424,9 @@ crashreporter:
   dontSendButton:
     value: No thanks
     comment: User does not agree to send a crash report.
+  crashOptInDialogDescription:
+    value:  Don't show this again.  Auto-send crash reports to Mozilla.
```
lesleyjanenorton, line 428: *Don’t (apostrophe)

translations/strings.yaml
```diff
@@ -422,6 +424,9 @@ crashreporter:
...
+    comment:  This is a descrption for a checked option on the crash reporter dialog.
```
lesleyjanenorton, line 429: *description

translations/strings.yaml
flodolo, line 428:
```suggestion
    value:  Don't show this again. Auto-send crash reports to Mozilla.
```

Drop the double space after punctuation.

Does `Auto-send` come from our Content Team? I've never seen it before, it's normally spelled out.

flodolo (changes requested):

translations/strings.yaml
flodolo, line 429:
```suggestion
    comment:  This is a description for a checkbox on the crash reporter dialog.
```

translations/strings.yaml
mkestler-mozilla: Auto-send was in the figma, but I've changed it.  I think Autosend as one word is the correct usage.

translations/strings.yaml
flodolo: This is still not fixed.

mkestler-mozilla: Forgot to push.  It is fixed now.

flodolo (approved):

PR #3007, opened 2022-03-14
Audit log messages to hide potentially sensitive information.

oskirby:
This should remove the following information from the logs:
 - User's public keys, e-mail address and account details.
 - Any identifying details of the chosen server.
 - Wi-Fi SSID/BSSID.
 - Captive portal redirection URL.

### DIFF
--- a/src/authenticationinapp/authenticationinapplistener.cpp
+++ b/src/authenticationinapp/authenticationinapplistener.cpp
@@ -149,7 +149,7 @@ void AuthenticationInAppListener::checkAccount(const QString& emailAddress) {
 
   connect(request, &NetworkRequest::requestCompleted, this,
           [this](const QByteArray& data) {
-            logger.debug() << "Account status checked" << data;
+            logger.debug() << "Account status checked";
 
             QJsonDocument json = QJsonDocument::fromJson(data);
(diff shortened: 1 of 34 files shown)


src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -149,7 +149,7 @@ void AuthenticationInAppListener::checkAccount(const QString& emailAddress) {
 
   connect(request, &NetworkRequest::requestCompleted, this,
           [this](const QByteArray& data) {
-            logger.debug() << "Account status checked" << data;
+            logger.debug() << "Account status checked";
```
bakulf, line 152: can we use the `sensitive()` method instead? These messages are still useful for debugging

bakulf (changes requested):
All good! But keep the auth-in-app messages with sensitive() for now.

src/authenticationinapp/authenticationinapplistener.cpp
oskirby: Okay, I'll convert these to sensitive instead.

mkestler-mozilla (approved):

bakulf (approved):

PR #3008, opened 2022-03-14
Use the new way to obtain tokens and codes via iaa:true

bakulf:


src/authenticationinapp/authenticationinapplistener.cpp
```diff
@@ -236,7 +258,9 @@ void AuthenticationInAppListener::signIn(const QString& unblockCode) {
 
 void AuthenticationInAppListener::signInInternal(const QString& unblockCode) {
   NetworkRequest* request = NetworkRequest::createForFxaLogin(
-      m_task, m_emailAddressCaseFix, generateAuthPw(), unblockCode, m_urlQuery);
+      m_task, m_emailAddressCaseFix, generateAuthPw(), unblockCode,
+      m_fxaParams.m_clientId, m_fxaParams.m_deviceId, m_fxaParams.m_flowId,
+      m_fxaParams.m_flowBeginTime);
```
birdsarah, line 263: I would have thought you needed, at least, scope. I was expecting you to need all of state, scope, accessType.

Just because the request works without these three doesn't mean it actually works:
* accessType, I thought, indicates we need a long-term refreshToken
* scopes are all the scopes we need for the support form, profile checking (including subscription checking), and in-app subscritpion to work.
* state is an important part of the integrity of the oauth flow (@jswinarton will have the details of what we need when on that bit)

birdsarah (commented):
I think this looks broadly right. I haven't done the most thorough job so apologies if my comment about scopes is out of place. 

I don't think either jeremy or i have context on the different types of requests you have to make for different login scenarios as we've just handed that all off to fxa. You may want to get an fxa engineer to review this.

But in terms of passing things around. The back end looks good i.e. you use the redirect url you get from fxa and redirect to it and get a code back. And the front looks good - where you get an object from us - although i'll let jeremy approve or not as he has the details on the apis.

src/authenticationinapp/authenticationinapplistener.cpp
bakulf: All is good here. See https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/api.md#post-accountcreate
The account creation doesn't need accessType, scopes, or state. Those are required only for the final step: `POST /oauth/authorization`

bakulf:
I haven't changed the use of the FxA API. I don't think I need a review from the FxA team. This review request is mainly about:
- code syntax, style, etc.
- interaction with guardian
- the temporary fix for the unit-test (no JSON for errors)

jswinarton (approved):

src/networkrequest.cpp
```diff
@@ -550,10 +548,10 @@ NetworkRequest* NetworkRequest::createForFxaSessionVerifyByEmailCode(
 
   QJsonObject obj;
   obj.insert("code", code);
-  obj.insert("service", query.queryItemValue("client_id"));
+  obj.insert("service", fxaClientId);
 
   QJsonArray scopes;
-  QStringList queryScopes = query.queryItemValue("scope").split("+");
+  QStringList queryScopes = fxaScope.split(" ");
```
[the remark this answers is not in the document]
bakulf: `/v1/session/verify_code` is a non-documented FxA API that wants the scopes as an array. I had to revert-engineer what the web page does to see the correct params of this endpoint.
I already spoke with the FxA team about this missing documentation, but it's better to file a bug. So far, this code works.

PR #3018, opened 2022-03-15
Update gh ios ci

mbirghan:


mbirghan:
After this lands we can remove the qt_ios repo

bakulf (approved):

PR #3019, opened 2022-03-15
grcov for linux

bakulf:
I cannot include lottie in this PR because grcov crashes with the following error:
```
Error: 0 [ERROR] A panic occurred at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/grcov-0.8.7/src/parser.rs:377: called `Result::unwrap()` on an `Err` value: Error("invalid type: floating point `2507390`, expected u64", line: 1, column: 212345)
```


mozrokafor (approved):
LGTM

bakulf:
> One thing i did notice is the diff from codecov, it appears off

Fixed :)

PR #3042, opened 2022-03-16
Add include all networks

mbirghan:


src/features/featureincludeallnetworks.h
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef FEATURE_INCLUDEALLNETWORKS_H
```
bakulf, line 5: I would call it "FeatureIOSKillSwitch". `IncludeAllNetworks` is just the API that allows us to implement the kill-switch.

src/features/featureincludeallnetworks.h
```diff
@@ -0,0 +1,40 @@
...
+#define FEATURE_INCLUDEALLNETWORKS_H
+
+#include "models/feature.h"
+
+constexpr const char* FEATURE_INCLUDEALLNETWORKS = "includeAllNetworks";
+
+class FeatureIncludeAllNetworks : public Feature {
+ public:
+  FeatureIncludeAllNetworks()
+      : Feature(FEATURE_INCLUDEALLNETWORKS, "Include All Networks",
```
bakulf, line 15: Also here: "IOS Kill-Switch"

src/features/featureincludeallnetworks.h
```diff
@@ -0,0 +1,40 @@
...
+                L18nStrings::Empty,  // LongDescr
+                "",                  // ImagePath
+                "",                  // IconPath
+                "",                  // link URL
+                "2.8",               // released
+                true                // Can be enabled in devmode
+        ) {}
+
+  bool checkSupportCallback() const override {
+#if defined(MVPN_IOS)
```
bakulf, line 28: Let's include this feature only on iOS. And to enforce it, At the top of this file:

```
#ifndef MVPN_IOS
#error "This feature should be exposed only to IOS - or a better error message here...
#endif
```

And here you can simply say: `return true;`

src/platforms/ios/ioscontroller.mm
```diff
@@ -12,6 +12,7 @@
 #include "mozillavpn.h"
 #include "server.h"
 #include "settingsholder.h"
+#include "features/featureincludeallnetworks.h"
```
bakulf, line 15: alphabetic order.

src/src.pro
```diff
@@ -246,6 +246,7 @@ HEADERS += \
         features/featureinappaccountcreate.h \
         features/featureinappauth.h \
         features/featureinapppurchase.h \
+        features/featureincludeallnetworks.h \
```
bakulf, line 249: Add it only in the `ios` block.

bakulf: Plus you need to create the feature in `featurelist.cpp` but only `#ifdef MVPN_IOS`

bakulf (changes requested):

src/featurelist.cpp
```diff
@@ -18,6 +18,7 @@
 #include "features/featureinappaccountcreate.h"
 #include "features/featureinappauth.h"
 #include "features/featureinapppurchase.h"
+#include "features/featureioskillswitch.h"
```
bakulf, line 21: `#ifdef MVPN_IOS`

src/featurelist.cpp
```diff
@@ -56,6 +57,7 @@ void FeatureList::initialize() {
   new FeatureInAppAccountCreate();
   new FeatureInAppAuth();
   new FeatureInAppPurchase();
+  new FeatureIosKillswitch();
```
bakulf, line 60: here too

src/src.pro
```diff
@@ -855,6 +855,7 @@ else:ios {
             tasks/purchase/taskpurchase.h
 
     OBJECTIVE_HEADERS += \
+            features/featureioskillswitch.h \
```
bakulf, line 858: move to the "HEADERS" block. I don't think it changes too much, but it's C++ and not obj-C++

bakulf (changes requested):

src/features/featureioskillswitch.h
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef FEATURE_IOS_KILLSWITCH_H
+#define FEATURE_IOS_KILLSWITCH_H
+
```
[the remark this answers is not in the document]
mbirghan: Sorry, I did not see this before. Will add it now.

bakulf (approved):

PR #3048, opened 2022-03-16
Telemetry events for connection stability

bakulf:


src/telemetry.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "telemetry.h"
+#include "leakdetector.h"
+#include "logger.h"
+#include "mozillavpn.h"
+
+#include "../../glean/telemetry/gleansample.h"
+
+// 10 seconds of connection.
+constexpr int CONNECTION_STABILITY_MSEC = 10000;
```
oskirby, line 13: As a minor improvement, I would recommend increasing this to 45-60 seconds. The ping statistics are averaged over a 30-second window, and this will ensure that any chaos from the connection change will fall out of that window and won't tarnish our statistics.

oskirby (changes requested):

oskirby (approved):

bakulf:
I'll add these events to the data-review doc together with the ohers.

PR #3072, opened 2022-03-18
Fix mobile onboarding white space on iOS

flozia:
This PR removes the safe area for the mobile onboarding on iOS devices.


mbirghan (approved):
LGTM

PR #3089, opened 2022-03-21
Add a new string to fix the connection label - VPN-1858

bakulf:


translations/strings.yaml
```diff
@@ -347,6 +347,7 @@ genericError:
 
 # Connection Info Screen
 connectionInfo:
+  openButton: Connection Info
```
flodolo, line 350: Given the use of "info", are there specific limitations on the size of the button that localizers should be mindful of? If so, it would be worth adding it as a comment.

flodolo (approved):

PR #3090, opened 2022-03-21
Fix way to enable debug features for Glean

brizental:
Hey folks, I noticed this issue while going through the code. 

This was our bad for not communicating the change better to our users, but in [v0.27.0](https://github.com/mozilla/glean.js/releases/tag/v0.27.0) Glean.js removed the `debug` option for the configration object. So the way that was set up right now would just be ignored by Glean.

This PR addresses that issue.

### DIFF
--- a/src/ui/main.qml
+++ b/src/ui/main.qml
@@ -309,13 +309,10 @@ Window {
         }
 
         function onInitializeGlean() {
-            var debug = {};
             if (VPN.debugMode) {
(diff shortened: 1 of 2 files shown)


bakulf (approved):

bakulf:
You should update the qml tests.

bakulf:
Merged! Thanks!

PR #3097, opened 2022-03-21
Add crash reporter opt in setting and UI

mkestler-rtp:
Add a setting check to enable autosend of crash reports.  Also add a checkbox on the crash reporter dialog to enable it there as well.


translations/strings.yaml
```diff
@@ -317,6 +317,9 @@ settings:
     value: Tunnel VPN through port 53/DNS
     comment: Tunnel is used as a verb, as in to direct traffic through
   tunnelPort53Description: Use port 53/DNS for Wireguard traffic, which may help evade firewalls
+  crashOptInLabel:
+    value: Autosend crash reports
+    comment: Main label for the auto send crash reports option.  Goes above the longer description.
```
flodolo, line 322:
```suggestion
    comment: Main label for the autosend crash reports option. Goes above the longer description.
```

Let's avoid double spaces everywhere.

flodolo (approved):

src/ui/settings/ViewPrivacySecurity.qml
```diff
@@ -55,19 +55,14 @@ Item {
...
-
             VPNCheckBoxRow {
                 id: dataCollection
                 objectName: "dataCollection"
                 width: parent.width - VPNTheme.theme.windowMargin
                 anchors.left: parent.left
                 anchors.right: parent.right
                 anchors.rightMargin: VPNTheme.theme.windowMargin
-
+                showDivider: false
```
lesleyjanenorton, line 65: Similar question as below but if `crashOptInCheckBox` is only visible on certain platforms, then this should be visible if `!crashOptInCheckBox.visible`.

mkestler-mozilla: Data collection is for glean, not crash reporting.  It is available everywhere.

lesleyjanenorton (approved):

src/ui/settings/ViewPrivacySecurity.qml
mkestler-mozilla: I've updated this now that I understand the comments.

bakulf:
@lesleyjanenorton do you think we still need this PR? It has never landed

lesleyjanenorton:
@bakulf not sure! It seems like nice to have but do you remember if there was some reason why it wasn't merged before? I can rebase tomorrow and try running.

birdsarah:
Closing as stale.

PR #3099, opened 2022-03-22
Fix the IOS compilation in CI - VPN-1955

bakulf:


.github/workflows/ios-build.yaml
```diff
@@ -47,10 +47,9 @@ jobs:
       - name: Importing translation files
         shell: bash
         run: |
-          export PATH=`pwd`/qt_ios/6.2.3/clang_64/bin:$PATH
-          export PATH=`pwd`/qt_ios/6.2.3/ios/bin:$PATH
+          export PATH=`pwd`/qt_ios/6.2.3/clang_64/bin:`pwd`/qt_ios/6.2.3/ios/bin:$PATH
           git submodule update --remote --depth 1 i18n
-          python3 scripts/utils/import_languages.py -m
+          python3 scripts/utils/import_languages.py -m -q "`pwd`/qt_ios/6.2.3/macos/bin"
```
mbirghan, line 52: this has to be `clang_64` instead of `macos`

.github/workflows/ios-build.yaml
```diff
@@ -47,10 +47,9 @@ jobs:
       - name: Importing translation files
         shell: bash
         run: |
-          export PATH=`pwd`/qt_ios/6.2.3/clang_64/bin:$PATH
-          export PATH=`pwd`/qt_ios/6.2.3/ios/bin:$PATH
+          export PATH=`pwd`/qt_ios/6.2.3/clang_64/bin:`pwd`/qt_ios/6.2.3/ios/bin:$PATH
```
mbirghan, line 50: we should switch the two so that `qmake` gets found in `ios` first and not in `clang_64`.

scripts/utils/import_languages.py
```diff
@@ -24,6 +24,9 @@
 parser.add_argument(
     '-m', '--macos', default=False, action="store_true", dest="ismacos",
     help='Include the MacOS bundle data')
+parser.add_argument(
```
mbirghan, line 27: I think we cannot do it this way. 

The thing is that `ios` does have its own `qmake` but only lacks `lupdate` and `lconvert`. I believe if we do it like this we will use the wrong `qmake`.

mbirghan (changes requested):

mbirghan (approved):

Issue #2774, opened 2022-02-16
VPN cannot be turned ON on Android


data-sync-user:
*VPN version:*

* 2.7.1 (16449500) - QT6 build;

*Platforms:*

* Android 10;

*Prerequisites:*

* Have an FxA account with a subscription for Mozilla VPN;

* Have a fresh install of the VPN QT6 build;

*Steps to reproduce:*

* Open the VPN app;
* Tap Sign up;
* Enter valid credentials;
* Finalize log in;
* Turn ON the VPN;

*Expected results:*

* VPN is turned on;

*Actual results:*

* VPN is blocked between states;

*Notes:*

* This is a QA-major issue because the VPN cannot be turned on;
* The workaround for this is to swipe-kill the app, you will see a crash, and reopen it.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1757)

data-sync-user:
➤ Sebastian Streich commented:

Is this still an issue? Just pulled up the latest build and can connect fine 😅 

If it is, can maybe get logs and tell me what android version this is?

data-sync-user:
➤ Valentina Virlics commented:

Verified this on the VPN 2.8 (1643750) build on Android 10.

The VPN can be turned on with success.

data-sync-user:
➤ Bianca Hidecuti commented:

Sebastian Streich , I am able to reproduce this issue on Mozilla VPN 2.8.0 (16474170) - link ( https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/aOSksCjhTwyfc7Yi5_ksaQ/runs/0/artifacts/public/build/mozillavpn-arm64-v8a-release.apk ), while using Android 10, after a fresh install, therefore I’m reopening this issue.

Attaching some logs.

[^Mozilla VPN Logs (88).txt]

data-sync-user:
➤ Sebastian Streich commented:

Bianca Hidecuti - Hi! I was wondering if you could check if the error still occurs with this test build ? I think I found a solution, but I would prefer a double check before I commit this 🙂 

[^android-build-arm64-v8a-release-signed.apk]

data-sync-user:
➤ Bianca Hidecuti commented:

Sebastian Streich , hello. The VPN is successfully turned on after a fresh install, while using this build: 2.8.0 (16479440).

data-sync-user:
➤ Bianca Hidecuti commented:

[...]

PR #3100, opened 2022-03-22
Android: Do not block UI thread for ipc

strseb:
We're deadlocking the UI and Qt thread, after you press "yes" on the permission prompt it seems. (on signed, release builds only )
Thinking about it, there is no reason to 
A: move `s_instance->resume_activate()`to the QT-Main thread anymore, as now IPC is no longer handled by QTIBinder.
B: Force "AndroidActivity::sendToService" to be on the Android-UI thread; it does not have side effects and the transaction later is happening on the ipc thread anyway. 

Should close https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2774

Update: Got confirmation of QA that this fixes the issue.

### DIFF
--- a/src/platforms/android/androidcontroller.cpp
+++ b/src/platforms/android/androidcontroller.cpp
@@ -268,9 +268,7 @@ void AndroidController::startActivityForResult(JNIEnv* env, jobject /*thiz*/,
 
     if (resultCode == ACTIVITY_RESULT_OK) {
       logger.debug() << "VPN PROMPT RESULT - Accepted";
-      AndroidUtils::dispatchToMainThread(
-          [&] { s_instance->resume_activate(); });
-
+      s_instance->resume_activate();
       return;
     }
     // If the request got rejected abort the current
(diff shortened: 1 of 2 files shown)


src/platforms/android/androidcontroller.cpp
```diff
@@ -268,9 +268,7 @@ void AndroidController::startActivityForResult(JNIEnv* env, jobject /*thiz*/,
 
     if (resultCode == ACTIVITY_RESULT_OK) {
       logger.debug() << "VPN PROMPT RESULT - Accepted";
-      AndroidUtils::dispatchToMainThread(
-          [&] { s_instance->resume_activate(); });
-
+      s_instance->resume_activate();
```
bakulf, line 271: can you add an assert to check we are on the main-thread here?

bakulf (changes requested):

src/platforms/android/androidcontroller.cpp
strseb: we are not on the main thread here - also resume_activate only calls AndroidVPNActivity::sendToService(...) - which can be called from any thread.

bakulf (approved):

PR #3120, opened 2022-03-23
Move macosx_build.sh script into the taskcluster folder

bakulf:


mbirghan (approved):
This looks good.

We just have to maybe clean this up later because I am currently working on #2873 which adds a similar file to a similar directory.

PR #3124, opened 2022-03-23
Move ConnectionInfo out of Nebula

flozia:
This PR moves the ConnectionInfo components out of Nebula.


nebula/ui/components/qmldir
```diff
@@ -17,13 +17,8 @@ VPNChevron 0.1 VPNChevron.qml
 VPNClickableRow 0.1 VPNClickableRow.qml
 VPNCollapsibleCard 0.1 VPNCollapsibleCard.qml
 VPNConnectionInfo 0.1 VPNConnectionInfo.qml
```
bakulf, line 19: I still don't understand what goes into Nebula and what does not.
Most of these files should not be in nebula. What do you think?

src/ui/ui.qrc
```diff
@@ -10,6 +10,11 @@
         <file>authenticationInApp/ViewAuthenticationUnblockCodeNeeded.qml</file>
         <file>authenticationInApp/ViewAuthenticationVerificationSessionByEmailNeeded.qml</file>
         <file>authenticationInApp/ViewAuthenticationVerificationSessionByTotpNeeded.qml</file>
+        <file>components/VPNControllerView.qml</file>
```
bakulf, line 13: We have a guide style to call QML components `VPN`+name. It can change, but do we want to start with this PR?

src/ui/views/ViewMain.qml
```diff
@@ -10,6 +10,8 @@ import Mozilla.VPN 1.0
 import components 0.1
 import compat 0.1
 
+import "qrc:/ui"
```
bakulf, line 13: why this is needed?

bakulf (changes requested):

nebula/ui/components/qmldir
flozia: More complex and not reusable components should be probably not part of nebula. I think cleaning that up would better be part of the design system and theming efforts.

PR #3132, opened 2022-03-23
A new tool to create country/city localization - fixes VPN-1967

bakulf:
The way we localize countries and cities is via SPARQL queries to WikiData. This is OK, but there are a few issues:
- the mapping between city/country names and WikiData resources is based on strings and it fails often.
- we don't know what has been localized and what is new when the script runs (except for `git diff`).
- the mapping is a magic box and it's hard to debug it

In this PR there is a tool that fixes these issues. The mapping is done via WikiData IDs stored in the `servers.json` file. The IDs can be easily tested by opening the corresponding WikiData page.

We can also include this script in a github action to spot when new servers are added. This is not done yet but can be easily done as a follow-up.


flodolo:
I'll take a look in the coming days, but it's probably a good idea to get a reviewer familiar with JS to look at the code.

tools/serverlocalizer/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "serverlocalizer",
+  "version": "1.0.0",
+  "description": "The localizator of countries and cities",
```
flodolo, line 4:
```suggestion
  "description": "Tool to localize country and city names",
```

"localizator" really sounds like a stange movie character ;)

flodolo (approved):
The output looks good, I confess I'm quite lost on the JS code, as I'm not as familiar.

PR #3137, opened 2022-03-23
Tunnel log entries were not decoded correctly

mkestler-mozilla:
On windows, the conversion from the fixed length buffer read from the tunnel's ringbuffer resulted in a 512b fixed length string being written to the log.  When encoded and sent to guardian for a support request that turned in to 512 /u0000s which caused it to be rejected by FxA for length.
I now truncate the buffer before conversion.  Also fixed a casting issue that was causing a warning.

### DIFF
--- a/src/platforms/windows/daemon/windowstunnellogger.cpp
+++ b/src/platforms/windows/daemon/windowstunnellogger.cpp
@@ -101,15 +101,14 @@ void WindowsTunnelLogger::process(int index) {
   uchar* data = m_logdata + RINGLOG_HEADER_SIZE + offset;
 
   quint64 timestamp;
-  memcpy(&timestamp, data, 8);
+  memcpy(&timestamp, data, RINGLOG_TIMESTAMP_SIZE);
   if (timestamp <= m_startTime) {
     return;
   }
-
   QByteArray message((const char*)data + RINGLOG_TIMESTAMP_SIZE,
                      RINGLOG_MESSAGE_SIZE);
   int nullIndex = message.indexOf(0x0);
-  if (nullIndex > 0) {
+  if (nullIndex >= 0) {
     message.truncate(nullIndex);
   }
   logger.info() << QString::fromUtf8(message);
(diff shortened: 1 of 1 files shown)


src/platforms/windows/daemon/windowstunnellogger.cpp
```diff
@@ -107,6 +108,10 @@ void WindowsTunnelLogger::process(int index) {
 
   QByteArray message((const char*)data + RINGLOG_TIMESTAMP_SIZE,
                      RINGLOG_MESSAGE_SIZE);
+  int nullIndex = message.indexOf(0x0);
```
oskirby, line 111: Oooh! Good catch! We could also have used `QString::fromUtf(const char *str, int size = -1)` which would simplify this even further.

src/platforms/windows/daemon/windowstunnellogger.cpp
```diff
@@ -107,6 +108,10 @@ void WindowsTunnelLogger::process(int index) {
...
+  if (nullIndex > 0) {
```
bakulf, line 112: `>=` ?

mkestler-mozilla: No, Owen's suggestion is better.  The true root is trying to decode UTF16 as UTF8.  I'm just verifying that now.

src/platforms/windows/daemon/windowstunnellogger.cpp
mkestler-mozilla: Correct.  I'll post an update doing the proper conversion.

mkestler-mozilla: Turns out it isn't.  Left in the current truncation.

src/platforms/windows/daemon/windowstunnellogger.cpp
mkestler-mozilla: Conversion didn't work, so I made this change.

bakulf (approved):

PR #3140, opened 2022-03-23
Allow the lib path to be set during build config

Cimbali:
Some firefox builds (all 64-bit builds? just some?) look for the `native-messaging-hosts` under `/usr/lib64/native-messaging-hosts`. These changes allow to build for those systems with `qmake ... LIBPATH=/usr/lib64`.

The manifest file is now also updated to contain the install-configured path to the `mozillavpnnp` binary, rather than the hardcoded default path as previously.  Changes to `USRPATH` (or the new `LIBPATH`) previously caused the installed json file to point to a wrong location.

The sed command does not depend on the current path in the file and just replaces the line that contains ”path” with the appropriate path. Not sure this is optimal, but it works. An alternate way that depends on the current path (not sure it’s more readable) could be something like:
```
manifestFile.commands = @$(SED) \'s,\"path\":\ *"/usr/lib/,\"path\":\ \"$${LIBPATH}/,\' $$ORIG_MOZILLAVPN_JSON >  $$manifestFile.output
```

Finally the `mozillavpnnp` binary is now also installed to the directory of appropriate bitness as selected by the `LIBPATH` variable.

Allows fixing mozilla/multi-account-containers#2315 at build time instead of manually doing `sudo mv /usr/lib/… /usr/lib64/…`

### DIFF
--- a/src/qmake/platforms/linux.pri
+++ b/src/qmake/platforms/linux.pri
@@ -141,7 +141,7 @@ INSTALLS += systemd_service
 
 ORIG_MOZILLAVPN_JSON = $$PWD/../../../extension/manifests/linux/mozillavpn.json
 manifestFile.input = ORIG_MOZILLAVPN_JSON
-manifestFile.output = $$PWD/../../../linux/mozillavpn.json
+manifestFile.output = $${OBJECTS_DIR}/mozillavpn.json
 manifestFile.commands = @python3 -c \'with open(\"$$ORIG_MOZILLAVPN_JSON\") as fin, open(\"$$manifestFile.output\", \"w\") as fout: [print(l.replace(\"/usr/lib/\", \"$${LIBPATH}/\"), end=\"\", file=fout) for l in fin]\'
 manifestFile.CONFIG = target_predeps no_link
 QMAKE_EXTRA_COMPILERS += manifestFile
--- a/src/qmake/platforms/linux.pri
+++ b/src/qmake/platforms/linux.pri
@@ -79,6 +79,9 @@ HEADERS += \
 isEmpty(USRPATH) {
     USRPATH=/usr
 }
+isEmpty(LIBPATH) {
+    LIBPATH=$${USRPATH}/lib
+}
 isEmpty(ETCPATH) {
     ETCPATH=/etc
 }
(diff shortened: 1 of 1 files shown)


src/qmake/platforms/linux.pri
```diff
@@ -136,20 +139,33 @@ systemd_service.files = $$PWD/../../../linux/mozillavpn.service
 systemd_service.path = $${USRPATH}/lib/systemd/system
 INSTALLS += systemd_service
 
-manifestFirefox.files = $$PWD/../../../extension/manifests/linux/mozillavpn.json
-manifestFirefox.path = $${USRPATH}/lib/mozilla/native-messaging-hosts
+ORIG_MOZILLAVPN_JSON = $$PWD/../../../extension/manifests/linux/mozillavpn.json
+manifestFile.input = ORIG_MOZILLAVPN_JSON
+manifestFile.output = $$PWD/../../../linux/mozillavpn.json
+manifestFile.commands = @$(SED) \'/\"path\":/c\ \ \"path\":\ \"$${LIBPATH}/mozillavpn/mozillavpnnp\",\' $$ORIG_MOZILLAVPN_JSON >  $$manifestFile.output
```
bakulf, line 145: Does it work on Windows too? Is there anything we can do using a non posix command? Maybe a python script...

Cimbali: I think the `linux.pri` is only ever called on on linux no? I can definitely try a python hack instead if you want to support some non-posix-to-posix cross compiling or something of the sort.

src/qmake/platforms/linux.pri
```diff
@@ -136,20 +139,33 @@ systemd_service.files = $$PWD/../../../linux/mozillavpn.service
 systemd_service.path = $${USRPATH}/lib/systemd/system
 INSTALLS += systemd_service
 
-manifestFirefox.files = $$PWD/../../../extension/manifests/linux/mozillavpn.json
-manifestFirefox.path = $${USRPATH}/lib/mozilla/native-messaging-hosts
+ORIG_MOZILLAVPN_JSON = $$PWD/../../../extension/manifests/linux/mozillavpn.json
+manifestFile.input = ORIG_MOZILLAVPN_JSON
+manifestFile.output = $$PWD/../../../linux/mozillavpn.json
```
oskirby, line 144: I think my only concern here is that this writes stuff into the source directory during the build, we should probably place the intermediate manifest file into `OBJECTS_DIR`, or make sure that `.gitignore` is updated accordingly.

oskirby (changes requested):

oskirby (approved):

oskirby:
@Cimbali awesome! Thank you!

PR #3157, opened 2022-03-24
WIP: Wasm pr preview

strseb:
WIP. Not ready to review. 

-> Convert some inpector files to typescript to add type-info 
-> Add a new connection type WasmClient to connect to wasm-client that opens the inspector
-> Extend the wasm shell to connect to an inspector instance 
-> Hopefully generate a preview for a branch in the github ci


.github/workflows/wasm.yaml
```diff
@@ -126,13 +126,50 @@ jobs:
...
+        with:
+          name: WebAssembly Build Qt6
+          # Destination path
+          path: _site/branches/${{github.ref_name}}
+      - name: Download a Build Artifact
+        uses: actions/download-artifact@v2.0.8
+        with:
+          name: Inspector Build
+          path: _site/branches/${{github.ref_name}}/inspector
+      - name: Uploading site Build
```
bakulf, line 153: we need to rewrite the git history otherwise we will create billions commits

strseb: This is currently only generating a .zip artifact to check if the structure is correct.
But the idea is to checkout the gh-pages branch and then just amend the commit. 
Removing the "Cecilapp/GitHub-Pages-deploy@3.1.0" which will just add an extra commit :)

bakulf (approved):

Issue #3126, opened 2022-03-23
armv7 builds from releases/2.8 cannot be installed


data-sync-user:
**Affected versions:**

* Mozilla VPN armv7_Release builds from [+releases/2.8+](https://github.com/mozilla-mobile/mozilla-vpn-client/tree/releases/2.8) branch
* Link: [https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/2024890400](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/2024890400|smart-link) 

**Tested Platforms:**

* Android;

**Prerequisites:**

* Have a device with a processor that accepts armv7 builds 
* I use an Android 9 Samsung Galaxy S10 Plus

**Steps to reproduce:**

1. download the .zip from Github
1. Unzip the folder and select the _android~~build~~universal~~release~~signed_ APK
1. Install it on the above mentioned device

**Expected result:**

* The build is successfully installed;

**Actual result:**

* The build is installed but the app does not work; “Your application encountered a fatal error and cannot continue” error is displayed at app launch

**Notes:**

* The armv7 builds from Treeherder can successfully be installed and launched
* The device also supports v8a builds, both from Treeherder and release branch

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1970)

data-sync-user:
➤ Luminita Buruian commented:

Verified this using builds from both the Main and the +releases/2.8+ ( https://github.com/mozilla-mobile/mozilla-vpn-client/tree/releases/2.8 ) branches. When unziping the packeges there are no more androidbuilduniversalreleasesigned APKs, only the androidbuildarmeabiv7arelease-signed APKs, which both work as expected on my android device.

Sebastian Streich If this change in builds is expected, we can consider this closed as fixed.

Thank you.

CC: Valentina Virlics

data-sync-user:
➤ Sebastian Streich commented:

Hi Luminita Buruian  yes this was the expected change, thank you!

data-sync-user:
➤ Luminita Buruian commented:

Thank you.

Closing, verified as fixed

PR #3161, opened 2022-03-24
Remove Broken Apk's in release tasks

strseb:
Currently the release task generates 5 apks (each arch, + universal) - 
However, as qt only builds one arch now - only the matching apk for the job actually contains everything. 
Therefore, we should delete the broken any other APKs - so noone is wondering why "universal" won't work for them :) 

Closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3126

### DIFF
--- a/.github/workflows/android_qt6.yaml
+++ b/.github/workflows/android_qt6.yaml
@@ -61,7 +61,7 @@ jobs:
             path: ${{ matrix.config.apk }}/android-build-${{ matrix.arch.qmake }}-release-signed.apk
       - name: Upload Unsigned APK${{ matrix.config.args }}
         uses: actions/upload-artifact@v1
-        if: ${{ matrix.config.sign }} == false
+        if: matrix.config.sign == false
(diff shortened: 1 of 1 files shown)


.github/workflows/android_qt6.yaml
```diff
@@ -52,7 +52,11 @@ jobs:
           AUTOGRAPH_TOKEN: ${{ secrets.AUTOGRAPH_KEY }}
         run: |
           bash ./scripts/android/sign.sh ${{ matrix.config.apk }}
+          # Delete all unsigned
           rm ${{ matrix.config.apk }}/*unsigned.apk
+          # Delete all other apks except the selected arch
+          cd ${{ matrix.config.apk }}
+          ls | grep -v "${{ matrix.arch.qmake }}" | xargs rm
```
bakulf, line 59: this is really really dangerous. Anything better?

bakulf (approved):

PR #3163, opened 2022-03-24
Fix handling of controller state changes while benchmarking connection

### LINKED ISSUE #3150: “Unexpected error” screen is not displayed when disconnecting from the Internet on iOS
**Affected versions:**

* Mozilla VPN 2.8.0 (2.202203240157);

**Tested Platforms:**

* iOS;

**Prerequisites:**

* Have a FXA account with an active subscription for Mozilla VPN;

* Have Mozilla VPN app installed and opened;

* Be signed in;

**Steps to reproduce:**

1. Turn ON the VPN;
1. Tap on the speed icon from the top left corner of the screen;
1. Disconnect from the internet;
1. Tap on the refresh button from the CI screen;
1. Observe the behavior;

**Expected result:**

* The “Unexpected error” screen is displayed after a few seconds;

**Actual result:**

* The “Unexpected error” screen is not displayed after a few seconds. The Loading screen does not disappear. The “X” button disappears from the Loading screen.

**Notes:**

* This is a qa-major issue because the user is not informed regarding his connection status and he cannot quit the Loading screen;
* See the attached video and logs;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1981)

flozia:
Fixes #3150


nebula/ui/components/VPNConnectionInfoScreen.qml
```diff
@@ -97,7 +97,8 @@ Rectangle {
         if (VPNConnectionBenchmark.state === VPNConnectionBenchmark.StateInitial) {
             VPNConnectionBenchmark.start();
         } else if (VPNConnectionBenchmark.state === VPNConnectionBenchmark.StateRunning
```
bakulf, line 99: does it mean: `} else {` ?

src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -95,8 +97,6 @@ void ConnectionBenchmark::start() {
           [this, downloadTask]() { m_benchmarkTasks.removeOne(downloadTask); });
   m_benchmarkTasks.append(downloadTask);
   TaskScheduler::scheduleTask(downloadTask);
-
-  setState(StateRunning);
```
bakulf, line 99: why this change?

bakulf (approved):

src/connectionbenchmark/connectionbenchmark.cpp
flozia: So that we have set `m_state == StateRunning` in case `handleStabilitiChange()` gets called in `L79`.

nebula/ui/components/VPNConnectionInfoScreen.qml
flozia: Yes, it does — changed!

bakulf (approved):

Issue #3119, opened 2022-03-23
QML warnings


data-sync-user:
Debugging other issues, I have seen these:

```
[21.03.2022 14:27:23.343] Warning: qrc:/nebula/components/VPNUIStates.qml:62:17: Unable to assign [undefined] to QColor (VPNUIStates.qml:62)
[21.03.2022 14:27:23.343] Warning: qrc:/nebula/components/VPNUIStates.qml:62:17: Unable to assign [undefined] to QColor (VPNUIStates.qml:62)
[21.03.2022 14:27:23.345] Debug: (networking - ConnectionHealth) Received DNS ping: 17 msec
[21.03.2022 14:27:24.997] Warning: qrc:/qt-project.org/imports/QtQuick/Controls/Windows/ScrollBar.qml:66:49: Unable to assign [undefined] to int (ScrollBar.qml:66)
[21.03.2022 14:27:24.997] Warning: qrc:/qt-project.org/imports/QtQuick/Controls/Windows/ScrollBar.qml:113:49: Unable to assign [undefined] to int (ScrollBar.qml:113)
```

And:

```
[21.03.2022 14:27:54.994] Warning: qrc:/nebula/components/VPNUIStates.qml:62:17: Unable to assign [undefined] to QColor (VPNUIStates.qml:62)
[21.03.2022 14:27:54.994] Warning: qrc:/nebula/components/VPNUIStates.qml:62:17: Unable to assign [undefined] to QColor (VPNUIStates.qml:62)
[21.03.2022 14:27:54.994] Warning: qrc:/nebula/components/VPNToggle.qml:248:9: Unable to assign [undefined] to double (VPNToggle.qml:248)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1966)

data-sync-user:
➤ Betty Fleming commented:

Lesley Norton do we want to keep this or open a new one?

PR #3173, opened 2022-03-26
Fix a bunch of assorted QML warnings.

oskirby:
There have been quite a few QML warnings appearing in the logs that tend to generate a lot of distracting clutter. Let's fix them.

We had quite a few colours in the theme which were missing their `buttonDisabled` value. For the most part I just picked a value by playing with the Value while keeping the Hue and Saturation the same as the button's base colour (in HSV colour space). I don't actually know if any of these colours appear anywhere in the UI.

This still leaves a few warnings about missing `accessibleName` properties, eg:
```
Warning: qrc:/nebula/components/VPNIconButton.qml:31:9: Unable to assign [undefined] to QString (VPNIconButton.qml:31)
Warning: qrc:/nebula/components/VPNIconButton.qml:21:5: Unable to assign [undefined] to QString (VPNIconButton.qml:21)
Warning: qrc:/nebula/components/VPNConnectionInfo.qml:58:9: Unable to assign [undefined] to QString (VPNConnectionInfo.qml:58)
```


oskirby:
It occurs to me that I might have duplicated some of the work needed for #3119

lesleyjanenorton (approved):

PR #3174, opened 2022-03-27
Fix 1615 - Checkboxes and radio buttons can be controlled by tapping their labels

### LINKED ISSUE #1615: Clicking on a radiobox/checkbox label should enable/disable the input field.
Currently, only the checkbox/radiobox graphical objects are clickable. Maybe we should include the label in the MouseArea. @lesleyjanenorton what do you think?

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-889)
┆author: Data Sync

MattLichtenstein:
This allows checkboxes to be toggled and radio buttons to be selected when clicking their labels

Fixes #1615

https://user-images.githubusercontent.com/15353801/160300647-70700a26-3f6f-459b-a82c-67e3f20c4ca2.mov

https://user-images.githubusercontent.com/15353801/160300648-e0b22696-fb59-401e-8f88-6ed7a759a69f.mov


nebula/ui/components/VPNMouseArea.qml
```diff
@@ -11,7 +11,8 @@ MouseArea {
 
     property var targetEl: parent
     property var uiState: VPNTheme.theme.uiState
-    property var onMouseAreaClicked: function() { parent.clicked() }
+    property bool propagateClickToParent: true
+    property var onMouseAreaClicked: function() { if(propagateClickToParent) parent.clicked() }
```
bakulf, line 15: `() => { if(propagateClickToParent) parent.clicked(); }`

nebula/ui/components/VPNMouseArea.qml
```diff
@@ -11,7 +11,8 @@ MouseArea {
 
     property var targetEl: parent
     property var uiState: VPNTheme.theme.uiState
-    property var onMouseAreaClicked: function() { parent.clicked() }
+    property bool propagateClickToParent: true
```
bakulf, line 14: `false` by default?

nebula/ui/components/VPNCheckBoxRow.qml
```diff
@@ -73,6 +74,11 @@ RowLayout {
             text: labelText
             color: VPNTheme.theme.fontColorDark
             horizontalAlignment: Text.AlignLeft
+
+            VPNMouseArea {
+                propagateClickToParent: false
```
bakulf, line 79: I like this but, not always we have a parent-child relationship between QML items.
What about if we implement a generic `VPNMouseAreaProxy` QML component with a: `propagateTo` property (or a better name). We can use the `propagateTo` property to link QML items.
Thoughts?

bakulf (commented):
Can you implement a QML test for this? Thank you!

nebula/ui/components/VPNMouseArea.qml
MattLichtenstein: It probably should be `false` by default, but the current functionality always propagates the click to the parent, so I left that as the default functionality to avoid having to make changes everywhere this component is used - mostly to avoid churn while I still feel my way around the codebase. 

I could change this though if you think it's worth it?

nebula/ui/components/VPNCheckBoxRow.qml
MattLichtenstein: Interesting. My question is, can't we just have whatever functionality we need in the `onClicked()` handler? Wouldn't that leave it more flexible (i.e. you can propagate the click to multiple other components if you wanted to)?

nebula/ui/components/VPNMouseArea.qml
bakulf: All good to have to `false`

src/ui/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -28,268 +28,245 @@ Item {
...
+            anchors.left: parent.left
 
-            labelText:  VPNl18n.SettingsDevShowOptionTitle
-            subLabelText: VPNl18n.SettingsDevShowOptionSubtitle
+            spacing: 0
 
-            isChecked: VPNSettings.developerUnlock
-            onClicked: VPNSettings.developerUnlock = !VPNSettings.developerUnlock
-        }
+            onImplicitHeightChanged: {
```
lesleyjanenorton, line 43: Nit: let's pad this number or add something like a `VPNVerticalSpacer{}` at the end of this `layout` so that there is a bit more of a margin between the `Qt version:` info text and the app window bottom. As is, I think this text will get covered up on iOS by the iPhone Home bar which is helpful for QA.

tests/qml/tst_VPNRadioDelegate.qml
```diff
@@ -0,0 +1,71 @@
+import QtQuick 2.15
```
bakulf, line 1: It seems we don't set the header in the qml test files. This is wrong, we should. Can you please add:

```
```

We can add the same header in the other files as a follow-up.

lesleyjanenorton (approved):

lesleyjanenorton:
🎉

PR #3176, opened 2022-03-28
Leak QmlEngineHolder

mbirghan:


src/qmlengineholder.cpp
```diff
@@ -9,6 +9,8 @@
 #include <QWindow>
 #include <QNetworkAccessManager>
 
+#include <QApplication>
```
bakulf, line 12: This is not needed

src/qmlengineholder.cpp
```diff
@@ -30,7 +32,9 @@ QmlEngineHolder::~QmlEngineHolder() {
 
 // static
 QmlEngineHolder* QmlEngineHolder::instance() {
-  Q_ASSERT(s_instance);
+  if(!s_instance) {
```
bakulf, line 35: Write a comment about why we do not set the parent object here.

bakulf (approved):
1. This should land to releases/2.8 only.
2. for the `main` branch, keep it as it is. No fix is needed. We need to investigate what we can do. And maybe file a QT bug if we are able to create a small test-case app.

PR #3182, opened 2022-03-28
Avoid a 0-division in CI

bakulf:


flozia (approved):
Thanks for catching this one!

PR #3183, opened 2022-03-28
Fix Not Loadable Lottie Animations on Android 12

strseb:
The QJSValue constructor for qstring actually constructs a js-number (a float actually) on android-11+ on arm64. (...Don't ask me why...)
Therefore getContext and loadanimation are not working there. 

Luckly engine->toScriptValue creates the right value here.
image


lottie/lib/lottieprivate.h
```diff
@@ -116,6 +116,8 @@ class LottiePrivate : public QQuickItem {
   LottieStatus m_status;
   bool m_autoPlay = false;
   QString m_fillMode = "stretch";
+  QString m_context_type = "2d";
```
bakulf, line 119: const QString

PR #3185, opened 2022-03-28
Update i18n repo in ci runs

strseb:
We had this line `git submodule update --remote --depth 1 i18n` in the package.sh prior. 
So therefore when adding the old gh-ci script back, that was not caught i guess :) 

Also let's print the current used translation commit in the script so we can easily check what commit was used in each build

### DIFF
--- a/.github/workflows/android_qt6.yaml
+++ b/.github/workflows/android_qt6.yaml
@@ -45,6 +45,7 @@ jobs:
         continue-on-error: true
         id: compile
         run: |
+          git submodule update --remote --depth 1 i18n 
           bash ./scripts/android/package.sh $QTPATH -A ${{matrix.arch.qmake}} -a ${{ secrets.ADJUST_SDK_TOKEN }} ${{ matrix.config.args }}
       - name: Sign Android
         if: ${{ matrix.config.sign }}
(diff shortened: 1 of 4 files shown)


.github/workflows/android_qt6.yaml
```diff
@@ -45,6 +45,7 @@ jobs:
         continue-on-error: true
         id: compile
         run: |
+          git submodule update --remote --depth 1 i18n
```
bakulf, line 48: not needed. `package.sh` calls `import_language.py`

taskcluster/scripts/build/android_build_release.sh
```diff
@@ -10,6 +10,7 @@ git submodule update
 ./scripts/utils/generate_glean.py
 # translations
 echo "Importing translations"
+git submodule update --remote --depth 1 i18n
```
bakulf, line 13: remove this

taskcluster/scripts/build/android_build_debug.sh
```diff
@@ -9,6 +9,7 @@ git submodule update
 # glean
 ./scripts/utils/generate_glean.py
 # translations
+git submodule update --remote --depth 1 i18n
```
bakulf, line 12: remove this too :)

bakulf (changes requested):

bakulf (approved):

PR #3191, opened 2022-03-29
Reduce the number of writing operations - fixes VPN-VPN-1996

bakulf:
We end up writing the setting file too many times. This PR reduces the number of writes to ~0. But it could introduce regressions. So, please, extra care when reviewing. Thanks!

### DIFF
--- a/src/settingsholder.cpp
+++ b/src/settingsholder.cpp
@@ -129,8 +129,10 @@ QString SettingsHolder::getReport() {
     return m_settings.value(key).toType();                              \
   }                                                                     \
   void SettingsHolder::setter(const type& value) {                      \
-    m_settings.setValue(key, value);                                    \
-    emit getter##Changed(value);                                        \
+    if (!has() || getter() != value) {                                  \
+      m_settings.setValue(key, value);                                  \
+      emit getter##Changed(value);                                      \
+    }                                                                   \
   }
 
 #include "settingslist.h"


src/models/servercountrymodel.cpp
```diff
@@ -127,6 +127,13 @@ bool ServerCountryModel::fromJsonInternal(const QByteArray& s) {
   return true;
 }
 
+void ServerCountryModel::writeSettings() {
```
oskirby, line 130: I'm noticing that all of the `FooModel::writeSettings()` are kind of exactly the same thing for every model. Is this something that we could just implement in the `BaseModel` class?

src/models/basemodel.h
```diff
@@ -0,0 +1,45 @@
...
+class BaseModel {
+ public:
+  BaseModel() = default;
+  virtual ~BaseModel() = default;
+
+  bool initialized() const { return !m_rawJson.isEmpty(); }
+
+  void forget() {
+    m_rawJson.clear();
+    m_writeNeeded = false;
```
oskirby, line 19: This seems like calling `forget()` could potentially leave us without the ability to erase settings from disk. If `m_writeNeeded` is false then we might never actually touch the disk to clear out what was there before.

src/models/basemodel.h
```diff
@@ -0,0 +1,45 @@
...
+ protected:
+  void storeJson(const QByteArray& json, bool setWriteNeeded) {
+    m_rawJson = json;
+    if (setWriteNeeded) {
+      m_writeNeeded = true;
+    }
+  }
+
+  bool compareJson(const QByteArray& json) const {
+    return json.isEmpty() || m_rawJson != json;
```
oskirby, line 31: I am slightly confused as to why `json.isEmpty()` is treated specially here. Surely if `m_rawJson` is empty, and it stays empty, then there's no reason to write anything.

oskirby (commented):
If I understand the approach being taken here, we are just checking if the settings values have changed, and then only calling the `SettingsHolder` setter function if there is a difference. Isn't this something that we could have implemented by adding an extra `if` statement to the `SETTING()` macro (see: settingholder.cpp:123) to do that in the setter itself?

bakulf:
> If I understand the approach being taken here, we are just checking if the settings values have changed, and then only calling the `SettingsHolder` setter function if there is a difference. Isn't this something that we could have implemented by adding an extra `if` statement to the `SETTING()` macro (see: settingholder.cpp:123) to do that in the setter itself?

I don't know why I was making my life so complex. Yes, we can.

oskirby (approved):

PR #3194, opened 2022-03-29
Fix high-DPI scaling on Android

### LINKED ISSUE #2770: White border for mobile on-boarding screens
*VPN version:*

* 2.7.0 (16447900) - QT6 build;

*Platforms:*

* Android 10;

*Prerequisites:*

* Have an FxA account with a subscription for Mozilla VPN;

* Have a fresh install of the VPN app QT6 build;

*Steps to reproduce:*

* Open the VPN app;
* Observe on-boarding screens borders;

*Expected results:*

* There are no white borders visible;

*Actual results:*

* While navigating through the on-boarding screen, some white borders are visible.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1753)

flozia:
⚠️ This PR possibly has wide-ranging consequences for scaling on Android devices. If we decide to make this change we should test the change on as many Android devices as possible.

On some Android devices the dimensions of the application window are not filling the screen entirely. The issue seems to be rooted in how the app is scaled for high-DPI screens. The default policy for rounding the DPI scale factor for Android is different from other devices. This behaviour can be adjusted by using [QGuiApplication::setHighDpiScaleFactorRoundingPolicy](https://doc.qt.io/qt-5/qguiapplication.html#setHighDpiScaleFactorRoundingPolicy). From the documentation: 

> The default value is Qt::HighDpiScaleFactorRoundingPolicy::Round. On Qt for Android the default is Qt::HighDpiScaleFactorRoundingPolicy::PassThrough, which preserves historical behavior from earlier Qt versions.

Setting `Qt::HighDpiScaleFactorRoundingPolicy::Round` for Android seems to resolve the issue and the window is scaled correctly.

**Notes**
- This change has possibly an effect on scaling for all Android devices. To keep track of the devices that the change was tested on I created the following sheet: [Checklist for high-DPI scaling on Android](https://docs.google.com/spreadsheets/d/1J6m-g0ARY4p3uWehuk0WjCoP5Be2jqOjubYP7hF2QHY/edit#gid=0)
- For a list of devices on which this issue is currently reproducible see: #2770

**Screenshots**
Device: Nexus 5X
Version: Android 11.0
Arch: arm64

|Before|After|
|---|---|
|image|image|

Resolves #2770

### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -164,6 +164,11 @@ int Command::runQmlApp(std::function<int()>&& a_callback) {
   QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 
+#ifdef MVPN_ANDROID
+  QGuiApplication::setHighDpiScaleFactorRoundingPolicy(
+      Qt::HighDpiScaleFactorRoundingPolicy::Round);
+#endif
+
   QApplication app(CommandLineParser::argc(), CommandLineParser::argv());
 
   QCoreApplication::setApplicationName("Mozilla VPN");
--- a/src/ui/main.qml
+++ b/src/ui/main.qml
@@ -60,7 +60,7 @@ Window {
     }
 
     screen: Qt.platform.os === "wasm" && Qt.application.screens.length > 1 ? Qt.application.screens[1] : Qt.application.screens[0]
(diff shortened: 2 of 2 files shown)


lesleyjanenorton (approved):
Nice. It looks like there is a very slight improvement to text rendering as well, at least in the example provided. I'd like to merge this and test it on as many devices and emulators as possible. This would be super easy to back out if there are side effects.

bakulf (approved):

PR #3202, opened 2022-03-29
Rescan for dumps

mkestler-mozilla:
This is a working theory that the dump files might not be written at the time the crash reporter is launched.  There is a 5x retry of the directory listing with an increasing interval (+ 1s each time).

### DIFF
--- a/src/crashreporter/crashreporterapp.cpp
+++ b/src/crashreporter/crashreporterapp.cpp
@@ -11,10 +11,6 @@
 #include "crashreporterfactory.h"
 
 int CrashReporterApp::main(int argc, char* argv[]) {
-#if QT_VERSION < 0x060000
-  // This flag is set by default in qt6.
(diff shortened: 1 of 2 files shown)


src/crashreporter/crashreporterapp.cpp
```diff
@@ -11,10 +11,6 @@
 #include "crashreporterfactory.h"
 
 int CrashReporterApp::main(int argc, char* argv[]) {
-#if QT_VERSION < 0x060000
-  // This flag is set by default in qt6.
-  QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-#endif
```
bakulf, line 17: Unrelated?

mkestler-mozilla: Yes, but since we are all in on Qt6, I thought I'd pull it.

PR #3203, opened 2022-03-29
Build functional tests using GCC-10

oskirby:
The functional tests are often failing due to a `grcov` parsing error which appears to be getting a float/integer conversion wrong. Hopefully we can work around the error by switching to a newer compiler.

See: https://github.com/mozilla/grcov/issues/570

### DIFF
--- a/.github/workflows/functional_tests.yaml
+++ b/.github/workflows/functional_tests.yaml
@@ -65,8 +65,8 @@ jobs:
 
           # Add external PPA, latest version of QT is 5.12.x for Ubuntu 20.04
           sudo add-apt-repository ppa:beineri/opt-qt-5.15.2-focal -y
-          sudo apt-get -o "Dir::Cache::archives=$(pwd)/build/archive" install -y \
-               git qt515base qt515tools qt515svg qt515networkauth-no-lgpl qt515charts-no-lgpl \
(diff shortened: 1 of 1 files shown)


bakulf (approved):

.github/workflows/functional_tests.yaml
```diff
@@ -85,7 +85,8 @@ jobs:
 
           # Delete unit tests, so we can get to testing faster
           sed -i '/tests\/unit/d' mozillavpn.pro
-          qmake CONFIG+=DUMMY QMAKE_CXXFLAGS+=--coverage QMAKE_LFLAGS+=--coverage CONFIG+=debug CONFIG+=inspector QT+=svg
+          qmake QMAKE_CXX=g++-10 QMAKE_LINK=g++-10 QMAKE_CXXFLAGS+=--coverage QMAKE_LFLAGS+=--coverage \
```
bakulf, line 88: Can you also change the grcov failure workaround in the GitHub action?

oskirby: Unfortunately, bumping the compiler to GCC-10 didn't actually solve the issue with our github workflow. I think this is something that we actually need to fix in the grcov project instead.

PR #3208, opened 2022-03-30
Verification code is a 8-digit string - fixes VPN-2000

bakulf:
This is the fix for the `main` branch. I'm about to submit a special PR for 2.8 because we are in string freeze and we don't want to add new content now.

### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -274,12 +274,12 @@ networkSettings:
 
 captivePortalAlert:
   title: Captive portal detected!
-  header: A captive portal is when the guest Wi-Fi you are using requires you to log in to get internet access. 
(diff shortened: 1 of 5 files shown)


flozia (approved):

translations/strings.yaml
```diff
@@ -274,12 +274,12 @@ networkSettings:
 
 captivePortalAlert:
   title: Captive portal detected!
-  header: A captive portal is when the guest Wi-Fi you are using requires you to log in to get internet access.
```
flodolo, line 277: Can we enforce no trailing whitespaces via linter? It's really not useful to blame a file, and I'd prefer to enforce this.

flodolo (approved):

translations/strings.yaml
bakulf: Yes. Let's do this. follow-up.

PR #3236, opened 2022-03-31
Only install on Win10 or higher, 64bit, x86

mkestler-mozilla:
Block install on non x86 64 bit platforms and < Win10
Added the 2 Condition statements under Product.  The rest is the clang format.

### DIFF
--- a/windows/installer/MozillaVPN_prod.wxs
+++ b/windows/installer/MozillaVPN_prod.wxs
@@ -11,81 +11,85 @@
 <?define WintunPlatform = "amd64" ?>
 
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-	<Product
+  <Product
 		Id="*"
 		Name="Mozilla VPN"
 		Language="1033"
 		Version="!(bind.FileVersion.MozillaVPNExecutable)"
 		Manufacturer="Mozilla Corporation"
 		UpgradeCode="$(var.UpgradeCode)">
-		<Package
+    <Condition Message="This application is only supported on Windows 10 or higher.">
+      <![CDATA[Installed OR (VersionNT >= 603)]]>
+    </Condition>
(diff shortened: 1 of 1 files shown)


windows/installer/MozillaVPN_prod.wxs
```diff
@@ -11,81 +11,85 @@
...
 		Name="Mozilla VPN"
 		Language="1033"
 		Version="!(bind.FileVersion.MozillaVPNExecutable)"
 		Manufacturer="Mozilla Corporation"
 		UpgradeCode="$(var.UpgradeCode)">
-		<Package
+    <Condition Message="This application is only supported on Windows 10 or higher.">
+      <![CDATA[Installed OR (VersionNT >= 603)]]>
+    </Condition>
+    <Condition Message="This application is only supported on 64 bit x86 platforms."> %PROCESSOR_ARCHITECTURE="AMD64" </Condition>
```
bakulf, line 24: https://wixtoolset.org/documentation/manual/v3/howtos/redistributables_and_install_checks/block_install_on_os.html
IN here they say to check the `VersionNT64` var. I cannot find the documentation for this `PROCESS_ARCHITECTURE="AMD64"` can you please share some?

mkestler-mozilla: I used it based on this:
https://stackoverflow.com/questions/8853653/wix-installer-differentiate-32bit-from-64bits
We can't use VersionNT64 because ARM64 would still be 64bit.

bakulf (approved):
We need the QA to test this.

bakulf:
I guess this should not be included in 2.8.

ValentinaPC:
@bakulf Verified this while using Mozilla VPN Windows build - 2.9.0 (2.202206271708) - from Taskcluster on a **Windows 7 x64** machine and on a **Windows 10 x32** machine.
Mozilla VPN app cannot be installed on any of the above mentioned devices. 
See below screenshots for reference.
image_2022_06_28T13_05_23_333Z
image_2022_06_28T14_01_16_761Z

If anything else needs to be checked here, please let us know.

Issue #3002, opened 2022-03-14
Open multiple network requests for download benchmark


flozia:
We should consider opening multiple network requests with different files to mitigate slow server responses.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1905)

Issue #3000, opened 2022-03-14
Add functional tests for connection info screen


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1721)

PR #3249, opened 2022-04-01
Parallelize the connection download benchmark

oskirby:
## Description

In an attempt to get benchmark results closer to what's reported by speedtest.net, we can open multiple connections to try and do a better job of saturating the VPN network connection. This patch does so by exploiting DNS load balancing to open connections to multiple servers that ought to be hosting the download file.

## Reference

Closes: #3002
Closes: #3000

## Checklist
    
### DIFF
--- a/src/connectionbenchmark/benchmarktaskdownload.cpp
+++ b/src/connectionbenchmark/benchmarktaskdownload.cpp
@@ -10,6 +10,7 @@
 
 #include <QByteArray>
 #include <QDnsLookup>
+#include <QHostAddress>
 
 constexpr const char* MULLVAD_DEFAULT_DNS = "10.64.0.1";
 
--- a/src/connectionbenchmark/benchmarktaskdownload.cpp
+++ b/src/connectionbenchmark/benchmarktaskdownload.cpp
@@ -12,7 +12,9 @@
 #include <QDnsLookup>
 #include <QHostAddress>
 
+#if !defined(MVPN_DUMMY)
 constexpr const char* MULLVAD_DEFAULT_DNS = "10.64.0.1";
+#endif
 
 namespace {
 Logger logger(LOG_MAIN, "BenchmarkTaskDownload");
@@ -39,7 +41,9 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
 
   if (state == BenchmarkTask::StateActive) {
     // Start DNS resolution
+#if !defined(MVPN_DUMMY)
     m_dnsLookup.setNameserver(QHostAddress(MULLVAD_DEFAULT_DNS));
+#endif
     m_dnsLookup.lookup();
   } else if (state == BenchmarkTask::StateInactive) {
     for (NetworkRequest* request : m_requests) {
--- a/src/connectionbenchmark/benchmarktaskdownload.cpp
+++ b/src/connectionbenchmark/benchmarktaskdownload.cpp
@@ -11,6 +11,7 @@
 #include <QByteArray>
 #include <QDnsLookup>
 #include <QHostAddress>
+#include <QScopeGuard>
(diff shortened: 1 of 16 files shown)


src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -31,43 +40,80 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
   if (state == BenchmarkTask::StateActive) {
-    m_request = NetworkRequest::createForGetUrl(this, m_fileUrl);
-    connect(m_request, &NetworkRequest::requestUpdated, this,
+    // Start DNS resolution
+#if !defined(MVPN_DUMMY)
+    m_dnsLookup.setNameserver(QHostAddress(MULLVAD_DEFAULT_DNS));
+#endif
+    m_dnsLookup.lookup();
+  } else if (state == BenchmarkTask::StateInactive) {
+    for (NetworkRequest* request : m_requests) {
```
bakulf, line 49: Is there a way to cancel the DNS lookup?

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -31,43 +40,80 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
+  }
+  if (m_dnsLookup.hostAddressRecords().isEmpty()) {
+    logger.error() << "DNS Lookup Failed: no records";
+    emit finished(0, true);
+    emit completed();
+    return;
+  }
+  logger.debug() << "DNS Lookup Finished";
+
+  for (const QDnsHostAddressRecord& record : m_dnsLookup.hostAddressRecords()) {
```
bakulf, line 71: What about if we have been cancelled in the meantime?

bakulf (changes requested):

src/connectionbenchmark/benchmarktaskdownload.cpp
oskirby: Yes, there is `QDnsLookup::abort()` that ought to cancel. I'll add another commit to address it.

src/connectionbenchmark/benchmarktaskdownload.cpp
oskirby: Ah yes, I neglected to think of that possibility. I'll add another commit to address it.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -31,43 +40,85 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
+      request->abort();
+    }
+    m_requests.clear();
+    m_dnsLookup.abort();
+  }
+}
+
+void BenchmarkTaskDownload::dnsLookupFinished() {
+  if (state() != BenchmarkTask::StateActive) {
+    logger.warning() << "DNS Lookup finished after task aborted";
```
bakulf, line 59: we still need to complete the task, otherwise, the task chain will be blocked.
I suspect you can treat this and the following if-stmt together.

bakulf (changes requested):

src/connectionbenchmark/benchmarktaskdownload.cpp
oskirby: Okay, I think I have gotten the cancellation behaviour correct now. Calling `m_dnsLookup.abort()` should trigger the `finished()` signal with a status of `QDnsLookup::OperationCancelledError`. We've kept the `StateActive` check just in case we happen to catch a race condition where the user clicks cancel before a pending DNS lookup gets processed.

bakulf (approved):

PR #3250, opened 2022-04-02
Delay initial handleStabilityChange call

flozia:
## Description

If the speed test is restarted from the error screen it seems like we do not run it again. That’s the case because if we have no signal at the time the test is retaken we are setting `StateError` again. Let’s delay `handleStabilityChange` so we do not abort the speed test immediately.

## Reference

Refresh button from the connection information unexpected error screen is not working ([VPN-1944](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-1944))

## Checklist
    
### DIFF
--- a/src/connectionbenchmark/connectionbenchmark.cpp
+++ b/src/connectionbenchmark/connectionbenchmark.cpp
@@ -12,6 +12,7 @@
 #include "logger.h"
 #include "mozillavpn.h"
 #include "taskscheduler.h"
+#include "timersingleshot.h"
 
 namespace {
 Logger logger(LOG_MODEL, "ConnectionBenchmark");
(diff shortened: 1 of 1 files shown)


src/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -75,7 +76,7 @@ void ConnectionBenchmark::start() {
   setState(StateRunning);
 
   if (vpn->connectionHealth()->stability() == ConnectionHealth::NoSignal) {
-    handleStabilityChange();
+    TimerSingleShot::create(this, 3000, [this]() { handleStabilityChange(); });
```
bakulf, line 79: This sounds like a hack. The "real" fix should be to monitor the connection stability and at the next good connection state, we restart the benckmark.

bakulf (changes requested):

PR #3252, opened 2022-04-04
Introduce the guide engine

bakulf:
## Description

This is the Guide engine. It does the following:
- retrieves a list of guide JSON files from a `qrc:/guides` path
- validates the JSON documents
- exposes the Guide objects to QML

The strings are generated from the JSON format using IDs. The eng test is the only one to included in the JSON file. The strings IDs are:
- Guide title: guide.<id>.title
- Single block (type title or text): guide.<id>.block.<id_block>
- List block (type list): guide.<id>.block.<id_block>.<sub_block>

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2025


scripts/utils/generate_strings.py
```diff
@@ -89,30 +91,114 @@ def generateStrings():
...
+    )
+
+    id_re = re.compile(r"[A-Za-z0-9_]+")
+    guide_ids = []
+
+    for guide_filename in os.listdir(guides_path):
+        if not guide_filename.endswith(".json"): continue
+        with open(os.path.join(guides_path, guide_filename)) as guide_file:
+            guide_json = json.load(guide_file)
+        if not "id" in guide_json:
```
flodolo, line 124: Wouldn't make more sense to have all these sanity checks as a linter when adding guide json files ? 

Also, would a JSON schema be better, given how many checks there are (I have never done that personally for my small projects, so no clue how much work it requires).

scripts/utils/generate_strings.py
```diff
@@ -89,30 +91,114 @@ def generateStrings():
...
+                            "value": [subblock["content"]],
+                            "comments": [f"A guide block content of type {block['type']}"],
+                        })
+            else:
+                string_ids.append(
+                    {
+                        "enum_id": enum_id,
+                        "string_id": f"guide.{guide_json['id']}.block.{block['id']}",
+                        "value": [block["content"]],
+                        "comments": [f"A guide block content of type {block['type']}"],
```
flodolo, line 189: The JSON should have space for comments associated to a string.

flodolo (commented):
It would be a good thing to pass all these Python files through black, maybe a separate follow-up.

scripts/utils/generate_strings.py
```diff
@@ -89,30 +95,210 @@ def generateStrings():
...
+                    enum_id = pascalize(
+                        f"guide_{guide_json['id']}_block_{block['id']}_{subblock['id']}"
+                    )
+                    if enum_id in guide_ids:
+                        exit(
+                            f"Duplicate id {enum_id} when parsing {guide_filename} - subblock {subblock['id']}"
+                        )
+                    guide_ids.append(enum_id)
+
+                    comment = f"A guide block content of type {block['type']}"
```
flodolo, line 195: The resulting comment sounds very clunky. What about something like this?

```
                    comment_types = {
                        "text": f"Standard text in a guide block",
                        "title": f"Title in a guide block",
                        "list": f"Bullet list item in a guide block",
                    }
                    if "comment" in subblock:
                        comment = subblock["comment"]
                    else:
                        comment = comment_types.get(block['type'], "")
```

You'd need to fix also line 208.

scripts/utils/generate_strings.py
```diff
@@ -89,30 +95,210 @@ def generateStrings():
...
+                "string_id": f"guide.{guide_json['id']}.title",
+                "value": [guide_json["title"]],
+                "comments": [comment],
+            }
+        )
+
+        if not "blocks" in guide_json:
+            exit(f"Guide file {guide_filename} does not have a blocks")
+        for block in guide_json["blocks"]:
+            if not "id" in block:
```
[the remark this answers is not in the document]
bakulf: Well, no because for instance, we will end up calling block["type"] with a null value. I prefer to keep these checks.

scripts/ci/check_jsonschema.py
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import argparse
+import json
+from jsonschema import validate, RefResolver
```
oskirby, line 8: These may need to be added to `requirements.txt`

oskirby (approved):

scripts/utils/generate_strings.py
```diff
@@ -3,21 +3,34 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import re
 import os
 import yaml
+import json
```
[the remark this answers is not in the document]
bakulf: It's a built-in module. No need to install it via pip.

PR #3266, opened 2022-04-05
Update connectionbenchmark.cpp

mkestler-mozilla:
The check for NoSignal caused the test to fast fail.  This made the UI appear to be unresponsive.  This relaunches the test in earnest, though with no connection it will fail, but it makes the app appear responsive.


bakulf:
OK what is happening with this PR is that we schedule the tasks. They will fail, but in the meantime, the user will see the animation.

bakulf (approved):

bakulf:
OK what is happening with this PR is that we schedule the tasks. They will fail, but in the meantime, the user will see the animation.

Issue #2231, opened 2021-11-09
Delete localStorage data when 'reset-and-quit'


data-sync-user:
We need to cleanup glean and in general QMLEngine offline data

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1313)

data-sync-user:
➤ Owen Kirby commented:

Unfortunately, this appears to be stuck in a bit of a chicken-and-egg state for Windows:

  * We cannot delete the local storage while it is opened by the QML engine.
  * Due to Qt bugs, we aren’t able to exit cleanly during reset-and-quit and don’t process signals properly to hook on shutdown.

data-sync-user:
➤ Santiago Andrigo commented:

Update from Owen Kirby for posterity: It is blocked due to upstream Qt bugs and given the low priority of the bug I don't think it's worth the effort to invest in a workaround. We just need to remember to pick it up again once the Qt bugs get resolved.

Removing from 2.10.

data-sync-user:
➤ Sarah Bird commented:

Going away with new glean.
Closed during bug squashing.

data-sync-user:
➤ Santiago Andrigo commented:

Moving back from Cancelled

PR #3270, opened 2022-04-05
Delete local storage on hard reset.

oskirby:
## Description

When issuing a hard reset through the developer menu, we should ensure that any data saved by the Glean local storage database should be purged. 

## Reference

Closes: #2231

## Checklist
    

bakulf:
I like this PR. Is it ready to go?

oskirby:
I mean to go test it on Windows, just to be sure that file-deletion semantics don't cause issue if the storage directory is still open, but then I got distracted and forgot about it :)

Otherwise, it's been tested on the UNIX-y platforms and it looks good.

oskirby:
Sadly, as expected, this does not work on Windows and we are unable to clear the offline storage due to a lingering file handle.

PR #3271, opened 2022-04-06
Add free trial strings

lesleyjanenorton:
This PR adds strings to support trial subscriptions.

### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -544,9 +544,6 @@ inAppAuth:
 
 freeTrials:
   startYourFreeTrial: 
-    value: "Sign up to start your %1-day free trial."
-    comment: "%1 is the number of free trial days"
-  freeTrialLabel: 
-    value: "%1-day free trial"
-    comment: "%1 is the number of free trial days"
-
+    value: Sign up to start your 7-day free trial.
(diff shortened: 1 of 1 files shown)


translations/strings.yaml
```diff
@@ -541,3 +541,12 @@ inAppAuth:
   termsOfServiceAndPrivacyDisclaimer:
     value: By continuing, you agree to the <a href="terms-of-service">Terms of Service</a> and the <a href="privacy-notice">Privacy Notice</a>.
     comment:  Do not remove or translate the markup '<a href="terms-of-service">', '<a href="privacy-notice">', '</a>'.
+
+freeTrials:
+  startYourFreeTrial: 
+    value: "Sign up to start your %1-day free trial."
```
flodolo, line 547: This really doesn't work for localization: we don't support plurals, and most locales will have to adapt `day` to the number. 

Do we already know how many days of free trial we want to support? My suggestion is to put the actual number, not a variable, otherwise we should try to get rid of the number alltogether.

`Sign up to start your 7-day free trial.` or `Sign up to start your free trial.`. The former will require a new string if we change the period. For that reason, we should get rid of the number of days in at least one of the strings (maybe the title).

flodolo (changes requested):

flodolo (approved):

PR #3272, opened 2022-04-06
modified layout to support long country names

lloan:
## Description

This PR addresses issues with long country names and city names after running a speed test. 

The fix makes long bolded labels wrap - by whole words, making sure we don't break words. Example: United States of America wraps at end of `of` instead of United States of Am-erica. See image below: 

image

Bug addressed can be seen in image below: 

image

## Reference

Jira Issue: https://mozilla-hub.atlassian.net/browse/VPN-2023


lesleyjanenorton (changes requested):

lesleyjanenorton (approved):
It works! thanks!

Issue #2693, opened 2022-02-08
VPN fails to initiate handshake


JKingZapata:
**Description**

After updating to 2.7, users are unable to connect to the VPN, when tapping the switch to connect it is stuck in "attempting to confirm connection" until it switches back to off position. From the VPN logs that have been shared, it keeps trying to initiate handshake 
```
[04.02.2022 17:12:46.473] Info: (windows - tunnel.dll) [TUN] [MozillaVPN] peer(c2if?MrBE) - Retrying handshake because we stopped hearing back after 15 seconds
[04.02.2022 17:12:46.473] Info: (windows - tunnel.dll) [TUN] [MozillaVPN] peer(c2if?MrBE) - Sending handshake initiation
```

**Technical Information**

Platform: Windows 10/11
FX product: Mozilla VPN 2.7

**Steps to Reproduce**
1. Purchase a VPN subscription
2. Install VPN on Windows
3. Flip switch to turn On VPN

**Actual Results**
- When tapping the switch to connect it is stuck on connecting status, then it goes to off

[mozillavpn-2022-2-4 (1).txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/8026316/mozillavpn-2022-2-4.1.txt)

![20220203_114444](https://user-images.githubusercontent.com/72564534/153046587-275edd78-07ed-4e54-afdb-62c55ed9aed4.jpg)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1698)

JKingZapata:
attaching another set of VPN logs for Windows 11 
[mozillavpn-2022-2-9.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/8036467/mozillavpn-2022-2-9.txt)

magsmoz:
It seems like a full reset helps for the moment, but as soon as the device goes into sleep mode and the user turns it back on the VPN won't connect again . I see this more often on Windows 11 than 10 
Here is another set of logs for Win11:
[mozillavpn-2022-2-7_5 (1).txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/8079725/mozillavpn-2022-2-7_5.1.txt)

data-sync-user:
➤ Joseph Anthony Cuevas commented:

regarding this bug
CS had a user giving the following feedback :
[...]

PR #3282, opened 2022-04-07
Windows: Ignore hidden interfaces for exclusion routes.

oskirby:
## Description

When handling the exclusion routes, Windows will sometimes report multiple default routes, including interfaces that may be inactive or hidden. If one of these interfaces is selected for the exclusion route, then this often results in a broken routing table and a failure to remain connected.

To resolve this properly, we must do two things:
- Rank duplicate routing table entries using their routing metric.
- Ignore routes on inactive interfaces. We utilize the `GetNetworkConnectivityHintForInterface` API call for this.

## Reference

Closes: #2693
JIRA Issue: [VPN-1698](https://mozilla-hub.atlassian.net/browse/VPN-1698)

## Checklist
    

oskirby:
@mkestler-mozilla if there is still time, we should try to cherry-pick this onto the next release.

bakulf (approved):

PR #3288, opened 2022-04-07
Android: Split the gradle monolith

strseb:
## Goals: 
Currently our /android folder is a template that qmake will use to compile a gradle monolith-project. This pr split's the app into it's modules and improves the developer-experience and making it finally testable!
## 1. Reworking the resource handling: 
 We have res/ and src/res, and a few dupes inbetween. i've cleaned that up so that we have resources/debug resources/release or resources/all for both modes

## 2. Split code into modules
Currently we have one big build.gradle and one big manifest. I've splitted them into smaller more manageble sub-projects: 
 - Daemon( Containing the vpn-service and all wireguard code)
 - QtBindings( handles the qmake added files)
 - src/ which is still the root project, containing the companion client-jni files for src/platforms/android 

This will allow us to add independent platform features pretty easily, and let's us write tests for them independently. 

## 3. Fix gradle project sync
As /android was never intended to be a project (just a template for qmake) - android studios gradle sync fails on this folder. This leaves dev's without any Linting, Code-Completion or Debugging support in android studio. 
As we now have a loose dependency tree here, we can mock qt-related stuff for android studio, so we can have: 
- Code Completion
- Dual debugging (c++ and kotlin!)
- Tests! ( we can now test each of the 3 modules in isolation, as they are now independend)
- Really nice linting hints from the android plugin. 

## 4. Ship ktlint via gradle. 
We now have ktlint as a dev-dep in gradle, so instead of getting it yourself you can do in /android 
```gradlew format``` :) - Also fixed a few lint things while at it, 

## 5. Qt5 Cleanup

Removed some hacks we did for qt5,
removed the leftover android qt5 docker images

## Reference

   none

## Checklist
    
### CODE REVIEW

File: android/buildSrc/src/main/kotlin/Dependencies.kt [hunk 1]
```diff
@@ -0,0 +1,14 @@
+object Dependencies {
```
bakulf, line 1: no header?

android/daemon/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
bakulf, line 1: can we move this into the root .gitignore file?

android/daemon/build.gradle
```diff
@@ -0,0 +1,92 @@
+buildscript {
```
bakulf, line 1: header?

bakulf (approved):
I have to admit this is big and hard to follow. I haven't found nothing totally wrong, so r+

PR #3290, opened 2022-04-07
Fix Wix schema error due to out-of-order Conditions.

oskirby:
## Description
During production code signing, we have encountered an XML schema error when trying to generate the installer package. This seems to be due to the `Condition` element being present in the wrong order. To fix this, we simply move it after the `Product` element, which is required to go first.

## Reference
See: #3236

## Checklist
    
### DIFF
--- a/windows/installer/MozillaVPN_prod.wxs
+++ b/windows/installer/MozillaVPN_prod.wxs
@@ -18,16 +18,17 @@
 		Version="!(bind.FileVersion.MozillaVPNExecutable)"
 		Manufacturer="Mozilla Corporation"
 		UpgradeCode="$(var.UpgradeCode)">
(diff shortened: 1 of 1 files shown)


oskirby:
This must be cherry-picked onto the release branch.

ahal (changes requested):
The repackage task still seems to fail with this change:
https://firefox-ci-tc.services.mozilla.com/tasks/Fj71qy5DSGKYX6k0Xpvtig

ahal (approved):
After editing the task's environment variables to point at this PR, looks like the task now succeeds:
https://firefox-ci-tc.services.mozilla.com/tasks/AKtBVZP5TnmwzRUBakV8FA

Thanks Owen!

PR #3295, opened 2022-04-11
initial commit for allure-hsitory

mozrokafor:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/.github/workflows/functional_tests.yaml
+++ b/.github/workflows/functional_tests.yaml
@@ -136,6 +136,20 @@ jobs:
           MVPN_API_BASE_URL: http://localhost:5000
           MVPN_BIN: ./build/mozillavpn
 
+      - name: Generate allure results
+        uses: actions/upload-artifact@master
+        with:
+          name: allure-results
+          path: allure-results
+          retention-days: 20
+
+      - name: Deploy report to github pages
+        if: always()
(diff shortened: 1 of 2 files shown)


.github/workflows/functional_tests.yaml
```diff
@@ -136,6 +136,20 @@ jobs:
...
           MVPN_BIN: ./build/mozillavpn
 
+      - name: Generate allure results
+        uses: actions/upload-artifact@master
+        with:
+          name: allure-results
+          path: allure-results
+          retention-days: 20
+
+      - name: Deploy report to github pages
```
strseb, line 146: This will would cause each task (so for each test file) to publish to gh-pages, which we're currently using for wasm and the inspector -  Probably overwriting our current deployments. Not sure whether this is what you want  😅

strseb (changes requested):

.github/workflows/functional_tests.yaml
strseb: I would suggest using the `destination_dir` to pick a unique subdir for the branch and have a 2nd task on merge that will clean that up

mozrokafor: lol thanks for looking at it, but no its not, still draft, pushing up some changes soon

Issue #3283, opened 2022-04-07
[Linux] App permissions not supported with cgroupsv2


data-sync-user:
**VPN version:**

* 2.8~rc82 (2.8~rc82-jammy1);

**Affected devices:**

* Ubuntu 22.04 - Jammy Jellyfish dev branch (Virtual machine);

**Prerequisites:**

* Have a subscription made for Mozilla VPN;

* Have Mozilla VPN installed;

* Have Mozilla VPN opened and OFF;

**Steps to reproduce:**

1. Go to Settings -> Network Settings;
1. Observe the listed options;

**Expected result:**

* The “App permissions” option is displayed;

**Actual result:**

* The “App permissions” option is missing;

**Additional notes:**

* see the attached screenshot;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2041)
┆author: Maria Berlinger

data-sync-user:
➤ Andrea Marchesini commented:

The latest versions of ubuntu do not support cgroup v1 anymore. We need to support cgroup2.

data-sync-user:
➤ Owen Kirby commented:

Unfortunately this will not be a trivial bug to solve, but I would like to try and add it in the 2.9 release.

In cgroupsv1, network traffic can be tagged with a classid, and we use a bit of nftables magic to match the classid, NAT-it and re-route it out the non-VPN interface.

In cgroupsv2, this network classifier is no longer supported and there exists no corresponding nftable rule that can be used to match a hierarchical cgroup path. It looks like we can workaround this behavior by writing an eBPF filter that marks outbound traffic in a way that nftables can act on it.

There is also an open question of how to match cgroups with application identifiers, and vice-versa.

data-sync-user:
➤ Owen Kirby commented:

Matching and rerouting packets based on cgroupsv2 is turning out to be a rather complicated task. There are a whole slew of packet matching options in the Linux kernel, but no one featureset is able to do what we need, and is incompatible with the others that do. In particular:

  * nftables has all the hooks we need to NAT and reroute packets, but it doesn’t support cgroupsv2.
  * eBPF is almost perfect for matching traffic that originates from a cgroupv2, but there’s no good hook point that we can used to do the NAT and rerouting. (the egress hook is too late in the output chain for re-routing).
[...]

PR #3303, opened 2022-04-11
Experiment: Support split tunneling with cgroups v2

oskirby:
## Description
This is an experiment to investigate how to pursue split tunnelling with Linux Control Groups v2. This presents some unique challenges because the support for network traffic classification was dropped in the transitions from Cgroups v1 to v2. And there doesn't appear to be a reliable mechanism to map between control group and application ID.

Roughly speaking, to resolve this issue, we need to accomplish the following tasks:
- [x] Track the creation and destruction of control groups, as they relate to user applications.
- [x] Associate control groups with their corresponding desktop application identifier.
- [x] Update netfilter.go to match traffic originating from cgroups v2 (xt_cgroup)
- [x] Tweak nftable rules to ensure that marked traffic is routed outside of the VPN when this feature is enabled.
- [x] Select v1 or v2 split tunnelling implementation depending on which cgroupfs exists.
- [x] Update packaging and test installation.

## Reference

See: #3283

## Checklist
    
### DIFF
--- a/src/platforms/linux/daemon/wireguardutilslinux.cpp
+++ b/src/platforms/linux/daemon/wireguardutilslinux.cpp
@@ -8,6 +8,7 @@
 #include "platforms/linux/linuxdependencies.h"
 
 #include <QHostAddress>
+#include <QFile>
 #include <QScopeGuard>
 
 #include <arpa/inet.h>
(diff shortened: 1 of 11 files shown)


src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -677,18 +689,66 @@ bool WireguardUtilsLinux::setupCgroupClass(const QString& path,
...
+    NetfilterResetCgroupV2(goCgroup);
   }
-  return m_cgroups + VPN_EXCLUDE_CGROUP;
 }
 
-QString WireguardUtilsLinux::getBlockCgroup() const {
-  if (m_cgroups.isNull()) {
-    return QString();
+void WireguardUtilsLinux::resetAllCgroups() {
+  logger.error() << "Permitting traffic from all cgroups";
```
bakulf, line 745: why error?

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -677,18 +689,66 @@ bool WireguardUtilsLinux::setupCgroupClass(const QString& path,
...
+  logger.error() << "Permitting traffic from" << cgroup;
+  if (m_cgroupVersion == 1) {
+    // Add all PIDs from the unified cgroup to the net_cls default cgroup.
+    moveCgroupProcs(m_cgroupUnified + cgroup, m_cgroupNetClass);
+  } else if (m_cgroupVersion == 2) {
+    QByteArray cgpath = cgroup.toLocal8Bit();
+    GoString goCgroup = {.p = cgpath.constData(),
+                         .n = (ptrdiff_t)cgpath.length()};
+    NetfilterResetCgroupV2(goCgroup);
   }
```
bakulf, line 741: do you want to add an `} else { unexpected cgroup version` ?

bakulf: or a Q_ASSERT(false)

src/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -677,18 +689,66 @@ bool WireguardUtilsLinux::setupCgroupClass(const QString& path,
...
 
-QString WireguardUtilsLinux::getExcludeCgroup() const {
-  if (m_cgroups.isNull()) {
-    return QString();
+// static
+bool WireguardUtilsLinux::moveCgroupProcs(const QString& src,
+                                          const QString& dest) {
+  QFile srcProcs(src + "/cgroup.procs");
+  FILE* fp = fopen(qPrintable(dest + "/cgroup.procs"), "w");
+  if (!fp) {
```
bakulf, line 697: why not a QFile?

bakulf (approved):

src/platforms/linux/daemon/wireguardutilslinux.cpp
oskirby: It turns out that when writing to the cgroupsfs pseudo-filesystem, the kernel is sensitive to system call, and you can only write a single PID per `write()` call. Unfortunately a `QFile` tends to batch up the file data and submits it to the kernel in a single system call - resulting in some processes failing to migrate between control groups.

src/platforms/linux/daemon/wireguardutilslinux.cpp
oskirby: There is also the possibility of `m_cgroupVersion == 0` indicating that the kernel doesn't support control groups at all.

PR #3304, opened 2022-04-12
Add postrm script to kill the mozillavpn instances when the linux package is uninstalled - fixes VPN-2048

bakulf:


linux/debian/mozillavpn.postrm
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+case "$1" in
+remove)
```
oskirby, line 3: It is perhaps food for thought about what do on other `postrm` events, such as an upgrade. But I think this is good to cover this particular issue.

oskirby (approved):

Issue #2712, opened 2022-02-10
Red line displayed when switching from single to multi hop


data-sync-user:
**VPN version:**

* 2.7.0 (16443620) - Android armv7 QT6 build;

**Platforms:**

* Android 9;

**Prerequisites:**

* Have and account with a subscription for Mozilla VPN;

**Steps to reproduce:**

1. Log into the Mozilla VPN app using the account from prerequisites;
1. From the Mozilla VPN main screen tap the Select location option;
1. On the Select location page, switch between single and multi-hop;

**Expected results:**

* The switch is made smoothly, without any UI issues;

**Actual results:**

* A red line displayed when switching from single to multi hop;

**Notes:**

* the issues reproduces for the DNS setting screen as well (Settings > Network settings > Advanced DNS Settings)
* see attached video;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1713)

data-sync-user:
➤ Lesley Norton commented:

Closing - the UI for this view changed in 2.9 and the line in question no longer exists.

PR #3309, opened 2022-04-12
VPN-1832: Improve visual cues from switching between single and multi-hop (and back)

MattLichtenstein:
## Description

- Creates two new components, `VPNSegmentedToggle` and `VPNSegmentedNavigation`
  - `VPNSegmentedToggle`: A multi-option segmented toggle control that allows the user to select from 2 or more options
  - `VPNSegmentedNavigation`: A navigation component, similar to `VPNTabNavigation`, that navigates screens based on selection of the embedded `VPNSegmentedToggle`
- Replaces the `VPNTabNavigation` component on the `ViewServers` screen with a `VPNSegmentedToggle`

https://user-images.githubusercontent.com/15353801/163046424-4401c5b2-ad97-494c-ae9f-08d05510ce7c.mov

## Reference

[VPN-1832: Improve visual cues from switching between single and multi-hop (and back)](https://mozilla-hub.atlassian.net/browse/VPN-1832)

## Checklist
    

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -75,7 +75,19 @@ Rectangle {
...
+
+                Accessible.role: Accessible.Button
+                Accessible.description: `${index + 1} of ${options.count}`
+                Accessible.name: {
+                    var name = ""
+                    if(root.selectedIndex === index) name += "selected,"
+                    name += label.text
+                    return name
+                }
+                Accessible.onPressAction: segmentClicked()
```
MattLichtenstein, line 87: Screen reader reads out the following: 

- Selected segment: "Selected, \<button text>, button, \<x> of \<y>" 
- Unselected segment: "\<button text>, button, \<x> of \<y>"

where `x` is the index of the segment, and `y` is the total number of segment options in the model. This reads out nearly the same way a native iOS UISegmentedControl reads (pausing is a little different).

Open to ideas if there is a better way to do this!

lesleyjanenorton:
Closes #2712

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -0,0 +1,133 @@
...
+            delegate: Item {
+                id: segment
+
+                Layout.fillWidth: true
+                Layout.fillHeight: true
+
+                objectName: segmentButtonId
+
+
+                Accessible.role: Accessible.Button
```
lesleyjanenorton, line 79: Since we're adding this, is there any reason not to just make the delegate a Button instead of an Item?

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -0,0 +1,133 @@
...
+                Accessible.description: `${index + 1} of ${options.count}`
```
lesleyjanenorton, line 80: We'll need to add this and a string for `Accessible.name` to `strings.yaml`.

nebula/ui/components/VPNSegmentedToggle.qml
lesleyjanenorton: Makes sense to me. The only other thing coming to mind is maybe changing the delegates to Buttons and adding them to a ButtonGroup.

nebula/ui/components.qrc
```diff
@@ -111,5 +111,7 @@
         <file>components/inAppAuth/VPNInAppAuthenticationLegalDisclaimer.qml</file>
         <file>components/inAppAuth/VPNInAppAuthenticationCancel.qml</file>
         <file>components/forms/VPNCursorDelegate.qml</file>
+        <file>components/VPNSegmentedToggle.qml</file>
```
lesleyjanenorton, line 114: ✨ **Segmented Toggle** ✨ ... I was wondering what this little piece of navigation should be called and segmented toggle seems to be exactly correct.

nebula/ui/components/VPNSegmentedNavigation.qml
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.15
+import QtQuick.Layouts 1.15
+
+import Mozilla.VPN 1.0
+
+Item {
```
lesleyjanenorton, line 10: Do we need this root `Item`?

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -0,0 +1,133 @@
...
+
+    onSelectedIndexChanged: {
+        selectedSegment = options.itemAt(selectedIndex)
+    }
+
+    Component.onCompleted: {
+        selectedSegment = options.itemAt(selectedIndex)
+    }
+
+    implicitHeight: 40
```
lesleyjanenorton, line 27: could be : `VPNTheme.theme.rowHeight`

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -0,0 +1,133 @@
...
+            rightMargin: 4
+            bottomMargin: 8
+        }
+
+        spacing: 0
+
+        Repeater {
+            id: options
+
+            delegate: Item {
```
lesleyjanenorton, line 70: We'll need to be sure these can be accessed via keyboard tabbing and may need to sync with @peterbenvenuto for guidance on what should happen visually when these have active focus.

lesleyjanenorton (changes requested):
A few questions and comments but over-all this is looking great!

nebula/ui/components/VPNSegmentedToggle.qml
MattLichtenstein: I think a ButtonGroup may be redundant here since the control already handles selecting a segment and preventing segments from being selected again while they are already selected.

Will definitely change the delegate to buttons though, since thats what they essentially are!

nebula/ui/components/VPNSegmentedNavigation.qml
MattLichtenstein: Nope!

translations/strings.yaml
```diff
@@ -541,3 +541,12 @@ inAppAuth:
   termsOfServiceAndPrivacyDisclaimer:
     value: By continuing, you agree to the <a href="terms-of-service">Terms of Service</a> and the <a href="privacy-notice">Privacy Notice</a>.
     comment:  Do not remove or translate the markup '<a href="terms-of-service">', '<a href="privacy-notice">', '</a>'.
+
+#Accessibility
+accessiblity:
```
flodolo, line 546:
```suggestion
accessibility:
```

Typo across the patch.

translations/strings.yaml
```diff
@@ -541,3 +541,12 @@ inAppAuth:
...
+  of:
+    value: of
```
flodolo, line 548: As a translator, I really wouldn't know how to translate this:
- We need a more descriptive ID that provides context about the string (and to prevent reuse in other situations). In this specific case, translating "of" is not enough, you need to expose the whole sentence (`x of y`?).
- Same for the next string. "Selected" needs to be tied to one element, as adjectives need to be declined (gender, number).

flodolo (changes requested):

nebula/ui/components/VPNSegmentedToggle.qml
```diff
@@ -0,0 +1,131 @@
...
+
+                Layout.preferredWidth: slider.implicitWidth
+                Layout.fillHeight: true
+
+                objectName: segmentButtonId
+
+                Accessible.description: `${index + 1} ${VPNl18n.AccessiblityOf} ${options.count}`
+                Accessible.name: {
+                    var name = ""
+                    if(root.selectedIndex === index) name += `${VPNl18n.AccessiblitySelected}.`
```
flodolo, line 82: If I understand the code (which I'm not sure about), these two strings should be exposed for localization in their entirety, i.e. 
- `%1 of %2`
- `%1 selected`

Hardcoding characters or imposing a specific order (e.g. noun followed by adjective) is generally a [red flag for localization](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#avoid-concatenations-use-placeholders-instead).


I'm still not sure what the referenced noun is (what is selected? 1 of 2 of?), but we should add that to comments in strings.yaml

P.S. I think the linter should have checks for strings.yaml, but there should be an empty line at the end of the file.

translations/strings.yaml
MattLichtenstein: oh wow 🤦 will fix

nebula/ui/components/VPNSegmentedToggle.qml
MattLichtenstein: I have made some changes based on your suggestions, but please let me know if and how it could be improved further!

translations/strings.yaml
```diff
@@ -541,3 +541,19 @@ inAppAuth:
...
+accessibility:
+  currentIndexFocusedOfTotalItemsInGroup:
+    value: "%1 of %2"
+    comment:
+      - Used by screen readers to explain to users which item they are currently focused on, and how many items there are total in a group.
+      - For example, given a list with 15 items. If a user selects the 3rd item, we want the screen reader to read out "3 of 15".
+      - The %1 represents the index of the currently focused item by the screen reader.
+      - The %2 represents the total number of items in the group.
+  selectedAndItemName:
+    value: "Selected, %1"
```
flodolo, line 555: As mentioned, this is tricky, because you'd need to know what it references (city, country, something else completely, given this looks reusable). 

Could we add a noun, e.g. `Selected item: %1`, so that languages can safely match the gender?

flodolo (commented):
This looks a lot better, only one more possible change/improvement

translations/strings.yaml
MattLichtenstein: Hmmm, I suppose we could do that. 

Just to clarify and make sure it is actually necessary, the word "Selected" here is basically it's own sentence (and we can make the comma a period to reflect that if that makes more sense). If we look at the example in the video from the PR summary, if single-hop is currently selected, then the screen reader should read out "Selected. Single hop. Button. 1 of 2". If "Selected" is it's own sentence, then does the rest of the context still matter?

Sorry for all the back and forth, as you can tell, I am not well versed in other languages 😄

flodolo: It's not important if "selected" is contained in a stand-alone sentence.

Practical example in Italian:
- If referred to a button (`pulsante`), you need to decline adjective as male (selected = selezionat**o**).
- If referred to an image (`immagine`), you need to decline it as female (selected = selezionat**a**).

Most of the major languages need this, English is the exception.

You could use male, and hope that it doesn't create confusion. But having an actual noun removes any risk.

BTW, in your example, where does the translation for "Button" come from? From the OS?

MattLichtenstein: The only hesitation I have with adding "item" after "selected" is that in a future case (not in the example in this PR), we may have components where more than one item can be selected, and I am afraid that saying "selected item" may sound like only one item can be selected, and refers to the item as part of a group, whereas just "selected" ignores the idea that the item is a part of a group, and only focuses on whether the current item is selected or not. Does that make sense?

Also, you are correct, "button" comes from and is translated by the OS

flodolo: I have bad news: adjectives need to be declined for gender and number. So, that scenario with multiple selected items makes things worse, also considering we don't support plurals :-( 

If we need to support that, we need two strings: `Selected item: %1` and `Selected items: %1`

MattLichtenstein: Haha, I guess we can worry about that when the time comes. I have changed the string to "Selected item: %1". Let me know if there is anything else.

Thanks for all of the insight!

lesleyjanenorton (approved):
🎉 this looks great- thank you!

flodolo (approved):

PR #3332, opened 2022-04-13
Remove unused environment variables from .taskcluster.yml

ahal:


strseb:
I've added those back, as it seems the windows-generic-worker need them 😅

ahal:
> @ahal I've added those back, as it seems the windows-generic-worker needs them 😅

Oh, weird.. Is it possible your branch wasn't based on top of the revision that removed them? Do you have a link to the error somewhere?

strseb:
See https://firefox-ci-tc.services.mozilla.com/tasks/S_A87Za5ROWmm297BvksUQ/runs/0/logs/public/logs/live.log

ahal:
Ah, that's good since we fixed that in a newer version of Taskgraph. I'll bump the version of Taskgraph being used in this PR as well and then that error won't happen anymore :)

ahal:
Lol, and sorry.. I totally didn't even realize the Decision task failed ><

ahal:
Ok that should fix it.

strseb (approved):

PR #3351, opened 2022-04-19
Pass iap free trial duration to qml

strseb:
## Description

This PR fetches the in-app subscription free trial info from playstore and passes it to qml. 
Given this is a bit hard to test, you need to use a signed build on prod. - When the dev menu is enabled, it will mock the 12-month plan with 7 days, otherwise it will return the values from google play (all 0 currently)  - so qa can test the ui flow for now, we might need a better way to test this. 

Screenshot_20220419-182839

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
+++ b/android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
@@ -29,6 +29,7 @@ import kotlinx.serialization.json.Json
 import org.mozilla.firefox.vpn.glean.GleanEvent
 import org.mozilla.firefox.vpn.qt.VPNUtils
 import java.text.NumberFormat
+import java.time.Duration
 import java.util.Currency
 import kotlin.collections.ArrayList
 import kotlin.collections.HashMap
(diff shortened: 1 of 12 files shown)


lesleyjanenorton (approved):

android/src/org/mozilla/firefox/vpn/InAppPurchase.kt
```diff
@@ -344,8 +346,25 @@ class InAppPurchase private constructor(ctx: Context) :
...
         val monthlyPriceString = formatter.format(monthlyPrice)
+
+        // freeTrialPeriod is a ISO 8601 duration i.e P7D == 7 days
+        val trialDays = if (monthCount == 12 && VPNUtils.isDevMode) {
+            // TODO: This is just some debugging stuff
+            // We need to be on prod so, for testing let's
+            // mock the PlayStore response for 12month if we're in isDevMode
+            // so QA can test when we go live
+            val duration = Duration.parse("P7D")
+            duration.toDays().toInt()
```
MattLichtenstein, line 357: I understand this is for QA to use for testing, but If a user happens to have dev options enabled and no free trial available for the 12mo option, and this option still appears, could this be misleading? Or do we just assume a user shouldn't have dev options enabled?

MattLichtenstein (approved):
One small question, otherwise looks good!

PR #3359, opened 2022-04-21
VPN-1839: Dinamically change status bar text color on mobile

brizental:
## Description

The default text color is black for this text. Contrary to what I
thought, iOS does not automatically verify background color and
change the text accordingly. Status bar text color needs to explicitly be
changed when necessary.

This is a draft because I still need to:

- [x] Stop using deprecated APIs on the iOS implementation
- [ ] ~Write tests~
   - Skipped for this functionality due to the lack of structue for adding platform specific tests at the moment.

I would like to get some early feedback on the approach I took here,
especially on how I handled this on the QML side, since this is my first PR on the project.
I thought it would be best to double check.

https://user-images.githubusercontent.com/25176023/164515154-2f296ad5-95b3-4228-b4b5-c0902815f451.mov

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-1839

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNMobileStatusBarModifier.qml
+++ b/nebula/ui/components/VPNMobileStatusBarModifier.qml
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
+
+import Mozilla.VPN 1.0
+
+QtObject {
+    id: statusBarModifier
+
+    enum StatusBarTextColor {
+        Light,
+        Dark
+    }
+    property int statusBarTextColor: VPNMobileStatusBarModifier.StatusBarTextColor.Dark
+
+    function applyChanges() {
+        VPN.setStatusBarTextColor(statusBarModifier.statusBarTextColor);
+    }
+
+    function resetDefaults() {
+        VPN.setStatusBarTextColor(VPNMobileStatusBarModifier.StatusBarTextColor.Dark);
+    }
+
+    Component.onCompleted: applyChanges()
+    Component.onDestruction: resetDefaults()
+}
--- a/nebula/ui/components/VPNMobileStatusBarModifier.qml
+++ b/nebula/ui/components/VPNMobileStatusBarModifier.qml
@@ -5,22 +5,19 @@
 import QtQuick 2.5
 
 import Mozilla.VPN 1.0
+import components 0.1
 
 QtObject {
     id: statusBarModifier
 
-    enum StatusBarTextColor {
-        Light,
-        Dark
-    }
-    property int statusBarTextColor: VPNMobileStatusBarModifier.StatusBarTextColor.Dark
+    property int statusBarTextColor: VPNTheme.StatusBarTextColorDark
 
     function applyChanges() {
-        VPN.setStatusBarTextColor(statusBarModifier.statusBarTextColor);
+        VPNTheme.setStatusBarTextColor(statusBarModifier.statusBarTextColor);
     }
 
     function resetDefaults() {
-        VPN.setStatusBarTextColor(VPNMobileStatusBarModifier.StatusBarTextColor.Dark);
+        VPNTheme.setStatusBarTextColor(VPNTheme.StatusBarTextColorDark);
     }
 
     Component.onCompleted: applyChanges()
--- a/src/mozillavpn.h
+++ b/src/mozillavpn.h
@@ -195,6 +195,7 @@ class MozillaVPN final : public QObject {
   Q_INVOKABLE void crashTest();
   Q_INVOKABLE void deleteAccount();
   Q_INVOKABLE void cancelAccountDeletion();
+  Q_INVOKABLE void setStatusBarTextColor(const quint8 color);
 #ifdef MVPN_ANDROID
   Q_INVOKABLE void launchPlayStore();
 #endif
--- a/src/mozillavpn.h
+++ b/src/mozillavpn.h
@@ -195,7 +195,6 @@ class MozillaVPN final : public QObject {
   Q_INVOKABLE void crashTest();
   Q_INVOKABLE void deleteAccount();
   Q_INVOKABLE void cancelAccountDeletion();
-  Q_INVOKABLE void setStatusBarTextColor(const quint8 color);
 #ifdef MVPN_ANDROID
   Q_INVOKABLE void launchPlayStore();
 #endif
--- a/src/platforms/ios/iosutils.h
+++ b/src/platforms/ios/iosutils.h
@@ -18,6 +18,8 @@ class IOSUtils final {
   static void shareLogs(const QString& logs);
 
   static int compareStrings(const QString& a, const QString& b);
+
+  static void setStatusBarTextColor(const quint8 color);
 };
 
 #endif  // IOSUTILS_H
--- a/src/platforms/ios/iosutils.h
+++ b/src/platforms/ios/iosutils.h
@@ -21,7 +21,7 @@ class IOSUtils final {
 
   static int compareStrings(const QString& a, const QString& b);
 
-  static void setStatusBarTextColor(const Theme::StatusBarTextColor color);
+  static void setStatusBarTextColor(Theme::StatusBarTextColor color);
 };
 
 #endif  // IOSUTILS_H
--- a/src/platforms/ios/iosutils.h
+++ b/src/platforms/ios/iosutils.h
@@ -5,6 +5,8 @@
 #ifndef IOSUTILS_H
 #define IOSUTILS_H
 
+#include "theme.h"
+
 #include <QString>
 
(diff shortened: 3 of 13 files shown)


src/platforms/ios/iosutils.mm
```diff
@@ -113,3 +113,12 @@
       return -1;
   }
 }
+
+// static
+void IOSUtils::setStatusBarTextColor(quint8 color) {
+  if (color == 0) {
+    [UIApplication sharedApplication].statusBarStyle = UIStatusBarStyleLightContent;
+  } else {
+    [UIApplication sharedApplication].statusBarStyle = UIStatusBarStyleDarkContent;
```
brizental, line 122: This is the deprecated APIs I mentioned on the PR description. They were deprecated on iOs 13.

src/ui/views/ViewMobileOnboarding.qml
```diff
@@ -23,6 +23,10 @@ VPNFlickable {
     height: parent.height
     interactive: flickContentHeight > height
 
+    VPNMobileStatusBarModifier {
+        statusBarTextColor: VPNMobileStatusBarModifier.StatusBarTextColor.Light
+    }
```
brizental, line 28: On the Jira ticket, it says this page is the only one that has issues with bad contrast and I believed it. I did not manually check all pages.

lesleyjanenorton: So happy you picked this up! The Jira ticket didn't lie but what if we move this into `main.qml` so that it can be accessed from anywhere once we enable theming/dark mode? Or maybe we don't actually need the `VPNMobileStatusBarModifier` component and we just call `VPN.setStatusBarTextColor(...)` directly as necessary?

nebula/ui/components/VPNMobileStatusBarModifier.qml
```diff
@@ -0,0 +1,28 @@
...
+
+import Mozilla.VPN 1.0
+
+QtObject {
+    id: statusBarModifier
+
+    enum StatusBarTextColor {
+        Light,
+        Dark
+    }
```
bakulf, line 15: If the enum is in the QML file, C++ needs to do what you have implemented: a quint8. And basically, we end up doing a mapping (int -> enum value). I wonder if it's cleaner to add an enum in C++ in `src/theme.h` and use the QT macro `Q_ENUM`. In this way, here we can do:

`property var statusBarTextColor: VPNTheme.StatusBarDarkTextColor` (or any better name).

We then propagate the VPNTheme.<NameOfTheEnum>.

nebula/ui/components/VPNMobileStatusBarModifier.qml
```diff
@@ -0,0 +1,28 @@
...
+    property int statusBarTextColor: VPNMobileStatusBarModifier.StatusBarTextColor.Dark
+
+    function applyChanges() {
+        VPN.setStatusBarTextColor(statusBarModifier.statusBarTextColor);
```
bakulf, line 19: What about if we move this method into VPNTheme (theme.cpp/h)?

nebula/ui/components/VPNMobileStatusBarModifier.qml
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    function resetDefaults() {
+        VPN.setStatusBarTextColor(VPNMobileStatusBarModifier.StatusBarTextColor.Dark);
+    }
+
+    Component.onCompleted: applyChanges()
```
bakulf, line 26: Maybe we can set the status-bar color in the CTOR of theme.cpp. And reset it when the property changes.

brizental: I implemented this component to overwrite whatever is the default status bar color. For the case described in the Jira ticket we want the text color to be light on onboarding pages and dark in other pages. That is why I chose to apply changes when the component is created and reset them when it's destroyed.

When there is a dark theme, it would be light in both and yes we would probably need to set that on the constructor of theme.cpp or main.qml. However, as I understand it, this is a different case.

src/ui/views/ViewMobileOnboarding.qml
brizental: So, IIUC this component should not be tied to the theme. I wrote it with the purpose to overwrite whatever is the theme default color for pages where a specific status bar text color is needed. That is why I only include it in this specific page, it is always going to have a dark background regardless of theme so it should reliably always have a light status bar text color.

The reason why I added the `VPNMobileStatusBarModifier` abstraction is because it includes the logic to apply the changes once the component is created and reset them once it is destroyed.

Let me know if the reasoning is incorrect :D

nebula/ui/components/VPNMobileStatusBarModifier.qml
bakulf: This one. Which is called on completed.

nebula/ui/components/VPNMobileStatusBarModifier.qml
bakulf: makes sense!

src/mozillavpn.h
```diff
@@ -195,6 +195,7 @@ class MozillaVPN final : public QObject {
   Q_INVOKABLE void crashTest();
   Q_INVOKABLE void deleteAccount();
   Q_INVOKABLE void cancelAccountDeletion();
+  Q_INVOKABLE void setStatusBarTextColor(const quint8 color);
```
bakulf, line 198: Let's move these into the VPNTheme object (theme.cpp/h)

src/platforms/ios/iosutils.mm
```diff
@@ -115,11 +116,16 @@
...
+@property (nonatomic, assign) UIStatusBarStyle preferredStatusBarStyle;
+@end
+
 void IOSUtils::setStatusBarTextColor(const Theme::StatusBarTextColor color) {
-  if (color == Theme::StatusBarTextColorLight) {
-    [UIApplication sharedApplication].statusBarStyle = UIStatusBarStyleLightContent;
-  } else {
-    [UIApplication sharedApplication].statusBarStyle = UIStatusBarStyleDarkContent;
-  }
+    StatusBarModifierViewController* rootViewController = static_cast<StatusBarModifierViewController *>([[UIApplication sharedApplication].windows[0] rootViewController]);
```
brizental, line 124: Not sure if this static_cast is too much of a hack. This is the most straight forward way I could find to have a setter for `preferredStatusBarStyle`. I am totally new to Objective-C so any input is appreciated :)

src/platforms/ios/iosutils.mm
```diff
@@ -12,6 +12,7 @@
 #include <QtGui>
 #include <QtGui/qpa/qplatformnativeinterface.h>
 #include <QtQuick>
+#include <iostream>
```
brizental, line 15: Reminder to self: Remove this unused include.

brizental:
Hey folks, this is now ready for a final review. I did not add Android changes because this is an iOS only problem. On Android the status bar is always behing a dark background.

I spoke to @lesleyjanenorton about a bug I found on this implementation. Because I decided to apply changes when a given component is created and reset them when it is destroyed, the status bar color remains white if the user navigates from the onboarding page to the "Get Help" page. See video below.


https://user-images.githubusercontent.com/25176023/165277645-17941ec2-0f8a-444d-8a8e-54f46e9997f4.mov

Lesley mentioned that the work she is currently doing should fix this issue by making sure the Onboarding view is actually destroyed when users navigate to Get Help from it, so I believe the changes can be merged as is as we test it again once Lesley's changes are merged.

src/platforms/ios/iosutils.h
```diff
@@ -18,6 +20,8 @@ class IOSUtils final {
   static void shareLogs(const QString& logs);
 
   static int compareStrings(const QString& a, const QString& b);
+
+  static void setStatusBarTextColor(const Theme::StatusBarTextColor color);
```
bakulf, line 24: Not need to be `const`

src/platforms/ios/iosutils.mm
```diff
@@ -4,13 +4,15 @@
 
 #include "iosutils.h"
 #include "logger.h"
+#include "theme.h"
```
bakulf, line 7: alphabetic order here.

src/platforms/ios/iosutils.mm
bakulf: No, it's actually correct. static_cast happens a compilation time.

src/theme.cpp
```diff
@@ -144,3 +148,10 @@ QVariant Theme::data(const QModelIndex& index, int role) const {
       return QVariant();
   }
 }
+
+void Theme::setStatusBarTextColor(
+    [[maybe_unused]] const StatusBarTextColor color) {
```
bakulf, line 153: Nice! We often use a macro for this:
`Q_UNUSED(color);`

The code would be:
```
#ifdef MVPN_IOS
  ...
#else
  Q_UNUSED(color)
#endif
```

But I like your solution too!

bakulf (approved):

lesleyjanenorton (approved):

PR #3360, opened 2022-04-21
Let's force openGL QT rendering on windows

bakulf:
This PR works on my computer(tm), but I would like to chat with other engineers to see if this approach is OK for real.
Reading https://doc.qt.io/qt-6/windows-graphics.html it seems that we do not need to include openGL dll files.

### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -15,6 +15,7 @@
 #include "simplenetworkmanager.h"
 
 #ifdef MVPN_WINDOWS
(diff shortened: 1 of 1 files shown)


strseb:
@bakulf - Should we close this ? As we now have dx11 detection in place?

oskirby:
I think this should be closed. The DX11 check should resolve the issue, and falling back to OpenGL doesn't work anyways due to missing DLLs on my affected machine.

PR #3365, opened 2022-04-22
Add and update account deletion views

flozia:
## Description

This PR adds and updates the view that are necessary for the account deletion flow, which enables the user to delete their FXA account from within the VPN.

## Reference

- Jira: [VPN-2064](https://mozilla-hub.atlassian.net/browse/VPN-2064)
- Figma spec: [Subscription Management In-App](https://www.figma.com/file/CIelhzmFVyDqn6gNvMXolc/Subscription-Management-In-App?node-id=798%3A2732)

## Checklist
    

nebula/ui/components/VPNCheckBoxRow.qml
```diff
@@ -22,7 +22,7 @@ RowLayout {
 
     signal clicked()
 
-    spacing: VPNTheme.theme.windowMargin
+    spacing: VPNTheme.theme.windowMargin * 0.5
```
flozia, line 25: The spacing looked a bit wide for me — also when comparing it to the designs.

src/ui/deleteAccount/ViewDeleteAccount.qml
```diff
@@ -39,30 +71,39 @@ Item {
             when: VPNAuthInApp.state === VPNAuthInApp.StateStart
             PropertyChanges {
                 target: loader
-                source: "ERROR ERROR! TODO"
+                source: "" // TODO: Error! Handle this case.
```
flozia, line 74: Have to check with UX what we would like to do in case we hit a state that indicates an error during the account deletion flow.

src/ui/deleteAccount/ViewDeleteAccountRequest.qml
```diff
@@ -7,25 +7,111 @@ import QtQuick.Layouts 1.14
...
+VPNInAppAuthenticationBase {
+    id: viewDeleteAccount
 
-    Text {
-        id: msg
-        text: "A" + VPNAuthInApp.attachedClients
-        anchors.top: parent.top
+    property var checkboxData: [
+        {
+            labelText: "Paid subscriptions", // TODO: Needs content
```
flozia, line 18: This needs to be localized if we want to keep the checkbox headers: UX and content are taking a look at this.

src/ui/settings/ViewSettingsMenu.qml
```diff
@@ -138,15 +138,17 @@ VPNFlickable {
             onClicked: settingsStackView.push(aboutUsComponent)
         }
 
-        VPNButton {
-            text: "TODO delete account"
-            anchors.top: vpnPanel.bottom
-            anchors.topMargin: VPNTheme.theme.vSpacing
-            anchors.horizontalCenter: parent.horizontalCenter
+        // TODO: Move to subscription management
```
flozia, line 141: Temporary placement until we added the “Subscription Management” view.

src/ui/deleteAccount/ViewDeleteAccount.qml
```diff
@@ -11,13 +11,41 @@ import components.forms 0.1
...
+        onStatusChanged: {
+            if (loader.status !== Loader.Ready) {
+                return;
+            }
+            
+            if (
+                prevState === ""
+                && viewDeleteAccount.state === "StateInitializing"
+            ) {
+                // `StateInitializing` is set for the first time
```
bakulf, line 34: I don't understand this status check. This function. Why do we need it? Let's talk on slack/zoom

src/ui/deleteAccount/ViewDeleteAccount.qml
bakulf: No needs to do it. This should never happen.

src/ui/deleteAccount/ViewDeleteAccount.qml
flozia: As discussed `isDeleteAccountAuth` is now set by default for all views.

src/ui/deleteAccount/ViewDeleteAccount.qml
flozia: OK — in that case I’ll remove the states entirely.

src/ui/deleteAccount/ViewDeleteAccount.qml
```diff
@@ -11,13 +11,40 @@ import components.forms 0.1
...
+                VPN.requestDeleteAccount();
+            } else if (
+                prevState !== ""
+                && viewDeleteAccount.state === "StateInitializing"
+            ) {
+                // `StateInitializing` is set during the account deletion flow,
+                // which means the user aborted the operation
+                settingsStackView.pop();
```
bakulf, line 43: can we add this where we call `VPN.cancelAccountDeletion()`?

src/ui/deleteAccount/ViewDeleteAccount.qml
```diff
@@ -80,25 +103,22 @@ Item {
...
+            )
             PropertyChanges {
                 target: loader
                 source: "../authenticationInApp/ViewAuthenticationVerificationSessionByTotpNeeded.qml"
             }
         },
 
-        State {
-            name: "StateFallbackInBrowser"
-            when: VPNAuthInApp.state === VPNAuthInApp.StateFallbackInBrowser
```
bakulf, line 92: Maybe add a comment about these remaining states and why we do not cover them here.

flozia: Added a comment above the `states` property.

src/ui/deleteAccount/ViewDeleteAccount.qml
flozia: Created a dedicated function `cancelAccountDeletion` that is being called from the account deletion flow templates when the user cancels the operation.

bakulf (approved):

flodolo (approved):
Do we support account creation directly from VPN? This content feels like something that really belongs to FxA more than a product, because we will end up duplicating it everywhere.

flozia:
> Do we support account creation directly from VPN? This content feels like something that really belongs to FxA more than a product, because we will end up duplicating it everywhere.

Yes we do! In `2.8` we introduced authentication and account creation in-app. This PR enables users to delete their FxA account as well from within the VPN. As far as I now this content is indeed taken from the FxA account deletion flow. Should we reconsider introducing the strings with this PR (I could use placeholders for now) or can I go ahead merging it @flodolo & @bakulf?

flodolo:
> Should we reconsider introducing the strings with this PR (I could use placeholders for now) or can I go ahead and merge it @flodolo & @bakulf?

That's a question for @bakulf 

As I said, my concern is we start doing this in every single project, it's going to end up with a lot of duplication (and maintenance concerns).

bakulf:
This is out-of-scope here, but I agree with @flod. This PR is ok as it is and it can land. About reducing the number of dup strings,
we are working on a new build system where the strings are handled differently. We can probably "steal" strings and translations from other projects in the future.

src/ui/authenticationInApp/ViewAuthenticationVerificationSessionByEmailNeeded.qml
```diff
@@ -58,7 +64,13 @@ VPNInAppAuthenticationBase {
 
         VPNCancelButton {
             anchors.horizontalCenter: parent.horizontalCenter
-            onClicked: VPN.cancelAuthentication()
+            onClicked: {
```
lesleyjanenorton, line 67: Would be nice if we could consolidate these `if (isDeleteAccountAuth) { cancelAccountDeletion() }.. else { VPN.cancelAuthentication() }` somehow.

src/ui/authenticationInApp/ViewAuthenticationSignIn.qml
```diff
@@ -10,31 +10,20 @@ import components 0.1
...
+        if (isDeleteAccountAuth) {
+            cancelAccountDeletion();
+        } else {
+            VPNAuthInApp.reset();
+        }
+    }
     _menuButtonAccessibleName: qsTrId("vpn.main.back")
     _headlineText: VPNAuthInApp.emailAddress
-    _subtitleText: VPNl18n.InAppAuthSignInSubtitle
+    _subtitleText: !isDeleteAccountAuth
```
lesleyjanenorton, line 24: We're doing the `isDeleteAccountAuth` quite a few times in this file and I wonder if it might feel more manageable to do it all in one place with something like:

```
states:[
  State {
    when: isDeleteAccountAuth
    PropertyChanges {...}

**Review by lesleyjanenorton — APPROVED** [review]
Looks good! A few questions/comments but nothing blocking. ✨

File: src/ui/authenticationInApp/ViewAuthenticationSignIn.qml (hunk 9 above)
Comment (flozia) [reply src/ui/authenticationInApp/ViewAuthenticationSignIn.qml]: Good suggestion — done!

File: src/ui/authenticationInApp/ViewAuthenticationSignIn.qml [hunk 10]
```diff
@@ -44,11 +33,18 @@ VPNInAppAuthenticationBase {
...
          }
-        _buttonText: VPNl18n.InAppAuthSignInButton
+        _buttonText: !isDeleteAccountAuth
+            ? VPNl18n.InAppAuthSignInButton
+            : VPNl18n.DeleteAccountAuthButtonLabel
         _inputPlaceholderText: VPNl18n.InAppAuthPasswordInputPlaceholder
     }
 
-    _disclaimers: VPNInAppAuthenticationLegalDisclaimer {}
+    _disclaimers: Loader {
```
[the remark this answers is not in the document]
flozia: Otherwise it’s not possible to conditionally set a component for the `_disclaimers` property.

PR #3371, opened 2022-04-25
Ensure `taskgraph` can be generated locally

ahal:
When working on Taskcluster CI changes it's useful to be able to debug locally by running the `taskgraph` command. However, a couple of exceptions have crept into this process lately necessitating workarounds. This ensure we can run `taskgraph <subcommand>` without errors.

Ideally we should set up some tests to cover this, but I haven't had a chance to work on those yet.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -111,3 +111,6 @@ jobs:
       - name: Test release-push
         run: |
           taskgraph full --p taskcluster/test/params/release-push.yml           
+      - name: Test default parameters
+        run: |
+          taskgraph full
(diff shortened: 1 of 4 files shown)


strseb:
Re testing: We have a small linting test that tests taskgraph runs with the test params, it should be easy to extend this to `taskgraph <subcommand>` :) 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/7f9b101021bf8de6840ed0e91d483b7c77761479/.github/workflows/linters.yaml#L93

ahal:
Neat, thanks for the pointer! Added.

strseb (approved):

PR #3375, opened 2022-04-26
WIP: Use the Glean-Android SDK on instead of glean.js on android

strseb:
## Description

We currently have the issue that glean pings might not be send on mobile as the app is not long enough in the foreground. 
So instead of manually creating background-tasks etc, let's use the android-glean SDK as the glean team already solved this🥳 

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-1729

## Checklist
    
- [c:] I have added thorough tests where needed


android/AndroidManifest.xml
```diff
@@ -31,7 +31,6 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
         <activity
             android:configChanges="orientation|uiMode|screenLayout|screenSize|smallestScreenSize|layoutDirection|locale|fontScale|keyboard|keyboardHidden|navigation|mcc|mnc|density"
             android:name="org.mozilla.firefox.vpn.qt.VPNActivity"
-            android:process=":QtOnlyProcess"
```
strseb, line 34: Note: glean expects to live on the main process - which is fine, we just don't want qt and WireGuard on the same, so let's move the daemon into its own process.

android/src/org/mozilla/firefox/vpn/qt/GleanUtil.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+    @JvmStatic
+    fun setGleanUploadEnabled(upload: Boolean) {
+        Glean.setUploadEnabled(upload)
+    }
+    @JvmStatic
+    fun setGleanSourceTag(tags: String = "") {
+        // val tag_list = tag.split(",")
+        // Glean.setSourceTags(tag_list);
+        // TODO: Glean-Android does not support this, yet.
```
brizental, line 43: I filed a bug about it, currently this APi is only available for debugging in Android which I believe does not make sense. See https://bugzilla.mozilla.org/show_bug.cgi?id=1766457

android/build.gradle
```diff
@@ -34,6 +44,18 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlinx-serialization'
 
+// Optionally, set any parameters to send to the plugin.
+try{
+    if(qtAndroidDir){
+        ext.gleanYamlFiles = [
+                "$rootDir/../../../glean/metrics.yaml",
+                "$rootDir/../../../glean/pings.yaml"
```
brizental, line 52: All metrics currently have 

```
 send_in_pings:
      - main
```

This means they will not be added to either the default `events` ping or `metrics` ping. You can change that by adding `- default` to the list of pings there. I am not sure this will address the problem of sending things in the background though, see my main review comment :)

brizental (commented):
> So instead of manually creating background-tasks etc, let's use the android-glean SDK as the glean team already solved this🥳

I think it's important to note that the Glean Android SDK does not create background tasks for sending pings.

It will send, out of the box the `baseline`, `metrics` and `events` ping (and the `deletion-request` ping which I will overlook for the current comment because that one already is sent by Glean.js in the same way as by other SDKs). Also, as I mentioned in one of my inline comments, with the current implementation, the `metrics` and `events` pings won't be sent at all.

Assuming we change the implementation to send these pings, they still won't be sent when the app is in background. The `baseline` ping is sent when the app goes from background to foreground and vice-versa. The Glean SDK will attempt to send the `metrics` ping once per day, but only when the app is actually active. And the `events` ping is sent when the app goes to background or when a given limit of events is hit. More details on: https://mozilla.github.io/glean/book/user/pings/sent-by-glean.html

The other thing I believe is worth flagging is that, by following this approach and using the Glean Android SDK default pings, we will end up with data sent in a completely different schedule only on Android. Which will make it harder to analyze the data side-by-side with other platforms.

badboy:
> > So instead of manually creating background-tasks etc, let's use the android-glean SDK as the glean team already solved this🥳
> 
> I think it's important to note that the Glean Android SDK does not create background tasks for sending pings.

It is correct that the Glean SDK does not keep a background task around to act on the app's behalf. 
For ping upload we rely on the work manager, so that the actual upload process is running separate from the main application (and can even run and finish when the main application is closed).

> It will send, out of the box the `baseline`, `metrics` and `events` ping (and the `deletion-request` ping which I will overlook for the current comment because that one already is sent by Glean.js in the same way as by other SDKs). Also, as I mentioned in one of my inline comments, with the current implementation, the `metrics` and `events` pings won't be sent at all.
> 
> Assuming we change the implementation to send these pings, they still won't be sent when the app is in background. The `baseline` ping is sent when the app goes from background to foreground and vice-versa. The Glean SDK will attempt to send the `metrics` ping once per day, but only when the app is actually active. And the `events` ping is sent when the app goes to background or when a given limit of events is hit. More details on: [mozilla.github.io/glean/book/user/pings/sent-by-glean.html](https://mozilla.github.io/glean/book/user/pings/sent-by-glean.html)
> 
> The other thing I believe is worth flagging is that, by following this approach and using the Glean Android SDK default pings, we will end up with data sent in a completely different schedule only on Android. Which will make it harder to analyze the data side-by-side with other platforms.
> 
> [EDIT] Flagging @badboy for a double check.

This is correct, to summarize:
* `baseline` is sent when putting the app into foreground or background.
* `metrics` is sent on a fixed schedule: 4am if the app is running or next start thereafter.
* `events` on putting the app to background or when reaching the event limit (currently 500)

If there are no events or metrics for the `metrics` ping around the pings won't get send. The `baseline` ping will always be sent.

If another scheduling is needed, especially to match up better with other applications a custom ping is the way to go.

(Note: I have not given this PR itself a look yet, but Bea is proficient with the SDK so my look is not necessary. I'm still around to help out of course and plan to give this a proper look this week)

strseb:
Hey - Thanks for the detailed feedback !

> we will end up with data sent in a completely different schedule only on Android.

This is the problem I'm trying to address. - As far as I understand the current qt-glean interactions, [we're sending the main ping every 30 minutes](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/052f6434aef287f48407f2d93c034eb638cd25ad/src/mozillavpn.cpp#L1105), while in foreground. However, when the qt-client is killed (i.e. the app was too long in the background) the timer is then reset, currently resulting in a weird schedule on android and iOS. 

The proposed solution was to send open metrics when the app goes into the background. So I guess adding the metrics to the "events" ping might be a right way forward?

>>So instead of manually creating background-tasks etc, let's use the android-glean SDK as the glean team already solved this🥳
>I think it's important to note that the Glean Android SDK does not create background tasks for sending pings.

Sorry, that was phrased super badly, I was referring to the use of the WorkManager for uploading. 😅

brizental:
> > we will end up with data sent in a completely different schedule only on Android.
> 
> This is the problem I'm trying to address. - As far as I understand the current qt-glean interactions, [we're sending the main ping every 30 minutes](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/052f6434aef287f48407f2d93c034eb638cd25ad/src/mozillavpn.cpp#L1105), while in foreground. However, when the qt-client is killed (i.e. the app was too long in the background) the timer is then reset, currently resulting in a weird schedule on android and iOS.
>
> The proposed solution was to send open metrics when the app goes into the background. So I guess adding the metrics to the "events" ping might be a right way forward?

I think this problem description points to the conclusion that the current data collection design and scheduling is not fit to answer the questions we want to in the platforms we support. Because of the cross-platform nature of VPN, simply defaulting to the pings Glean sends out of the box to address the issue might not be the best solution. 

My suggestion would be to take a step back and take a look at what questions we want to answer with the data we are collecting and what are the platform limitations for sending that data and then defining pings and their schedules in a way that will dodge most of the issues we are having. 

Also, it's worth noting, that just by using the Glean Android/iOS SDKs we are already sending different data from mobile. Even if all the rest is the same, there will still be the `baseline` data that will be different. I am mostly sure it even requires that we do a new data review for the inclusion of the SDK alone. Just to be clear, I am not saying we should not do it, the fact that these SDKs have a better infrastructure to send pings reliably from these platforms is indeed compelling, I am just saying it's something to keep in mind :)

birdsarah:
Thanks everyone for the detailed discussion.

Here's my thoughts based on the discussion so far.

We're already using a custom ping `main` and we already send it on our own schedule. To my knowledge, other than `deletionRequest`, we don't send any other pings. Happy to be corrected.

I am not understanding why we can't do what @strseb has proposed in this PR with the change that we'll continue not using any out of the box pings and just use main, like we do in our JS implementation. We would then use whatever built in mechanism android has to wake and send pings as needed.

While the app is foregrounded and all processes are running then all events will be sent to the android glean process and they'll be sent along.

However, if I understand correctly, when the app is back-grounded, the qt process will go dormant and none of the code that generates event will be running so waking up the android glean process to send events on the schedule of our choosing will have little value as there will be no events.

But:
* once we're in the background, at least once every x hours, i'd like to get a ping that says whether the vpn is on or off and, if on, how long has it been on for. There's a lot of fudge factor here because we don't currently have a usage metric like this and I'm happy to iterate. We know we want a usage metric, but we haven't specified anything yet.
* this background ping will mean that we can get a much better sense of daily active clients and monthly active clients.
* I'll note that working on a ping like this will also likely improve desktop data too because there's no guarantee that the desktop clients will send a ping if no events have been generated and if the connection has been completely stable then I don't think any event is generated.
* One additional class of events I'd like us to think about is the current connectionHealth events. We [turn the connectionHealth events off when we background the android and iOS apps](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/afee48ee2769f18f59c93bcabe412e82935754fe/src/connectionhealth.cpp#L237). Can we bring the connectionHealth code to the background android process task? Perhaps, tying into a previous point, we can send a "connection still stable" event as a first proxy for a usage indicator that helps give us a positive signal on whether user's are having a solid experience rather than only surfacing the negative experience.

cc @oskirby 

One last question are we already running an "on background event" task to send all outstanding events. If not then we wouldn't need to implement as the android glean process would pick up these events to send on its first background wake-up.

Other note: I am guessing that this will have the effect of resetting the client ids for all the android clients. But I'm fine with that. We don't have good android data at the moment. This is part of the journey and I don't think we need to sweat it.

birdsarah:
Made some edits in long comment above.

brizental:
> We're already using a custom ping main and we already send it on our own schedule. To my knowledge, other than deletionRequest, we don't send any other pings. Happy to be corrected.

This is correct. My point is that it seems the current schedule of this custom ping is not really fit to answer the questions we have, given the platform limitations we are dealing with. There are multiple ways we can come up to deal with this e.g. we can have different pings that serve different purposes and are sent in different schedules. I am hesitant to give any ideas here because I have not investigated the issues we have that much.

> -  once we're in the background, at least once every x hours, I'd like to get a ping that says whether the vpn is on or off and, if on, how long has it been on for. There's a lot of fudge factor here because we don't currently have a usage metric like this and I'm happy to iterate. We know we want a usage metric, but we haven't specified anything yet.

This is a good use case for creating a different lighter ping that is sent in a more frequent schedule than the `main` ping.

> - this main-sent-while-backgrounded ping will mean that we can get a much better sense of daily active clients and monthly active clients. yes, this is analagous to baseline ping, but as @brizental said it's useful to have comparable data across platforms so lets double down on our existing main ping and send as we need.

Just a quick note that the `baseline` ping is not sent while the app is backgrounded. It is sent immediatelly when the app goes to background or foreground, but Glean does not wake the app up to send a `baseline` ping periodically.

> - I'll note that working on a ping like this will also likely improve desktop data too because there's no guarantee that the desktop clients will send a ping if no events have been generated and if the connection has been completely stable then I don't think any event is generated.

I agree.

> One last question are we already running an "on background event" task to send all outstanding events. If not then we wouldn't need to implement as the android glean process would pick up these events to send on its first background wake-up.

This is already true. Glean persists metrics until they are submitted in a ping. Right now even if we don't send all events in a single run, once the app is woken up and a ping is submitted it will contain outstanding from previous runs metrics.

> Other note: I am guessing that this will have the effect of resetting the client ids for all the android clients. But I'm fine with that. We don't have good android data at the moment. This is part of the journey and I don't think we need to sweat it (famous last words!).

Yeap, thanks for noting this. Indeed changing to the Glean Android SDK will reset all client_ids. We should also think about cleaning the QML local storage on first run after the update that contains this change, because that is the storage used by Glean.js and if we don't clean it up we will leave all the Gean data lying around in users devices. It should not be a lot of data, but it's worth taking this into consideration.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
```diff
@@ -0,0 +1,134 @@
...
+            putBoolean("glean_enabled", upload)
+            apply()
+        }
+        Glean.setUploadEnabled(upload)
+    }
+
+    fun setGleanSourceTag(tags: String = "") {
+        // val tag_list = tag.split(",")
+        // Glean.setSourceTags(tag_list);
+        // TODO: Glean-Android does not support this, yet.
```
brizental, line 58: This is now supported from [v44.2.0](https://github.com/mozilla/glean/blob/main/CHANGELOG.md#v4420-2022-05-16) on.

strseb: Nice! Thank you, so I guess we need to wait for android-components to pick that new version up. - 
https://github.com/mozilla-mobile/android-components/blob/main/buildSrc/src/main/java/Dependencies.kt#L35

brizental: /o\ Yeap, I guess we do.

badboy: Right. That requires us updating it in m-c, however we're preparing a major update for next week (v50) and so I won't get to a 44.2.0 update there.
Are you already depending on other Android Components? If not then you can depend on the Glean release directly using `org.mozilla.telemetry:glean:44.2.0`. That way you don't need to wait for the (sometimes lengthy) a-c release cycles.

src/platforms/android/androidglean.cpp
```diff
@@ -0,0 +1,133 @@
...
+  QJsonDocument doc(args);
+  logger.debug() << " gleanEnabledChanged" << enabled;
+  AndroidVPNActivity::instance()->sendToService(
+      ServiceAction::ACTION_GLEAN_ENABLED_CHANGED,
+      doc.toJson(QJsonDocument::Compact));
+}
+
+void AndroidGlean::daemonConnected() {
+  // Daemon is now ready
+  gleanEnabledChanged(SettingsHolder::instance()->gleanEnabled());
```
brizental, line 117: Does this mean Glean will be started with `uploadEnabled=false` and only when the deamon is connected the upload status will be changed?

brizental (commented):
A few notes:

1. I believe a bit of migration code would be nice to clean up the Glean.js storage of clients that are moved from Glean.js to Glean Android SDK.
2. The data coming from the Android SDK will land in different tables than the Glean.js data, are we alright with that? [On Android the app id is extracted from the application context](https://mozilla.github.io/glean/book/reference/general/initializing.html#gleaninitializeconfiguration), in this case Android data would land on the table `org_mozilla_firefox_vpn.<ping-name>`, worth double checking the package name there just in case. This also means that, before this lands we need to redo step 4. of the [Glean integration checklist](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/index.html#glean-integration-checklist) and register the new app id, otherwise the data will go to the error stream.
3. Since the Android SDK collects new data and sends new pings (`baseline`), we need to request a data review for adding it to our project. See step 2. of the integrations checklist.

Observation: I am not an Android expert in any means, so I am focusing this review only on the Glean integration bit.

strseb:
>I believe a bit of migration code would be nice to clean up the Glean.js storage of clients that are moved from Glean.js to Glean Android SDK.

Good Point! I've updated the pr to do that :) 

>The data coming from the Android SDK will land in different tables than the Glean.js data, are we alright with that? [On Android the app id is extracted from the application context](https://mozilla.github.io/glean/book/reference/general/initializing.html#gleaninitializeconfiguration), in this case Android data would land on the table org_mozilla_firefox_vpn.<ping-name>, worth double checking the package name there just in case. 

Personally, I would think this sounds nice. @birdsarah - knows probably better if that is fine.  

>This also means that, before this lands we need to redo step 4. of the [Glean integration checklist](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/index.html#glean-integration-checklist) and register the new app id, otherwise the data will go to the error stream.
    Since the Android SDK collects new data and sends new pings (baseline), we need to request a data review for adding it to our project. See step 2. of the integrations checklist.

That is a very good callout, I will do that when we're ready to move forward with this pr :)

birdsarah:
@strseb can we meet next week to discuss this PR. We need to see it land, but we also need to be able to use the generated data.

e.g. "The data coming from the Android SDK will land in different tables than the Glean.js data, are we alright with that?" Yes, we're alright with it because I don't see a choice. But we will probably need to update almost all our dashboards and I'd like to make sure that we're not making data incomparable.

src/commands/commandui.cpp
```diff
@@ -189,13 +190,17 @@ int CommandUI::run(QStringList& tokens) {
...
     Lottie::initialize(engine, QString(NetworkManager::userAgent()));
     Nebula::Initialize(engine);
     L18nStrings::initialize();
 
     MozillaVPN vpn;
     vpn.setStartMinimized(minimizedOption.m_set);
 
+#ifdef MVPN_ANDROID
+    AndroidGlean::initialize();
+#endif
```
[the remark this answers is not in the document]
brizental: Sounds good, is there a ticket filed for this refactor work?

brizental (approved):
From a Glean integration stand point this looks correct. It's best that someone more familiar with our Android / Qt integration looks at this before merging though. I do not feel confortable reviewing that.

I am also missing tests for the Glean collection, however I think it might be best to do that in a separate PR since this is one is already large enough. Would mind filing a ticket for that @strseb ?

Note that we cannot move forward with this without the data-review. It can be done in a comment in this PR directly. We _can_ merge without registering the new application id, however until that is done all Android data will go directly to the error stream.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
```diff
@@ -0,0 +1,134 @@
+package org.mozilla.firefox.vpn.daemon
```
bakulf, line 1: Headers

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
```diff
@@ -0,0 +1,134 @@
...
+        val gleanEnabled = Prefs.get(mParent).getBoolean("glean_enabled", false) // Don't send telemetry unless explicitly asked to.
+        val channel = if (mParent.packageName.endsWith(".debug")) {
+            "staging"
+        } else {
+            "production"
+        }
+        val conf = Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, channel, 500, HttpURLConnectionUploader())
+        val build = BuildInfo("versionCode", "VersionName", Calendar.getInstance())
+        Glean.registerPings(Pings)
+        Log.e("ANDROID-GLEAN", "initializeGlean on:$gleanEnabled, ch:$channel ")
```
bakulf, line 36: extra space before the final "

src/inspector/inspectorhandler.cpp
```diff
@@ -758,6 +762,19 @@ static QList<InspectorCommand> s_commands{
           obj["value"] = countryArray;
           return obj;
         }},
+#ifdef MVPN_ANDROID
+    InspectorCommand{"dm", "Send a request to the Daemon {type} {args}", 2,
```
bakulf, line 766: can we have a better name than 'dm'?

bakulf: maybe with 'android' in the name?

src/platforms/android/androidglean.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <QFile>
+
+namespace {
+AndroidGlean* s_instance = nullptr;
+Logger logger(LOG_ANDROID, "AndroidGlean");
+
+}  // namespace
+
+// static
+void AndroidGlean::initialize(QQmlEngine* e) {
```
bakulf, line 35: s/e/engine/ ?

src/platforms/android/androidglean.cpp
```diff
@@ -0,0 +1,138 @@
...
+  if (!s_instance) {
+    Q_ASSERT(qApp);
+    s_instance = new AndroidGlean(qApp);
+  }
+  auto glean_db_path = e->offlineStorageDatabaseFilePath("Glean") + ".sqlite";
+  auto gleanDB = QFileInfo(glean_db_path);
+  if (gleanDB.exists()) {
+    QFile::remove(glean_db_path);
```
bakulf, line 43: why do we remove it?

src/platforms/android/androidglean.h
```diff
@@ -0,0 +1,36 @@
...
+  void setGleanSourceTags(const QStringList& tags);
+
+  void applicationStateChanged(Qt::ApplicationState state);
+
+  void daemonConnected();
+
+  static void initialize(QQmlEngine* e);
+
+ private:
+  AndroidGlean(QObject* parent);
```
bakulf, line 32: explicit

src/ui/main.qml
```diff
@@ -350,7 +348,7 @@ Window {
         }
 
         function onRecordGleanEvent(sample) {
-            console.debug("recording Glean event");
+            console.debug("recording Glean event" + sample);
```
bakulf, line 351: s/+/,/ or ``

src/ui/settings/ViewAppPermissions.qml
```diff
@@ -36,9 +34,10 @@ Item {
         Component.onCompleted: {
             console.log("Component ready");
             VPNAppPermissions.requestApplist();
-            Sample.appPermissionsViewOpened.record();
+            VPN.recordGleanEvent("appPermissionsViewOpened");
+
             if (!vpnIsOff) {
-                Sample.appPermissionsViewWarning.record();
```
bakulf, line 41: Can we enforce this? so that we do not use `Sample` by mistake?

src/ui/states/StateSubscriptionBlocked.qml
```diff
@@ -37,6 +37,6 @@ VPNStackView {
             getHelpLinkVisible: false
             }
         );
-        Sample.subscriptionBlockedViewed.record();
```
bakulf, line 40: Do we need to remove `import org.mozilla.Glean 0.30` ?

bakulf (changes requested):

src/platforms/android/androidglean.cpp
strseb: As that is only used by glean.js, it's just wasting disk space for users who upgrade from 2.8 -> 2.9

src/ui/settings/ViewAppPermissions.qml
strseb: Not right now. I've filed https://mozilla-hub.atlassian.net/browse/VPN-2390 - To refactor how we load glean.js. If we were to use it similar as lottie.js, we could remove the global qml sample.

src/commands/commandui.cpp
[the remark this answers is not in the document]
strseb: yes :) https://mozilla-hub.atlassian.net/browse/VPN-2390

src/platforms/android/androidglean.cpp
bakulf: OK write a comment.

src/platforms/android/androidglean.cpp
```diff
@@ -0,0 +1,139 @@
...
+  // Daemon is now ready
+  gleanUploadEnabledChanged(SettingsHolder::instance()->gleanEnabled());
+}
+
+void AndroidGlean::applicationStateChanged(Qt::ApplicationState state) {
+  switch (state) {
+    case Qt::ApplicationState::ApplicationActive:
+      break;
+
+    case Qt::ApplicationState::ApplicationSuspended:
```
bakulf, line 131: `[[fallthrough]];`

src/platforms/android/androidglean.cpp
```diff
@@ -0,0 +1,139 @@
...
+    case Qt::ApplicationState::ApplicationInactive:
```
bakulf, line 132: here too

bakulf (approved):

PR #3382, opened 2022-04-27
Enforce empty line at the end of strings.yaml

flodolo:
No empty line means that the last string changes every time we add new content at the end of the file.

### DIFF
--- a/scripts/utils/generate_strings.py
+++ b/scripts/utils/generate_strings.py
@@ -56,6 +56,11 @@ def parseTranslationStrings():
         exit("Unable to find translations/strings.yaml")
 
     with open(yaml_path, "r", encoding="utf-8") as yaml_file:
+        # Enforce a new line at the end of the file
+        last_line = yaml_file.readlines()[-1]
+        if last_line == last_line.rstrip():
+            exit("The yaml file must have an empty line at the end")
+
         yaml_content = yaml.load(yaml_file, UniqueKeyLoader)
 
         if yaml_content is not None:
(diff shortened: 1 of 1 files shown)


scripts/utils/generate_strings.py
```diff
@@ -56,6 +56,11 @@ def parseTranslationStrings():
         exit("Unable to find translations/strings.yaml")
 
     with open(yaml_path, "r", encoding="utf-8") as yaml_file:
+        # Enforce a new line at the end of the file
```
flodolo, line 59: This is the actual change.

scripts/utils/generate_strings.py
```diff
@@ -133,7 +138,9 @@ def parseGuideStrings():
     for guide_filename in os.listdir(guides_path):
         if not guide_filename.endswith(".json"):
             continue
-        with open(os.path.join(guides_path, guide_filename), "r", encoding="utf-8") as guide_file:
+        with open(
```
flodolo, line 141: This is just reformatting from black (in a separate commit)

bakulf (approved):

bakulf (changes requested):

scripts/utils/generate_strings.py
flodolo: Bah, I even thought of checking that, and then forgot 🤦🏼 

What about using `.seek()`. Trying with that, but happy to create two read cycles instead.

bakulf: `.seek()` works too.

bakulf (approved):

PR #3386, opened 2022-04-27
Enable mac signing task

ahal:
Adds a `mac-signing` task that runs after the Taskcluster MacOS build.

### DIFF
--- a/taskcluster/ci/build/macos.yml
+++ b/taskcluster/ci/build/macos.yml
@@ -23,10 +23,8 @@ macos/opt:
         taskcluster-proxy: true
         chain-of-trust: true
         max-run-time: 3600
-        artifacts:
-            - type: directory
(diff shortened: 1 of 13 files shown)


hneiva:
@ahal the worker is up and running now, should be able to test dep signing.

ahal:
@strseb btw we do have level 1 signing available here. The tasks currently aren't configured to run on PRs (since I figured it might just be adding noise), but we can configure them to if you think it would be useful. Otherwise you can either test by pushing to the `staging` repo, or by temporarily adding `github-pull-request` to the `run_on_tasks_for` attribute in your PR.

ahal:
Actually @hneiva and I were chatting and he has one additional fix on top of this over in #3534, so doing to close this in favour of that.

PR #3392, opened 2022-04-27
Fix binding loops in mobile onboarding

lesleyjanenorton:
## Description

    Fixes binding loops in mobile onboarding panel. This should go out with the Free Trials updates.

## Checklist
    
### DIFF
--- a/src/ui/views/ViewMobileOnboarding.qml
+++ b/src/ui/views/ViewMobileOnboarding.qml
@@ -19,7 +19,7 @@ VPNFlickable {
     property real panelHeight: window.safeContentHeight
     property bool shouldRestoreSlide: false
 
-    flickContentHeight: window.safeContentHeight / 2 + col.implicitHeight - spacerBottom.height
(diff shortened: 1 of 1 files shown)


src/ui/views/ViewMobileOnboarding.qml
```diff
@@ -260,29 +260,31 @@ VPNFlickable {
             Layout.minimumHeight: VPNTheme.theme.windowMargin
         }
 
-        ColumnLayout {
+        Column {
```
MattLichtenstein, line 263: Yeaa, I had a feeling making this a layout would break something.

giphy

I was having trouble getting the spacing to match the spec with a positioner, but didn't notice the binding loop. I think this looks good nonetheless!

MattLichtenstein (approved):

src/ui/views/ViewMobileOnboarding.qml
lesleyjanenorton: haha nw it's too easy to do! I'd love to find a more resilient approach to responsive QML layouts.

PR #3398, opened 2022-04-28
Guess graphic API for windows

bakulf:


src/commands/commandgraphicguess.cpp
```diff
@@ -0,0 +1,93 @@
...
+                      const QString& message) {
+  if (message.contains("0x80070057")) {
+    ++s_vertexShaderErrorCount;
+  }
+}
+
+}  // namespace
+
+CommandGraphicGuess::CommandGraphicGuess(QObject* parent)
+    : Command(parent, "gg", "Guess the graphic configuration") {
```
oskirby, line 31: I don't particularly like the name `gg` for this test. I would recommend changing it to something more like `gpucheck` which I feel has a stronger association with running a test to get a result.

Also, "gg" is old internet slang for good game - something you would say to someone when the game is over. And I hope user's aren't over with our product :)

src/commands/commandgraphicguess.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+CommandGraphicGuess::~CommandGraphicGuess() {
+  MVPN_COUNT_DTOR(CommandGraphicGuess);
+}
+
+int CommandGraphicGuess::run(QStringList& tokens) {
+  Q_ASSERT(!tokens.isEmpty());
+
+  // Let's take the location before creating QGuiApplication.
+  QString location =
```
oskirby, line 43: In terms of standards paths, this will result in `%AppData%\Roaming` and the output of this test is one of the few artifacts that is more appropriate to go into `%AppData%\Local`, which is for files that should *not* be migrated across machines.

bakulf: I would like to keep this new file next to the setting file. In this way we do not spread config files in random locations.

oskirby (approved):
Looks good

flodolo:
@bakulf 

This looks very weird in the export…

```
  <file original="../src/ui/gpucheck/main.qml" datatype="plaintext" source-language="en" target-language="en-US">
    <body>
      <group restype="x-trolltech-linguist-context" resname="main">
        <trans-unit id="_msg1">
        <source>Checking the graphical card</source>
        </trans-unit>
      </group>
    </body>
  </file>
````

I'm particularly concerned about the group, because that's not something we have.

Any way you could change this code, and verify how the xliff looks like? Maybe move the string to the YAML file?

flodolo:
Alternative thought: is this string displayed to user? Who's going to see it?

PR #3399, opened 2022-04-28
Maybe fix appstore connect build in 2.8.2?

strseb:
@mbirghan , @bakulf - It seems on the release branch we are not currently building ios in appstore connect...

Looking at the logs
```
==> Running `brew cleanup go`...

Disable this behaviour by setting HOMEBREW_NO_INSTALL_CLEANUP.

Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).

/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin/qmake

/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin/qmake: line 7: /Volumes/workspace/repository/qt_ios/6.2.3/macos/bin/qmake: No such file or directory

/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin/qmake: line 7: /Volumes/workspace/repository/qt_ios/6.2.3/macos/bin/qmake: No such file or directory

This script compiles MozillaVPN for MacOS/iOS

Using the MVPN_IOS_ADJUST_TOKEN value for the adjust token

qmake path: /Volumes/workspace/repository/qt_ios/6.2.3/ios/bin/qmake

qmake doesn't exist or it fails
```

It seems like ios/bin/qmake does expect the destop qmake not under `clang_64` but to be in the macos folder. Not sure if this will fix it? But i do not understand how this would have worked for 2.8.0 then 😅

### DIFF
--- a/ci_scripts/ci_post_clone.sh
+++ b/ci_scripts/ci_post_clone.sh
@@ -32,9 +32,9 @@ else
   pip3 install aqtinstall
   aqt install-qt -O /Volumes/workspace/repository/qt_ios mac desktop 6.2.3 -m qtcharts qtwebsockets qt5compat
   aqt install-qt -O /Volumes/workspace/repository/qt_ios mac ios 6.2.3 -m qtcharts qtwebsockets qt5compat
-  mv /Volumes/workspace/repository/qt_ios/6.2.3/macos /Volumes/workspace/repository/qt_ios/6.2.3/clang_64
+ # mv /Volumes/workspace/repository/qt_ios/6.2.3/macos /Volumes/workspace/repository/qt_ios/6.2.3/clang_64
   export QT_IOS_BIN=/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin
-  export PATH=/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin:/Volumes/workspace/repository/qt_ios/6.2.3/clang_64/bin:$PATH
+  export PATH=/Volumes/workspace/repository/qt_ios/6.2.3/ios/bin:/Volumes/workspace/repository/qt_ios/6.2.3/macos/bin:$PATH
 fi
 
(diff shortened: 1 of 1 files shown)


mbirghan:
I don't know why but this is already fixed on `main` but not on the release branch.
Don't know if this patch is needed or if we just have to cherry pick a patch.

mbirghan:
#3252 is fixing this. 
@bakulf is this something that we can put onto the release branch or should we create a separate patch?

bakulf:
I would like to know why changing the `macos` folder to `clang64` fixes the issue. Can you tell me more?

mbirghan:
Ok so this explanation is a little bit longer.

TL;DR: iOS `qmake` references MacOS `qmake`. That reference changed with the new Qt version.

Qt5 had two different `qmake` files for iOS and MacOS. So if you installed only one of those two platforms everything worked as expected. 
In Qt6 there was a change so that the iOS `qmake` in now referencing the MacOS `qmake` using a hardcoded path. So far so good (or bad).

If you install Qt6 using the online installer it automatically installs the MacOS files even if you only want to install iOS. This is good since iOS wouldn't work otherwise. Now this isn't the case if we use `aqt`.

`aqt` actually only installs the platform we specify which is why we have to install the MacOS files separately. The issue with that is that previously (for Qt 6.2.1 or so), the iOS files did not reference the MacOS files at `.../macos/...` but at `.../clang64/...`. Since this is hardcoded into the iOS `qmake` at installation (at least as far as I could determine) we had to move the MacOS files to the correct folder. 

Now it seems like something in the iOS installation has changed (maybe for 6.2.3) and the iOS `qmake` is now referencing `.../macos/...` so we had to take out the line that moves the MacOS files.

bakulf (approved):

mbirghan (approved):

Issue #2850, opened 2022-02-24
Unstable connection and No internet connection error displayed when connecting to VPN


data-sync-user:
**VPN version:**

* 2.7.1 (2.202202232320);

**Affected Platforms:**

* Windows 10;

**Prerequisites:**

* Have Mozilla VPN installed on your device;
* Have one FxA account with a VPN subscription;

**Steps to reproduce:**

1. Log into the app using the account from prerequisites;
1. Connect to the VPN using different locations;

**Expected result:**

* The user is connected to the VPN without any issues;

**Actual result:**

* The Unstable connection message is briefly displayed each time, sometimes accompanied by the No Internet Connection error;

**NOTE**: Please see attached video and logs.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1823)

data-sync-user:
➤ Rebecca Billings commented:

Luminita Buruian would you please confirm in the Actual results if this fixes itself after a few seconds?

data-sync-user:
➤ Luminita Buruian commented:

Rebecca Billings Done.

data-sync-user:
➤ Sebastian Streich commented:

can repro, will look into it 🙂

data-sync-user:
➤ Luminita Buruian commented:

The issue no longer rproduces on Windows 10 build 2.9.0 (2.202207061818) - signed RC build

Closing this issue and will reopen/log another one if this reproduces again later on.

CC: Valentina Virlics Bianca Hidecuti

PR #3412, opened 2022-05-02
Use whole sliding-window when calulating Ping-Loss

strseb:
## Description
Currently, we're calculating loss only considering the send-packages from our sliding window.
When estimating the stability, we want to enforce a max loss of 10% of the last 32 packages. 
This leads to a weird situation when we have sent less than 32 packages.
 i.e that if any of the first 9 packages gets lost, the loss is above 10% and switch servers, which is probably not what we want for one lost packet. 

On Windows, the first ICMP package after creating the wintun device is failing very reliably, therefore we're seeing https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2850. 
 
## Reference

   https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2850

## Checklist
    

bakulf:
I prefer @oskirby reviewing this.

oskirby (approved):
I am okay with these changes, since it just makes the math a bit more conservative when computing ping loss.

However, as a solution to the initial connection instability on Windows, it feels like we are hacking around the real problem, which is that it seems to take windows an additional 5-10 seconds before it can actually route anything.

I am in favour of merging this patch.

Issue #3394, opened 2022-04-27
Installer hangs on Windows 10 1909.00 and older


data-sync-user:
While fixing the handshake initiation error with an active-but-disconnected Wi-Fi interface, we added a call to `GetNetworkConnectivityHint` to determine if the interface was suitable for use as a default route before sending any traffic its way. Unfortunately this API call requires a minimum windows build of 2004.00 (19041), and will cause the VPN executable to throw an error at startup.

This manifests at installation time by hanging the installer as the service fails to start.

**Affected Platforms:**

* Windows 10 build 1909.00 and earlier

**Steps to Reproduce:**

* Download the 2.8.0 installer from archive.mozilla.org
* Attempt to install the VPN client

**Expected Result:**

* Installation succeeds, and the VPN client is launched to the sign-in view.

**Actual Result:**

* The installer hangs, and eventually reports an error of `Service "Mozilla VPN (broker)" (MozillaVPNBroker) failed to start. Verify that you have sufficient privileges to start system services.`

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2135)
┆author: Owen Kirby

data-sync-user:
➤ Valentina Virlics commented:

Tried to verify this on my Windows 10 19043.1645 build, but while using Mozilla VPN 2.8, I was not able to see the error. I am able to install the client with success. Owen Kirby Should I check something specific? Maybe in the Inspector? Thanks!

data-sync-user:
➤ Owen Kirby commented:

Hey Valentina, you will need to test with an older version of Windows, the 19043.1645 build you are using corresponds to Windows 10 Version 21H1 (May 2021 Update)

But this bug only manifests on Windows 10 version 1909.00 (builds 10.0.18363.2212) and older, which should be included in our Mozilla VPN supported platforms ( https://docs.google.com/spreadsheets/d/1vaKIfTLIpmAfnBor3ZTzhOMzmoqhVEA-6tb9Ugf4dDE/edit#gid=0 )

data-sync-user:
➤ Valentina Virlics commented:

When I read the issue, I’ve understood that is 1909.00 and older. My bad.

This would be difficult, but will try to find a device with correct version.

data-sync-user:
➤ Valentina Virlics commented:

We could not yet find a colleague with a device that has requested configuration.

Moving this into Blocked.

data-sync-user:
[...]

PR #3418, opened 2022-05-02
Windows: Replace connectivity hint with GetIpInterfaceTable()

oskirby:
## Description
In the 2.8.0 release train, we encountered an issue where the VPN client would fail to launch on Windows 10 versions 1909.00 and older due to a new dependency introduced on the `GetNetworkConnectivityHint` system call. Unfortunately this would mean dropping support for some LTSC versions of windows that may still be out in the wild.

This patch attempts to work around this system call by inspecting the IP interface table, using the `GetIpInterfaceTable` system call, which is supported from Windows Vista and onwards. This is also slightly more correct, and we now respect the routing policy if users opt to disable IPv6 on one or more interfaces.

## Reference

JIRA ticket [VPN-2135](https://mozilla-hub.atlassian.net/browse/VPN-2135)
Github issue #3394
Related issue #2693

## Checklist
    
### DIFF
--- a/src/platforms/windows/daemon/windowsroutemonitor.cpp
+++ b/src/platforms/windows/daemon/windowsroutemonitor.cpp
@@ -6,6 +6,8 @@
 #include "leakdetector.h"
 #include "logger.h"
 
+#include <QScopeGuard>
+
 namespace {
 Logger logger(LOG_WINDOWS, "WindowsRouteMonitor");
 };  // namespace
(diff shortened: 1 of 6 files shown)


src/platforms/windows/daemon/windowsroutemonitor.cpp
```diff
@@ -68,6 +68,45 @@ WindowsRouteMonitor::~WindowsRouteMonitor() {
...
+      logger.debug() << "Interface" << row->InterfaceIndex << "is valid for IPv4 routing";
+      m_validInterfacesIpv4.append(row->InterfaceLuid.Value);
+    }
+    if (row->Family == AF_INET6) {
+      logger.debug() << "Interface" << row->InterfaceIndex << "is valid for IPv6 routing";
+      m_validInterfacesIpv6.append(row->InterfaceLuid.Value);
+    }
+  }
+
+  FreeMibTable(table);
```
bakulf, line 107: To be 100% sure we do not add a random `return` in this function, what about if we use a `qScopeGuard()` for this free?

bakulf (approved):

PR #3421, opened 2022-05-03
Upgrade Taskgraph to 1.5.0

ahal:
This includes a fix which should help fix the cleanup error in Windows
builds.

PR #3422, opened 2022-05-03
Use hashIndex for Openssl

strseb:
## Description
I kinda introduced a chicken and egg situation into our ci. 
Creating tasks with index-routes require level-3. Pr's are level-1 
When we now have a pr that invalidates our index-search (i.e a taskgraph update like #3421) - it thus creates a level-1 task that now has a route, which is rejected. 

The hashIndex transform solves this. It auto-removes routes on level-1 tasks. Also it will make sure we will re-compile openssl when the build script changes. We're using this for qt already.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/build-openssl/kind.yml
+++ b/taskcluster/ci/build-openssl/kind.yml
@@ -8,6 +8,7 @@ kind-dependencies:
     - fetch
 
 transforms:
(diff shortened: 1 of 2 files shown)


ahal (approved):
Thanks, makes sense!

bakulf (approved):

PR #3425, opened 2022-05-03
Ignore: Taskcluster cache experiment

strseb:
- Use hashIndex for Openssl
- Can we cache the ccache folder on taskgraph?

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/build-openssl/kind.yml
+++ b/taskcluster/ci/build-openssl/kind.yml
@@ -8,6 +8,7 @@ kind-dependencies:
     - fetch
 
 transforms:
+    - mozillavpn_taskgraph.transforms.hashIndex:transforms
     - taskgraph.transforms.job:transforms
     - taskgraph.transforms.task:transforms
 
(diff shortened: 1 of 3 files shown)


strseb:
@ahal - Just wondering if, using the docker-worker cache for ccache, go-cache and the gradle-cache is something that would make sense. Locally it speeds up the task significantly but this would be of no use if we are constantly swapping out the workers.

ahal:
> @ahal - Just wondering if, using the docker-worker cache for ccache, go-cache and the gradle-cache is something that would make sense. Locally it speeds up the task significantly but this would be of no use if we are constantly swapping out the workers.

This seems like a good thing to land regardless IMO.. Worker manager tries not to spin workers up / down unless necessary and there's even an `idleShutdown` attribute we can set for a given pool which controls how long a worker is allowed to linger around idle before being released. So if we find we are getting too many cache misses for a given pool, we could try increasing that value.

ahal (changes requested):

PR #3431, opened 2022-05-03
Tips and tricks page

MattLichtenstein:
## Description

- Create "Guides" card component
- Create "Tutorials" card component
- Implement Tips & Tricks UI
- Add Tips & Tricks functional tests
- Guide and tutorial cards DO NOT do anything when clicked (in this PR)

https://user-images.githubusercontent.com/15353801/167491100-b4a94b82-2d18-49ac-a3fc-16083d61ffe2.mov

https://user-images.githubusercontent.com/15353801/167491368-4cd9a533-a12c-461c-acf6-e91640ec9821.mov

*Video note: The "Connect VPN on startup" tutorial would not actually be visible on iOS (due to not meeting "Start on boot" condition) but is present in the video for demonstration purposes

## Reference

[VPN-2074: As a VPN user,  I would like a dedicated Tips and Tricks page where I can view guides, start tutorials, and get a feature tour so that I can get up to speed faster.](https://mozilla-hub.atlassian.net/browse/VPN-2074)

## Checklist
    

src/ui/settings/ViewTipsAndTricks.qml
```diff
@@ -0,0 +1,130 @@
...
+
+            Flow {
+                Layout.fillWidth: true
+                spacing: VPNTheme.theme.vSpacingSmall
+
+                VPNTutorialCard {
+                    width: vpnFlickable.screenWidth < root.tabletSize ? parent.width : (parent.width - parent.spacing) / 2
+                    height: 144
+
+                    imageSrc: "qrc:/tutorials/01_get_started.svg"
```
bakulf, line 50: This is not good. You must use the model (or other c++ to qml layer). nothing has to be hardcoded for the following reasons:

1. Not all the tutorials can be exposed. See the `conditions` attribute in the JSON files. (Platforms, features, etc).
2. This must be dynamic. We cannot change the qml code each time we add/remove/change a tutorial.
3. It breaks the abstraction layer of the tutorial component.

Let's see how to do it properly. I'm sure we can find an elegant solution between you, Lesley and me :)

@lesleyjanenorton do you have any good hints to improve this code?

MattLichtenstein: This was a part of an older commit from a few days ago and these were just "dummy" cards I used to start laying out the page. The cards do get their data from the model as follows: 

```
VPNTutorialCard {
    property variant tutorial: VPNTutorial.getTutorialById("01_get_started")

    width: vpnFlickable.screenWidth < root.tabletSize ? parent.width : (parent.width - parent.spacing) / 2
    height: 144

    imageSrc: tutorial.image
    imageBgColor: VPNTheme.theme.ink //  TODO: REPLACE WITH DATA FROM MODEL ONCE BACKGROUND COLOR IS ADDED TO SCHEMA
    title: VPNl18n[tutorial.titleId]
    description: VPNl18n[tutorial.subtitleId]
}
```
You'll notice that this is still somewhat hardcoded by the property getting the tutorial object by knowing the tutorialId. That is because this card component is **NOT** a part of a component that takes in a model and creates delegates (eg ListView, Repeater). The reason for this is because of the way the tips and tricks page designed, not all tutorial cards are laid out next to each other, but they are divided into sections

tipsandtricks

Right now we have one tutorial at the top (not including whats new because that doesn't get data from the model so presumably should remain completely hardcoded), and one at the bottom, which means they really can't be a part of the same component that lays out a whole model. 

The only way we can use those objects that lay out models is if we filter our models or have multiple models only containing the indices that we need for designated locations in the UI. For example, we could have a "header" tutorial model for the cards laid out at the top, and an "other" tutorial model for the headers laid out at the bottom, and we could add this data to this "designated location" to the json schema, or set up the file structure in such a way so that tutorials get put in the correct model, but because the future of this page is determined by feedback and usage, the layout may change slightly or dramatically, so it may be too soon to start splitting the monolithic tutorial model into sub/filtered models based on where they go in the view.

All that being said, if we think that there might be a better way to do this, I am definitely open to suggestions/brainstorming 😄 

Good call on the conditions though, I was not aware of those as I haven't gotten to actually implementing the guides/tutorials - just the card components right now, but I shall add logic to check this before displaying the cards.

MattLichtenstein: I have pushed [another, more up-to-date commit](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3431/commits/2b0a602c7caa201034987a4b17add803f515322e) to better represent the direction I am going in. Let me know your thoughts!

bakulf: Ok i see! 
I would do the following:
1. Let's talk with Peter to see if we can simplify the view. 
2. The challenge here is too have a totally dynamic view: it must be able to support 0, 1, many tutorials. Same thing for guides.
3. The condition can not be checked in the view because..  unsupported tutorials are not loaded at all. Same thing for the guides
4. If needed, we can pick the first element from the model. Then we can use a VPN filterable model to take all the others. Could this be a solution?

src/ui/settings/ViewTipsAndTricks.qml
```diff
@@ -0,0 +1,169 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.15
+import QtQuick.Layouts 1.15
+
+import Mozilla.VPN 1.0
+import components 0.1
+
+Item {
```
lesleyjanenorton, line 13: Can probably get rid of this `Item {}` (though I know a few of the settings sub-pages follow this pattern)

src/ui/settings/ViewTipsAndTricks.qml
```diff
@@ -0,0 +1,186 @@
...
+
+                            imageSrc: firstTutorial.image
+                            imageBgColor: firstTutorial.imageBgColor
+                            title: VPNl18n[firstTutorial.titleId]
+                            description: VPNl18n[firstTutorial.subtitleId]
+                        }
+                    }
+
+                    VPNTutorialCard {
+                        width: vpnFlickable.screenWidth < root.tabletSize ? parent.width : (parent.width - parent.spacing) / 2
```
lesleyjanenorton, line 76: Any reason to not specify the width/height in VPNTutorialCard.qml ?

src/ui/settings/ViewTipsAndTricks.qml
lesleyjanenorton: Let's bring this up in the design meeting. I think my preference would be to push the What's New card to the top and keep all of the tutorial cards together in a single block, like the guide cards.

lesleyjanenorton:
@MattLichtenstein let's also remove the 'What's new' option from the settings menu since that will be accessed via Tips & Tricks now

src/ui/settings/ViewTipsAndTricks.qml
MattLichtenstein: I remember trying to avoid using this at the start and was running into either some weird qml warning or layout issues and figured it was some hacky workaround, but let me try again!

src/ui/settings/ViewTipsAndTricks.qml
MattLichtenstein: In this specific case, the width of the tutorial cards are sometimes reliant on ancestor components which seems like bad practice to refer to from inside the component file. 

The height is always the same for all tutorial cards here, but I guess in the off chance that these cards are used in some other view at some point with a different height, it may be weird to have 'default' height 🤷

nebula/ui/themes/themes.js
```diff
@@ -200,6 +200,12 @@ theme.vpnToggleDisconnected = {
   'focusBorder': theme.darkFocusBorder,
 };
 
+theme.card = {
+    'defaultColor': theme.white,
+    'focusOutline': theme.bgColorTransparent,
+    'focusBorder': theme.bgColorTransparent,
```
lesleyjanenorton, line 206: Do we know what should happen when a card has active focus or is hovered/clicked? I don't think I've seen it called out in the spec.

lesleyjanenorton (approved):

PR #3438, opened 2022-05-04
Fix the tutorial engine

bakulf:
## Description

With Qt6 a few things were broken. Others were missing.

## Reference

VPN-1745

### DIFF
--- a/src/models/guide.cpp
+++ b/src/models/guide.cpp
@@ -9,6 +9,7 @@
 #include "logger.h"
 #include "models/feature.h"
 #include "mozillavpn.h"
+#include "settingsholder.h"
 
 #include <QFile>
 #include <QJsonArray>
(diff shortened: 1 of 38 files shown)


flodolo (approved):
Please pass the Python script through black.

src/models/guide.cpp
```diff
@@ -19,8 +20,121 @@
 namespace {
 Logger logger(LOG_MAIN, "Guide");
 
+bool evaluateConditionsEnabledFeatures(const QJsonArray& enabledFeatures) {
+  for (QJsonValue enabledFeature : enabledFeatures) {
+    // If the feature doesn't exist, we crash.
+    const Feature* feature = Feature::get(enabledFeature.toString());
+    Q_ASSERT(feature);
```
oskirby, line 27: I think an assert is the wrong approach here, we should just treat it as an unsupported guide in this case.

src/models/guide.cpp
```diff
@@ -19,8 +20,121 @@
...
+
+  return true;
+}
+
+bool evaluateConditionsSettingsOp(const QString& op, bool result) {
+  if (op == "eq") return result;
+
+  if (op == "neq") return !result;
+
+  logger.error() << "Invalid settings operator" << op;
```
oskirby, line 58: In the taxonomy of logging levels, this should be a warning(): something has gone wrong in an unexpected way but the application can continue to function.

oskirby (approved):

PR #3451, opened 2022-05-05
Fix German Free Trial layout

strseb:
## Description

Currently, the free trial chip is cut off in German, so let's take the Headline-Overflow into account when calculating the box size
Before German:

| Live(German) | Fixed (German) | Fix applied  (English) |
| ------------- | ------------- | -----|
| Screenshot_20220505-151213(1) | Screenshot_20220505-155911 | Screenshot_20220505-155943 |

Don't mind the sorting, that is from my other pr.

### DIFF
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -21,7 +21,7 @@ RadioDelegate {
     ButtonGroup.group: subscriptionOptions
 
     Layout.fillWidth: true
-    Layout.preferredHeight: chipFreeTrial.visible ? 96+headLineLabel.overflow : 96
+    Layout.preferredHeight: Math.max(96, row.implicitHeight + row.anchors.topMargin + row.anchors.bottomMargin)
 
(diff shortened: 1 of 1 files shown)


nebula/ui/components/VPNSubscriptionOption.qml
```diff
@@ -21,7 +21,7 @@ RadioDelegate {
     ButtonGroup.group: subscriptionOptions
 
     Layout.fillWidth: true
-    Layout.preferredHeight: 96
+    Layout.preferredHeight: chipFreeTrial.visible ? 96+headLineLabel.overflow : 96
```
MattLichtenstein, line 24: I think you can achieve the same thing by using the height of the layout, which may be a little more robust. Basically making it at least 96 points tall, but expanding if necessary

something like: `Layout.preferredHeight: Math.max(96, row.implicitHeight + row.anchors.topMargin + row.anchors.bottomMargin)`

MattLichtenstein (approved):
Excellent find! Just one suggestion

Issue #3413, opened 2022-05-02
[taskcluster] Android build failures do not cause the task to fail


ahal:
Noticed this on latest push to main:
https://firefox-ci-tc.services.mozilla.com/tasks/CoACVcV8TyKmGdriyI8y5w/runs/0/logs/public/logs/chain_of_trust.log

Looks like the `release_artifacts` might have changed? Unsure what's going on here, but will poke around.

/cc @gabrielBusta

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2143)
┆author: Data Sync

ahal:
On the subsequent push, the tasks started passing again. I'm going to leave this open a bit to try and understand what happened here.

escapewindow:
This is bizarre; the chain of trust artifact generation, signing, and upload should happen after all the other artifacts are generated, so we should have all of the artifact shas in the [chain of trust artifact](https://firefoxci.taskcluster-artifacts.net/Dtj2rO61TcOxQmDvUxqT8g/0/public/chain-of-trust.json). But we only have ` certified.log`:

```
{
  "chainOfTrustVersion": 1,
  "artifacts": {
    "public/logs/certified.log": {
      "sha256": "4f915b023a39930990a5b26a8bfbc9a60d30f8cbe3fe98fbf5f247ed9bbe3a52"
    }
  },
```

This is the first time I've seen something like this, so I'm thinking it's an anomaly. Otherwise it may be an issue in the worker daemon.

ahal:
Looks like I spoke too soon and one of the Android signing tasks *did* fail in the subsequent push:
https://firefox-ci-tc.services.mozilla.com/tasks/AwFVaSSFQLyszK0H3dd2jg/runs/0

So appears to be intermittent?

escapewindow:
We're using the exact same AMIs as the other docker-worker pools: https://hg.mozilla.org/ci/ci-configuration/file/d3a8d38297b615d875ba05392032eded89c0dfc6/worker-pools.yml#l2877

[...]

PR #3461, opened 2022-05-06
ci(taskcluster): make sure Android build failures turn the task red (…

ahal:
…#3413)

We noticed that sometimes the Android builds fail but the task is still
green. Partly this is taskcluster/taskcluster#4590 as the task should
fail if one of its declared artifacts doesn't exist.

But partly the build script itself should fail the task if one of its
commands returns non-zero. This PR fixes that by enabling `set -e` in
all the build scripts.

### DIFF
--- a/taskcluster/scripts/build/android_build_debug.sh
+++ b/taskcluster/scripts/build/android_build_debug.sh
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+set -e
 
 # This script is used in the Android Debug (universal) build task
 git submodule init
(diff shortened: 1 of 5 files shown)


ahal:
There's a chance this may introduce build failures if there are commands that are expected to return non-zero.. So there may be some follow-up work to fix those...

ahal:
Also the Android builds are currently intermittently failing.. So this *will* appear to introduce regressions there, but really it's just making the failures visible.

ahal:
Looks like the `wasm` builds were also failing.

strseb (approved):

mbirghan (approved):

ahal:
Looks like the `wasm` task was also potentially failing?
https://firefox-ci-tc.services.mozilla.com/tasks/cvvWz1VBRhyF3w7X8yeUfA/runs/0/logs/public/logs/live.log

Or maybe the same OOM error the Androind ones are hitting?

ahal:
@strseb I'll leave it up to you whether you want to merge this now or block on fixing the Android / wasm builds first.

PR #3474, opened 2022-05-09
Get android builds a bigger taskcluster vm

strseb:
## Description

we're running OOM a lot, so let's use a bigger taskcluster vm :) 
## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/config.yml
+++ b/taskcluster/ci/config.yml
@@ -41,7 +41,7 @@ workers:
             provisioner: 'mozillavpn-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: b-linux-large
+            worker-type: b-linux-large-gcp
         b-win2012:
             provisioner: 'mozillavpn-{level}'
             implementation: generic-worker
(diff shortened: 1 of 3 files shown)


taskcluster/ci/config.yml
```diff
@@ -37,6 +37,11 @@ workers:
             implementation: docker-worker
             os: linux
             worker-type: b-linux
+        b-linux-large:
+            provisioner: 'mozillavpn-{level}'
+            implementation: docker-worker
+            os: linux
+            worker-type: b-linux-large
```
ahal, line 44: Oh, I think I gave you some bad information. Looks like this might need to be `b-linux-large-gcp`.. Sorry!

You can keep the name of this config as `b-linux-large` and that way you can still use `b-linux-large` in the task definitions.

ahal (approved):
Nice!

I'll just leave a note here that an `xlarge` pool is also an option. Though we'd want to find a balance between CI cost / task runtime.

bakulf (approved):

PR #3485, opened 2022-05-09
Remove the 'aia' params in the client user-agent - PSP-393

bakulf:
## Description

Remove the 'aia' params in the client user-agent

## Reference

PSP-393

### DIFF
--- a/src/networkmanager.cpp
+++ b/src/networkmanager.cpp
@@ -73,10 +73,6 @@ QByteArray NetworkManager::userAgent() {
       flags.append("iap:true");
     }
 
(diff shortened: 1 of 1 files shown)


oskirby (approved):
I think this looks good and don't forsee any problems with the change on the client side. We're really just pushing the complexity into the guardian-website, which should have all the data it needs to figure the correct auth mode out.

birdsarah (approved):
Should we also clean out FeatureInAppAuth. I got the impression that it'll be everywhere now and it's not really a question of "isSupported" but whether a given platform choses to go through one flow or another.

PR #3490, opened 2022-05-10
Fetching the balrog content before emitting the signal

bakulf:
## Description

Fetching the balrog content before emitting the signal

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2196


bakulf:
Actually, we can do even better:
- we can download the file if the update is triggered by the client
- we don't download the file is this has been triggered by user-interaction
- we can add a progress bar to inform the user when the downloading is actually happening

bakulf:
Talking with Peter, we want to postpone this PR. This consumes too much bandwidth.

PR #3497, opened 2022-05-10
Add highlighted tutorial attribute

bakulf:
## Description

We need to show the highlighted tutorial on top of the view. Let's expose it

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2200


MattLichtenstein (approved):
Looks good, though this won't work if we ever need to have more than one tutorial highlighted (for example, one for mobile, and a different one for desktop) without some kind of priority or sorting system

PR #3502, opened 2022-05-11
Remove webview auth code for iOS and Android

bakulf:
## Description

This is dead code. In 2.8.0 we have released the auth in app.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2206


strseb (approved):

bakulf:
@strseb there some taskcluster failures. Can you please tell me if they are connected to my PR or not? and if not how to have them fixed?

strseb:
This [mac runner](https://firefox-ci-tc.services.mozilla.com/provisioners/releng-hardware/worker-types/mozillavpn-b-1-osx/workers/mdc1/macmini-r8-179) seems to be busted again. @ahal - I think we had our mac runner stuck once, can you maybe help here again?
Wasm seems to be something about the translations - i can look into that.

ahal:
Thanks. For some reasons the pool (including the level 3 one) keeps getting into a bad state. Dave House has been getting them unstuck, but it keeps happening..

I'll file another relops JIRA ticket, but I think we're going to have to address the underlying cause here as this is happening much too frequently :/

PR #3505, opened 2022-05-11
VPN-2260 - Create resilient connection with Guardian Websocket

brizental:
## Description

This sets up the client to connect to the Guardian websocket implemented on https://github.com/mozilla-services/guardian-website/pull/1504.

Right now it's not sending any messages so over here we are also not doing anything to process messages. Those will come as follow up work.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2260

## Checklist
    
### DIFF
--- a/src/mozillavpn.h
+++ b/src/mozillavpn.h
@@ -30,6 +30,7 @@
 #include "statusicon.h"
 #include "telemetry.h"
 #include "theme.h"
+#include "websockethandler.h"
 
 #include <QList>
 #include <QNetworkReply>
@@ -291,7 +292,7 @@ class MozillaVPN final : public QObject {
 
   void setUpdateRecommended(bool value);
 
-  UserState userState() const { return m_userState; }
+  UserState userState() const;
 
   bool startMinimized() const { return m_startMinimized; }
 
@@ -447,6 +448,7 @@ class MozillaVPN final : public QObject {
     SurveyModel m_surveyModel;
     Telemetry m_telemetry;
     Theme m_theme;
+    WebSocketHandler m_webSocketHandler;
     WhatsNewModel m_whatsNewModel;
     User m_user;
   };
--- a/src/websockethandler.cpp
+++ b/src/websockethandler.cpp
@@ -38,7 +38,7 @@ QString WebSocketHandler::s_customWebSocketServerUrl = "";
 
 // static
 QString WebSocketHandler::webSocketServerUrl() {
-  if (s_customWebSocketServerUrl.isEmpty()) {
+  if (!s_customWebSocketServerUrl.isEmpty()) {
     return s_customWebSocketServerUrl;
   }
 
--- a/src/websockethandler.cpp
+++ b/src/websockethandler.cpp
@@ -29,12 +29,17 @@ WebSocketHandler::WebSocketHandler() {
 
   connect(&m_pingTimer, &QTimer::timeout, this,
           &WebSocketHandler::onPingTimeout);
+
+  m_pingTimer.setSingleShot(true);
 }
 
 // static
+QString WebSocketHandler::s_customWebSocketServerUrl = "";
+
+// static
 QString WebSocketHandler::webSocketServerUrl() {
-  if (WebSocketHandler::m_customWebSocketServerUrl.count() > 0) {
-    return WebSocketHandler::m_customWebSocketServerUrl;
+  if (s_customWebSocketServerUrl.isEmpty()) {
+    return s_customWebSocketServerUrl;
   }
 
   QString httpServerUrl;
@@ -55,12 +60,10 @@ bool WebSocketHandler::isUserAuthenticated() {
   return vpn->userState() == MozillaVPN::UserAuthenticated;
 }
 
+#ifdef UNIT_TEST
 // static
-QString WebSocketHandler::m_customWebSocketServerUrl = "";
-
-// static
-void WebSocketHandler::testOverrideWebSocketServerUrl(QString url) {
-  WebSocketHandler::m_customWebSocketServerUrl = url;
+void WebSocketHandler::testOverrideWebSocketServerUrl(const QString& url) {
+  WebSocketHandler::s_customWebSocketServerUrl = url;
 }
 
 void WebSocketHandler::testOverridePingInterval(int newInterval) {
@@ -70,6 +73,7 @@ void WebSocketHandler::testOverridePingInterval(int newInterval) {
 void WebSocketHandler::testOverrideRetryInterval(int newInterval) {
   m_retryInterval = newInterval;
 }
+#endif
 
 void WebSocketHandler::initialize() {
   logger.debug() << "Initialize";
@@ -94,7 +98,7 @@ void WebSocketHandler::onUserStateChanged() {
   logger.debug() << "User state change detected:"
                  << MozillaVPN::instance()->userState();
 
-  if (WebSocketHandler::isUserAuthenticated()) {
+  if (isUserAuthenticated()) {
     open();
   } else {
     close();
@@ -115,12 +119,12 @@ void WebSocketHandler::open() {
   }
 
   logger.debug() << "Attempting to open WebSocket connection."
-                 << WebSocketHandler::webSocketServerUrl();
+                 << webSocketServerUrl();
 
   QNetworkRequest request;
   request.setRawHeader("Authorization",
                        SettingsHolder::instance()->token().toLocal8Bit());
-  request.setUrl(QUrl(WebSocketHandler::webSocketServerUrl()));
+  request.setUrl(QUrl(webSocketServerUrl()));
   m_webSocket.open(request);
 }
 
@@ -162,7 +166,7 @@ void WebSocketHandler::close() {
 void WebSocketHandler::onClose() {
   logger.debug() << "WebSocket closed";
 
-  if (WebSocketHandler::isUserAuthenticated()) {
+  if (isUserAuthenticated()) {
     logger.debug()
         << "User is authenticated. Attempting to reopen WebSocket in:"
         << m_retryInterval;
@@ -185,8 +189,6 @@ void WebSocketHandler::onClose() {
  */
 void WebSocketHandler::sendPing() {
   m_webSocket.ping();
-
-  m_pingTimer.setSingleShot(true);
   m_pingTimer.start(m_pingInterval);
 }
 
@@ -198,8 +200,6 @@ void WebSocketHandler::sendPing() {
  * out.
  */
 void WebSocketHandler::onPong(quint64 elapsedTime) {
-  qDebug() << "PONG PONG PONG";
-
   logger.debug() << "WebSocket pong" << elapsedTime;
 
   m_pingTimer.stop();
--- a/src/websockethandler.cpp
+++ b/src/websockethandler.cpp
@@ -33,6 +33,10 @@ WebSocketHandler::WebSocketHandler() {
 
 // static
 QString WebSocketHandler::webSocketServerUrl() {
+  if (WebSocketHandler::m_customWebSocketServerUrl.count() > 0) {
+    return WebSocketHandler::m_customWebSocketServerUrl;
+  }
+
   QString httpServerUrl;
   if (Constants::inProduction()) {
     httpServerUrl = Constants::API_PRODUCTION_URL;
@@ -49,6 +53,22 @@ bool WebSocketHandler::isUserAuthenticated() {
   Q_ASSERT(vpn);
 
   return vpn->userState() == MozillaVPN::UserAuthenticated;
+}
+
+// static
+QString WebSocketHandler::m_customWebSocketServerUrl = "";
+
+// static
+void WebSocketHandler::testOverrideWebSocketServerUrl(QString url) {
+  WebSocketHandler::m_customWebSocketServerUrl = url;
+}
+
+void WebSocketHandler::testOverridePingInterval(int newInterval) {
+  m_pingInterval = newInterval;
+}
+
+void WebSocketHandler::testOverrideRetryInterval(int newInterval) {
+  m_retryInterval = newInterval;
 }
 
 void WebSocketHandler::initialize() {
@@ -145,9 +165,9 @@ void WebSocketHandler::onClose() {
   if (WebSocketHandler::isUserAuthenticated()) {
     logger.debug()
         << "User is authenticated. Attempting to reopen WebSocket in:"
-        << WEBSOCKET_RETRY_INTERVAL;
+        << m_retryInterval;
 
-    QTimer::singleShot(WEBSOCKET_RETRY_INTERVAL, this, &WebSocketHandler::open);
+    QTimer::singleShot(m_retryInterval, this, &WebSocketHandler::open);
   } else {
     logger.debug()
         << "User is not authenticated. Will not attempt to reopen WebSocket.";
@@ -167,7 +187,7 @@ void WebSocketHandler::sendPing() {
   m_webSocket.ping();
 
   m_pingTimer.setSingleShot(true);
-  m_pingTimer.start(WEBSOCKET_PING_INTERVAL);
+  m_pingTimer.start(m_pingInterval);
 }
 
 /**
@@ -183,8 +203,7 @@ void WebSocketHandler::onPong(quint64 elapsedTime) {
   logger.debug() << "WebSocket pong" << elapsedTime;
 
   m_pingTimer.stop();
-  QTimer::singleShot(WEBSOCKET_PING_INTERVAL, this,
-                     &WebSocketHandler::sendPing);
+  QTimer::singleShot(m_pingInterval, this, &WebSocketHandler::sendPing);
 }
 
 /**
--- a/src/websockethandler.cpp
+++ b/src/websockethandler.cpp
@@ -1,0 +1,222 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "websockethandler.h"
+#include "constants.h"
+#include "leakdetector.h"
+#include "logger.h"
+#include "mozillavpn.h"
+#include "networkrequest.h"
+#include "settingsholder.h"
+#include "urlopener.h"
+
+namespace {
+Logger logger(LOG_MAIN, "WebSocketHandler");
+}  // namespace
+
+WebSocketHandler::WebSocketHandler() {
+  MVPN_COUNT_CTOR(WebSocketHandler)
+
+  connect(&m_webSocket, &QWebSocket::connected, this,
+          &WebSocketHandler::onConnected);
+  connect(&m_webSocket, &QWebSocket::disconnected, this,
+          &WebSocketHandler::onClose);
+  connect(&m_webSocket,
+          QOverload<QAbstractSocket::SocketError>::of(&QWebSocket::error), this,
+          &WebSocketHandler::onError);
+  connect(&m_webSocket, &QWebSocket::pong, this, &WebSocketHandler::onPong);
+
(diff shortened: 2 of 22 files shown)


brizental:
This a draft because I still have to (figure out how to) write tests.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "urlopener.h"
+#include "mozillavpn.h"
+
+#include <QtCore/QDebug>
+#include <QAbstractSocket>
+#include <QTimer>
+
+constexpr const char* WEBSOCKET_SERVER_URL = "ws://localhost:3000";
+constexpr const int WEBSOCKET_PING_INTERVAL = 30 * 1000;   // 30s
+constexpr const int WEBSOCKET_RETRY_INTERVAL = 30 * 1000;  // 30s
```
brizental, line 17: Note: The way that I chose this interval was completely arbitrary.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+ * attempts.
+ */
+void WebSocketHandler::ping() {
+  if (!this->m_isAlive) {
+    qDebug() << "WebSocket was disconnected silently. Closing.";
+    m_webSocket.close();
+    return;
+  }
+
+  m_isAlive = false;
```
bakulf, line 90: Do we want to add a timeout between a ping and a pong? Something like, if we have not received a pong() in  < 20 seconds, let's consider the connection as gone.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+    qDebug() << "WebSocket closed explicitly. Will not attempt to reopen.";
+  }
+}
+
+/**
+ * @brief Ackowledges there was a WebSocket error.
+ *
+ */
+void WebSocketHandler::onError(QAbstractSocket::SocketError error) {
+  qDebug() << "WebSocket error:" << error;
```
bakulf, line 145: here i would try to reconnect (after X seconds).

bakulf: You can do it using:

`TimerSingleShot::create(this, X-seconds, [this]() { reconnect(); });`

bakulf (changes requested):
Excellent starting point! Thanks! I think I wrote too many confusing comments.
Happy to chat with you tomorrow to go through them.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "networkrequest.h"
+#include "settingsholder.h"
+#include "urlopener.h"
+#include "mozillavpn.h"
+
+#include <QtCore/QDebug>
+#include <QAbstractSocket>
+#include <QTimer>
+
+constexpr const char* WEBSOCKET_SERVER_URL = "ws://localhost:3000";
```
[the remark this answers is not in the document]
brizental: Hm, actually this is development leftover. We can use https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/constants.h#L83-L85 with different protocols.

src/websockethandler.cpp
brizental: If the Websocket is closed, the onClose handler is also called and that already attempts a reconnect.

src/websockethandler.cpp
brizental: That is how this is working. The timeout is 30s. We send a ping and change `m_isAlive` to `false` i.e. we assume the connection is broken. If it is not broken we will get a pong and change `m_isAlive` to `true`. If we get to the next ping without ever getting a pong the connections is assumed to be broken and the socked is closed. That trigger the `onClose` handler which attempts a reconnections.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+ */
+void WebSocketHandler::close() {
+  if (m_webSocket.state() == QAbstractSocket::UnconnectedState ||
+      m_webSocket.state() == QAbstractSocket::ClosingState) {
+    qDebug() << "Attempted to close a WebSocket, but it's already closed.";
+    return;
+  }
+
+  qDebug() << "Closing WebSocket";
+  m_reconnectOnClose = false;
```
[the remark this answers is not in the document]
brizental: Hm, `m_reconnectOnClose` is used to check if we should attempt to reconnect when the socket is closed. This is to address the situation of closing the websocket on logout. We do not want to reconnect when the user is logged out.

If the connection fails / is broken / errors we already attempt a reconnect.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+  m_isAlive = true;
+}
+
+/**
+ * @brief Ackowledges the WebSocket has been succesfully connected.
+ */
+void WebSocketHandler::onConnected() {
+  qDebug() << "WebSocket connected";
+
+  m_isAlive = true;
```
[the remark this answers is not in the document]
brizental: Hm, maybe we could. I am not sure m_webSocket state will hold up to the assumptions I make in the strategy to check for connection though, I'd rather keep m_isAlive which we control. This is related to my other comment https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3505#discussion_r871095160

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+#include <QtCore/QDebug>
+#include <QAbstractSocket>
+#include <QTimer>
+
+constexpr const char* WEBSOCKET_SERVER_URL = "ws://localhost:3000";
+constexpr const int WEBSOCKET_PING_INTERVAL = 30 * 1000;   // 30s
+constexpr const int WEBSOCKET_RETRY_INTERVAL = 30 * 1000;  // 30s
+
+WebSocketHandler::WebSocketHandler() {
```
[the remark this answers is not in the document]
brizental: For the destructor I get an error that I am implementing something that is already implicitly implemented. I guess it's fine to omit in this case? I updated the constructor though.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * @brief Opens the websocket connection.
+ *
+ * No-op in case the connection is already open.
+ */
+void WebSocketHandler::open() {
+  if (m_webSocket.state() != QAbstractSocket::UnconnectedState &&
+      m_webSocket.state() != QAbstractSocket::ClosingState) {
+    qDebug()
```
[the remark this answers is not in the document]
brizental: Would you mind expanding on why? I am relying on the user state, because I need the auth token to connect to the websocket. The `State` enum has many more variants, UserState looks more straight forward for this case IMO.

src/websockethandler.cpp
brizental: Ack, I just realized I am confusing this. On Node.js everytime there is an error it is also followed by a close event. It seems it's not really the case on Qt. I'll fix this.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * @brief Acknowledges the user state has changed.
+ *
+ * When the user has authenticated, a WebSocket connection is openened.
+ * Otherwise it's closed.
+ *
+ * The Guardian WebSocket does not accept unauthenticated users. Attempting to
+ * connect while unauthenticated would result is an infinite retry loop.
+ */
+void WebSocketHandler::userStateChanged() {
```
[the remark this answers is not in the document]
brizental: Would you mind expanding on why? I am relying on the user state, because I need the auth token to connect to the websocket. The State enum has many more variants, UserState looks more straight forward for this case IMO.

src/websockethandler.cpp
brizental: Addressed on [4f882d5](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3505/commits/4f882d5bef590f04c0cf0be2c8f6e6f537f1c280)

src/websockethandler.cpp
```diff
@@ -0,0 +1,217 @@
...
+void WebSocketHandler::close(bool reconnect) {
+  if (m_webSocket.state() == QAbstractSocket::UnconnectedState ||
+      m_webSocket.state() == QAbstractSocket::ClosingState) {
+    logger.debug()
+        << "Attempted to close a WebSocket, but it's already closed.";
+    return;
+  }
+
+  logger.debug() << "Closing WebSocket";
+  m_reconnectOnClose = reconnect;
```
brizental, line 104: Drop this state and instead use User.isauthenticated or smth.

src/websockethandler.h
```diff
@@ -0,0 +1,42 @@
...
+ signals:
+  void closed();
+
+ private:
+  QWebSocket m_webSocket;
+  QUrl m_url;
+  QTimer* m_pingTimer = nullptr;
+
+  bool m_reconnectOnClose = true;
+  bool m_isAlive = false;
```
brizental, line 27: Rename to `m_waitingForPong`.

src/websockethandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+ * state.
+ */
+void WebSocketHandler::onClose() {
+  m_pingTimer->stop();
+
+  if (m_reconnectOnClose) {
+    qDebug() << "WebSocket closed. Attempting to reopen in:"
+             << WEBSOCKET_RETRY_INTERVAL;
+
+    QTimer::singleShot(WEBSOCKET_RETRY_INTERVAL, this, &WebSocketHandler::open);
```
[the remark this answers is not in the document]
brizental: Follow-up filed: https://mozilla-hub.atlassian.net/browse/PSP-398

src/websockethandler.cpp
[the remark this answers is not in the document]
brizental: We discussed this offile and decided to keep using UserState for now. The reason for moving to StateMain was that it has a more loaded meaning. Not only does it mean the user is authenticated, it also means the VPN is ready for connection, subscription is valid and so on. 

Because we are not sure the scope of the messages that will be passed through this WebSocket connection, relying on UserState or State right now doesn't make too big a difference. State is more restrictive. If we ever see the need we can change to rely on it instead.

src/websockethandler.cpp
brizental: We spoke offline, yes, this makes sense.

src/websockethandler.cpp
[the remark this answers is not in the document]
brizental: We spoke offline, we will drop reconnectOnClose and instead just check for the user authentication state.

src/websockethandler.h
```diff
@@ -0,0 +1,43 @@
...
+  void ping();
+  static QString webSocketServerUrl();
+  static bool isUserAuthenticated();
+
+  void onUserStateChanged();
+  void onConnected();
+  void onClose();
+  void onMessageReceived(const QString& message);
+  void onError(QAbstractSocket::SocketError error);
+  void onPong(quint64 elapsedTime);
```
bakulf, line 40: Usually our class headers are:

```
class Foo {
  public:
    Foo();

  signals:
    void signal();

  private:
    methods();

  private:
   bool m_var = true;
}
```

I would move the member variables at the bottom in their 'private' block.

bakulf (commented):
To make the tests green-ish, I think you need to:
1. add the QT webSocket dependency in the MozillaVPN and QML projects (both qmake and cmake). See other QT libraries such as Gui, Test,Widgets
2. Something is wrong with github actions because your builds are unable to find the github secrets. I'll check why
3. We also need to add the websocket dependencies for linux packages! Talk with @oskirby if you need to know more
4. probably we need to do disable this feature for wasm

src/websockethandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+ */
+void WebSocketHandler::ping() {
+  // Never got a pong for the last ping sent, assume connection is broken.
+  if (m_waitingForPong) {
+    logger.debug() << "WebSocket was disconnected silently. Closing.";
+    m_webSocket.close();
+    return;
+  }
+
+  m_waitingForPong = true;
```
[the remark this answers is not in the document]
brizental: This approach would not work. When the WebSocket is silently disconnected, the `m_webSocket.state()` will still be `Connected` even though the connection is lost. (I just verified that this is indeed the case, by testing locally. Network disconnection is something that will cause a silent disconnection, for example).

I cannot think of another approach to this issue that does not involve keeping some sort of connection state on the WebSocketHandler. I am 100% open for suggestions though :) Maybe we can use an enum instead of boolean?

src/mozillavpn.h
```diff
@@ -30,6 +30,7 @@
 #include "statusicon.h"
 #include "telemetry.h"
 #include "theme.h"
+#include "websockethandler.h"
```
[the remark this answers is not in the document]
brizental: This gives a "field has incomplete type error". Should I start `m_webSocketHandler` as a `nullptr` and so on?

src/websockethandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+  connect(&m_webSocket, &QWebSocket::connected, this,
+          &WebSocketHandler::onConnected);
+  connect(&m_webSocket, &QWebSocket::disconnected, this,
+          &WebSocketHandler::onClose);
+  connect(&m_webSocket,
+          QOverload<QAbstractSocket::SocketError>::of(&QWebSocket::error), this,
+          &WebSocketHandler::onError);
+  connect(&m_webSocket, &QWebSocket::pong, this, &WebSocketHandler::onPong);
+
+  m_pingTimer = new QTimer(this);
```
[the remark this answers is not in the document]
brizental: C++ newbie question: what do you mean by "no need to allocate it" here?

src/websockethandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// static
+bool WebSocketHandler::isUserAuthenticated() {
+  MozillaVPN* vpn = MozillaVPN::instance();
+  Q_ASSERT(vpn);
+
+  if (vpn->userState() == MozillaVPN::UserAuthenticated) {
+    return true;
+  } else {
```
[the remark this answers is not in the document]
brizental: Nice. Would it be possible to have a linter warning for this? Happy to file a bug if so.

src/websockethandler.cpp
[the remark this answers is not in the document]
brizental: Hum, ok I re-read your comment and I am wrong! The approach should work indeed because we are closing the connection when there is a timeout. Alright. I'll go change it 🎉

src/websockethandler.cpp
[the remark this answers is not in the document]
bakulf: Maybe there is already something we can use. It would be nice! I agree.

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+          &WebSocketHandler::onError);
+  connect(&m_webSocket, &QWebSocket::pong, this, &WebSocketHandler::onPong);
+
+  connect(&m_pingTimer, &QTimer::timeout, this,
+          &WebSocketHandler::onPingTimeout);
+}
+
+// static
+QString WebSocketHandler::webSocketServerUrl() {
+  if (WebSocketHandler::m_customWebSocketServerUrl.count() > 0) {
```
bakulf, line 36: `if (!m_customWebSocketServerUrl.isEmpty()) {`

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+    return WebSocketHandler::m_customWebSocketServerUrl;
```
bakulf, line 37: Why `WebSocketHandler::` ? Just `return m_customSocketServerUrl;`

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+// static
+bool WebSocketHandler::isUserAuthenticated() {
+  MozillaVPN* vpn = MozillaVPN::instance();
+  Q_ASSERT(vpn);
+
+  return vpn->userState() == MozillaVPN::UserAuthenticated;
+}
+
+// static
+QString WebSocketHandler::m_customWebSocketServerUrl = "";
```
bakulf, line 59: Move it to the top.

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+// static
+void WebSocketHandler::testOverrideWebSocketServerUrl(QString url) {
```
bakulf, line 62: `const QString& url`

src/websockethandler.cpp
bakulf: If static, `s_customWebSocketServerUrl`

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+ * Otherwise it's closed.
+ *
+ * The Guardian WebSocket does not accept unauthenticated users. Attempting to
+ * connect while unauthenticated would result is an infinite retry loop.
+ */
+void WebSocketHandler::onUserStateChanged() {
+  logger.debug() << "User state change detected:"
+                 << MozillaVPN::instance()->userState();
+
+  if (WebSocketHandler::isUserAuthenticated()) {
```
bakulf, line 97: No need `WebSocketHandler::`

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * @brief Ackowledges the WebSocket has been closed.
+ *
+ * May trigger a reconnection attempt depending on whether the user is
+ * authenticated.
+ */
+void WebSocketHandler::onClose() {
+  logger.debug() << "WebSocket closed";
+
+  if (WebSocketHandler::isUserAuthenticated()) {
```
bakulf, line 165: Also here, remove `WebSocketHandler::`

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * @brief Acknowledges a pong response from the WebSocket server. Schedules the
+ * next ping.
+ *
+ * @param elapsedTime How long it took to get the response since a ping was sent
+ * out.
+ */
+void WebSocketHandler::onPong(quint64 elapsedTime) {
+  qDebug() << "PONG PONG PONG";
```
bakulf, line 201: logger.debug() ?

src/websockethandler.h
```diff
@@ -0,0 +1,54 @@
...
+constexpr const int WEBSOCKET_PING_INTERVAL = 30 * 1000;   // 30s
+constexpr const int WEBSOCKET_RETRY_INTERVAL = 30 * 1000;  // 30s
+
+class WebSocketHandler final : public QObject {
+  Q_OBJECT
+ public:
+  WebSocketHandler();
+  void initialize();
+
+  // These functions are here for testing purposes.
```
bakulf, line 20:
```
#ifdef UNIT_TEST
```

tests/unit/testwebsockethandler.cpp
```diff
@@ -0,0 +1,207 @@
...
+ */
+void MockServer::onNewConnection() {
+  QWebSocket* pSocket = m_pWebSocketServer->nextPendingConnection();
+
+  emit newConnection(pSocket->request());
+
+  connect(pSocket, &QWebSocket::disconnected, this,
+          &MockServer::onSocketDisconnected);
+
+  m_clients << pSocket;
```
bakulf, line 43: This is a test, and I guess it's fine, but I would add this line before calling `emit newConnection`. Because maybe, in the future, we want to retrieve the list of clients in the slot, or we want to delete it immediately.

bakulf (approved):
Really cool!

src/websockethandler.cpp
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+/**
+ * @brief Opens the websocket connection.
+ *
+ * No-op in case the connection is already open.
+ */
+void WebSocketHandler::open() {
+  if (m_webSocket.state() != QAbstractSocket::UnconnectedState &&
```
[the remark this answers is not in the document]
brizental: Hum, but this is checking the websocket state not the user state.

src/websockethandler.cpp
brizental: I was using qDebug because the logger.debug statements do not show up in the test logs for some reason.

PR #3508, opened 2022-05-11
Reduce clickable area for checkboxrow labels

MattLichtenstein:
## Description

- Reduce the clickable area of a checkboxrow label to just the width (and height) of the text

## Reference

[VPN-889: Clicking on a radiobox/checkbox label should enable/disable the input field.](https://mozilla-hub.atlassian.net/browse/VPN-889)

## Checklist
    
### DIFF
--- a/tests/qml/tst_VPNCheckBoxRow.qml
+++ b/tests/qml/tst_VPNCheckBoxRow.qml
@@ -59,6 +59,12 @@ Item {
             expected = false
             actual = checkboxrow.isChecked
             verify(expected === actual, `isChecked was ${actual} not ${expected}.`);
+
+            //Simulate clicking to the right of the label of the checkbox. The last parameter (2) represents the topMargin for the VPNInterLabel in VPNCheckBoxRow
+            mouseClick(checkboxrow, checkboxrow.width + 1, 2)
+            expected = false
+            actual = checkboxrow.isChecked
+            verify(expected === actual, `isChecked was ${actual} not ${expected}.`);
         }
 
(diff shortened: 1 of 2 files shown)


tests/qml/tst_VPNCheckBoxRow.qml
```diff
@@ -59,6 +59,12 @@ Item {
             expected = false
             actual = checkboxrow.isChecked
             verify(expected === actual, `isChecked was ${actual} not ${expected}.`);
+
+            //Simulate clicking to the right of the label of the checkbox. The last parameter (2) represents the topMargin for the VPNInterLabel in VPNCheckBoxRow
+            mouseClick(checkboxrow, checkboxrow.width + 1, 2)
```
flozia, line 64: suggestion (non-blocking): That’s definitely a smart alternative for emulating a click on the label. Just in case you are hitting the label indirectly here because the label is not accessible from the test: If you add an alias to the root item of `VPNCheckBoxRow` (e.g. `property alias checkboxLabel: label`) you should be able to access it via `checkboxrow.checkboxLabel`.

flozia (approved):
Works as expected! Just left a suggestion for the test.

tests/qml/tst_VPNCheckBoxRow.qml
MattLichtenstein: yep, totally a valid option - I was just trying to get fancy with it to avoid adding an extra property since the label extends all the way to the end of the component

PR #3510, opened 2022-05-11
Reduce the level of information shared through logs

bakulf:
Fixes - VPN-2215


birdsarah (approved):
I'm hoping that obfuscating the response body doesn't cause debugging issues with other network requests. Maybe the consumer of the network request errors can be responsible for logging additional extra info if needed.

PR #3516, opened 2022-05-12
Compile wasm using cmake

bakulf:


CMakeLists.txt
```diff
@@ -71,10 +71,13 @@ option(BUILD_TESTING "Build the testing tree." OFF)
 include(CTest)
 add_custom_target(build_tests)
 set_target_properties(build_tests PROPERTIES EXCLUDE_FROM_ALL $<NOT:$<BOOL:${BUILD_TESTING}>>)
-add_subdirectory(tests/auth)
-add_subdirectory(tests/nativemessaging)
-add_subdirectory(tests/unit)
-add_subdirectory(tests/qml)
+
+if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten")
```
oskirby, line 75: Out of curiosity, does building for WASM set the `CROSS_COMPILING` variable? The best approach here would be to setup the tests to be built for the host system, and failing that we should just drop them when cross compiling (which will apply to iOS and Android too).

src/networkrequest.cpp
```diff
@@ -2,27 +2,30 @@
...
-#include <QJsonDocument>
-#include <QJsonArray>
-#include <QJsonObject>
-#include <QNetworkAccessManager>
-#include <QNetworkReply>
-#include <QNetworkRequest>
-#include <QRegularExpression>
-#include <QUrl>
+// Wasm has its own networkrequest implementation
+#ifndef MVPN_WASM
```
oskirby, line 6: I have a better solution. Since this is a one-off quirk for WASM, set the `HEADER_FILE_ONLY` property on `networkrequest.cpp` in `src/cmake/wasm.cmake` which should disable compilation of this file in just this one case.

oskirby (commented):
I am amazed that this works, and requires this few changes.

PR #3536, opened 2022-05-13
Disable features

bakulf:
## Description

For 2.8.2 we want to have a way to disable the free-trial feature in case something goes wrong. With this PR, each `Feature` can be defined as flippable on and flippable off, independently. If they are flipped on/off this setting overwrites the default Feature state.

As Lesley said, this PR transforms a boolean into a quantum boolean. Because now a feature can be:
- on or off
- forced on
- forced off

To make it happen, this PR does:

1. It allows the flip on/off of features from the dev-menu.
2. Now the dev menu makes toggleable only the features that can be enabled for real
3. Similar toggles/checkboxes are for features that can be disabled
4. rename the SettingsList devModeFeatureFlags to featuresFlippedOn
5. introduce SettingsList featuresFlippedOff
6. tests

## Reference
https://mozilla-hub.atlassian.net/browse/VPN-2228

### DIFF
--- a/src/featurelist.cpp
+++ b/src/featurelist.cpp
@@ -80,11 +80,28 @@ void FeatureList::initialize() {
   m_featurelist = Feature::getAll();
 }
 
-void FeatureList::devModeFlipFeatureFlag(const QString& feature) {
-  logger.debug() << "Flipping " << feature;
+void FeatureList::toggleForcedEnable(const QString& feature) {
+  logger.debug() << "Flipping on" << feature;
+
+  const Feature* f = Feature::get(feature);
+  if (!f) {
+    logger.debug() << "Feature" << feature << "does not exist";
+    return;
+  }
+
+  if (!f->isFlippableOn()) {
+    logger.debug() << "Feature" << feature << "cannot be flipped on";
+    return;
+  }
+
+  if (f->isSupportedIgnoringFlip()) {
+    logger.debug() << "This is an internal bug. Why flipping on a pref that is "
+                      "already supported?";
+    return;
+  }
 
   auto const settings = SettingsHolder::instance();
-  QStringList flags = settings->devModeFeatureFlags();
+  QStringList flags = settings->featuresFlippedOn();
 
   logger.debug() << "Got List - size:" << flags.size();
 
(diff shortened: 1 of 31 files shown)


src/featurelist.cpp
```diff
@@ -132,43 +188,50 @@ void FeatureList::updateFeatureList(const QByteArray& data) {
...
-    return;
-  }
-
-  QJsonObject featuresObj = featuresValue.toObject();
-  for (const QString& key : featuresObj.keys()) {
-    QJsonValue value = featuresObj.value(key);
-    if (!value.isBool()) {
-      logger.error() << "Error in parsing feature enabling:" << key;
-      continue;
+  if (json.contains("featuresOverwrite")) {
```
bakulf, line 192: I decided to change the name of the attribute because I don't want to break any custom feature-enabling config.
The new approach is that, if guardian returns a list of features, _those_ must be applied, overwriting everything.

oskirby: I have concerns about this approach that it could make things difficult for us as developers. I could forsee a situation where we need Guardian to disable a feature due to some bug, and we can't turn it on as developers on order to debug it because the contents of `featuresOverwrite` takes precedence over our local settings.

tests/unit/testfeature.cpp
```diff
@@ -3,35 +3,151 @@
...
+  QVERIFY(!settingsHolder.featuresFlippedOff().contains("testFeatureB"));
+  QVERIFY(Feature::get("testFeatureB")->isSupported());
+
+  // Flipping on a feature which was off but enabled -> it becomes off
+  fl->flipFeatureOn("testFeatureB");
+  QVERIFY(!settingsHolder.featuresFlippedOn().contains("testFeatureB"));
+  QVERIFY(!settingsHolder.featuresFlippedOff().contains("testFeatureB"));
+  QVERIFY(!Feature::get("testFeatureB")->isSupported());
+
+  // Flipping on an flippable feature doesn't produce changes.
```
oskirby, line 78: I think you mean "unflippable" here in this comment.

tests/unit/testfeature.cpp
```diff
@@ -3,35 +3,151 @@
...
+  fl->flipFeatureOn("testFeatureC");
+  QVERIFY(!settingsHolder.featuresFlippedOn().contains("testFeatureC"));
+  QVERIFY(!settingsHolder.featuresFlippedOff().contains("testFeatureC"));
+  QVERIFY(!Feature::get("testFeatureC")->isSupported());
+
+  // Flipping on an flippable feature doesn't produce changes.
```
oskirby, line 84: I think you mean "unflippable" here in this comment.

src/featurelist.cpp
oskirby: Also, I wouldn't worry about changing the name in the JSON, the featurelist is behind a dev menu, so we don't have to feel too bad about potential user breakage here. We're changing the name of the underlying settings anyways, so any previous feature settings aren't going to carry over across an upgrade anyways.

oskirby (approved):
I think this is good enough to satisfy our needs for 2.8.2, though I have some questions about code clarity. In particular I'm confused by what `flipFeatureOn()` is supposed to leave the feature as.

PR #3539, opened 2022-05-14
Properly disable qmljsdebugger in release mode

andersk:
## Description

* In qmake, the `debug` variable and the `release` variable are both present by default.  We need to check which one is active, by virtue of having been set later.

  https://doc.qt.io/qt-6/qmake-test-function-reference.html#config-config

* In CMake, the `"${CMAKE_BUILD_TYPE}"` test did not do what it looks like it does.  CMake evaluates `if("${CMAKE_BUILD_TYPE}")` to `if(Release)`, which tests whether the *variable named `Release`* is defined; it isn’t, so we were unconditionally setting `CMAKE_BUILD_TYPE` to `Debug`.

  https://cmake.org/cmake/help/latest/command/if.html#variable-expansion

  We need to check whether `CMAKE_BUILD_TYPE` variable itself is defined, and we need to do so before `project()` defines it by default.

## Checklist
    
### DIFF
--- a/src/qmake/debug.pri
+++ b/src/qmake/debug.pri
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-debug {
+CONFIG(debug, debug|release) {
     # If in debug mode, set mvpn_debug flag too.
     CONFIG += mvpn_debug
 }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,17 +39,17 @@ endif()
 
 option(BUILD_DUMMY "Build for the dummy platform" OFF)
 
-project("Mozilla VPN" VERSION 2.10.0 LANGUAGES C CXX
-        DESCRIPTION "A fast, secure and easy to use VPN. Built by the makers of Firefox."
-        HOMEPAGE_URL "https://vpn.mozilla.org"
-)
-
 message("Configuring for ${CMAKE_GENERATOR}")
-if(NOT (DEFINED CMAKE_CONFIGURATION_TYPES OR "${CMAKE_BUILD_TYPE}"))
+if(NOT (DEFINED CMAKE_CONFIGURATION_TYPES OR DEFINED CMAKE_BUILD_TYPE))
     ## Ensure the build type is set for single-config generators.
     set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "default build type" FORCE)
     message("Setting build type ${CMAKE_BUILD_TYPE}")
 endif()
+
+project("Mozilla VPN" VERSION 2.10.0 LANGUAGES C CXX
+        DESCRIPTION "A fast, secure and easy to use VPN. Built by the makers of Firefox."
+        HOMEPAGE_URL "https://vpn.mozilla.org"
+)
 
 if(NOT DEFINED BUILD_ID)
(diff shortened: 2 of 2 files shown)


src/qmake/debug.pri
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-debug {
+CONFIG(debug, debug|release) {
```
bakulf, line 5: We are switching to cmake. Can you please use cmake instead? In this way you probably do not have this issue.

andersk: I can’t yet, because the current release v2.8.0 doesn’t have it, and Qt6 is not yet in nixpkgs (NixOS/nixpkgs#169296). When you cut a new release and Qt6 enters nixpkgs I will switch.

But meanwhile, this code is still present and should be fixed, because unexpectedly exposing QML debugging on [port 1234 of all interfaces](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/commands/commandui.cpp#L174-L175) is [insecure](https://doc.qt.io/qt-6/qtquick-debugging.html#qml-debugging-infrastructure).

oskirby (approved):

andersk:
I discovered that the CMake build had a similar misconfiguration for an entirely different reason. The `"${CMAKE_BUILD_TYPE}"` test did not do what it looks like it does. CMake evaluates `if("${CMAKE_BUILD_TYPE}")` to `if(Release)`, which tests whether the *variable named `Release`* is defined; it isn’t, so we were unconditionally setting `CMAKE_BUILD_TYPE` to `Debug`.

I’ve added a fix for this second bug.

Please make sure the 2.9.0 release won’t expose the QML debugger on 0.0.0.0:1234 for all users. Also, if it’s so easy to misconfigure build systems into debug mode, perhaps we should not have debug builds start this insecure listener automatically—there’s a reason it ordinarily requires a command line argument.

bakulf (approved):
If Owen says yes, it's OK

oskirby (approved):

andersk:
Ping. @strseb merged this into `releases/2.9.0` at #3803, but this still needs to be merged into `main` here to avoid regressing in a future 2.10 release.

ahal:
This broke the Linux Taskcluster builds:
https://treeherder.mozilla.org/jobs?repo=mozilla-vpn-client

Unfortunately they didn't run on this PR due to the "collaborators only" policy being set. This is a known pain point (currently tc-github only supports a "public" or "collaborators" policy). I have an [RFC proposal](https://github.com/taskcluster/taskcluster-rfcs/pull/175) to make this situation better, though not sure when someone will have time to implement it.

andersk:
This was mismerged. Look at the difference between the commit I pushed here (258eb87d01ad1ced55e14058fd122b29f861bfca) and the commit that was merged (9e1e3d383f7eb532290a258d9d7ae68a50d9080e).

andersk:
And that explains the failure I see in those taskcluster logs.

```
[task 2022-07-13T18:12:53.857Z] Computing the version... 2.10.0
[task 2022-07-13T18:12:53.857Z] 2.10.0
[task 2022-07-13T18:12:53.859Z] Update the submodules...
…
[task 2022-07-13T18:12:56.702Z] Creating the working directory... done.
[task 2022-07-13T18:12:56.702Z] Generating glean samples...
[task 2022-07-13T18:12:56.703Z] scripts/linux/script.sh: line 167: cd: too many arguments
```

Fixed in #4015.

PR #3553, opened 2022-05-16
Use Authentication in-app for the functional tests

bakulf:
## Description

To improve our functional tests, first step is to make the auth faster. Let's use the in-app flow.

## Reference

VPN-2218

## Checklist
    

tests/functional/helper.js
```diff
@@ -243,36 +243,45 @@ module.exports = {
 
   // TODO - The expected staging urls are hardcoded, we may want to
   // move these hardcoded urls out if testing in alternate environments.
-  async authenticate(clickOnPostAuthenticate = false, acceptTelemetry = false) {
+  async authenticateInBrowser(
+      clickOnPostAuthenticate = false, acceptTelemetry = false) {
+console.log("A");
```
oskirby, line 248: Please remove these debugging `console.log()` messages.

tests/functional/testHeartbeat.js
```diff
@@ -27,16 +27,18 @@ describe('Backend failure', function() {
     await vpn.wait();
   });
 
-  it('Backend failure in the help menu', async () => {
-    await vpn.waitForElementProperty('getHelpLink', 'visible', 'true');
-    await vpn.clickOnElement('getHelpLink');
-    await vpn.waitForElementProperty('getHelpLink', 'visible', 'false');
+  /* TODO
```
oskirby, line 30: What's the reason for this test being commented out? We should either document the reason or delete the code.

src/localizer.cpp
```diff
@@ -215,9 +215,6 @@ QHash<int, QByteArray> Localizer::roleNames() const {
 }
 
 int Localizer::rowCount(const QModelIndex& index) const {
-  if (!index.isValid()) {
```
oskirby, line 218: I think this has broken the localizer unit tests, could you explain the reasoning for the change? Maybe we need to rethink how the localizer is tested.

oskirby (changes requested):

tests/functional/testHeartbeat.js
bakulf: I'm about to file an issue. This test is totally broken but fixing it is out of scope.

tests/functional/helper.js
bakulf: Yes I was debugging something to fix the in-browser auth. We have an issue in the iOS auth flow.

src/localizer.cpp
bakulf: Are you able to see the languages? Because in the latest build the languages, at least on my computer are totally empty if I don't apply this change.

tests/functional/testHeartbeat.js
bakulf: When this PR is reviewed, before landing I'll file the issue and I'll write the ID here.

src/localizer.cpp
bakulf: Actually, I confirm it, `rowCount()` should return 0 only when used in tables. Any other `rowCount()` impl ignores the model index.

oskirby: You are correct, without this change I am not seeing the languages.

lesleyjanenorton (approved):
🎉

Issue #3537, opened 2022-05-14
Restore gcov for qml, lottie, auth and native messaging unit-tests


data-sync-user:
We do not submit those grcov info files after the cmake porting

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2229)
┆author: Andrea Marchesini

PR #3350, opened 2022-04-19
Experiment: Building with CMake

oskirby:
## Description
This is an experiment to see if CMake can work as a viable build system, and one that can be used to clean up the myriad of scripts and obscure tools needed to build the VPN client.

TODO List:
- [x] Build on Linux
- [x] Build on Windows
- [ ] Build on MacOS
- [ ] Build for Android
- [ ] Build for iOS
- [x] Build for functional testing
- [x] Build the MAC extension
- [x] Build Unit/Auth/QML tests
- [x] Update CI tests to use CMake
- [x] Build the Ubuntu/Debian packages
- [x] Build the Windows installer
- [ ] Build the MacOS bundle

## Reference
GitHub issue #3331
JIRA Issue [VPN-2095](https://mozilla-hub.atlassian.net/browse/VPN-2094)

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #3557, opened 2022-05-17
Fix auth tests and improve unit testing

oskirby:
## Description
During the omnibus CMake PR #3350 we neglected to upload code coverage for many tests and our test coverage appears to have degraded badly as a result.

And a recent change to Guardian has meant that we now examine the client version in the user agent string. While this is an awesome simplification of some gnarly backend code, it resulted in unit test breakage due to the use of a mocked client version string.

Changes to address these two bugs include:
 - Upload coverage artifacts from all unit tests (auth, QML, native messaging and unit).
 - Use the genuine `APP_VERSION` and `BUILD_ID` in all tests except QML (fixes the auth tests via CMake).
 - Create a `mocconstants.cpp` mock for QML testing, where we need a predictable version.
 - Update qmake builds to generate a `version.h` file, just like CMake (fixes the auth tests via QMake).

And along the way, we included the following for general cleanup:
 - Build tests in parallel on Linux.
 - Autogenerate `BUILD_ID` on windows in a manner consistent with all the UNIX-y platforms.
 - Cache `grcov` between workflow runs, which should save about 4mins before testing starts.
 - Add a `Log::operator<<(const QJsonObject&);`
 - Convert Lottie to a standalone CMake project.
 - Use CMake for the Windows and Linux lottie tests in CI.
 - Upload code coverage for Lottie tests on Linux.

## Reference
GitHub: #3537
JIRA: [VPN-2229](https://mozilla-hub.atlassian.net/browse/VPN-2229)

## Checklist
    

src/logger.cpp
```diff
@@ -46,6 +47,11 @@ Logger::Log& Logger::Log::operator<<(const QStringList& t) {
   return *this;
 }
 
+Logger::Log& Logger::Log::operator<<(const QJsonObject& t) {
+  m_data->m_ts << QJsonDocument(t).toJson(QJsonDocument::Indented) << ' ';
```
bakulf, line 51: why the extra space at the end?

bakulf (approved):

src/logger.cpp
oskirby: This just maintains consistency with all the other log `operator<<()` methods, which also include a space at the end.

Personally, I find it ever so slightly obnoxious that the logger does this.

Issue #3367, opened 2022-04-22
Ubuntu 20.04 / Elementary OS 6.1 segmentation fault  mozillavpn


immutable-pro:
When opening the Mozilla VPN client, installed from PPA

```
➜  ~ mozillavpn 
[22.04.2022 20:27:13.670] Info: (main - Command) MozillaVPN 2.8.0
[22.04.2022 20:27:13.670] Info: (main - Command) User-Agent: MozillaVPN/2.8.0 (sys:elementary 6.1)
[22.04.2022 20:27:13.738] Warning: QApplication: invalid style override 'adwaita' passed, ignoring it.
	Available styles: Windows, Fusion
[22.04.2022 20:27:13.746] Error: (main - EventListener) Connection failed.
[22.04.2022 20:27:13.795] Info: (networking - NetworkRequest) Imported cert from: ISRG Root X1
[22.04.2022 20:27:13.805] Info: (captiveportal - CaptivePortalDetection) Not connected, starting background captive-portal Monitor
[22.04.2022 20:27:13.816] Warning: QQmlApplicationEngine failed to load component
[22.04.2022 20:27:13.816] Warning: qrc:/ui/main.qml: module "adwaita" is not installed (main.qml)
[1]    21580 segmentation fault  mozillavpn
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2113)
┆author: Data Sync

peteruithoven:
Related / duplicate issue: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3447

peteruithoven:
I'm also running into this issue on 2 laptops running elementary OS 6.1 Jólnir (Built on Ubuntu 20.04.3 LTS). Seeing [this comment](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3447#issuecomment-1118750514) I thought it might be interesting to share GPU/CPU info (`cat /proc/cpuinfo`):
<details>
<summary>Dell XPS 9550</summary>

```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 94
model name	: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
stepping	: 3
microcode	: 0xea
cpu MHz		: 2600.000
cache size	: 6144 KB
physical id	: 0
siblings	: 8
core id		: 0
cpu cores	: 4
apicid		: 0
[...]

PR #3558, opened 2022-05-17
Clear QT_STYLE_OVERRIDE if the style was not found.

oskirby:
## Description
Some distros may set `QT_STYLE_OVERRIDE` by default when attempting to apply a uniform style their Qt5 applications. However, this style may not necessarily exist for Qt6, which will result in a failure to load QML content. To work around this issue, check that the style exists before creating the `QApplication`, and unset the environment variable if no such style was found.

The QML fails to load `QtQuick.Controls` when this situation occurs.

## Reference
Github: #3367
JIRA: [VPN-2113](https://mozilla-hub.atlassian.net/browse/VPN-2113)

## Checklist
    
### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -26,6 +26,7 @@
 
 #include <QApplication>
 #include <QIcon>
+#include <QStyleFactory>
 #include <QTextStream>
 
 namespace {
@@ -147,6 +148,9 @@ int Command::runQmlApp(std::function<int()>&& a_callback) {
   }
 
   qInstallMessageHandler(LogHandler::messageQTHandler);
+
+  // Ensure that external styling hints are disabled.
+  qunsetenv("QT_STYLE_OVERRIDE");
 
   logger.info() << "MozillaVPN" << Constants::versionString();
   logger.info() << "User-Agent:" << NetworkManager::userAgent();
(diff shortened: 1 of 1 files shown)


linux/debian/control.qt6
```diff
@@ -17,6 +17,8 @@ Build-Depends: debhelper (>= 9.20160709),
                libqt6opengl6-dev (>=6.2.0~),
                libqt6svg6-dev (>=6.2.0~),
                libqt6websockets6-dev (>=6.2.0~),
+               libssl-dev,
```
bakulf, line 20: Unrelated change

src/command.cpp
```diff
@@ -181,6 +182,12 @@ int Command::runQmlApp(std::function<int()>&& a_callback) {
 
   qInstallMessageHandler(LogHandler::messageQTHandler);
 
+  QString styleName = qEnvironmentVariable("QT_STYLE_OVERRIDE");
```
bakulf, line 185: Can we simply unset it always? Maybe on Linux only

linux/debian/control.qt6
oskirby: I can move this into a separate PR

bakulf (approved):

PR #3564, opened 2022-05-18
Fix and improve unit tests

oskirby:
## Description
A recent update to guardian makes use of the client's version during the authentication flow, which was an awesome simplification. Unfortunately, this resulted in unit test breakage due to the use of a mocked client version. So let's remove it and use the real client version instead.

## Checklist
    

oskirby:
On second thought, I am going to merge this into #3557 and bundle a bunch of unit testing cleanup together.

PR #3565, opened 2022-05-18
Guide view for tips and tricks

MattLichtenstein:
## Description

- Introduces a guide view when clicking on a guide card in the tips and tricks page

| The top uses and benefits of using Mozilla VPN  | Is my VPN connected to the right server location? | How do I add and remove devices? | Connecting to other devices while using VPN |
| ------------ | ----------------- | --------------- | --------------- |
| Screen Shot 2022-05-18 at 3 59 53 PM | Screen Shot 2022-05-18 at 3 59 59 PM | Screen Shot 2022-05-18 at 4 00 05 PM | Screen Shot 2022-05-18 at 4 00 12 PM

## Reference

[VPN-2121: Implement Guide UI](https://mozilla-hub.atlassian.net/browse/VPN-2121)

## Checklist
    

nebula/ui/components/VPNFlickable.qml
```diff
@@ -106,7 +106,7 @@ Flickable {
         minimumSize: 0
 
         opacity: hideScollBarOnStackTransition && (vpnFlickable.StackView.status !== StackView.Active) ? 0 : 1
-        visible: !windowHeightExceedsContentHeight
```
MattLichtenstein, line 109: The old logic here basically says only show scroll bars when `flickable.contentHeight > window.height`. This logic for showing scroll bars didn't make a whole lot of sense to me. What if the flickable's viewport is only 100px tall, and needs to show 200px of content? 200px is surely going to be smaller than the window height, but a scroll bar should still be shown because it is scrollable.

We seem to typically base `flickable.interactive` on `flickable.contentHeight > flickable.height`, meaning, if all the content can't be shown in the viewport, allow scrolling. We currently always show scrolls bars when scrolling is available, so this remains consistent with what already exists

flodolo (approved):

src/ui/settings/ViewGuide.qml
```diff
@@ -0,0 +1,239 @@
...
+    ColumnLayout {
+        anchors.fill: parent
+
+        spacing: 0
+
+        Rectangle {
+            id: imageBg
+
+            Layout.fillWidth: true
+            Layout.preferredHeight: Screen.height * 0.33
```
MattLichtenstein, line 24: needs to be changed to window.height

src/models/guide.cpp
```diff
@@ -1,4 +1,4 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
+﻿/* This Source Code Form is subject to the terms of the Mozilla Public
```
bakulf, line 1: What is this change?

src/models/guide.h
```diff
@@ -25,6 +25,8 @@ class Guide final : public QObject {
 
   static bool evaluateConditions(const QJsonObject& conditions);
 
+  QString titleId() const { return m_titleId; }
```
bakulf, line 28: `const QString& titlteId() const { return m_titleId; }`

src/ui/settings/ViewGuide.qml
```diff
@@ -0,0 +1,243 @@
...
+                        Component {
+                            id: listBlock
+
+                            VPNInterLabel {
+                                property var guideBlock
+                                property string listType
+
+                                property var tagsList: {
+                                    var list = []
+                                    guideBlock.subBlockIds.forEach(subBlockId => {
```
bakulf, line 188: You can also use `map`, I guess:
```
guideBlock.subBLockIds.map(subBlockId => `<li>${VPNl18n[subBlockId]}</li>`);
```

bakulf (approved):

lesleyjanenorton (approved):

src/models/guide.cpp
flodolo: > $file guide.cpp
> guide.cpp: c program text, Unicode text, UTF-8 (with BOM) text

I think it's adding BOM, which shouldn't happen. Which text editor are you using? VSCode shows clearly if a file is _with BOM_ or not

flozia (approved):
Looks really good!

I only noticed that you are already using `VPNTheme` variables in a lot of places, but there are a few margins and spacings that still have hardcoded values. Feel free to ignore the inline suggestions.

src/models/guide.cpp
MattLichtenstein: I believe @flodolo is right, though not sure how it happened. I use Qt Creator and thought I might've had some messed up auto-format settings but everything looks good.

Anyways, I will revert this!

src/ui/main.qml
```diff
@@ -138,7 +137,6 @@ Window {
         anchors.top: parent.top
         anchors.topMargin: iosSafeAreaTopMargin.height + wasmMenuHeader.height
         height: safeContentHeight
-        clip: true
```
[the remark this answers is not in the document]
MattLichtenstein: I am using negative margins in ViewGuide to get the colored background to show up behind the status bar, clipping would prevent that

src/ui/settings/ViewGuide.qml
```diff
@@ -0,0 +1,243 @@
...
+    ColumnLayout {
+        anchors.fill: parent
+
+        spacing: 0
+
+        Rectangle {
+            id: imageBg
+
+            Layout.fillWidth: true
+            Layout.preferredHeight: window.height * 0.33
```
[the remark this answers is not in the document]
MattLichtenstein: It's meant to be the whole window - this is the background color behind the hero image

Issue #3528, opened 2022-05-13
Github\main branch builds cannot be installed on Windows 10


data-sync-user:
**Platforms:**

* Windows 10;

**Prerequisites:**

* Have the [last build](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/2315005671) from GitHub/main branch downloaded;

**Steps to reproduce:**

1. Install the .msi;
1. Observe the behavior;

**Expected results:**

* Mozilla VPN app installation is finalized with success;

**Actual results:**

* Mozilla VPN app installation fails with an error.

**Notes:**

* Reproducing on Windows 11 also;
* Other platform work;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2224)
┆author: Valentina Virlics

data-sync-user:
➤ Owen Kirby commented:

Most likely this is a missing runtime DLL. If I had to take a random guess it’s libssl.dll and libcrypto.dll I had a TODO message about them in my CMake PR and I couldn’t tell if they were still necessary without spinning up a VM.

There’s also an outside chance that we might be missing a Qt runtime DLL that the Wix installer isn’t gathering after windeployqt does its magic - That would be a harder thing to fix.

data-sync-user:
➤ Sebastian Streich commented:

Those are missing but i’ve tried feeding the MozillaVPN.exe that cmake produces into the legacy-installer build process, with the same result. Oddly enough when i open the exe it produces windows wants to download some dependency from the windows store - but it never opens or tells me what it’s looking for :c

!image-20220518-180130.png|width=853,height=616!

data-sync-user:
➤ Valentina Virlics commented:

Verified as fixed while using the last build from mai ( https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/2360315073 )n branch.

Installation is finalized with success.

Issue #3515, opened 2022-05-12
The windows cmake installer must be improved


data-sync-user:
During the cmake refactoring, we have introduced a change in the Mozilla VPN windows exe file. Now the filename has a space between "Mozilla" and "VPN".
This is great, but there are a few remaining things to do:

1. the registry key for the Windows Error Reporing should refer to Mozilla<space>VPN.exe
2. the installer must kill both MozillaVPN.exe and Mozilla<space>VPN.exe
3. We need to update the prod WXS file too

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2211)
┆author: Andrea Marchesini

data-sync-user:
➤ Andrea Marchesini commented:

If we decide to release cmake win builds for 2.9, this issue must be fixed.

If we postpone it, we can reduce the priority to high/medium

data-sync-user:
➤ Owen Kirby commented:

Except for the creation of a production WXS file, I think this was completed by PR #3567 ( https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3567 )

data-sync-user:
➤ Andrea Marchesini commented:

This issue is already fixed.

PR #3567, opened 2022-05-19
Fix the cmake-generated windows installer

oskirby:
## Description
Some cleanup of the executable paths in the Wix installer after renaming the client from `MozillaVPN.exe` to `Mozilla VPN.exe` and the sources for the GPU check command got lost in the rebase when merging PR #3350 

## Reference
GitHub: #3528 ([VPN-2224](https://mozilla-hub.atlassian.net/browse/VPN-2224))
GitHub: #3515 ([VPN-2211](https://mozilla-hub.atlassian.net/browse/VPN-2211))

## Checklist
    
### DIFF
--- a/.github/workflows/windows-build.yaml
+++ b/.github/workflows/windows-build.yaml
@@ -37,16 +37,9 @@ jobs:
         run: |
           auth_header="$(git config --local --get http.https://github.com/.extraheader)"
           git clone https://github.com/mozilla-mobile/qt_static_windows --depth 1
-          cd qt_static_windows
-          cat qt6* > qt_static.tar.bz2
-          tar xf qt_static.tar.bz2
-          tar xf msm.tar.gz
           mkdir /c/MozillaVPNBuild
-          cp -r * /c/MozillaVPNBuild
-          cd ..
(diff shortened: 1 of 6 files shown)


bakulf (approved):

bakulf:
We still need to fix the `kill task` part. If we are updating to 2.9, more likely the running process will be "MozillaVPN.exe". We need to kill that too and not just "Mozilla VPN.exe". Follow up?

oskirby:
> We still need to fix the `kill task` part. If we are updating to 2.9, more likely the running process will be "MozillaVPN.exe". We need to kill that too and not just "Mozilla VPN.exe". Follow up?

I wasn't sure how to test this feature. I was never able to get the `CloseApplication` step to do anything other than produce an error, which gets silently ignored by the installer. Is there a particular version I should try upgrading from, that might clause this clause to do anything?

bakulf:
> 

In theory, if you have MozillaVPN 2.8 and you try to update to 2.9, 2.8 should be killed. But I guess it's not. Maybe an STR is the following:

- Have 2.8 running
- run the 2.9 installer (ignore the update system in app, just run the MSI file)
- 2.8 should be killed

Note that this can be a follow up :)

Issue #3568, opened 2022-05-19
VPN blank UI after 2.8.3 update


JKingZapata:
Users are seeing a white screen when opening the VPN after having updated to 2.8 and cannot connect to the client, select a location, or access settings. Some options are accessible by right-clicking the VPN icon in the taskbar such as connecting to the latest location used, 

**Technical Information**
Platform Windows 10/11
FX product Mozilla VPN 2.8.3

**Steps to Reproduce**
1. Have Mozilla VPN 2.7 installed
2. Open VPN client
3. Click on VPN update banner/icon to install latest version

**Actual Results**
-  Once update completes the VPN client reopens and only shows a blank screen

**Expected Results**
- VPN Client interface should display options to connect, select location and access settings

![screenshot 4](https://user-images.githubusercontent.com/72564534/169315553-05550d8b-95a1-4832-8125-281b5916b03c.jpg)

[mozillavpn-2022-5-18_1.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/8730046/mozillavpn-2022-5-18_1.txt)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2250)

data-sync-user:
➤ Juan Zapata commented:

related to previous bug https://mozilla-hub.atlassian.net/browse/VPN-2106 ( https://mozilla-hub.atlassian.net/browse/VPN-2106|smart-link )

data-sync-user:
➤ Magdalena Schwaighofer commented:

atting logs as per Trello card - also DxDiag file

[^DxDiag.txt]
[^mozillavpn-2022-5-24.txt]

data-sync-user:
➤ Magdalena Schwaighofer commented:

please find another set of logs and DxDiag file from user attached 

[^DxDiag (1).txt]
[^mozillavpn-2022-5-24_1.txt]

data-sync-user:
➤ Juan Zapata commented:

Attaching another Dxdiag file

[^DxDiag (54ed12aa-7ac7-475b-91d7-4774305105e3).txt]

data-sync-user:
➤ Manuel Calidonio commented:

Attaching Dxdiag

[^DxDiag1.txt]

data-sync-user:
➤ Magdalena Schwaighofer commented:

Attaching Dxdiag

[^DxDiag (2).txt]

data-sync-user:
[...]

PR #3572, opened 2022-05-19
Check Direct3D shader version to select software rendering

oskirby:
## Description
In the 2.8.3 update, it seems that the GPU check is not running correctly for some users, resulting in a situation where they continue to receive the blank window after updating to Qt6. It's hard to know why this is the case, but maybe we can take another approach and check for the Direct3D version as a mechanism to determine if software rendering is required.

The common denominator for the affected users seems to be that they all have graphics cards that only support Direct3D version 10 or older.

**Note: This PR is branched from releases/2.8.3** it will need to be rebased onto a different release should we intend on getting this out to our users.

## Reference
GitHub: #3568 ([VPN-2250](https://mozilla-hub.atlassian.net/browse/VPN-2250))

## Checklist
    

src/command.cpp
```diff
@@ -32,31 +31,6 @@
...
 namespace {
 Logger logger(LOG_MAIN, "Command");
-
-#ifdef MVPN_WINDOWS
-QSGRendererInterface::GraphicsApi maybeUseCustomGraphicApi() {
-  QString location =
-      QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-  QDir appDataLocation(location);
-
-  QFile gpuCheckSettings = appDataLocation.filePath("moz.vpn.gpucheck");
```
bakulf, line 42: We can revert the gpucheck thing too then.

oskirby:
I've given this a test on my T420s with an i7-2620 and Intel HD3000 and it correctly detects the D3D version of 10.0 with shader version 4 and falls back to software rendering. So I think this should be a viable solution to the problem.

bakulf (approved):

PR #3586, opened 2022-05-23
First experiments with addons

bakulf:


lesleyjanenorton (approved):
Alive and kicking!
Screen Shot 2022-05-23 at 10 57 17 PM
This is neat.

PR #3620, opened 2022-05-25
Use toolchain tasks for building openssl and qt

ahal:
Currently the openssl and qt tasks needed by the Windows and Mac builds are modeled after toolchains but contain a lot of custom logic since Taskgraph didn't support toolchain tasks on `generic-worker`. As of Taskgraph 1.6.0 this is no longer the case and toolchain tasks are supported there (along with powershell scripts).

Using Taskgraph's toolchain tasks allows us to simplify a bunch of logic in `mozilla-vpn-client`.


ahal:
I think I should hold off landing this until the Mac worker is fixed. Otherwise there's a high likelihood I break it without noticing..

I'm unsure what's up with the Windows qt task and why it would timeout trying to download that URL. Will investigate this one later.

PR #3626, opened 2022-05-26
Start update with a detached process - VPN-116

bakulf:
On windows, we see the installer killed together with the app. This happens because the program is still attached to the VPN client instance. Let's use `startDetached` instead.


oskirby (approved):
Looks good to me, but I am a little bit concerned that we are no longer able to gather logs from the MSI installation in the event that something goes wrong (not that our logs have every been particularly reliable in this scenario anyways).

PR #3630, opened 2022-05-26
ci(taskcluster): temporarily disable caches for mac builds and toolch…

ahal:
…ains

This works around the recent generic-worker related breakage by
disabling caches entirely. This does mean the tasks will take longer to
run, but at least they'll run!

Once the root issue
(https://github.com/taskcluster/taskcluster/issues/5396) is fixed, we
can turn caches back on again.


ahal:
Aki suggested disabling caches and it worked.. I should have thought of this a long time ago.

hneiva (approved):

strseb (approved):

PR #3638, opened 2022-05-28
Generate tutorial addons

bakulf:


scripts/utils/generate_strings.py
```diff
@@ -388,10 +218,6 @@ def serialize(string):
         help='YAML strings file to process')
     parser.add_argument('-o', '--output', metavar='DIR', type=str, action='store',
         help='Output directory for generated files')
-    parser.add_argument('-g', '--guides', metavar='DIR', type=str, action='store',
-        help='Parse JSON guides from DIR')
-    parser.add_argument('-t', '--tutorials', metavar='DIR', type=str, action='store',
-        help='Parse JSON tutorials from DIR')
     args = parser.parse_args()
 
     ## If no source was provided, find it relative to this script file.
```
flodolo, line 223: I can't comment on the actual string, but please fix the fallback to actually use rootpath (as explained in the l10n original issue)

scripts/addon/build.py
```diff
@@ -0,0 +1,361 @@
+#! /usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import argparse
+import json
+import os
+from pathlib import Path
```
flodolo, line 9: This is not used.

scripts/addon/build.py
```diff
@@ -0,0 +1,361 @@
...
+import xml.etree.ElementTree as ET
+import tempfile
+import shutil
+import sys
+
+comment_types = {
+    "text": f"Standard text in a guide block",
```
flodolo, line 16: f-strings are not necessary when there's no variable in them (there are a few of these in this file)

```suggestion
    "text": "Standard text in a guide block",
```

scripts/addon/build.py
```diff
@@ -0,0 +1,361 @@
...
+    "ulist": f"Bullet unordered list item in a guide block",
+    "olist": f"Bullet ordered list item in a guide block",
+}
+
+
+def retrieve_strings_tutorial(manifest, filename):
+    tutorial_strings = {}
+
+    tutorial_json = manifest["tutorial"]
+    if not "id" in tutorial_json:
```
flodolo, line 27:
```suggestion
    if "id" not in tutorial_json:
```

nit: [Pythonic way](https://www.flake8rules.com/rules/E713.html)

scripts/addon/build.py
```diff
@@ -0,0 +1,361 @@
...
+    return guide_strings
+
+
+def write_en_language(filename, strings):
+    ts = ET.Element("TS")
+    ts.set("version", "2.1")
+    ts.set("language", "en")
+
+    context = ET.SubElement(ts, "context")
+    name = ET.SubElement(context, "name")
```
flodolo, line 158: This is not used anywhere.

scripts/addon/build.py
```diff
@@ -0,0 +1,361 @@
...
+            copy_files(file_path, dir_path)
+
+
+def get_file_list(path, prefix):
+    file_list = []
+    for file in os.listdir(path):
+        file_path = os.path.join(path, file)
+        if (
+            os.path.isfile(file_path)
+            and not file_path.endswith(".ts")
```
flodolo, line 204: `.endswith` accepts also a tuple, so this would work

```
        if (
            os.path.isfile(file_path)
            and not file_path.endswith((".ts", ".qrc", ".rcc"))
       )
```

scripts/addon/generate_all.py
```diff
@@ -0,0 +1,43 @@
...
+generated_path = os.path.join(generated_path, "addons")
+if not os.path.isdir(generated_path):
+    os.mkdir(generated_path)
+
+for file in os.listdir(addons_path):
+    if not file.startswith("tutorial_") and not file.startswith("guide_"):
+        continue
+    addon_path = os.path.join(addons_path, file, "manifest.json")
+
+    build_args = [sys.executable, build_path, addon_path, generated_path]
```
flodolo, line 39: nit: that's a `build_cmd`,  not really a `build_args` at this point.

scripts/utils/generate_strings.py
```diff
@@ -9,14 +9,6 @@
 import json
```
flodolo, line 9: `json` and `re` are unused

scripts/utils/generate_strings.py
flodolo: About this line though: comments should only have one `#`.

flodolo (commented):
Don't forget to use black, at least on the files you're touching (all Python files would need a pass…)

scripts/addon/build.py
bakulf: Right, but the TS file seems to need an empty `<name/>`

flodolo: I assume you should be able to just call `ET.SubElement(context, "name")`, without assigning it to a variable?

scripts/utils/generate_strings.py
```diff
@@ -381,34 +209,26 @@ def serialize(string):
...
-        help='Parse JSON guides from DIR')
-    parser.add_argument('-t', '--tutorials', metavar='DIR', type=str, action='store',
-        help='Parse JSON tutorials from DIR')
     args = parser.parse_args()
 
-    ## If no source was provided, find it relative to this script file.
+    # If no source was provided, find it relative to this script file.
     if args.source is None:
         rootpath = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir))
         args.source = os.path.join('translations', 'strings.yaml')
```
flodolo, line 224:
```suggestion
        args.source = os.path.join(rootpath, 'translations', 'strings.yaml')
```

flodolo (commented):
I think `generate_ts.sh` should generate the .ts files also for the addons, and the translations workflow should upload all of them?

flodolo:
> I think `generate_ts.sh` should generate the .ts files also for the addons, and the translations workflow should upload all of them?

We also need to define the behavior between main and branches. We need to support branches, in case an add-on gets removed.

Do we merge the ts if there is already a file with the same name? That seems like the most sensible approach.

flodolo:
Not 100% sure it works or how the resulting XLIFF looks like, but this is the idea
https://github.com/flodolo/mozilla-vpn-client/commit/2a0f7f9399ee311ec7fdeb5875ccaeeb91c9368c

BTW, how are you going to import back translations? There's nothing in the current PR.

scripts/addon/build.py
```diff
@@ -0,0 +1,359 @@
...
+
+def write_en_language(filename, strings):
+    ts = ET.Element("TS")
+    ts.set("version", "2.1")
+    ts.set("language", "en")
+
+    context = ET.SubElement(ts, "context")
+    ET.SubElement(context, "name")
+
+    for key, value in strings.items():
```
flodolo, line 159: I noticed that this generates XML like this

```
    <message id="guide.how_to_vpn.title">
      <source>The top uses and benefits of using Mozilla VPN</source>
      <translation type="unfinished"/>
      <extracomment>Title for a guide view</extracomment>
    </message>
```

In comparison, this is a string from the normal .ts file

```
    <message id="vpn.settings.aboutUs">
        <location filename="nebula/ui/components/VPNAboutUs.qml" line="19"/>
        <location filename="nebula/ui/components/VPNAboutUs.qml" line="56"/>
        <location filename="src/ui/settings/ViewSettingsMenu.qml" line="135"/>
        <source>About us</source>
    </message>
````

This results in XLIFF with empty origin attribute

```
  <file original="" datatype="plaintext" source-language="en" target-language="en-US">
```

Could you add a location? Even something hardcoded, e.g. `experiment.qml` should probably be enough.

File: src/cmake/linux.cmake [hunk 11]
```diff
@@ -112,6 +112,8 @@ install(FILES platforms/linux/daemon/org.mozilla.vpn.conf
 install(FILES platforms/linux/daemon/org.mozilla.vpn.dbus.service
     DESTINATION ${CMAKE_INSTALL_DATADIR}/dbus-1/system-services)
 
+add_definitions(-DADDONS_PATH=\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/mozillavpn/addons\")
```
oskirby, line 115: We should really be moving away from using hard-coded paths like this, and we should instead perform a runtime search using the `XDG_DATA_DIRS` environment variable.

scripts/addon/build.py
```diff
@@ -0,0 +1,362 @@
...
+        xliff_path = os.path.join(
+            i18n_path, locale, "addons", manifest["id"], "locale.xliff"
+        )
+
+        if os.path.isfile(xliff_path):
+            locale_file = os.path.join(tmp_path, "i18n", f"locale_{locale}.ts")
+            os.system(f"{lconvert} -if xlf -i {xliff_path} -o {locale_file}")
+            os.system(f"{lrelease} -idbased {locale_file}")
+
+    print("Generate the RC file...")
```
oskirby, line 346: The comment should probably read "Generate the RCC file..."

scripts/addon/build.py
```diff
@@ -0,0 +1,362 @@
...
+        if os.path.isdir(file_path):
+            file_list += get_file_list(file_path, f"{prefix}{file}/")
+
+    return file_list
+
+
+parser = argparse.ArgumentParser(description="Generate an addon package")
+parser.add_argument(
+    "source",
+    metavar="manifest",
```
oskirby, line 221: I usually make the `metavar` uppercase to distinguish it from the argument itself.

oskirby (approved):
Overall, I think this is pretty good and aside from a few minor changes I don't see any major problems with merging this work. We should still have enough time for these changes to settle before the fix and code freeze deadlines.

flodolo (approved):

flodolo:
Note: as mentioned before, this is still missing the "import strings" part. But I'd be OK with keeping that as a separate follow-up, when you actually have the strings in the repo to use.

src/cmake/linux.cmake
bakulf: I removed this part.

oskirby (approved):
Looks good!

.github/workflows/wasm.yaml
```diff
@@ -93,9 +93,52 @@ jobs:
...
+    env:
+      QTVERSION: 6.2.4
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
flodolo, line 107: I think I've commented in a commit before: why node here, and not Python?

PR #3640, opened 2022-05-30
ci: temporarily disable production mac signing tasks

ahal:
The level 3 pools these tasks are running on are using an outdated Chain
of Trust private key. We'll need to provision the workers with the new
key before we these will be useful.

Instead of turning Mac signing off completely, just turning them into
dep signing tasks to help avoid regressions.


ahal:
Tested on `staging-mozilla-vpn-client` here:
https://firefox-ci-tc.services.mozilla.com/tasks/G82na7nXRXWGI_ombISt_g

hneiva (approved):

ahal:
Having trouble getting a green run, but this only touches Taskcluster tasks so can't affect GH actions. Help merging would be appreciated!

ahal:
We managed to fix the build workers so this PR is no longer needed.

PR #3648, opened 2022-06-01
Expose a way to localize currency values from iso4217 - VPN-2307

bakulf:


flozia (approved):
That’s a really cool workaround! 🙂

PR #3651, opened 2022-06-01
Glean: Add Host's internet connection type to glean events

strseb:
## Description
This pr adds the ability to detect the type of network used (i.e wifi / 4g etc) and add them to the glean ping's releated to connectivity issues. 

# Todo
- [ ] Linux detection ( probably will happen in a follow-up, i can't seem to get the client to build locally :c ) 
- [ ] Data review (will happen once we're fine with the impl) 

## Reference
Data Review on : https://bugzilla.mozilla.org/show_bug.cgi?id=1774439

https://mozilla-hub.atlassian.net/browse/VPN-2279

## Checklist
    

src/platforms/ios/iosnetworkwatcher.mm
```diff
@@ -0,0 +1,117 @@
...
+#include "iosnetworkwatcher.h"
+#include "leakdetector.h"
+#include "logger.h"
+#include "mozillavpn.h"
+
+#import <Network/Network.h>
+
+namespace {
+Logger logger(LOG_IOS, "IOSNetworkWatcher");
+dispatch_queue_t sQueue = dispatch_queue_create("VPNNetwork.queue", DISPATCH_QUEUE_SERIAL);
```
bakulf, line 14: s_q

src/platforms/ios/iosnetworkwatcher.mm
```diff
@@ -0,0 +1,117 @@
...
+IOSNetworkWatcher::~IOSNetworkWatcher() {
+  MVPN_COUNT_DTOR(IOSNetworkWatcher);
+  if (mNetworkMonitor != nil) {
+    nw_path_monitor_cancel(mNetworkMonitor);
+    nw_release(mNetworkMonitor);
+  }
+}
+
+void IOSNetworkWatcher::initialize() {
+  auto mon = nw_path_monitor_create();
```
bakulf, line 30: why 'mon' variable when you can immediately use `m_networkMonitor`?

src/platforms/ios/iosnetworkwatcher.mm
```diff
@@ -0,0 +1,117 @@
...
+  auto str = server.ipv4AddrIn().toStdString();
+  auto endpoint = nw_endpoint_create_host(str.c_str(), "80");
+  auto params = nw_parameters_create_secure_udp(NW_PARAMETERS_DISABLE_PROTOCOL,
+                                                NW_PARAMETERS_DEFAULT_CONFIGURATION);
+  auto con = nw_connection_create(endpoint, params);
+  nw_connection_set_queue(con, sQueue);
+  nw_connection_set_path_changed_handler(con, ^(nw_path_t _Nonnull path) {
+    mCurrentVPNTransport = toTransportType(path);
+  });
+  nw_connection_start(con);
```
bakulf, line 115: in this way you create a connection, but how to keep it alive? Do we need to?

src/platforms/linux/linuxnetworkwatcher.h
```diff
@@ -22,6 +22,11 @@ class LinuxNetworkWatcher final : public NetworkWatcherImpl {
 
   void start() override;
 
+  NetworkWatcherImpl::TransportType getTransportType() {
+    // TODO: Find out how to do that on linux generally.
```
bakulf, line 26: file an issue. This is totally doable via DBus.

src/platforms/macos/macosnetworkwatcher.h
```diff
@@ -18,8 +22,21 @@ class MacOSNetworkWatcher final : public NetworkWatcherImpl {
 
   void checkInterface();
 
+  NetworkWatcherImpl::TransportType getTransportType() override;
+
+  void controllerStateChanged();
+
  private:
   void* m_delegate = nullptr;
+  NetworkWatcherImpl::TransportType mCurrentDefaultTransport =
```
bakulf, line 31: m_cu

src/platforms/macos/macosnetworkwatcher.mm
```diff
@@ -129,3 +140,87 @@ - (void)bssidDidChangeForWiFiInterfaceWithName:(NSString*)interfaceName {
...
+    return mCurrentDefaultTransport;
+  }
+  if (mObservableConnection != nil) {
+    return mCurrentVPNTransport;
+  }
+  // If we don't have an open tunnel-observer, mCurrentVPNTransport is probably wrong.
+  return NetworkWatcherImpl::TransportType_Unknown;
+}
+
+void MacOSNetworkWatcher::controllerStateChanged() {
```
bakulf, line 157: can you share code between the IOS and the mac impls?

src/qmake/platforms/ios.pri
```diff
@@ -52,6 +52,7 @@ LIBS += -framework UIKit
 LIBS += -framework Foundation
 LIBS += -framework StoreKit
 LIBS += -framework UserNotifications
+LIBS += -framework Network
```
bakulf, line 55: do we need permissions?

src/telemetry.cpp
```diff
@@ -37,7 +37,10 @@ void Telemetry::initialize() {
 
             emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
                 GleanSample::connectivityHandshakeTimeout,
-                {{"server", publicKey}});
+                {{"server", publicKey},
+                 {"transport", MozillaVPN::instance()
+                                   ->networkWatcher()
```
bakulf, line 42: document all these params in the data-review

bakulf (changes requested):

android/src/org/mozilla/firefox/vpn/qt/VPNNetworkWatcher.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        val connectivityManager: ConnectivityManager = con.getSystemService(Context.CONNECTIVITY_SERVICE) as(ConnectivityManager)
+        val network = connectivityManager.activeNetwork
+
+        val capabilities = connectivityManager.getNetworkCapabilities(network)
+            ?: return TransportTypes.Unknown
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
+            return TransportTypes.Unknown
+        }
+        if (capabilities.hasTransport(NetworkCapabilities.TRANSPORT_BLUETOOTH)) {
```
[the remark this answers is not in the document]
strseb: This requires "ACCESS_NETWORK_STATE" - Which we already require.

src/platforms/ios/iosnetworkwatcher.mm
strseb: I haven't found anything in the docs for that. But as it's a udp connection i would suspect it's "stateless" and therefore not really getting closed by the os for not using it.

src/mozillavpn.cpp
```diff
@@ -423,6 +423,11 @@ void MozillaVPN::maybeStateMain() {
...
 }
 
+void MozillaVPN::setEntryServerPublicKey(const QString& publicKey) {
+  logger.debug() << "Set entry-server public key:" << logger.keys(publicKey);
+  m_entryServerPublicKey = publicKey;
+}
+
 void MozillaVPN::setServerPublicKey(const QString& publicKey) {
   logger.debug() << "Set server public key:" << logger.keys(publicKey);
   m_serverPublicKey = publicKey;
```
oskirby, line 433: Though it requires a broader bit of refactoring, we should rename `m_serverPublicKey` to `m_exitServerPublicKey` to maintain consistency between these two and to make it clear which of the two servers we are referring to. In the case of single-hop, the entry and exit servers should be set to the same value.

This should eliminate a lot of `vpn->multiphop() ? entry : exit` kind of checks.

src/networkwatcher.cpp
```diff
@@ -11,6 +11,8 @@
 #include "platforms/dummy/dummynetworkwatcher.h"
 #include "settingsholder.h"
 
+#include <QMetaEnum>
+
 #ifdef MVPN_WINDOWS
```
oskirby, line 16: If every single platform now has a `NetworkWatcher` would could move this into a factory class, like we have done with other platform-specific classes (eg: `PingSenderFactory`)

src/networkwatcher.cpp
```diff
@@ -153,3 +166,9 @@ void NetworkWatcher::notificationClicked(NotificationHandler::Message message) {
     MozillaVPN::instance()->activate();
   }
 }
+
+QString NetworkWatcher::getCurrentTransport() {
+  auto type = m_impl->getTransportType();
+  QMetaEnum metaEnum = QMetaEnum::fromType<NetworkWatcherImpl::TransportType>();
+  return metaEnum.valueToKey(type);
```
oskirby, line 173: Since the transport type enums all have the form `TransportType_xxxx` should we strip off the leading `TransportType_` before this string gets consumed by glean?

src/platforms/windows/windowsnetworkwatcher.cpp
```diff
@@ -134,3 +144,28 @@ void WindowsNetworkWatcher::processWlan(PWLAN_NOTIFICATION_DATA data) {
...
+  }
+
+  switch (device.IanaInterfaceType()) {
+    case 6:
+      return TransportType_Ethernet;
+    case 71:
+      return TransportType_WiFi;
+  }
+  return TransportType_Other;
+}
```
oskirby, line 171: Please add a newline at the end of this file.

oskirby (changes requested):

bakulf (commented):
Why cannot we use the same code for VPN on and off?

strseb:
> Why cannot we use the same code for VPN on and off?
@bakulf - In case this is about iosnetworkwatcher -> The default route monitor will once the vpn is on, try to check the utun device (as thats now the default route) and will report Type::Other.

oskirby (approved):

PR #3654, opened 2022-06-01
Support for the touch events

bakulf:


src/itempicker.cpp
```diff
@@ -19,9 +19,20 @@ bool ItemPicker::eventFilter(QObject* obj, QEvent* event) {
...
+  bool isMouseEvent = event->type() == QEvent::MouseButtonPress ||
+                      event->type() == QEvent::MouseButtonDblClick ||
+                      event->type() == QEvent::MouseButtonRelease;
+  bool isTouchEvent =
+      event->type() == QEvent::TouchBegin &&
+      !!qobject_cast<QQuickWindow*>(static_cast<QTouchEvent*>(event)->target());
+
+  if (
+#if defined(MVPN_ANDROID) || defined(MVPN_IOS)
+      !isTouchEvent
```
bakulf, line 31: We need to test this with a desktop build with a touch screen.

src/itempicker.cpp
```diff
@@ -75,6 +98,46 @@ QList<QQuickItem*> ItemPicker::pickItem(QMouseEvent* event, QQuickItem* item) {
...
+    }
+  }
+
+  return list;
+}
+
+QList<QQuickItem*> ItemPicker::pickItem(QTouchEvent* event, QQuickItem* item) {
+  QList<QQuickItem*> list;
+
+  if (event->touchPoints().length() != 1) {
```
bakulf, line 111: Also this needs to be checked. Do we want to support multi-touch interactions?

lesleyjanenorton: Let's start with this for now. If folks are prevented from clicking the qmlEmitter with two fingers instead of one than maybe we reconsider.

src/itempicker.cpp
lesleyjanenorton: good point

lesleyjanenorton (approved):

PR #3664, opened 2022-06-02
Add a script to generate a translation file

bakulf:


scripts/utils/generate_ts.sh
```diff
@@ -0,0 +1,64 @@
...
+. $(dirname $0)/commons.sh
+
+print N "This script generates a 'ts' file for the location of the Mozilla VPN client"
+print N ""
+
+cd $(dirname $0)/../.. || die
+
+printn Y "Caching the dep scripts... "
+mkdir -p cache || die
+cp scripts/utils/generate_strings.py cache || die
```
flodolo, line 16: Are you caching this to work around the fact that the script is different between main and branches, or am I missing the actual reason? If that's the case, it would make more sense to have the same logic that we have in the [l10n workflow](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/workflows/update.yaml#L55-L75):
- It simplifies the whole $PARAMS definition.
- It's easier to clean up, and the same approach would work if we break things again in the future.

flodolo (commented):
Questions:
* Should this script run in automation, at least it's easy to spot if things break?
* Should this file be actually stored in the repository? That would simplify the automation even further on the l10n side.

We should have notes in the depending scripts (e.g. generate_strings.py) to make sure we don't break again so easily.

bakulf:
>     * Should this script run in automation, at least it's easy to spot if things break?
Yes. Let's add an action to run the script and count the number of strings.

>     * Should this file be actually stored in the repository? That would simplify the automation even further on the l10n side.
No, better to generate the translations. Otherwise the file will be always our of sync.

scripts/utils/generate_ts.sh
bakulf: Yes that is the point. About the params, we are going to remove them in an existing PR. `-g` and `-t` are basically gone in 2.9 already. See this PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3638
In 2.9 we are also starting the migration to cmake and soon-ish we will remove the `src.pro` file. Having 1 single dummy.pro file seems a good solution to be back-compatible.

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - 'releases/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  translations:
+    name: Translations
+    runs-on: ubuntu-20.04
```
flodolo, line 20: Can also use `ubuntu-latest` at this point.

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+    env:
+      QTVERSION: 6.2.4
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
```
flodolo, line 26: All these actions have new versions (v3 for checkout, upload-artifact)

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Checkout submodules
+        shell: bash
```
flodolo, line 29: That should be already the default on non Windows.

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Install Qt
+        shell: bash
+        run: |
+          python3 -m pip install aqtinstall
+          python3 -m aqt install-qt -O /opt linux desktop $QTVERSION
+
+      - name: Install python dependencies
+        shell: bash
+        run: |
+          pip install -r requirements.txt
```
flodolo, line 44: Unrelated: I think the requirements file should really have versions ;-)

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+        shell: bash
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+      - name: Install Qt
+        shell: bash
+        run: |
+          python3 -m pip install aqtinstall
```
flodolo, line 38: Interesting, I suppose I should follow this system to install qt also in the l10n actions.

flodolo (commented):
Leaving a few comments on the workflow, but I still need to test the script locally, and won't be able until next week (Monday is public holiday here).

flodolo:
Note: for sake of review, it would really help not to have force-pushes and squash at merge ;-)

.github/workflows/translations.yaml
bakulf: Yes, we can update the l10n actions too

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+          pip install -r requirements.txt
+
+      - name: Generating translations
+        shell: bash
+        run: |
+          export PATH=/opt/$QTVERSION/gcc_64/bin:$PATH
+          ./scripts/utils/generate_ts.sh
+
+      - name: Uploading
+        uses: actions/upload-artifact@v1
```
flodolo, line 53:
```suggestion
        uses: actions/upload-artifact@v3
```

scripts/utils/generate_ts.sh
```diff
@@ -0,0 +1,72 @@
...
+
+git branch -r
+for branch in $(git branch -r | grep origin/releases); do
+  printn Y "Importing strings from $branch..."
+
+  git checkout $branch &>/dev/null || die
+  PARAMS=
+  [ -d src/ui/guides ] && PARAMS="$PARAMS -g src/ui/guides"
+  [ -d src/ui/tutorials ] && PARAMS="$PARAMS -t src/ui/tutorials"
+  python cache/generate_strings.py -o translations/generated $PARAMS || die
```
flodolo, line 54: If you don't plan to keep the script with parameters even in 2.9, shouldn't we just get rid of $PARAMS? Also, the script on the 2.8.x branches doesn't have an `-o` output.

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  push:
+    branches:
+      - main
+      - 'releases/**'
+  pull_request:
+    branches:
+      - main
+      - 'releases/**'
```
flodolo, line 11: Can you add `workflow_dispatch:`, so it can be run manually for testing?

scripts/utils/generate_ts.sh
bakulf: 2 things:
- that part where we remove the -g and -t params has not been reviewed yet.
- right. This is why we use the cached script and not what is in the current branch.

.github/workflows/translations.yaml
flodolo: To be clear, I mean that this line is not necessary as far as I understand, since we're only running Linux ;-)

scripts/utils/generate_ts.sh
```diff
@@ -0,0 +1,72 @@
...
+RESOURCES += \$\$files(../../nebula/*.qrc, true)
+EOF
+print G "done"
+
+print Y "Generating the main translation file... "
+lupdate translations/generated/dummy.pro -ts translations.ts || die
+
+git branch -r
+for branch in $(git branch -r | grep origin/releases); do
+  printn Y "Importing strings from $branch..."
```
flodolo, line 48: Is it just me or I don't see this part of the code running in the action?
https://github.com/mozilla-mobile/mozilla-vpn-client/runs/6738300986?check_suite_focus=true

Also don't see the branch name printed from line 17.

```
022-06-04T12:46:25.9413195Z Branch name: 
2022-06-04T12:46:25.9449174Z Caching the dep scripts... done.
2022-06-04T12:46:26.0604284Z Generating strings... done.
2022-06-04T12:46:26.0616738Z Generating a dummy PRO file... done
2022-06-04T12:46:26.0617327Z Generating the main translation file... 
2022-06-04T12:46:26.1787790Z Info: creating stash file /home/runner/work/mozilla-vpn-client/mozilla-vpn-client/.qmake.stash
2022-06-04T12:46:26.1839241Z lupdate warning: TS files from command line will override TRANSLATIONS in /home/runner/work/mozilla-vpn-client/mozilla-vpn-client/translations/generated/dummy.pro.
2022-06-04T12:46:26.2991234Z Updating 'translations.ts'...
2022-06-04T12:46:26.2997391Z     Found 493 source text(s) (493 new and 0 already existing)
2022-06-04T12:46:26.3022964Z   pull/3664/merge
2022-06-04T12:46:26.3062422Z Remove cache... done.
```

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: Translations
+    runs-on: ubuntu-latest
+    env:
+      QTVERSION: 6.2.4
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Checkout submodules
```
flodolo, line 29: BTW, do you really need to update submodules? I assume there are no strings in there, but I might be wrong. We definitely don't do it in the l10n action.

scripts/utils/generate_ts.sh
```diff
@@ -0,0 +1,72 @@
...
+HEADERS += \$\$files(../../nebula/*.h, true)
+SOURCES += \$\$files(../../nebula/*.cpp, true)
+RESOURCES += \$\$files(../../nebula/*.qrc, true)
+EOF
+print G "done"
+
+print Y "Generating the main translation file... "
+lupdate translations/generated/dummy.pro -ts translations.ts || die
+
+git branch -r
```
flodolo, line 46: This looks like debug stuff?

flodolo (changes requested):
Found the issue (cloning with depth 1, so no branches). See fixes here
https://github.com/flodolo/mozilla-vpn-client/commit/e7f1af31e2737aa954fef607341e912410759ea6

flodolo (approved):
Seems to work now 👍🏼

flodolo:
@bakulf 
Is it OK to merge this, so I can check on the l10n side?

PR #3671, opened 2022-06-03
Fallback for the download task DNS lookup

flozia:
## Description

In case the DNS lookup fails in the for the download task we need a fallback.

## Reference

[VPN-2157](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-2157)

## Checklist
    

bakulf:
This looks good but it's important to see why it used to work with previous builds. Maybe we are just hiding a bigger bug.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -55,43 +55,60 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
+  connect(request, &NetworkRequest::requestUpdated, this,
+          &BenchmarkTaskDownload::downloadProgressed);
+  connect(request, &NetworkRequest::requestFailed, this,
+          &BenchmarkTaskDownload::downloadReady);
+  connect(request, &NetworkRequest::requestCompleted, this,
+          [&](const QByteArray& data) {
+            downloadReady(QNetworkReply::NoError, data);
+          });
+
+  logger.debug() << "Starting request:" << (void*)request;
```
oskirby, line 70: Since we're making changes to this file, could you remove the logging of the `(void*)request` pointer?

oskirby (approved):

flozia:
> This looks good but it's important to see why it used to work with previous builds. Maybe we are just hiding a bigger bug.

This bug was only discovered while testing `2.9`, because the the DNS lookup was not included in the `2.8` release.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -55,43 +55,60 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
   }
 }
 
+void BenchmarkTaskDownload::connectNetworkRequest(NetworkRequest* request) {
```
bakulf, line 58: If we know that QDnsLoop does not work on android, can we simply add a:
```
#ifdef MVPN_ANDROID
  NetworkRequest* request =
      NetworkRequest::createForGetUrl(this, m_fileUrl.toString());
  connectNetworkRequest(request);
#endif
....

**Review by bakulf — CHANGES_REQUESTED** [review]
And I would like to add something like:

```
if QT_VERSION >= 0x060400
#error Check if QT has finally implemented QDnsLoop
#endif
```

File: src/connectionbenchmark/benchmarktaskdownload.cpp [hunk 3]
```diff
@@ -55,43 +55,60 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
     emit finished(0, true);
     emit completed();
   });
 
-  if (m_dnsLookup.error() != QDnsLookup::NoError) {
+  if (state() != BenchmarkTask::StateActive) {
+    logger.warning() << "DNS Lookup finished after task aborted";
+    return;
+  }
+  if (m_dnsLookup.error() != QDnsLookup::NoError || true) {
```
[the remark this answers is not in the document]
flozia: :)

src/connectionbenchmark/benchmarktaskdownload.cpp
flozia: Removed the fallback.

src/connectionbenchmark/benchmarktaskdownload.cpp
```diff
@@ -55,43 +67,49 @@ void BenchmarkTaskDownload::handleState(BenchmarkTask::State state) {
...
+  m_requests.append(request);
+}
+
 void BenchmarkTaskDownload::dnsLookupFinished() {
   auto guard = qScopeGuard([&] {
     emit finished(0, true);
     emit completed();
   });
 
+  if (state() != BenchmarkTask::StateActive) {
```
bakulf, line 92: We dont need this any more. remove the changes in this method.

bakulf (approved):
Nice!  Apply the changes before merging it. Thanks!

PR #3675, opened 2022-06-03
VPN-2259 - Implement exponentially increasing interval between WebSocket server connection retries

brizental:
## Description

The title is descriptive enough for the main changes here.

I also did some clean up and added an extra test for the `close` method.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2259

## Checklist
    
### DIFF
--- a/src/websockethandler.cpp
+++ b/src/websockethandler.cpp
@@ -15,6 +15,60 @@ namespace {
 Logger logger(LOG_MAIN, "WebSocketHandler");
 }  // namespace
 
+ExponentialBackoffStrategy::ExponentialBackoffStrategy() {
+  MVPN_COUNT_CTOR(ExponentialBackoffStrategy);
+
+  m_retryTimer.setSingleShot(true);
+}
+
+#ifdef UNIT_TEST
+void ExponentialBackoffStrategy::testOverrideBaseRetryInterval(
+    int newInterval) {
+  m_baseInterval = newInterval;
+}
+
+void ExponentialBackoffStrategy::testOverrideMaxRetryInterval(int newInterval) {
+  m_maxInterval = newInterval;
+}
+#endif
+
+/**
+ * @brief Schedules the next attempt to execute a given function.
+ *
+ * Everytime a new attempt is scheduled the interval will be exponentially
+ * larger.
+ */
+int ExponentialBackoffStrategy::scheduleNextAttempt(
+    std::function<void()> action) {
+  int retryInterval = qPow(m_baseInterval, m_retryCounter);
+  if (retryInterval < m_maxInterval) {
+    m_retryCounter++;
+  }
+
+  QMetaObject::Connection* const connection = new QMetaObject::Connection;
+  *connection =
+      connect(&m_retryTimer, &QTimer::timeout, this, [action, connection]() {
+        action();
+
+        // Immediatelly disconnect after first execution.
+        QObject::disconnect(*connection);
+        delete connection;
+      });
+
+  m_retryTimer.start(retryInterval);
+
+  return retryInterval;
+}
+
+/**
+ * @brief Stops the timer for an ongoing execution attempt, if any. Resets the
+ * interval to base interval.
+ */
+void ExponentialBackoffStrategy::reset() {
+  m_retryTimer.stop();
+  m_retryCounter = 1;
+}
+
 WebSocketHandler::WebSocketHandler() {
   MVPN_COUNT_CTOR(WebSocketHandler);
 
@@ -70,8 +124,8 @@ void WebSocketHandler::testOverridePingInterval(int newInterval) {
   m_pingInterval = newInterval;
 }
 
-void WebSocketHandler::testOverrideRetryInterval(int newInterval) {
-  m_retryInterval = newInterval;
+void WebSocketHandler::testOverrideBaseRetryInterval(int newInterval) {
+  m_backoffStrategy.testOverrideBaseRetryInterval(newInterval);
 }
 #endif
 
@@ -134,6 +188,8 @@ void WebSocketHandler::open() {
 void WebSocketHandler::onConnected() {
   logger.debug() << "WebSocket connected";
 
+  m_backoffStrategy.reset();
+
   connect(&m_webSocket, &QWebSocket::textMessageReceived, this,
           &WebSocketHandler::onMessageReceived);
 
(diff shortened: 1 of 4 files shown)


src/websockethandler.cpp
```diff
@@ -15,6 +15,60 @@ namespace {
...
+#endif
+
+/**
+ * @brief Schedules the next attempt to execute a given function.
+ *
+ * Everytime a new attempt is scheduled the interval will be exponentially
+ * larger.
+ */
+int ExponentialBackoffStrategy::scheduleNextAttempt(
+    std::function<void()> action) {
```
bakulf, line 42: `std::function<void>&& action`

src/websockethandler.cpp
```diff
@@ -15,6 +15,60 @@ namespace {
...
+  int retryInterval = qPow(m_baseInterval, m_retryCounter);
+  if (retryInterval < m_maxInterval) {
+    m_retryCounter++;
+  }
+
```
bakulf, line 47: Can you store the action in a `std::function<void()> m_nextAction;`?

src/websockethandler.cpp
```diff
@@ -15,6 +15,60 @@ namespace {
...
+  QMetaObject::Connection* const connection = new QMetaObject::Connection;
+  *connection =
+      connect(&m_retryTimer, &QTimer::timeout, this, [action, connection]() {
```
bakulf, line 50: I wonder if this object should simply emit a signal and we can connect the signal to `open`. Would it be the same?
If we follow this QT pattern, we can connect the signal `GoodNameHere` of `m_backoffStrategy` in the CTOR of the `WebSocketHandler`. Then we can call: `schedule()` that will emit the signal, eventually.

src/websockethandler.cpp
```diff
@@ -15,6 +15,60 @@ namespace {
...
+
+  return retryInterval;
+}
+
+/**
+ * @brief Stops the timer for an ongoing execution attempt, if any. Resets the
+ * interval to base interval.
+ */
+void ExponentialBackoffStrategy::reset() {
+  m_retryTimer.stop();
```
bakulf, line 68: This will leak the `connection` if there is a pending operation.

src/websockethandler.cpp
```diff
@@ -15,6 +15,60 @@ namespace {
...
+  }
+
+  QMetaObject::Connection* const connection = new QMetaObject::Connection;
+  *connection =
+      connect(&m_retryTimer, &QTimer::timeout, this, [action, connection]() {
+        action();
+
+        // Immediatelly disconnect after first execution.
+        QObject::disconnect(*connection);
+        delete connection;
```
bakulf, line 55: this connection will be leaked if `scheduleNextAttempt` is called multiple times by mistake.

tests/unit/testwebsockethandler.cpp
```diff
@@ -173,7 +347,7 @@ void TestWebSocketHandler::tst_reconnectionAttemptsOnPingTimeout() {
   QSignalSpy socketDisconnectedSpy(&server, SIGNAL(socketDisconnected()));
 
   WebSocketHandler handler;
-  handler.testOverrideRetryInterval(100);
+  handler.testOverrideBaseRetryInterval(100);
```
bakulf, line 350: You can move the `ExponentialBackoffStrategy` object in a separate class and test it independently.

bakulf (changes requested):

src/websockethandler.cpp
brizental: Heh, that is what I get for copying code from stack overflow... This will be gone on my next commit thanks to your previous comment.

src/websockethandler.cpp
```diff
@@ -15,6 +15,57 @@ namespace {
 Logger logger(LOG_MAIN, "WebSocketHandler");
 }  // namespace
 
+ExponentialBackoffStrategy::ExponentialBackoffStrategy() {
+  MVPN_COUNT_CTOR(ExponentialBackoffStrategy);
+
+  connect(&m_retryTimer, &QTimer::timeout, this,
+          &ExponentialBackoffStrategy::onNextAttemptTimeout);
```
bakulf, line 22: connect the timeout signal with the exeucteNextAtempt directly: `&ExponentialBackoffStrategy::executeNextAttempt)`. In this way you can remove `onNextattemptTimeout`

src/websockethandler.h
```diff
@@ -8,8 +8,36 @@
...
+#ifdef UNIT_TEST
+  void testOverrideBaseRetryInterval(int newInterval);
+  void testOverrideMaxRetryInterval(int newInterval);
+#endif
+
+ signals:
+  void executeNextAttempt();
+
+ private:
+  void onNextAttemptTimeout();
```
bakulf, line 31: you can remove this. See my comment below/above :)

tests/unit/testwebsockethandler.cpp
bakulf: All good! you are testing the exponentialBackoffstrategy already. I was suggesting to move it in a seprate class "file" (I forgot to write this word:) 
Ignore this comment.

src/websockethandler.cpp
```diff
@@ -15,6 +15,49 @@ namespace {
...
+
+/**
+ * @brief Schedules the next attempt to execute a given function.
+ *
+ * Everytime a new attempt is scheduled the interval will be exponentially
+ * larger.
+ */
+int ExponentialBackoffStrategy::scheduleNextAttempt() {
+  int retryInterval = qPow(m_baseInterval, m_retryCounter);
+  if (retryInterval < m_maxInterval) {
```
bakulf, line 45: I have a question here. `m_baseinterval` is 5000. `m_maxInterval` is 3600000. So, the interval will be only 5000.
Because 5000^2 is 25000000 which is greater than 3600000.
I wonder if we should multiply `* 1000` later. In this way the intervals will be: [ 5, 25, 125, 625, 3125 ].
Am I right?

brizental: !!! Yeap, you are totally right. I missed this in tests because I am overwriting the interval over there to 5...

src/websockethandler.h
```diff
@@ -8,8 +8,31 @@
 #include <QWebSocket>
 #include <QTimer>
 
-constexpr const int WEBSOCKET_PING_INTERVAL = 30 * 1000;   // 30s
-constexpr const int WEBSOCKET_RETRY_INTERVAL = 30 * 1000;  // 30s
+constexpr const int BASE_RETRY_INTERVAL = 5;              // 5s
+constexpr const int MAX_RETRY_INTERVAL = 60 * 60 * 1000;  // 1hr
```
bakulf, line 12: Can you maybe rename these two to `_SEC` and `_MSEC` ?

bakulf (approved):

PR #3707, opened 2022-06-07
Update Get help menus

lesleyjanenorton:
## Description

This changes the 'Contact us' string to 'Contact support' in the macOS top bar menu and moves 'Give feedback' to the bottom of the list in ViewGetHelp.qml. 

## Reference

   VPN-2198

## Checklist
    
### DIFF
--- a/src/models/helpmodel.cpp
+++ b/src/models/helpmodel.cpp
@@ -49,7 +49,7 @@ void maybeInitialize() {
   s_helpEntries.append(
       HelpEntry(L18nStrings::instance()->id(
                     L18nStrings::InAppSupportWorkflowSupportNavLinkText),
-                true, false, MozillaVPN::LinkContact));
+                false, false, MozillaVPN::LinkContact));
 
   //% "View log"
   logger.debug() << "Adding:" << qtTrId("help.viewLog");
--- a/src/models/helpmodel.cpp
+++ b/src/models/helpmodel.cpp
@@ -43,10 +43,10 @@ void maybeInitialize() {
   s_helpEntries.append(
       HelpEntry("help.helpCenter2", true, false, MozillaVPN::LinkHelpSupport));
 
-  //% "Contact us"
-  logger.debug() << "Adding:" << qtTrId("help.contactUs");
-  s_helpEntries.append(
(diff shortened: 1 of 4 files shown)


src/models/helpmodel.cpp
```diff
@@ -43,10 +43,10 @@ void maybeInitialize() {
   s_helpEntries.append(
       HelpEntry("help.helpCenter2", true, false, MozillaVPN::LinkHelpSupport));
 
-  //% "Contact us"
-  logger.debug() << "Adding:" << qtTrId("help.contactUs");
-  s_helpEntries.append(
-      HelpEntry("help.contactUs", false, false, MozillaVPN::LinkContact));
+  logger.debug() << "Adding:"
+                 << qtTrId("vpn.inAppSupportWorkflow.supportNavLinkText");
+  s_helpEntries.append(HelpEntry("vpn.inAppSupportWorkflow.supportNavLinkText",
```
bakulf, line 48: I don't like this because it exposes the string ID generated from the `strings.yaml`. I wonder if we can do something different:
- we can add a `const char* id(String)` method in ` l18nstrings.h`.
- here we can use it in this way: `s_helpEntries.append(HelpEntry(L18nStrings::instance()->id(L18nStrings::InAppSupportWorkflowSupportNavLinkText)))`

It's a bit more work because you need to change `scripts/utils/generate_strings.py`: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/scripts/utils/generate_strings.py#L327 adding `const char* id(String) const;`

and also: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/translations/l18nstrings.cpp#L39
Adding something like:

```
const char* L18nString::id(L18nString::String string) const {
  Q_ASSERT(string < __Last);
   return _ids[string];
}
```

Actually! If you add this method, you can also simplify the `t()` one:

```
                                                     
QString L18nStrings::t(L18nStrings::String string) const {                                                       
  QString id = id(string);                                                                                     
  if (id.isEmpty()) return "";                                                                                   
  return qtTrId(id);                                                                                   
}                                                              
```

bakulf (approved):

bakulf:
Probably you need to update a functional test

Issue #2643, opened 2022-01-26
Telemetry: Add VPN usage metrics


data-sync-user:
As a product manager and organization, I want to monitor actual VPN usage so that we can build metrics around delivering value to our customers.

ToDo: Add glean telemetry for vpn connection on and off.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1619)

data-sync-user:
➤ Sarah Bird commented:

Additionally, an uptime metric will help with https://docs.google.com/document/d/1RuFijLAwRNmWKwH6XhqWC19iBzVfz0VFlkJwWwnbyeo/edit ( https://docs.google.com/document/d/1RuFijLAwRNmWKwH6XhqWC19iBzVfz0VFlkJwWwnbyeo/edit|smart-link )

PR #3717, opened 2022-06-09
Add periodic check (3 hours / 1 minute) for controller state

birdsarah:
## Description

Adds two new events for particular controller states (to ease analysis as opposed to using the controller_step data) and sets a long timer (3 hours / 10 minutes) to report the controller state for desktop. Follow-up work will be done to incorporate the controllerStateOn event for mobile in separate PRs.

## Reference

* [Jira](https://mozilla-hub.atlassian.net/browse/VPN-1619)
* [Github](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2643)

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -80,6 +80,13 @@ Controller::Controller() {
 
   connect(&m_handshakeTimer, &QTimer::timeout, this,
           &Controller::handshakeTimeout);
+
+  // Setup regular check for controller up state
+  connect(&m_gleanControllerUpTimer, &QTimer::timeout, this,
+          &Controller::periodicStateRecorder);
+  m_gleanControllerUpTimer.start(
+      Constants::controllerPeriodicStateRecorderMsec());
+  m_gleanControllerUpTimer.setSingleShot(false);
 }
 
 Controller::~Controller() { MVPN_COUNT_DTOR(Controller); }
@@ -609,8 +616,34 @@ void Controller::setState(State state) {
   emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
       GleanSample::controllerStep,
       {{"state", QVariant::fromValue(state).toString()}});
+
+  // Specific events for on and off state to aid with analysis
+  if (m_state == StateOn) {
+    emit MozillaVPN::instance()->recordGleanEvent(
+        GleanSample::controllerStateOn);
+  }
+  if (m_state == StateOff) {
+    emit MozillaVPN::instance()->recordGleanEvent(
+        GleanSample::controllerStateOff);
+  }
+
   m_state = state;
   emit stateChanged();
+}
+
+void Controller::periodicStateRecorder() {
+// On mobile this is handled in the background process that talks to the
+// controller
+#if defined(MVPN_WINDOWS) || defined(MVPN_LINUX) || defined(MVPN_MACOS)
+  if (m_state == StateOn) {
+    emit MozillaVPN::instance()->recordGleanEvent(
+        GleanSample::controllerStateOn);
+  }
+  if (m_state == StateOff) {
+    emit MozillaVPN::instance()->recordGleanEvent(
+        GleanSample::controllerStateOff);
+  }
+#endif
 }
 
 qint64 Controller::time() const {
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -606,20 +606,6 @@ void Controller::setState(State state) {
     return;
   }
   logger.debug() << "Setting state:" << state;
-  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
-      GleanSample::controllerStep,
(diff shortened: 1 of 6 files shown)


birdsarah:
Bug is draft while waiting for data review. But is ready for review @strseb.

src/controller.cpp
```diff
@@ -80,6 +80,13 @@ Controller::Controller() {
 
   connect(&m_handshakeTimer, &QTimer::timeout, this,
           &Controller::handshakeTimeout);
+
+  // Setup regular check for controller up state
+  connect(&m_gleanControllerUpTimer, &QTimer::timeout, this,
+          &Controller::periodicStateRecorder);
+  m_gleanControllerUpTimer.start(
```
bakulf, line 87: should we reset the time each time we go to `StateOn`/`StateOff` ? In order to have fix step data. Otherwise it could be this timer is triggered immediately after a connection/disconnection.

birdsarah:
Moved out of draft as have data review +.

src/controller.cpp
birdsarah: I thought about this but didn't want to over complicate it. I think this will be fine actually for now and keeps the code much simpler. If it turns out analysis is complicated we can improve. But this was my motivation for putting this into the state change too so that we can use this instead of needing to also read from a state change event. So we can consider these events for usage metrics and state change events for troubleshooting our connection dramas.

src/controller.cpp
```diff
@@ -609,10 +616,36 @@ void Controller::setState(State state) {
...
+  }
+
   m_state = state;
   emit stateChanged();
 }
 
+void Controller::periodicStateRecorder() {
+// On mobile this is handled in the background process that talks to the
+// controller
+#if defined(MVPN_WINDOWS) || defined(MVPN_LINUX) || defined(MVPN_MACOS)
```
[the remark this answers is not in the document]
birdsarah: Only windows and mac use LocalSocketController unless I'm missing something.

src/controller.h
```diff
@@ -176,6 +176,7 @@ class Controller final : public QObject {
 
   QTimer m_connectingTimer;
   QTimer m_handshakeTimer;
+  QTimer m_gleanControllerUpTimer;
```
bakulf, line 179: I have a better solution. Tell me what you think about it:
Can you move this timer in `telemetry.h/cpp`? https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/telemetry.h

In the end, this timer can live everywhere and I've already started a cleanup of controller.h/cpp.
If you do that, then activating and deactivating the timer when we are in StateOn/Off can be easily done.

bakulf (changes requested):

birdsarah:
@bakulf I'm not sure if this is what you intended with the timer change, but I've tested it locally and it's working as I expect.

src/controller.cpp
```diff
@@ -609,6 +609,17 @@ void Controller::setState(State state) {
   emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
       GleanSample::controllerStep,
       {{"state", QVariant::fromValue(state).toString()}});
+
+  // Specific events for on and off state to aid with analysis
+  if (state == StateOn) {
```
bakulf, line 614: move these to telemetry.cpp. See https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/telemetry.cpp#L43

src/telemetry.cpp
```diff
@@ -21,6 +21,12 @@ Telemetry::Telemetry() {
   m_connectionStabilityTimer.setSingleShot(true);
   connect(&m_connectionStabilityTimer, &QTimer::timeout, this,
           &Telemetry::connectionStabilityEvent);
+
+  connect(&m_gleanControllerUpTimer, &QTimer::timeout, this,
+          &Telemetry::periodicStateRecorder);
+  m_gleanControllerUpTimer.start(
```
bakulf, line 27: I would do in `StateChanged`:

```
if (state == Controller::StateOn || state == Controller::SateOff) {
   m_gleanControllerUpTimer.start(Constants::controllerPeriodicStateRecorderMsec());
} else {
   m_gleanControllerUpTimer.stop();
}
```

bakulf (changes requested):

src/telemetry.cpp
birdsarah: Agreed in conversation to not do this.

birdsarah:
Hopefully this is good. I also shortened the stage timer to 1minute as it was really tedious waiting for 10 minutes.

birdsarah:
I also moved the other telemetry from controller.cpp while I was at it as it was all the same stuff.

strseb (approved):

bakulf (approved):

PR #3744, opened 2022-06-13
fix: misnamed path in Windows qt toolchain script

ahal:
This fixes a bug in the Windows qt toolchain task which was resulting in an invalid artifact (and therefore also invalid Windows builds. This was regressed by #3620 ,

### DIFF
--- a/taskcluster/scripts/toolchain/compile_qt_6.ps1
+++ b/taskcluster/scripts/toolchain/compile_qt_6.ps1
@@ -84,7 +84,7 @@ Set-Location $FETCHES_PATH/qt-everywhere-src-$QT_VERSION
 
 
 Set-Location $REPO_ROOT_PATH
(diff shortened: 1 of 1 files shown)


ahal:
Hmm, so the [first run](https://firefox-ci-tc.services.mozilla.com/tasks/ZMlD4qLPQX-kQXdCOrMETQ/runs/0) of the `qt` task failed with exit status `1073807364`. I'm not sure what that means, but it's listed in the task payload as a `retry` exit code, so the task was re-tried.

But then the [second](https://firefox-ci-tc.services.mozilla.com/tasks/ZMlD4qLPQX-kQXdCOrMETQ/runs/1) run hit an error in `Invoke-Webrequest` much earlier than the place this PR is changing.. so I *think* it's either another intermittent or is somehow caused due to the failure in the first run?

PR #3761, opened 2022-06-15
Add level-3 latest-builds to a version-keyed index

strseb:
## Description
This adds the app_version to the taskgraph params. It's determined by reading the cmake lists. 
This is then used to add an index route to the sign tasks. Also i've added this for now to the build/windows task as we don't have signing for now :) 

Here is an example task generated by `taskgraph full --p taskcluster/test/params/release-push.yml`
Signing/Linux therefore gets: "index.mozillavpn.v2.mozillavpn.releases.2.9.0.linux.latest",

```
  "signing-linux/opt": {
    "attributes": {
      "always_target": false,
      "build-type": "linux/opt",
      "kind": "signing",
      "release-artifacts": [
        {
          "name": "public/build/target.tar.gz",
          "path": "/builds/worker/artifacts/target.tar.gz",
          "type": "file"
        }
      ],
      "run_on_projects": [
        "all"
      ],
      "run_on_tasks_for": [
        "github-push"
      ],
      "signed": true
    },
    "dependencies": {
      "build": "build-linux/opt"
    },
    "description": "Sign MozillaVPN",
    "if_dependencies": [],
    "kind": "signing",
    "label": "signing-linux/opt",
    "optimization": null,
    "soft_dependencies": [],
    "task": {
      "created": {
        "relative-datestamp": "0 seconds"
      },
      "deadline": {
        "relative-datestamp": "1 day"
      },
      "expires": {
        "relative-datestamp": "1 year"
      },
      "extra": {
        "index": {
          "rank": 1641494741
        },
        "parent": "",
        "treeherder": {
          "collection": {
            "all": true
          },
          "jobKind": "build",
          "machine": {
            "platform": "linux"
          },
          "symbol": "Bs",
          "tier": 1
        },
        "treeherder-platform": "linux/all"
      },
      "metadata": {
        "description": "Sign MozillaVPN ([Treeherder push](https://treeherder.mozilla.org/#/jobs?repo=mozilla-vpn-client&revision=5d9e12a6b60a28218aa8094a45111777d73f7c1d))",
        "name": "signing-linux/opt",
        "owner": "user@example.com",
        "source": "https://github.com/mozilla-mobile/mozilla-vpn-client/blob/5d9e12a6b60a28218aa8094a45111777d73f7c1d/taskcluster/ci/signing"
      },
      "payload": {
        "maxRunTime": 600,
        "upstreamArtifacts": [
          {
            "formats": [
              "autograph_debsign"
            ],
            "paths": [
              "public/build/target.tar.gz"
            ],
            "taskId": {
              "task-reference": "<build>"
            },
            "taskType": "build"
          }
        ]
      },
      "priority": "highest",
      "provisionerId": "scriptworker-k8s",
      "routes": [
        "index.mozillavpn.v2.mozillavpn.releases.2.9.0.linux.latest",
        "tc-treeherder.v2.mozilla-vpn-client.5d9e12a6b60a28218aa8094a45111777d73f7c1d.0",
        "checks"
      ],
      "scopes": [
        "project:mozillavpn:releng:signing:cert:release-signing",
        "project:mozillavpn:releng:signing:format:autograph_debsign"
      ],
      "tags": {
        "createdForUser": "user@example.com",
        "kind": "signing",
        "label": "signing-linux/opt",
        "os": "scriptworker",
        "worker-implementation": "scriptworker"
      },
      "workerType": "mozillavpn-3-signing"
    }
  },


```

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -5,6 +5,7 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
+    - mozillavpn_taskgraph.transforms.release_index:transforms
     - mozillavpn_taskgraph.transforms.build:transforms
     - taskgraph.transforms.job:transforms
     - taskgraph.transforms.task:transforms
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -29,7 +29,7 @@ job-template:
     description: Sign MozillaVPN
     worker: {} # Will be configured by signing:transforms
     run-on-tasks-for: []
-    push-to-release-index: true
+    add-release-index: true
     signing-format:
         by-build-type:
             android.*: autograph_apk
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -7,6 +7,7 @@ loader: mozillavpn_taskgraph.loader.multi_dep:loader
 transforms:
     - mozillavpn_taskgraph.transforms.multi_dep:transforms
     - mozillavpn_taskgraph.transforms.signing:transforms
+    - mozillavpn_taskgraph.transforms.release_index:transforms
     - taskgraph.transforms.task:transforms
 
 kind-dependencies:
@@ -28,6 +29,7 @@ job-template:
     description: Sign MozillaVPN
     worker: {} # Will be configured by signing:transforms
     run-on-tasks-for: []
+    push-to-release-index: true
     signing-format:
         by-build-type:
             android.*: autograph_apk
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -29,7 +29,7 @@ job-template:
     description: Sign MozillaVPN
     worker: {} # Will be configured by signing:transforms
     run-on-tasks-for: []
-    add-release-index: true
+    add-branch-index: true
     signing-format:
         by-build-type:
             android.*: autograph_apk
--- a/taskcluster/mozillavpn_taskgraph/parameters.py
+++ b/taskcluster/mozillavpn_taskgraph/parameters.py
@@ -16,14 +16,13 @@ def get_defaults(repo_root):
     }
 
 def getAppVersion():
-    f = open("CMakeLists.txt","r")
-    lines = f.readlines()
-    f.close()
-    isVersionLine = lambda x: "project(\"Mozilla VPN\" VERSION" in x
-    version_line = next(filter(isVersionLine,lines))
-    tokens = version_line.split(" ")
-    version = tokens[tokens.index("VERSION")+1]
-    return version
+    with open("CMakeLists.txt") as f:
+        lines = f.readlines()
+        isVersionLine = lambda x: "project(\"Mozilla VPN\" VERSION" in x
+        version_line = next(filter(isVersionLine,lines))
+        tokens = version_line.split(" ")
+        version = tokens[tokens.index("VERSION")+1]
+        return version
 
 
 extend_parameters_schema(
--- a/taskcluster/mozillavpn_taskgraph/parameters.py
+++ b/taskcluster/mozillavpn_taskgraph/parameters.py
@@ -12,25 +12,13 @@ def get_defaults(repo_root):
     return {
         "pull_request_number": None,
         "version": "",
-        "app_version":"0.0.0"
     }
-
-def getAppVersion():
-    with open("CMakeLists.txt") as f:
-        lines = f.readlines()
-        isVersionLine = lambda x: "project(\"Mozilla VPN\" VERSION" in x
-        version_line = next(filter(isVersionLine,lines))
-        tokens = version_line.split(" ")
-        version = tokens[tokens.index("VERSION")+1]
-        return version
 
 
 extend_parameters_schema(
     {
         Required("pull_request_number"): Any(All(int, Range(min=1)), None),
         Required("version"): str,
-        Required("app_version"): str,
-        
     },
     defaults_fn=get_defaults
 )
@@ -39,7 +27,6 @@ extend_parameters_schema(
 def get_decision_parameters(graph_config, parameters):
     head_tag = parameters["head_tag"]
     parameters["version"] = head_tag[1:] if head_tag else ""
-    parameters["app_version"] = getAppVersion()
 
     pr_number = os.environ.get("MOZILLAVPN_PULL_REQUEST_NUMBER", None)
     parameters["pull_request_number"] = None if pr_number is None else int(pr_number)
--- a/taskcluster/mozillavpn_taskgraph/parameters.py
+++ b/taskcluster/mozillavpn_taskgraph/parameters.py
@@ -12,13 +12,26 @@ def get_defaults(repo_root):
     return {
         "pull_request_number": None,
         "version": "",
+        "app_version":"0.0.0"
     }
+
+def getAppVersion():
+    f = open("CMakeLists.txt","r")
+    lines = f.readlines()
+    f.close()
+    isVersionLine = lambda x: "project(\"Mozilla VPN\" VERSION" in x
+    version_line = next(filter(isVersionLine,lines))
+    tokens = version_line.split(" ")
+    version = tokens[tokens.index("VERSION")+1]
+    return version
 
 
 extend_parameters_schema(
     {
         Required("pull_request_number"): Any(All(int, Range(min=1)), None),
         Required("version"): str,
+        Required("app_version"): str,
+        
     },
     defaults_fn=get_defaults
 )
@@ -27,6 +40,7 @@ extend_parameters_schema(
 def get_decision_parameters(graph_config, parameters):
     head_tag = parameters["head_tag"]
     parameters["version"] = head_tag[1:] if head_tag else ""
+    parameters["app_version"] = getAppVersion()
 
     pr_number = os.environ.get("MOZILLAVPN_PULL_REQUEST_NUMBER", None)
     parameters["pull_request_number"] = None if pr_number is None else int(pr_number)
--- a/taskcluster/mozillavpn_taskgraph/transforms/release_index.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/release_index.py
@@ -6,15 +6,18 @@
 from taskgraph.transforms.base import TransformSequence
 transforms = TransformSequence()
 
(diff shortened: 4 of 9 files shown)


taskcluster/mozillavpn_taskgraph/parameters.py
```diff
@@ -12,13 +12,26 @@ def get_defaults(repo_root):
     return {
         "pull_request_number": None,
         "version": "",
+        "app_version":"0.0.0"
     }
 
+def getAppVersion():
+    f = open("CMakeLists.txt","r")
+    lines = f.readlines()
+    f.close()
```
ahal, line 21: A bit simpler:
```
with open("CMakeLists.txt") as f:
    lines = f.readlines()
```

taskcluster/ci/build/kind.yml
```diff
@@ -5,6 +5,7 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
+    - mozillavpn_taskgraph.transforms.release_index:transforms
```
ahal, line 8: Open question, do we want to add both the unsigned and signed build tasks to this index? I was kind of thinking we'd only index the signed tasks, but maybe there's benefit to also adding the build tasks.

taskcluster/ci/signing/kind.yml
```diff
@@ -7,6 +7,7 @@ loader: mozillavpn_taskgraph.loader.multi_dep:loader
 transforms:
     - mozillavpn_taskgraph.transforms.multi_dep:transforms
     - mozillavpn_taskgraph.transforms.signing:transforms
+    - mozillavpn_taskgraph.transforms.release_index:transforms
```
ahal, line 10: I'm kind of on the fence about this comment, so feel free to leave this as is..

But the `task` transforms support a (undocumented) feature called `index_builders`, e.g:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/transforms/task.py#L890

So basically you could create a "release_index" builder. You'd want to define it somewhere that gets imported from [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/taskcluster/mozillavpn_taskgraph/__init__.py#L21) (you'd likely want to add a new module called "index_builders" or something like that).

Then rather than have your tasks import the transform, you'd add something like:
```
index:
    type: release
    product: mozillavpn
    job-name: build
```
to its definition.

That said, the transform approach works too and is probably easier to trace, so I don't mind going with that approach either.

taskcluster/ci/build/kind.yml
ahal: Oh, I think you're doing this just because there are no signed Windows builds yet. If this is temporary, then that makes sense to me!

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,23 @@
...
+from taskgraph.transforms.base import TransformSequence
+transforms = TransformSequence()
+
+
+@transforms.add
+def add_pr_number(config, tasks):
+   for task in tasks:
+        if task.__contains__("push-to-release-index"):
+            release_index = task.pop("push-to-release-index")
+            if release_index and int(config.params["level"]) == 3:
```
ahal, line 15: I think we'll want to check `config.params["tasks_for"] == "github-release"` here. Otherwise, any old push to a branch will overwrite this index.

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,23 @@
...
+                name = task['name'].split("/")[0]
+                version = config.params["app_version"]
+                route = f"index.mozillavpn.v2.mozillavpn.releases.{version}.{name}.latest"
```
ahal, line 18: I guess it's a matter of taste, but I think the `latest` here is superfluous.

We may want to add a `index.mozillavpn.v2.mozillavpn.releases.latest.{name}` route, but then we'd need some way to check whether the current version is the highest or not. So maybe we can ignore that for now.

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,23 @@
...
+                if task.__contains__("routes"):
+                    task["routes"].append(route)
+                else:
+                    task["routes"] = [route]
```
ahal, line 22: Instead of this if/else, try:
```
task.setdefault("routes", []).append(route)
```

taskcluster/ci/signing/kind.yml
```diff
@@ -28,6 +29,7 @@ job-template:
     description: Sign MozillaVPN
     worker: {} # Will be configured by signing:transforms
     run-on-tasks-for: []
+    push-to-release-index: true
```
ahal, line 32: nit: I think `push` is probably the wrong verb here as indexes are just pointers to a task. Maybe `add-release-index`?

Note that if you go with the `index_builders` suggestion above, this key won't be necessary.

taskcluster/ci/build/kind.yml
ahal: Also if you move this transform below the `build:transforms`, you wouldn't need to add `routes` to the build schema. That said, I propose a better fix below.

ahal (changes requested):
Nice, thanks! Despite the # of comments, this is good and looks like it would work as is.

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
strseb: I think that is what we want rn. Even as we are in "feature freeze" we will update the release branch probably multiple times per day, and we want to have the index to always point at the latest build here. 
Doing a gh-pre-release for that would not really scale here and make qa depend on us pushing that often :D 
Also once we are in codefreeze main will host 2.10, so i'm not worried about main or other branches overwriting this.

strseb: Thinking about... maybe in that case the index should not be `mozillavpn.releases.{version}.{name}` rather `mozillavpn.preview.{version}.{name}` ? And keep the mozillavpn.releases for "github-release" graphs?

taskcluster/ci/build/kind.yml
strseb: Yes, i've only added this to have the windows.msi also there, for now :)

taskcluster/ci/signing/kind.yml
strseb: That is super cool buuut i'm afraid, qa would like the index route ready tommorow (eu) morning, so i would probably look into this in a followup, filed a bug to look into that later :)

taskcluster/mozillavpn_taskgraph/parameters.py
```diff
@@ -12,13 +12,26 @@ def get_defaults(repo_root):
...
+    tokens = version_line.split(" ")
+    version = tokens[tokens.index("VERSION")+1]
+    return version
+
 
 extend_parameters_schema(
     {
         Required("pull_request_number"): Any(All(int, Range(min=1)), None),
         Required("version"): str,
+        Required("app_version"): str,
```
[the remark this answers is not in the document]
ahal: Ah ok, so these indexes aren't for releases then, but rather builds. We may want to rethink the naming a bit in that case.

Instead of introducing a new `app_version` parameter, could we change `version` to read from the `CMakeLists` file instead?

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
ahal: Yeah, I think that's what I was missing.

So in this case I'd recommend using the `branch` namespace, e.g:
```
index.mozillvpn.v2.mozilla-vpn-client.branch.{branch}.latest.{name}
```
We already index the decision task there, so would just be a matter of indexing the builds as well:
https://firefox-ci-tc.services.mozilla.com/tasks/index/mozillavpn.v2.mozilla-vpn-client.branch.main

The good news (or maybe bad if you spent a lot of time on this) is that the `generic` index builder should do this for us already. So you can likely just add the `index` key to the signing tasks and it might work.

ahal: My bad, the `generic` index build does not add these `branch` routes.

So your patch is still necessary, but I'd suggest copying something similar to the above route. We'll likely want to use the unsigned builds here as well as I think we may stop doing build signing on every push eventually.

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,20 @@
...
+
+@transforms.add
+def add_release_index(config, tasks):
+   for task in tasks:
+        if "add-release-index" in task:
+            release_index = task.pop("add-release-index")
+            if release_index and int(config.params["level"]) == 3:
+                name = task['name'].split("/")[0]
+                version = config.params["app_version"]
+                route = f"index.mozillavpn.v2.mozillavpn.releases.{version}.{name}"
```
ahal, line 18: So given these aren't actually releases, I think we should use the following namespace:
`mozillavpn.v2.mozilla-vpn-client.branch.{branch}.{name}`

This lines up well with index route conventions across other projects, and we also already index Decision tasks there:
https://firefox-ci-tc.services.mozilla.com/tasks/index/mozillavpn.v2.mozilla-vpn-client.branch.main

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,23 @@
...
+from taskgraph.transforms.base import TransformSequence
+transforms = TransformSequence()
+
+_GIT_REFS_HEADS_PREFIX = "refs/heads/"
+
+@transforms.add
+def add_branch_index(config, tasks):
+   for task in tasks:
+        if "add-branch-index" in task:
+            release_index = task.pop("add-branch-index")
```
ahal, line 15: nit: s/release_index/branch_index

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -0,0 +1,23 @@
...
+            if release_index and int(config.params["level"]) == 3:
```
ahal, line 16: nit: You can combine your two if statements like so:
```
add_branch_index = task.pop("add-branch-index", False)
if add_branch_index and int(config.params["level"]) == 3:
```

ahal (approved):
Awesome, thanks!

PR #3765, opened 2022-06-16
Fix an error in the translation string script

bakulf:


scripts/utils/generate_ts.sh
```diff
@@ -82,6 +82,7 @@ done
 
 printn Y "Remove cache... "
 rm -rf cache || die
+rm -rf addon_ts || die
```
flodolo, line 85: You can't remove this cache, it's needed by the extraction scripts on the l10n side. That's also why it's in `.gitignore`. Maybe add a comment? I thought I did, or at least commented on why I was using a different folder.

https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/workflows/update.yaml#L53

flodolo (changes requested):

flodolo (approved):

PR #3781, opened 2022-06-18
Add more tests for the auth-in-app

bakulf:


src/ui/authenticationInApp/ViewAuthenticationSignUp.qml
```diff
@@ -12,15 +12,8 @@ import components.inAppAuth 0.1
 
 
 VPNInAppAuthenticationBase {
-    // TODO
```
flozia, line 15: Dang — I was under the impression I already removed those.

flozia (approved):
Nice!

bakulf:
I added more tests to cover all the error code. And I found 2 issues:
- a error code that is not used
- a typo in a QML file

PR #3803, opened 2022-06-21
[2.9.0] Properly disable qmljsdebugger in release mode

andersk:
## Description

This is a 2.9.0 backport of the security fix for both qmake and CMake in #3539.

> * In qmake, the `debug` variable and the `release` variable are both present by default.  We need to check which one is active, by virtue of having been set later.
>
>   https://doc.qt.io/qt-6/qmake-test-function-reference.html#config-config
>
> * In CMake, the `"${CMAKE_BUILD_TYPE}"` test did not do what it looks like it does.  CMake evaluates `if("${CMAKE_BUILD_TYPE}")` to `if(Release)`, which tests whether the *variable named `Release`* is defined; it isn’t, so we were unconditionally setting `CMAKE_BUILD_TYPE` to `Debug`.
>
>   https://cmake.org/cmake/help/latest/command/if.html#variable-expansion
>
>   We need to check whether `CMAKE_BUILD_TYPE` variable itself is defined, and we need to do so before `project()` defines it by default.

(There was a trivial merge conflict on the line with the version number, and I wanted to make sure this isn’t forgotten for the 2.9.0 release that appears to be imminent.)

## Checklist
    
### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,17 +6,17 @@ cmake_minimum_required(VERSION 3.16)
 
 option(BUILD_DUMMY "Build for the dummy platform" OFF)
 
-project("Mozilla VPN" VERSION 2.9.0 LANGUAGES C CXX
-        DESCRIPTION "A fast, secure and easy to use VPN. Built by the makers of Firefox."
-        HOMEPAGE_URL "https://vpn.mozilla.org"
-)
-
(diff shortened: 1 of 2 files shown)


bakulf (approved):

andersk:
Thanks. Please also merge #3539 to `main` so this doesn’t regress later.

PR #3809, opened 2022-06-22
Remove RPM github action

bakulf:
They break the linux package generation because we are unable to include the addons in the RPM package yet.


strseb (approved):
:c

PR #3815, opened 2022-06-22
Fix Missing Text in segmented button

### LINKED ISSUE #3437: [Android] “Single-hop” / “Multi-hop” strings are missing from the “Select location” screen
**Affected versions:**

* Mozilla VPN 2.9.0 (16511830);

**Tested Platforms:**

* Android;

**Prerequisites:**

* Have Mozilla VPN app installed and opened;

* Have a FxA account with an active subscription for Mozilla VPN;

* Be signed in; 

**Steps to reproduce:**

1. Tap on the “Select location” option form the activation screen;
1. Observe the “Select location” screen toggle options;

**Expected result:**

* “Single-hop” and “Multi-hop” strings are correctly displayed on the “Select location” screen;

**Actual result:**

* “Single-hop” and “Multi-hop” strings are not displayed at all on the “Select location” screen. 

**Notes:**

* See the attached screenshot;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2158)
┆author: Bianca Hidecuti

strseb:
## Description

ElideRight seems to cause android to simply not render any text, so lets make it conditional 

Screenshot 2022-06-22 142416

## Reference

Fixes https://mozilla-hub.atlassian.net/browse/VPN-2158
Closes #3437

### DIFF
--- a/nebula/ui/components/VPNSegmentedToggle.qml
+++ b/nebula/ui/components/VPNSegmentedToggle.qml
@@ -106,7 +106,9 @@ Rectangle {
                     text: VPNl18n[segmentLabelStringId]
                     horizontalAlignment: Text.AlignHCenter
                     verticalAlignment: Text.AlignVCenter
-                    elide: Text.ElideRight
+                    // Bug VPN-2158 - Apparently QText is not rendered on android
+                    // When we select any other elide then none
+                    elide: Qt.platform.os === "android" ? Text.ElideNone : Text.ElideRight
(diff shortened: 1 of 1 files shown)


MattLichtenstein (approved):
Wow, how strange. My only concern is that this is not the only location where we elide text / use `ElideRight` - do we need an app-wide audit of the usage of eliding text?

strseb:
We're using it in quite a few places. So maybe there are a few more pre-conditions that need to be met for this to occur :/

PR #3824, opened 2022-06-22
Support back-button on android in the guide view - fixes VPN-2435

bakulf:
The issue is in how we load ViewGuide.qml: always on the mainview stack. We do
it also when we are in settings.  When we go back, the app removes a layer on
top stack view (settings) and the tips&tricks view goes away, instead of the
guide.


MattLichtenstein (approved):
👍

lesleyjanenorton (approved):

PR #3833, opened 2022-06-23
Parse subscription data depending on subscription type

flozia:
## Description

Moves parsing of the subscription data per subscription type in the client.

## Reference

[VPN-2428](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-2428)

## Checklist
    

src/subscriptiondata.cpp
```diff
@@ -33,36 +34,8 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
...
-  m_status = status.toString();
-
-  QJsonValue type = obj.value("type");
-  if (!type.isString()) {
-    return false;
-  }
-  m_type = type.toString();
-
   // Plan
+  logger.debug() << "Parse plan start";
```
bakulf, line 38: remove these comments. It seems to me that we have too many :)

src/subscriptiondata.cpp
```diff
@@ -83,59 +56,138 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
...
+          type.toString().toUtf8()));
+
+  // Parse subscription data depending on subscription platform
+  switch (m_type) {
+    case web:
+      if (!parseSubscriptionDataWeb(subscriptionData)) {
+        return false;
+      }
+      break;
+    case iap_apple:
```
bakulf, line 131: [[fallthrough]]

src/subscriptiondata.cpp
```diff
@@ -83,59 +56,138 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
...
+  m_expiresOn = subscriptionData.value("current_period_end").toInt();
+  if (!m_expiresOn) {
+    return false;
+  }
+  QJsonValue isCancelled = subscriptionData.value("cancel_at_period_end");
+  if (!isCancelled.isBool()) {
+    return false;
+  }
+  m_isCancelled = isCancelled.toBool();
+  QJsonValue status = subscriptionData.value("status");
```
bakulf, line 185:
```
m_status = subscriptionData["status"].toString();
if (m_status.isEmpty()) {
  return false;
}
```

src/subscriptiondata.cpp
```diff
@@ -83,59 +56,138 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
...
+
+  m_createdAt = subscriptionData.value("created").toInt();
+  if (!m_createdAt) {
+    return false;
+  }
+  m_expiresOn = subscriptionData.value("current_period_end").toInt();
+  if (!m_expiresOn) {
+    return false;
+  }
+  QJsonValue isCancelled = subscriptionData.value("cancel_at_period_end");
```
bakulf, line 180: Use `subscriptionData["foo"]` instead of `.value()`

bakulf (changes requested):
Nice!

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  QJsonObject obj = doc.object();
+
+  // Plan
+  logger.debug() << "Parse plan start";
+  QJsonObject planData = obj["plan"].toObject();
+
+  m_planAmount = planData["amount"].toInt();
+  if (!m_planAmount && m_planAmount != 0) {
```
bakulf, line 42: What does this mean? :)
` if (!m_planAmount) {`

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  QJsonValue planCurrency = planData["currency"];
+  if (!planCurrency.isString()) {
+    return false;
+  }
+  // We transform currency code to uppercase in order to be compliant with
+  // ISO 4216. FxA passes the lowercase currency code values without
+  // transformation from Stripe.
+  m_planCurrency = planCurrency.toString().toUpper();
```
bakulf, line 53: Let's write lines 46-53 in this way:
```
  // We transform currency code to uppercase in order to be compliant with
  // ISO 4216. FxA passes the lowercase currency code values without
  // transformation from Stripe.
  m_planCurrency = planData["currency"].toString().toUpper();
  if (m_planCurrency.isEmpty()) {
    return false;
  }
```

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return false;
+  }
+  logger.debug() << "Parse plan ready";
+
+  // Payment
+  QJsonObject paymentData = obj["payment"].toObject();
+
+  // If we do not receive payment data from FxA we don’t show it instead of
+  // throwing an error. There is a known bug with FxA which causes FxA to not
+  // show payment info: https://mozilla-hub.atlassian.net/browse/FXA-3856.
```
bakulf, line 66: we need to have a test for this!

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+      if (!m_creditCardExpYear) {
+        return false;
+      }
+    }
+  }
+
+  // Subscription
+  QJsonObject subscriptionData = obj["subscription"].toObject();
+
+  QJsonValue type = subscriptionData["_subscription_type"];
```
bakulf, line 105: here as well, more than the type, we care if we have the value:
```
QString type = subscriptionData["_subscription_type"].toString();
if (type.isEmpty()) {
  return false;
}
```

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+  if (!type.isString()) {
+    return false;
+  }
+
+  // Enum from string
+  m_type = static_cast<TypeSubscription>(
+      QMetaEnum::fromType<TypeSubscription>().keyToValue(
+          type.toString().toUtf8()));
```
bakulf, line 113: Use the `ok` boolean check:

```
bool ok = false;
m_type = static_cast<TypeSubscription>(
      QMetaEnum::fromType<TypeSubscription>().keyToValue(
          type.toUtf8(), &ok));
if (!ok) {
logger.error() << "Unsupported subscription type:" << type;
  return false;
}
```

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,177 @@
...
+bool SubscriptionData::parseSubscriptionDataIap(
+    const QJsonObject& subscriptionData) {
+  logger.debug() << "Parse IAP start" << m_type;
+
+  m_expiresOn = subscriptionData["expiry_time_millis"].toInt();
+  if (!m_expiresOn) {
+    return false;
+  }
+  QJsonValue autoRenewing = subscriptionData["auto_renewing"];
+  if (!autoRenewing.isBool()) {
```
bakulf, line 150: this check can be removed:
`m_isCancelled = !subscriptionData["auto_renewing"].toBool();`

bakulf (changes requested):
We also need to check why state is inactive for IAP.

src/models/subscriptiondata.cpp
bakulf: Follow up?

src/models/subscriptiondata.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  m_expiresOn = subscriptionData["expiry_time_millis"].toVariant().toLongLong();
+  if (!m_expiresOn) {
+    return false;
+  }
+  m_isCancelled = !subscriptionData["auto_renewing"].toBool();
+
+  quint64 now = QDateTime::currentMSecsSinceEpoch();
+  if (now <= m_expiresOn) {
+    m_status = "active";
```
bakulf, line 158: I would like to see an enum here. m_status should not be parsed twice (here and in the QML code). Follow-up.

tests/functional/testSubscription.js
```diff
@@ -0,0 +1,426 @@
...
+        subscription: {
+          value: {
+            _subscription_type: 'iap_google',
+            expiry_time_millis: 2147483647000,
+            auto_renewing: false,
+          },
+          expected: {
+            cancelled: '1/19/38',
+            label: 'Expires',
+            status: 'Active'
```
bakulf, line 251: Because the tests should not know how the code works, can we have inactive/active for both google/apple? Basically 4 tests.

bakulf (approved):

PR #3839, opened 2022-06-24
tips and tricks telemetry

MattLichtenstein:
## Description

- Record telemetry for tips and tricks when:
  - The tips and tricks page is opened
  - The tips and tricks intro modal is shown
  - The tips and tricks intro modal is closed (with action <dismissed | cta>)
  - A guide is opened (with guide id)
  - A guide is closed (with guide id and duration of how long the guide was viewed for (in ms))
  - A tutorial was started, ended, or aborted (with tutorial id)
  - A tutorial step was viewed (with tutorial id and step id)

Data Review: https://bugzilla.mozilla.org/show_bug.cgi?id=1776499
Data Review Status: r+

## Reference

[VPN-2292: Instrumentation for Tips & Tricks](https://mozilla-hub.atlassian.net/browse/VPN-2292)

## Checklist
    

src/ui/settings/ViewGuide.qml
```diff
@@ -252,5 +252,16 @@ Item {
         target: VPNCloseEventHandler
     }
 
-    Component.onCompleted: VPNCloseEventHandler.addView(root)
+    Component.onCompleted: {
+//        console.log("Guide view created")
```
lesleyjanenorton, line 256: don't think we need this 🙂

src/ui/views/ViewMain.qml
```diff
@@ -206,8 +207,15 @@ VPNFlickable {
...
             ]
 
-            onOpened: VPNSettings.tipsAndTricksIntroShown = true
-            onClosed: tipsAndTricksIntroPopupLoader.active = false
+            onOpened: {
+                VPNSettings.tipsAndTricksIntroShown = true
+                Sample.tipsAndTricksModalShown.record();
+            }
+
+            onClosed: {
```
lesleyjanenorton, line 215: Does this `onClosed` event fire regardless of which button is clicked? If so, and if the purpose of the ping is to track modal _dismissals_, we may want to attach this specifically to the `onClicked` handlers for `tipsAndTricksIntroPopupGoBackButton` and `tipsAndTricksIntroPopupCloseButton`.

MattLichtenstein: Right, I was sort of under the impression that "dismissed" meant closed, which, technically all of the buttons do, but I can go ahead and change it so the event is not fired when the primary button is clicked.

MattLichtenstein: I've kept the trigger in `onClosed` because there are technically other ways to close the popup besides the "X" and "Go back" CTA (i.e. hitting "Esc", clicking/pressing outside the modal) but I've prevented the event from firing when closed via the primary "Discover now" CTA

src/ui/settings/ViewGuide.qml
```diff
@@ -253,4 +253,11 @@ Item {
     }
 
     Component.onCompleted: VPNCloseEventHandler.addView(root)
+
+    Component.onDestruction: {
+        VPN.recordGleanEventWithExtraKeys("guideClosed",{
+                                          "id": guide.id,
+                                          "duration": `${((new Date().getTime() - timeOfOpen) / 1000)} seconds`
```
birdsarah, line 260: This should be a number and the key should indicate the units e.g. "duration_s": 1234

As a string analysis will be a big pain.

src/ui/views/ViewMain.qml
birdsarah: The goal is analysis that makes sense.

We'll have a count for tipsAndTricksModalShown

For every modal shown there should be a "modal gone" type event. That could be because it was dismissed or maybe people interacted with it and went to the content.

What events can I sum to see the same number of "modal gone" as "modal shown"?

birdsarah (changes requested):

src/ui/settings/ViewGuide.qml
birdsarah: I don't think you need to worry about handling that. Just do the math. You could also just make it milliseconds so "duration_ms" and then you don't need to do the division.

src/ui/views/ViewMain.qml
birdsarah: @MattLichtenstein posting your note here:

So as it stands right now (after making the change Lesley suggested), there is no combination of metrics that would equate to the "modal shown" count. Are you alluding to adding another metric for primary button clicks. Because that, alongside dismissals, sums up everything the user can do when they are shown the modal.

I didn't add that originally because it was specifically called out in the spec. What I did have originally was that those primary button clicks also triggered a "modal dismissed" event (so dismissed events would equal shown events), but I have since removed that. I could bring that back if you think that is the better option

@lesleyjanenorton I'd like your input here.

bakulf (changes requested):
There are 3 places where telemetry can be added:
1. UI: as this PR does
2. in the C++ layer
3. in the addon engine

Can you tell me why you pick this approach? I would vote for 2 or 3. 3 is probably better because it means that each addon can in the future has its own telemetry events. But I'm opened to different solutions if well motivated :)

birdsarah:
> There are 3 places where telemetry can be added:
> 
>     1. UI: as this PR does
> 
>     2. in the C++ layer
> 
>     3. in the addon engine
> 
> 
> Can you tell me why you pick this approach? I would vote for 2 or 3. 3 is probably better because it means that each addon can in the future has its own telemetry events. But I'm opened to different solutions if well motivated :)

@bakulf I'd be interested in working up some guidance on this and adding it to the README (https://github.com/mozilla-mobile/mozilla-vpn-client#working-on-tickets-with-new-glean-events). I don't have firm opinions about 1 vs 2 and I haven't seen glean events in addon engine yet (excited to).

src/ui/views/ViewMain.qml
lesleyjanenorton: > Given that this event is called "Sample.tipsAndTricksModalClosed" we could extra key of "how" with "dismissed" | "cta" as values.

I like it. Let's handle it this way.

MattLichtenstein:
> > 
> 
> There are 3 places where telemetry can be added:
> 
>     UI: as this PR does
>     in the C++ layer
>     in the addon engine
> 
> Can you tell me why you pick this approach? I would vote for 2 or 3. 3 is probably better because it means that each addon can in the future has its own telemetry events. But I'm opened to different solutions if well motivated :)

I picked the UI approach mainly because these are interaction-based events that are triggered directly from the UI, so I thought that made the most sense. I do see how it would be appropriate to abstract some of them out to the c++ layer, so I have moved `tutorialStarted`, `tutorialCompleted` and `tutorialStepViewed` into `tutorial.cpp`

src/tutorial/tutorial.cpp
```diff
@@ -35,6 +37,8 @@ void Tutorial::allowItem(const QString& objectName) {
   m_allowedItems.append(objectName);
 }
 
+QString Tutorial::currentTutorialId() { return m_currentTutorial->id(); }
```
bakulf, line 40: 1. What about m_currentTutorial is null?
2. `Tutorial::currentTutorialId() const`

src/tutorial/tutorial.h
```diff
@@ -26,6 +26,7 @@ class Tutorial final : public QObject {
   Q_INVOKABLE void play(Addon* tutorial);
   Q_INVOKABLE void stop();
   Q_INVOKABLE void allowItem(const QString& objectName);
+  Q_INVOKABLE QString currentTutorialId();
```
bakulf, line 29: Instead of exposing the ID, what about if you expose the whole `m_currentTutorial` ? And it can be a property:
`Q_PROPERY(Addon* currentTutorial MEMBER m_currentTutorial NOTIFY playingChanged)`

nebula/ui/components/VPNTutorialPopups.qml
```diff
@@ -235,9 +235,11 @@ Item {
         }
 
         tutorialPopup.secondaryButtonOnClicked = () => {
+            VPN.recordGleanEventWithExtraKeys("tutorialAborted", {"id": VPNTutorial.currentTutorialId()});
```
bakulf, line 238: `VPNTutorial.currentTutorial.id`

bakulf (changes requested):

src/tutorial/tutorial.h
MattLichtenstein: That makes sense, but why `Q_PROPERTY(Addon*...)` instead of `Q_PROPERTY(AddonTutorial*...)`

src/tutorial/tutorial.cpp
MattLichtenstein: Will remove this in favor of the `Q_PROPERTY`

src/tutorial/tutorial.h
```diff
@@ -40,6 +42,7 @@ class Tutorial final : public QObject {
   void tooltipNeeded(const QString& tooltipText, QObject* targetElement);
   void tooltipShownChanged();
   void tutorialCompleted(const QString& completionMessageText);
+  void currentTutorialChanged();
```
bakulf, line 45: Why this?

bakulf (approved):
Remove the unused signal. It looks good !

birdsarah (approved):

PR #3876, opened 2022-06-29
Add tests for the auth in subscription and account deletion

bakulf:
VPN-2454

### DIFF
--- a/tests/functional/testSubscription.js
+++ b/tests/functional/testSubscription.js
@@ -5,6 +5,25 @@
 const assert = require('assert');
 const vpn = require('./helper.js');
 
+const SUBSCRIPTION_DETAILS = {
+  plan: {amount: 123, currency: 'usd', interval: 'year', interval_count: 1},
+  payment: {
+    payment_provider: 'stripe',
+    payment_type: 'credit',
+    last4: '1234',
+    exp_month: 12,
+    exp_year: 2022,
+    brand: 'visa',
+  },
+  subscription: {
+    _subscription_type: 'web',
+    created: 1,
+    current_period_end: 2,
+    cancel_at_period_end: true,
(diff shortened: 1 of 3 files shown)


tests/functional/testSubscription.js
```diff
@@ -19,14 +38,195 @@ describe('Subscription view', function() {
...
+    await vpn.waitForElementProperty('authSignIn-button', 'enabled', 'true');
+
+    this.ctx.guardianSubscriptionDetailsCallback = req => {
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/subscriptionDetails']
+          .status = 200;
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/subscriptionDetails']
+          .body = SUBSCRIPTION_DETAILS;
+    };
+
+    await vpn.clickOnElement('authSignIn-button');
```
flozia, line 131: The button is not clicked if I’m running the test locally. Adding `await vpn.wait();` before this line fixes the issue. Which is weird, because we are already waiting for the button to enabled above.

tests/functional/testSubscription.js
```diff
@@ -19,14 +38,195 @@ describe('Subscription view', function() {
...
+    this.ctx.fxaLoginCallback = (req) => {
+      this.ctx.fxaOverrideEndpoints.POSTs['/v1/account/login'].body = {
+        sessionToken: 'session',
+        verified: false,
+        verificationMethod: 'totp-2fa'
+      };
+      this.ctx.fxaOverrideEndpoints.POSTs['/v1/account/login'].status = 200;
+    };
+
+    await vpn.clickOnElement('authSignIn-button');
```
flozia, line 187: Same thing here.

PR #3898, opened 2022-06-30
Deprecate qmake for Linux

oskirby:
## Description
We have been using CMake exclusively for Linux builds for the last couple of months, and this has proven to be a better development experience, and one that tends to have fewer bugs overall. I think it means that we are finally ready to remove support for `qmake`

## Checklist
    

src/src.pro
```diff
@@ -32,8 +32,6 @@ include($$PWD/qmake/sources.pri)
 # Platform-specific entries:
 DUMMY {
    include($$PWD/qmake/platforms/dummy.pri)
-} else:linux:!android {
-   include($$PWD/qmake/platforms/linux.pri)
```
bakulf, line 36: What about if we keep these 2 lines and we add: `error(Please use cmake)`

bakulf (approved):

bakulf (approved):

PR #3939, opened 2022-07-06
Document that right now we rely on CMAKE_INSTALL_PREFIX=/usr

emilio:
We hard-code /usr/bin/mozillavpn in a couple places on Linux.

The right fix would be to make the relevant files:

 * `linux/extra/mozillavpn-startup.desktop`
 * `linux/extra/mozillavpn.desktop`
 * `linux/mozillavpn.service`
 * `src/platforms/linux/daemon/org.mozilla.vpn.dbus.service`

Generated at build-time, using the specified `CMAKE_INSTALL_PREFIX`
instead. But I'm not a CMake expert and I'm not sure what's the best way
to do that.

Instead, document that this flag is needed, since a lot of distros
default to /usr/local otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ PATH the build process is effectively the same on each of the supported platform
 1. Start by creating a build directory and using `cmake` to generate the Makefiles.
 
 ```bash
-mkdir build && cmake -S . -B build
+mkdir build && cmake -S . -B -DCMAKE_INSTALL_PREFIX=/usr build
 ```
 
 This generation step can be augmented by providing variable definitions on the


oskirby (approved):
I think this is fine, but it should no longer be necessary as soon as #3939 gets merged, and fixes our hard-coded dependency on `/usr/`

emilio:
Ah, great, I was not aware of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3888, happy to close in favor of that (or if you think it's gonna take a while we should probably merge this in the meantime, your call)

PR #3943, opened 2022-07-06
Fix the windows installer path for 2.9

bakulf:


windows/installer/MozillaVPN_prod.wxs
```diff
@@ -121,7 +118,7 @@
       Merge modules
     -->
     <DirectoryRef Id="MozillaVPNFolder">
-      <Merge Id="VCRedistModule" SourceFile="../../Microsoft_VC142_CRT_$(var.Platform).msm" DiskId="1" Language="0"/>
```
oskirby, line 124: I'm suspicious of this change, as `scripts/windows/compile.bat` is still copying this file as `Microsoft_VC142_CRT_x64.msm` in which case the `.wxs` file will not locate the merge module correctly.

oskirby (approved):

windows/installer/MozillaVPN_prod.wxs
bakulf: we use cmake for 2.9.

PR #3979, opened 2022-07-08
Removed the survey models and alerts and tests

bakulf:
If we move the survey logic to addons, we don't need this code.


oskirby (approved):
Wonderful! I love to see stale code get deleted.

Issue #3960, opened 2022-07-07
[iOS] Crash after swipe kill while on Profile screen


data-sync-user:
**VPN version:**

* 2.9.0 (2.202207052010);

**Platforms:**

* iOS;

**Prerequisites:**

* Have a fresh install of the VPN app;
* Be in Profile screen;

**Steps to reproduce:**

1. Swipe kill the app;
1. Observe the behavior;

**Expected results:**

* App is closed with success.

**Actual results:**

* App crashes.

**Notes:**

* Issue is reproducing if swipe-killing the app while in other screens as well;

* Attaching logs;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2548)
┆author: Valentina Virlics

data-sync-user:
➤ Florian Zia commented:

I’m able to reproduce this on every settings page I tried except the main settings view.

data-sync-user:
➤ Valentina Virlics commented:

This might have the same root cause as https://mozilla-hub.atlassian.net/browse/VPN-1979 ( https://mozilla-hub.atlassian.net/browse/VPN-1979|smart-link )

data-sync-user:
➤ Rebecca Billings commented:

Confirmed that this issue still repros. Earlier it was suggested that this requires a QT6 update which we haven’t seen merged. When this is fixed, please include build and testing info.

data-sync-user:
➤ Valentina Virlics commented:

Verified this using the Mozilla VPN 2.9.0 (2.202207122010) build on iOS and the crash is still reproducing.

Reopening.

data-sync-user:
➤ Valentina Virlics commented:

Verified this on the Mozilla VPN 2.9.0 (2.202207142010) build from TestFlight.

The crash is no longer reproducing at swipe kill from different settings screens.

Issue #3465, opened 2022-05-06
Mozilla VPN app for Android crashes when tapping the Sign Up button


ManuelCalidonio:
**Description**

Users are reporting a crash when tapping on the Sign up button on the Mozilla VPN app for Android.

**Technical Information**

Platform: Android 11
FX product: Mozilla VPN:
VPN version: 2.8.0

**Steps to Reproduce**
1. Open the Mozilla VPN app
2. Log out of the Mozilla VPN app (if you are logged in)
2. Tap Sign up
3. Tap the back arrow on the top left corner
4. Tap Sign up again

**Actual Results**

The app crashes. This is not a constant behavior, the app sometimes crashes when Sign up is tapped the first time. Sometimes it does not crash at all when following the steps above. 

**Expected Results**

The app should not crash

data-sync-user:
➤ Sebastian Streich commented:

It seems this was fixed inbetween 6.2.2->6.2.4, ryan cannot reproduce with the 2.9 client from taskcluster.

PR #3980, opened 2022-07-08
Bump Qt version / Maybe fix iOS crash?

strseb:
## Description
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3960
Sounds very much like a bug we had in android that we had resolved by upgrading to 6.2.4. - Let's do that for ios too, we can always roll back if it breaks things

Note: Merging this into main will bump the next ios build but not windows/macos. - :)

### DIFF
--- a/taskcluster/ci/toolchain/qt.yml
+++ b/taskcluster/ci/toolchain/qt.yml
@@ -5,7 +5,7 @@
 job-defaults:
     worker:
         env:
-            QT_VERSION: "6.2.3"
+            QT_VERSION: "6.2.4"
             QT_MAJOR: "6.2"
         max-run-time: 14400
 


bakulf (approved):

strseb:
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3465 - This was the one i was thinking about. Here qt crashed at running destructors for qml objects.

bakulf:
would that update this URL `https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozillavpn.cache.level-3.toolchains.v3.qt-ios.latest/artifacts/public%2Fbuild%2Fqt6_ios.zip` ?

PR #3981, opened 2022-07-08
Dismiss addon messages

bakulf:


src/addons/addonmessage.cpp
```diff
@@ -16,6 +17,15 @@ Logger logger(LOG_MAIN, "AddonMessage");
 Addon* AddonMessage::create(QObject* parent, const QString& manifestFileName,
                             const QString& id, const QString& name,
                             const QJsonObject& obj) {
+  SettingsHolder* settingsHolder = SettingsHolder::instance();
+  Q_ASSERT(settingsHolder);
+
+  QStringList dismissedAddonMessages = settingsHolder->dismissedAddonMessages();
```
oskirby, line 23: I wonder if it might be worthwhile to consider some kind of generic local storage for the addon to handle this sort of usecase. For example an addon that notifies a user of an update might want to be a bit more dynamic and offer "dismiss forever/remind me later" kind of behaviours, or reset the dismissal state on the next update.

This is not actually requesting any change. I'm just airing my half-formed thoughts as they appear.

oskirby (approved):

PR #3983, opened 2022-07-09
VPN-2261 - Implement handler for DEVICE_DELETED push message

brizental:
## Description

I also did some housekeeping work to break the websockets module into multiple files, because with the `pushmessage` module also in there, it would get too crowded.

Note: I suggest reviewing this commit-by-commit.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2261

## Checklist
    
### DIFF
--- a/src/websocket/pushmessage.cpp
+++ b/src/websocket/pushmessage.cpp
@@ -1,0 +1,114 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "pushmessage.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QJsonObject>
+#include <QJsonDocument>
+#include <QJsonValue>
+
+namespace {
+Logger logger(LOG_MAIN, "PushMessage");
+}  // namespace
+
+/**
+ * @brief Abstraction over a message sent from the push server.
+ *
+ * Messages are expected to be a valid JSON with the `type` and (optional)
+ * `payload` attributes. Example:
+ * ```json
+ * {
+ *   "type": "DEVICE_DELETED",
+ *   "payload": {
+ *     "deviceId": "e95cfedd-119b-4ef1-aa51-46f9427e2337"
+ *   }
+ * }
+ * ```
+ *
+ * The `type` can be one of a set of known possibilities. See
+ * `PushMessage::messageTypeFromString`, for a comprehensive list of known
+ * message types.
+ *
+ * The payload format will depend on the `type` of the message and may be
+ * omiited. When present,  it will contain information necessary for executing
+ * whatever action is according to the message `type`.
+ */
+PushMessage::PushMessage(const QString& message)
+    : m_messageType(MessageType_UnknownMessage),
+      m_messagePayload(QJsonObject()) {
+  MVPN_COUNT_CTOR(PushMessage);
+  parseMessage(message);
+}
+
+PushMessage::~PushMessage() { MVPN_COUNT_DTOR(PushMessage); }
+
+/**
+ * @brief Parses messages send from the push server.
+ *
+ * If the message is wrongly formatted or of an unknown type,
+ * this function will not error.
+ */
+void PushMessage::parseMessage(const QString& message) {
+  QJsonObject obj;
+  QJsonDocument doc = QJsonDocument::fromJson(message.toUtf8());
+
+  if (!doc.isNull()) {
+    if (doc.isObject()) {
+      obj = doc.object();
+
+      m_messagePayload = obj.take("payload").toObject();
+
+      QString messageType = obj.take("type").toString();
+      m_messageType = messageTypeFromString(messageType);
+    } else {
+      logger.debug() << "Push message is a not a JSON object:" << message;
+    }
+  } else {
+    logger.debug() << "Push message is invalid JSON:" << message;
+  }
+}
+
+/**
+ * @brief Infers the message type from the `type` extracted from the given JSON
+ * string.
+ *
+ * Note: These should be in sync with the type sent by the push server.
+ * See:
+ * https://github.com/mozilla-services/guardian-website/blob/7da6d52b334e35bd858bd22c7add5cde7ced9ef2/server/websocket/messages.ts
+ */
+// static
+PushMessage::MessageType PushMessage::messageTypeFromString(QString& str) {
+#ifdef UNIT_TEST
+  if (str == "TEST_MESSAGE") return MessageType_TestMessage;
+#endif
+
+  logger.debug() << "Unknown message type" << str;
+  return MessageType_UnknownMessage;
+}
+
+/**
+ * @brief Executes the action related to the given message.
+ *
+ * If the message is malformatted or of unknown type, this is a no-op.
+ *
+ * The `onExecutionCompleted` is emited this function is done.
+ * If nothing was done e.g. if the message was unknown, it will emit a `false`
+ * value.
+ */
+bool PushMessage::executeAction() {
+  switch (m_messageType) {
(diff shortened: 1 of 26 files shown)


src/websocket/pushmessage.cpp
```diff
@@ -0,0 +1,162 @@
...
+ * If the current device was the one that got deleted, user will be logged out.
+ * Otherwise, deleted device will be deleted from the settings.
+ */
+// static
+bool PushMessage::handleDeviceDeleted(QJsonObject& payload) {
+  bool executionResult = true;
+
+  const QString& publicKey = payload.take("publicKey").toString();
+  logger.debug() << publicKey;
+  if (publicKey.isNull() || publicKey.isEmpty()) {
```
bakulf, line 146: isEmpty() is enough.

src/websocket/pushmessage.cpp
```diff
@@ -0,0 +1,162 @@
...
+ *    "publicKey": "<device public key>",
+ * }
+ * ```
+ *
+ * If the current device was the one that got deleted, user will be logged out.
+ * Otherwise, deleted device will be deleted from the settings.
+ */
+// static
+bool PushMessage::handleDeviceDeleted(QJsonObject& payload) {
+  bool executionResult = true;
```
bakulf, line 142: this variable is not used for real.

src/websocket/pushmessage.cpp
```diff
@@ -0,0 +1,162 @@
...
+        << "Malformed message payload for DeviceDeleted message. Ignoring.";
+    return false;
+  }
+
+  MozillaVPN* vpn = MozillaVPN::instance();
+  Q_ASSERT(vpn);
+  if (vpn->keys()->publicKey() == publicKey) {
+    logger.error() << "Current device has been deleted from this subscription.";
+    vpn->reset(true);
+    return executionResult;
```
bakulf, line 157: return true

src/websocket/pushmessage.cpp
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  MozillaVPN::instance()->deviceRemoved(publicKey);
+  return executionResult;
```
bakulf, line 161: return true

tests/functional/testPushMessageHandlers.js
```diff
@@ -0,0 +1,157 @@
...
+
+    it('attempting to delete a third device updates the device list', async () => {
+      // Wait for viewMain to load, then wait a bit more for good measure.
+      await vpn.waitForElement("viewMainFlickable");
+      await vpn.wait();
+
+      let devices = await vpn.getDevices();
+      // Get a device from the list of devices that is not the current one.
+      let thirdDevice = devices.find(device => !device.currentDevice);
+      while(thirdDevice) {
```
bakulf, line 107: why a loop? is it not enough 1 single broadcastMessage?

bakulf (changes requested):

macos/pkg/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ add_custom_command(
     DEPENDS mozillavpn
     COMMAND ${CMAKE_COMMAND} -E remove_directory -f ${CMAKE_CURRENT_BINARY_DIR}/staging
     COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/staging
-    COMMAND ${CMAKE_COMMAND} -E copy_directory $<TARGET_BUNDLE_DIR:mozillavpn>
+    COMMAND ${CMAKE_COMMAND} -E copy_directory $<TARGET_FILE_DIR:mozillavpn>
```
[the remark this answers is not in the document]
brizental: Yeap, it's unrelated, however without this change I was not able to work on the changes I made here and it was so small I decided to keep it. However, I am not too sure about it. From the commit message:

> My experience with CMake is none, so I am not really sure what was going
on here. What I know is that I was getting the following error:
> 
> TARGET_BUNDLE_DIR is allowed only for Bundle targets
> 
> And with the change in this commit I don't get it anymore. Added this in
a separate commit so that it can be easily dropped if this was just me
doing something wrong.

brizental:
Hey @baku, so there is this error hapenning on the CI now 

```
/usr/bin/ld: CMakeFiles/unit_tests.dir/__/__/src/websocket/websockethandler.cpp.o: in function `WebSocketHandler::onMessageReceived(QString const&)':
websockethandler.cpp:(.text+0x1c7f): undefined reference to `PushMessage::PushMessage(QString const&)'
/usr/bin/ld: websockethandler.cpp:(.text+0x1ca8): undefined reference to `PushMessage::executeAction()'
/usr/bin/ld: CMakeFiles/unit_tests.dir/__/__/src/websocket/websockethandler.cpp.o: in function `PushMessage::~PushMessage()':
websockethandler.cpp:(.text._ZN11PushMessageD2Ev[_ZN11PushMessageD5Ev]+0x1c): undefined reference to `vtable for PushMessage'
collect2: error: ld returned 1 exit status
```

If I explicitly declare the destrutor for PushMessage it goes away. However if I do that it fails for me locally with an error that is something like "Redeclaration of implicitly declared dtor". I am not sure how to address this do you have any idea?

tests/unit/CMakeLists.txt
```diff
@@ -212,6 +212,10 @@ target_sources(unit_tests PRIVATE
     ${MVPN_SOURCE_DIR}/update/webupdater.h
     ${MVPN_SOURCE_DIR}/urlopener.cpp
     ${MVPN_SOURCE_DIR}/urlopener.h
+    ${MVPN_SOURCE_DIR}/websocket/exponentialbackoffstrategy.cpp
+    ${MVPN_SOURCE_DIR}/websocket/exponentialbackoffstrategy.h
+    ${MVPN_SOURCE_DIR}/websocket/websockethandler.cpp
+    ${MVPN_SOURCE_DIR}/websocket/websockethandler.h
```
bakulf, line 218: add here the pushmessage.cpp/h to fix the unit-test compilation issue

bakulf (approved):

tests/functional/testPushMessageHandlers.js
bakulf: yes please. Because it should simply work. I don't want that this "covers" other bugs. What do you think?

brizental: Oh, I think I misunderstood your initial question. There is a loop there because we are deleting more than one device, not because we attempt to delete the same device more than once. I chose a poor variable name, `thirdDevice` makes it sound like there are only three, but in fact there are 5 devices there (see the Guardian overrides on the beaforeEach block). I'll change that.

I'd still rather leave the loop there, so that we know deletion works no matter how many devices are in the list. WDYT?

macos/pkg/CMakeLists.txt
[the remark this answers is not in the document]
brizental: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4030

brizental:
Filed an issue about the Windows unit test failures, looks like a perma failure: https://mozilla-hub.atlassian.net/browse/VPN-2587

The functional tests failurs also seem to be unrelated. Is there already a ticket to track that @bakulf ? I did not find any, but just checking.

tests/functional/testPushMessageHandlers.js
bakulf: good!

bakulf:
Yes the test failures are unrelated. And they are connected to VPN-2586

brizental:
Reminder to my Monday self on why I have not merged this yet:

I have attempted re-running the functional tests, but they are still failing. Since the same is not happening on my [other PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4046) I think this might be this PR's fault, so I need to check that next week, before I merge.

brizental:
The one test failure here is realted to: https://mozilla-hub.atlassian.net/browse/VPN-2635.

PR #3993, opened 2022-07-09
An optimization related to saved addons

bakulf:
As we know, we don't load all the addons. But because of this, without this PR, at any refresh, we try to download the non-loaded ones. With this PR, we keep track of them, storing them on disk without loading them.


oskirby (approved):
A more accurate title might be "remove survey models in favor of addons" but I am delighted to see so much code being removed.

PR #3998, opened 2022-07-11
basti/maybeFixIOS

strseb:
Removing some whitespace should trigger the update of the toolchain task :) - As apprently changing the .env does not trigger that

### DIFF
--- a/taskcluster/ci/toolchain/qt.yml
+++ b/taskcluster/ci/toolchain/qt.yml
@@ -5,7 +5,7 @@
 job-defaults:
     worker:
         env:
-            QT_VERSION: "6.2.3"
+            QT_VERSION: "6.2.4"
             QT_MAJOR: "6.2"
         max-run-time: 14400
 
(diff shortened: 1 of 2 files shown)


taskcluster/scripts/toolchain/bundle_qt_ios.sh
```diff
@@ -5,8 +5,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 set -x
 
-
```
bakulf, line 8: this is a random change :)

taskcluster/ci/toolchain/qt.yml
```diff
@@ -5,7 +5,7 @@
 job-defaults:
     worker:
         env:
-            QT_VERSION: "6.2.3"
+            QT_VERSION: "6.2.4"
```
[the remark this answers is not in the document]
bakulf: We still need to land this, but it's not the real fix.

bakulf (approved):

taskcluster/scripts/toolchain/bundle_qt_ios.sh
strseb: its intentional :) - 

changeing the `env:` vom 2.3 to 2.4 will not invalidate the old toolchain task. It's only updated if anything it depends on (taskcluster-version, any task it depends on, or changes in bundle_qt_ios.sh ) changes. Therefore as we landed the patch initially, we did not get a new task that generates that. 

I'm currently working on expressing those types of things - see #3999

PR #4013, opened 2022-07-14
Populate glean app info 

strseb:
Sorry!
Sets  the glean data using the package manager info
```
Version Code: 1657817315
Version Name: 2.10.0(-debug)
```


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
```diff
@@ -34,7 +36,12 @@ class GleanUtil(aParent: Context) {
             "production"
         }
         val conf = Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, channel, 500, HttpURLConnectionUploader())
-        val build = BuildInfo("versionCode", "VersionName", Calendar.getInstance())
+        val info: PackageInfo = mParent.packageManager.getPackageInfo(mParent.packageName, 0)
+        val build = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
birdsarah, line 40: I don't understand why you need this if statement because I don't have enough familiarity with the code.

birdsarah (approved):
Looks good to the best of my limited abilities.

brizental (changes requested):
glean_parser generates the GleanBuildInfo object to be passed as the build info on initialize. See example on https://mozilla.github.io/glean/book/reference/general/initializing.html.

Would you mind switching to use the Glean generated build info object?

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
strseb: On newer android versions we need to use `longVersionCode` which is not present on android 7, so we fallback to  `versionCode`

strseb:
Is there a special flag i need to add to the build.gradle? It does not seem that glean generates that class :c - Could it be because we're building an aar instead of an apk at this stage?

brizental:
No flag needed as far as I know. In fact it seems you can use a flag to turn it off, but the default is on. 

With regards to 
> Could it be because we're building an aar instead of an apk at this stage?

I actually have no idea about this, however I don't see why that would have an impact. The metrics and pings modules are there and these should all have been during the same step... Hm. This is weird.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/GleanUtil.kt
badboy: fwiw glean_parser can generate this at build time (it uses `BuildConfig.VERSION_CODE` and `BuildConfig.VERSION_NAME`).

If you're using the Glean Gradle plugin it does that for you and you should be able to import it:

```
import org.mozilla.firefox.vpn.daemon.GleanMetrics.GleanBuildInfo
```	

and then pass that in as `buildInfo = GleanBuildInfo.buildInfo`

PR #4014, opened 2022-07-14
Revert "Properly disable qmljsdebugger in release mode"

strseb:
Reverts mozilla-mobile/mozilla-vpn-client#3539


andersk:
The real problem here is that #3539 was mismerged due to an undetected conflict with #3935. I have a proper fix in #4015.

PR #4022, opened 2022-07-16
Proxy implementation

bakulf:
This is a simple socks5 proxy implementation that can be activated when the VPN is on.

### DIFF
--- a/socks5proxy/src/socks5client.cpp
+++ b/socks5proxy/src/socks5client.cpp
@@ -1,0 +1,337 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "socks5client.h"
+#include "socks5.h"
+
+#ifdef Q_OS_WIN
+#  include <winsock2.h>
+#  include <ws2ipdef.h>
+#else
+#  include <arpa/inet.h>
+#endif
+
+#include <QHostAddress>
+
+namespace {
+
+#ifdef Q_OS_WIN
+#  define PACK(__Declaration__) \
+    __pragma(pack(push, 1)) __Declaration__ __pragma(pack(pop))
+#else
+#  define PACK(__Declaration__) __Declaration__ __attribute__((__packed__))
+#endif
+
+PACK(struct ClientGreetingPacket {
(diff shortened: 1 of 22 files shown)


socks5proxy/src/socks5client.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+        ClientConnectionAddressIpv6Packet packet;
+        if (m_inSocket->read((char*)&packet,
+                             sizeof(ClientConnectionAddressIpv6Packet)) !=
+            sizeof(ClientConnectionAddressIpv6Packet)) {
+          m_inSocket->close();
+          return;
+        }
+
+        struct sockaddr_in6 sa;
```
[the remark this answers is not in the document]
oskirby: There is also a weird MacOS/BSD quirk that might require settings `sin6_len = sizeof(struct sockaddr_in6)`

socks5proxy/src/socks5client.cpp
```diff
@@ -0,0 +1,337 @@
...
+});
+
+PACK(struct ClientConnectionAddressIpv6Packet {
+  uint8_t ip_dst[16];
+  uint16_t port;
+});
+
+PACK(struct ServerResponsePacket {
+  uint8_t m_version = 0x05;
+  uint8_t m_rep = 0x01;
```
oskirby, line 60: For cleanliness, I feel like we need to add an enumeration for the response codes instead of using hardcoded values and comments spread across this file.

socks5proxy/src/socks5client.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+qint64 Socks5Client::proxy(QTcpSocket* a, QTcpSocket* b) {
+  Q_ASSERT(a && b);
+
+  qint64 bytes = 0;
+  char buffer[4096];
+  while (a->bytesAvailable()) {
+    qint64 val =
+        a->read(buffer, qMin(a->bytesAvailable(), (qint64)sizeof(buffer)));
+    if (val <= 0 || b->write(buffer, val) != val) {
```
oskirby, line 279: This does leave open the possibility that too much incoming data can overwhelm the outbound socket when connecting to a slow (or unreachable?) host. In such a situation this could result in unexpected connection drops when the more correct behaviour is to throttle the reception of data on the incoming socket.

birdsarah:
Closing as stale.

PR #4015, opened 2022-07-14
Repair mismerge of #3539

andersk:
## Description

While #3539 was waiting to be merged, it accumulated an undetected merge conflict with commit 7369bb3dc5792059389295504b56d2e7917fb615 (#3935), which moved the `project()` directive to the top of the file. GitHub incorrectly resolved this by duplicating the `project()` directive.  Then commit 28cc02cdd9de5f1c00e258a3fe32f132d875bb31 (#4023) deleted one of the duplicates, but left the original fix ineffective, because the `set(CMAKE_BUILD_TYPE …)` directive is still incorrectly after the remaining `project()` directive.

Fix it.

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #4023, opened 2022-07-17
Remove a double project entry in CMakeLists.txt

bakulf:


flozia (approved):

andersk:
This was not the right resolution. I already gave you the right resolution in #4015, which I’ve now updated to apply on top of this. Can you **please just merge #4015 right now** and stop kicking this down the road to accumulate even more incorrectly resolved merge conflicts?

PR #4028, opened 2022-07-19
Reset profile flow on account deletion

flozia:
## Description

Reset the profile flow on account deletion.

## Reference

[VPN-2554: Profile screen cannot be accessed after a delete-account action](https://mozilla-hub.atlassian.net/browse/VPN-2554)

## Checklist
    
### DIFF
--- a/src/profileflow.cpp
+++ b/src/profileflow.cpp
@@ -50,7 +50,9 @@ void ProfileFlow::start() {
 
   setState(StateLoading);
 
-  User* user = MozillaVPN::instance()->user();
+  MozillaVPN* vpn = MozillaVPN::instance();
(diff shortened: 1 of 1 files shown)


src/profileflow.cpp
```diff
@@ -80,6 +82,7 @@ void ProfileFlow::start() {
       setState(StateError);
     }
   });
+  connect(vpn, &MozillaVPN::accountDeleted, this, [&] { reset(); });
```
bakulf, line 85: mm what about if we reset it when the VPN->State goes to != StateMain ?

bakulf (changes requested):

bakulf (approved):

PR #4030, opened 2022-07-19
BUGFIX: Fix macos cmake config file

brizental:
## Description

My experience with CMake is none, so I am not really sure what was going
on here. What I know is that I was getting the following error:

`TARGET_BUNDLE_DIR is allowed only for Bundle targets`

And with the change in this commit I don't get it anymore.

This was preventing me from building on MacOS.

Moved over from: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/3983

## Reference

n/a

## Checklist
    

README.md
```diff
@@ -209,7 +209,9 @@ Some variables that might be useful when configuring the project:
 cmake --build build
 ```
 
-This will produce the application bundle in `build/src/Mozilla VPN.app`
+This will produce the application bundle in `build/src/Mozilla VPN.app`.
+
+> **Note**: to open the application from the command line, run `open -n build/src/Mozilla\ VPN.app`.
```
brizental, line 214: Actually this is better:
```suggestion
> **Note**: to open the application from the command line, run `build/src/Mozilla\ VPN.app/Contents/MacOS/Mozilla\ VPN`.
```

macos/pkg/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ add_custom_command(
     DEPENDS mozillavpn
     COMMAND ${CMAKE_COMMAND} -E remove_directory -f ${CMAKE_CURRENT_BINARY_DIR}/staging
     COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/staging
-    COMMAND ${CMAKE_COMMAND} -E copy_directory $<TARGET_BUNDLE_DIR:mozillavpn>
```
oskirby, line 54: This change surprises me because I thought the `mozillavpn` target was a bundle (ie: we set the `MACOSX_BUNDLE` target). Could you confirm that we get the same directory structure built by the `pkg` target after making this change as we used to get with our old qmake builds after making this change?

oskirby (commented):
I think these changes are great, my only concern is that we might have changed the installer package by switching `TARGET_FILE_DIR` for `TARGET_BUNDLE_DIR`.

macos/pkg/CMakeLists.txt
brizental: Yes, I think it is the same directory structure. This is where pkg was located: `.../mozilla-vpn-client/build/macos/pkg/staging/MozillaVPN.pkg`, is that as expected?

I was unable to install though, here is the error I got:

Screenshot 2022-07-22 at 11 48 19

Do you think this is related? If not maybe I can go ahead and merge this one and file a bug to look into the other one.

oskirby: Yeah, I think there's a problem with this change to the installer package. I gave it a test and I found that the installer package only contains the VPN binary and is missing the rest of the bundled contents. After building the installer package, you should be able to take a look in `<builddir>/macos/pkg/staging/Applications/` and there should be a directory named `Mozilla VPN.app` which has the same directory structure as in `<builddir>/src/Mozilla VPN.app`

oskirby (changes requested):
We will need to fix the `copy_directory` step, or address that change in a separate PR. Thanks.

brizental:
>  We will need to fix the copy_directory step, or address that change in a separate PR. Thanks.

I dropped that commit. Somehow it's not giving me any errors anymore, so I guess all good.

oskirby (approved):

PR #4049, opened 2022-07-22
Wasm network via JS callback

bakulf:
OK This is a cool thing!
With this PR I have done the following:

1. remove the custom WASM components in our QML code
2. move the network handling for wasm in JS - reusing the same callback system as we have in the functional tests \o/
3. implement the "preset" concept for wasm

Current presets:
1. default - same as now in the wasm client
2. authentication in app
3. account creation in app
4. mobile onboarding

We can now have fun implementing all the remaining scenarios:
- subscription
- errors
- etc


strseb (approved):
Uuuhh this is nice!

PR #4060, opened 2022-07-23
Wasm errors

bakulf:
This is a big PR because it introduces:

- IAP for wasm - with all the possible error scenarios
- backend failure scenario in wasm
- better messaging handler in the wasm viewer
- more auth error scenarios (basically all..?)

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -920,6 +920,7 @@ void MozillaVPN::reset(bool forceInitialState) {
   logger.debug() << "Cleaning up all";
 
   TaskScheduler::deleteTasks();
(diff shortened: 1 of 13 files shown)


src/mozillavpn.cpp
```diff
@@ -920,6 +920,7 @@ void MozillaVPN::reset(bool forceInitialState) {
   logger.debug() << "Cleaning up all";
 
   TaskScheduler::deleteTasks();
+  m_private->m_closeEventHandler.removeAllStackViews();
```
bakulf, line 923: This is needed because after a reset, we can still have stack views. We need to remove them all to go to the main view.

strseb (approved):

PR #4072, opened 2022-07-25
Bottom Navbar UI

MattLichtenstein:
## Description

- Introduces a Bottom navigation bar to the UI

## Reference

[VPN-2487: Create custom navigation tab bar component](https://mozilla-hub.atlassian.net/browse/VPN-2487)

## Checklist
    

src/ui/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
```diff
@@ -119,17 +119,13 @@ VPNFlickable {
 
                 onClicked: {
                     VPN.recordGleanEvent("deleteAccountRequested");
+                    navbar.visible = false
```
MattLichtenstein, line 122: Even though it is a somewhat common pattern in this project, I don't love referencing ancestor components in other files like this since it's very fragile, still considering other implementations

src/ui/settings/ViewGiveFeedback.qml
```diff
@@ -375,7 +378,7 @@ Item {
                        anchors.top = undefined;
                        anchors.topMargin = undefined;
                        anchors.bottom= parent.bottom
-                       anchors.bottomMargin = VPNTheme.theme.windowMargin * 4
+                       anchors.bottomMargin = VPNTheme.theme.navBarHeightWithMargins
```
MattLichtenstein, line 381: For non-scrolling page views, we will have to remember to account for the navbar by anchoring the bottom margin of the layout to this value. Alternatively, we can create a new base component (maybe something like `SinglePageView`) that has this built in so we don't have to remember to add it for every new page view, and can be the root component of all of our pages. That's a decent amount of churn so I haven't done that yet, but would like to get some thoughts

translations/strings.yaml
```diff
@@ -677,6 +677,17 @@ tutorialPopup:
   tutorialLeaveSubtitle: Looks like you haven’t finished the tutorial quite yet. Would you like to continue and finish now or quit and come back later?
   tutorialLeavePrimaryButtonLabel: Resume tutorial
 
+navBar:
+  homeTab:
+    value: Home
+    comment: Label for the tab on the nav bar that takes the user to the home screen
```
flodolo, line 683: übernit: `for the tab on the nav bar` is incredibly hard to parse for me (because of the amount of short words), can we go with `for the tab on the navigation bar`?

flodolo (approved):

nebula/ui/components/VPNBottomNavigationBarButton.qml
```diff
@@ -0,0 +1,38 @@
...
+        radius: parent.height / 2
+        opacity: .2
+        color: VPNTheme.theme.white
+    }
+
+    Image {
+        id: image
+        anchors.centerIn: parent
+        sourceSize.height: 32
+        sourceSize.width: 32
```
flozia, line 36: Nit: Maybe we could use `VPNTheme.theme.iconSize * 2` here.

src/ui/settings/ViewGiveFeedback.qml
flozia: I like the idea and I’m also wondering if we rather should have all views in a `PageView` that can be potentially scrollable. If we feel like that is too much work for now that would be probably a good task to add to a design system epic.

nebula/ui/components/VPNBottomNavigationBar.qml
```diff
@@ -0,0 +1,55 @@
...
+        hoverEnabled: true
+    }
+
+    RowLayout {
+        id: layout
+        anchors.fill: parent
+        anchors.leftMargin: tabs.length === 3 ? 48 : 32
+        anchors.rightMargin: anchors.leftMargin
+
+        spacing: tabs.length === 3 ? 51.5 : 24
```
flozia, line 46: We’ll probably won’t have less than three tabs, but in the spirit of making the layout dynamic: `tabs.length <= 3`? Also: Could we have variables for those to make it a bit easier to grasp how the layout adjust to a changing amount of tabs?

nebula/ui/components/VPNServerList.qml
```diff
@@ -128,19 +125,31 @@ FocusScope {
...
             Repeater {
                 id: countriesRepeater
                 model: countriesModel
-                delegate: VPNServerCountry{}
+                delegate: VPNServerCountry{
+                    //Necessary because VPNClickableRow has hardcoded anchors (that we aren't using here because this is a layout)
+                    anchors.left: undefined
+                    anchors.right: undefined
+                    anchors.leftMargin: undefined
+                    anchors.rightMargin: undefined
```
flozia, line 144: Nit: For setting multiple `anchors` properties I really like using:
```suggestion
                    anchors {
                        left: undefined
                        right: undefined
                        leftMargin: undefined
                        rightMargin: undefined
                    }
```

flozia (commented):
The Navbar looks awesome. 🎉  Left some minor remarks and until it’s fully functional we should consider to hide it for now.

nebula/ui/components/VPNBottomNavigationBarButton.qml
```diff
@@ -0,0 +1,38 @@
+import QtQuick 2.0
```
MattLichtenstein, line 1: needs source code header

nebula/ui/components/VPNBottomNavigationBar.qml
```diff
@@ -0,0 +1,55 @@
+import QtQuick 2.0
```
MattLichtenstein, line 1: needs source code header

nebula/ui/components/VPNBottomNavigationBar.qml
```diff
@@ -0,0 +1,55 @@
...
+        id: layout
+        anchors.fill: parent
+        anchors.leftMargin: tabs.length === 3 ? 48 : 32
+        anchors.rightMargin: anchors.leftMargin
+
+        spacing: tabs.length === 3 ? 51.5 : 24
+        data: tabs
+
+        Component.onCompleted: {
+            for(var i = 0 ; i < tabs.length; i++) {
```
lesleyjanenorton, line 50: supernit: `let`

src/ui/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
lesleyjanenorton: You could also do this by adding an objectName to the parent Item {} in `ViewDeleteAccount.qml` and then adding something along the lines of `&& mainStackView.currentItem.objectName !== "viewDeleteAccountObjectName"` to toggle the nav bar visibility in main.qml.

src/ui/views/ViewContactUs.qml
```diff
@@ -80,8 +80,7 @@ Item {
             id: vpnFlickable
             property var appRating
             property var feedbackCategory
-            flickContentHeight: col.childrenRect.height
-            interactive: flickContentHeight > height
+            flickContentHeight: col.implicitHeight + col.anchors.topMargin
```
lesleyjanenorton, line 83: binding loop

MattLichtenstein:
> I like the idea and I’m also wondering if we rather should have all views in a PageView that can be potentially scrollable. If we feel like that is too much work for now that would be probably a good task to add to a design system epic.

@flozia That sounds like we should just make every page based on a VPNFlickable, regardless of whether or not we expect it to scroll. I am fine with this, maybe we just start using flickables for all pages going forward?

nebula/ui/components/VPNServerList.qml
MattLichtenstein: I have no preference either way - but I know this is whats recommended by Qt's own conventions. My vote is for consistency. If we can get buy-in from all UI engineering to do this going forward, I am for it. 

Maybe we should start a document on our own qml coding conventions?

src/ui/main.qml
```diff
@@ -121,6 +121,77 @@ Window {
...
+        Behavior on opacity {
+            PropertyAnimation {
+                duration: 500
+            }
+        }
+
+        Connections {
+            target: VPNConnectionBenchmark
+            onStateChanged: {
+                navbar.opacity = VPNConnectionBenchmark.state === VPNConnectionBenchmark.StateInitial ? 1 : 0
```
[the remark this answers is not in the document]
MattLichtenstein: its disabled on zero opacity, so not accessible to screen readers 😄

src/ui/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
MattLichtenstein: haha seems hacky, but probably better than all the rebinding

src/ui/main.qml
```diff
@@ -121,6 +121,77 @@ Window {
...
+            }
+        }
+
+        Connections {
+            target: VPNAuthInApp
+            onStateChanged: {
+                if(VPNAuthInApp.state === VPNAuthInApp.StateInitializing) {
+                    navbar.visible = Qt.binding(() => VPN.state === VPN.StateMain)
```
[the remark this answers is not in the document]
MattLichtenstein: the binding may be broken on line 122 in ViewSubscriptionManagement.qml, but will make the change you suggested [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4072#discussion_r933641617) and get rid of this

PR #4074, opened 2022-07-26
test

strseb:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/addons/kind.yml
+++ b/taskcluster/ci/addons/kind.yml
@@ -1,0 +1,30 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+jobs:
+  addons/any:
+      description: "Addons Build"
+      treeherder:
+          symbol: A
+          kind: build
+          tier: 1
+          platform: addons/all
+      worker-type: b-linux
+      worker:
+          docker-image: {in-tree: wasm}
+          max-run-time: 3600
+          artifacts:
+              - type: directory
+                name: public/build
+                path: /builds/worker/artifacts
+      run:
+          using: run-task
+          use-caches: true
+          cwd: '{checkout}'
+          command: ./taskcluster/scripts/addons/build.sh
--- a/taskcluster/scripts/addons/build.sh
+++ b/taskcluster/scripts/addons/build.sh
@@ -1,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+set -e
+
+# Fetch strings
+auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+git submodule sync --recursive
+git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+# Reqs
+pip3 install -r requirements.txt
+
+python3 scripts/addon/generate_all.py
+
+cp -r addons/generated/addons /builds/worker/artifacts 
+


taskcluster/ci/push-addon/kind.yml
```diff
@@ -0,0 +1,45 @@
...
+          max-run-time: 3600
+          artifacts:
+              - type: directory
+                name: public/build
+                path: /builds/worker/artifacts
+      run:
+          using: run-task
+          use-caches: true
+          cwd: '{checkout}'
+          command: ./taskcluster/scripts/push-addons/deploy.sh
```
ahal, line 45: Is the intent of this kind to push addons to archive.mozilla.org?

If so unfortunately you'll have to redo this task to use our "beetmover" worker (beetmover -> bit mover -> move bits to archive.mozilla.org.. don't ask). This is the standard and secure method we have in place for moving things over to archive.mozilla.org, so I think it'll be a requirement.

The good news is I've already requested and got access to the credentials we can use to move things under `pub/vpn/addons`, and these credentials are already in the right place. @gabrielBusta is looking into standing up some tasks that will be able to move the main client there. So if you block on his work, a lot of the scaffolding will be done for you already.

Alternatively, I can put you on the right path if you want to go ahead with it yourself.

taskcluster/ci/build/kind.yml
```diff
@@ -20,3 +20,4 @@ jobs-from:
     - android.yml
     - macos.yml
     - wasm.yml
+    - addons.yml
```
ahal, line 23: I might reconsider rolling this into the build kind. My gut feeling is down the road keeping it as a separate kind will end up being easier to maintain. But up to you!

taskcluster/ci/push-addon/kind.yml
```diff
@@ -0,0 +1,45 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - mozillavpn_taskgraph.transforms.build:transforms # So we can do "requires-level"
```
ahal, line 8: Might be worth pulling `requires-level` processing into a separate transform?

ahal: Note: even if it is in a separate transform, the build transforms could still import and use it implicitly.

taskcluster/ci/push-addon/kind.yml
strseb: Happy to dig into this, once we have the beetmover stuff setup :) 

I do think the "end goal" is exactly this. For now, as we are testing - we're fine with this task just unpacking the bundle and using the release-index work to get a permalink to each file. (I kind of to have this even without archive, so we can stop using gh-pages as our current "file" storage for vpn :D )

ahal: Cool, sounds good!

gabrielBusta: @strseb @ahal reviving this thread. We can now beetmove TC artifacts to `pub/vpn/addons`, is the `addons.zip` bundle what we want to push to the archive? And how should we trigger this beetmoving? a Taskcluster hook?

ahal: Yeah, I was thinking there'd be a "release promotion" action (https://mozilla-hub.atlassian.net/browse/RELENG-981). So to trigger a release, you'd go to the Task Group page in Taskcluster of the revision you want to promote, select a `Promote Add-on` action (or similar), enter in the name of the add-on to promote, and then submit.

strseb: > We can now beetmove TC artifacts to pub/vpn/addons
Nice! - However we still have a dependency to setup proper-release-sigining for this.

What we could do if we can push to`pub/vpn/candidates/addons`.
The current `push-addon`- already does test-sign the addons with a garbage-key. 
On any push to main - if we could get the output of that beetmoved to archive, ([unzipped, so we can address each file like here](https://firefox-ci-tc.services.mozilla.com/tasks/index/mozillavpn.v2.mozilla-vpn-client.branch.releases.2.10.latest.addons/stage), that would be super cool!

ahal: I'm happy to use whatever URL paths the VPN team likes, but I was thinking it would be something like:
```
pub/vpn/addons/{addon-name}/candidates/{version}/
pub/vpn/addons/{addon-name}/releases/{version}/
```

Since the addons don't need to be signed, there's not really any harm in beetmoving ahead of time IMO (at least to the candidates dir).

gabrielBusta: Trying out an [experiment in staging](https://github.com/mozilla-releng/staging-mozilla-vpn-client/tree/releases/gabriel/addons_beetmover) to beetmove the addons from the `push-addon` task.

[Running into this issue with the `addons-bundle` task](https://firefox-ci-tc.services.mozilla.com/tasks/K2d6XQEdRtCeVXKkVS9b7w/runs/0/logs/public/logs/live.log#L53) ❌ 

`bash: ./taskcluster/scripts/addons/build.sh: Permission denied`

I looked into it, not much luck, thought it might be something with the in-tree wasm image, but [everything in the working directory should have permissions to run](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/taskcluster/docker/wasm/Dockerfile#L15) 😞

gabrielBusta:
@strseb I ran into [the issue on that latest decision task run](https://firefox-ci-tc.services.mozilla.com/tasks/VsvSZIViQDOU3Q8Le40svw/runs/0/logs/public/logs/live.log) in staging. [This is how I fixed it](https://github.com/mozilla-releng/staging-mozilla-vpn-client/commit/0ae16496120d81430ebd03238aad41cf59e817e4)

taskcluster/ci/push-addon/kind.yml
```diff
@@ -0,0 +1,45 @@
...
+      requires-level: 3
+      add-index-routes: addons # Name of the index
+      scopes:
+         - 'secrets:get:project/mozillavpn/tokens'
+      dependencies:
+        build: "addons-bundle"
+      fetches:
+        build:
+            - artifact: addons.zip
+      worker:
```
gabrielBusta, line 34: Before landing please add `chain-of-trust: true` to `worker` ([ex](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/taskcluster/ci/repackage/kind.yml#L32)). The beetmover worker needs this task to generate a `chain-of-trust.json` artifact

taskcluster/ci/addons/kind.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  bundle:
+      description: "Addons Build"
+      treeherder:
+          symbol: A
+          kind: build
+          tier: 1
+          platform: addons/all
+      worker-type: b-linux
+      worker:
```
gabrielBusta, line 19: Should probably add `chain-of-trust: true` to `worker` here too

gabrielBusta (changes requested):

taskcluster/scripts/push-addons/deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+# Reqs
+pip3 install -r requirements.txt
+
+# Get Secrets for building
+echo "Fetching Tokens!"
+./taskcluster/scripts/get-secret.py -s project/mozillavpn/tokens -k addons_key -f addons_key.pem
+ls 
+cd /builds/worker/fetches/
+ls
+openssl dgst -sha256 -sign /builds/worker/checkouts/vcs/addons_key.pem -out addons/manifest.json.sign addons/manifest.json
```
gabrielBusta, line 19: This push-addon task might just turn into an addon signing task (and we index that one?)

gabrielBusta:
A couple of thoughts:

- Are we beetmoving just the compiled `rcc` files + `manifest.json` + `manifest.json.sig`? (i.e. not the `ts` / `qrc` files)
- Wrt to the addons path in the archive, after taking a closer look at `addons/generate_all.py` and `addons/build.py`, paths like `pub/vpn/candidates/addons/<moz_build_date>/manifest.json`, `pub/vpn/candidates/addons/latest/manifest.json`, and `pub/vpn/candidates/addons/latest/guide_01_how_to_vpn.rcc` make sense to me

gabrielBusta (approved):

strseb:
> Are we beetmoving just the compiled rcc files + manifest.json + manifest.json.sig? (not the ts files) 

Yes only the json/json.sig and rcc files. 

>Wrt to the addons path in the archive, after taking a closer look at addons/generate_all.py and addons/build.py, paths like pub/vpn/candidates/addons/<moz_build_date>/<file_name> and pub/vpn/candidates/addons/latest/<file_name> make sense to me

Those look good to me :) 

cc @bakulf

bakulf:
+1 for anything @strseb said in the previous comment.

PR #4085, opened 2022-07-27
Port wasm build to cmake

bakulf:
Wasm compiled using cmake


src/cmake/wasm.cmake
```diff
@@ -0,0 +1,30 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Linux platform source files
```
oskirby, line 5: This comment should be updated

src/CMakeLists.txt
```diff
@@ -34,13 +39,17 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
...
+    set(MVPN_PLATFORM_NAME "wasm")
 endif()
 
 add_compile_definitions("$<$<CONFIG:Debug>:MVPN_DEBUG>")
 add_compile_definitions("MVPN_$<UPPER_CASE:${MVPN_PLATFORM_NAME}>")
 include(cmake/${MVPN_PLATFORM_NAME}.cmake)
 
-add_subdirectory(crashreporter)
-target_link_libraries(mozillavpn PRIVATE crashreporter)
+if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten")
```
oskirby, line 50: Would `NOT CROSS_COMPILING` be a better test here? Not that it really matters until we support Android/iOS with CMake.

src/addonmanager.cpp
```diff
@@ -182,8 +182,6 @@ bool AddonManager::validateIndex(const QByteArray& index,
   for (const AddonData& addonData : addons) {
     if (!m_addons.contains(addonData.m_addonId) &&
         validateAndLoad(addonData.m_addonId, addonData.m_sha256)) {
-      Q_ASSERT(m_addons.contains(addonData.m_addonId));
```
oskirby, line 185: This feels unrelated, or has something changed with wasm and the addons?

oskirby (approved):

src/CMakeLists.txt
bakulf: Well I hope to have crashreport on mobile eventually.

src/addonmanager.cpp
bakulf: you are right. It's unrelated but with a debug build, the wasm client crashes when validating a blocked addon

src/CMakeLists.txt
oskirby: Don't we already get crash reporting through the App/Play stores?

taskcluster/scripts/build/wasm.sh
```diff
@@ -12,18 +12,28 @@ qmake --version
 git submodule init
 git submodule update
 
+wget https://github.com/Kitware/CMake/releases/download/v3.24.0-rc4/cmake-3.24.0-rc4-linux-x86_64.sh
+chmod +x cmake-3.24.0-rc4-linux-x86_64.sh
+mkdir /tmp/cmake
+./cmake-3.24.0-rc4-linux-x86_64.sh --skip-license --prefix=/tmp/cmake
```
strseb, line 18: NIT: I think this should go here :)
 https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/taskcluster/docker/wasm/Dockerfile

src/CMakeLists.txt
bakulf: not via crashpad

PR #4094, opened 2022-07-28
Filter and sortable model

bakulf:
Let's make the model sortable and filterable. VPN-2532

### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -10,7 +10,6 @@
 #include "closeeventhandler.h"
 #include "commandlineparser.h"
 #include "constants.h"
-#include "filterproxymodel.h"
 #include "fontloader.h"
 #include "iaphandler.h"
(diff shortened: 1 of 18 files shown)


src/commands/commandui.cpp
```diff
@@ -203,8 +203,8 @@ int CommandUI::run(QStringList& tokens) {
     }
 #endif
     // This object _must_ live longer than MozillaVPN to avoid shutdown crashes.
-    QmlEngineHolder* engineHolder = new QmlEngineHolder();
-    QQmlApplicationEngine* engine = QmlEngineHolder::instance()->engine();
+    QQmlApplicationEngine* engine = new QQmlApplicationEngine();
+    QmlEngineHolder engineHolder(engine);
```
strseb, line 207: This seems unrelated, is it?

strseb (approved):

PR #4104, opened 2022-07-29
Add signing tasks for Taskcluster windows builds

ahal:
This will add 3 tasks that run one after the other:

1. Signing internals
2. Repackage MSI
3. Signing MSI

Still a WIP

### DIFF
--- a/taskcluster/ci/repackage/kind.yml
+++ b/taskcluster/ci/repackage/kind.yml
@@ -1,0 +1,45 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - mozillavpn_taskgraph.transforms.release_artifacts:transforms
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - signing
+
+jobs:
+    msi:
+        description: repackage mozillavpn msi
+        run-on-tasks-for: [github-push, github-release]
+        fetches:
+            signing:
+                - artifact: unsigned.zip
+                  # Workers are using an old version of PowerShell that doesn'tcontain the fix in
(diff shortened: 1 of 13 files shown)


ahal:
Blocked on issue #4107

ahal:
Have this working on the staging repo now. I'll wait for https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4112 to land and rebase before requesting review here.

taskcluster/ci/repackage/kind.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+    msi:
+        description: repackage mozillavpn msi
+        run-on-tasks-for: [github-push, github-release]
+        fetches:
+            signing:
+                - artifact: unsigned.zip
+                  # Workers are using an old version of PowerShell that doesn'tcontain the fix in
+                  # https://github.com/PowerShell/Microsoft.PowerShell.Archive/issues/48, and the
+                  # `fetch-content` scrpit is not able to properly extract it. Until we can use a
```
hneiva, line 24: nit: s/scrpit/script

taskcluster/ci/signing/kind.yml
```diff
@@ -24,21 +24,24 @@ only-for-build-types:
...
+    description: sign mozillavpn
     run-on-tasks-for: []
     add-index-routes: build
     signing-format:
         by-build-type:
             android.*: autograph_apk
             linux.*: autograph_debsign
             macos.*: macapp
-            default: ""
+            windows/opt: autograph_authenticode
```
hneiva, line 38: Should we still have a default?

hneiva (approved):
Good stuff!

taskcluster/ci/signing/kind.yml
ahal: Nah I don't think so. If we ever add another platform to `only-for-build-types`, it would be invalid to have an empty signing-format anyway, so throwing an error (due to not being matched here) seems like a better outcome anyway.

ahal:
The `fetch-win-rustup` failure should be temporarily fixed in #4163

strseb (approved):

Issue #4107, opened 2022-07-29
Taskcluster Windows build missing files in `unsigned.zip` artifact


ahal:
The Taskcluster Windows build upload an `unsigned.zip` artifact which is then repackaged into an msi and released. While working on automated Windows signing, I noticed several files are missing, causing the MSI package task to fail.

Here's the contents of `unsigned.zip` from the latest [2.9.0 release build](https://firefox-ci-tc.services.mozilla.com/tasks/efV7lpSpTk2CyRrCfpek-Q):
```
addons
balrog.dll
libcrypto-1_1-x64.dll
libssl-1_1-x64.dll
Microsoft_VC142_CRT_x64.msm
mozillavpn.exe
mozillavpn.json
MozillaVPN.pdb
mozillavpnnp.exe
mullvad-split-tunnel.cat
mullvad-split-tunnel.inf
mullvad-split-tunnel.sys
tunnel.dll
WdfCoinstaller01011.dll
wintun.dll
```

And here are the contents of a [recent task](https://firefox-ci-tc.services.mozilla.com/tasks/TsHQHKnkR-S5BEzdSPllJA) from `main`:
```
logo.ico
Microsoft_CRT_x64.msm
Mozilla VPN.exe
mozillavpn.json
Mozilla VPN.pdb
mozillavpnnp.exe
mullvad-split-tunnel.cat
mullvad-split-tunnel.inf
mullvad-split-tunnel.sys
tunnel.dll
WdfCoinstaller01011.dll
wintun.dll
```

I also noticed that `MozillaVPN.{exe,pdb}` changed to `Mozilla VPN.{exe,pdb}`. If this was intentional we'll also need to update the `wxs` files.

Fixing this blocks both releasing and automated Windows signing so should likely be a high priority in case there's a need for a chemspill.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2633)

strseb:
```
addons
balrog.dll
libcrypto-1_1-x64.dll
libssl-1_1-x64.dll
```
Those are fine that they are missing. The .dll's are now static linked. The addons folder will be fetched on the fly from taskcluster. 
I've created #4112 to address the .wxs adjustments :)

Issue #4105, opened 2022-07-29
Mozilla VPN app cannot be installed on Windows10


data-sync-user:
**VPN version:**

* 2.10

**Platforms:**

* Windows 10;

**Prerequisites:**

* Download the latest build from [Treeherder](https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/EBqEms9sRWOY1vgmo54Gdg/runs/0/artifacts/public/build/MozillaVPN.msi) or Giyhub for Windows 10;

**Steps to reproduce:**

1. Try to install the app;
1. Observe the behavior;

**Expected results:**

* The app is installed with success.

**Actual results:**

* The app fails installation with a privilege error.

**Notes:**

* This was not reproducible with 2.9 builds;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2631)
┆author: Valentina Virlics

data-sync-user:
➤ Basti (Sebastian) Streich commented:

Valentina Virlics - Can you recheck with the newest build from taskcluster? 🙂

data-sync-user:
➤ Valentina Virlics commented:

Basti (Sebastian) Streich Sure!

Verified this while using the last 2.10.0 (2.202208050620) Windows build from Taskcluster and the VPN app installation is finalized with success. Yay!!

PR #4112, opened 2022-08-01
TC: Fix "windows build does not install"

strseb:
## Description

The issue with #4105: 

We need to set the build type during the build-system-gen, as we are using ninja. So we were creating a debug build, which meant the app was looking for the vc-redist-debug dll's (which we are not shipping in the msi ofc).  Therefore the install failed on non-dev systems.

So we're creating a release build now with forced debug info by cl.exe, as using cmake "relwithdebug" will create wierd golang <-> cl.exe linking issues. Not sure how to fix that. 

While at it touching #4107:
Removed the now missing files from the wxs, so the rebundle does not fail after signing :)


src/cmake/windows.cmake
```diff
@@ -110,13 +117,3 @@ install(FILES ${VC_TOOLS_REDIST_PATH}/MergeModules/Microsoft_VC${MSVC_TOOLSET_VE
     DESTINATION . RENAME Microsoft_CRT_x64.msm)
 
 install(FILES ui/resources/logo.ico DESTINATION .)
-
-## TODO: Are these still needed? It's not clear.
```
strseb, line 114: We know now, the build works fine without them :)

taskcluster/scripts/build/windows.ps1
```diff
@@ -53,16 +53,18 @@ $ErrorActionPreference = "Stop"
...
+cmake -S . -B $BUILD_DIR -GNinja -DCMAKE_BUILD_TYPE=Release # TODO: Linking breaks horribly with RelWithDebInfo
+cmake --build $BUILD_DIR #--config RelWithDebInfo Ignored as we are using ninja
+cmake --build $BUILD_DIR --target msi
 cmake --install $BUILD_DIR --prefix "$TASK_WORKDIR/unsigned"
 
 Write-Output "Writing Artifacts"
 New-Item -ItemType Directory -Path "$TASK_WORKDIR/artifacts" -Force
 $ARTIFACTS_PATH =resolve-path "$TASK_WORKDIR/artifacts"
 Copy-Item -Path $BUILD_DIR/windows/installer/MozillaVPN.msi -Destination $ARTIFACTS_PATH/MozillaVPN.msi
-Copy-Item -Path "$TASK_WORKDIR/unsigned/Mozilla VPN.pdb" -Destination "$ARTIFACTS_PATH/Mozilla VPN.pdb"
```
oskirby, line 65: Could you explain what's going on with this change and where `vc140.pdb` comes from?

src/cmake/windows.cmake
```diff
@@ -9,6 +9,13 @@ set_target_properties(mozillavpn PROPERTIES
     VERSION ${CMAKE_PROJECT_VERSION}
     WIN32_EXECUTABLE ON
 )
+# Todo: This will force the generation of a .pdb
```
oskirby, line 12: This should use a generator statement so that it's only takes effect when using the Release configuration, otherwise this may interfere with an attempt to create a real debug build.

IIRC that should look something like: `target_compile_options(mozillavpn PRIVATE $<$<CONFIG:Release>:/ZI>>`

taskcluster/scripts/build/windows.ps1
strseb: Building with `/ZI` will force the generation of the .pdb and vc140 is just the [default name](https://docs.microsoft.com/en-us/cpp/build/reference/fd-program-database-file-name?view=msvc-170) - tried to add flags to re-name it, but that failed so i did not investigate further.

oskirby (approved):

Issue #4090, opened 2022-07-28
[l10n] “Contact Support” screen has UI issues for Hungarian language


data-sync-user:
**Affected versions:**

* 2.10.0 (2.202207270720);

**Tested Platforms:**

* MacOS 11.6;

**Prerequisites:**

* Have a FxA account with an active subscription;

* Have VPN app opened and be signed in;

* Have app language set to Hungarian;

**Steps to reproduce:**

1. Go to Settings - Get Help menu;
1. Click on Contact support;
1. Observe the “Contact Support” title while hovering over the “<-” button;

**Expected result:**

* The screen title does not interfere with the navigation arrow;

**Actual result:**

* The screen title interferes with the navigation arrow;

**Notes:**

* Reproducible for the "VPN tips and tricks" screen as well, while having the language set to Icelandic;
* See the attached ss;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2621)
┆author: Bianca Hidecuti

data-sync-user:
➤ Santiago Andrigo commented:

Thanks Bianca Hidecuti Does this prevent the user from being able to click on the arrow to come back?

data-sync-user:
➤ Santiago Andrigo commented:

... and is the implication that this only affects Hungarian and Icelandic?

data-sync-user:
➤ Valentina Virlics commented:

Santiago Andrigo No, the functionality is not blocked, this is just UI.

Regarding the different languages implications, it is hard to say, because not all strings (at least when we talk about recent features) are translated in all languages. 

Here ( https://pontoon.mozilla.org/projects/mozilla-vpn-client/ ) you can see the progress.

data-sync-user:
➤ Santiago Andrigo commented:

Thanks!

data-sync-user:
➤ Santiago Andrigo commented:

Given the bug prioritization framework ( https://docs.google.com/document/d/1ryjue_UIYc_JZbONgj1Q1CXaha7pXtElEYoO814AHnw/edit# ) criteria, this goes down to ‘low’ priority. This is extremely low reach given the languages impacted, and it’s aesthetic, not functional.

PR #4117, opened 2022-08-01
2621/4090 - Fix menu UI breakages with lengthier menu titles

lesleyjanenorton:
## Description

This refactors VPNMenu.qml to handle longer menu titles in some locales.
Screen Shot 2022-08-01 at 11 53 23 AM
Screen Shot 2022-08-01 at 12 45 45 PM

## Reference
VPN-2621
#4090
## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNMenu.qml
+++ b/nebula/ui/components/VPNMenu.qml
@@ -16,7 +16,7 @@ Item {
     property bool accessibleIgnored: false
     property bool btnDisabled: false
     property alias forceFocus: iconButton.focus
-    property string _menIconButtonSource: "qrc:/nebula/resources/back.svg"
+    property string _menuIconButtonSource: "qrc:/nebula/resources/back.svg"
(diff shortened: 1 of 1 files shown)


nebula/ui/components/VPNMenu.qml
```diff
@@ -29,57 +29,76 @@ Item {
...
+            enabled: !btnDisabled
+            opacity: enabled ? 1 : .4
+
+            Image {
+                objectName: "menuIcon"
+                source: _iconButtonSource
+                sourceSize.width: VPNTheme.theme.iconSize
+                fillMode: Image.PreserveAspectFit
+                anchors.centerIn: iconButton
+            }
```
flozia, line 65: Suggestion: Definitely not part of this PR, but maybe we could file a ticket under tech debt — Looking at the `VPNIconButton` and places where it’s currently used I think it would be nice if `Image` could be moved into the component, so that we can just pass the `source` property.

flozia (approved):
Nice and works as expected!

Issue #3749, opened 2022-06-13
Support OpenSuSe


mcorino:
Description
I managed to build mozillavpn on OpenSuSE Leap 15.4 without errors as far as I can tell and it works just fine as long as I stick to the commandline commands. Installation went fine and the mozillavpn.service runs ok.
When I attempt to start the UI though the app crashes with an abort message.

Technical Information
Platform: OpenSuSE Leap 15.4 (with Qt 6.3.0 installed)
FX product: Mozilla VPN 2.8.4

Steps to Reproduce

Turn on the computer
Click on the VPN menu item to open the app or use command line 'ui' command to start it, both fail.

Actual Results
When starting from the commandline the app terminates with the message "Aborted (core dumped)".

Journalctl shows:

Jun 13 19:51:18 mco-desktop.m2c-software.nl kernel: Code: ab f4 ff 48 8b 04 24 48 85 c0 74 07 8b 10 83 fa 01 7e 12 31 c9 31 d2 31 f6 4c 89 e7 e8 6c de fb ff 48 8b 04 24 48 8b 54 24 08 <4c> 8b 22 48 85 c0 74 06 f0 83 28 01 74 25 4d 85 e4 74 >
Jun 13 19:51:18 mco-desktop.m2c-software.nl kernel: mozillavpn[10802]: segfault at 0 ip 00000000004e12bd sp 00007fff3d2aa850 error 4 in mozillavpn[400000+5a6000]
Jun 13 19:51:18 mco-desktop.m2c-software.nl mozillavpn[10768]: [13.06.2022 19:51:18.471] Debug: (linux - DBusService) Version request
Jun 13 19:51:18 mco-desktop.m2c-software.nl systemd[1546]: Started MozillaVPN.
Jun 13 19:51:17 mco-desktop.m2c-software.nl plasmashell[1757]: file:///usr/share/plasma/plasmoids/org.kde.plasma.kicker/contents/ui/RunnerResultsList.qml:50:9: Unable to assign [undefined] to int
Jun 13 19:51:17 mco-desktop.m2c-software.nl plasmashell[1757]: file:///usr/share/plasma/plasmoids/org.kde.plasma.kicker/contents/ui/RunnerResultsList.qml:50:9: Unable to assign [undefined] to int
[...]

PR #4121, opened 2022-08-01
Properly handle QML that failed to load

Cimbali:
## Description

Added early bail-out when qml fails to load in UI command, since the current callback and proper exit only get called too late (in `qApp.exec()`), i.e. after the error caused a segfault.

An alternate fix would be to delay all `connect()` calls that require the window (like the one below) until after the execution has started and the window has been confirmed to exist.

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3ae9aea3c262c0aa7a8d313976f80f38dcc957e5/src/systemtraynotificationhandler.cpp#L79-L80

## Reference

See #3749

## Checklist
    
   I was unable to find style guidelines in the README, in the wiki, in other top-level md files in the repo, on the support website…
      A test that the code does not segfault on failed QML loading would be nice I suppose, but I don’t know much about where or how to add it.

### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -553,16 +553,11 @@ int CommandUI::run(QStringList& tokens) {
 
     // Here is the main QML file.
     const QUrl url(QStringLiteral("qrc:/ui/main.qml"));
-    QObject::connect(
-        engine, &QQmlApplicationEngine::objectCreated, qApp,
-        [url](QObject* obj, const QUrl& objUrl) {
-          if (!obj && url == objUrl) {
-            logger.error() << "Failed to load " << objUrl.toString();
-            QGuiApplication::exit(-1);
-          }
-        },
-        Qt::QueuedConnection);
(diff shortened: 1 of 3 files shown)


bakulf:
Can you please tell me more about what you want to achieve here? Most of the cases, the app should be able to load the main QML file because it's a QRC resource, embedded in the app. Do you have a particular scenario?

Cimbali:
Yes @bakulf the scenario is reported in #3749.

When there is an error in loading the QML file (in this case because of a missing dependency), the `app.qExec()` is never reached because accessing the window (`QmlEngineHolder::instance()->window()` called from `NotificationHandler::create(…)`) causes a segfault.

The callback currently handling the loading error is never executed because of this segfault (it’s just queued to be run in the main loop), it’s dead code.

This PR fixes that by handling the error without callbacks and before calling `QmlEngineHolder::window()`.

Cimbali:
Closing in favour of #5237.

PR #4129, opened 2022-08-02
VPN-2646 - Remove websocket feature flag and leave it on by default

brizental:
## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2646

## Checklist
    
- [ ] ~I have commented my code PARTICULARLY in hard to understand areas~
- [x] I have ~added~ updated tests where needed

### DIFF
--- a/src/featureslist.h
+++ b/src/featureslist.h
@@ -261,4 +261,4 @@ FEATURE_SIMPLE(websocket,      // Feature ID
                true,           // Can be flipped on
                true,           // Can be flipped off
                QStringList(),  // feature dependencies
-               FeatureCallback_false)
+               FeatureCallback_true)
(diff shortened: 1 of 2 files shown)


src/featureslist.h
```diff
@@ -254,11 +254,3 @@ FEATURE_SIMPLE(unsecuredNetworkNotification,      // Feature ID
...
                QStringList(),                     // feature dependencies
                FeatureCallback_unsecuredNetworkNotification)
-
-FEATURE_SIMPLE(websocket,      // Feature ID
-               "WebSocket",    // Feature name
-               "2.9.0",        // released
-               true,           // Can be flipped on
-               true,           // Can be flipped off
-               QStringList(),  // feature dependencies
-               FeatureCallback_false)
```
bakulf, line 264: I think we can simply use `FeatureCallback_true` here. In this way, if we need to disable the feature we can do it from guardian.

bakulf (changes requested):

bakulf (approved):

PR #4132, opened 2022-08-03
Do not reset the journal public/private keys when something goes wrong

bakulf:
This fixes VPN-2643, probably just partially, but it's a good starting point.

### DIFF
--- a/src/tasks/adddevice/taskadddevice.cpp
+++ b/src/tasks/adddevice/taskadddevice.cpp
@@ -59,7 +59,6 @@ void TaskAddDevice::run() {
             MozillaVPN* vpn = MozillaVPN::instance();
             Q_ASSERT(vpn);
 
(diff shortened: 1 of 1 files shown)


src/constants.h
```diff
@@ -119,7 +119,7 @@ PRODBETAEXPR(const char*, addonPublicKeyFile,
              ":/addons_signature/production.der",
              ":/addons_signature/staging.der");
 
-PRODBETAEXPR(qint64, keyRegeneratorTimeSec, 604800, 300);
+PRODBETAEXPR(qint64, keyRegeneratorTimeSec, 604800, 30);
```
oskirby, line 122: Was this intentional? I still think that 30 seconds is too fast.

bakulf: no it was not. It was me testing the bug.

oskirby (approved):

PR #4140, opened 2022-08-03
Add MacOS beetmover tasks on the mozilla-vpn-client github-release taskgraph

gabrielBusta:
## Description

These code changes add tasks to upload VPN macos release candidates to https://ftp.mozilla.org/pub/vpn/candidates/ on `github-release` events.

I tried it out in staging. First I created a release with the tag `v0.0.16`. This triggered the release graph and uploaded release candidate [20220803200417](https://ftp.stage.mozaws.net/pub/vpn/candidates/0.0.16/)

To upload a second release candidate I deleted the release tag:

`git push --delete origin v0.0.16`

The release then turned back into a pre-release. I republished the release with the same tag. That uploaded release candidate [20220803202328](https://ftp.stage.mozaws.net/pub/vpn/candidates/0.0.16/)

[Here's is the latest release beetmover task I ran in staging](https://github.com/mozilla-releng/staging-mozilla-vpn-client/runs/7660304101)

## Reference

[RELENG-937](https://mozilla-hub.atlassian.net/browse/RELENG-937)

## Checklist
    
### DIFF
--- a/taskcluster/ci/beetmover/kind.yml
+++ b/taskcluster/ci/beetmover/kind.yml
@@ -21,11 +21,7 @@ primary-dependency: signing
 
 job-template:
     worker-type: beetmover
-    run-on-tasks-for: [action, github-release]
-    bucket:
-        by-level:
-            "3": "release"
-            default: "dep"
+    run-on-tasks-for: [action:release-promotion, github-release]
     treeherder:
         job-symbol: BM
         kind: build
--- a/taskcluster/ci/beetmover/kind.yml
+++ b/taskcluster/ci/beetmover/kind.yml
@@ -1,0 +1,35 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: mozillavpn_taskgraph.loader.multi_dep:loader
+
+transforms:
+    - mozillavpn_taskgraph.transforms.multi_dep:transforms
+    - mozillavpn_taskgraph.transforms.beetmover:transforms
+    - taskgraph.transforms.task:transforms
+
+group-by: build-type
+
+only-for-build-types:
+    - macos/opt
+
+kind-dependencies:
+    - signing
+
+primary-dependency: signing
+
+job-template:
+    worker-type: beetmover
+    run-on-tasks-for: [action, github-release]
+    bucket:
+        by-level:
+            "3": "release"
+            default: "dep"
+    treeherder:
+        job-symbol: BM
+        kind: build
+        tier: 1
+        platform:
+            by-build-type:
+                macos/opt: macos/all
--- a/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
@@ -39,7 +39,7 @@ def add_beetmover_worker_config(config, tasks):
         )
         destination_paths = [candidates_path]
         archive_url = "https://ftp.stage.mozaws.net/" if config.params["level"] != 3 else "https://ftp.mozilla.org/"
-        task_description = f"This {worker_type} task will upload a release candidate for v{app_version} of the {app_name} on {build_os} to {archive_url}{candidates_path}"
+        task_description = f"This {worker_type} task will upload a {build_os} release candidate for {app_name} v{app_version} to {archive_url}{candidates_path}/"
         branch = config.params["head_ref"]
 
         def get_artifact_path(path):
--- a/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
@@ -1,0 +1,104 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+import os.path
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import resolve_keyed_by
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def add_beetmover_worker_config(config, tasks):
+    for task in tasks:
+        worker_type = task["worker-type"]
+        task_name = task["name"]
+        task_label = f"{worker_type}-{task_name}"
+        resolve_keyed_by(
+            task,
+            "bucket",
+            item_name=task_label,
+            **{"level": config.params["level"]},
+        )
+        bucket = task["bucket"]
+        # When we add balrog releases we will need to add an action to beetmoverscript
+        action = "push-to-nightly"
+        app_name = "vpn"
+        attributes = task["attributes"]
+        run_on_tasks_for = task["run-on-tasks-for"]
+        build_id = config.params["moz_build_date"]
+        build_type = attributes["build-type"]
+        build_os = os.path.dirname(build_type)
+        dependencies = task["dependencies"]
+        app_version = config.params["version"]
+        candidates_path = os.path.join(
+            "pub", app_name, "candidates", app_version, build_id, build_os
+        )
+        destination_paths = [candidates_path]
+        archive_url = "https://ftp.stage.mozaws.net/" if config.params["level"] != 3 else "https://ftp.mozilla.org/"
+        task_description = f"This {worker_type} task will upload a release candidate for v{app_version} of the {app_name} on {build_os} to {archive_url}{candidates_path}"
+        branch = config.params["head_ref"]
+
+        def get_artifact_path(path):
+            # iscript turns .tar.gz files into signed .pkg files
+            if build_os == "macos":
+                return path.replace(".tar.gz", ".pkg")
+            return path
+
+        upstream_artifacts = [
+            {
+                **upstream_artifact,
+                "paths": [
+                    get_artifact_path(path) for path in upstream_artifact["paths"]
+                ],
+            }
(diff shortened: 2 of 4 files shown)


taskcluster/ci/beetmover/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+    - mozillavpn_taskgraph.transforms.beetmover:transforms
+    - taskgraph.transforms.task:transforms
+
+group-by: build-type
+
+only-for-build-types:
+    - macos/opt
+
+kind-dependencies:
+    - signing
```
ahal, line 18: This will get a little bit weird because on Windows the kind we will be beetmoving is `repackage-signing`.. but I guess we can deal with that later.

ahal (commented):
Nice! I'll take a closer look at this tomorrow.

taskcluster/ci/beetmover/kind.yml
gabrielBusta: Ah yeah, I dunno, maybe add another kind that depends on repackage-signing and put it thru the same beetmover transform? Seems kinda hacky

ahal: We might want to make it not use the `multi_dep` loader and just use the transform loader directly with each task explicitly listing a dependency.. but we can punt that to when we stand up the Windows task.

gabrielBusta: That sounds better

taskcluster/ci/beetmover/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+
+primary-dependency: signing
+
+job-template:
+    worker-type: beetmover
+    run-on-tasks-for: [action, github-release]
```
ahal, line 24: I'm not sure `action` is a valid tasks_for. Instead the action we'll use to trigger this will just use a `target_tasks_method` that explicitly selects this task. It won't even need to look at the `run-on-tasks-for` property.

taskcluster/ci/beetmover/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+    bucket:
+        by-level:
+            "3": "release"
+            default: "dep"
```
ahal, line 28: I'm not sure if setting this `by-level` makes sense. Only a `github-release` or action should be using the `release` bucket, so a regular push should use the `dep` bucket even though it is still level 3. It's not a big deal as long as we have the scopes configured correctly though. Maybe if someone manages to run a BM on a regular push, a scope error would be a good thing?

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,108 @@
...
+        task_label = f"{worker_type}-{task_name}"
+        resolve_keyed_by(
+            task,
+            "bucket",
+            item_name=task_label,
+            **{"level": config.params["level"]},
+        )
+        bucket = task["bucket"]
+        # When we add balrog releases we will need to add an action to beetmoverscript
+        action = "push-to-nightly"
```
ahal, line 28: In `beetmoverscript` it looks like `push-to-candidates` is an alias for `push-to-nightly`. I propose we use that string here as it matches closer to reality what is actually happening.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,108 @@
...
+        app_version = config.params["version"]
+        candidates_path = os.path.join(
+            "pub", app_name, "candidates", app_version, build_id, build_os
+        )
+        destination_paths = [candidates_path]
+        archive_url = (
+            "https://ftp.stage.mozaws.net/"
+            if config.params["level"] != 3
+            else "https://ftp.mozilla.org/"
+        )
```
ahal, line 45: Same comment as before around the `bucket`. Probably doesn't matter as long as scopes are correct.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,108 @@
...
+        task_description = f"This {worker_type} task will upload a {build_os} release candidate for {app_name} v{app_version} to {archive_url}{candidates_path}/"
+        branch = config.params["head_ref"]
+
+        def get_artifact_path(path):
+            # iscript turns .tar.gz files into signed .pkg files
+            if build_os == "macos":
+                return path.replace(".tar.gz", ".pkg")
```
ahal, line 52: I'm confused about this. If `iscript` converts it to a `.pkg` file, then shouldn't the artifact we're beetmoving already be in this format?

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,108 @@
...
+        # When we add balrog releases we will need to add an action to beetmoverscript
+        action = "push-to-nightly"
+        app_name = "vpn"
+        attributes = task["attributes"]
+        run_on_tasks_for = task["run-on-tasks-for"]
+        build_id = config.params["moz_build_date"]
+        build_type = attributes["build-type"]
+        build_os = os.path.dirname(build_type)
+        dependencies = task["dependencies"]
+        app_version = config.params["version"]
```
ahal, line 36: I suspect you may have copied this file, but for some of these that are only used in one place, why bother storing them into an intermediate variable?

ahal (changes requested):
Nice! Just have a few minor comments / suggestions / questions.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: This is just to display a useful link in the Github task description (on both the staging repo & here)

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: This is confusing. The artifact we are beetmoving is indeed a `.pkg` file. The upstream artifact to sign (from the build) is a `tar.gz` file, but the artifact produced by the iscript signing task (the one want to beetmove) is a `.pkg` file. I tried to see if I could get the artifacts from the upstream signing task from some attribute in the signing task, but I was not able to. Maybe the build should produce a `.pkg` file? Not sure... iscript seems to expect a `tar.gz` file.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: Some of that is copy paste. It makes the code more terse below I guess? I didn't bother making them not intermediate variables

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: Nice, didn't realize we had that alias 👍🏼

taskcluster/ci/beetmover/kind.yml
ahal: Right, I just mean should we add even more logic here? E.g, it's `release` if level 3 *and* `tasks_for in ["github-release", "action"]`. Would probably need to set it in a transform at that point unless you wanted nested `by-` statements.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: [I was thinking I could use the `release-artifacts` but that's also suffixed `tar.gz`](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/0fab5e0f5e00a6895313812148086dd4c0a1fb3f/taskcluster/ci/build/macos.yml#L24)

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@transforms.add
+def add_beetmover_worker_config(config, tasks):
+    for task in tasks:
+        app_name = "vpn"
+        worker_type = task["worker-type"]
+        task_name = task["name"]
+        task_label = f"{worker_type}-{task_name}"
+        l3_relpro = (
```
ahal, line 21: Is this supposed to be `is_relpro` ? The variable name is confusing to me.

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,99 @@
...
+            config.params["level"] == "3"
+            and config.params["tasks_for"] in task["run-on-tasks-for"]
+        )
+        bucket = "release" if l3_relpro else "dep"
+        build_id = config.params["moz_build_date"]
+        build_type = task["attributes"]["build-type"]
+        build_os = os.path.dirname(build_type)
+        app_version = config.params["version"]
+        candidates_path = os.path.join(
+            "pub", app_name, "candidates", app_version, build_id, build_os
```
ahal, line 31: So I guess we're going to implement the task to copy these over to the `releases` dir as a follow-up? We'll want to refactor this to re-use logic then, but happy to leave that for later.

ahal (approved):
Awesome, thanks!

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: Makes sense. I'll do it later

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
gabrielBusta: Is L3 relpro, but I see what u mean, will probably land a rename with the windows task

PR #4144, opened 2022-08-04
More countries localized

bakulf:
VPN-2654


flodolo (approved):
Should this be a GitHub action opening pull requests?

bakulf:
the localization of new countries/cities is still a manual operation. There is a script that generates the diff, but I still run it manually

PR #4146, opened 2022-08-04
Cmake mac tests

bakulf:


.github/workflows/test_lottie.yaml
```diff
@@ -104,11 +104,22 @@ jobs:
         run: |
           cargo install grcov --root ${{github.workspace}}/grcov-build --version 0.8.9
 
+      - name: Building tests
+        shell: bash
+        run: |
+          export PATH=/opt/qt6/bin:${{github.workspace}}/grcov-build/bin:$PATH
+          mkdir -p build
+          cmake -S lottie -B $(pwd)/build -DBUILD_TESTING=ON
+          cmake --build $(pwd)/build --config Release
```
oskirby, line 113: `--config Release` probably isn't doing anything here. CMake has two modes of operation depending on the type of make system being generated. There are multi-config generators like Visual Studio and XCode which need to set the build configuration at build-time (which is where the `--config` option gets used) and single config generators like GNU Make and Ninja where the build configuration is set at configuration time.

The default generator produces GNU Makefiles, in which case you would need to set the Release mode on the previous line with `-DCMAKE_BUILD_TYPE=Release`

oskirby (changes requested):
Now that all of the tests have been migrated to using CMake, we can get rid of the `QML tests` workflow, since the unit tests workflows actually run both the unit and QML tests together.

oskirby (approved):

oskirby:
Superseded by https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4821

PR #4151, opened 2022-08-04
VPN-2593 - Provide staging only flag to disable addon signature verification

brizental:
## Description

Opening this as a work in progress to get a bit of early feedback if possible as well as to make the branch available in case I need help debugging this. All builds here, the app runs nicely and signature verification is skipped if the feature is flipped off. However even with that I am unable to see changes made to local addons.

Steps I used for testing:

1. Changed the manifest file of one of the addons. I was changing the contents of the title of a guide.
2. Re-generated addons using `python ./scripts/addon/generate_all.py`
3. Changed the URL [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/constants.h#L112-L116) to point to the right folder in my local server. Note: Just setting the env var was not working.

With this I was able to see when `TaskAddonIndex` downloaded new stuff and the AddonManager updated itself. Still I saw no changes in the UI. I am not sure what is still missing to see that.

[UPDATE]

Somewhere there seems to be a _very_ persistent cache that I was unable to clean and that is why I was not seeing any of the updates. If I add a completely new guide I am able to see the new guide, but then subsequent changes to this new guide also do not show up afterwards. I think this is something that can be straightened out when writing up the docs for local addon testing (VPN-2596) so I will not consider it a blocking for opening this up for review and merging.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2593

## Checklist
    

src/addonindex.cpp
```diff
@@ -0,0 +1,225 @@
...
+    return;
+  }
+
+  QJsonObject indexObj;
+  if (!validate(index, indexSignature, &indexObj)) {
+    logger.debug() << "Unable to validate the index";
+    return;
+  }
+
+  // Question: Should we not emit the signal in case write is unsuccessfull?
```
brizental, line 134: @bakulf WDYT?

src/addondirectory.cpp
```diff
@@ -19,6 +19,8 @@ QString rootAppFolder() {
 #ifdef MVPN_WASM
   // https://wiki.qt.io/Qt_for_WebAssembly#Files_and_local_file_system_access
   return "/";
+#elif defined(UNIT_TEST)
+  return QStandardPaths::writableLocation(QStandardPaths::TempLocation);
```
brizental, line 23: I tried using [QStandardPaths::setTestModeEnabled](https://doc.qt.io/qt-6/qstandardpaths.html#setTestModeEnabled) in the tests/unit/helper.h module instead of doing this, but it did not seem to work.

src/signature.cpp
```diff
@@ -23,6 +23,10 @@ bool verify_rsa(const char* publicKey, size_t pubKeyLen, const char* message,
 // static
 bool Signature::verify(const QByteArray& publicKey, const QByteArray& content,
                        const QByteArray& signature) {
+  if (publicKey == nullptr || content == nullptr || signature == nullptr) {
+    return false;
+  }
```
brizental, line 28: Not too sure about the consequences of this change, I needed something that would make signature verification fail in testing mode. This seemed logical, so I went with it, but calling it out in case it has bad consequences.

brizental:
Addin @MattLichtenstein for awareness, @bakulf mentioned you were waiting for this feature :)

tests/unit/testaddonindex.cpp
```diff
@@ -69,7 +69,7 @@ void TestAddonIndex::update() {
 
   QSignalSpy indexUpdatedSpy(&ai, SIGNAL(indexUpdated(QList<AddonData>)));
 
-  ai.update(index, QByteArray());
+  ai.update(index, QByteArray("test signature"));
```
brizental, line 72: Needed to add _something_ here. `QByteArray()` evaluates to `nullptr`.

src/addondirectory.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <QObject>
+
+class QString;
+class QByteArray;
+class QDir;
+
+constexpr const char* ADDON_FOLDER = "addons";
+
+class AddonDirectory final : public QObject {
```
bakulf, line 16: this does not need to be a QObject.

src/addondirectory.h
```diff
@@ -0,0 +1,33 @@
...
+  Q_OBJECT
+
+ public:
+  AddonDirectory();
+
+  bool getDirectory(QDir* dir);
```
bakulf, line 22: All of these methods can be `const`. Like: `bool getDirectory(QDir* dir) const;`

src/addonindex.cpp
```diff
@@ -0,0 +1,250 @@
...
+ * id and sha256.
+ *
+ * The main function of this manager is to provide an API to update the index
+ * files, get the in memory index of addons, and whenever an update is
+ * completed successfully to emit a signal with the updated list of addons.
+ */
+AddonIndex::AddonIndex(AddonDirectory* addonDirectory) {
+  MVPN_COUNT_CTOR(AddonIndex);
+
+  m_addonDirectory = addonDirectory;
```
bakulf, line 37: `Q_ASSERT(addonDirectory);`

src/addonindex.cpp
bakulf: Yes, good point. we do something similar in case the validation fails.

src/addonindex.cpp
```diff
@@ -0,0 +1,250 @@
...
+    return;
+  }
+
+  // Question: Should we not emit the signal in case write is unsuccessfull?
+  write(index, indexSignature);
+
+  QList<AddonData> addons = extractAddonsFromIndex(indexObj);
+
+  if (!addons.isEmpty()) {
+    emit indexUpdated(addons);
```
bakulf, line 140: I wonder if we should emit `indexUpdated` also with an empty list of addons. I imagine this scenario:
something goes wrong, and we want to disable all the addons. We ship an empty manifest, and.. nothing happens because of this if-stmt.

src/addonindex.cpp
```diff
@@ -0,0 +1,250 @@
...
+  if (!Signature::verify(publicKey, index, indexSignature)) {
+    logger.warning() << "Unable to verify the signature of the addon index";
+    return false;
+  }
+
+  return true;
+}
+
+// static
+QList<AddonData> AddonIndex::extractAddonsFromIndex(QJsonObject indexObj) {
```
bakulf, line 240: const QJsonObject&

src/addonindex.h
```diff
@@ -0,0 +1,61 @@
...
+  QString m_addonId;
+  Addon* m_addon;
+};
+
+class AddonIndex final : public QObject {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(AddonIndex);
+
+ public:
+  AddonIndex(AddonDirectory* dir);
```
bakulf, line 35: explicit

src/addonindex.h
```diff
@@ -0,0 +1,61 @@
...
+
+  bool getOnDiskAddonsList(QList<AddonData>* addonsList);
+  void update(const QByteArray& index, const QByteArray& indexSignature);
+
+ signals:
+  void indexUpdated(QList<AddonData> addons);
```
bakulf, line 41: should this be: `const QList<AddonData>& addons)`

src/addonindex.h
```diff
@@ -0,0 +1,61 @@
...
+  // Helpers
+  static bool validateIndex(const QByteArray& index, QJsonObject* indexObj);
+  static bool validateSingleAddonIndex(const QJsonValue& addonValue);
+  static bool validateIndexSignature(const QByteArray& index,
+                                     const QByteArray& indexSignature);
+
+  static QList<AddonData> extractAddonsFromIndex(QJsonObject indexObj);
+
+ private:
+  AddonDirectory* m_addonDirectory;
```
bakulf, line 58: = nullptr;

src/addonmanager.cpp
```diff
@@ -3,15 +3,15 @@
...
-#include "constants.h"
 #include "leakdetector.h"
 #include "logger.h"
 #include "models/feature.h"
 #include "qmlengineholder.h"
-#include "signature.h"
 #include "taskscheduler.h"
 #include "tasks/addon/taskaddon.h"
 #include "tasks/function/taskfunction.h"
+#include "addondirectory.h"
```
bakulf, line 13: what about if we move all the addon* files into the `addons` folder? This one included.

bakulf: alphabetic order of the includes.

src/signature.cpp
bakulf: In theory, publicKey is never null because it's from a QRC file. Probably it could be in UNIT_TEST.
Same for content.

bakulf: if you are worried for the unit-test, we can add a `#if defined(UNIT_TEST) ... return true..`

src/tasks/addonindex/taskaddonindex.cpp
```diff
@@ -48,23 +52,27 @@ void TaskAddonIndex::run() {
...
               m_indexSignData = data;
+              m_indexSignatureRequestComplete = true;
               maybeComplete();
             });
   }
 }
 
 void TaskAddonIndex::maybeComplete() {
-  if (m_indexData.isEmpty() || m_indexSignData.isEmpty()) {
+  if (!m_indexSignatureRequestComplete || !m_indexRequestComplete) {
```
bakulf, line 72: I would like to suggest a different approach. If the signature verification is not needed, we could skip the second network-request.
And in here, we can check the feature again. If the signature verification is not requested, we can ignore the indexSignatureRequestComplete.

tests/unit/testaddonindex.cpp
bakulf: QByteArray() == nullptr is misleading, at least to me because it's not a pointer. Can we use QByteArray.isEmpty() instead?

src/signature.cpp
bakulf: Plus, I don't understand the type of comparison here. == nullptr seems to me that you want to check if a pubKey pointer is null. But this is a reference. Maybe we want to check this `isEmpty()` instead.

bakulf (changes requested):
This is such great improvement! I love this cleanup. Just small changes.

src/addonindex.cpp
brizental: Ha, that is a great catch actually. This is totally a bug in the implementation, the AddonManager even deletes the addons once they are not in the index anymore. I'll fix it.

src/tasks/addonindex/taskaddonindex.cpp
brizental: Yeah that was my first approach here, but I was worried about adding signature checks in places outside of the AddonIndex module and then if we ever change it, forget to change it here. That is something that should be mitigated by having a test that fails in case we forget to change it, now that I think about it... I'll change it.

src/addonindex.h
brizental: Yes indeed, I need to get better at the C++ annotations clearly. Thanks for catching all of these!

src/signature.cpp
brizental: > In theory, publicKey is never null because it's from a QRC file. Probably it could be in UNIT_TEST.

Maybe I can just check if signature is empty?

> if you are worried for the unit-test, we can add a #if defined(UNIT_TEST) ... return true..

That is what we have currently. Regardless of the arguments we return `true` in tests. However I needed it to return `false` in tests in case the signature was missing, otherwise my tests were failing. 

> Plus, I don't understand the type of comparison here. == nullptr seems to me that you want to check if a pubKey pointer is null. But this is a reference. Maybe we want to check this isEmpty() instead.

I'll try with isEmpty!

tests/unit/testaddonindex.cpp
brizental: The only reason why it was not misleading to me is because I don't have such a deep knowledge of C++ data structures yet :/ It makes sense now that you explain it, that it is weird... It is the behaviour that I noticed, I could be wrong though. I'll make the changes you suggested and see what happens.

src/tasks/addonindex/taskaddonindex.cpp
```diff
@@ -34,37 +35,47 @@ void TaskAddonIndex::run() {
...
   }
 
-  // Index file signature
-  {
-    NetworkRequest* request = NetworkRequest::createForGetUrl(
-        this, QString("%1manifest.json.sign").arg(Constants::addonSourceUrl()),
-        200);
+  if (Feature::get(Feature::Feature_addonSignature)->isSupported()) {
+    // Index file signature
+    {
```
bakulf, line 46: we can remove this '{' and the corresponding '}'. It's already in the if-stmt block.

bakulf (approved):

brizental:
Yes, there are test failures on the MacOS tests... tl;dr; If you run the tests with cmake yo uget no failures not even on the other AddonManager tests that were already failing even priot to this PR. So I will merge this with the failures and file a separate PR for the test fixes where I switch over to cmake.

PR #4153, opened 2022-08-04
VPN-2384 - Fix intermittent MacOS failure on websocket tests

brizental:
## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2384

### DIFF
--- a/src/websocket/websockethandler.cpp
+++ b/src/websocket/websockethandler.cpp
@@ -37,6 +37,14 @@ WebSocketHandler::WebSocketHandler() {
           this, &WebSocketHandler::open);
 }
 
+WebSocketHandler::~WebSocketHandler() {
+  MVPN_COUNT_DTOR(WebSocketHandler);
+
+  m_webSocket.disconnect();
+  m_backoffStrategy.disconnect();
+  m_pingTimer.disconnect();
+}
+
 // static
 QString WebSocketHandler::s_customWebSocketServerUrl = "";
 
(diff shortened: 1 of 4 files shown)


brizental:
Un-drafting this, even though the tests are still failing. Two reasons:

1. The MacOS tests are failing on a different test now (AddonManager tests, to be specific). The issue with them on the WebSocketHandler tests was addressed reliably here by using QTRY_* testing APIs instead of relying on random wait intervals before asserting on things. I filed a separate ticket to address the AddonManager test failures: https://mozilla-hub.atlassian.net/browse/VPN-2664?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel
2. The Windows unit tests are indeed failing on the WebSocket tests, however that is a different issue completely. It seems the intermittent backoff is not increasing on the Windows tests. I tested on a Windows machine and in normal usage it is increasing so this is not a fire. However, I would like to postpone the fixing of the Windows tests and merge the fix for the MacOS tests now. The Windows tests are a beast of their own because 1. I don't have a Windows machine to test on 2. I cannot see the failures be cause QTest output is not showing on CI. I filed a separate ticket to address the Windows QTest output issue: https://mozilla-hub.atlassian.net/browse/VPN-2686?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

strseb (approved):
out

PR #4158, opened 2022-08-05
Remove The wasm bin from the gh-pages branch

strseb:
This pr's goals is to sunset the gh-pages branch, which takes up quite a bit of space: 

-> A: Using an artifact instead of the old gh-pages branch https://github.com/actions/upload-pages-artifact 
-> B: Not storing the wasm-client in the gh-pages at all. As we have now have wasm builds in taskcluster, i wrote a small "chrome" that can load up the wasm for any branch. 
-> C: Splitting the gh pages and wasm workflows, as they now do completly diffrent things :) 
Untitled


.github/workflows/gh_pages.yaml
```diff
@@ -0,0 +1,146 @@
+name: GH Pages
```
bakulf, line 1: header

bakulf (approved):
2 questions:
- does it work for non-github users?
- the artifacts expire after 90 days, is it correct? I guess this is currently acceptable.

strseb:
> does it work for non-github users?

We're using the public api, without auth. That gets you per ip ~ 50 requests / h. No github account reqiured.
That's why i'm doing a bit of manual response cache-ing. :) 

>    the artifacts expire after 90 days, is it correct? I guess this is currently acceptable.

Level-3 (main, releases/*) -> 1 year
Level-1 (any other) -> 30days

PR #4161, opened 2022-08-08
ci: put Mac and Linux signing tasks on the same line as their builds …

ahal:
…in Treeherder


ahal:
Sorry for the double review, couldn't get CI to pass on the last one because I accidentally pushed it from my fork.

strseb (approved):

ahal:
Hm, I'm still getting the same failures, maybe the unittest task is just busted :/

PR #4169, opened 2022-08-08
Fix Taskcluster release indexes

ahal:
I accidentally regressed the index paths for non-Windows builds in my recent Windows signing PR. I also noticed that the Windows path is `msi` rather than `windows. This PR addresses both issues.

### DIFF
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -5,7 +5,6 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
-    - mozillavpn_taskgraph.transforms.release_index:transforms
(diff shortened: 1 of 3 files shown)


taskcluster/ci/build/kind.yml
```diff
@@ -5,7 +5,6 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
-    - mozillavpn_taskgraph.transforms.release_index:transforms
```
ahal, line 8: This was accidentally leftover and is no longer being used as no `build` tasks are setting the `add-index-route` key.

gabrielBusta (approved):

PR #4200, opened 2022-08-15
Remove a non-used feature for tips&tricks

bakulf:


MattLichtenstein (approved):
Ahhh, I forgot to remove this when we decided to not show it in the feature tour! Nice catch 😄

PR #4209, opened 2022-08-17
Navigation refactoring

bakulf:


translations/strings.yaml
```diff
@@ -697,3 +698,6 @@ global:
     value: Go back
     comment: Label for button used in intro modals to return to the app.
 
+getHelp:
+  helpCenter: Help Center
+  viewLogs: View Logs
```
flodolo, line 703: I assume this is a button or a link? Adding it to the ID or a comment would help, because English can be confusing.

flodolo (approved):
No specific issues with the export, only one comment for one of the new strings.

Issue #3655, opened 2022-06-01
VPN keeps losing connection after various hours of stable use


data-sync-user:
**Description**

The VPN loses connection after checking the uptime (basically opening the app to see how much time you've been successfully connected to the VPN). All data traffic stops even though the VPN looks like it is connected. The only way to get it working again is to open the VPN application. When they do this they see the uptime counter reset to 0 and the connection starts working. Happening on Celluar and WiFI.

**Technical Information**

Platform Android 11
FX product MozillaVPN v.2.8.2

**Steps to Reproduce**
1. Make sure the user is using Mozilla VPN v2.8.2
2. Open the VPN and toggle the button On
3. After a couple of minutes of hours connected, check back on the VPN app to see the total connected time

**Actual Results**

* The VPN app starts losing connection randomly
* It is not always reproducible by the user
* Has also happened while using a streaming service

**Expected Results**

* Mozilla VPN should open up normally and show the total connection time without losing the connection.

[^Android VPN logs.txt]

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2309)
┆author: Edwin Dubon

data-sync-user:
➤ Santiago Andrigo commented:

Hey Basti (Sebastian) Streich why is this market as blocked? Anything I can do to help?

data-sync-user:
➤ Santiago Andrigo commented:

Nevermind, now I understand on what’s its blocked 🙂

data-sync-user:
➤ Santiago Andrigo commented:

Hey Basti (Sebastian) Streich and Edwin Dubon 
I’m not sure I understand this issue: 

  * Are we saying this only happens after various hours of use (see the title) or even after a few minutes (see reproduction steps)?
  * (based on the description) What does it mean to ‘check the uptime’ without opening the app?
  * Is this happening in v2.9?

For what’s worth, I’m not seeing this problem in v2.9.

Basti (Sebastian) Streich did you see anything useful on the logs?
Unless somebody finds a thread to pull on this, it seems like it might not be actionable with the amount of information we have. Thoughts?

PR #4216, opened 2022-08-18
ConnectionHealth for Android Daemon

strseb:
Issue:
>The VPN loses connection after checking the uptime (basically opening the app to see how much time you've been successfully connected to the VPN). All data traffic stops even though the VPN looks like it is connected. The only way to get it working again is to open the VPN application. When they do this they see the uptime counter reset to 0 and the connection starts working. Happening on Celluar and WiFI.

Looking at the description - "When they do this they see the uptime counter reset to 0 and the connection starts working. " This clearly is ConnectionHealth, switching to the next server :) 

Of course that can't be used while the app is in the background, so this re-implements part's of this in the daemon. 
Essentially, it run's on a timer, checks the connection. The client, will provide a fallback server in the same city and we will silently switch to that in case the original server becomes unavailable :) 

Closes https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3655

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
@@ -1,128 +1,199 @@
 package org.mozilla.firefox.vpn.daemon
 
-import android.annotation.SuppressLint
 import android.content.Context
 import android.net.*
 import android.os.Build
 import android.os.CountDownTimer
-import android.os.Handler
-import androidx.annotation.RequiresApi
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
 
-
-@RequiresApi(Build.VERSION_CODES.R)
-class ConnectionHealth (service: VPNService): ConnectivityDiagnosticsManager.ConnectivityDiagnosticsCallback(){
-    private val mService: VPNService = service;
+class ConnectionHealth(service: VPNService) {
+    private val TAG = "DaemonConnectionHealth"
+    private val mService: VPNService = service
     private val PING_TIMEOUT = 3000 // ms
 
-    private var mEndPoint : String =""
-    private var mGateway : String =""
+    private var mEndPoint: String = ""
+    private var mDNS: String = ""
+    private var mGateway: String = ""
+    private var mAltEndpoint: String = ""
+    private var mResetUsed = false
 
-    var mActive = false;
-    var mVPNNetwork:Network? = null;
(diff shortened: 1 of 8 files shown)


bakulf (commented):
I prefer @oskirby to review this PR.

src/platforms/android/androidcontroller.cpp
```diff
@@ -181,6 +181,27 @@ void AndroidController::activate(const HopConnection& hop, const Device* device,
     excludedApps.append(QJsonValue(appID));
   }
 
+  // Find a Server as Fallback in the Same Location in case
```
oskirby, line 184: This is a pretty neat idea, and I like it better than the silent server switching we currently have. Although such a change is probably out of scope for the size of this PR, I would be interested to know if we could generalize the "fallback server" into something that's supported on all platforms.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -0,0 +1,199 @@
...
+                taskDone()
+                return@Runnable
+            }
+            // Step 4: Either the server is unreachable or the "reconnect" did not
+            // fix the issue. Try to switch to the fallback server.
+            val fallbackServerIsReachable = mConnectivityManager.allNetworks.firstOrNull() {
+                it.getByName(fallbackEndpoint).isReachable(PING_TIMEOUT)
+            } != null
+            if (fallbackServerIsReachable) {
+                Log.i("BASTI", "Switch to fallback VPN server")
```
oskirby, line 182: You should probably remove your name from the logs before this gets merged.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -0,0 +1,199 @@
...
+                mService.mainLooper.run {
+                    // Silent server switch to the same server
+                    mService.reconnect(true)
+                    NotificationUtil.instance?.show(mService)
+                }
+                mResetUsed = true
+                taskDone()
+                return@Runnable
+            }
+            // If we land here: The server and the fallback are unreachable
```
oskirby, line 193: If we get this far, can we emit something for the C++ code to handle? For example, we could display the server unavailable notification message or otherwise integrate with the UX via the `Controller::readyToServerUnavailable` signal.

oskirby (approved):
I think this looks good and we could probably merge it as-is, but I have added a couple ideas on how to make this a bit more awesome.

PR #4232, opened 2022-08-22
BUGFIX: Do not attempt to close websocket, if not connected

brizental:
~No ticket for this one, found it and fixed it immediatelly.~ Eh, ended up filing a ticket: https://mozilla-hub.atlassian.net/browse/VPN-2713

### DIFF
--- a/src/websocket/websockethandler.cpp
+++ b/src/websocket/websockethandler.cpp
@@ -163,6 +163,19 @@ void WebSocketHandler::onConnected() {
  * No-op in case the connection is already closed.
  */
 void WebSocketHandler::close() {
+  // If there is an attempt to close the websocket when it is not connected,
+  // QAbstractSocket may throw a write error before even attempting to close the
+  // underlying socket (see:
+  // https://code.woboq.org/qt5/qtwebsockets/src/websockets/qwebsocket_p.cpp.html#357).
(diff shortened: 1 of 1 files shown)


brizental:
I was not able to figure out how to write a test for this, unfortunately. Let me know if you can think of something @bakulf

brizental:
Ok I am marking this as a draft, because it requires a bit more digging. For starters, the premise of the fix is incorrect. Qt itself already checks if the socket is not connected and makes `close` a no-op in that case (https://code.woboq.org/qt5/qtbase/src/network/socket/qabstractsocket.cpp.html#_ZN15QAbstractSocket5closeEv). 

The actual issue here is that `close` may cause the underlying socket to error before it even attempts to close, by calling `write` (https://code.woboq.org/qt5/qtwebsockets/src/websockets/qwebsocket_p.cpp.html#357). `write` will throw an error when the websocket is not a in a writable state and _that_ is what causes the infinite loop. Anyways, this kind of looks like a Qt bug, but I am not sure I'll investigate it enough to file an issue. I'll see if I can just find a workaround here.

bakulf (approved):

brizental:
Well turns out the initial check was the proper workaroud. I just corrected the comment.

PR #4234, opened 2022-08-22
Measure and display latency in server view.

oskirby:
## Description
This adds storage of server latency data into the `Server` class, and adds a mechanism to refresh the server latency information whenever we update the server list, and then display that latency data to the user through a wifi with a variable number of bars.

## Reference
JIRA: [VPN-2526](https://mozilla-hub.atlassian.net/browse/VPN-2526)

## Checklist

### DIFF
--- a/nebula/ui/components/VPNServerCountry.qml
+++ b/nebula/ui/components/VPNServerCountry.qml
@@ -170,7 +170,7 @@ VPNClickableRow {
                 property string _countryCode: code
                 property string _localizedCityName: modelData[1]
                 property string locationScore: modelData[2]
-                property bool isAvailable: locationScore > 0
+                property bool isAvailable: locationScore >= 0
                 property int itemHeight: 54
 
                 id: del
@@ -216,13 +216,13 @@ VPNClickableRow {
                         top: del.top
                         verticalCenter: del.verticalCenter
                     }
-                    source: del.locationScore == 0 ? "qrc:/nebula/resources/warning.svg" :
-                        "qrc:/nebula/resources/wifi-" + (del.locationScore-1) + ".svg"
+                    source: del.locationScore < 0 ? "qrc:/nebula/resources/warning.svg" :
+                        "qrc:/nebula/resources/wifi-" + del.locationScore + ".svg"
                     sourceSize {
                         height: VPNTheme.theme.iconSizeSmall
                         width: VPNTheme.theme.iconSizeSmall
                     }
-                    visible: true
+                    visible: del.locationScore != 0
                 }
 
             }
--- a/nebula/ui/components/VPNServerCountry.qml
+++ b/nebula/ui/components/VPNServerCountry.qml
@@ -166,10 +166,10 @@ VPNClickableRow {
             model: cities
 
             delegate: VPNRadioDelegate {
-                property string _cityName: modelData[0]
+                property string _cityName: modelData.name
                 property string _countryCode: code
-                property string _localizedCityName: modelData[1]
-                property string locationScore: modelData[2]
+                property string _localizedCityName: modelData.localizedName
+                property string locationScore: VPNServerCountryModel.cityConnectionScore(code, modelData.code)
                 property bool isAvailable: locationScore >= 0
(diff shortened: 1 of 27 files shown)


src/serverlatency.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  connect(m_pingSender, SIGNAL(recvPing(quint16)), this,
+          SLOT(recvPing(quint16)));
+  connect(m_pingSender, SIGNAL(criticalPingError()), this,
+          SLOT(criticalPingError()));
+  connect(&m_timeout, &QTimer::timeout, this, &ServerLatency::stop);
+
+  // Send a ping to every server.
+  quint16 seq = 1;
+  for (const Server& server : scm->servers()) {
```
bakulf, line 48: I don't like the idea to ping all the servers at the same time. This means hundreds of ICMP packets and that can reach the rate limit on some OS.
I would suggest to use chunks of ICMP packets.

bakulf (changes requested):

src/serverlatency.cpp
oskirby: Done!

src/models/servercountrymodel.h
```diff
@@ -24,6 +24,13 @@ class ServerCountryModel final : public QAbstractListModel {
     CodeRole,
     CitiesRole,
   };
+  enum ServerConnectionScores {
+    Unavailable = -1,
+    NoData = 0,
+    Poor = 1,
+    Moderate = 2,
+    Good = 3,
+  };
```
bakulf, line 33: Q_ENUM ?

src/models/servercountrymodel.h
```diff
@@ -74,10 +82,14 @@ class ServerCountryModel final : public QAbstractListModel {
...
   QVariant data(const QModelIndex& index, int role) const override;
 
+ signals:
+  void changed();
+
  private:
   [[nodiscard]] bool fromJsonInternal(const QByteArray& data);
 
   void sortCountries();
+  int cityConnectionScore(const ServerCity& city) const;
```
bakulf, line 92: ServerConnectionScore?

src/models/servercountrymodel.cpp
```diff
@@ -156,22 +158,14 @@ QVariant ServerCountryModel::data(const QModelIndex& index, int role) const {
...
-        for (QString pubkey : city.servers()) {
-          if (m_servers.value(pubkey).cooldownTimeout() <= now) {
-            activeServerCount++;
-          }
-        }
-
         QStringList names{
             city.name(),
             ServerI18N::translateCityName(country.code(), city.name()),
-            QString::number(activeServerCount)};
```
bakulf, line 174: We should expose an object instead of this QStringList

bakulf (approved):

src/models/servercountrymodel.cpp
oskirby: I agree, this feels quite messy and it makes it quite error-prone to extend the city data.

src/models/servercountrymodel.h
oskirby: In this case we are actually scoring the city and not any one particular server at that location, so I feel that calling this a `cityConnectionScore` is slightly more accurate.

nebula/ui/components/VPNServerCountry.qml
```diff
@@ -216,12 +216,13 @@ VPNClickableRow {
...
                     }
-                    source: "qrc:/nebula/resources/warning.svg"
+                    source: del.locationScore == 0 ? "qrc:/nebula/resources/warning.svg" :
+                        "qrc:/nebula/resources/wifi-" + (del.locationScore-1) + ".svg"
                     sourceSize {
                         height: VPNTheme.theme.iconSizeSmall
                         width: VPNTheme.theme.iconSizeSmall
                     }
-                    visible: !del.isAvailable
+                    visible: true
```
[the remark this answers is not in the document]
oskirby: This prop is now used again when enabling and disabling the feature.

Issue #3982, opened 2022-07-09
Instrument DEVICE_DELETED push message handler


data-sync-user:
This is the first push message handler it will be useful for debugging and understanding of the new feature that is it instrumented so we can monitor it after release.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2563)
┆author: Beatriz Rizental Machado

data-sync-user:
➤ Beatriz Rizental Machado commented:

I feel like we should not only instrument the message handler itself, but also the general websocket message handler. Following are the questions I would like to answer with this data collection.

With regards to the WebSocketHandler:

  * How long does it take from “attempting connection” to “connected”, during testing I noticed this can sometimes take a very long time which would be concerning and might mean we want to queue up messages on the server.
  * How many times do we get connection errors and which errors are we getting.
  * How many times is the connection silently broken?
  * When is the connection closed and why.

With regards to the device deletion message:

  * When when deviceRemoved is called, is it being called by the push message or by the task delete device?
  * How many messages do we get?
    * From these, how many are malformed?
    * How many result in a reset?
    * How many result in device deletion?
    * How many result in… nothing?

  * How many unknown message types do we get?

Andrea Marchesini what do you think about these? Can you think of any other question that might come up when validating this feature?

PR #4275, opened 2022-08-26
VPN-2563 - Instrument websocket handler and device removal push message

brizental:
## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2563

## Checklist
    
- [ ] ~I have commented my code PARTICULARLY in hard to understand areas~
- [ ] ~I have added thorough tests where needed~
  - I would have loved to add tests for these. With the current structure that is not really straight forward so I skipped that for now.
  - I tested manually though, see https://debug-ping-preview.firebaseapp.com/pings/beavpntesting

### DIFF
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -1325,3 +1325,165 @@ sample:
         description: |
           The id of the step that was viewed
         type: string
+
+  # Start of push messages validation metrics
+  websocket_connection_attempted:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      Connection attempt has started.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+
+  websocket_connected:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      Connection was succesfull.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+
+  websocket_errored:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      There was an error with the connection.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+    extra_keys:
+      type:
+        description: |
+          The type of the error. One of these:
+          https://doc.qt.io/qt-6/qabstractsocket.html#SocketError-enum
+        type: quantity
+
+  websocket_pong_timed_out:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      Pong timed out. Probably the connection was silently broken.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+
+  websocket_close_attempted:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      Close attempt has started.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+
+  websocket_closed:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      Pong timed out. Probably the connection was silently broken.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+    extra_keys:
+      reason:
+        description: |
+          The reason why the websocket was closed.
+          https://doc.qt.io/qt-6/qwebsocketprotocol.html#CloseCode-enum
+        type: quantity
+
+  push_message_received:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      A device has been removed from the user's account.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+    extra_keys:
+      type:
+        description: |
+          The message type.
+          Either 'DEVICE_DELETED', 'TEST_MESSAGE' or 'unknown'.
+        type: string
+
+  device_removed:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      A device has been removed from the user's account.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - todo
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
+    extra_keys:
+      source:
+        description: |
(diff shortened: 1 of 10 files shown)


brizental:
@bakulf This is ready for review :) It's in draft mode just until I get the data-review r+.

travis79:
> DATA REVIEW REQUEST
> 
>     1. What questions will you answer with this data?
> 
> 
> With regards to the WebSocketHandler:
> 
>     * How long does it take from “attempting connection” to “connected”;
> 
>     * How many times do we get connection errors and which errors are we getting.
> 
>     * How many times is the connection silently broken?
> 
>     * When is the connection closed and why.
> 
> 
> With regards to the device deletion push message:
> 
>     * When deviceRemoved is called, is it being called by the push message or by the task delete device?
> 
>     * How many messages do we get?
>       
>       * From these, how many are malformed?
>       * How many unknown message types do we get?
> 
> 
>     2. Why does Mozilla need to answer these questions? Are there benefits for users?
>        Do we need this information to address product or business requirements?
> 
> 
> In order to monitor/validate the newly implemented push messages on the VPN client app.
> 
>     3. What alternative methods did you consider to answer these questions?
>        Why were they not sufficient?
> 
> 
> There is server instrumentation on the push messages. However that only tells half the story and does not inform us about this feature implementation on the client.
> 
>     4. Can current instrumentation answer these questions?
> 
> 
> Nope. See answer to 3.
> 
>     5. List all proposed measurements and indicate the category of data collection for each
>        measurement, using the Firefox data collection categories found on the Mozilla wiki.
> 
> 
> Measurement Name 	Measurement Description 	Data Collection Category 	Tracking Bug
> `sample.websocket_connection_attempted` 	Connection attempt has started. 	technical 	#3982
> `sample.websocket_connected` 	Connection was succesfull. 	technical 	#3982
> `sample.websocket_errored` 	There was an error with the connection. 	technical 	#3982
> `sample.websocket_pong_timed_out` 	Pong timed out. Probably the connection was silently broken. 	technical 	#3982
> `sample.websocket_close_attempted` 	Close attempt has started. 	technical 	#3982
> `sample.websocket_closed` 	Pong timed out. Probably the connection was silently broken. 	technical 	#3982
> `sample.push_message_received` 	A device has been removed from the user's account. 	technical 	#3982
> `sample.device_removed` 	A device has been removed from the user's account. 	technical 	#3982
> 
>     6. Please provide a link to the documentation for this data collection which
>        describes the ultimate data set in a public, complete, and accurate way.
> 
> 
> This collection is Glean so is documented [in the Glean Dictionary](https://dictionary.telemetry.mozilla.org).
> 
>     7. How long will this data be collected?
> 
> 
> This collection has expiry '2022-12-31'. The primary purpose of this data collection is to validate the performance of the push messages implementation. I am not sure all of it will be necessary indefinitely, probably only a subset of it. So for now I'll have this short expiry date and likely a part of these metrics will be renewed while the other goes away.
> 
>     8. What populations will you measure?
> 
> 
> All channels, countries, and locales. No filters.
> 
>     9. If this data collection is default on, what is the opt-out mechanism for users?
> 
> 
> These collections are Glean. The opt-out can be found in the product's preferences.
> 
>     10. Please provide a general description of how you will analyze this data.
> 
> 
> Through a periodically monitored dashboard.
> 
>     11. Where do you intend to share the results of your analysis?
> 
> 
> Mozilla VPN team.
> 
>     12. Is there a third-party tool (i.e. not Telemetry) that you
>         are proposing to use for this data collection?
> 
> 
> No.

# Data Review

1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way?

Yes, through the metrics.yaml file and the [Glean Dictionary](https://dictionary.telemetry.mozilla.org/).

2) Is there a control mechanism that allows the user to turn the data collection on and off?

Yes, through the application preferences

3) If the request is for permanent data collection, is there someone who will monitor the data over time?

N/A, collection to end 2022-12-31

4) Using the **[category system of data types](https://wiki.mozilla.org/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Category 1, Technical data

5) Is the data collection request for default-on or default-off?

Default-on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?

No

7) Is the data collection covered by the existing Firefox privacy notice?

Yes

8) Does the data collection use a third-party collection tool?

No

## Result

data-review+

travis79 (approved):
Please don't forget to update the metrics.yaml with the data-review link!

data-review+

glean/metrics.yaml
```diff
@@ -1325,3 +1325,165 @@ sample:
...
+      There was an error with the connection.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4275#issuecomment-1228532390
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
```
bakulf, line 1379: maybe this should not expire. What do you think?

glean/metrics.yaml
```diff
@@ -1325,3 +1325,165 @@ sample:
...
+      A device has been removed from the user's account.
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3982
+    data_reviews:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4275#issuecomment-1228532390
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - brizental@mozilla.com
+    expires: 2022-12-31
```
bakulf, line 1482: same here.

src/websocket/pushmessage.cpp
```diff
@@ -76,15 +77,27 @@ void PushMessage::parseMessage(const QString& message) {
...
-
   logger.debug() << "Unknown message type" << str;
   return MessageType_UnknownMessage;
 }
 
+// static
+const QString PushMessage::messageTypeToString(MessageType& type) {
+  switch (type) {
+    case MessageType_DeviceDeleted:
+      return "DEVICE_DELETED";
```
bakulf, line 91: what about if we use QVariant to do this? If MessageType is a Q_ENUM, we can do: ` QVariant::fromValue(type).toString()`

src/websocket/pushmessage.h
```diff
@@ -20,18 +20,19 @@ class PushMessage final : public QObject {
...
     MessageType_DeviceDeleted,
-
 #ifdef UNIT_TEST
     MessageType_TestMessage,
 #endif
-
     MessageType_UnknownMessage
   };
   static MessageType messageTypeFromString(const QString& str);
+  const QString type() { return messageTypeToString(m_messageType); }
```
bakulf, line 29: probably this is not needed.

src/websocket/websockethandler.cpp
```diff
@@ -270,5 +293,8 @@ void WebSocketHandler::onMessageReceived(const QString& message) {
   logger.debug() << "Message received:" << message;
 
   PushMessage parsedMessage(message);
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+      GleanSample::pushMessageReceived, {{"type", parsedMessage.type()}});
```
bakulf, line 297: QVariant::fromValue(type).toString()

bakulf (changes requested):

src/websocket/websockethandler.cpp
brizental: Yeah, so this is what I had initially and it was reporting empty strings. That is why I implemented that whole method to make the enum into a string. Maybe I was doing it wrong though, that is quite plausible. I'll try again.

glean/metrics.yaml
brizental: I added the expiry because I think these metrics don't need to be here indefinitely, maybe a subset of them, but not all. I will get an email when they are about to expire and by then I should know which are the keepers and then I can go ahead and change this. WDYT?

src/websocket/pushmessage.cpp
brizental: Ah I think that is the issue, MessateType is not yet a Q_ENUM!

src/tasks/removedevice/taskremovedevice.cpp
```diff
@@ -49,6 +50,9 @@ void TaskRemoveDevice::run() {
   connect(request, &NetworkRequest::requestCompleted, this,
           [this](const QByteArray&) {
             logger.debug() << "Device removed";
+
+            emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
```
bakulf, line 54: I would move this into the VPN.deviceRemoved to be sure we don't forget similar removal calls.

We can pass an enum as extra param to the identify the source of the call.

src/websocket/pushmessage.cpp
```diff
@@ -145,6 +144,8 @@ bool PushMessage::handleDeviceDeleted(const QJsonObject& payload) {
     return true;
   }
 
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
```
bakulf, line 147: Here as well.

bakulf (changes requested):

bakulf (approved):

PR #4296, opened 2022-08-29
Fix the macos Unit test - addon signature

bakulf:


brizental (approved):

src/models/feature.h
```diff
@@ -160,6 +160,10 @@ class Feature : public QObject {
   bool m_released = false;
   // Is true if the Feature was released in this Version
   bool m_new = false;
+
+#ifdef UNIT_TEST
+  friend class TestAddonIndex;
+#endif
```
brizental, line 166: Question: why is this necessary?

bakulf: because maybeFlipOnOff is private.

PR #4297, opened 2022-08-29
Fix tips and tricks layout with no addons

MattLichtenstein:
## Description

Fix the layout of the tips and tricks "all" tab when no addons are active

## Reference

[VPN-2753: Tips & Tricks feature has UI issues for a inexistent add-on URL](https://mozilla-hub.atlassian.net/browse/VPN-2753)

Screen Shot 2022-08-29 at 11 09 01 AM

## Checklist
    

brizental (approved):
Nice! Thanks for taking this!

src/ui/screens/settings/ViewTipsAndTricks.qml
```diff
@@ -155,9 +156,10 @@ VPNViewBase {
                 ColumnLayout {
                     id: layoutGuide
 
-                    anchors.fill: parent
```
brizental, line 158: So we remove this because if the Layout doesn't require the whole window height with this it will try to stretch and then margins will get weird, correct?

MattLichtenstein: Pretty much. Inside a scroll view/flickable you should never really need to fill the parent. It can cause problems 1.) for the reason you mentioned (things align to the middle of the layout and give you weird margins/spacing) 2.) it doesn't really make sense, since the scroll view only fills the screen, but the columnlayout is going to be longer than the screen (it's more analogous to the flickables contentHeight)

PR #4301, opened 2022-08-30
🏗️ Prototype: Android + Cmake 

strseb:
## Description
Switch build pipeline from:
```
Script.->Qmake->make->qmlimportscanner->gradle
```
We will build the client with cmake. Cmake will be invoked by gradle. 
This allow's us to build the client with one command (theoreticly) ```gradlew assemble```, and gives us the ability to define cmake/ndk versions as a classic gradle dependency.

Handling can stay the same as the package.sh now will just set gradle up, instead of qmake :)


android/buildSrc/src/main/kotlin/Config.kt
```diff
@@ -7,5 +7,5 @@ object Config {
     const val buildToolsVersion = "30.0.3"
     const val minSdkVersion = 24
     const val targetSdkVersion = 30
-    const val ndkVersion = "23.1.7779620"
+    const val ndkVersion = "24.0.8215888"
```
bakulf, line 10: do we need to update the documentation in the README file or elsewhere?

android/buildSrc/src/main/kotlin/QtConfig.kt
```diff
@@ -0,0 +1,226 @@
+import org.gradle.api.Plugin
```
bakulf, line 1: headers

android/buildSrc/src/main/kotlin/VPNVersion.kt
```diff
@@ -0,0 +1,43 @@
+import org.gradle.api.Plugin
```
bakulf, line 1: headers

bakulf (approved):
Can you remove `qmake/platforms/android.pri` and add something like this:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/src.pro#L44

PR #4303, opened 2022-08-31
Add JS conditions for addons

bakulf:
The idea here is to have conditions implementable in JS. As you can see from the unit-tests, they work fine.
I also introduce the AddonAPI object. We can extend this for the ComposerBlockButton with JS callbacks.


src/addons/addon.cpp
```diff
@@ -209,41 +210,51 @@ QList<ConditionCallback> s_conditionCallbacks{
...
+     }},
+
+    {"javascript",
+     [](const QJsonValue&) -> bool {
+       // dynamic condition
+       return true;
+     },
+     [](Addon* addon, const QJsonValue& value) -> AddonConditionWatcher* {
+       return AddonConditionWatcherJavascript::maybeCreate(addon,
+                                                           value.toString());
```
brizental, line 257: Question: can't we substitute almost all of the current condition watchers with JS condition watchers?

brizental (approved):

src/addons/addon.cpp
bakulf: in theory yes, but it would make all more complex to follow and implement.

PR #4305, opened 2022-08-31
Disable timers for add-device requests

bakulf:
This can be a good fix for the key-rotation


oskirby (approved):
I am still concerned that we might just be pushing the problem of ever-increasing response times for this API onto Guardian, but this should mitigate the issue on the client end at least.

PR #4313, opened 2022-09-01
Button - composer block

bakulf:
With JS and unit-tests


src/composer/composerblockbutton.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  QJSValue function =
+      QmlEngineHolder::instance()->engine()->evaluate(file.readAll());
+  if (function.isError()) {
+    logger.debug() << "Execution throws an error:" << function.toString();
+    return nullptr;
+  }
+
+  if (!function.isCallable()) {
+    logger.debug() << "The button js script should expose a callbale function"
```
brizental, line 56:
```suggestion
    logger.debug() << "The button js script should expose a callable function"
```

brizental (approved):
Awesome that you also added tests for the other composer elements. LGTM :D

PR #4316, opened 2022-09-01
No TS files for non-translatable add-ons

bakulf:


flodolo (approved):
Looks good, thanks for the quick fix
https://github.com/mozilla-mobile/mozilla-vpn-client/runs/8134907645?check_suite_focus=true

PR #4319, opened 2022-09-01
VPN-2787 - Update design on Tips & Tricks view

brizental:


translations/strings.yaml
```diff
@@ -595,9 +595,13 @@ tipsAndTricks:
   featureTourCardTitle: Tour the VPN
   featureTourCardDescription: Take a tour of Mozilla VPN to learn how the latest features help you protect your online privacy.
   quickTipsTitle: Quick tips
-  quickTipsDescription: Let us help you get the most out of Mozilla VPN
+  quickTipsDescription: Let us help you get the most out of Mozilla VPN.
```
flodolo, line 598: You need new string IDs for this type of change, to make sure every locale is updated.

Long explanation available here: https://developer.mozilla.org/en-US/docs/Making_String_Changes

brizental: Interesting, even if this hasn't made it into a release yet? BTW, the link is a 404.

flodolo: Also: https://transvision.flod.org/string/?entity=vpn_client/mozillavpn.xliff:f8f24fcadf191e23cb6e7479c3f14b61&repo=vpn_client

nebula/ui/components/VPNTipsAndTricksSection.qml
```diff
@@ -0,0 +1,67 @@
...
+
+ColumnLayout {
+    id: layout
+
+    property string title
+    property string description
+
+    property string type
+    property var customFilter
+    property var featureTourCardVisible
```
MattLichtenstein, line 21: `featureTourCardVisible` either needs to bound to a default value or always be defined in your `ObjectModel` in `ViewTipsAndTricks`

also, lets change `var` to `bool`

src/ui/screens/tipsAndTricks/ViewTipsAndTricks.qml
```diff
@@ -40,21 +40,42 @@ VPNViewBase {
...
+            Item {
+                property string title: VPNl18n.TipsAndTricksMoreTutorialsTitle
+                property string description: VPNl18n.TipsAndTricksMoreTutorialsDescription
+                property var filter: (addon) => !addon.highlighted
+            }
+        }
+
         stackContent: [
             // All
             VPNFlickable {
```
MattLichtenstein, line 81: Instead of flickables, we can use `VPNViewBase` which is essentially a flickable that factors in the nav bar so we don't have to set a weird flickContentHeight

nebula/ui/components/VPNTipsAndTricksSection.qml
brizental: It is bound to a default value [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4319/files#diff-bfa251922231da58128bc4d90863ab50cae6fd6c0277d75b6159d70133b1d819R16) does it still need a default value in this view or is that one enough?

MattLichtenstein: I believe you are overwriting that binding when you [pass in your options here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4319/files#diff-b037a6fe4a4973ce595c2734e31689591ae4d44b49f47b2f55e4fcefe220fee7R62)

brizental: Hm, I thought QML would work like JavaScript. If the value is `undefined` it will go to the defaut. I'll fix it.

brizental: I changed that line to cource an `undefined` into a `false`.

MattLichtenstein (approved):

flodolo (approved):

PR #4327, opened 2022-09-02
VPN-2517 - Notify user when a new in-app message is received or enabled

brizental:
## Description

I have tested this on both MacOS and Linux. 
- On **MacOS** nothing happened :-1: . 
- On **Linux** I got the notification :+1: . 
- On **iOS** I could not figure out how to test this :shrug:. 
- On **Android**... Well I have never been able to fully build for Android :shrug:. 
- I do not have access to Windows yet :shrug: .

I am pretty sure this does not work as described on the ticket for mobile devices though. All of the notification code is on the client and when the app is closed / on background that is just not running.

I have filed tickets to look at these issues, the scope of that works is likely too large to take on for 2.10 though.

- [VPN-2804 - MacOS desktop notification seem not to be working](https://mozilla-hub.atlassian.net/browse/VPN-2804)
- [VPN-2803 - Provide an easy way for developers and QA to send test messages to the app](https://mozilla-hub.atlassian.net/browse/VPN-2803)
- [VPN-2805 - How can we send push notifications to mobile devices when the app is on background / closed?](https://mozilla-hub.atlassian.net/browse/VPN-2805)

Note 1: filling as a draft, because this is still missing tests.
Note 2: I will continue the attempts to test this on mobile devices on Monday.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2517


.gitignore
```diff
@@ -192,3 +192,18 @@ extension/bridge/target
...
 extension/bridge/.cargo
+
+# Generated by Cargo
+# will have compiled files and executables
+debug/
+target/
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
+# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
+Cargo.lock
```
bakulf, line 203: I suspect we commit them. Can you check if we need to remove them?

src/addons/addonmessage.cpp
```diff
@@ -224,3 +227,23 @@ qint64 AddonMessage::planDateRetranslationInternal(
...
+  NotificationHandler* notificationHandler = NotificationHandler::instance();
+  if (!notificationHandler) {
+    // When loading at startup the notification handler is not available.
+    // That is fine, we no not want to notify when loading at startup.
+    return;
+  }
+
+  if (enabled() && !m_isRead) {
+    NotificationHandler::instance()->newInAppMessageNotification(m_title.get(),
+                                                                 m_title.get());
```
bakulf, line 247: This means that we notify at any startup right?

src/notificationhandler.h
```diff
@@ -58,7 +62,7 @@ class NotificationHandler : public QObject {
 
  private:
   virtual void notifyInternal(Message type, const QString& title,
-                              const QString& message, int timerMsec);
+                              const QString& message, int timerMsec = 5);
```
bakulf, line 65: Why this default value? Move it to constants.h

bakulf (changes requested):

src/addons/addonmessage.cpp
brizental: No, because the NotificationHandler is not available at startup and we check for that right before that check. That is fragile though, I'll see if I can figure out something better.

src/notificationhandler.h
brizental: I was using `0`, but that will throw and error on MacOS. So I swtiched to 5... Could have been 1. Anyways, I'll move it and add a comment.

.gitignore
brizental: Interestingly, doesn't look like we commited them, no. https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/signature

src/addons/addonmessage.h
```diff
@@ -19,7 +19,7 @@ class AddonMessage final : public Addon {
   ADDON_PROPERTY(subtitle, m_subtitle, retranslationCompleted)
 
   Q_PROPERTY(Composer* composer READ composer CONSTANT)
-  Q_PROPERTY(bool isRead MEMBER m_isRead NOTIFY isReadChanged)
```
brizental, line 22: Note: I removed the NOTIFY here because it was not being used anywhere.

tests/unit/main.cpp
```diff
@@ -42,15 +43,15 @@ int main(int argc, char* argv[]) {
   pe.insert("LANG", "en");
   pe.insert("LANGUAGE", "en");
 
-  QCoreApplication a(argc, argv);
+  QApplication app(argc, argv);
```
brizental, line 46: Had to make this change after importing the systray modules. Is this an issue @bakulf?

brizental: Hm, I think this is the culprit in breaking the Linux tests :thinking:

src/addons/addonmessage.h
brizental: TODO: Change this back, this is totally used [https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3e62d1245fa553311763b26d[…]ed9b5d2ad1836dba/src/ui/screens/messaging/ViewMessagesInbox.qml](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3e62d1245fa553311763b26ded9b5d2ad1836dba/src/ui/screens/messaging/ViewMessagesInbox.qml#L290)... Thanks @MattLichtenstein for confirming!

src/addons/addonmessage.cpp
```diff
@@ -74,35 +74,97 @@ AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
...
-void AddonMessage::maskAsRead() {
-  if (m_isRead) {
-    return;
+  QStringList readAddonMessages = settingsHolder->readAddonMessages();
+  if (readAddonMessages.contains(id)) {
+    return State::Read;
   }
 
-  m_isRead = true;
+  QStringList dismissedAddonMessages = settingsHolder->dismissedAddonMessages();
```
bakulf, line 91: This can introduce bugs where, for funny reasons, we have the same ID in the dismiss setting and in the read setting...and so on. What about if we change this all and we have 1 setting only: `addonMessageStates` (or something better).

I was wondering if it makes sense to do this:
```
Q_ASSERT(settingsHolder);

AddonMessage::State state = settingsHolder->addonProperty(addon->id(), "state")->castingToStateSomehow();
...
```

Basically we can add in the SettingsHolder a group for each addon: https://doc.qt.io/qt-5/qsettings.html#beginGroup
The group can be called `addons/<id>`.

src/addons/addonmessage.cpp
```diff
@@ -74,35 +74,97 @@ AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
...
+    case State::Received: {
+      // Do nothing. This state is not saved in settings,
+      // but it's also not an error.
+      break;
+    }
+    case State::Notified: {
+      QStringList notifiedAddonMessages =
+          settingsHolder->notifiedAddonMessages();
+      notifiedAddonMessages.removeAll(id());
+      settingsHolder->setNotifiedAddonMessages(notifiedAddonMessages);
```
bakulf, line 114: if we do a setting group per addon, this code can be simplified a lot. Thoughts?

src/addons/addonmessage.cpp
```diff
@@ -224,3 +286,28 @@ qint64 AddonMessage::planDateRetranslationInternal(
...
 }
+
+void AddonMessage::enable() {
+  Addon::enable();
+
+  maybePushNotification();
+}
+
+void AddonMessage::maybePushNotification() {
+  if (!m_conditionsParsed) {
```
bakulf, line 297: in theory, `::enable()` is called the first time _after_ maybeCreateConditionWatchers(): https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/addons/addon.cpp#L349-L352

If this happens before, it's a bug.

src/notificationhandler.h
```diff
@@ -57,8 +63,9 @@ class NotificationHandler : public QObject {
                       const QString& message, int timerMsec) = 0;
 
  private:
-  virtual void notifyInternal(Message type, const QString& title,
-                              const QString& message, int timerMsec);
+  virtual void notifyInternal(
+      Message type, const QString& title, const QString& message,
+      int timerMsec = Constants::DEFAULT_NOTIFICATION_TIMER);
```
bakulf, line 68: can we keep timerMSec as a "fix" param?

bakulf (changes requested):

src/addons/addonmessage.h
```diff
@@ -19,7 +23,7 @@ class AddonMessage final : public Addon {
...
+  Q_PROPERTY(bool isRead READ isRead)
```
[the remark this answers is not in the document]
brizental: Yeap I figured hehe https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4327#discussion_r963994333 WOMP WOMP

tests/unit/helper.h
```diff
@@ -1,5 +1,5 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * License; v. 2.0. If a copy of the MPL was not distributed with this
```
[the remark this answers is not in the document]
brizental: LOL probably a search and replace gone wrong... Thanks for catching it!

src/addons/addon.cpp
```diff
@@ -388,6 +388,13 @@ void Addon::retranslate() {
 }
 
 void Addon::maybeCreateConditionWatchers(const QJsonObject& conditions) {
+  if (m_conditionsParsed) {
```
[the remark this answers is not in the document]
brizental: I added this just because it with `m_conditionsParsed` there is an assumption that `maybeCreateConditionWatchers` will only be called once. That assumption was not enforced by code though so I added the enforcement... Q_ASSERT should work as well.

src/addons/addonmessage.cpp
brizental: Yeap, agreed. I did not know about settings groups when implementing. Will make the change.

src/addons/addonmessage.cpp
brizental: Correct. Development was a bit rocky on his feature and this was added before the AddonMessage::State which makes this whole `m_conditionsParsed` unnecessary it seems like :thinking:  I did not realize that until reading this comment. I'll just say goodbye to that state.

src/notificationhandler.h
brizental: Hm, I am not sure I know what you mean. Would you mind explaining what you mean by "fix param"?

src/addons/addonmessage.h
```diff
@@ -19,7 +23,7 @@ class AddonMessage final : public Addon {
...
   Q_PROPERTY(qint64 date MEMBER m_date)
```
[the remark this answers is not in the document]
brizental: Hm, I don't think I can. 

> The presence of the CONSTANT attribute indicates that the property value is constant. For a given object instance, the READ method of a constant property must return the same value every time it is called. This constant value may be different for different instances of the object. A constant property cannot have a WRITE method or a NOTIFY signal.

This value will require a `NOTIFY` and is also not a constant value, it can change.

src/addons/addonmessage.cpp
brizental: Addressed in [1320edd](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4327/commits/1320edd8c83262044dcc0837c16504d5aa3158ac)

src/addons/manager/addonmanager.cpp
```diff
@@ -153,7 +154,11 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
     beginResetModel();
   }
 
-  Q_ASSERT(m_addons.contains(addon->id()));
+  // If loading from the inspector command,
+  // this is likely the first time we are seeing this addon.
+  if (!fromInspectorCommand) {
+    Q_ASSERT(m_addons.contains(addon->id()));
+  }
   m_addons[addon->id()].m_addon = addon;
```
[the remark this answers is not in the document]
brizental: Is it an issue even though I am only skipping it when coming from the inspector?

tests/unit/main.cpp
brizental: Addressed in [eef9b69](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4327/commits/eef9b69cf5c213755eb842d591752a6150ae19eb)

src/addons/addonmessage.h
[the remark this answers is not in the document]
bakulf: no `m_date` cannot change. It's the date of the message and it's read from the manifest. It should be set as CONSTANT.

src/addons/addonmessage.h
```diff
@@ -31,10 +35,24 @@ class AddonMessage final : public Addon {
...
+    Dismissed
+  };
+  Q_ENUM(State);
+
   Q_INVOKABLE void dismiss();
-  Q_INVOKABLE void maskAsRead();
+  Q_INVOKABLE void markAsRead();
   Q_INVOKABLE bool containsSearchString(const QString& query) const;
 
+  bool isRead() { return m_state == State::Read; }
```
bakulf, line 54: `bool isRead() const`

src/addons/manager/addonmanager.cpp
```diff
@@ -424,8 +429,7 @@ QJSValue AddonManager::reduce(QJSValue callback, QJSValue initialValue) const {
 void AddonManager::reinstateMessages() const {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
   Q_ASSERT(settingsHolder);
-  settingsHolder->setDismissedAddonMessages(QStringList());
-  settingsHolder->setReadAddonMessages(QStringList());
+  settingsHolder->clearGroup(Constants::ADDON_MESSAGE_SETTINGS_GROUP);
```
bakulf, line 432: nice :)

src/constants.h
```diff
@@ -104,6 +104,11 @@ constexpr const char* APPLE_SUBSCRIPTIONS_URL =
 constexpr const char* GOOGLE_SUBSCRIPTIONS_URL =
     "https://play.google.com/store/account/subscriptions";
 
+// This is the default time a notification will be shown when received.
+constexpr uint32_t DEFAULT_NOTIFICATION_TIMER = 1 * 1000;  // 1s
```
bakulf, line 108: Is this been decided by Peter? 1 sec is really a short time. What do you think?

src/settingslist.h
```diff
@@ -422,63 +460,71 @@ SETTING_BOOL(unsecuredNetworkAlert,     // getter
...
 )
 
 SETTING_STRING(userEmail,     // getter
                setUserEmail,  // setter
                hasUserEmail,  // has
                "user/email",  // key
                "",            // default value
-               true           // remove when reset
+               true,          // remove when reset
+               ""             // group name
```
bakulf, line 500: these are actually grouped as `'user` but we can change it as a follow up.

src/systemtraynotificationhandler.cpp
```diff
@@ -27,6 +28,9 @@ SystemTrayNotificationHandler::SystemTrayNotificationHandler(QObject* parent)
   MozillaVPN* vpn = MozillaVPN::instance();
   Q_ASSERT(vpn);
 
+  m_menu = new QMenu();
```
bakulf, line 31: `new QMenu(this)`

src/systemtraynotificationhandler.cpp
```diff
@@ -27,6 +28,9 @@ SystemTrayNotificationHandler::SystemTrayNotificationHandler(QObject* parent)
...
+  m_systemTrayIcon = new QSystemTrayIcon();
```
bakulf, line 32: `new QSystemTrayIcon(this)`

src/systemtraynotificationhandler.cpp
```diff
@@ -245,8 +255,7 @@ void SystemTrayNotificationHandler::showHideWindow() {
   }
 }
 
-void SystemTrayNotificationHandler::maybeActivated(
-    QSystemTrayIcon::ActivationReason reason) {
+void SystemTrayNotificationHandler::maybeActivated(int reason) {
```
bakulf, line 258: why `int` and not `QSystemTrayIcon::ActivationReason` ?

src/systemtraynotificationhandler.h
```diff
@@ -32,13 +33,13 @@ class SystemTrayNotificationHandler : public NotificationHandler {
...
-  QMenu m_menu;
-  QSystemTrayIcon m_systemTrayIcon;
+  QMenu* m_menu;
+  QSystemTrayIcon* m_systemTrayIcon;
 
  private:
   void createStatusMenu();
 
-  void maybeActivated(QSystemTrayIcon::ActivationReason reason);
+  void maybeActivated(int reason);
```
bakulf, line 42: Let's keep the include `QSystemTrayIcon>` and let's have a reason enum. It seems easier to understand when reading the code.

src/systemtraynotificationhandler.cpp
bakulf: I see! It's for the include. I think we should keep the include<QSystemTrayIcon> and make the code more readable.

tests/unit/mocsystemtraynotificationhandler.cpp
```diff
@@ -0,0 +1,40 @@
...
+SystemTrayNotificationHandler::SystemTrayNotificationHandler(QObject* parent)
+    : NotificationHandler(parent) {}
+
+SystemTrayNotificationHandler::~SystemTrayNotificationHandler() {}
+
+void SystemTrayNotificationHandler::createStatusMenu() {}
+
+void SystemTrayNotificationHandler::setStatusMenu() {}
+
+void SystemTrayNotificationHandler::notify(NotificationHandler::Message type,
```
bakulf, line 19: nice!

bakulf (changes requested):

src/addons/addonmessage.h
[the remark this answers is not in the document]
brizental: Oh! You were referring to `m_date` hehe I thought you were referring to `isRead`

src/systemtraynotificationhandler.cpp
brizental: If we keep the include we can't import this module in tests :/

src/constants.h
brizental: Yeap I agree, I just noticed all the other notifications are using 4s so I'll stick to that.

src/systemtraynotificationhandler.h
brizental: How would you feel about having a comment here about QSystemTrayIcon::ActivationReason? Unless there is a way for me to import only this enum and not the whole QSystemTrayIcon class here

src/systemtraynotificationhandler.cpp
brizental: This fails.

src/addons/addonmessage.cpp
```diff
@@ -76,35 +78,45 @@ AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
...
+// static
+AddonMessage::State AddonMessage::loadMessageState(const QString& id) {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
-  Q_ASSERT(settingsHolder);
 
-  QStringList dismissedAddonMessages = settingsHolder->dismissedAddonMessages();
-  dismissedAddonMessages.append(id());
-  settingsHolder->setDismissedAddonMessages(dismissedAddonMessages);
-}
+  bool isValidState;
```
bakulf, line 85: move this to the if-stmt block.

src/addons/addonmessage.cpp
```diff
@@ -76,35 +78,45 @@ AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
...
+  QString stateSetting = settingsHolder->getAddonSetting(StateQuery(id));
+  QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
 
-void AddonMessage::maskAsRead() {
-  if (m_isRead) {
-    return;
+  if (!stateSetting.isEmpty()) {
```
bakulf, line 89: what happens if we remove check? Would the code work?

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
...
 
+  // Addon specific
+
+  struct AddonSettingQuery {
+    QString addonId;
+    QString addonGroup;
+    QString setting;
+    QString defaultValue;
+
+    AddonSettingQuery(QString ai, QString ag, QString s, QString dv)
```
bakulf, line 81: `const QString& ai, const QString& ag, const QString& s, const QString& dv`

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
 
   QString settingsFileName() const;
 
+  // Addon specific
+
+  struct AddonSettingQuery {
+    QString addonId;
```
bakulf, line 76: do we want to rename them to m_<foo> ?

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
...
+    QString addonGroup;
+    QString setting;
+    QString defaultValue;
+
+    AddonSettingQuery(QString ai, QString ag, QString s, QString dv)
+        : addonId(ai), addonGroup(ag), setting(s), defaultValue(dv) {}
+  };
+  QString getAddonSetting(AddonSettingQuery query);
```
bakulf, line 84: ` QString getAddonSetting(const AddonSettingQuery& query) const;`

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
...
+  bool hasAddonSetting(AddonSettingQuery query);
```
bakulf, line 85: ` bool hasAddonSetting(const AddonSettingQuery& query) const;`

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
...
+  void setAddonSetting(AddonSettingQuery query, const QString& value);
```
bakulf, line 86: ` void setAddonSetting(const AddonSettingQuery& query, const QString& value);`

src/settingsholder.h
```diff
@@ -69,11 +70,30 @@ class SettingsHolder final : public QObject {
...
+
  private:
   explicit SettingsHolder(QObject* parent);
 
   QString placeholderUserDNS() const;
 
+  // Addon specific
+
+  static QString getAddonSettingKey(AddonSettingQuery query);
```
bakulf, line 95: `const AddonSettingQuery& query`

src/systemtraynotificationhandler.cpp
bakulf: ah right. QMenu wants QWidget as parent.
Let's use: `QScopedPointer<QMenu> m_menu`;

src/addons/addonmessage.h
```diff
@@ -52,15 +75,28 @@ class AddonMessage final : public Addon {
...
  signals:
-  void isReadChanged();
+  void stateChanged(State state);
 
  private:
   AddonMessage(QObject* parent, const QString& manifestFileName,
                const QString& id, const QString& name);
 
+  struct StateQuery : public SettingsHolder::AddonSettingQuery {
+    StateQuery(const QString& ai)
```
bakulf, line 87: explicit

src/addons/addonmessage.h
```diff
@@ -52,15 +75,28 @@ class AddonMessage final : public Addon {
...
+
  signals:
-  void isReadChanged();
+  void stateChanged(State state);
 
  private:
   AddonMessage(QObject* parent, const QString& manifestFileName,
                const QString& id, const QString& name);
 
+  struct StateQuery : public SettingsHolder::AddonSettingQuery {
```
bakulf, line 86: struct StateQuery final : ...

bakulf (changes requested):

src/addons/addonmessage.cpp
brizental: Hm, I guess we can remove it since 1. there is a default value, so this should neve really be empty 2. isValidState will be `false` if it's empty.

brizental: I will just add two tests cases for that in case I am wrong.

src/addons/addonmessage.cpp
```diff
@@ -76,35 +78,43 @@ AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
...
-  dismissedAddonMessages.append(id());
-  settingsHolder->setDismissedAddonMessages(dismissedAddonMessages);
-}
+  QString stateSetting = settingsHolder->getAddonSetting(StateQuery(id));
+  QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
 
-void AddonMessage::maskAsRead() {
-  if (m_isRead) {
-    return;
+  bool isValidState;
```
bakulf, line 88: `bool isValidState = false;` just in case...

bakulf (approved):
Thanks!

PR #4329, opened 2022-09-05
Separate AddonApi per addon

bakulf:
Having an API object per add-on means that we can store data and we can keep separate state machines for each addon without conflicts.


brizental (approved):
I like this change. 💯

PR #4330, opened 2022-09-05
Remove shader based globe

strseb:
Okay, so the shader-based globe is broken on osx. We already have the fallback, which we need to maintain anyway and that does the same job - How about we just use it as default :) 

Screenshot 2022-09-05 at 13 30 21


flozia (approved):

flozia (commented):
We can remove `import compat 0.1` in `VPNMainImage.qml`.

nebula/ui/components/VPNMainImage.qml
```diff
@@ -367,47 +367,11 @@ Rectangle {
...
-    Image {
-        id: globe
-        source: "qrc:/nebula/resources/globe.svg";
-        sourceSize.height: logo.height
-        sourceSize.width: logo.width
-        visible: false
-    }
-
     Image {
         id: globeFallback
```
flozia, line 371: Nit: Maybe we should use another name for `id` since it’s not the fallback graphic anymore.

lesleyjanenorton (commented):
Do we know why this is works on local osx builds but not on treeherder builds? I'm fine with this approach but let's get that shield icon realigned.

strseb:
For some unknown reason to me, the Qt5 Compat Shaders are packaged incorrectly. So when we load them we get a "Z-Compress data error" - The shader won't get loaded, thus we never render the globe. 
I really am out of clues here, as baku/i/taskcluster are using the same src-zip and the same compile script. But only baku's build has the shaders in a non-broken state :c

nebula/ui/components/VPNMainImage.qml
```diff
@@ -367,47 +367,11 @@ Rectangle {
...
         source: logo.showVPNOnIcon ? "qrc:/ui/resources/main-img-vpn-on.svg" : "qrc:/ui/resources/main-img-vpn-off.svg"
         sourceSize.height: globe.height
         sourceSize.width: globe.width
-        visible: window._fallbackQtQuickRenderer
+        visible: true
```
bakulf, line 375: Remove this prop. Is visible=true the default value?

bakulf (commented):
What's the impact of this change?

PR #4335, opened 2022-09-05
fix: Always set the status icon dimensions to the status bar height

flozia:
## Description

Always scale the status bar icon to fit the status bar height.

## Reference
- [VPN-2520](https://mozilla-hub.atlassian.net/browse/VPN-2520): Status bar scaling issue: https://mozilla-hub.atlassian.net/browse/VPN-2520?focusedCommentId=582118

## Checklist
    
### DIFF
--- a/src/platforms/macos/macosutils.mm
+++ b/src/platforms/macos/macosutils.mm
@@ -190,11 +190,9 @@ void MacOSUtils::patchNSStatusBarSetImageForBigSur() {
 - (void)setImagePatched:(NSImage*)image {
   NSImage* img = image;
 
-  if (@available(macOS 11.0, *)) {
-    if (image != nil) {
-      int thickness = [[NSStatusBar systemStatusBar] thickness];
-      img = [NSImageScalingHelper imageByScaling:image size:NSMakeSize(thickness, thickness)];
-    }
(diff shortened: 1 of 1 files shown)


src/platforms/macos/macosutils.mm
```diff
@@ -190,11 +190,9 @@ @implementation NSStatusBarButton (Swizzle)
 - (void)setImagePatched:(NSImage*)image {
   NSImage* img = image;
 
-  if (@available(macOS 11.0, *)) {
```
bakulf, line 193: Can you tell me more what happens on MacOS <11? Is the API available? And why we had this it-stmt?

bakulf (changes requested):

src/platforms/macos/macosutils.mm
flozia: The API is available on `< 11`. The if-stmt — and the entire function was added to fix a Qt bug that caused status bar icons to be displayed stretched. The idea behind removing this statement is: Now this function resizes the image always to fit the size of the status bar so we can use one image for retina and non-retina displays. Which would be necessary now we use `png` instead of `svg` graphics.

bakulf (approved):

PR #4359, opened 2022-09-08
Optimize the filtering of messages

bakulf:
Without this PR, we close all the swipe at each filtering callback execution which means at least for each message. We just need to do it once when the search input-field value changes.

### DIFF
--- a/src/ui/screens/messaging/ViewMessagesInbox.qml
+++ b/src/ui/screens/messaging/ViewMessagesInbox.qml
@@ -87,15 +87,9 @@ VPNViewBase {
             _searchBarHasError: !root.isEmptyState && listView.count === 0
 
             _filterProxySource: VPNAddonManager
-            _filterProxyCallback: obj => {
-                                      root.closeAllSwipes()
-                                      root.isEditing = false
-                                      const filterValue = getSearchBarText();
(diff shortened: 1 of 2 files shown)


src/ui/screens/messaging/ViewMessagesInbox.qml
```diff
@@ -87,15 +87,12 @@ VPNViewBase {
             _searchBarHasError: !root.isEmptyState && listView.count === 0
 
             _filterProxySource: VPNAddonManager
-            _filterProxyCallback: obj => {
+            _filterProxyCallback: obj => obj.addon.type === "message" && obj.addon.containsSearchString(getSearchBarText())
+            _sortProxyCallback: (obj1, obj2) => obj1.addon.date > obj2.addon.date
+            _editCallback: () => {
                                       root.closeAllSwipes()
```
MattLichtenstein, line 93: If the model is refreshing with every new search query, do we even need to call `closeAllSwipes` since they close by default when the cells are re-initialized?

bakulf: You are right!

MattLichtenstein (approved):

PR #4377, opened 2022-09-08
Fix the navigation API for add-ons

bakulf:


brizental (approved):

src/addons/addonapi.cpp
```diff
@@ -94,7 +94,7 @@ QJSValue AddonApi::settings() const {
 QJSValue AddonApi::navigator() const {
   QJSEngine* engine = QmlEngineHolder::instance()->engine();
 
-  QObject* obj = SettingsHolder::instance();
+  QObject* obj = Navigator::instance();
```
brizental, line 97: Would it be possible to also add test cases that check this?

bakulf: because the test is silly: it checks if `vpn.navigator` exists but it does not call method because the navigator is not available in unit-tests.

brizental: :/ Alright!

PR #4385, opened 2022-09-08
Do not use lower-case feature names for Addon feature condition watcher - VPN-2827

bakulf:


brizental (approved):

src/addons/addon.cpp
```diff
@@ -68,7 +68,7 @@ QList<ConditionCallback> s_conditionCallbacks{
      [](Addon* addon, const QJsonValue& value) -> AddonConditionWatcher* {
        QStringList features;
        for (const QJsonValue& v : value.toArray()) {
-         features.append(v.toString().toLower());
```
brizental, line 71: Should we add some tests to prevent this from happening again?

bakulf: good point. We have tests for the condition watcher but not connected to the manifest parser.

PR #4392, opened 2022-09-09
Fix the message notification - VPN-2857

bakulf:
I also reduced the number of getAddonSettingsKey call because each time we were generating a key string.

### DIFF
--- a/src/addons/addonmessage.cpp
+++ b/src/addons/addonmessage.cpp
@@ -101,11 +101,12 @@ void AddonMessage::updateMessageState(State newState) {
 
   QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
   QString newStateSetting = stateMetaEnum.valueToKey(newState);
-  SettingsHolder* settingsHolder = SettingsHolder::instance();
-  settingsHolder->setAddonSetting(StateQuery(id()), newStateSetting);
 
   m_state = newState;
   emit stateChanged(m_state);
+
+  SettingsHolder* settingsHolder = SettingsHolder::instance();
+  settingsHolder->setAddonSetting(StateQuery(id()), newStateSetting);
 }
 
 void AddonMessage::dismiss() {
--- a/src/settingsholder.cpp
+++ b/src/settingsholder.cpp
@@ -80,14 +80,6 @@ void SettingsHolder::clear() {
 #undef SETTING
 }
 
-void SettingsHolder::clearGroup(const QString& group) {
-  logger.debug() << "Clean up the settings for group" << group;
-
-  m_settings.beginGroup(group);
-  m_settings.remove("");
-  m_settings.endGroup();
-}
-
 void SettingsHolder::sync() { m_settings.sync(); }
 
 void SettingsHolder::hardReset() {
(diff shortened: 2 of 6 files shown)


src/addons/addonmessage.cpp
```diff
@@ -101,11 +101,12 @@ void AddonMessage::updateMessageState(State newState) {
 
   QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
   QString newStateSetting = stateMetaEnum.valueToKey(newState);
-  SettingsHolder* settingsHolder = SettingsHolder::instance();
-  settingsHolder->setAddonSetting(StateQuery(id()), newStateSetting);
 
   m_state = newState;
   emit stateChanged(m_state);
+
```
bakulf, line 107: This is needed to receive the `addonSettingsChanged` after `stateChanged`.

src/settingsholder.cpp
```diff
@@ -186,16 +191,19 @@ bool SettingsHolder::hasAddonSetting(const AddonSettingQuery& query) {
 }
 
 QString SettingsHolder::getAddonSetting(const AddonSettingQuery& query) {
-  if (!hasAddonSetting(query)) return query.m_defaultValue;
-
   QString key = getAddonSettingKey(query);
+
+  if (!m_settings.contains(key)) return query.m_defaultValue;
```
brizental, line 196: We can delete `hasAddonSetting`, because with this change it's not being used anymore.

src/ui/navigator/navigationBar/VPNBottomNavigationBar.qml
```diff
@@ -150,26 +150,18 @@ Rectangle {
...
-            root.getUnreadNotificationStatus()
-        }
-        function onDismissedAddonMessagesChanged() {
+        function onAddonSettingsChanged() {
             root.getUnreadNotificationStatus()
         }
     }
 
     function getUnreadNotificationStatus() {
         messagesNavButton._hasNotification = VPNAddonManager.reduce((addon, initialValue) => initialValue + (addon.type === "message" && !addon.isRead), 0)
```
brizental, line 161: I know this is not part of this PR, but it's confusing, 1. Can we change `initialValue` to `sum` or `result` or something more expressive? 2. We are adding a number to a boolean here, WDYT about:
```suggestion
        messagesNavButton._hasNotification = VPNAddonManager.reduce((addon, sum) => {
            if (addon.type === "message" && !addon.isRead) {
               sum++;
            }
            return sum;
           }, 0)
```
There is no change in behaviour, it's just less confusing and dependant or weird implicit JavaScript type coersions i.e. `true -> 1`.

brizental: Also, I just noticed this was actually broken, because it needs to account for dismissed messages too. Would you liek to fix that in this PR or file a follow up to do that?

src/addons/addonmessage.cpp
brizental: I think one of the reasons this was broken by my PR is that there no tests to guarantee the behaviour. Is it possible to write tests for it? If so, would you prefer to do that on this PR or file a follow up?

brizental (changes requested):

src/ui/navigator/navigationBar/VPNBottomNavigationBar.qml
bakulf: that works because when a message is dismissed, it's disabled, so the counting will change.

brizental (approved):

Issue #4366, opened 2022-09-08
Optimise status bar indicator implementation


flozia:
Refactor status bar indicator to only be generated when needed and separate logic for different platforms.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2852)

Issue #4364, opened 2022-09-08
Black status bar indicator in state main


flozia:
On Linux and Windows the status indicator is black if the VPN is turned off. We need to add another state to match the behaviour with MacOS.

PR #4395, opened 2022-09-09
Fix and simplify status icon indicator

flozia:
## Description

Fixes and simplifies the status indicator color indicator.

## Reference

- #4364
- #4366 

## Checklist
    

src/statusicon.cpp
```diff
@@ -67,38 +72,41 @@ void StatusIcon::activateAnimation() {
...
-const QIcon& StatusIcon::icon() {
-  if (m_icon.isNull()) {
-    stateChanged();
-    Q_ASSERT(!m_icon.isNull());
+      m_indicatorColor = QColor();
+      break;
   }
 
-  return m_icon;
+  return m_indicatorColor;
```
bakulf, line 109: Why do we need this variable?

src/statusicon.cpp
```diff
@@ -176,29 +172,14 @@ QIcon StatusIcon::drawStatusIndicator(const QString& iconUrl) const {
...
+void StatusIcon::setIcon(const QString& iconUrl) {
   logger.debug() << "Set icon" << iconUrl;
 
-  m_iconUrl = iconUrl;
-
-  // If we are in state main and the VPN is on we draw the status indicator
-  // directly onto the image Windows and Linux.
-  if (shouldDrawIndicator) {
+  // Draw the status indicator directly onto the image if needed
+  if (!m_icon.isNull()) {
```
bakulf, line 179: Here you can simply invalidate m_icon

src/statusicon.cpp
```diff
@@ -48,17 +50,20 @@ StatusIcon::StatusIcon() : m_iconUrl(LOGO_GENERIC) {
...
-          [this] { setIcon(m_iconUrl, false); });
-#endif
 }
 
 StatusIcon::~StatusIcon() { MVPN_COUNT_DTOR(StatusIcon); }
 
+const QIcon& StatusIcon::icon() {
+  if (m_icon.isNull()) {
+    // Inits m_icon according to the current state
+    stateChanged();
```
bakulf, line 60: m_icon = draw()

src/statusicon.cpp
```diff
@@ -67,38 +72,41 @@ void StatusIcon::activateAnimation() {
...
-    Q_ASSERT(!m_icon.isNull());
+      m_indicatorColor = QColor();
+      break;
   }
 
-  return m_icon;
+  return m_indicatorColor;
 }
 
 void StatusIcon::stateChanged() {
```
bakulf, line 112: Here you can simply invalidate the m_icon and emit the signal. 
If you do this, you can remove `setIcon`

src/statusicon.h
```diff
@@ -16,37 +16,32 @@ class StatusIcon final : public QObject {
...
-
  public:
   StatusIcon();
   ~StatusIcon();
 
-  QUrl iconUrl() const;
-  const QString& iconString() const { return m_iconUrl; }
   const QIcon& icon();
-  const QColor& indicatorColor() const { return m_indicatorColor; }
+  const QString& iconString() const { return m_iconUrl; }
```
bakulf, line 24: Compute the iconString at each call. And remove m_iconUrl

src/statusicon.h
```diff
@@ -16,37 +16,32 @@ class StatusIcon final : public QObject {
...
   QIcon drawStatusIndicator(const QString& iconUrl) const;
-  void setIcon(const QString& iconUrl, bool shouldDrawIndicator);
-  void setIndicatorColor(const QColor& color);
+  void generateIcon();
+  void setIcon(const QString& iconUrl);
 
  private:
-  QString m_iconUrl;
   QIcon m_icon;
+  QString m_iconUrl;
```
bakulf, line 44: You can remove this or to optimize the code, keep it and use it as a cache like m_icon.
In stateChanged, set it to null.

src/statusicon.h
```diff
@@ -16,37 +16,32 @@ class StatusIcon final : public QObject {
...
   QColor m_indicatorColor;
```
bakulf, line 45: And this

tests/unit/teststatusicon.cpp
```diff
@@ -10,64 +10,49 @@
...
 #else
-  QCOMPARE(si.iconUrl().toString(), "qrc:/ui/resources/logo-generic-mask.png");
   QCOMPARE(si.iconString(), ":/ui/resources/logo-generic-mask.png");
 #endif
 
   TestHelper::vpnState = MozillaVPN::StateMain;
   TestHelper::controllerState = Controller::StateOn;
   si.stateChanged();
 #if defined(MVPN_LINUX) || defined(MVPN_WINDOWS)
-  QCOMPARE(si.iconUrl().toString(), "qrc:/ui/resources/logo-on.png");
```
bakulf, line 33: Why this change? Here we can probably remove all of these #ifdef macro

bakulf (changes requested):

src/statusicon.cpp
flozia: We do not need it and can return the colors directly.

src/statusicon.cpp
```diff
@@ -43,117 +46,122 @@ constexpr const char* LOGO_GENERIC_ON =
...
     default:
-      setIcon(LOGO_GENERIC, false);
+      m_animatedIconTimer.stop();
+      return LOGO_GENERIC;
       break;
   }
 }
 
-QIcon StatusIcon::drawStatusIndicator(const QString& iconUrl) const {
+const QColor& StatusIcon::indicatorColor() {
```
bakulf, line 125: I would not return a reference. Return a QColor

src/statusicon.cpp
```diff
@@ -43,117 +46,122 @@ constexpr const char* LOGO_GENERIC_ON =
...
+      return RED_COLOR;
+    default:
+      logger.error() << "Unhandled status indicator for connection stability";
+      return INVALID_COLOR;
+  }
+
+  return m_indicatorColor;
+}
+
+void StatusIcon::stateChanged() {
```
bakulf, line 151: What about if we call this method `refreshNeeded` or `resetIcon`? Asking because it's not connected to state changing anymore.

tests/unit/teststatusicon.cpp
```diff
@@ -10,64 +10,49 @@
 void TestStatusIcon::basic() {
   StatusIcon si;
 #if defined(MVPN_LINUX) || defined(MVPN_WINDOWS)
-  QCOMPARE(si.iconUrl().toString(), "qrc:/ui/resources/logo-generic.png");
```
bakulf, line 13: You can still keep these tests

bakulf (changes requested):
Great refactoring! Can you please improve the tests? I think you can remove the ifdef and test the single methods on all the platforms.

src/statusicon.cpp
flozia: Sounds good!

tests/unit/teststatusicon.cpp
flozia: The method `iconUrl` is not used anymore.

bakulf (approved):

PR #4403, opened 2022-09-09
Create a task to sign the addons-manifest

ahal:
It turns out that Chain of Trust has a hardcoded list of kinds that are allowed to be part of a release pipeline and fails due to this being an 'addons' task.

We could update this list, but that would involve a new scriptworker release + rollout. It's simpler to move this over to the build kind instead.

## Checklist
    
- [ x ] My code follows the style guidelines for this project
- [ x ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ x ] I have performed a self review of my own code
- [ x ] I have commented my code PARTICULARLY in hard to understand areas
- [ x ] I have added thorough tests where needed


ahal:
There's currently no release signer for this in autograph, so the task will fail if we attempt to do any release signing with it. Though currently this patch is set up so that it won't run anyway. This is mostly just to get the code in place to help unblock gabriel.

strseb (approved):

PR #4409, opened 2022-09-12
adding multihop functional tests, some optimizations to helper class

mozrokafor:
## Description

    added multihop connectivity functional tests

## Reference

    [i.e Jira or Github issue URL](https://mozilla-hub.atlassian.net/browse/PSP-493)

## Checklist
    
### DIFF
--- a/tests/functional/actions.js
+++ b/tests/functional/actions.js
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */ 
+
+const vpn = require('./helper.js');
+ 
+const connectionChangeNotification = async () => {
+    return await vpn.setSetting('connection-change-notification', 'true');
+}
+
+const serverSwitchNotification = async () => {    
+    return await vpn.setSetting('server-switch-notification', 'true');
+}
+ 
+
+module.exports = {
+    actions: {
+        settings: {
+            connectionChangeNotification,
+            serverSwitchNotification,
+        }
+    }
+}
--- a/tests/functional/actions.js
+++ b/tests/functional/actions.js
@@ -2,22 +2,36 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */ 
 
+const elements = require('./elements.js');
 const vpn = require('./helper.js');
  
-const connectionChangeNotification = async () => {
-    return await vpn.setSetting('connection-change-notification', 'true');
+const connectionChangeNotification = async (status = 'true') => {
+    return await vpn.setSetting('connection-change-notification', status);
 }
 
-const serverSwitchNotification = async () => {    
-    return await vpn.setSetting('server-switch-notification', 'true');
+const serverSwitchNotification = async (status = 'true') => {    
+    return await vpn.setSetting('server-switch-notification', status);
 }
- 
+
+const selectCountryFromList = async (countryId) => {
+  await vpn.setElementProperty(elements.SERVER_COUNTRY_VIEW, 'contentY', 'i', parseInt(await vpn.getElementProperty(countryId, 'y')));
+  await vpn.wait()
+}
+
+const selectCityFromList = async (cityId, countryId) => {
+  await vpn.setElementProperty(elements.SERVER_COUNTRY_VIEW, 'contentY', 'i', parseInt(await vpn.getElementProperty(cityId, 'y')) + parseInt(await vpn.getElementProperty(countryId, 'y')));
+  await vpn.wait()
+}
 
 module.exports = {
     actions: {
         settings: {
             connectionChangeNotification,
             serverSwitchNotification,
+        },
+        locations: {
+            selectCountryFromList,
+            selectCityFromList
         }
     }
 }
--- a/tests/functional/actions.js
+++ b/tests/functional/actions.js
@@ -5,12 +5,12 @@
 const elements = require('./elements.js');
 const vpn = require('./helper.js');
  
-const connectionChangeNotification = async (status = 'true') => {
-    return await vpn.setSetting('connection-change-notification', status);
+const setConnectionChangeNotification = async (status) => {
+    return await vpn.setSetting('connection-change-notification', status ? "true" : "false");
 }
 
-const serverSwitchNotification = async (status = 'true') => {    
-    return await vpn.setSetting('server-switch-notification', status);
+const setServerSwitchNotification = async (status) => {    
+    return await vpn.setSetting('server-switch-notification', status ? "true" : "false");
 }
 
 const selectCountryFromList = async (countryId) => {
@@ -26,8 +26,8 @@ const selectCityFromList = async (cityId, countryId) => {
 module.exports = {
     actions: {
         settings: {
-            connectionChangeNotification,
-            serverSwitchNotification,
+            setConnectionChangeNotification,
+            setServerSwitchNotification,
         },
         locations: {
             selectCountryFromList,
--- a/tests/functional/helper.js
+++ b/tests/functional/helper.js
@@ -162,6 +162,7 @@ module.exports = {
   async waitForElementAndClick(id) {
     await this.waitForElementAndProperty(id, 'visible', 'true');
     await this.clickOnElement(id)
+    await this.wait()
   },
 
   async clickOnNotification() {
@@ -251,8 +252,8 @@ module.exports = {
     }
   },
 
-  wait() {
-    return new Promise(resolve => setTimeout(resolve, 500));
+  wait(waitTimeInMilliSecs = 1000) {
+    return new Promise(resolve => setTimeout(resolve, waitTimeInMilliSecs));
   },
 
   // TODO - The expected staging urls are hardcoded, we may want to
--- a/tests/functional/helper.js
+++ b/tests/functional/helper.js
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 const assert = require('assert');
 const {URL} = require('node:url');
 const http = require('http')
@@ -160,8 +159,8 @@ module.exports = {
         `Command failed: ${json.error}`);
   },
 
-  async waitForPropertyAndClickOnElement(id, property) {    
-    await this.waitForElementProperty(property ? property : id, 'visible', 'true');
+  async waitForElementAndClick(id) {
+    await this.waitForElementAndProperty(id, 'visible', 'true');
     assert(await this.hasElement(id), 'Clicking on an non-existing element?!?');
     const json = await this._writeCommand(`click ${id}`);
     assert(
@@ -206,7 +205,6 @@ module.exports = {
   },
 
   async setElementProperty(id, property, type, value) {
-    await this.waitForElement(id)
     assert(
         await this.hasElement(id),
         'Property checks must be done on existing elements');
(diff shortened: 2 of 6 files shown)


tests/functional/helper.js
```diff
@@ -160,6 +160,15 @@ module.exports = {
         `Command failed: ${json.error}`);
   },
 
+  async waitForPropertyAndClickOnElement(id, property) {    
+    await this.waitForElementProperty(property ? property : id, 'visible', 'true');
```
bakulf, line 164: I don't understand the logic here. The first argument should be the element, but you are passing the property. Can you tell me how do you want to use this?

bakulf: Plus it seems that a second argument is not used in this code.

tests/functional/helper.js
```diff
@@ -197,6 +206,7 @@ module.exports = {
   },
 
   async setElementProperty(id, property, type, value) {
+    await this.waitForElement(id)
```
bakulf, line 209: this breaks the logic. `set` should be used only when the element already exists.

tests/functional/helper.js
```diff
@@ -208,6 +218,7 @@ module.exports = {
   },
 
   async waitForElementProperty(id, property, value) {
+    await this.waitForElement(id)
```
bakulf, line 221: mmm maybe rename this to `waitForElementAndProperty()`

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ const { beforeEach } = require('mocha');
```
bakulf, line 4: empty line between line 3 and 4.

bakulf: plus this line can be removed. See other tests.

bakulf (changes requested):
Can you add these tests:
- how to migrate from single hop to multi-hop
- test the connection of the VPN and the migration single-hop -> multi-hop
- test the notifications (optional)

tests/functional/testMultihop.js
mozrokafor: ah thought i removed this already, good catch

tests/functional/helper.js
mozrokafor: the thought behind it, was to wait for element before setting it. what do you think?

tests/functional/helper.js
```diff
@@ -160,6 +159,15 @@ module.exports = {
         `Command failed: ${json.error}`);
   },
 
+  async waitForElementAndClick(id) {
+    await this.waitForElementAndProperty(id, 'visible', 'true');
+    assert(await this.hasElement(id), 'Clicking on an non-existing element?!?');
+    const json = await this._writeCommand(`click ${id}`);
```
bakulf, line 165: can you use clickOnElement() ?

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,119 @@
...
+  
+      // select city      
+      await vpn.waitForElementAndClick(cityTwoId);
+      await vpn.wait()
+      
+      // navigate back to 
+      await vpn.waitForElementAndClick('serverListBackButton');
+      await vpn.wait();
+
+      await vpn.activate();
```
bakulf, line 104: where do you deactivate the VPN?

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,119 @@
...
+
+      await vpn.waitForCondition(async () => {
+        return await vpn.getElementProperty('controllerTitle', 'text') ==
+            'VPN is on';
+      });
+      await vpn.wait();
```
bakulf, line 110: you should check the name of the cities somehow.

bakulf (changes requested):

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,125 @@
...
+      await vpn.waitForElementAndClick('multiHopSelector/tabMultiHop');
+      await vpn.wait()    
+       
+      // select entry
+      await vpn.waitForElementAndClick('buttonSelectEntry');    
+      
+      // wait for country of choice to be visible    
+      const server = servers[i]
+      const countryId = 'serverCountryList/serverCountry-' + server.code;
+      const exitCountryId = 'serverCountryList/serverCountry-' + servers[0].code;
```
bakulf, line 50: you  always pick the same exit node?

bakulf (changes requested):
I think we need to add more tests here:
- is the 'what is multi-hop VPN?' block visible? Can it be expanded? etc.
- add tests for the exit location
- can you search in the server list?
- do we have the correct 'select location' string after the selection?
- we should test that going back to 1 single top all is correctly changed

tests/functional/testMultihop.js
mozrokafor: it'll need to be updated once bug is fixed

bakulf (changes requested):
I would remove all the getRandomInt() use and make a strong and reliable test. I think testing everything is more important than run fast. Functional tests have 3 countries with 1 or 2 cities each. In total 4 cities. All the possible combinations are 4^2 => 16 configurations.

I still don't see tests for:
* is the 'what is multi-hop VPN?' block visible? Can it be expanded? etc.
* add tests for the exit location
* can you search in the server list?
* single hop -> multi hop -> single hop -> in the server-switching context.

I still think that covering the search is important. Having a separate test is good but it means that we need to have a lot of dup code. The search needs to be implemented both in the single-hop and in the multi-hop (exit and enter lists).

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,152 @@
...
+      await vpn.wait()      
+      await vpn.waitForElementAndClick(cityTwoId);
+      
+      // navigate back to 
+      await vpn.waitForElementAndClick('serverListBackButton');
+
+      currentCountry = servers[idx].localizedName;
+      currentCity = cityTwo.localizedName;      
+
+      if(i < 1){
```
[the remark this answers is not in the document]
mozrokafor: that would not cause a server switch, it would simply connect again everytime

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,152 @@
...
+      await vpn.wait();
+
+      if (await vpn.getElementProperty(countryId, 'cityListVisible') === 'false') {
+        await vpn.clickOnElement(countryId);
+      }  
+      
+      await vpn.waitForElementProperty(countryId, 'cityListVisible', 'true');
+  
+      // navigate and select city    
+      const cityOne = server.cities[getRandomInt(server.cities.length)]
```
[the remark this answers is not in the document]
mozrokafor: long tests are not ideal especially when the same coverage can be achieved in a shorter manner. as this is being assigned to a constant it'll be used throughout the loop as if it was individually looped through all

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,152 @@
...
+         }
+       }
+     }
+     console.log(
+         'Current city (localized):', currentCity,
+         '| Current country code:', currentCountryCode);
+   });
+
+   it('Checking multihop functionality', async () => {
+    for(let i=0; i<2; i++){
```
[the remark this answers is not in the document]
mozrokafor: good catch, moving back to servers.length

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,152 @@
...
+      // wait for select entry      
+      await vpn.waitForElementAndClick('multiHopSelector/tabMultiHop');
+             
+      // select entry
+      await vpn.waitForElementAndClick('buttonSelectEntry');
+      
+      // wait for country of choice to be visible    
+      const server = servers[i]
+      const countryId = 'serverCountryList/serverCountry-' + server.code;
+      const exitCountryId = 'serverCountryList/serverCountry-' + servers[idx].code;
```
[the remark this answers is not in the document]
mozrokafor: this can be simply fixed by doing a getRandomInt(servers.length), that way it will always pick a server within range

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,152 @@
...
+      await vpn.waitForElementAndClick('serverListButton');
+    }
+
+    // Verify switch to single hop
+    await vpn.waitForElementAndClick('multiHopSelector/tabSingleHop');
+    await vpn.waitForElementAndClick('serverListBackButton');
+
+    await vpn.waitForCondition(async () => {
+      return vpn.lastNotification().title == "VPN Switched Servers"
+    }, 5000)
```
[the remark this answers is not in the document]
mozrokafor: > 

long tests are equivalent to tickets that are 13 story points, they should be split. The longer the test the more flakiness and instability that is being introduced. Even looking at the QA manual test cases, they do not include search within multihop test case. I recommend search be its own test that is concisely only about searching both single and multihop as search does not effect single or multihop's main function.

fReshGreenb (changes requested):
[nebula/ui/components/VPNCollapsibleCard.qml](tests/functional/testMultihop.js)

bakulf (changes requested):
0. I like the elements.js and actions.js!

1. Are you planning to cleanup also the other tests?

2. There are still random variables. Tests should be deterministic. If I pick servers randomly in the test it means that I make the test not deterministic and I increase the probability to have intermittent failures or to have non-easily-reproducible bugs. Because we have only 4 cities, I suggest having a server-switching tests in this way:

```
server(cityA, countryA) -> server(city B,country A)
server(cityB, country A -> server(city D, country D)
```

This should cover all the options, I guess (different cities in the same country, different city in a different country).

tests/functional/actions.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */ 
+
+const elements = require('./elements.js');
+const vpn = require('./helper.js');
+ 
+const connectionChangeNotification = async (status = 'true') => {
+    return await vpn.setSetting('connection-change-notification', status);
```
bakulf, line 9: the best is to have status as boolean and convert it to string here. Something like `status ? "true" : "false"`

tests/functional/actions.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const serverSwitchNotification = async (status = 'true') => {    
+    return await vpn.setSetting('server-switch-notification', status);
```
bakulf, line 13: here as well

tests/functional/testMultihop.js
```diff
@@ -0,0 +1,383 @@
...
+
+  it('Server switching', async () => {
+    let newCurrentCountry;
+    let newCurrentCity;
+
+    let serverOne = servers[Math.floor(Math.random() * servers.length)]
+    let serverTwo = await _chooseNewServer(serverOne, servers)    
+    let serverThree;
+    
+    await actions.settings.serverSwitchNotification()
```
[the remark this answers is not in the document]
bakulf: `serverSwitchNotification` does not communicate that the setting flag is set to true. What about:

`action.settings.setServerSwitchNotification(true)` with the true/false passed explicitly?

tests/functional/actions.js
```diff
@@ -0,0 +1,37 @@
...
+const selectCityFromList = async (cityId, countryId) => {
+  await vpn.setElementProperty(elements.SERVER_COUNTRY_VIEW, 'contentY', 'i', parseInt(await vpn.getElementProperty(cityId, 'y')) + parseInt(await vpn.getElementProperty(countryId, 'y')));
+  await vpn.wait()
+}
+
+module.exports = {
+    actions: {
+        settings: {
+            connectionChangeNotification,
+            serverSwitchNotification,
```
bakulf, line 30: setServerSwitchNotification and setConnectionChangeNotification ?
Otherwise it seems that those methods returns the current values. They sounds like getters and not setters.

bakulf (approved):
Apply the comments before landing the PR. Thanks!

PR #4452, opened 2022-09-14
Reschedulable tasks

bakulf:
This is the algorithm I was discussing with you. It seems to work fine. Not for 2.10.

### DIFF
--- a/src/tasks/group/taskgroup.cpp
+++ b/src/tasks/group/taskgroup.cpp
@@ -51,9 +51,35 @@ void TaskGroup::cancel() {
   }
 }
 
-bool TaskGroup::deletable() const {
+Task::DeletePolicy TaskGroup::deletePolicy() const {
+  // The policy for task group is:
+  // - if there is at least 1 non-deletable task, the group is non-deletable
+  // - if the tasks are all reschedulable, the group is reschedulable
+  // - if there are no reschedulable tasks, the group is deletable
+  // - fallback: the task is not deletable
+
+  int reschedulable = 0;
   for (Task* task : m_tasks) {
-    if (!task->deletable()) return false;
+    switch (task->deletePolicy()) {
+      case NonDeletable:
+        return NonDeletable;
+
+      case Reschedulable:
+        ++reschedulable;
+        break;
+
+      default:
+        break;
+    }
   }
-  return true;
+
+  if (reschedulable == m_tasks.length()) {
+    return Reschedulable;
+  }
+
+  if (reschedulable == 0) {
+    return Deletable;
+  }
+
+  return NonDeletable;
 }
--- a/src/tasks/group/taskgroup.cpp
+++ b/src/tasks/group/taskgroup.cpp
@@ -56,7 +56,7 @@ Task::DeletePolicy TaskGroup::deletePolicy() const {
   // - if there is at least 1 non-deletable task, the group is non-deletable
   // - if the tasks are all reschedulable, the group is reschedulable
   // - if there are no reschedulable tasks, the group is deletable
-  // - fallback: the task is not deletable
+  // - if there is at least 1 reschedulable task, the group in non-deletable
 
   int reschedulable = 0;
   for (Task* task : m_tasks) {
--- a/src/taskscheduler.cpp
+++ b/src/taskscheduler.cpp
@@ -79,8 +79,6 @@ void TaskScheduler::taskCompleted() {
 }
 
 void TaskScheduler::deleteTasksInternal() {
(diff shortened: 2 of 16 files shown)


src/tasks/group/taskgroup.cpp
```diff
@@ -51,9 +51,35 @@ void TaskGroup::cancel() {
   }
 }
 
-bool TaskGroup::deletable() const {
+Task::DeletePolicy TaskGroup::deletePolicy() const {
+  // The policy for task group is:
+  // - if there is at least 1 non-deletable task, the group is non-deletable
+  // - if the tasks are all reschedulable, the group is reschedulable
+  // - if there are no reschedulable tasks, the group is deletable
+  // - fallback: the task is not deletable
```
brizental, line 59: Hum, it is not obvious to me why you landed on this policy for groups. Would you mind expanding?

src/taskscheduler.cpp
```diff
@@ -79,19 +79,40 @@ void TaskScheduler::taskCompleted() {
...
-  if (m_running_task && m_running_task->deletable()) {
+  if (m_running_task && m_running_task->deletePolicy() == Task::Deletable) {
     m_running_task->cancel();
     m_running_task->deleteLater();
     m_running_task->disconnect();
     m_running_task = nullptr;
   }
+
+  if (!reschedulableTasks.isEmpty()) {
+    QTimer::singleShot(500, this, [this, reschedulableTasks]() {
```
brizental, line 112: Why wait 500 seconds pushing the task to the end of the line? Also, what if we have multiple reschedulable tasks in a row? One single run of this timer would reschedule all of them, but there would be multiple timers running, no? Wouldn't it make more sense to schedule a timer per task and ditch the `reschedulabeTasks` list?

tests/unit/testtasks.cpp
```diff
@@ -104,8 +104,58 @@ void TestTasks::function() {
   QVERIFY(completed);
 }
 
-void TestTasks::removeDevice() {
```
brizental, line 107: Would you mind filing a ticket as a reminder that we need to write removeDevice task tests?

tests/unit/testtasks.cpp
```diff
@@ -104,8 +104,58 @@ void TestTasks::function() {
...
+
+  // The first one is t2.
+  QCOMPARE(sequence.at(0), 2);
+
+  // Then we have t4 because t3 is rescheduable.
+  QCOMPARE(sequence.at(1), 4);
+
+  // The 2 reschedulable tasks: t3 and the sentinel.
+  QCOMPARE(sequence.at(2), 3);
+  QCOMPARE(sequence.at(3), 99);
```
brizental, line 158: I think there should also be tests for the TaskGroup delete policy.

src/tasks/group/taskgroup.cpp
bakulf: My idea is that the group should not override the policy of the sub-tasks.
1. The first point seems reasonable: if there is at least 1 non-deletable task the whole group is non-deletable.
2. The second one makes sense too: if all are reschedulable, the group is reschedulable.
3. The third one seems correct: if all are deletable -> let's delete the entire group
4. the last one is the complex one. If we are here, we have some reschedulable tasks, and some deletable ones. We don't know if changing the order of the execution we can introduce bugs, so my approach is: do nothing, keep the group as it is.

tests/unit/testtasks.cpp
bakulf: https://mozilla-hub.atlassian.net/browse/VPN-2896

src/tasks/group/taskgroup.cpp
brizental:
```suggestion
  // - if there is at least 1 reschedulable task, the group in non-deletable
```
Just to make it clear.

tests/unit/testtasks.cpp
```diff
@@ -104,8 +105,97 @@ void TestTasks::function() {
...
+                     new TaskFunction([&]() {}, Task::Reschedulable)});
+    QCOMPARE(g->deletePolicy(), Task::Reschedulable);
+    delete g;
+  }
+
+  // Not deletable group if mixed
+  {
+    TaskGroup* g = new TaskGroup(
+        QList<Task*>{new TaskFunction([&]() {}, Task::NonDeletable),
+                     new TaskFunction([&]() {}, Task::Reschedulable)});
```
brizental, line 195: This group falls on the same code path as the group in line 172. Maybe do Deletable, Reschedulable instead?

PR #4457, opened 2022-09-15
VPN-2882 - Add header to tutorials tab

brizental:


brizental:
@MattLichtenstein Any idea why this change broke the tutorial tests? It's the same thing as last time we spoke, it can't find the highlighted tutorial... None of the object names changed or anything I don't get it /o\

MattLichtenstein:
> @MattLichtenstein Any idea why this change broke the tutorial tests? It's the same thing as last time we spoke, it can't find the highlighted tutorial... None of the object names changed or anything I don't get it /o\

The object names haven't changed, but the nesting structure has.

Any component created by a Repeater (and Loader I believe as well - @bakulf can you confirm?) can only be accessed through it's parent.

We went from:

`VPNTutorialList (GridLayout) > Repeater`

to:

`ColumnLayout > Repeater > VPNTipsAndTricksSection (ColumnLayout) > Loader > VPNTutorialList (GridLayout) > Repeater`

So we'll have to set the `objectName` property for all the new nesting components. A little messy, but I believe it is the only way to gain access to these components. It should look something like this: 

Screen Shot 2022-09-15 at 1 57 32 PM
Screen Shot 2022-09-15 at 1 58 06 PM
Screen Shot 2022-09-15 at 1 58 40 PM

brizental:
> The object names haven't changed, but the nesting structure has.

Actually it hasn't. The nesting structure changes on the "Tutorials" tab but not on the "All" tab, which is where the tests broke... The change makes sense to me, even though I don't like it!

Thank you so much for looking into this though. I'll apply the suggestions!

MattLichtenstein (approved):

PR #4458, opened 2022-09-15
VPN-2883 - Do not send new message notifications to logged out users

brizental:


tests/unit/addons/message2.json
```diff
@@ -4,13 +4,14 @@
   "name": "Test Message 2",
   "translatable": false,
   "type": "message",
-  "conditions": { },
+  "conditions": {
+    "end_time": 1599424544
```
bakulf, line 8: this is in the past. So message2 will not be loaded.

bakulf (approved):

tests/unit/addons/message2.json
brizental: yeap that is on purpose. that was the way i found to load the message and be sure it is disabled.

PR #4459, opened 2022-09-15
Recompute the device-limit view when the device is removed - VPN-2847

bakulf:
I moved the logic in `deviceRemoved()`

### DIFF
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -25,6 +25,7 @@
 #include "settingsholder.h"
 #include "task.h"
 #include "urlopener.h"
+#include "websocket/pushmessage.h"
 
 #include <functional>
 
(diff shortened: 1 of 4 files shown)


src/mozillavpn.cpp
```diff
@@ -662,36 +681,19 @@ void MozillaVPN::removeDeviceFromPublicKey(const QString& publicKey) {
...
-    }
+  // Let's inform the UI about what is going to happen.
+  emit deviceRemoving(publicKey);
+  TaskScheduler::scheduleTask(new TaskRemoveDevice(publicKey));
 
-    maybeStateMain();
-  }));
+  if (m_state != StateDeviceLimit) {
+    // To have a faster UI, we inform the device-model that this public key
+    // is going to be removed.
```
brizental, line 694: At which point is the UI informed about the device removal, here, on line 688 or both? And if both, why? This is confusing.

brizental (changes requested):
Would it be possible to write a test for this behaviour?

src/inspector/inspectorhandler.cpp
```diff
@@ -939,6 +940,22 @@ static QList<InspectorCommand> s_commands{
...
+          payload["publicKey"] = QString(arguments[1]);
+
+          QJsonObject msg;
+          msg["type"] = "DEVICE_DELETED";
+          msg["payload"] = payload;
+
+          PushMessage message(QJsonDocument(msg).toJson());
+          message.executeAction();
+          return QJsonObject();
+        }},
```
brizental, line 958: Nice!

brizental (approved):

tests/functional/testDevices.js
```diff
@@ -169,8 +169,126 @@ describe('Devices', function() {
...
+      await vpn.waitForElementProperty('deviceListView', 'visible', 'true');
+
+      await vpn.waitForElement('deviceLimitHeader');
+      await vpn.waitForElementProperty('deviceLimitHeader', 'visible', 'true');
+
+      const key = UserData.devices[0].pubkey;
+      UserData.devices.splice(0, 1);
+      await vpn.sendPushMessageDeviceDeleted(key);
+
+      await vpn.waitForElement('controllerTitle');
```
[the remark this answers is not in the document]
bakulf: It means that the UI shows the controller main view. The title is "VPN is On" or things like this.

Issue #3682, opened 2022-06-06
[foxfooding] Random VPN location is unexpected


pdehaan:
> _"The apparently random selection of the first VPN location is surprising. Or I presume it's random; why otherwise would the desktop client have picked Bucharest for me, and the Android client Vienna?"_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2333)

strseb:
Closing this as we as of 2.14 select the closest location based on ping :)

PR #4467, opened 2022-09-15
Experiment: Auto-select server via GeoIP proximity

oskirby:
## Description
A significant number of our users will often find themselves installing the VPN client, and using it in its default configuration indefinitely, meaning that they may never actually change their server away from the randomly-selected initial value. This can result in poor VPN performance if the random selection chooses a server on the opposite side of the internet.

A better approach for this initial selection would be to choose the nearest location and hope that it happens to correlate to the best network performance.

At present, we don't receive the client's latitude and longitude from the guardian, so for a proof of concept we are making use of a third party API to receive GeoIP data.

## Reference
Issue #3682 ([VPN-2333](https://mozilla-hub.atlassian.net/browse/VPN-2333))

## Checklist
- [ ] Implement or update a guardian API to provide GeoIP data (latitude/longitude).

### DIFF
--- a/src/cmake/sources.cmake
+++ b/src/cmake/sources.cmake
@@ -314,6 +314,8 @@ target_sources(mozillavpn PRIVATE
     tasks/sendfeedback/tasksendfeedback.h
     tasks/servers/taskservers.cpp
     tasks/servers/taskservers.h
+    tasks/serverselect/taskserverselect.cpp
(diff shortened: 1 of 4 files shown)


oskirby:
Closing this PR for now, I will pick the work back up once 1.28 is released to Guardian staging.

PR #4481, opened 2022-09-19
VPN-2886 - Fix warning about titleComponent not being defined

brizental:
Not sure why this warning was only showing on the Tips & Tricks page. AFAICT it could be shown in any page (?).

### DIFF
--- a/nebula/ui/components/VPNViewBase.qml
+++ b/nebula/ui/components/VPNViewBase.qml
@@ -15,7 +15,6 @@ Item {
 
    property string _menuTitle: ""
    property var _menuOnBackClicked
-   property var titleComponent
    property alias _viewContentData: viewContent.data
    property alias _interactive: vpnFlickable.interactive
 
--- a/nebula/ui/components/VPNViewBase.qml
+++ b/nebula/ui/components/VPNViewBase.qml
@@ -21,7 +21,7 @@ Item {
 
    anchors {
        top: if (parent) parent.top
-       topMargin:  _menuTitle !== "" || titleComponent || rightMenuButton ? VPNTheme.theme.menuHeight : 0
+       topMargin: menu.visible ? VPNTheme.theme.menuHeight : 0
    }
 
    Rectangle {
--- a/nebula/ui/components/VPNViewBase.qml
+++ b/nebula/ui/components/VPNViewBase.qml
@@ -15,6 +15,7 @@ Item {
 
    property string _menuTitle: ""
    property var _menuOnBackClicked
+   property var titleComponent
    property alias _viewContentData: viewContent.data
    property alias _interactive: vpnFlickable.interactive
 
--- a/src/ui/screens/tipsAndTricks/ViewTipsAndTricks.qml
+++ b/src/ui/screens/tipsAndTricks/ViewTipsAndTricks.qml
@@ -79,6 +79,7 @@ VPNViewBase {
             // All
             VPNViewBase {
                 objectName: 'allTab'
+                anchors.topMargin: 0
 
                 _viewContentData: ColumnLayout {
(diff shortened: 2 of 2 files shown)


nebula/ui/components/VPNViewBase.qml
```diff
@@ -15,6 +15,7 @@ Item {
 
    property string _menuTitle: ""
    property var _menuOnBackClicked
+   property var titleComponent
```
MattLichtenstein, line 18: You're still going to have the same issue with `rightMenuButton` - but anyways, all this is doing is creating an undefined variable - we need to refer to the the `titleComponent` and `rightButtonComponent` variables from  the VPNMenu component set in VPNScreenBase. We could just change 

`topMargin: _menuTitle !== "" || titleComponent || rightMenuButton ? VPNTheme.theme.menuHeight : 0`

to: 

`topMargin: _menuTitle !== "" || menu.titleComponent || menu.rightMenuButton ? VPNTheme.theme.menuHeight : 0`

but that same check is actually already done in [VPNMenu](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/nebula/ui/components/VPNMenu.qml#L32) - so really we can just say `topMargin: menu.visible ? VPNTheme.theme.menuHeight : 0`

brizental: > all this is doing is creating an undefined variable

Yes I am aware, but QML is not happy when we refer to a variable that has not been declared. Declaring it, event hough in practice it should all be the same, removes the error message.

> so really we can just say topMargin: menu.visible ? VPNTheme.theme.menuHeight : 0

This does not work as expected. I did not dig to figre out why, but for the TipsAndTricks view for example it end up giving a false positive.

You are right about the `rightButtonComponent`, I guess we did not see the error for that before because it was declared somewhere. Maybe I can just declare that one as empty varible like so:

```suggestion
   property var titleComponent
   property var rightButtonComponent
```

WDYT?

MattLichtenstein: > Yes I am aware, but JavaScript is not happy when we refer to a variable that has not been declared. The only intent i have in declaring it, is to suppress the error message.

Right, I suppose this would just be something of a bandaid solution though

> You are right about the rightButtonComponent, I guess we did not see the error for that before because it was declared somewhere. Maybe I can just declare that one as empty varible like so:

You wouldn't see it because the line was erroring on the `titleComponent` part and short circuiting 😅

> This does not work as expected. I did not dig to figure out why, but for the TipsAndTricks view for example it end up giving a false positive.

I assume you mean it didn't work visually. Probably some extra margin at the top because all of the tabs in tips and tricks utilize `VPNViewBase` and are adding an extra menu-height's worth of top margin. Try adding `anchors.topMargin: 0` to all of the `VPNViewBase` components in the tips and tricks view

brizental: That worked. Sorry for the delay here, this fell through the cracks.

MattLichtenstein: Does that mean we can remove `property var titleComponent`?

brizental:
```suggestion
```

brizental: Ah, looks like I did that already in my second commit.

brizental: ...or github just has a bug and no I haven't. Help me out here GitHub!!

brizental: Ha, you can't apply a suggestion on a resolved thread. TIL!

MattLichtenstein (approved):
🥳

PR #4455, opened 2022-09-14
Cross-Plattform Crash Reporting

strseb:
## Description

This PR adds experimental crash reporting support using sentry-native. 
The feature is disabled by default, only enabled when we're on guardian-stage. 

By Crossplattform, i mean osx and windows for now :) - Linux should be straightforward to add, i don't have a vm to test that on right now. 

Android & iOS are theoretically supported. We need to either figurre out how to get this into the qmake structure, or port those plattforms finally to cmake again like (#4301) 

##Ref 
https://mozilla-hub.atlassian.net/browse/VPN-2822

## Checklist
    
- [x] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)  -> I DID. 
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #4484, opened 2022-09-19
Cross-Plattform Crash Reporting: Add Android Support

strseb:
This adds support to report crashes & errors for android to #4455


.github/workflows/linters.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: |
           auth_header="$(git config --local --get http.https://github.com/.extraheader)"
           git submodule sync --recursive
-          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=2
```
bakulf, line 31: Why?

src/sentry/sentryadapter.cpp
```diff
@@ -14,10 +14,13 @@
 #include "logger.h"
 #include "settingsholder.h"
 #include "mozillavpn.h"
+#include "tasks/sentry/tasksentry.h"
+#include "taskscheduler.h"
 
 namespace {
 SentryAdapter* s_instance = nullptr;
 Logger logger(LOG_MAIN, "Sentry");
+void* transport_state = 0;
```
bakulf, line 23: 1. Nullptr
2. Where do you change this value?

bakulf (changes requested):

.github/workflows/linters.yaml
strseb: the sentry_native repo itself has submodules: 
https://github.com/getsentry/sentry-native/tree/master/external

PR #4487, opened 2022-09-19
Emit Glean pings on addon, message state changes

lesleyjanenorton:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -45,7 +45,6 @@ sample:
           Installed, Enabled, or Disabled
         type: string
 
-
   addon_message_state_changed:
     type: event
     lifetime: ping
@@ -73,13 +72,13 @@ sample:
         type: string
 
 
-  addon_message_cta_clicked:
+  addon_cta_clicked:
     type: event
     lifetime: ping
     send_in_pings:
       - main
     description: |
-      User clicked primary CTA in message
+      User clicked primary CTA in an addon
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1711123
     data_reviews:
@@ -87,13 +86,12 @@ sample:
     data_sensitivity:
       - interaction
     extra_keys:
-      message_id:
-        description: Message identifier
+      addon_id:
+        description: Addon identifier
         type: string
     notification_emails:
       - amarchesini@mozilla.com
     expires: never
-
 
   iap_g_no_sku_details:
     type: event
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -51,7 +51,7 @@ sample:
     send_in_pings:
       - main
     description: |
-      An addon was disabled
+      The addon message' state changed
     bugs:
       - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/1620
     data_reviews:
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -67,13 +67,10 @@ sample:
         description: |
           Addon identifier
         type: string
-      new_state:
+      message_state:
         description: |
           Received, Notified, Read, or Dismissed
         type: string
-      old_state:
-        description: |
-          Received, Notified, Read, or Dismissed
 
 
   addon_message_cta_clicked:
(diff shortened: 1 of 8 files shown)


src/addons/addon.cpp
```diff
@@ -375,6 +376,25 @@ Addon::~Addon() {
   disable();
 }
 
+void Addon::updateAddonState(State newState) {
+  m_state = newState;
```
bakulf, line 380: Before this, `if (m_state == newState) { return;}`

src/addons/addon.cpp
```diff
@@ -375,6 +376,25 @@ Addon::~Addon() {
...
+
+  SettingsHolder* settingsHolder = SettingsHolder::instance();
```
bakulf, line 382: You need to load the state in the CTORto keep m_state in sync with the real value.

src/addons/addon.h
```diff
@@ -27,7 +31,11 @@ class Addon : public QObject {
 
   static bool evaluateConditions(const QJsonObject& conditions);
 
+  enum State { Installed, Enabled, Disabled };
+  Q_ENUM(State);
+
   virtual ~Addon();
+  void updateAddonState(State newState);
```
bakulf, line 38: This is a private method.

src/addons/addonmessage.cpp
```diff
@@ -101,12 +103,18 @@ void AddonMessage::updateMessageState(State newState) {
...
-
+  QString oldStateSetting = stateMetaEnum.valueToKey(m_state);
   m_state = newState;
   emit stateChanged(m_state);
 
   SettingsHolder* settingsHolder = SettingsHolder::instance();
   settingsHolder->setAddonSetting(StateQuery(id()), newStateSetting);
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+      GleanSample::addonMessageStateChanged, {{"message_id", id()},
+                                              {"old_state", oldStateSetting},
```
bakulf, line 114: Why here you care about the old state and in the addon.cpp you don't?
I think we need to rename State to MessageState to avoid confusion

src/addons/manager/addonmanager.cpp
```diff
@@ -324,6 +324,9 @@ void AddonManager::storeAndLoadAddon(const QByteArray& addonData,
   if (!validateAndLoad(addonId, sha256, false)) {
     logger.warning() << "Unable to load the addon";
   }
+
```
bakulf, line 327: Add here a `Q_ASSERT(m_addons.contains(addonId));`

src/addons/addon.h
```diff
@@ -27,7 +31,11 @@ class Addon : public QObject {
 
   static bool evaluateConditions(const QJsonObject& conditions);
 
+  enum State { Installed, Enabled, Disabled };
```
bakulf, line 34: Maybe you can do the following:
1. Add an Unknown state and set it as default in m_state
2. In the CTOR of the addon class, you read the state from the settings. If you don't find it, call updateState(installed)
3. Remove the change in storeAndLoad

It should work. What do you think? In this way, it does not matter why and when we create the addon obj. At the first creation we set it as installed.

If we do this, we can add an assertion in updateState to see if the new state != Unknown.

bakulf (changes requested):

src/addons/addonmessage.cpp
lesleyjanenorton: Updated! I was sanity checking things in the Glean debug view but we don't need to care about the old_state.

glean/metrics.yaml
```diff
@@ -18,6 +18,81 @@
...
+    data_sensitivity:
+      - interaction
+    extra_keys:
+      addon_id:
+        description: Addon identifier
+        type: string
+    notification_emails:
+      - amarchesini@mozilla.com
+    expires: never
+
```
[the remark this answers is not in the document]
bakulf: We want to see if all the add-ons are correctly installed and enabled. It's mainly for debugging.

brizental (approved):
r+, let's do the data-review before merging though.

Also, would you mind filing a follow up for the validation of these metrics after release?

bakulf (approved):

Issue #4476, opened 2022-09-19
Mozilla VPN EU subscribers are able to access the Bundle upgrade flow


data-sync-user:
**Affected versions:**

* 2.10.0 (2.202209161150);

**Tested Platforms:**

* MacOS 11.6;

**Prerequisites:**

* Have Mozilla VPN app installed and opened;

* Have a Fx account with an **active web** subscription in **EURO** currency for Mozilla VPN;
* Be signed in;

* Have the system language / app language set to **en~~US/en~~CA**;

**Steps to reproduce:**

1. Navigate to Settings - Profile view;
1. Observe the “Upgrade” button displayed in Profile view;
1. Click on the “Upgrade” button;
1. Observe the behavior;

**Expected result:**

* For step 2: The “Upgrade” button should not be displayed for EU users in Profile view(?);

* For step 4: The user should be informed that upgrading to Bundle is not possible for EU users;

**Actual result:**

* For step 2: The “Upgrade” button IS displayed for EU users in Profile view;

* For step 4: The user is redirected to the upgrade flow. A general error message is displayed only after agreeing with the ToS/PN and clicking on the “Pay now” button;

**Notes:**

* Bundle upsell message is also displayed in the Messages screen;

* See the attached ss;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2907)
┆author: Bianca Hidecuti

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this on Mozilla VPN 2.10.0 (2.202209210215), while using MacOS. The “Upgrade” button is no longer displayed in the Profile view for EU users. 

Bundle upsell message is still displayed in the Messages screen  - Florian Zia , I assume that this is expected, as long as the user has the language set to en-US/en-CA? Thanks. 

!Screen Shot 2022-09-21 at 12.13.20.png|width=353,height=653!

data-sync-user:
➤ Bianca Hidecuti commented:

Moving this into "Cancelled" as per the following decision ( https://mozilla.slack.com/archives/C01A2L7GRL0/p1663716918154579 ).

data-sync-user:
➤ Santiago Andrigo commented:

Moving back to Cancelled

Issue #4470, opened 2022-09-16
Upsell Bundle message displayed for British users


data-sync-user:
**Mozilla VPN affected versions:**

* 2.10.0 (2.202209151741);

**Tested Platforms:**

* all;

**Prerequisites:**

* Have the Mozilla VPN app freshly installed;
* Have an active WEB subscription for Mozilla VPN;

**Steps to reproduce:**

1. Open the app and log in with the account from prerequisites;
1. Go to Settings-> System preferences-> Languages;
1. Set the en-GB language.
1. Go to Messages;
1. Observe displayed message;

**Expected results:**

* The bundle upsell message is not displayed.

**Actual results:**

* The bundle upsell message is displayed.

**Notes:**

* No bundle promotion displayed in Profile screen;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2903)
┆author: Valentina Virlics

data-sync-user:
➤ Santiago Andrigo commented:

UK accounts are almost 6% of our user base but this is a pretty bad way to start with in-app messages.
Sidenote: Is this dependent on whether the account has a subscription or not?

data-sync-user:
➤ Valentina Virlics commented:

I’ve checked the Bundle upsell message from inside the VPN client, which requires having a subscription for Mozilla VPN.

data-sync-user:
➤ Santiago Andrigo commented:

That’s what I thought but how should I interpret “Open the app and log in with an account without subscription;” in the description?

data-sync-user:
➤ Valentina Virlics commented:

That should have been: “Open the app and log in with the account from prerequisites“. My bad!

data-sync-user:
➤ Santiago Andrigo commented:

We've decided not to go forward with neither the Bundle upsell message in the Messages view nor in the Profile view.

data-sync-user:
➤ Santiago Andrigo commented:

Back to cancelled

Issue #4468, opened 2022-09-16
Bundle upsell message and “Upgrade” button not displayed on mobile for web subscribers


data-sync-user:
**Affected versions:**

* 2.10.0 (16632630);

* 2.10 (2.202209152007);

**Tested Platforms:**

* Android 10

* iOS 14.7.1;

**Prerequisites:**

* Have Mozilla VPN app installed and opened;

* Have a FxA account with an web active subscription for Mozilla VPN;

* Be signed in;

* Have the system language / app language set to en-US/en-CA

**Steps to reproduce:**

1. Navigate to Messages screen and observe the Bundle upsell message;
1. Navigate to “Profile” view and observe the “Upgrade” button;

**Expected result:**

* For step 1: “Privacy Bundle” message is displayed in the Messages screen;

* For step 2:  “Upgrade” button is displayed in the Profile view;

**Actual result:**

* For step 1: “Privacy Bundle” message is NOT displayed in the Messages screen;

* For step 2:  “Upgrade” button is NOT displayed in the Profile view;

**Notes:**

* While using the same account on desktop, both “Upgrade” button and Bundle message are displayed;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2901)
┆author: Bianca Hidecuti

data-sync-user:
➤ Bianca Hidecuti commented:

The Bundle upsell message is also not displayed or en-US/en-CA IAP subscribers.

data-sync-user:
➤ Santiago Andrigo commented:

The bundle upsell message hasn’t been implemented still: https://mozilla-hub.atlassian.net/browse/VPN-2544 ( https://mozilla-hub.atlassian.net/browse/VPN-2544|smart-link ). This can be done via the AddOns system, so it’s less of a priority (it doesn’t need to get bundled in the binary). For the sake of clarity, I’m removing this from the bug title and description.

Having said that, the upgrade button should be available in the Profile view for Web subscribers (on both desktop and mobile). I see here that it’s not showing up on Mobile and I’m not seeing it on Desktop either.

data-sync-user:
➤ Santiago Andrigo commented:

We've decided not to go forward with neither the Bundle upsell message in the Messages view nor in the Profile view.

data-sync-user:
➤ Santiago Andrigo commented:

Back to cancelled

PR #4474, opened 2022-09-18
Hide bundle upgrade if already subscribed

lesleyjanenorton:
## Description
This PR:
1.) looks through the user's list of subscriptions for a bundle subscription and
2.) hides bundle upgrade paths in ViewSubscriptionManagement if a bundle subscription exists

@flozia , @bakulf These changes were requested by PM on Thursday. You two are more familiar with the gotchas and things that can go wrong here. If this doesn't seem scary, it'd be nice to get it in but I don't want to introduce something that could cause us a lot of potential grief at this late stage. What do we think?

## Reference

[VPN-2913](https://mozilla-hub.atlassian.net/browse/VPN-2913)
## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #4489, opened 2022-09-19
Use subscription plan currency as a proxy for bundle upgrade availability

flozia:
## Description

Instead of relying on the users locale we decided to use the users subscription plan currency as a proxy for privacy bundle upgrade availability. Also only show the bundle upgrade for users with a web subscription.

## Reference

- Issue #4476
- Issue #4468
- Issue #4470

## Checklist
    
### DIFF
--- a/src/models/subscriptiondata.cpp
+++ b/src/models/subscriptiondata.cpp
@@ -147,6 +147,12 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
     return false;
   }
 
+  // Check if the privacy bundle upgrade is available. For now we use the
+  // currency as a proxy to determine if the upgrade is available for a user.
+  const QStringList bundleUpgradeWhiteList = {"USD", "CAD"};
+  m_bundleUpgradeAvailable = bundleUpgradeWhiteList.contains(m_planCurrency)
+      && m_type == SubscriptionWeb;
+
   // Payment
   QJsonObject paymentData = obj["payment"].toObject();
 
@@ -241,6 +247,7 @@ void SubscriptionData::resetData() {
   m_planBillingInterval = BillingIntervalUnknown;
   m_planAmount = 0;
   m_planCurrency.clear();
+  m_bundleUpgradeAvailable = false;
 
   m_paymentProvider.clear();
   m_creditCardBrand.clear();
--- a/src/models/subscriptiondata.cpp
+++ b/src/models/subscriptiondata.cpp
@@ -10,6 +10,7 @@
 #include "telemetry/gleansample.h"
 
 #include <QDateTime>
+#include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QJsonValue>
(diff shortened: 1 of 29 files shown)


src/models/subscriptiondata.cpp
```diff
@@ -147,6 +147,12 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
     return false;
   }
 
+  // Check if the privacy bundle upgrade is available. For now we use the
+  // currency as a proxy to determine if the upgrade is available for a user.
+  const QStringList bundleUpgradeWhiteList = {"USD", "CAD"};
+  m_bundleUpgradeAvailable = bundleUpgradeWhiteList.contains(m_planCurrency)
+      && m_type == SubscriptionWeb;
```
flozia, line 154: The bundle upgrade is only supposed to be available to users with a web subscription, right?

src/models/subscriptiondata.cpp
```diff
@@ -147,6 +147,12 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
     return false;
   }
 
+  // Check if the privacy bundle upgrade is available. For now we use the
+  // currency as a proxy to determine if the upgrade is available for a user.
+  const QStringList bundleUpgradeWhiteList = {"USD", "CAD"};
```
bakulf, line 152: 1. Static
2. Use the CTOR list{a, b c} directly. Basically remove the =

src/models/subscriptiondata.cpp
```diff
@@ -147,6 +147,12 @@ bool SubscriptionData::fromJson(const QByteArray& json) {
...
+  m_bundleUpgradeAvailable = bundleUpgradeWhiteList.contains(m_planCurrency) &&
+                             m_type == SubscriptionWeb;
```
bakulf, line 154: Can you invert the conditions? First a boolean check and then the list?

src/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
```diff
@@ -159,7 +159,9 @@ VPNViewBase {
                     VPNSubscriptionData.planCurrency,
                     VPNSubscriptionData.planAmount,
                 ),
-                type: (Qt.locale().name === "en_US" || Qt.locale().name === "en_CA" ) && VPNFeatureList.get("bundleUpgrade").isSupported ? "text-upgrade" : "text",
+                type: VPNSubscriptionData.bundleUpgradeAvailable && VPNFeatureList.get("bundleUpgrade").isSupported
```
bakulf, line 162: This is confusing. Why part of the conditions are in c++ and part in Js? I would do everything here.

bakulf (changes requested):

src/models/subscriptiondata.cpp
lesleyjanenorton: Correct!

src/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
flozia: Since we also introduce a check to see if a user is already a bundle subscriber with PR #4474 I thought it might better idea to not move this logic to QML. Are you also OK with moving `VPNFeatureList.get("bundleUpgrade").isSupported` to the C++ layer instead of moving everything to QML?

src/models/subscriptiondata.cpp
flozia: Sure!

src/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
flozia: Moved the check to QML.

bakulf (approved):

PR #4493, opened 2022-09-20
Add Time to main Metric

strseb:
RE: [VPN-2916](https://mozilla-hub.atlassian.net/browse/VPN-2916)

This measures time between calling main and finishing the load of the Main view. 

Still needs a data-review tho :)

### DIFF
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -1538,4 +1538,26 @@ sample:
           The source of the device removal information.
           Either 'TaskRemoveDevice' or 'PushMessage'.
         type: string
-# End of push messages validation metrics
+  # End of push messages validation metrics
+
+  perf_time_to_main_screen:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      The time the app took between starting and rendering the main screen
+    bugs:
+      - https://github.com/mozilla-mobile/mozilla-vpn-client/issues/4486
+    data_reviews:
+      - TODO
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - sstreich@mozilla.com
+    expires: 2023-12-31
+    extra_keys:
+      duration:
+        description: |
+          The Millisecounds this it took for the App to Load the main screen.
+        type: string
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -1557,7 +1557,11 @@ sample:
       - sstreich@mozilla.com
     expires: 2023-12-31
     extra_keys:
-      duration:
+      start:
         description: |
-          The Millisecounds this it took for the App to Load the main screen.
+          Timestamp of the Start
+        type: string
+      end:
+        description: |
+          Timestamp of the end
         type: string
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,11 +5,14 @@
 #include "commandlineparser.h"
 #include "leakdetector.h"
 
+#include "telemetry.h"
+
 int main(int argc, char* argv[]) {
 #ifdef MVPN_DEBUG
   LeakDetector leakDetector;
   Q_UNUSED(leakDetector);
 #endif
+  Telemetry::recordAppStartTimestamp();
 
   CommandLineParser clp;
   return clp.parse(argc, argv);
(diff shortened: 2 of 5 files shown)


glean/metrics.yaml
```diff
@@ -1538,4 +1538,26 @@ sample:
...
-# End of push messages validation metrics
+  # End of push messages validation metrics
+
+  perf_time_to_main_screen:
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - main
+    description: |
+      The time the app took between starting and rendering the main screen
```
brizental, line 1549: I think we would need to instrument this on all pages, not just the main page. That bug seemes to happen in random pages at random times.

src/main.cpp
```diff
@@ -5,11 +5,14 @@
...
 #include "leakdetector.h"
 
+#include "telemetry.h"
+
 int main(int argc, char* argv[]) {
 #ifdef MVPN_DEBUG
   LeakDetector leakDetector;
   Q_UNUSED(leakDetector);
 #endif
+  Telemetry::recordAppStartTimestamp();
```
brizental, line 15: We can have two events here, "main screen started loading" / "main screen lof complete" instead of calculating the time ourselves. Then we record the first event here and the other even later and let Glean do the timestamp calculations. WDYT?

brizental (changes requested):

glean/metrics.yaml
strseb: While I agree in general, I'm concerned that generating events for every navigation is more on the creepy scale. Given we know this also happens on load, I would prefer we have one probe and see what we are working with.

src/main.cpp
strseb: My problem in general would rather be, can we record glean events before glean is initialised or what do you think about 2 events that have a "timestamp" extra ?

brizental: Events have timestamps by default, so the extra is not necessary. And yes, we can record Glean events before init, they will just be sent after init. There is a limit for how many events can be recorded before init, but that is very high and will not be hit here.

glean/metrics.yaml
brizental: Hm, I am not sure I would agree it's "creepy"... I guess since we are using events only here that might feel that way because it kinda feel like following the user everywhere, however we do have a bunch of "open this" "click that" events that are basically exactly that already. 

Well, no strong opinions. We can go with just this for and see what we can get out of it at analysis time.

src/main.cpp
strseb: > And yes, we can record Glean events before init, they will just be sent after init. 

ie. during "Telemetry::recordAppStartTimestamp();" - Neither the QJs engine is created nor the android daemon started yet - so i mean glean code has not yet been loaded, which can store the event for later. That's why i'd use an extra timestamp.

brizental: Sheesh, I see... Ok yeah I don't see how that would work in this case. Well, then a single event with the two timestamps as extras should suffice no? The Glean event timestamps are relative to the first event recorded on each app run, they are not absolute. So if we want to compare our two timestamps we would need to calculate them both ourselves. 

Ugh, that's messy for sure.

src/telemetry.cpp
```diff
@@ -116,3 +117,27 @@ void Telemetry::periodicStateRecorder() {
...
+
+  emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
+      GleanSample::perfTimeToMainScreen,
+      {
+          {"start", QString::number(s_appStart.toMSecsSinceEpoch())},
+          {"end", QString::number(now.toMSecsSinceEpoch())},
+      });
+  logger.info() << "Time to load up the App:" << s_appStart.msecsTo(now)
+                << "ms";
+}
```
brizental, line 143: Missing newline.

src/telemetry.cpp
```diff
@@ -13,6 +13,7 @@ constexpr int CONNECTION_STABILITY_MSEC = 45000;
 
 namespace {
 Logger logger(LOG_MAIN, "Telemetry");
+QDateTime s_appStart = QDateTime();
```
bakulf, line 16: you don't need to add `= QDateTime()`.

src/telemetry.h
```diff
@@ -15,6 +15,9 @@ class Telemetry final : public QObject {
 
   void initialize();
 
+  static void recordAppStartTimestamp();
+  void recordAppLoadedTimeStamp();
```
bakulf, line 19: static? at least this is what the C++ code says.

bakulf (changes requested):

PR #4502, opened 2022-09-20
Use DNS ping latency only when enabled - VPN-2647

bakulf:
I don't know if this is the reason of this bug. But for sure it sounds like a bug the use of m_dnsPingLatency if not initialized.


oskirby (approved):
I don't know if this would be the cause of the bugs you mention either, but it certainly looks a little fragile and suspicious without this check.

PR #4513, opened 2022-09-21
Remove bundle upsell message and UI

lesleyjanenorton:
## Description

This PR removes...
- The privacy bundle in-app message
- The upgrade paths in ViewSubscriptionManagement and related tidbits

## Reference

https://mozilla.slack.com/archives/C0291CWG53K/p1663716976010969

## Checklist
    

bakulf (commented):
I would like to postpone this after 2.11. It's a big change and I don't think we want to land it at this point. Plus maybe we can use this code to run A/B tests.

lesleyjanenorton:
@bakulf sounds good! Will close this for now in that case.

PR #4341, opened 2022-09-06
Add addons beetmovers, create relpro action, upgrade taskgraph, and run taskgraph validation on TC

gabrielBusta:
## Description

- Adds beetmover tasks for the addons
- Adds a release promotion action to trigger beetmover tasks
- There are three `release-promotion` flavors available in the release promotion action
    - `promote-addons` uploads the addons to [`/pub/vpn/addons/candidates/<moz_build_date>/`](https://ftp.stage.mozaws.net/pub/vpn/addons/candidates/20220919220000/)
    - `ship-addons` uploads the addons to [`/pub/vpn/addons/releases/<moz_build_date>/`](https://ftp.stage.mozaws.net/pub/vpn/addons/releases/20220919220000/) and [`/pub/vpn/addons/releases/lastest/`](https://ftp.stage.mozaws.net/pub/vpn/addons/releases/latest/)
    - `promote-client` uploads the client to the candidates directory [`/pub/vpn/candidates/<app_version>-candidates/build<moz_build_date>/<build_os>/`](https://ftp.stage.mozaws.net/pub/vpn/candidates/0.0.2-candidates/build20220919200551/)
    - `ship-client` copies the client from the candidates directory to [`/pub/vpn/releases/<app_version>/<build_os>/`](https://ftp.stage.mozaws.net/pub/vpn/releases/0.0.2/)
- Major version bump to Taskgraph v3.1.0
- Migrates the `Test Taskgraph Definition` job to a Taskcluster task

## Reference

https://mozilla-hub.atlassian.net/browse/RELENG-969
https://mozilla-hub.atlassian.net/browse/RELENG-981

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #4531, opened 2022-09-23
Remove beetmover release triggers

gabrielBusta:
## Description

Stop uploading `release` graph client builds to the Mozilla archive automatically.

cc @bakulf @ahal 

## Reference

#4341 

## Checklist
    
### DIFF
--- a/taskcluster/ci/beetmover/kind.yml
+++ b/taskcluster/ci/beetmover/kind.yml
@@ -20,7 +20,7 @@ tasks:
         worker:
             chain-of-trust: true
             max-run-time: 1800
-        run-on-tasks-for: [github-release, action:release-promotion]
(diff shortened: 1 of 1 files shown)


taskcluster/ci/beetmover/kind.yml
```diff
@@ -37,7 +37,7 @@ tasks:
         worker:
             chain-of-trust: true
             max-run-time: 1800
-        run-on-tasks-for: [github-release, action:release-promotion]
```
ahal, line 40: Technically we could do leave the promote phase to be triggered on release creation, and then we'd only have a `ship-client` phase in the action. But that's a bit clunky.. Just calling it out in case the VPN team would prefer that though.

PR #4540, opened 2022-09-30
Wasm_Chrome: Replace Select with a combo box

strseb:
This Pr replaces the old select with a combo-box from fluent. So we have better keyboard support and autofill (!) on the page. Also moved the buttons to a small menu :) 

image


flozia (approved):
Nice! :)

Issue #4472, opened 2022-09-16
When VPN turned On, device loses internet connection on iOS 16


JKingZapata:
After updating to iOS 16, users connecting to VPN get a "No signal" message and device loses internet connection. Changing server location, reinstalling, port53, and resetting device network settings do not help. Internet connection is restored after VPN is turned Off.

Does it impact functionality or aesthetics? Yes
Does this endanger users? What would be compromised and how likely is the threat? No
How annoying is this for affected users? (Think about how much 'in the way' this is, relative to what the user is trying to do) Very annoying
Does it degrade the user experience significantly? Yes
Is it legally sensitive? No

**Centrality** 
Core functionality. 

**Technical Information**
Platform iOS 16
FX product Mozilla VPN 2.9
Region/Languages
Network type (Mobile/Wifi/Wired): Both

**Steps to Reproduce**

1. Purchase a subscription
1. Install VPN
1. Complete setup
1. Connect to server location

**Actual Results**
After VPN connects, "No Signal" message appears and user cannot browse the web or use any apps that requires internet

**Expected Results**
VPN should show "VPN is On, secure and private" and the device continues to have internet access

[MozillaVPN-logs-14.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/9587032/MozillaVPN-logs-14.txt)

![width=200,height=183](https://user-images.githubusercontent.com/72564534/190725697-2d962a4b-df2c-406c-a754-1957592117b5.jpeg)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2905)

mdsfau:
Seeing the same exact issue. Totally broken on iOS 16.

Austionian:
I'm also seeing this issue.

mthomasuk:
me too

data-sync-user:
➤ Santiago Andrigo commented:

Joseph Anthony Cuevas is reporting that this is a growing problem with iOS 16, and that it’s beginning to show up even in App Store ratings.

data-sync-user:
➤ Sarah Bird commented:

I have been experiencing exactly the same symptoms with exactly the same steps to reproduce on my iPad for a very long time…..like at least 9 months.

[...]

PR #4547, opened 2022-09-30
Update stale copy of ios/gobridge

oskirby:
## Description
To work correctly with iOS 16, we needed to update the `wireguard-apple` submodule, which we did in f97739eff4a3766fc00e14709a150d37467c5f90, however, we didn't notice a vendored copy of the iOS gobridge which also needed to be updated. This resulted in some old code that failed to create the wireguard listener sockets correctly.

## Reference
Github #4472 ([VPN-2905](https://mozilla-hub.atlassian.net/browse/VPN-2905))

## Checklist


ios/gobridge/go.mod
```diff
@@ -1,8 +1,10 @@
 module golang.zx2c4.com/wireguard/apple
 
-go 1.16
+go 1.17
```
bakulf, line 3: Are we ok with this change in all the platforms?

bakulf (approved):
We need to set up a dependency bot ... I filed a bug. We should prioritize it.

ios/gobridge/go.mod
oskirby: This module is only used on iOS, so the remaining platforms should continue to be able to use go 1.16.

PR #4572, opened 2022-10-04
Deprecate the message bundle addon - VPN-2979

bakulf:


lesleyjanenorton (approved):
Do we want to add `"env": "staging"` to the manifest to match the addon's name, in case we revive this at some point?

PR #4574, opened 2022-10-05
Cleanup ios NetworkExtension

bakulf:
We don't need to have our own go-bridge implementation. We can use the wireguardKitGo one instead.
The code is exactly the same plus 1 PR: https://github.com/WireGuard/wireguard-apple/pull/15


oskirby (approved):
Yes, I think this is a much cleaner approach, and it should prevent us from letting our gobridge fall out of date in the future.

PR #4578, opened 2022-10-05
Use a timer to check the status when connected - ios

bakulf:


oskirby (approved):
LTGM.

We could make it a little tighter by checking that `last_handshake_time_nsec` which will prevent a false-negative from occurring if the connection is established within 1 second of of the epoch (which is extremely unlikely).

Issue #4569, opened 2022-10-04
“Unable to connect. Try again” red banner displayed when connecting to some servers on Android


data-sync-user:
**Affected versions:**

* 2.10.0 (16648200);

**Tested Platforms:**

* Android 10;

**Prerequisites:**

* Have Mozilla VPN app installed and opened;

* Have a Fx account with an active subscription for Mozilla VPN;
* Be signed in;

**Steps to reproduce:**

1. Navigate to “Select location” screen;
1. Select one of the following servers: Latvia - Riga / Moldova - Chisinau / North Macedonia - Skopje;
1. Turn ON the VPN;
1. Observe the behavior;

**Expected result:**

* The VPN is successfully turned ON;

**Actual result:**

* “Unable to connect. Try again” red banner is displayed. The VPN is not turned ON;

**Notes:**

* After swipe-killing the app and reopening it, the VPN is connected;

* Not reproducing on other platforms;

* See the attached video and logs;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2978)
┆author: Bianca Hidecuti

data-sync-user:
➤ Santiago Andrigo commented:

Bianca Hidecuti When you say ‘some servers’, do you mean random ones or the ones you specifically listed here?

data-sync-user:
➤ Bianca Hidecuti commented:

Santiago Andrigo , the ones specifically listed in the bug description: Latvia - Riga / Moldova - Chisinau / North Macedonia - Skopje

data-sync-user:
➤ Owen Kirby commented:

All of these locations are unique in that they tend to have very few servers (Skopje, Macedona only has one server), and we seem to be getting an exception in the VPN service at these locations as follows:

{{[2022-10-04T16:45:58.371] [info] - (VPNService) - Dispatch Daemon State -> connected  }}
{{[2022-10-04T16:45:58.620] [error] - (VPNServiceBinder) - An Error occurred while enabling the VPN: No value for serverFallback  }}
{{[2022-10-04T16:45:58.621] [error] - (VPNServiceBinder) - StackTrace:  }}
{{[2022-10-04T16:45:58.621] [error] - (VPNServiceBinder) - 	 org.json.JSONObject.get(JSONObject.java:399)  }}
[...]

PR #4581, opened 2022-10-06
Fix android exception when no fallback server is present.

oskirby:
## Description
For some locations, it is possible that we might find ourselves without any fallback servers (eg: Skopje, Macedonia has only a single server). When trying to connect to such a location, we encounter an exception in `VPNService.kt` when parsing the connection parameters from JSON.

To fix this, we attempt to check for the existence of `serverFallback` before attempting to parse it.

## Reference
Github #4569 ([VPN-2978](https://mozilla-hub.atlassian.net/browse/VPN-2978))

## Checklist


lesleyjanenorton (approved):
Approving this so that we can get it over to QA, since Basti is on PTO for the next few weeks.

Issue #4587, opened 2022-10-06
[Windows] Mozilla VPN not responding while trying to force the update required


data-sync-user:
**Mozilla VPN affected versions:**

* 2.10.0 (2.202210052004);

**Platform tested:**

* Windows 10/11;
* MacOS;
* Linux;

**Prerequisites:**

* Be logged in with an account with subscription for Mozilla VPN;
* Be in Stage;

**Steps to reproduce:**

1. Open the [+Inspector+](https://mozilla-mobile.github.io/mozilla-vpn-client/inspector/);
1. Run “force_update_check 1.5”;
1. Click anywhere inside the Mozilla VPN app;

**Expected results:**

* An “update required” screen is displayed;

**Actual results:**

* App is blocked into “Not responding” state;

**Notes:**

* On macOS the “updated_required” screen is not displayed but the app it is not blocked in not responding state;
* On Linux the “updated_required” screen is triggered;

[^mozillavpn-2022-10-6_1.txt]

!notresp.png|width=294,height=509!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2987)
┆author: Valentina Virlics

data-sync-user:
➤ Matt Lichtenstein commented:

Contrary to what the notes say, I am experiencing a not-responding state on macOS (12.5.1) after running force_update_check 1.5 from the inspector

Version: 2.10.0 (2.202210061009)

data-sync-user:
➤ Matt Lichtenstein commented:

Only seeing this on my local macOS builds and not CI builds… issue may still be Windows-specific

data-sync-user:
➤ Andrea Marchesini commented:

This issue is about how we compile the balrog go module. Owen Kirby changed it in 2.10 for the CMake migration task and it has not been tested yet.

Somehow the balrog module is incompatible with the rest of the C++ app and when we try to execute it, the app crashes.

data-sync-user:
➤ Bianca Hidecuti commented:

After further investigation, we noticed that this has started to reproduce with the first 2.10 build - 2.10.0 (2.202207261255). 

On the 2.9 version, the issue is not reproducible and the “Update required” screen is displayed.

data-sync-user:
➤ Valentina Virlics commented:

Verified as fixed on Windows 10, while using the last 2.11.0 (2.202210090428) build from main branch.

[...]

PR #4600, opened 2022-10-07
Convert windows balrog library back to a DLL

oskirby:
## Description
During the update to CMake, we refactored the build of the balrog library to be linked statically instead of importing a DLL. Unfortunately, this appears to result in breakage during update on the Windows platform. To resolve the problem, let's just go back to using a DLL for this purpose.

Unfortunately, statically linking between golang and the MSVC toolchain is badly supported and this is not something likely to be fixed any time soon.

## Reference
Github #4587 [VPN-2987](https://mozilla-hub.atlassian.net/browse/VPN-2987)

## Checklist


src/update/balrog.cpp
```diff
@@ -25,10 +25,8 @@
 #  include "windows.h"
 #  include "platforms/windows/windowscommons.h"
 #  include "platforms/windows/golang-msvc-types.h"
```
bakulf, line 27: Do we still need this?

src/update/balrog.cpp
```diff
@@ -25,10 +25,8 @@
...
-#endif
-
+#else
 // Import balrog C/Go library (unless we are building on Windows with qmake)
-#if !(defined(MVPN_WINDOWS) && defined(BUILD_QMAKE))
```
bakulf, line 31: Update the comment above

src/cmake/golang.cmake
```diff
@@ -46,17 +54,15 @@ function(add_go_library GOTARGET SOURCE)
...
     add_dependencies(${GOTARGET} golang_${GOTARGET})
     set_target_properties(${GOTARGET} PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_CURRENT_BINARY_DIR}
         INTERFACE_SOURCES ${CMAKE_CURRENT_BINARY_DIR}/${HEADER_NAME}
         IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/${ARCHIVE_NAME})
-
-    if(MSVC)
-        # prevent error LNK2019: unresolved external symbol fprintf referenced in function ...
-        set_property(TARGET ${GOTARGET} APPEND PROPERTY
-            INTERFACE_SOURCES ${CMAKE_SOURCE_DIR}/src/platforms/windows/golang-msvc-fixup.cpp)
```
bakulf, line 59: Can we remove this file?

bakulf (approved):

src/cmake/golang.cmake
oskirby: Yes, we can remove it.

src/update/balrog.cpp
oskirby: Yes, we still need to define the C/golang types manually, since the C/go compiler doesn't produce an importable header file on the MSVC toolchain.

PR #4604, opened 2022-10-10
Better error propagation

bakulf:


src/errorhandler.h
```diff
@@ -48,6 +48,14 @@ class ErrorHandler final : public QObject {
   };
   Q_ENUM(AlertType)
 
+  // This can be an boolean but it's better to have an enum to improve the
+  // readability of the code.
+  enum ErrorPropagation {
+    NoErrorPropagation,
+    PropagateError,
+  };
+  Q_ENUM(ErrorPropagation);
```
brizental, line 57: It's weird to me that this enum is defined on the error handler when in fact its being used elsewhere in the code. AFAICT it's only used in the Tasks, what about we move this declaration there?

src/errorhandler.h
```diff
@@ -48,6 +48,14 @@ class ErrorHandler final : public QObject {
   };
   Q_ENUM(AlertType)
 
+  // This can be an boolean but it's better to have an enum to improve the
+  // readability of the code.
+  enum ErrorPropagation {
+    NoErrorPropagation,
+    PropagateError,
```
brizental, line 55: This is kinda nitpicky, but how about `DoNotPropagateError` / `PropagateError`?

src/tasks/account/taskaccount.cpp
```diff
@@ -25,8 +25,15 @@ void TaskAccount::run() {
...
       request, &NetworkRequest::requestFailed, this,
       [this](QNetworkReply::NetworkError error, const QByteArray&) {
-        logger.error() << "Account request failed" << error;
-        ErrorHandler::instance()->errorHandle(ErrorHandler::toErrorType(error));
+        logger.error() << "Account request failed" << error
+                       << "error-propagation:"
+                       << (m_errorPropagation == ErrorHandler::PropagateError
+                               ? "yes"
+                               : "no");
+        if (m_errorPropagation == ErrorHandler::PropagateError) {
```
brizental, line 33: I am not a big fan of re-declaring this conditional in each task, it makes it easier for the meaning or "ErrorPropagation" to be changed in sutil ways as new tasks are implemented or changed. What if we declare a function on the Task super class like so:

```cpp
void handleTaskError(Error error) {
   if (m_errorPropagation == ErrorHandler::PropagateError) {
       ErrorHandler::instance()->errorHandle(
              ErrorHandler::toErrorType(error));
   }
}
```

brizental (changes requested):

src/releasemonitor.cpp
```diff
@@ -1,72 +1,75 @@
...
-    connect(task, &Task::completed, [] {
-      MozillaVPN* vpn = MozillaVPN::instance();
-      Q_ASSERT(vpn);
-      vpn->setUpdating(false);
-    });
-
-    TaskScheduler::scheduleTask(task);
-  });
-}
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
bakulf: because the file was encoded using CR+LF (dos/windows)...

src/errorhandler.h
bakulf: I thought about that, but why tasks? we can have the same enum in every single class or implement it centralized. ErrorHandler seems to be the right choice. Balrog is not a task and still need the ErrorHandler enum

brizental: Ah! I missed Balrog. Ok yeah, that makes it tricker.

src/errorhandler.cpp
```diff
@@ -226,3 +226,12 @@ void ErrorHandler::setAlert(AlertType alert) {
   m_alert = alert;
   emit alertChanged();
 }
+
+// static
+void ErrorHandler::networkErrorHandle(
```
brizental, line 231: nit: `handleNetworkError`? `networkErrorHandle` sounds like the function is going to return a handle to something.

src/errorhandler.h
brizental: Ah, the new commit changes thing in a way that it makes complete sense for the enum to be on the ErrorHandler class.

brizental (approved):
r+wc

src/errorhandler.cpp
brizental: Oh I see that `errorHandle` also has "handle" at the end. IMO it also sounds like that function returns a handle to something... But we don't need to make this PR about renaming a bunch of things.

dlmchugh1212:
please stop sending

PR #4609, opened 2022-10-10
Fix a wrong connection between addonMessage and its composer - VPN-2997

bakulf:


src/addons/addonmessage.cpp
```diff
@@ -66,16 +66,17 @@ Addon* AddonMessage::create(QObject* parent, const QString& manifestFileName,
...
+
   return message;
 }
 
 AddonMessage::AddonMessage(QObject* parent, const QString& manifestFileName,
                            const QString& id, const QString& name)
     : Addon(parent, manifestFileName, id, name, "message") {
   MVPN_COUNT_CTOR(AddonMessage);
-
-  connect(this, &Addon::retranslationCompleted, m_composer,
```
bakulf, line 77: at this point, m_composer is still null.

brizental (approved):

PR #4623, opened 2022-10-12
VPN-2934 - Fix device limit issues when logging out manually from other device

brizental:


src/mozillavpn.cpp
```diff
@@ -352,13 +352,18 @@ void MozillaVPN::setState(State state) {
   emit MozillaVPN::instance()->recordGleanEventWithExtraKeys(
       GleanSample::appStep, {{"state", QVariant::fromValue(state).toString()}});
 
-  // If we are activating the app, let's initialize the controller.
-  if (m_state == StateMain) {
-    m_private->m_controller.initialize();
+  // If we have a token, we can start periodic operations.
+  // If the timer is already started, this is a no-op.
+  if (SettingsHolder::instance()->hasToken()) {
```
brizental, line 357: Ok, so theoretically this should not be necessary because the client should be notified through the websocket connection that a device has been deleted. Since that has not been validated yet (and in my local testing it failed intermittently to received the message) I will leave this.

bakulf (approved):

PR #4630, opened 2022-10-13
Cache the conditionEnabled status to avoid race conditions - VPN-3006

bakulf:
This is a nasty bug. The STR/bug is the following:

1. using the jira STR, we have only 2 addons with a conditional-watcher locale set to 'it'
2. set the locale to 'ru'; the addons are disabled. AddonManager has 0 active addons.
3. change the language to 'it'
4. the conditional watcher of addon A emits `conditionChanged(true)`
5. the addonManager adds the addon in the list.. but to do this, QT calls `rowCount` which calls, for each addon `->enabled()`, which calls `addon->m_conditionalWatcher->conditionApplied()` which returns `true` for both addons. `rowCount` _before_ adding the first addon return 2 - the bug.
6. QT thinks there are 2 addons and we are adding a 3rd one.
7. in the log we see an error message about an invalid QModelIndex
8. In the frontend code we use a FilterableProxyModel which calls `lessThan` for all the 3(?) addons
9. when we compare the second one with the 3rd one(?), the `QModelIndex` of the 3rd one is invalid
10. the frontend code we do: `_sortProxyCallback: (obj1, obj2) => obj1.addon.date > obj2.addon.date` and the code throws an exception because `date` is not a property of `undefined`.

So, the fix is to do not call `conditionApplied()` in `enabled` but to use a cached value. The code will process 1 signal/slot at the time.

### DIFF
--- a/src/addons/manager/addonmanager.cpp
+++ b/src/addons/manager/addonmanager.cpp
@@ -174,11 +174,9 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
       }
       if (!enabled) {
         beginRemoveRows(QModelIndex(), pos, pos);
-        removeRow(pos);
         endRemoveRows();
       } else {
         beginInsertRows(QModelIndex(), pos, pos);
-        insertRow(pos);
         endInsertRows();
       }
       break;
--- a/src/addons/manager/addonmanager.cpp
+++ b/src/addons/manager/addonmanager.cpp
@@ -89,6 +89,11 @@ void AddonManager::updateIndex(const QByteArray& index,
 void AddonManager::updateAddonsList(QList<AddonData> addons) {
   logger.debug() << "Updating addons list";
 
+  // Maybe this is not the first time we load the addon list. So, technically,
+  // the addonManager has already complete the first addon loading. No need to
+  // reset m_loadCompleted.
+  bool loadCompleted = false;
+
   // Remove unknown addons
   QStringList addonsToBeRemoved;
   for (QMap<QString, AddonData>::const_iterator i(m_addons.constBegin());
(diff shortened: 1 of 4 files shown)


bakulf:
I also fixed VPN-3005. See the comment in AddonManager.cpp.

src/addons/manager/addonmanager.cpp
```diff
@@ -174,11 +174,9 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
       }
       if (!enabled) {
         beginRemoveRows(QModelIndex(), pos, pos);
-        removeRow(pos);
         endRemoveRows();
       } else {
         beginInsertRows(QModelIndex(), pos, pos);
-        insertRow(pos);
```
brizental, line 181: Why remove these two lines?

brizental (commented):
The changes look good to me. I have just one question. 

Would you mind filing a follow up for the refactor work we discussed on Slack for the addon manager lists?

bakulf:
About refactoring the m_addons list in 2, sure. I think it can cleanup the implementation. Let's do it as a follow up.

src/addons/manager/addonmanager.cpp
brizental: But then what is the point of calling begin/end remove/insert rows if these functions are not being called anymore?

src/addons/manager/addonmanager.cpp
```diff
@@ -174,11 +172,9 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
       }
       if (!enabled) {
```
brizental, line 173:
```suggestion
      // Note that `removeRow` and `insertRow` are not called between the `begin`/`end` calls below.
      // Those are not necessary, because the modelling of the addons list in this class is different from what
      // we need to convey to QML i.e. in here the addons list contails all addons while QML needs
      // enabled addons only. See the `AddonManager::rowCount` and `AddonManager::data` for more clarity.
      //
      // The `begin`/`end` calls are still necessary though to trigger specific signals for QML to
      // know that the list has changed and update the UI accordingly.
      if (!enabled) {
```

brizental (approved):

PR #4633, opened 2022-10-14
Drop support for qt5

bakulf:


oskirby (approved):
Can we add a very loud error in the top-level `CMakeLists.txt` that refuses to build if someone tries to build with Qt5? I anticipate we are likely to get bug reports filed from users on older Linux distributions who don't notice our required Qt version.

PR #4642, opened 2022-10-14
Static analysis

bakulf:
I used clazy to run a bit of static analysis of the code. This is a good chunk of issues.


oskirby (approved):
Looks good

Issue #1668, opened 2021-08-31
Add accessible name to recent connection buttons


lesleyjanenorton:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/336455944346cc9db1b4ea97592a863b29d77f53/src/ui/components/VPNRecentConnections.qml#L105

We need to add an accessibleName to recent connection buttons that changes based on whether the connection is single or multi-hop.

PR #4677, opened 2022-10-19
Add accessible label for recent connections

flozia:
## Description

Add an accessible label to the recent connections.

## Reference

- #1668

## Checklist
    

nebula/ui/components/VPNRecentConnections.qml
```diff
@@ -80,13 +80,22 @@ ColumnLayout {
...
+                            countryCode: countryCode,
+                            serverCityName: serverCityName,
+                            localizedCityName: VPNLocalizer.localizedCityName(countryCode, serverCityName)
+                        });
                     }
 
-                    append({isMultiHop, connection });
+                    const [{ firstCityLocalizedName: localizedCityName }, secondServer] = connection;
+                    const accessibleLabel = secondServer
+                        ? `${firstCityLocalizedName} ${VPNl18n.MultiHopFeatureMultiHopArrowAltText} ${secondServer.localizedCityName}`
```
bakulf, line 91: @flodolo what do you think about this string? We can also have a "$1 $2" in strings.yaml and let the translator decide what is better.

bakulf (approved):

nebula/ui/components/VPNRecentConnections.qml
flodolo: Sorry, I don't understand what the ask is: are you suggesting to make this label localizable, e.g. `$1 to $2" or something else?

flozia: Happy to make the string `$1 to $2` if we run the risk that this does not translate correctly in all cases.

flodolo: Let's make that, thanks 👍🏼

flozia: Done!

translations/strings.yaml
```diff
@@ -137,6 +137,9 @@ multiHopFeature:
     value: To
     comment: This is alt text for an arrow showing the multi-hop route between servers. e.g. Atlanta TO Copenhagen
   serverSearchResultsError: No results. Try a different search
+  accessibleNameRecentConnection:
+    value: "%1 to %2"
+    comment: Accessible label that is used for recent multi-hop server connections. “%1” reprensents the entry and “%2” the exit server.
```
flodolo, line 142:
```suggestion
    comment: Accessible label that is used for recent multi-hop server connections. “%1” represents the entry server and “%2” the exit server.
```

flodolo (approved):

Issue #4097, opened 2022-07-29
Rating pictograms from “Give feedback” screen have issues


data-sync-user:
**Affected versions:**

* 2.10.0 (2.202207282101);

**Tested Platforms:**

* MacOS 11.6;

**Prerequisites:**

* Have a FxA account with an active subscription for Mozilla VPN;
* Have Mozilla VPN installed and opened;
* Be signed in;

**Steps to reproduce:**

1. Navigate to Settings - Get help - Give feedback screen;
1. Observe the “Give feedback” screen;
1. Hover over each rating pictogram and observe the behavior;
1. Click on any rating pictogram from the screen;

**Expected result:**

* For step 2: Each rating pictogram is displayed in gray,  as no rating is selected.
* For step 3: Each rating pictogram changes its color from gray to blue when hovering over. 
* For step 4: The rating pictogram changes its color from gray to blue and is outlined in a lighter shade of blue when clicking on it. 

**Actual result:**

* For step 2: The first rating pictogram (“very poor”) is displayed in blue by default, without clicking on it. 
* For step 3: The rating pictograms do not change their color from gray to blue when hovering over. 
* For step 4: The rating pictogram does not change its color from gray to blue when clicking on it, but is outlined in a lighter shade of blue. 

**Notes:**

* See the attached video;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2624)
┆Reporter: Bianca Hidecuti

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this on Mozilla VPN 2.10.0 (2.202209120606), while using MacOS 11.6, and the following AR is no longer reproducing: “_For step 2: The first rating pictogram (“very poor”) is displayed in blue by default, without clicking on it.”_

However, rating pictograms still do not change their color from gray to blue when hovering over and also when clicking on a rating pictogram the color is still not changing from gray to blue  - it is outlined in a lighter shade of blue. 

Attaching a video as well.

!Screen Recording 2022-09-12 at 12.41.57.mov|width=716,height=1308!

data-sync-user:
➤ Bianca Hidecuti commented:

When being in the “Give feedback” screen, while drag clicking any rating pictogram, the “Continue” button remains disabled, even though the rating pictograms are displayed as selected.

Attaching a video as well.

!bandicam 2022-09-27 09-01-36-604.mp4|width=448,height=812!

[...]

PR #4684, opened 2022-10-19
Fix user feedback icon ui states

flozia:
## Description

Fixes the icon ui states in the “Give feedback” form.

https://user-images.githubusercontent.com/13835474/198679646-85613c5f-041a-4897-8b9c-53ee8446c691.mov

## Reference

- #4097

## Checklist
    

MattLichtenstein:
Everything looks good, just one question

In #4097, the steps to reproduce and expected results say: 

**Step to repro**: Hover over each rating pictogram and observe the behavior;
**Expected result**: Each rating pictogram changes its color from gray to blue when hovering over. 

which appears to be the opposite of what is implemented. Implementation has a gray icon with blue outline on hover. Just trying to figure out if the expected results or the implementation is correct

flozia:
> Everything looks good, just one thought
> 
> In #4097, the steps to reproduce and expected results say:
> 
> **Step to repro**: Hover over each rating pictogram and observe the behavior; **Expected result**: Each rating pictogram changes its color from gray to blue when hovering over.
> 
> which appears to be the opposite of what is implemented. Implementation has a gray icon with blue outline on hover. Just trying to figure out if the expected results or the implementation is correct

Good point. Updated the icon to be blue on hover and only have an outline when selected and added a screencast to the description. Also checking with you @peterbenvenuto just to make sure this is the correct behaviour.

PR #4695, opened 2022-10-20
VPN-3020 - Remove unused VPN.currentView

brizental:


nebula/ui/components/VPNStackView.qml
```diff
@@ -13,7 +13,6 @@ StackView {
 
     onCurrentItemChanged: {
         var objString = currentItem.toString().split("(")[0];
```
bakulf, line 15: You can remove the entire function

bakulf (changes requested):

bakulf (approved):

Issue #4650, opened 2022-10-17
[Windows] Mozilla VPN cannot be turned ON after an update


data-sync-user:
**Affected versions:**

* 2.10.0 (2.202210111527);

**Tested Platforms:**

* Windows 10/11;

**Prerequisites:**

* Have **2.9** Mozilla VPN app installed and opened;
* Have a Fx account with an active

subscription for Mozilla VPN;

* Be signed in;

**Steps to reproduce:**

1. Click on the “Update now” link from the blue banner;
1. Click on the “Update now” blue button from the “Update recommended” screen;
1. Finalize the update;
1. Turn ON the VPN;
1. Observe the behavior;

**Expected result:**

* The VPN is successfully turned ON;

**Actual result:**

* The VPN cannot be turned ON;

**Notes:**

* Quitting the app does not resolve the issue;

* Not reproducible after directly installing the 2.10 version;

* Not reproducible on MacOS;

* See the attached video and logs;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3029)
┆author: Bianca Hidecuti

data-sync-user:
➤ Andrea Marchesini commented:

[17.10.2022 11:29:58.840] Debug: (windows - WindowsTunnelService) Service: C:/Program Files/Mozilla/Mozilla VPN/Mozilla VPN.exe
[17.10.2022 11:29:58.885] Error: (main - WindowsCommons) Failed to start the service - The service did not respond to the start or control request in a timely fashion.
[17.10.2022 11:29:58.886] Error: (windows - WireguardUtilsWindows) Failed to activate the tunnel service
[17.10.2022 11:29:58.886] Error: (main - Daemon) Interface creation failed.
[17.10.2022 11:29:58.886] Error: (main - DaemonLocalServerConnection) Failed to activate the interface
[17.10.2022 11:30:00.501] Debug: (main - DaemonLocalServerConnection) Read Data
[17.10.2022 11:30:00.501] Debug: (main - DaemonLocalServerConnection) Command received: activate
[17.10.2022 11:30:00.501] Debug: (main - Daemon) Activating interface
[17.10.2022 11:30:00.501] Debug: (main - WindowsCommons) Getting Current Internet Adapter that routes to XXXXXXXX
[17.10.2022 11:30:00.507] Debug: (main - WindowsCommons) Internet Adapter: wireless_32768
[...]

Issue #4649, opened 2022-10-17
[Windows] “Start on boot” feature does not work after an update


data-sync-user:
**Mozilla VPN affected versions:**

* 2.10.0 (2.202210111527);

**Platform tested:**

* Win 10;

**Prerequisites:**

* Log in with an account with subscription for Mozilla VPN;
* Have the “start on boot” option enabled;
* Have a 2.9 build installed;

**Steps to reproduce:**

1. Click the “update now” link from the blue banner;
1. Wait so that the update is finalized;
1. Observe VPN status;

**Expected results:**

* After update, the VPN is ON;

**Actual results:**

* After the update the VPN is OFF;

**Notes:**

* For 1 second it looks like it tries to start, but cannot;
* The feature works as expected after a normal install;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3028)
┆author: Valentina Virlics

data-sync-user:
➤ Owen Kirby commented:

I think this issue has already been fixed on main by with PR #4624 ( https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4624 ). We just needed to cherry-pick that commit onto the release branch.

PR #4697, opened 2022-10-20
Revert windows executable rename back to MozillaVPN.exe

oskirby:
## Description
In the 2.10 release, we have a problem where the upgrade from 2.9 to 2.10 over an existing installation leaves the client partially installed, with many of the supporting DLLs missing. This appears to be a result of the WIX installer bundling everything together into a single component, which appears to be discouraged by [MSDN](https://learn.microsoft.com/en-us/windows/win32/msi/defining-installer-components).

This leads to the installer assuming that the component has been renamed and that it is safe to install the new component in parallel with the old one, and then deleting the old one, but because some of the files didn't change names this results in some files being deleted after the upgrade.

I think the proper long term fix is to create unique components for each file being installed, so that they can be versioned independently by MSI. Unfortunately that's going to be tricky to get right in the short term, but we can avoid the immediate breakage by reverting the executable filename back to `MozillaVPN.exe`.

## Reference
Github #4650 ([VPN-3029](https://mozilla-hub.atlassian.net/browse/VPN-3029))
Github #4649 ([VPN-3028](https://mozilla-hub.atlassian.net/browse/VPN-3028))

## Checklist
    
### DIFF
--- a/src/cmake/windows.cmake
+++ b/src/cmake/windows.cmake
@@ -5,7 +5,7 @@
 add_definitions(-DWIN32_LEAN_AND_MEAN)
 
 set_target_properties(mozillavpn PROPERTIES
-    OUTPUT_NAME "Mozilla VPN"
+    OUTPUT_NAME "MozillaVPN"
     VERSION ${CMAKE_PROJECT_VERSION}
     WIN32_EXECUTABLE ON
 )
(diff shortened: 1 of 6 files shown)


oskirby:
I am going to close this PR, since a better workaround was already implemented on `main` with #4624 and I think we just need to cherry-pick it onto the release branch.

PR #4698, opened 2022-10-20
Readme improvements for Windows

jps-moz:
## Description

-Noted recommended Qt version and required Qt packages
-Noted required VS studio packages
-Added hint about x64 prompt
-Added UI debug instructions
-Note about OpenSSL


oskirby (approved):
I think this looks good.

PR #4699, opened 2022-10-21
Fix a crash in the addon disabling procedure

bakulf:


brizental (commented):
What was the crash exactly? Were we not cleaning the `m_addons` list as required?

Would it be possible to add a unit test for it?

tests/functional/testAddons.js
```diff
@@ -0,0 +1,53 @@
...
+    await vpn.waitForCondition(async () => {
+      return parseInt(
+                 await vpn.getElementProperty('VPNAddonManager', 'count'),
+                 10) === count;
+    });
+
+    await vpn.setSetting('addon/customServer', 'true');
+    await vpn.setSetting(
+        'addon/customServerAddress',
+        'https://archive.mozilla.org/pub/vpn/addons/releases/latest/');
```
brizental, line 49: Wouldn't a reset do this?

tests/functional/testAddons.js
```diff
@@ -0,0 +1,53 @@
...
+    await vpn.wait();
+    await vpn.wait();
```
brizental, line 51: Why is this necessary?

src/addons/manager/addonmanager.h
```diff
@@ -18,6 +18,7 @@ class AddonManager final : public QAbstractListModel {
 
   Q_PROPERTY(
       bool loadCompleted MEMBER m_loadCompleted NOTIFY loadCompletedChanged)
+  Q_PROPERTY(int count READ count NOTIFY countChanged)
```
brizental, line 21: Can we drop the sinal and just expose the prop without it? IIUC the signal is not being used in the tests.

brizental (commented):
Thank you for adding the tests. Left some questions, otherwise looks good.

tests/functional/testAddons.js
bakulf: yes but it does much more: logout the user, etc...

src/addons/manager/addonmanager.h
bakulf: A property can be CONSTANT or NOTIFY. If we dont set the signal, the property will not be usable in QML. We are not using it now, but if we implement the property, let's implementing correctly. What do you think?

brizental: Ah, ok I did not know it was either or. Ok then!

tests/functional/testAddons.js
brizental: Alright, would you mind adding a comment? This is not obvious.

tests/functional/testAddons.js
bakulf: tests do not run on the same machine and they do not share the same configuration. So, no. All should be fine.

brizental: Well, in that case we probably don't even to do all of these resetting operations? Your call, no strong opinion.

brizental (approved):

PR #4725, opened 2022-10-26
Update the what's new 2.10 date - VPN-3076

bakulf:


brizental (approved):

addons/message_whats_new_v2.10/manifest.json
```diff
@@ -11,7 +11,7 @@
     "id": "message_whats_new_v2.10",
     "title": "Welcome to Mozilla VPN 2.10",
     "subtitle": "You’re using the latest version of Mozilla VPN. Take a look at the newest features and updates available in Mozilla VPN 2.10:",
-    "date": 1661021990,
+    "date": 1666659600,
```
brizental, line 14: Whats the point in changing this now if both dates are in the past?

bakulf: at least it's in sync with the release date of 2.10.

PR #4728, opened 2022-10-26
Android: Fix white screen of death.

strseb:


nebula/ui/components/VPNStackView.qml
```diff
@@ -12,8 +12,15 @@ StackView {
...
 
     Component.onCompleted: function(){
-        if(!currentItem && typeof initialItem === "number" ){
+        if(!currentItem && initialItem) {
+            // We don't show anything right now and inital item is set, 
+            // On android if initialItem is anything But a component
+            // it will totaly parse that into garbage values and fail
+            // 
+            // See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/2638
             console.error("Failed to parse initialItem, try Component.OnComplete:push(someURI)");
```
brizental, line 21:
```suggestion
            console.error("Using the initialItem property does not work on some platforms. Use Component.onCompleted: stackview.push(someURI)");
```
Question, is the documentation we can link to on why this is illegal on Android? Or a bug maybe?

brizental: Also, is this only for Android? Is so we can replace "some platforms" with "Android".

nebula/ui/components/VPNStackView.qml
```diff
@@ -12,8 +12,15 @@ StackView {
...
+            VPN.exitForUnrecoverableError("Setting initialItem on a StackView is illegal, dont do that.");
```
brizental, line 22:
```suggestion
            VPN.exitForUnrecoverableError("Setting initialItem on a StackView is illegal. See previous logs for more information.");
```

brizental (approved):

nebula/ui/components/VPNStackView.qml
strseb: >Question, is the documentation we can link to on why this is illegal on Android? Or a bug maybe? 
The pr linked in the comment goes a bit more into detail why, which is the only thing i have found about this issue so far 😅 

>Also, is this only for Android? 
Only android-arm as far as i know.

brizental: Ack sorry, for some reason I thought the linked PR was this one...

PR #4732, opened 2022-10-26
VPN-3053 - Implement Glean Rust for VPN package and initialize it behind feature

brizental:
A few things are implemented here, but in essence this is just the bare minimum package and integration to use Glean Rust from C++. Commit by commit:

- ae055fe65 Add mozilla/glean as submodule
  - Not strictly necessary right now, but will be later so I am just doing it now
- 960127f52 Implement Glean Rust lib and initialize it with the VPN
  - This is the bulk of the PR here I create the Rust package, integrate it to the build process (cmake only for now)
- 7c7d0ac6d Change Glean Rust upload status when setting changes
  -  Also part of the bare mininum Glean integration
- d4bab7c86 Put Glean initialization behind feature flag
  - I am still unsure wheter to put init behind a feature flag or initialize but leave uploadEnabled always false when the flag is false. or wrap all Glean functions on a feature condition so that it is no-op? I wonder if C++ has something like decorators...

Only tested the build on Linux, CI is probably going to be red.


vpnglean/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+)
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(CARGO_CMD ${CARGO_BUILD_ENV} cargo build)
+    set(TARGET_DIR "debug")
+else ()
+    set(CARGO_CMD ${CARGO_BUILD_ENV} cargo build --release)
+    set(TARGET_DIR "release")
+endif ()
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
oskirby, line 31: This can be done platform agnostically with:
`set(LIBNAME ${CMAKE_STATIC_LIBRARY_PREFIX}vpnglean${CMAKE_STATIC_LIBRARY_SUFFIX})`

vpnglean/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    set(OPENSSL_USE_STATIC_LIBS TRUE)
+    find_package(OpenSSL REQUIRED)
+    cmake_path(GET OPENSSL_SSL_LIBRARY PARENT_PATH OPENSSL_LIB_DIR)
+endif()
+
+set(CARGO_BUILD_ENV
+    BUILD_ID=${BUILD_ID}
+    APP_VERSION=${APP_VERSION}
```
oskirby, line 17: I don't think `APP_VERSION` exists as a CMake variable, did you want `${CMAKE_PROJECT_VERSION}` instead?

src/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten")
     )
 endif()
 
-target_link_libraries(mozillavpn PRIVATE glean lottie nebula translations)
+target_link_libraries(mozillavpn PRIVATE glean vpnglean lottie nebula translations)
```
oskirby, line 28: As some thoughts for a future PR, why not merge `glean` and `vpnglean` into one library?

oskirby (changes requested):
For Linux build purposes, we also need to be sure to vendor the `vpnglean` module so that it can be built offline. This is done by adding a `cargo vendor` command to `scripts/linux/script.sh` somewhere around line 160.

vpnglean/CMakeLists.txt
brizental: Ha, yeap you are right. Why did I think it existed?! :eyes: :exploding_head:

src/CMakeLists.txt
brizental: So, the idea is to have both and leave vpnglean behind a feature flag untill it's fully implemented / validated. It's not pretty and `glean` and `vpnglean` is probably the worst terminology choice... Got any better naming ideas?

brizental: Hm, upon reading this again today I think I misunderstood what you said :sweat_smile:

src/glean/glean.cpp
```diff
@@ -0,0 +1,75 @@
...
+#else
+  return QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+#endif
+}
+}  // namespace
+
+Glean::Glean() {
+  MVPN_COUNT_CTOR(Glean);
+
+  connect(SettingsHolder::instance(), &SettingsHolder::gleanEnabledChanged,
```
bakulf, line 31: I would move this to `::initialize`

src/glean/glean.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef GLEAN_H
+#define GLEAN_H
+
+#include <QObject>
+
+constexpr const char* GLEAN_DATA_DIRECTORY = "glean";
+
+class Glean final : public QObject {
+  Q_OBJECT
```
bakulf, line 13: `Q_DISABLE_COPY_MOVE(Glean)`

src/mozillavpn.cpp
```diff
@@ -42,6 +42,7 @@
 #include "update/versionapi.h"
 #include "urlopener.h"
 #include "websocket/websockethandler.h"
+#include "glean/glean.h"
```
bakulf, line 45: alphabetic order.

src/CMakeLists.txt
oskirby: Well, since we don't change the build based on feature flags, we are always linking the glean.rs FFI library, so we might as well just bundle all of the "glean stuff" together into a single library as far as the `mozillavpn` target is concerned. For example, the `glean` library can depend on an internally-built glean.rs FFI library and export it as a public dependency.

Something along the lines of:
```
add_custom_command(
    OUTPUT ${CMAKE_CURRENT_BUILD_DIR}/libglean_ffi.a
    COMMAND cargo build ...
)
target_link_libraries(glean PUBLIC ${CMAKE_CURRENT_BUILD_DIR}/libglean_ffi.a)
```

src/mozillavpn.cpp
```diff
@@ -233,6 +234,10 @@ void MozillaVPN::initialize() {
 
   AddonManager::instance();
 
+  if (Feature::get(Feature::Feature_gleanRust)->isSupported()) {
```
bakulf, line 237: can we move this `if-stmt` in `::initialize`?

bakulf (approved):

oskirby (approved):

linux/mozillavpn.spec
```diff
@@ -12,13 +12,14 @@ Requires:  qt6-qtbase >= 6.0
...
 Requires:  qt6-qtsvg >= 6.0
-REquires:  qt6-qt5compat >= 6.0
+Requires:  qt6-qt5compat >= 6.0
 Requires:  wireguard-tools
 
 BuildRequires: golang >= 1.13
 BuildRequires: polkit-devel
 BuildRequires: python3-yaml
 BuildRequires: cargo
+BuildRequires: openssl
```
brizental, line 22: @oskirby I think this is necessary... Is it though?

.github/workflows/linux.yaml
```diff
@@ -97,21 +97,36 @@ jobs:
...
+          # so we have to do it the hard way.
+          yum -y install perl git
+          git clone https://github.com/openssl/openssl.git
+          cd openssl
+          ./Configure
+          make
+          # I think the issue with `yum install` is that it will install the package
+          # in a folder Cmake is not expecting, while this command always users /usr/local
+          make install
+          cd ..
```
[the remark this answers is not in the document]
brizental: I tried `dnf` with `--installroot` because `yum` doesn't have an option like that (that I could find). However didn't have success with that -- also didn't try very hard, gotta be honest.

Anyways, the main issue is that this makes this specific CI job exceedingly slow. Maybe we can cache the OpenSSL build skip the build which is the longest running script? Let me know.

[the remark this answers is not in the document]
brizental: Talked offline, que tl;dr; is: no this is not a good workaround, but we can fix it in a follow-up. Filed: https://mozilla-hub.atlassian.net/browse/VPN-3116?atlOrigin=eyJpIjoiY2NkMDQyZTVjNmU0NDYzZjk4ZWI5NmY0N2RmYzc4NzAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

linux/mozillavpn.spec
brizental: Also talked offline, yes seems to be necessary.

badboy (commented):
I can't say much about the build integration. cmake is not my strong suit.
The bit of calling into the Glean RLB looks correct.

I do remember though that we talked about _also_ using Glean Kotlin and initializing from there to get the benefits of the platform integration (metrics/baseline ping & app cycle, Kotlin uploader).
What's the status on that?

brizental:
> I do remember though that we talked about also using Glean Kotlin and initializing from there to get the benefits of the platform integration (metrics/baseline ping & app cycle, Kotlin uploader).
> What's the status on that?

Coming on the next PRs! I'm splitting it up a bit so it's not just one massive PR :D

.github/workflows/linux.yaml
```diff
@@ -97,21 +97,36 @@ jobs:
...
+          name: Sources
 
       - name: Install Build Dependencies
         run: |
+          # Install and build OpenSSL from scratch.
+          #
+          # Using `yum install openssl` doesn't seem to make cmake happy,
+          # so we have to do it the hard way.
+          yum -y install perl git
+          git clone https://github.com/openssl/openssl.git
```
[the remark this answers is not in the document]
brizental: Noted!

Issue #4628, opened 2022-10-13
[Linux] Multi-hop Mozilla VPN cannot be turned ON if the default Firewall is enabled


data-sync-user:
**Mozilla VPN affected versions:**

* 2.10.0~rc78-focal1;

**Platform tested:**

* Linux Ubuntu 20.04;

**Prerequisites:**

* Be logged in with an account with subscription for Mozilla VPN;
* Be on Stage;
* Have the Ubuntu default firewall (ufw) enabled;

**Steps to reproduce:**

1. Go to Select Location and enable Multi-Hop option
1. Turn the ufw on -> “sudo ufw enable”
1. Turn the VPN on;

**Expected results:**

* The VPN is turned on with success;

**Actual results:**

* The VPN cannot be turned on; after few seconds the “server unavailable” modal is displayed;

**Notes:**

* Not reproducing for single hop;
* Not reproducing on other platforms;
* Reproducing for both stage and prod;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3007)
┆Reporter: Valentina Virlics
┆Link To Issue: https://mozilla-hub.atlassian.net/browse/VPN-3007

data-sync-user:
➤ Valentina Virlics commented:

Owen Kirby Please take a look over this to see if this is not actually an “expected“ behavior; thanks!

data-sync-user:
➤ Santiago Andrigo commented:

Corner case of UFW interfering with multi-hop.
Valentina Virlics does this also happen if UFW is enabled before the multi-hop selection? (step 1)

data-sync-user:
➤ Valentina Virlics commented:

Santiago Andrigo  Yes, it does. The apps enters in “No signal“.

oskirby:
I have been able to reproduce this. It seems like the connection to the entry server works just fine, as expected, but it's the connection to the exit server that never completes.

data-sync-user:
➤ Valentina Virlics commented:

Owen Kirby should’nt we expect to see the “firewall block“ strings in the server unavailable modal, in this case? (more details https://mozilla-hub.atlassian.net/browse/VPN-1748 ( https://mozilla-hub.atlassian.net/browse/VPN-1748|smart-link ))

data-sync-user:
➤ Owen Kirby commented:

[...]

Issue #2788, opened 2022-02-16
Multi-hop doesn't work on Fedora 35 without Gnome Shell


dannycolin:
### Step to reproduce
1. [Download the RPM package from GitHub Actions on the official repository][download] (Last build tested: 2.7~rc15)
1. Install the package with `dnf install ./path/to/RPM/package` (this is needed to ensure that all dependencies are installed)
1. Open Mozilla VPN Client
1. Configure Multi-Hops

### Actual behavior
Mozilla VPN fails to connect to the servers configured in the multi-hops preference and return a "Server unavailable" error. I have tested multiple server permutations to be sure that it wasn't limited to only a specific set (or specific order) but it didn't look like it.

![mozilla-vpn-unavailable-server-multi-hops](https://user-images.githubusercontent.com/7339076/154329384-9fa0f46a-01e7-467a-9414-3194dc66d058.png)

### Expected behavior
Mozilla VPN successfully connects to the servers (enter and exit nodes) configured in multi-hops.

[download]: https://github.com/mozilla-mobile/mozilla-vpn-client/actions/workflows/linux.yaml?page=2&query=branch%3Amain

### Additional informations
* strace log: https://gist.github.com/dannycolin/811e03f5b35875f1598b903245e825df
* systemd log: https://gist.github.com/dannycolin/5d6567aabfcbece8697f46c7a1e5dcf8

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1767)

data-sync-user:
➤ Valentina Virlics commented:

This might have the same root cause as: https://mozilla-hub.atlassian.net/browse/VPN-1748 ( https://mozilla-hub.atlassian.net/browse/VPN-1748|smart-link ) Andrea Marchesini, could you confirm? Thanks!

data-sync-user:
➤ Andrea Marchesini commented:

I’m not sure yet.

dannycolin:
I've done a fresh install of Fedora 35 (the default install with Gnome) and it now works. I seems like it must be related to a dependency issue or a system configuration that isn't present on (some) non-Gnome environment. 

@oskirby any idea what dependency or configuration could be missing here? or if it could be the same underlaying issue than #1869?

dannycolin:
Still doesn't work on Fedora 36 minimal install but does on Fedora Workstation (Gnome).

data-sync-user:
➤ Lesley Norton commented:

[...]

PR #4735, opened 2022-10-26
Fix bug in netfilter inbound traffic marking

oskirby:
## Description
It seems that enabling the firewall in Ubuntu prevents the VPN client from being able to connect to the exit server when multihop is enabled. In tracing through this issue it seems that this is due to a bug in how inbound wireguard traffic is marked and we find ourselves in a situation where packets sent to the exit server are marked, but the inbound traffic is left unmarked, and the difference causes `conntrack` to miss the handshake.

To address this issue, we revise how we set the inbound traffic mark and use a netfilter set to match the packet source address against the known wireguard peers. This ensures that the inbound traffic is marked correctly, establishes a valid `conntrack` state and passes the firewall as intended.

## Reference
Github #4628 ([VPN-3007](https://mozilla-hub.atlassian.net/browse/VPN-3007))
Github #2788 ([VPN-1767](https://mozilla-hub.atlassian.net/browse/VPN-1767))

## Checklist
    
### DIFF
--- a/src/platforms/linux/daemon/wireguardutilslinux.cpp
+++ b/src/platforms/linux/daemon/wireguardutilslinux.cpp
@@ -241,8 +241,8 @@ bool WireguardUtilsLinux::updatePeer(const InterfaceConfig& config) {
   }
 
   // Update the firewall to mark inbound traffic from the server.
-  GoString goAddress = { .p = qPrintable(config.m_serverIpv4AddrIn),
-                         .n = (ptrdiff_t)config.m_serverIpv4AddrIn.length() };
+  GoString goAddress = {.p = qPrintable(config.m_serverIpv4AddrIn),
+                        .n = (ptrdiff_t)config.m_serverIpv4AddrIn.length()};
   NetfilterMarkInbound(goAddress, config.m_serverPort);
 
   // Set/update peer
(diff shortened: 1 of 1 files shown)


linux/netfilter/netfilter.go
```diff
@@ -120,43 +121,56 @@ func (ctx* nftCtx) nftIfup(ifname string) {
...
   ctx.conn.AddRule(&nftables.Rule{
     Table: ctx.table_inet,
     Chain: ctx.preroute,
     Exprs: []expr.Any{
-      // Match packets arriving from outside the tunnel
-			&expr.Meta{
-        Key:      expr.MetaKeyIIFNAME,
-        Register: 1,
+      // For now, we only support IPv4 in this check.
+      // TODO: Support IPv6.
```
bakulf, line 131: is this something we want to cover as follow up? Do we want to file a separate jira issue?

bakulf (approved):

linux/netfilter/netfilter.go
oskirby: We probably should, but the VPN client assumes IPv4 in many other areas of the code so it won't be a simple fix.

PR #4745, opened 2022-10-28
Taskcluster: Check compat with "next" QT

strseb:
I want to have a task that will assert we can get builds with the "next" version of QT - so we can either QA that or throw this in some automation ... or let me check if my build tools break :)

### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -199,8 +199,8 @@ int CommandUI::run(QStringList& tokens) {
     // https://bugreports.qt.io/browse/QTBUG-82617
     // Currently there is a crash happening on exit with Huawei devices.
     // Until this is fixed, setting this variable is the "official" workaround.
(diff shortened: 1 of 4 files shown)


src/commands/commandui.cpp
```diff
@@ -199,8 +199,8 @@ int CommandUI::run(QStringList& tokens) {
     // https://bugreports.qt.io/browse/QTBUG-82617
     // Currently there is a crash happening on exit with Huawei devices.
     // Until this is fixed, setting this variable is the "official" workaround.
-    // We certainly should look at this once 6.4 is out.
-#  if QT_VERSION >= 0x060400
+    // We certainly should look at this once 6.6 is out.
+#  if QT_VERSION >= 0x060600
```
strseb, line 203: No reason to remove it at 6.4

bakulf (approved):

Issue #4730, opened 2022-10-26
Cargo not installed for Taskcluster Linux builds


ahal:
The Taskcluster builds are failing because `cargo` is not installed:
https://firefox-ci-tc.services.mozilla.com/tasks/SIYGl5JqT1yugl672LkgSw/runs/0/logs/public/logs/live.log#L92

According to @oskirby , this is why they failed to build on launchpad.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-3080)

data-sync-user:
➤ Valentina Virlics commented:

Raphael Okafor If this does not require any QA work, could you please add the “qa-not-actionable“ label? Thanks!

PR #4749, opened 2022-10-28
Add `set -e` to all the bash scripts - VPN-3081

bakulf:


ahal (approved):
Thanks! Note this will make the Linux builds red due to https://github.com/mozilla-mobile/mozilla-vpn-client/issues/4730, but that's an improvement over silently failing :)

PR #4751, opened 2022-10-28
Better logout procedure - VPN-3037

bakulf:
Running the removal-device together with the reset() operation means that, if something goes wrong, at the next execution we show the initial screen immediately.

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -859,8 +859,18 @@ void MozillaVPN::logout() {
   }
 
   if (m_private->m_deviceModel.hasCurrentDevice(keys())) {
-    TaskScheduler::scheduleTask(new TaskRemoveDevice(keys()->publicKey()));
+    TaskScheduler::scheduleTask(new TaskGroup(
+        {new TaskRemoveDevice(keys()->publicKey()),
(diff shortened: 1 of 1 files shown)


brizental (commented):
Can we add tests to guarantee the expected behaviour?

src/mozillavpn.cpp
```diff
@@ -859,8 +859,18 @@ void MozillaVPN::logout() {
...
+    TaskScheduler::scheduleTask(new TaskGroup(
+        {new TaskRemoveDevice(keys()->publicKey()),
+         // Immediately after the scheduling of the device removal, we want to
+         // delete the session token, so that, in case the app is terminated, at
+         // the next execution we go back to the init screen.
+         new TaskFunction([this]() { reset(false); })}));
+
+    // In case the app is closed even before scheduling the previous TaskGroup,
+    // removing the key we will enforce a new authentication at the first
+    // TaskAccount execution.
```
brizental, line 871: Why is this logic only necessary when the user has the current device? What if the app is closed even before scheduling the reset task on line 876?

brizental: Nevermind, I get it on a second look.

brizental (approved):

PR #4764, opened 2022-10-31
Make the TaskAddonIndex reschedulable - VPN-3094

bakulf:
I don't know if we want to write a unit-test to check the delete state of each task. It seems not particularly useful. Note the the logic is fully tested already.


brizental:
Question: How does this address VPN-3094?

Issue #3520, opened 2022-05-12
Localizer unit tests perma failure due to incorrectly assuming locale value


data-sync-user:
This is a way confusing bug title, apologies. 

Here is a the failure:

```
FAIL!  : TestLocalizer::systemLanguage() Compared values are not the same
   Actual   (l.previousCode()): "en-DE"
   Expected ("en")            : en
   Loc: [[...]/mozilla-vpn-client/tests/unit/testlocalizer.cpp(31)]
```

This only happens because my computer doesn’t use the `en` locale.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2214)
┆author: Beatriz Rizental Machado

data-sync-user:
➤ Owen Kirby commented:

Yes! Thank you, this has been bothering me for ages.

data-sync-user:
➤ Andrea Marchesini commented:

Beatriz Rizental Machado This is fixed now, right?

data-sync-user:
➤ Beatriz Rizental Machado commented:

The test hasn’t changed so I don’t see how it would have been fixed. I changed to a computer where the locale is set to “en” hehe, so I haven’t seen this anymore. You know what, this should be an easy enough fix, I’ll take it and fix it.

PR #4767, opened 2022-10-31
Fix 'es' language on iOS - VPN-2212

bakulf:
A bit of context: on IOS, bc47name() returns `es-US` when the language is set to spanish. The fix to this issue is to use the `uiLanguages` method (https://doc.qt.io/qt-6/qlocale.html#uiLanguages).

In this PR I also centralize the fetching of system language code in all the code base.


oskirby (approved):
This might enable us to properly mock out the system language detection that was leading to #3520/[VPN-2214](https://mozilla-hub.atlassian.net/browse/VPN-2214)

PR #4769, opened 2022-10-31
Android CMake MVP

strseb:
The original pr -> https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4301 was a bit too much to chew of. 
Turns out this worked fine with the current Qt, but on 6.4 we would have pain upgrading. 

Let's be more iterative: this pr just switches from qmake to cmake for debug builds, as adjust files are missing. That will be the next followup :) 
The whole android-deploy step remains unchanged.


oskirby (approved):

bakulf (approved):

bakulf:
Can you also change `src.pro` to do something like:
```
} else:android {
error(qmake is not supported for Android. Please use cmake.)
```

And remove `src/qmake/platform/android.pri`

strseb:
> Can you also change `src.pro` to do something like:
> 
> ```
> } else:android {
> error(qmake is not supported for Android. Please use cmake.)
> ```
> 
> And remove `src/qmake/platform/android.pri`

We can do that, once we can do release builds with cmake. The adjust stuff is missing rn :)

PR #4779, opened 2022-11-01
VPN-2804 - Get notifications to work (again) on MacOS

brizental:
This does _not_ cause the other systray icon to blink when a new notification is show. At least not that I could see.

### DIFF
--- a/src/platforms/macos/macossystemtraynotificationhandler.cpp
+++ b/src/platforms/macos/macossystemtraynotificationhandler.cpp
@@ -58,6 +58,18 @@ void MacosSystemTrayNotificationHandler::showHideWindow() {
   }
 }
 
+void MacosSystemTrayNotificationHandler::notify(Message type,
+                                                const QString& title,
+                                                const QString& message,
+                                                int timerMsec) {
+  // This is very hacky, but necessary to circumvent a Qt bug where
+  // notifications are not shown when the icon is not visible.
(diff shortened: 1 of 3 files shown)


src/platforms/macos/macossystemtraynotificationhandler.cpp
```diff
@@ -58,6 +58,18 @@ void MacosSystemTrayNotificationHandler::showHideWindow() {
...
+void MacosSystemTrayNotificationHandler::notify(Message type,
+                                                const QString& title,
+                                                const QString& message,
+                                                int timerMsec) {
+  // This is very hacky, but necessary to circumvent a Qt bug where
+  // notifications are not shown when the icon is not visible.
+  // See: https://bugreports.qt.io/browse/QTBUG-108134
+  m_systemTrayIcon->show();
+  SystemTrayNotificationHandler::notify(type, title, message, timerMsec);
+  m_systemTrayIcon->hide();
```
flozia, line 70: Would it make sense to check the visibility with `QSystemTrayIcon::isVisible` before or do we always want to call `hide`?

brizental: We always want to hide at least in MacOS where we create our own system tray icon. If we don't hide an empty icon shows up next to the MozVPN icon.

flozia (approved):

PR #4785, opened 2022-11-01
No bug - Check for OpenSSL in Android too when building Glean

brizental:


bakulf (approved):
@strseb do you see issues with adding this check?

strseb (changes requested):
`find_package(OpenSSL REQUIRED)`
is going to fail on android, you would need to add 3rdparty/openssl to the includepath first.

brizental:
Thanks folks for taking a look. This was just a quick test I did yesterday before leaving without much thinking to see if it would fix the Android issue. I'll take a proper look today. Sorry for not flagging the PR is a test sooner!

PR #4786, opened 2022-11-02
VPN-3071 Unify mobile and desktop auth (WIP)

birdsarah:
## Description

* Use the same on-boarding screen for desktop and mobile
* Uses IAP flow for all platforms
* Defines a new dummy product "web" (need to work on UI for this)
* Uses DummyIAPHandler (this needs to move) to redirect back to the webflow in the case of a subscription being needed
* Adds email to the authenticate call if it's available in the settings holder (which it won't be if a user most recently logged out, but will be in this new flow) - this lets us pre-populate fxa screen nicely.

## Reference

    [VPN-3071](https://mozilla-hub.atlassian.net/browse/VPN-3071) (and some others)

## Checklist
    
### DIFF
--- a/src/platforms/dummy/dummyiaphandler.cpp
+++ b/src/platforms/dummy/dummyiaphandler.cpp
@@ -27,7 +27,8 @@ void DummyIAPHandler::nativeRegisterProducts() {
 
 void DummyIAPHandler::nativeStartSubscription(Product* product) {
   Q_ASSERT(product->m_name == "web");
-  MozillaVPN::instance()->logout();
+  // This is a hack. Should we make a new state?
+  MozillaVPN::instance()->setUserState(MozillaVPN::UserNotAuthenticated);
   MozillaVPN::instance()->authenticateWithType(
       MozillaVPN::AuthenticationType::AuthenticationInBrowser);
 }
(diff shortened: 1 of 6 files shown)


src/platforms/dummy/dummyiaphandler.cpp
```diff
@@ -18,10 +20,17 @@ DummyIAPHandler::DummyIAPHandler(QObject* parent) : IAPHandler(parent) {
...
+void DummyIAPHandler::nativeRegisterProducts() {
+  QTimer::singleShot(200, this,
+                     [this]() { emit productsRegistrationCompleted(); });
+}
 
 void DummyIAPHandler::nativeStartSubscription(Product* product) {
-  Q_UNUSED(product)
+  Q_ASSERT(product->m_name == "web");
+  // This is a hack. Should we make a new state?
+  MozillaVPN::instance()->setUserState(MozillaVPN::UserNotAuthenticated);
```
bakulf, line 31: This is bad, because technically, the user is already authenticated.
I think you need to call the `AuthenticationListener::create() + start()` in a task.

src/tasks/authenticate/taskauthenticate.cpp
```diff
@@ -81,8 +82,12 @@ void TaskAuthenticate::run() {
             m_authenticationListener->aboutToFinish();
           });
 
+  SettingsHolder* settingsHolder = SettingsHolder::instance();
+  Q_ASSERT(settingsHolder);
```
bakulf, line 86: I would revert this change.

PR #4793, opened 2022-11-02
Android: Support "Start On Boot" 

strseb:
Adds Support for start on boot.

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/BootReceiver.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/BootReceiver.kt
@@ -2,33 +2,33 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
- package org.mozilla.firefox.vpn.daemon
+package org.mozilla.firefox.vpn.daemon
 
- import android.content.BroadcastReceiver
- import android.content.Context
- import android.content.Intent
- import android.os.Build
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.os.Build
 
- class BootReceiver : BroadcastReceiver() {
-     private val TAG = "BootReceiver"
+class BootReceiver : BroadcastReceiver() {
+    private val TAG = "BootReceiver"
 
-     override fun onReceive(context: Context, arg1: Intent) {
-         Log.init(context)
-         if(!Prefs.get(context).getBoolean(START_ON_BOOT,false) ){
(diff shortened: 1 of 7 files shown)


src/platforms/android/androidvpnactivity.cpp
```diff
@@ -142,3 +149,9 @@ void AndroidVPNActivity::onServiceDisconnected(JNIEnv* env, jobject thiz) {
   logger.debug() << "service disconnected";
   emit AndroidVPNActivity::instance()->serviceDisconnected();
 }
+void AndroidVPNActivity::startAtBootChanged(const bool& startAtBoot) {
+  QJsonObject args;
+  args["startOnBoot"] = startAtBoot;
+  QJsonDocument doc(args);
+  sendToService(ServiceAction::ACTION_SET_START_ON_BOOT, doc.toJson());
+}
```
brizental, line 157: nit: New line

src/featureslistcallback.h
```diff
@@ -150,7 +150,7 @@ bool FeatureCallback_splitTunnel() {
 
 bool FeatureCallback_startOnBoot() {
 #if defined(MVPN_LINUX) || defined(MVPN_MACOS) || defined(MVPN_WINDOWS) || \
-    defined(MVPN_DUMMY) || defined(MVPN_WASM)
+    defined(MVPN_DUMMY) || defined(MVPN_WASM) || defined(MVPN_ANDROID)
```
brizental, line 153: nit: At this point should we just check if it's **not** iOS instead?

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/BootReceiver.kt
```diff
@@ -0,0 +1,34 @@
...
+    override fun onReceive(context: Context, arg1: Intent) {
+        Log.init(context)
+        if (!Prefs.get(context).getBoolean(START_ON_BOOT, false)) {
+            Log.i(TAG, "This device did not enable start on boot - exit")
+            return
+        }
+        Log.i(TAG, "This device did enable start on boot - try to start")
+        val intent = Intent(context, VPNService::class.java)
+        intent.putExtra("startOnBoot", true)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
brizental, line 24: Just to see if I get it. We are checking if the current Android SDK version is greater than Android SDK version `Oreo`? Ha... That took me a while, why does 0 look so much like O???????? Anyways, would you mind adding a comment here on why we need to check this?

brizental (approved):
Not sure I am the right reviewer here but... r+wc :trophy

bakulf (approved):

PR #4821, opened 2022-11-03
Deprecate qmake for MacOS

oskirby:
## Description
The time is almost upon us where we can finally get rid of qmake for MacOS. This requires us to:
- [x] Migrate CI tests to use CMake for all platforms.
- [x] Migrate taskcluster to use CMake for MacOS.
- [ ] Verify that package signing works as expected.
- [ ] Verify that installer signing works as expected.

## Reference
Much inspiration and cherry-picking was done from PR #4146 and PR #4536

## Checklist


tests/unit/websocket/testwebsockethandler.cpp
```diff
@@ -33,12 +33,18 @@ void MockServer::close() {
 }
 
 void MockServer::open() {
-  if (m_pWebSocketServer->listen(QHostAddress::Any, 5000)) {
+  if (m_pWebSocketServer->listen(QHostAddress::LocalHost, m_port)) {
```
bakulf, line 36: m_port is not initialized. Am I right?

bakulf (changes requested):

tests/unit/websocket/testwebsockethandler.cpp
oskirby: It is intentionally initialized to zero in the class declaration (`testwebsockethandler.h`). This gives us the behaviour that the port is chosen randomly from any available port when creating the class, but then persists if we open/close the websocket server.

This was needed for the `tst_reconnectionAttemptsAfterUnexpectedClose` test

oskirby:
So, it seems that signing of the binaries and installer packages aren't done except for the `main` branch, so the best I can really do is verify that this PR produces identical-looking build artifacts to what we get from qmake.

bakulf (approved):

PR #4824, opened 2022-11-03
Add 2.11 whats new message manifest

MattLichtenstein:
## Description

- Adds 2.11 what's new addon message manifest

## Reference

[VPN-3157: Implement 2.11 Manifest](https://mozilla-hub.atlassian.net/browse/VPN-3157)

## Checklist
    
### DIFF
--- a/addons/message_whats_new_v2.11/manifest.json
+++ b/addons/message_whats_new_v2.11/manifest.json
@@ -1,0 +1,33 @@
+{
+  "api_version": "0.1",
+  "id": "message_whats_new_v2.11",
+  "name": "What's new message",
+  "type": "message",
+  "conditions": {
+    "min_client_version": "2.11.0",
+    "max_client_version": "2.11.9"
+  },  
+  "message": {
+    "id": "message_whats_new_v2.11",
+    "title": "Welcome to Mozilla VPN 2.11",
+    "subtitle": "You’re using the latest version of Mozilla VPN. Take a look at the newest features and updates:",
+    "date": 1668020390,
+    "badge": "whats_new",
+    "blocks": [
+      { "id": "c_1",
+        "type": "ulist",
+        "content": [
+          { "id": "l_1",
+            "content": "If you have enabled “Connect VPN on startup” on desktop, the app will now launch minimized whenever you start your computer." },
+          { "id": "l_2",
+            "content": "We’ve made bug fixes, UI adjustments and performance improvements to make Mozilla VPN even better." }
+        ]
+      },
+      {
+        "id": "c_2",
+        "type": "text",
+        "content": "Thank you for installing the latest version of Mozilla VPN!" 
+      }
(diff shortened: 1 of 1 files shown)


addons/message_whats_new_v2.11/manifest.json
```diff
@@ -0,0 +1,33 @@
+{
+  "api_version": "0.1",
+  "id": "message_whats_new_v2.11",
+  "name": "What's new message",
+  "type": "message",
+  "conditions": {
+    "min_client_version": "2.11.0",
+    "max_client_version": "2.11.9"
+  },
```
MattLichtenstein, line 9: Is this okay? making `max_client_version` 2.12.0 doesn't make much sense since there will be a new message for 2.12. No need for 2 what's new messages right?

addons/message_whats_new_v2.11/manifest.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "message",
+  "conditions": {
+    "min_client_version": "2.11.0",
+    "max_client_version": "2.11.9"
+  },  
+  "message": {
+    "id": "message_whats_new_v2.11",
+    "title": "Welcome to Mozilla VPN 2.11",
+    "subtitle": "You’re using the latest version of Mozilla VPN. Take a look at the newest features and updates:",
+    "date": 1668020390,
```
MattLichtenstein, line 14: This date is `Wednesday, November 9, 2022 6:59:50 PM` 
Does doing dates this way make sense? I feel like this should be generated client side... obviously out of scope for this ticket

bakulf: Correct! Good for now. We can do better in 2.12

bakulf (approved):

addons/message_whats_new_v2.11/manifest.json
lesleyjanenorton: Right. This makes sense although I wonder if it should `2.11.1` in case there is dot release and we need to show something else. @bakulf what do you think?

lesleyjanenorton (approved):

addons/message_whats_new_v2.11/manifest.json
brizental: The addons could provide a "version" condition instead, that would probably be better for this type of use case. I created a ticket: https://mozilla-hub.atlassian.net/browse/VPN-3179

brizental: Nahhh, my first comment is wrong ignore me. It should be right though, that's a bug! :bug:

flodolo:
I wasn't pinged by this PR :-( 

Pretty sure the CODEOWNERS file is not valid (it doesn't specify a user in the first line)
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/CODEOWNERS

PR #4833, opened 2022-11-04
Auto re-publish archive stuff

strseb:
## Description

@albionx suggested it'd be nice if we have the binaries also under the gh releases tab. 
-> This task will automaticly on each release check if archive has published bin's for that release and attach them to the gh release. 

Demo: 
This release -> https://github.com/strseb/mozilla-vpn-client/releases/tag/v2.10.0
Triggered this task -> https://github.com/strseb/mozilla-vpn-client/actions/runs/3394219420

### DIFF
--- a/.github/workflows/fetch_binaries.yaml
+++ b/.github/workflows/fetch_binaries.yaml
@@ -1,0 +1,37 @@
+name: Publish Binaries to GH releases
+
+# This task will on every created release automaticly look
+# at archive.mozilla.org and upload files under the 
+# GH-Release tab. 
+#
+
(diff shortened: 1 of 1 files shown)


oskirby:
Oh my, this is amazing, there are so many things I want to publish now!

birdsarah (changes requested):
I'm not sure about this.

My feeling is that there should only be one canonical place that builds live.

If this workflow updated the release text with links to archive.mozilla.org then that would be :shipit:.

bakulf (approved):

PR #4838, opened 2022-11-04
Transaction for settings

bakulf:
This PR implements transactions for the settings. Addon tutorials will be able to start a transaction, change a few things, than revert all.

In order to support other setting changes during a tutorial, I marked all the settings with a new flag: 'user-settings'. A user-setting is something that is controlled by the user via the front-end code. Only those settings are reverted in case of a rollback or a crash.

### DIFF
--- a/scripts/ci/jsonSchemas/tutorial.json
+++ b/scripts/ci/jsonSchemas/tutorial.json
@@ -40,6 +40,10 @@
       "type": "string",
       "description": "The image of the tutorial"
     },
+    "settings_rollback_needed": {
+      "type": "boolean",
+      "description": "Do we need to revert all the settings changes at the end of the tutorial?"
+    },
     "steps": {
       "type": "array",
       "description": "The list of text steps",
--- a/src/settingsholder.cpp
+++ b/src/settingsholder.cpp
@@ -10,7 +10,10 @@
 #include "logger.h"
 #include "models/feature.h"
 
+#include <QDir>
+#include <QFile>
 #include <QSettings>
+#include <QStandardPaths>
 
 namespace {
 
@@ -24,6 +27,9 @@ QVector<QString> SENSITIVE_SETTINGS({
 
 SettingsHolder* s_instance = nullptr;
 
+const QSettings::Format MozFormat = QSettings::registerFormat(
+    "moz", CryptoSettings::readFile, CryptoSettings::writeFile);
+
 }  // namespace
 
 // static
@@ -32,20 +38,36 @@ SettingsHolder* SettingsHolder::instance() {
   return s_instance;
 }
 
-#ifndef UNIT_TEST
-const QSettings::Format MozFormat = QSettings::registerFormat(
-    "moz", CryptoSettings::readFile, CryptoSettings::writeFile);
-#endif
-
 SettingsHolder::SettingsHolder()
-    :
+    : m_settings(MozFormat, QSettings::UserScope,
 #ifndef UNIT_TEST
-      m_settings(MozFormat, QSettings::UserScope, "mozilla", "vpn")
+                 "mozilla",
 #else
-      m_settings("mozilla_testing", "vpn")
+                 "mozilla_testing",
 #endif
-{
+                 "vpn") {
   MVPN_COUNT_CTOR(SettingsHolder);
+
+  if (QFile::exists(journalSettingFileName())) {
+    QString journalSettingFile(journalSettingFileName());
+    logger.warning() << "journal file exists" << journalSettingFile;
+
+    // We cannot simply rename the journal settings file because the main
+    // QSettings could be not a file on some platforms (windows, for instance).
+    {
+      QSettings journalSettings(journalSettingFile, MozFormat);
+      for (const QString& key : journalSettings.allKeys()) {
+        m_settings.setValue(key, journalSettings.value(key));
+      }
+    }
+
+    if (!QFile::remove(journalSettingFile)) {
+      logger.warning() << "Unable to remove the journal settings file"
+                       << journalSettingFile;
+    }
+
+    logger.warning() << "Recovering completed";
+  }
 
   Q_ASSERT(!s_instance);
   s_instance = this;
@@ -69,7 +91,9 @@ SettingsHolder::~SettingsHolder() {
   s_instance = nullptr;
 
 #ifdef UNIT_TEST
-  m_settings.clear();
+  if (!m_doNotClearOnDTOR) {
+    m_settings.clear();
+  }
 #endif
 }
 
@@ -77,10 +101,10 @@ void SettingsHolder::clear() {
   logger.debug() << "Clean up the settings";
 
 #define SETTING(type, toType, getter, setter, has, key, defvalue, \
-                removeWhenReset)                                  \
+                userSettings, removeWhenReset)                    \
   if (removeWhenReset) {                                          \
     m_settings.remove(key);                                       \
-    emit getter##Changed(defvalue);                               \
+    emit getter##Changed();                                       \
   }
 
 #include "settingslist.h"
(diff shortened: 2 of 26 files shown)


src/settingsholder.cpp
```diff
@@ -200,3 +225,101 @@ void SettingsHolder::setAddonSetting(const AddonSettingQuery& query,
...
+  return true;
+}
+
+bool SettingsHolder::finalizeTransaction() {
+  m_transactionChanges.clear();
+  delete m_settingsJournal;
+  m_settingsJournal = nullptr;
+
+  if (!QFile::remove(journalSettingFileName())) {
+    logger.warning() << "Unable to remove the setting journal file"
```
brizental, line 300: Is a warning really enough here? What if we can't remove and then start a new transaction?

src/settingsholder.cpp
```diff
@@ -31,21 +37,37 @@ SettingsHolder* SettingsHolder::instance() {
...
-      m_settings("mozilla_testing", "vpn")
+                 "mozilla_testing",
 #endif
-{
+                 "vpn") {
   MVPN_COUNT_CTOR(SettingsHolder);
 
+  if (QFile::exists(journalSettingFileName())) {
+    QString journalSettingFile(journalSettingFileName());
+    logger.warning() << "journal file exists" << journalSettingFile;
```
brizental, line 52: If a file exists it means the app was closed unexpectedly with a transaction was happening. Correct?

Does this really need to be a warning? Also, can you add to the log message that a recovery process will start?

src/settingsholder.cpp
```diff
@@ -31,21 +37,37 @@ SettingsHolder* SettingsHolder::instance() {
...
+
+    // We cannot simply rename the journal settings file because the main
+    // QSettings could be not a file on some platforms (windows, for instance).
```
brizental, line 55: I am not sure I understand why this comment here and what it means.

src/settingsholder.cpp
```diff
@@ -144,8 +167,10 @@ QString SettingsHolder::getReport() const {
   }                                                                     \
   void SettingsHolder::setter(const type& value) {                      \
     if (!has() || getter() != value) {                                  \
+      maybeSaveInTransaction(key, getter(), value, #getter "Changed",   \
+                             userSettings);                             \
       m_settings.setValue(key, value);                                  \
-      emit getter##Changed(value);                                      \
+      emit getter##Changed();                                           \
```
brizental, line 173: Why remove `value` from the signal?

src/settingslist.h
```diff
@@ -143,6 +156,7 @@ SETTING_STRING(entryServerCity,     // getter
                hasEntryServerCity,  // has
                "entryServer/city",  // key
                nullptr,             // default value
+               true,                // user setting
```
brizental, line 159: Is there any way we could enforce this is used correctly in code? What makes a setting a user setting is it the fact that it can be invoked from QML? If so, can we make it not invokable if not a user setting? I am thinking about us and other developers in the future just copy pasting other settings and not thinking about this exact property.

tests/unit/testsettings.cpp
```diff
@@ -0,0 +1,101 @@
...
+    QVERIFY(settingsHolder.beginTransaction());
+    settingsHolder.setTheme("BBB");
+    QCOMPARE(settingsHolder.theme(), "BBB");
+    settingsHolder.setToken("TOKEN 2");
+    QCOMPARE(settingsHolder.token(), "TOKEN 2");
+  }
+
+  {
+    SettingsHolder settingsHolder;
+    QCOMPARE(settingsHolder.theme(), "AAA");
```
brizental, line 96: Ha! I'm confused. Why does theme not change?

scripts/ci/jsonSchemas/tutorial.json
```diff
@@ -40,6 +40,10 @@
       "type": "string",
       "description": "The image of the tutorial"
     },
+    "settings_rollback_needed": {
+      "type": "boolean",
+      "description": "Do we need to revert all the settings changes at the end of the tutorial?"
+    },
```
brizental, line 46: I think unless we are going to alert users that the tutorials will change their settings or not, we should just make this default `true` and not even provide it as a property on the manifest.

src/settingsholder.cpp
bakulf: Because it sounds like a bug having 2 nested transactions or canceling the previous one.

src/settingsholder.cpp
bakulf: Yes, let's replace `warning` with `info`

src/settingsholder.cpp
bakulf: because:
1. we don't use it too much
2. it's complex to "cast" the value when we rollback values

src/settingslist.h
bakulf: well, some values are exposed to the dev-menu. It's not so easy to implement what you are proposing.

bakulf: can we use CODEOWNERS to protect the settings list...

tests/unit/testsettings.cpp
bakulf: because we are simulating a recovering from a missing rollback/commit

scripts/ci/jsonSchemas/tutorial.json
bakulf: we are going to alert the user. This is a task for @MattLichtenstein

src/settingsholder.cpp
bakulf: I would add a Q_ASSERT() to prevent this from happening. So far, this is used only by the Tutorial engine and we cannot have 2 tutorials happening at the same time.

tests/unit/testsettings.cpp
bakulf: because token is not an user-setting

src/settingsholder.cpp
brizental: Hm, but the issue is not really if two tutorials happen at same time. It's if one tutorial happens and cannot delete the journal file, leaves it there corrupted and then another one starts and has to deal with the corrupted file.

tests/unit/testsettings.cpp
brizental: Ah, of course. Ok makes sense. Would you mind adding some comments just to make the test easier to read?

src/settingsholder.cpp
bakulf: Good point. I added a telemetry event to monitor how often it happens.

src/settingslist.h
```diff
@@ -462,6 +514,7 @@ SETTING_STRINGLIST(vpnDisabledApps,     // getter
                    hasVpnDisabledApps,  // has
                    "vpnDisabledApps",   // key
                    QStringList(),       // default value
+                   false,               // user setting
```
MattLichtenstein, line 517: This should be a user setting, no?

glean/metrics.yaml
```diff
@@ -1467,6 +1467,23 @@ sample:
...
+    bugs:
+      - https://mozilla-hub.atlassian.net/browse/VPN-3184
+    data_reviews:
+      - TODO
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - amarchesini@mozilla.com
+    expires: never
+
```
brizental, line 1486: I think it's good to add telemetry to monitor this functionality. However I wonder if this metric by itself is enough. What questions would we answer with it? We would simply have an absolute count of how many times the journal file fails to be deleted, but nothing else to analyse it against. 

If we are going to add telemetry here I think we would need also events that check:

1. A transaction has started
2. A trasaction has completed succesfully
3. A transaction has failed, with a `reason` extra key on why it failed (which would include this case)

This would allow us to create analyses that check the percentage of failures on this feature and what are the most prominent failures. 

What do you think? Do you think we already have metrics that would help us answer that without adding new ones?

Also, it's fine to do this as a follow up... So I'll r+ this PR for now :)

brizental (approved):
r+wc

src/settingslist.h
brizental: As discussed on Slack, let's file a follow up to deal with this scenario as it would not be a simple change.

PR #4839, opened 2022-11-06
VPN-3071(Part 1) Split products

birdsarah:
## Description

Split products management out of IAPHandler. This paves the way for more purchase management in app on desktop.

This PR should be a no-op functionally. It is just separating concerns to lay the groundwork for the next PR.

## Reference

Prep work for [VPN-2253](https://mozilla-hub.atlassian.net/browse/VPN-2253)

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -200,11 +200,11 @@ RadioDelegate {
 
             function getSubscriptionDuration(product) {
                 switch (product) {
-                    case VPNIAP.ProductMonthly:
+                    case VPNProducts.ProductMonthly:
                         return 1;
-                    case VPNIAP.ProductHalfYearly:
+                    case VPNProducts.ProductHalfYearly:
                         return 6;
-                    case VPNIAP.ProductYearly:
+                    case VPNProducts.ProductYearly:
                         return 12;
                     default:
                         return -1;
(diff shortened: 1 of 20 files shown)


bakulf (changes requested):
I don't understand why we have IAPHandler and ProductHandler. We can have 1 single class for both?

birdsarah:
@bakulf re-requesting review.

I fixed the comment comment but not your other two notes.

To recap what we discussed earlier:

* Splitting the IAPHandler into IAPHandler and ProductsHandler paves the way for the next PR where I will create PurchaseHandler with subclasses PurchaseIAPHandler and PurchaseWebHandler. Because PurchaseWebHandler has no product handling dependencies this makes this hierarchy clean.
* What is not clean is that I'm short on time and instead of splitting the native implementations (wasmiaphandler, iosiaphandler, and androidiaphandler) into two pieces also I've left them together. I've done this for expediency for delivering this feature by Monday. It's not as clean as it could be. I can pick this clean-up work up for 2.13.
* Also, for posterity, to recap our conversation that my initial preference for implementing this was PurchaseHandler (QObject) has subclasses PurchaseIAPHandler and PurchaseWebHandler where the PurchaseIAPHandler multi-inherited from PurchaseHandler and QAbastractList so that PurchaseIAPHandler could be the model for the QML view. However, Qt does not allow multiple inheritance of two QObjects (QAbstractList is derived from QObject) so this wasn't an option and this is how I ended up with what's in this PR.

birdsarah:
For reference @bakulf the follow-on PR is here: #4883

The code isn't finished, but the re-org (so you can see the shape of the final product), is mostly done.

bakulf (approved):

PR #4857, opened 2022-11-07
Tech-Dept: Simplify VPN-Permission flow

strseb:
__This needs a rebase - This is currently not based ontop of main... because of the build failures.__

This PR simplifies the way we ask for the VPN permission. Since we now can pass object with the parcelable trait between the IPC border, we no longer need: 
-> A Fake VPN service generateing a permission Prompt
-> Private QT Apis for both fireing aswell as recieving results
-> c++ client code to feed the result back into the service. 

I'm pretty sure we had deadlock conditions in there, as we have at least 3 threads involved. Also GCP is pretty clear this piece of code is 90% of all "application is not responding cases" .

--------

This pr moves all of that permission stuff into the binder thread, cutting out all native code of the path. So even if this does not solve our ANR's we will have more visibility what the problem here is, instead of having the stackflow hide behind `QTPrivate::something` :)


android/build.gradle
```diff
@@ -211,4 +211,27 @@ task ktlintFormat(type: JavaExec, group: "formatting") {
     classpath = configurations.ktlint
     main = "com.pinterest.ktlint.Main"
     args "-F", "src/**/*.kt", "!**/build/**/*.kt", "buildSrc/**/*.kt",  "daemon/**/*.kt", "**/*.kts"
-}
\ No newline at end of file
+}
+
+
+// Tasks to Import the Platform plugins
```
strseb, line 217: This is due to missing rebase :)

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -174,24 +174,19 @@ class VPNService : android.net.VpnService() {
       * If the permission is given, returns true
       * Requests permission and returns false if not.
       */
-    fun checkPermissions(): Boolean {
+    fun checkPermissions(): Intent? {
         // See https://developer.android.com/guide/topics/connectivity/vpn#connect_a_service
         // Call Prepare, if we get an Intent back, we dont have the VPN Permission
         // from the user. So we need to pass this to our main Activity and exit here.
         val intent = prepare(this)
```
bakulf, line 181: `return prepare(this);`

bakulf (approved):

Issue #4219, opened 2022-08-18
The input-text fields should not be editable after clicking buttons in the auth-in-app


data-sync-user:
See the video.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-2702)
┆Reporter: Andrea Marchesini
┆Link To Issue: https://mozilla-hub.atlassian.net/browse/VPN-2702

data-sync-user:
➤ Florian Zia commented:

We have to investigate if that’s true for other inputs.

data-sync-user:
➤ Valentina Virlics commented:

Not sure why this has Fix Version 2.10. I’ll update it, and add 2.12 label. Thank you!

PR #4864, opened 2022-11-07
Disable auth inputs after clicking submit - Fix VPN-2702

lesleyjanenorton:
## Description
Opening as draft until I can smoke test on a physical device (tomorrow). 

Disable authentication inputs after clicking submit. Re-enable auth inputs if there is an error and we are unable to proceed to next step in flow. 

## Reference

VPN-2702, #4219 

## Checklist
    
### DIFF
--- a/nebula/ui/components/inAppAuth/VPNInAppAuthenticationInputs.qml
+++ b/nebula/ui/components/inAppAuth/VPNInAppAuthenticationInputs.qml
@@ -26,7 +26,7 @@ ColumnLayout {
     }
 
     function disableActiveInput() {
-        return activeInput().enabled = false;
+        activeInput().enabled = false;
     }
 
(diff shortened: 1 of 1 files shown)


nebula/ui/components/inAppAuth/VPNInAppAuthenticationInputs.qml
```diff
@@ -25,13 +25,21 @@ ColumnLayout {
         return _isSignUpOrIn ? passwordInput : textInput
     }
 
+    function disableActiveInput() {
+        return activeInput().enabled = false;
```
bakulf, line 29: The `return` is confusing. Do you use the returned value?

bakulf (approved):

nebula/ui/components/inAppAuth/VPNInAppAuthenticationInputs.qml
lesleyjanenorton: good point. no. i'll remove

Issue #2, opened 2020-08-25
Complete the UI


bakulf:
List of unfinished views:

* [x] Initial view
* [x] Connecting view
* [x] Main view
* [ ] Settings
* [x] onboarding views
* [x] device view

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-312)

data-sync-user:
➤ Owen Kirby commented:

Closing this issue on Jira as it seems to have been created in error by the data sync.

PR #4897, opened 2022-11-09
Enable Start On Boot-ish for iOS

strseb:
## Description

This PR adds support for the "start-on-boot" settings option using the ios `onDemandRules` provider. 
calling this only "Start On Boot-ish" (and marking as WIP) as there is a catch, it will start the vpn on boot, if it was not manually disconnected at the time of of the reboot.

-> i.e. connecting. Closing app, reboot -> will restart vpn on reboot.
-> disconnecting. Closing app, reboot -> will not. 

Reason for this: 
The `onDemand` rule matching `any` will apparently trigger on a manual disconnect, so the OS would just bring the tunnel back up. I'm not really smart enough to know if we can somehow write a rule that would match only a reboot?

Smart ideas welcome on how to better do that :D 

## Reference

VPN-3113

## Checklist
    
### DIFF
--- a/src/apps/vpn/appfeaturelistcallback.h
+++ b/src/apps/vpn/appfeaturelistcallback.h
@@ -146,6 +146,8 @@ bool FeatureCallback_startOnBoot() {
 #if defined(MZ_LINUX) || defined(MZ_MACOS) || defined(MZ_WINDOWS) || \
     defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_IOS)
   return true;
+#endif
+  return false;
 }
 
 bool FeatureCallback_unsecuredNetworkNotification() {
--- a/src/apps/vpn/appfeaturelistcallback.h
+++ b/src/apps/vpn/appfeaturelistcallback.h
@@ -144,7 +144,8 @@ bool FeatureCallback_splitTunnel() {
 
 bool FeatureCallback_startOnBoot() {
 #if defined(MZ_LINUX) || defined(MZ_MACOS) || defined(MZ_WINDOWS) || \
-    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_IOS)
+    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_ANDROID) ||  \
+    defined(MZ_IOS)
   return true;
 #else
   return false;
--- a/src/apps/vpn/appfeaturelistcallback.h
+++ b/src/apps/vpn/appfeaturelistcallback.h
@@ -146,8 +146,9 @@ bool FeatureCallback_startOnBoot() {
 #if defined(MZ_LINUX) || defined(MZ_MACOS) || defined(MZ_WINDOWS) || \
     defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_IOS)
   return true;
-#endif
+#else
   return false;
+#endif
 }
 
 bool FeatureCallback_unsecuredNetworkNotification() {
--- a/src/apps/vpn/appfeaturelistcallback.h
+++ b/src/apps/vpn/appfeaturelistcallback.h
@@ -144,11 +144,8 @@ bool FeatureCallback_splitTunnel() {
 
 bool FeatureCallback_startOnBoot() {
 #if defined(MZ_LINUX) || defined(MZ_MACOS) || defined(MZ_WINDOWS) || \
-    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_ANDROID)
+    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_IOS)
   return true;
-#else
-  return false;
-#endif
 }
 
 bool FeatureCallback_unsecuredNetworkNotification() {
--- a/src/apps/vpn/platforms/ios/ioscontroller.swift
+++ b/src/apps/vpn/platforms/ios/ioscontroller.swift
@@ -257,24 +257,21 @@ public class IOSControllerImpl : NSObject {
         }
     }
     @objc func setAlwaysOn(setEnabled: Bool){
-            if(!setEnabled){
-                tunnel!.isOnDemandEnabled = false;
-                tunnel!.onDemandRules = []
-                return;
-            }
-            let alwaysConnect = NEOnDemandRuleConnect()
-            alwaysConnect.interfaceTypeMatch = .any
-            tunnel!.isOnDemandEnabled = true
-            tunnel!.onDemandRules = [alwaysConnect]
+        if(!setEnabled){
+            tunnel!.isOnDemandEnabled = false;
+            tunnel!.onDemandRules = []
+            return;
+        }
+        let alwaysConnect = NEOnDemandRuleConnect()
+        alwaysConnect.interfaceTypeMatch = .any
+        tunnel!.isOnDemandEnabled = true
+        tunnel!.onDemandRules = [alwaysConnect]
     }
     
     @objc func getAlwaysOn() -> Bool{
         return tunnel!.isOnDemandEnabled;
     }
-    // Only if the onDemand rules is set
-    // and contains our "always on rule"
-    // the use would have had the option
-    // to make a choise
+    
     @objc func hasAlwaysOn() -> Bool{
         let rules = tunnel!.onDemandRules;
         if( rules == nil){
@@ -288,8 +285,8 @@ public class IOSControllerImpl : NSObject {
     @objc func disconnect() {
         Logger.global?.log(message: "Disconnecting")
         assert(tunnel != nil)
-        // Turn off always-on, as this would
-        // oterwise imediatly trigger a reconnect from the OS
+        // Turn off "always-on", as this rule would
+        // trigger a reconnect from the OS
         setAlwaysOn(setEnabled: false)
         tunnel!.saveToPreferences { [unowned self] saveError in
             if saveError != nil {
--- a/src/apps/vpn/platforms/ios/ioscontroller.swift
+++ b/src/apps/vpn/platforms/ios/ioscontroller.swift
@@ -165,7 +165,7 @@ public class IOSControllerImpl : NSObject {
         return true
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, enableAlwaysConnect:Bool , failureCallback: @escaping () -> Void) {
+    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, failureCallback: @escaping () -> Void) {
         Logger.global?.log(message: "Connecting")
         assert(tunnel != nil)
 
--- a/src/apps/vpn/platforms/ios/ioscontroller.swift
+++ b/src/apps/vpn/platforms/ios/ioscontroller.swift
@@ -2,30 +2,29 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import Foundation import NetworkExtension
+import Foundation
+import NetworkExtension
 
-    let vpnName = "Mozilla VPN" var vpnBundleID = "";
+let vpnName = "Mozilla VPN"
+var vpnBundleID = "";
 
 @objc class VPNIPAddressRange : NSObject {
- public
-  var address : NSString = "" public var networkPrefixLength
-      : UInt8 = 0 public var isIpv6 : Bool = false
+    public var address: NSString = ""
+    public var networkPrefixLength: UInt8 = 0
+    public var isIpv6: Bool = false
 
-                                      @objc
-                                      init(address
-                                           : NSString, networkPrefixLength
-                                           : UInt8, isIpv6
-                                           : Bool) {
-    super
-        .init()
+    @objc init(address: NSString, networkPrefixLength: UInt8, isIpv6: Bool) {
+        super.init()
 
-            self.address = address self.networkPrefixLength =
-        networkPrefixLength self.isIpv6 = isIpv6
-  }
+        self.address = address
+        self.networkPrefixLength = networkPrefixLength
+        self.isIpv6 = isIpv6
+    }
 }
 
 public class IOSControllerImpl : NSObject {
- private var tunnel: NETunnelProviderManager? = nil
+
+    private var tunnel: NETunnelProviderManager? = nil
     private var stateChangeCallback: ((Bool) -> Void?)? = nil
     private var privateKey : PrivateKey? = nil
     private var deviceIpv4Address: String? = nil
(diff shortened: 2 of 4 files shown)


mcleinman:
> ## Description
> 
> This PR adds support for the "start-on-boot" settings option using the ios `onDemandRules` provider. calling this only "Start On Boot-ish" (and marking as WIP) as there is a catch, it will start the vpn on boot, if it was not manually disconnected at the time of of the reboot.
> 
> -> i.e. connecting. Closing app, reboot -> will restart vpn on reboot. -> disconnecting. Closing app, reboot -> will not.
> 
> Reason for this: The `onDemand` rule matching `any` will apparently trigger on a manual disconnect, so the OS would just bring the tunnel back up. I'm not really smart enough to know if we can somehow write a rule that would match only a reboot?
> 
> Smart ideas welcome on how to better do that :D
> ## Reference
> 
> VPN-3113
> ## Checklist
> 
>     * [ ]  My code follows the style guidelines for this project
> 
>     * [ ]  I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
> 
>     * [ ]  I have performed a self review of my own code
> 
>     * [ ]  I have commented my code PARTICULARLY in hard to understand areas
> 
>     * [ ]  I have added thorough tests where needed

Just spent some more time in the docs, and some more time searching. Given that all the rules are about network state and request destination, I don't think there is an "always on boot, even if it was previously disconnected" possibility.

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,45 @@ public class IOSControllerImpl : NSObject {
             }
         }
     }
+    @objc func setAlwaysOn(setEnabled: Bool){
```
mcleinman, line 259: Currently, nothing in the code explicitly says this is for "connect on reboot". It seems like this function might be a good spot. Can you add a quick comment here (or another spot that seems appropriate)?

src/platforms/ios/ioscontroller.swift
```diff
@@ -165,7 +165,7 @@ public class IOSControllerImpl : NSObject {
         return true
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, failureCallback: @escaping () -> Void) {
+    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int,enableAlwaysOn:Bool, failureCallback: @escaping () -> Void) {
```
mcleinman, line 168: Lower priority -
```suggestion
    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, enableAlwaysOn: Bool, failureCallback: @escaping () -> Void) {
```
And similar spacing in a few other spots.

mcleinman (changes requested):
Some code changes, and also a few issues:
1. We don’t update the iOS settings when a user toggles the setting in the app - we toggle the Connect On Demand setting the next time we connect the VPN. I think users would expect that setting to be immediately reflected.
2. Since users are able to toggle “Connect On Demand” in the iOS Settings app, should we treat that as an identical toggle to the “connect on start up” setting in the app? If so, we need to push the “keep system setting in check with app settings” elsewhere - potentially on every connect and disconnect. (Does it run the disconnect code every time a user shuts down the phone when the VPN is active? I’d hope so, but am not sure and would have to check.)
3. Relatedly, when toggling “Connect on Demand” in iOS Settings app currently, it turns off the VPN. If possible, flicking that switch shouldn’t turn off the entire VPN, it would just change the option.

mcleinman:
Let me know when this is ready for another round of review. Thanks!

strseb:
Thank you for the review, adjusted some things so it's ready for round 2 :D 

 > 1.  We don’t update the iOS settings when a user toggles the setting in the app - we toggle the Connect On Demand setting the next time we connect the VPN. I think users would expect that setting to be immediately reflected.

I agree that this is not really great, however if we were to sync the flag from client->System OS instantly, that would also force a connection/disconnection... which is probably not expected. 

> 2. Since users are able to toggle “Connect On Demand” in the iOS Settings app, should we treat that as an identical toggle to the “connect on start up” setting in the app? If so, we need to push the “keep system setting in check with app settings” elsewhere - potentially on every connect and disconnect. (Does it run the disconnect code every time a user shuts down the phone when the VPN is active? I’d hope so, but am not sure and would have to check.)
 
I've added an extra check, so every time the app goes into foreground it will now get checked aswell. 

> 3. Relatedly, when toggling “Connect on Demand” in iOS Settings app currently, it turns off the VPN. If possible, flicking that switch shouldn’t turn off the entire VPN, it would just change the option.

Afaik this is done by the OS and not by us. Could we even change that?

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,45 @@ public class IOSControllerImpl : NSObject {
...
         assert(tunnel != nil)
-        (tunnel!.connection as? NETunnelProviderSession)?.stopTunnel()
+        // Turn off "always-on", as this rule would
+        // trigger a reconnect from the OS
+        setAlwaysOn(setEnabled: false)
+        tunnel!.saveToPreferences { [unowned self] saveError in
+            if saveError != nil {
+                Logger.global?.log(message: "Save Error")
+            }
+            (tunnel!.connection as? NETunnelProviderSession)?.stopTunnel()
```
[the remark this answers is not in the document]
mcleinman: Thinking about this more, if alwaysOn is false and saving preferences failed, we'd fail to stop the tunnel. Could this be more defensive, and we move the `stopTunnel` to before we save the preferences?

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
+    * @brief Enables / Disables the Always-On Rule on the runnel
+    * This is used for the "start-on-boot" feature. 
+    * We are setting a rule matching * any network change
+    * which will cause the phone to start the VPN after the phone boots.
+    *
+    * Note: For the Setting to be applied to the iOS settings: 
+    *       tunnel!.saveToPreferences needs to be called.  
+    * @param setEnabled: bool If the rule should be added or removed.
+    **/
+    @objc func setAlwaysOn(setEnabled: Bool){
```
mcleinman, line 270:
```suggestion
    @objc func setAlwaysOn(setEnabled: Bool) {
```

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
+        if(!setEnabled){
```
mcleinman, line 271:
```suggestion
        if (!setEnabled) {
```

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
+            tunnel!.isOnDemandEnabled = false;
+            tunnel!.onDemandRules = []
+            return;
+        }
+        let alwaysConnect = NEOnDemandRuleConnect()
+        alwaysConnect.interfaceTypeMatch = .any
+        tunnel!.isOnDemandEnabled = true
+        tunnel!.onDemandRules = [alwaysConnect]
+    }
```
mcleinman, line 280: Can we log somewhere in this function that we're setting onDemand, and whether we're setting it on or off?

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
+    
+    @objc func getAlwaysOn() -> Bool{
```
mcleinman, line 282:
```suggestion
    @objc var getAlwaysOn: Bool {
```

Personal preference, but IMO a bit cleaner - making this a computed var. Feel free to ignore if you want.

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
+        return tunnel!.isOnDemandEnabled;
+    }
+    
+    @objc func hasAlwaysOn() -> Bool{
```
mcleinman, line 286:
```suggestion
    @objc var hasAlwaysOn: Bool {
```

Same as above

src/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,55 @@ public class IOSControllerImpl : NSObject {
...
     }
+    
+    /**
+    * @brief Enables / Disables the Always-On Rule on the runnel
+    * This is used for the "start-on-boot" feature. 
+    * We are setting a rule matching * any network change
+    * which will cause the phone to start the VPN after the phone boots.
+    *
+    * Note: For the Setting to be applied to the iOS settings: 
+    *       tunnel!.saveToPreferences needs to be called.
```
mcleinman, line 267: If we need to saveToPreferences afterward, we could include the save as part of this function - to ensure we never forget in the future - and provide this function a completion block to run after the successful save.

Let me know if you want me to write some code for this bit.

mcleinman:
> Thank you for the review, adjusted some things so it's ready for round 2 :D
> 
> > 1. We don’t update the iOS settings when a user toggles the setting in the app - we toggle the Connect On Demand setting the next time we connect the VPN. I think users would expect that setting to be immediately reflected.
> 
> I agree that this is not really great, however if we were to sync the flag from client->System OS instantly, that would also force a connection/disconnection... which is probably not expected.
> 
> > 2. Since users are able to toggle “Connect On Demand” in the iOS Settings app, should we treat that as an identical toggle to the “connect on start up” setting in the app? If so, we need to push the “keep system setting in check with app settings” elsewhere - potentially on every connect and disconnect. (Does it run the disconnect code every time a user shuts down the phone when the VPN is active? I’d hope so, but am not sure and would have to check.)
> 
> I've added an extra check, so every time the app goes into foreground it will now get checked aswell.

The interaction between #1 and #2 here causes a bug. I *think* we should save when the user toggles this setting, even if it causes the VPN to flicker off. 

Steps:
1. Turn VPN on in app.
2. Turn "turn on at startup" setting on, in the app. (At this point, it hasn't been set in the iOS settings.)
3. Background the app.
4. Open the app again. The app sees that its setting (turn on at startup) is out of sync w/ the iOS settings (do NOT turn on at startup), and updates itself to match iOS. Thus the app (and iOS) show as do NOT turn on at startup - the opposite of what the user just set.

I don't think "toggle this setting, then leave the app before toggling the VPN" is a rare path, and thus I think this should probably be updated.

 
> > 3. Relatedly, when toggling “Connect on Demand” in iOS Settings app currently, it turns off the VPN. If possible, flicking that switch shouldn’t turn off the entire VPN, it would just change the option.
> 
> Afaik this is done by the OS and not by us. Could we even change that?

I think you're right, there isn't anything we can do to change this. Was hoping you knew some secret. 😅

mcleinman (changes requested):
A bunch of small things, as well as the bigger inconsistency (mentioned in the comment replying to your comment) that we should discuss or fix before this gets merged.

src/platforms/ios/ioscontroller.swift
strseb: Good call, added a call to savePrefs but I was not sure how to add this type of handling or generally... really what the controller should do in case this fails :/

strseb:
>The interaction between https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1 and https://github.com/mozilla-mobile/mozilla-vpn-client/issues/2 here causes a bug. I think we should save when the user toggles this setting, even if it causes the VPN to flicker off.
Steps:
Turn VPN on in app.
Turn "turn on at startup" setting on, in the app. (At this point, it hasn't been set in the iOS settings.)
Background the app.
Open the app again. The app sees that its setting (turn on at startup) is out of sync w/ the iOS settings (do NOT turn on at startup), and updates itself to match iOS. Thus the app (and iOS) show as do NOT turn on at startup - the opposite of what the user just set.

Thanks! So that should be fixed, we're not updating the iOS setting aswell when it was changed :) 

Other general question, should we maybe inclide https://github.com/apple/swift-format in the CI? I appreciate the formatting comments, but you should not have to do that by hand :D

src/apps/vpn/appfeaturelistcallback.h
```diff
@@ -144,7 +144,7 @@ bool FeatureCallback_splitTunnel() {
 
 bool FeatureCallback_startOnBoot() {
 #if defined(MZ_LINUX) || defined(MZ_MACOS) || defined(MZ_WINDOWS) || \
-    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_ANDROID)
+    defined(MZ_DUMMY) || defined(MZ_WASM) || defined(MZ_IOS)
```
mcleinman, line 147: Is this  removing the feature from Android?

src/apps/vpn/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,53 @@ public class IOSControllerImpl : NSObject {
             }
         }
     }
+    
+    /**
+    * @brief Enables / Disables the Always-On Rule on the runnel
```
mcleinman, line 261:
```suggestion
    * @brief Enables / Disables the Always-On Rule on the tunnel
```

src/apps/vpn/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,53 @@ public class IOSControllerImpl : NSObject {
...
+        alwaysConnect.interfaceTypeMatch = .any
+        tunnel!.isOnDemandEnabled = true
+        tunnel!.onDemandRules = [alwaysConnect]
+    }
+    
+    @objc func getAlwaysOn() -> Bool {
+        return tunnel!.isOnDemandEnabled;
+    }
+    
+    @objc func hasAlwaysOn() -> Bool {
```
mcleinman, line 284: I believe this function will return `false` whenever the user has actively chosen to *not* enable "launch on boot", though we'd want it to return `true`. `onDemandRules` will have 0 rules, and so `isEmpty` will return `true`.

src/apps/vpn/platforms/ios/ioscontroller.swift
```diff
@@ -254,11 +256,53 @@ public class IOSControllerImpl : NSObject {
...
+    
 
     @objc func disconnect() {
         Logger.global?.log(message: "Disconnecting")
         assert(tunnel != nil)
         (tunnel!.connection as? NETunnelProviderSession)?.stopTunnel()
+        // Turn off "always-on", as this rule would
+        // trigger an immediate reconnect from the OS
+        setAlwaysOn(setEnabled: false)
+        tunnel!.saveToPreferences { [unowned self] saveError in
```
mcleinman, line 300: I don't believe we need `[unowned self]` here.

mcleinman (changes requested):
Marking as `Request Changes` until we resolve whether `hasAlwaysOn` needs a logic tweak.

Also, I should have thought of this before - given that we can't save this while the VPN is on, should we use the same "setting can't be toggled when VPN is on" type of messaging that we do for DNS server changes? And just prevent the user from toggling this unless the VPN is off? That's more of a Santiago call, I suppose, but realized you may be working hard for functionality we may not want.

And I'll ping you about linters when you're back in the office - I'm curious how we handle linting across our project.

src/apps/vpn/platforms/ios/ioscontroller.swift
strseb: >I believe this function will return false whenever the user has actively chosen to not enable "launch on boot"

If the user has actively chosen to not enable "launch on boot" on the Client, that is correct. Given that we clear out the rules when this option is disabled. (Which i think is good, otherwise we can have weird edge-cases)

However, if the user has chosen to disable it on the iOS system settings, the `onDemandRules` will not be empty only `isOnDemandEnabled` will be false. Therefore, I can use it to check if we need to sync the value of `isOnDemandEnabled` back to the client :)

src/apps/vpn/appfeaturelistcallback.h
strseb: Oooops. that should get removed, yes but not in this PR :)

strseb:
> Also, I should have thought of this before - given that we can't save this while the VPN is on, should we use the same "setting can't be toggled when VPN is on" type of messaging that we do for DNS server changes? And just prevent the user from toggling this unless the VPN is off? That's more of a Santiago call, I suppose, but realized you may be working hard for functionality we may not want.

Given that this now behaves as the iOS toggle - therefore (enabling enables), (disabling disables) the vpn i don't have the concern we have an issue with "dangling state". If we were to lock this setting to the vpn being off - that would mean a user can't untick the setting directly after enabling it as the VPN  get's turned on and therefore would be locked. While this sounds good for this settings activation rate, it might be unwanted :D

src/apps/vpn/platforms/ios/ioscontroller.swift
mcleinman: I just tested this:
1) In app, activate "always on".
2) In app, deactivate "always on".
3) In iOS Settings, activate "always on".
4) Relaunch app. Our app shows inactive "always on", and iOS settings is updated to also have an inactive "always on".

In this scenario, don't we want our app to match what the user did in iOS settings, and not vice versa?

mcleinman:
> > Also, I should have thought of this before - given that we can't save this while the VPN is on, should we use the same "setting can't be toggled when VPN is on" type of messaging that we do for DNS server changes? And just prevent the user from toggling this unless the VPN is off? That's more of a Santiago call, I suppose, but realized you may be working hard for functionality we may not want.
> 
> Given that this now behaves as the iOS toggle - therefore (enabling enables), (disabling disables) the vpn i don't have the concern we have an issue with "dangling state". If we were to lock this setting to the vpn being off - that would mean a user can't untick the setting directly after enabling it as the VPN get's turned on and therefore would be locked. While this sounds good for this settings activation rate, it might be unwanted :D

I'd love to get a thumbs up from Santiago and/or one of our designers, to make sure this is the way they'd want it. Given that there are multiple paths here that make sense (and many of our users may never know this toggle also exists in the iOS settings), I don't feel like I know their intentions well enough to approve the PR on my own. I'll ask Santiago to chime in here.

PR #4901, opened 2022-11-10
 VPN-3071(Part 2) Implement in-app auth and desktop purchase

birdsarah:
## Description

Follow on to #4839.

Implements the main feature:
* [x] in-app auth
* [x] in-app account creation   
* [x] new screen for purchase for web (calls to PurchaseWebHandler)
  * [x] Basic screen
  * [x] Get help from front-end engineer on layout (thank you @flozia) 

## Reference

* [VPN-3071](https://mozilla-hub.atlassian.net/browse/VPN-3071)
* [VPN-3072](https://mozilla-hub.atlassian.net/browse/VPN-3072)
* Also fixes [VPN-3217](https://mozilla-hub.atlassian.net/browse/VPN-3217)

## Checklist
    

src/commands/commandui.cpp
```diff
@@ -11,7 +11,7 @@
 #include "constants.h"
 #include "fontloader.h"
 #include "frontend/navigator.h"
-#include "iaphandler.h"
+#include "purchasehandler.h"
```
bakulf, line 14: Alphabetical order

src/mozillavpn.cpp
```diff
@@ -8,7 +8,7 @@
 #include "dnshelper.h"
 #include "frontend/navigator.h"
 #include "glean/glean.h"
-#include "iaphandler.h"
+#include "purchasehandler.h"
```
bakulf, line 11: Alphabetical order

src/networkmanager.cpp
```diff
@@ -67,10 +67,7 @@ QByteArray NetworkManager::userAgent() {
   {
     QStringList flags;
     flags.append(QString("sys:") + NetworkManager::osVersion());
-
-    if (Feature::get(Feature::Feature_inAppPurchase)->isSupported()) {
-      flags.append("iap:true");
-    }
+    flags.append("iap:true");
```
bakulf, line 70: It would be nice if we can get rid of this flag and use just the client version in Guardian

src/purchasehandler.h
```diff
@@ -46,18 +44,15 @@ class IAPHandler : public QObject {
...
+  PurchaseHandler(QObject* parent);
+  ~PurchaseHandler();
 
   enum State {
     eActive,
     eInactive,
   } m_subscriptionState = eInactive;
 
-  QString m_currentSKU;
+  QString m_currentSKU = QString();
```
bakulf, line 55: This is not needed. You already run the ctor of qstring

src/purchaseiaphandler.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "leakdetector.h"
+#include "logger.h"
+#include "models/feature.h"
+
+#include <QCoreApplication>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonValue>
+#include <QScopeGuard>
```
bakulf, line 17: Cleanup these headers. You don't need them.

src/purchaseiaphandler.h
```diff
@@ -0,0 +1,33 @@
...
+  void startSubscription(const QString& productIdentifier) override;
+  void startRestoreSubscription() override;
+
+ protected:
+  virtual void nativeStartSubscription(ProductsHandler::Product* product) = 0;
+  virtual void nativeRestoreSubscription() = 0;
+
+  bool m_hasRestore = true;
+  bool m_hasAlreadySubscribed = true;
+  bool m_hasBillingMethods = true;
```
bakulf, line 30: You don't use these vars

src/purchasewebhandler.cpp
```diff
@@ -0,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "purchasewebhandler.h"
+#include "leakdetector.h"
+#include "logger.h"
+
+#include <QCoreApplication>
```
bakulf, line 9: Remove this

src/purchaseiaphandler.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <QObject>
+
+class PurchaseIAPHandler : public PurchaseHandler {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(PurchaseIAPHandler)
+
+ public:
+  PurchaseIAPHandler(QObject* parent);
+  ~PurchaseIAPHandler();
```
bakulf, line 18: Virtual dtor

src/purchasehandler.h
```diff
@@ -46,18 +44,15 @@ class IAPHandler : public QObject {
   void stopSubscription();
 
  protected:
-  IAPHandler(QObject* parent);
-  ~IAPHandler();
-
-  virtual void nativeStartSubscription(ProductsHandler::Product* product) = 0;
-  virtual void nativeRestoreSubscription() = 0;
+  PurchaseHandler(QObject* parent);
+  ~PurchaseHandler();
```
bakulf, line 48: Virtual dtor

src/mozillavpn.cpp
```diff
@@ -1432,9 +1430,6 @@ void MozillaVPN::subscriptionFailedInternal(bool canceledByUser) {
         ErrorHandler::SubscriptionFailureError);
   }
 
-  TaskScheduler::scheduleTask(
```
birdsarah, line 1435: @bakulf need your help with this deletion.

The problem that I was finding was that when I intentionally cancelled the web subscription this new TaskAccount causes accountChecked to run which then causes the emptying of the settings holder. This caused subsequent restarts of my subscribe task to fail.

This change also seems to resolve VPN-3217 (canceling and then resubscribing causes error in subscription screen).

So, removing it seems like a win in these two cases. However, calling TaskAccount in subscriptionFailedInternal was added by you a *long* time ago (https://github.com/mozilla-mobile/mozilla-vpn-client/commit/234345d89ad362024254a6d0061bc1ca208847d7). Removing code that's been around for so long makes me nervous. Would love a consult from you on workflows where we would want this to happen and your thoughts on this change.

src/networkmanager.cpp
birdsarah: This is a good idea, but not small, i've opened https://mozilla-hub.atlassian.net/browse/VPN-3248.

src/purchaseiaphandler.h
birdsarah: I've added the virtual. For my own learning. Why do we specify it as virtual but still keep the implementation of the destructor in purchaseiaphandler.cpp. I see this is pattern we've done elsewhere e.g. networkmanager.h and .cpp just looking to learn more about the c++ mechanics here.

birdsarah:
Hey @bakulf this is finished and ready for another review.

translations/strings.yaml
```diff
@@ -214,6 +214,13 @@ restorePurchaseExpiredError:
 restorePurchaseInUseError:
   restorePurchaseInUseErrorText: Another Firefox Account has already subscribed using this Apple ID. Please sign out and try again or contact our support team for help.
 
+purchaseWeb:
+  title: Subscribe to Mozilla VPN
+  message:
+    value: We couldn’t find a subscription for %1. Subscribe now to protect your connection.
+    comment: The %1 is the user's email address.
```
flodolo, line 221:
```suggestion
    comment: %1 is the user's email address.
```

flodolo (approved):

src/mozillavpn.cpp
bakulf: I see why this fixes a lot of issues, and I'm happy with this change. The only concern I have is that, if the user completes the subscription on another device, there is no way to resume from this view except closing the app and restart. But this can be fixed in a follow up.

bakulf (approved):
Sorry for the delay.

PR #4904, opened 2022-11-10
Add the controller API to addons - VPN-3236

bakulf:
This is needed for the split-tunnel tutorial because we want to expose it only if the VPN is off.


brizental (approved):

bakulf:
> Is the list of APIs exposed to JS documented somewhere? If not, would you mind filing a follow up ticket?

We have a non-completed page in the `docs` folder.

PR #4907, opened 2022-11-10
No ticket - Check count before removing rows from AddonManager

brizental:
This was triggering a Q_ASSERT and thus causing a segfault.

### DIFF
--- a/src/addons/manager/addonmanager.cpp
+++ b/src/addons/manager/addonmanager.cpp
@@ -170,7 +170,7 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
         if (i.value().m_addon->enabled()) ++pos;
         continue;
       }
-      if (!enabled) {
+      if (!enabled && pos < count()) {
         beginRemoveRows(QModelIndex(), pos, pos);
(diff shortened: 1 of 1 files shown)


src/addons/manager/addonmanager.cpp
```diff
@@ -170,7 +170,7 @@ bool AddonManager::loadManifest(const QString& manifestFileName) {
         if (i.value().m_addon->enabled()) ++pos;
         continue;
       }
-      if (!enabled) {
+      if (!enabled && pos < count()) {
```
bakulf, line 173: I suspect the bug is elsewhere. I don't see why this should happen. Are you able to reproduce it consistently?
`count` impl should be in sync with this for-stmt.

brizental: Nope, it's totally intermitting. Yeah I agree it's probably somewhere else. I can try to dig a little bit and come back here.

brizental: Eh... let's close it. If I stumble on it again I can file a proper ticket and investigate as necessary.

PR #4916, opened 2022-11-11
2.12 tutorials!

bakulf:
This is mainly @MattLichtenstein's work. I added:
- the restore of the VPN connectivity after the tutorial 04
- fix a bug in the setting journal file name


src/apppermission.cpp
```diff
@@ -108,6 +108,9 @@ void AppPermission::flip(const QString& appID) {
 
   int index = m_applist.indexOf(AppDescription(appID));
   emit dataChanged(createIndex(index, 0), createIndex(index, 0));
+
+  beginResetModel();
```
bakulf, line 112: @MattLichtenstein why do we need this?

addons/tutorial_04_split_tunneling/manifest.json
```diff
@@ -0,0 +1,94 @@
...
+        },
+        "next": {
+          "op": "signal",
+          "vpn_emitter": "controller",
+          "signal": "stateChanged"
+        }
+      }, {
+        "id": "s1",
+        "element": "navigationLayout/navButton-settings",
+        "tooltip": "TODO TODO",
```
flodolo, line 34: We shouldn't expose TODO strings.

flodolo (changes requested):

addons/tutorial_04_split_tunneling/manifest.json
bakulf: the addon manifest is marked as "translatable" false because of this. As soon as we have the final strings I mark the addon as translatable.

flodolo: Thanks, I forgot about that flag. The rest looks good.

flodolo (approved):

addons/tutorial_04_split_tunneling/manifest.json
```diff
@@ -0,0 +1,94 @@
+{
+  "api_version": "0.1",
+  "id": "tutorial_04_split_tunneling",
+  "name": "Tutorial: Split tunneling",
+  "conditions": {
+    "enabled_features": ["splitTunnel"],
+    "javascript": "splitTunnelingOff.js"
```
MattLichtenstein, line 7: Just to confirm, we are requiring split tunneling to be turned off for the tutorial to even be visible to the user?

If so, I think we should update the A/C because right now it states: 

> If a user had already disabled the “Protect all apps with VPN” toggle, the tutorial should flip it back on so that the tutorial can advance on firm footing telling users to flip it off.

addons/tutorial_04_split_tunneling/manifest.json
```diff
@@ -0,0 +1,94 @@
...
+        "tooltip": "Toggle this switch to activate the VPN",
+        "conditions": {
+           "javascript": "vpnWasOn.js"
+        },
+        "next": {
+          "op": "signal",
+          "vpn_emitter": "controller",
+          "signal": "stateChanged"
+        }
+      }
```
MattLichtenstein, line 91: Should there be a step between these two to get the user to the home tab?

src/apppermission.cpp
MattLichtenstein: I was using this to reset the app permission list model so that the UI would update without having to reload the list component (aka popping the view off the stack and re-pushing it). Do you think there is there a better way to do this?

addons/tutorial_03_multi_hop/manifest.json
```diff
@@ -0,0 +1,99 @@
+{
```
MattLichtenstein, line 1: Are we still implementing the "Rollback to original servers" at the end of this tutorial?

bakulf: yes, we do.

lesleyjanenorton (approved):

addons/tutorial_03_multi_hop/manifest.json
lesleyjanenorton: Hmmmm - just ran this locally and I remained on Vienna -> Melbourne after the tutorial ended.

src/apppermission.cpp
bakulf: OK! we can work on this as a follow up.

addons/tutorial_03_multi_hop/manifest.json
bakulf: ah! multi-hop sorry. I think we are OK with this.

PR #4917, opened 2022-11-11
Recommended servers list

flozia:
## Description

Adds the QML layout for an additional tab to the server list with recommended locations. The feature sits behind the `recommendedServers` flag and is disabled by default. Currently the list is rendered with dummy data.

## Reference

- [Figma](https://www.figma.com/file/3HRiJ8mPABFpXx0vEd9q6L/Server-Latency-Reporting?node-id=3%3A5521&t=CtQnYumfqG4VDoKG-0): Design specs
- [VPN-3023](https://mozilla-hub.atlassian.net/browse/VPN-3023): Implement the Recommended Server UI

## Checklist
    

translations/strings.yaml
```diff
@@ -73,6 +73,30 @@
...
+    value: These locations are sorted by expected performance.
+    comment: Body text of an expandable info card on the recommended servers list.
+  recommendedTextStatus:
+    value: "Last updated %1 ago."
+    comment: Time passed since the last update to the recommended server list. %1 is the time value and unit.
+  recommendedTextStatusWithInfo:
+    value: "Last updated %1 ago. To update this list please first disconnect from the VPN."
+    comment:
+      - Time passed since the last update to the recommended server list and additional info.
+      - "%1 is the time value and unit since the list was updated."
```
flodolo, line 89: For this and the string before we should specify how the time value and unit are generated, i.e. if they're localized and how they will look like (I see a `// TODO: Replace string arg with the actual formatted date` comment in the code). How are these going to be generated?

nebula/ui/components/VPNServerCountry.qml
```diff
@@ -207,24 +201,15 @@ VPNClickableRow {
...
-
+                VPNServerLatencyIndicator {
                     anchors {
-                        right: del.right
+                        right: parent.right
                         rightMargin: VPNTheme.theme.hSpacing
-                        top: del.top
-                        verticalCenter: del.verticalCenter
+                        top: parent.top
+                        verticalCenter: parent.verticalCenter
```
MattLichtenstein, line 209: don't need to assign top anchors if setting verticalCenter anchors

nebula/ui/components/VPNServerList.qml
```diff
@@ -62,77 +129,306 @@ FocusScope {
...
+                            horizontalAlignment: Text.AlignLeft
+                            // TODO: Replace string arg with the actual formatted date
+                            text: !statusComponent.rowShouldBeDisabled
+                                ? VPNl18n.ServersViewRecommendedTextStatus.arg("5m")
+                                : VPNl18n.ServersViewRecommendedTextStatusWithInfo.arg("5m")
+                            wrapMode: Text.WordWrap
+                        }
+
+                        Item {
+                            Accessible.name: VPNl18n.ServersViewRecommendedTextInfo
```
MattLichtenstein, line 226: I don't see this in strings.yaml? And should the `Accessible.name` field be attached to the `ClickableRow`?

nebula/ui/components/VPNServerList.qml
```diff
@@ -62,77 +129,306 @@ FocusScope {
...
+                                source: refreshIcon
+                            }
+                        }
+                    }
+                }
+
+                Repeater {
+                    id: recommendedRepeater
+                    model: testRecommendedModel
+                    delegate: VPNClickableRow {
```
MattLichtenstein, line 255: The recommended servers need accessible names

MattLichtenstein (changes requested):
See comments

nebula/ui/components/VPNServerList.qml
flozia: Yes, it does. Added!

nebula/ui/components/VPNServerList.qml
flozia: Dang — changed the string id and moved it to the `ClickableRow`.

translations/strings.yaml
flozia: I was thinking that generating the time value and unit like we do  in [AddonMessage::dateInternal](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/addons/addonmessage.cpp#L165) would work, but thinking about it again this might be subject to change. I’d rather not introduce these strings with this PR and do it in a follow up.

flodolo (approved):

MattLichtenstein (approved):
🥳

PR #4918, opened 2022-11-11
Update issue templates

strseb:
Let's have an issue template that Forces the "bug" label. So that it will sync as a jira bug, not a task :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,10 +11,13 @@ assignees: ''
 A clear and concise description of what the bug is.
 
 **VPN version:**
+  <Please fill this in>
 
 **Affected Platforms:**
+  <Please fill this in>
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**VPN version:**
```
bakulf, line 13: I see a lot of dup fields here. In theory, linking the jira issue should contain most of these fields.
The jira issue should be enough.

strseb: This will be used if a User is filing a new issue. - The template is just copied from QA :D

birdsarah (approved):

Issue #4920, opened 2022-11-11
 [macOS] Mozilla VPN 2.12 build from theeherder is crashing after installation


data-sync-user:
**Affected versions:**

* Mozilla VPN 2.12.0;

**Tested Platforms:**

* MacOS 11.6;

**Prerequisites:**

* Have Mozilla VPN 2.12 build [+downloaded+](https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/co18JYOfSdKkblQHYRfJzA/runs/0/artifacts/public/build/MozillaVPN.pkg); 

**Steps to reproduce:**

1. Install the 2.12 build;
1. After the installation is finished, observe the behavior;

**Expected result:**

* The client is successfully opened after the installation;

**Actual result:**

* The client is crashing after installation;

**Notes:**

* Reproducible if opening the client from the Application folder;
* This is a blocker for QA, as the 2.12 version cannot be tested on MacOS;
* Not reproducing on Windows while using the latest 2.12 build from treeherder;

* See the attached video;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3247)
┆Reporter: Bianca Hidecuti

data-sync-user:
➤ Andrea Marchesini commented:

[14.11.2022 12:14:19.129] Warning: qrc:/ui/main.qml:5:1: module "QtQuick" plugin "qtquick2plugin" not found (main.qml:5)

Owen Kirby Sebastian Streich  Have we changed something in how we compile/link QT for macos?

data-sync-user:
➤ Owen Kirby commented:

We recently switched the taskcluster builds to use cmake for MacOS, so that is probably why we’re seeing this now.

I was able to reproduce the issue and make it go away with a tweak to the MacOS platform. I’m guessing that I simply never tested this with a Qt static build. With any luck this PR here ( https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4932 ) should fix the issue, but we won’t know for certain until it gets merged into main and we can test the signed taskcluster artifacts.

data-sync-user:
➤ Bianca Hidecuti commented:

[...]

PR #4932, opened 2022-11-14
Possible fix to QML plugin scanner on static MacOS builds

oskirby:
## Description
For static builds of Qt using CMake, we need to perform a scan to link in plugins required for the QML imports. For our MacOS builds, we are attempting to do this with an explicit call to `qt6_import_qml_plugins()` however, this only picks up dependencies that exist at the time of this call. It is far better to defer this to target finalization (eg: in `qt_finalze_target()`)

## Reference
Github issue #4920 ([VPN-3247](https://mozilla-hub.atlassian.net/browse/VPN-3247))

## Checklist
    
### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,8 @@ lottie/liblottie.a
 nebula/nebula_autogen/
 nebula/libnebula.a
 src/mozillavpn_autogen/
+/build-*/
+/build/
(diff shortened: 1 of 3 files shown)


bakulf (approved):
Let's see if it works!

PR #4956, opened 2022-11-16
add dummybuild to gitignore

mozrokafor:
## Description

    Add dummybuild to git ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ src/mozillavpn.app
 src/mozillavpn.framework
 src/version.h
 mozillavpn_wg
+dummybuild
 mozillavpn.qmltypes
 mozillavpn_metatypes.json
 mozillavpn_qmltyperegistrations.cpp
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -47,6 +47,7 @@ nebula/libnebula.a
 src/mozillavpn_autogen/
 /build-*/
 /build/
+dummybuild
```
bakulf, line 50: where does this dummybuild come from?

mozrokafor: cmake command ran before being able to run a test

cmake -S . -B $(pwd)/dummybuild -DBUILD_DUMMY=ON 
cmake --build $(pwd)/dummybuild -j$(nproc)

mozrokafor: functional test*

bakulf: Right but you can call this dir `build` too. Why dummybuild?

bakulf: Is it written in the README.md?

PR #4963, opened 2022-11-17
Fix settings menu scroll for some languages

MattLichtenstein:
## Description

- Removes the excess height in the settings screen for some languages that would cause the settings screen to scroll unnecessarily 
- **MORE IMPORTANTLY:** Replace anchors with layouts in `VPNScreenBase` which gets rid of that _annoyingggggg_ issue where determining position of a component on a view (by using something like `mapToItem()`, especially in `VPNFlickable` when calculating it's contentHeight) would initially be off by the size of a `VPNMenu` and would take an extra couple of milliseconds to figure itself out

## Reference

[VPN-3200: [l10n] Scroll in Settings screen for non-US languages](https://mozilla-hub.atlassian.net/browse/VPN-3200)

## Checklist
    

src/ui/screens/ScreenGetHelp.qml
```diff
@@ -16,13 +17,6 @@ Item {
...
     }
 
-    VPNMenu {
-        id: menu
-        objectName: "getHelpBack"
-        _menuOnBackClicked: () => VPNNavigator.requestPreviousScreen()
-        _menuIconButtonSource: getHelpStackView.depth === 1 ? "qrc:/nebula/resources/close-dark.svg" : "qrc:/nebula/resources/back.svg"
-    }
-
     VPNMouseArea {
```
lesleyjanenorton, line 20: I think we'll need to rehome this `MouseArea`, which seems to have been orphaned by the move of `VPNMenu`. I can't trigger the dev menu and seeing this in console: `Warning: qrc:/ui/screens/ScreenGetHelp.qml:20:5: QML VPNMouseArea: Cannot anchor to an item that isn't a parent or sibling. (ScreenGetHelp.qml:20)`

nebula/ui/components/VPNViewBase.qml
```diff
@@ -18,11 +18,10 @@ Item {
    property alias _viewContentData: viewContent.data
    property alias _interactive: vpnFlickable.interactive
 
-   height: parent.implicitHeight
```
lesleyjanenorton, line 21: I don't think this patch caused it but since we're here - seeing a lot of these complaints: `Warning: qrc:/nebula/components/VPNViewBase.qml:21:4: TypeError: Cannot read property 'height' of null (VPNViewBase.qml:21)` so maybe we just add an `if (parent)` to cheer up the console.

src/ui/screens/ScreenGetHelp.qml
```diff
@@ -4,67 +4,66 @@
...
+
+            property int unlockCounter: 0
 
-    VPNStackView {
-        id: getHelpStackView
+            objectName: "getHelpBack"
+            _menuOnBackClicked: () => VPNNavigator.requestPreviousScreen()
+            _menuIconButtonSource: getHelpStackView.depth === 1 ? "qrc:/nebula/resources/close-dark.svg" : "qrc:/nebula/resources/back.svg"
+            titleClicked: () => {
+                            if (unlockCounter >= 5) {
```
flozia, line 32: Nit: I think it would be nice to have threshold values — in this case the amount of clicks needed in order “unlock” the developer options — in `constants`. Wdyt?

src/ui/screens/ScreenGetHelp.qml
```diff
@@ -4,67 +4,66 @@
...
-            menu._menuOnBackClicked = currentItem._menuOnBackClicked ? currentItem._menuOnBackClicked : () => getHelpStackView.pop()
+            Component.onCompleted: function() {
+                VPNNavigator.addStackView(VPNNavigator.ScreenGetHelp, getHelpStackView)
+                getHelpStackView.push("qrc:/ui/screens/getHelp/ViewGetHelp.qml")
+            }
+
+            onCurrentItemChanged: {
+                menu.title = Qt.binding(() => currentItem._menuTitle || "");
+                menu.visible = Qt.binding(() => menu.title !== "");
+                menu._menuOnBackClicked = currentItem._menuOnBackClicked ? currentItem._menuOnBackClicked : () => getHelpStackView.pop()
```
flozia, line 65: Nit: Could this also be simplified:`menu._menuOnBackClicked || () => getHelpStackView.pop()`?

nebula/ui/components/VPNViewBase.qml
```diff
@@ -18,11 +18,10 @@ Item {
...
+//   height: parent.height
```
flozia, line 21: `// :)`

src/ui/screens/ScreenGetHelp.qml
MattLichtenstein: We could consider exposing `constants.cpp` to the QML engine, or creating a frontend only constants.js file (similar to themes.js)

flozia: Thinking about it again: It might be nice to have, but let’s do it in case there is a need for that in other places as well. Doing this for the developer options as an isolated case might be not the best incentive.

flozia (approved):

lesleyjanenorton (changes requested):
I'm not sure what about `Layout.fillWidth` it is that sometimes causes this - but _something_ about the `Layout.fillWidth` change on L54 in `ScreenGetHelp` causes a regression where part of the UI goes off-screen after transitioning between portrait and landscape modes a few times on iPad. Replacing that line with `Layout.preferredWidth: parent.width` fixes it.

Issue #4937, opened 2022-11-15
Update from Mozilla VPN 2.10 to 2.11 not working on Production server


data-sync-user:
**Affected versions:**

* Mozilla VPN 2.10.0 (2.202210211057);

**Tested Platforms:**

* Windows 10/11;

**Prerequisites:**

* Have Mozilla VPN **2.10** version installed and opened;
* Have a Fx account with an active subscription for Mozilla VPN;

* Be signed in - **Production** server;

**Steps to reproduce:**

1. Go to Settings - About us and click the “Check for updates” button;
1. Observe the behavior;

**Expected result:**

* “Update available” modal is displayed;

**Actual result:**

* “You are up to date” modal is displayed;

**Notes:**

* Quitting the app does not solve the issue;

* The update is working as expected if starting from the 2.9 version;

* Not reproducible on Stage server;

* See the attached video and logs;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3255)
┆Reporter: Bianca Hidecuti

data-sync-user:
➤ Bianca Hidecuti commented:

update: I am also able to reproduce it on MacOS while using the 2.10 build (Prod sever). On MacOS I was able to see the “Update available” modal after turning ON the VPN.

!Screen Recording 2022-11-15 at 13.50.27.mov|width=414,height=686!

data-sync-user:
➤ Sarah Bird commented:

There’s a discussion with the releng team here: https://mozilla.slack.com/archives/C01DCUKG95E/p1668538988105999 ( https://mozilla.slack.com/archives/C01DCUKG95E/p1668538988105999|smart-link )

It is expected that balrog may return a cached value (so no update) for up to 24 hours after the change.

Please check again on the morning of Nov 16 Romanian time. If you are still consistently getting this issue then there’s a problem.

data-sync-user:
➤ Bianca Hidecuti commented:

Sarah Bird, the issue is still reproducible while using the 2.10 version → “You are up to date” modal is displayed.

The update from 2.9 to 2.11 works as expected. 

!bandicam 2022-11-16 10-30-33-928.mp4|width=464,height=828!

[^mozillavpn-2022-11-16.txt]

data-sync-user:
➤ Jon Buckley commented:

[...]

PR #4968, opened 2022-11-19
Update windows/taskcluster builds to go1.18.8

oskirby:
## Description
During the conversion to CMake, we began relying on the environment's golang version when building `tunnel.dll` and `balrog.dll`. This had a wonderful effect on build times, but we recently found that the resulting `balrog.dll` seems to have issues parsing certificates, and results in breakage of the update flow.

In my local builds, updating to the latest stable go1.18 seems to resolve the issue.

## Reference
Github issue #4937 ([VPN-3255](https://mozilla-hub.atlassian.net/browse/VPN-3255))

## Checklist
    
### DIFF
--- a/balrog/build.cmd
+++ b/balrog/build.cmd
@@ -14,7 +14,7 @@ if exist .deps\prepared goto :build
 	rmdir /s /q .deps 2> NUL
 	mkdir .deps || goto :error
 	cd .deps || goto :error
-	call :download go.zip https://dl.google.com/go/go1.13.3.windows-amd64.zip 9585efeab37783152c81c6ce373b22e68f45c6801dc2c208bfd1e47b646efbef || goto :error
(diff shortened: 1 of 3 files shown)


hneiva (approved):
assuming you've tested those URLs :)

oskirby:
@hneiva I have tested these URLs to the best of my ability, and they are the same ones I am using for my local builds now, and the domain is signed by Google Trust Services, if that matters.

However, looking under the hood, it seems that URL actually redirects to https://dl.google.com/go/go1.18.8.windows-amd64.zip, would that be a more appropriate URL to use in lieu of the `go.dev` domain?

bakulf (approved):

hneiva:
> @hneiva I have tested these URLs to the best of my ability, and they are the same ones I am using for my local builds now, and the domain is signed by Google Trust Services, if that matters.
> 
> However, looking under the hood, it seems that URL actually redirects to https://dl.google.com/go/go1.18.8.windows-amd64.zip, would that be a more appropriate URL to use in lieu of the `go.dev` domain?

I personally don't think it would be better to go directly to their storage URL.
We check the hash of the binary, so following whatever CDN they have in front of it will avoid some headaches and it's still safe.

PR #4979, opened 2022-11-21
Mfbt - checkedInt - VPN-3315

bakulf:


src/mfbt/mozilla/CheckedInt.h
```diff
@@ -0,0 +1,800 @@
+/* -*- Mode: C++; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
```
brizental, line 1: Please add comments to this file and IntegerTypeTraits, mentioning they are copy pasted from m-c and linking to the copied files.

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
```diff
@@ -4,23 +4,24 @@
 
 #include "addonconditionwatchertimeend.h"
 #include "leakdetector.h"
+#include "mfbt/checkedint.h"
 #include "settingsholder.h"
 
 #include <QDateTime>
 
 AddonConditionWatcherTimeEnd::AddonConditionWatcherTimeEnd(QObject* parent,
```
brizental, line 12: Is it possible to merge `AddonConditionWatcherTimeEnd` and `AddonConditionWatcherTimeStart` into a single class? They seem to be mostly redundant.

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
```diff
@@ -4,23 +4,24 @@
...
+  connect(&m_timer, &QTimer::timeout, this, [this]() {
+    if (maybeStartTimer()) {
+      emit conditionChanged(false);
+    }
+  });
 
-    connect(&m_timer, &QTimer::timeout, this,
-            [this]() { emit conditionChanged(false); });
-  }
+  maybeStartTimer();
```
brizental, line 24: Don't we need to emit a `conditionChanged` if this returns `true`? What is the condition is already met when the condition watcher is started?

src/addons/conditionwatchers/addonconditionwatchertriggertimesecs.cpp
```diff
@@ -20,21 +21,17 @@ AddonConditionWatcher* AddonConditionWatcherTriggerTimeSecs::maybeCreate(
 
 AddonConditionWatcherTriggerTimeSecs::AddonConditionWatcherTriggerTimeSecs(
```
brizental, line 22: This one is also redundant with the other time condition watchers.

src/models/devicemodel.h
```diff
@@ -42,7 +42,7 @@ class DeviceModel final : public QAbstractListModel {
 
   void removeDeviceFromPublicKey(const QString& publicKey);
 
-  int activeDevices() const { return m_devices.count(); }
+  int activeDevices() const { return static_cast<int>(m_devices.count()); }
```
brizental, line 45: n00b question: Why do we need the static_cast here? Weren't we returning an `int` before already?

src/models/licensemodel.cpp
```diff
@@ -60,8 +60,8 @@ void LicenseModel::initialize() {
 
   beginResetModel();
 
-  for (uint32_t i = 0, length = content.length(); i < length; ++i) {
-    int32_t lineLength = content[i].length();
+  for (qsizetype i = 0, length = content.length(); i < length; ++i) {
+    qsizetype lineLength = content[i].length();
```
brizental, line 64: Is this change actually related to the CheckedInt stuff or is it just a case of some implicit type coersion was going on before and is not anymore?

src/platforms/ios/iosiaphandler.h
```diff
@@ -23,7 +23,7 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
 
  protected:
   void nativeStartSubscription(ProductsHandler::Product* product) override;
-  void nativeRestoreSubscription();
+  void nativeRestoreSubscription() override;
```
brizental, line 26: How is this change related to the chekedInt change?

tests/unit/testcheckedint.cpp
```diff
@@ -0,0 +1,596 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: If these are also copied from m-c, please also add a comment :)

brizental (commented):
I left a few clarificaiton questions. Not sure I fully grasp what was done here yet :) Also, what does `mfbt` stand for?

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
bakulf: follow up?

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
bakulf: no signals in the CTOR. Nobody will receive them.

src/addons/conditionwatchers/addonconditionwatchertriggertimesecs.cpp
bakulf: Why? this is not a specific time. It's X seconds from the first activation of the client.

src/models/devicemodel.h
bakulf: count() returns a `qsizetype`, which means a `size_t`which is a `unsigned int`.

src/models/licensemodel.cpp
bakulf: I removed a bunch of int warnings too.

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
brizental: Sure!

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
brizental: Ah, ok. But still what happens if the condition is already met when the condition watcher is started?

src/addons/conditionwatchers/addonconditionwatchertriggertimesecs.cpp
brizental: Just because the new code looks like it's the same.

src/models/devicemodel.h
brizental: Ok, but the function signature was not changed, which would mean it was already returning an `int` before. What that some implicit type coersion? What is different about that and the static_cast?

src/models/licensemodel.cpp
brizental: Ah! Got it? Should we be failing CI on warning to avoid having things like these in the future?

src/platforms/ios/iosiaphandler.h
brizental: No need for a separate PR, but I have the same question as above https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4979#discussion_r1029423504

src/models/devicemodel.h
brizental: Nevermind, this was probably to address warnings. Right?

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
bakulf: we call: `conditionApplied()` after the CTOR.

src/models/devicemodel.h
bakulf: Right!

src/models/licensemodel.cpp
bakulf: These warnings are shown only on iOS...

src/addons/conditionwatchers/addonconditionwatchertimeend.cpp
bakulf: VPN-3333

src/models/licensemodel.cpp
brizental: Sheesh, ok yeah let's forget about it now then 👀

brizental (approved):

PR #4987, opened 2022-11-21
Fix tab navigation glitch / VPN-2850

lesleyjanenorton:
## Description

This fixes the glitch in `VPNTabNavigation` when the `currentItem` changes, which was caused by explicitly assigning a `width` to tab buttons. There is a trade-off: It also removes the sliding "active tab" indicator. This is because the sliding indicator relied on the `x` value of the tab buttons, which becomes a rubbish value when the tab button `width` is not explicitly set. If we feel strongly about the sliding indicator I am happy to open a ticket to re-implement.

https://user-images.githubusercontent.com/22355127/203120760-be36d138-5d4c-4f6b-ba29-719ef5c0d973.mov

## Reference

VPN-2850

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNTabNavigation.qml
+++ b/nebula/ui/components/VPNTabNavigation.qml
@@ -47,7 +47,6 @@ Item {
                 id: btn
                 objectName: tabButtonId
                 height: bar.contentHeight
-                width: stack.children.length > 0 ? (bar.width / stack.children.length) : bar.width
                 onClicked: handleTabClick(btn)
 
                 background: Rectangle {
@@ -59,7 +58,7 @@ Item {
                         anchors.right: parent.right
                         anchors.bottom: parent.bottom
                         color: VPNTheme.colors.purple70
-                        opacity: btn.activeFocus ? 1 : 0
+                        opacity: btn.checked || btn.activeFocus ? 1 : 0
                         Behavior on opacity {
                             PropertyAnimation {
                                 duration: 100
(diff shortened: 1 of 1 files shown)


MattLichtenstein:
Did some digging here and realized the `x` value for each of the tab buttons was weirdly negative. Anyways, I found [this](https://stackoverflow.com/a/73287249) so it doesn't appear to be a bug, and it seems if we don't set the `width` of the TabButton's, remove the rectangle from the TabButton repeater, and set the `x` value of the sliding rectangle to `currentTab.x - currentTab.ListView.view.originX`, this should be working as expected with the sliding functionality.

lesleyjanenorton:
> Did some digging here and realized the `x` value for each of the tab buttons was weirdly negative. Anyways, I found [this](https://stackoverflow.com/a/73287249) so it doesn't appear to be a bug, and it seems if we don't set the `width` of the TabButton's, remove the rectangle from the TabButton repeater, and set the `x` value of the sliding rectangle to `currentTab.x - currentTab.ListView.view.originX`, this should be working as expected with the sliding functionality.

This worked!

birdsarah (approved):
I've run this locally and confirmed it's working. I don't have anywhere close to enough qml knowledge to know if this is the best solution but it looks reasonable enough albeit a bit magical :)

MattLichtenstein:
@lesleyjanenorton Are we able to remove the rectangle inside the repeater?

lesleyjanenorton:
> @lesleyjanenorton Are we able to remove the rectangle inside the repeater?

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b2bccec40ee60f10dcdab146015fef10c2f66447/nebula/ui/components/VPNTabNavigation.qml#L53 
@MattLichtenstein this one? Not without working out another way to overwrite the default background color.

MattLichtenstein:
@lesleyjanenorton this one. It seems redundant

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b2bccec40ee60f10dcdab146015fef10c2f66447/nebula/ui/components/VPNTabNavigation.qml#L56-L67

PR #4991, opened 2022-11-22
Improve error handling + Telemetry for alerts - VPN-3320

bakulf:


glean/metrics.yaml
```diff
@@ -1627,4 +1627,29 @@ sample:
...
+      - amarchesini@mozilla.com
+    expires: never
+    extra_keys:
+      task:
+        description: The name of the task that has generated this error.
+        type: string
+      context:
+        description: |
+          An extra string to describe more in detail the source of the error.
+        type: string
```
brizental, line 1655: As discussed offline, I think this is a bit too generic. Let's split this into line number / file name.

glean/metrics.yaml
```diff
@@ -1627,4 +1627,29 @@ sample:
...
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - amarchesini@mozilla.com
+    expires: never
+    extra_keys:
+      task:
+        description: The name of the task that has generated this error.
+        type: string
+      filename_and_line:
```
brizental, line 1652: Can we split this into two different extra keys? one that is a string "filename" and another that is a number "line"?

brizental (approved):

PR #5002, opened 2022-11-22
Addon condition cleanup - VPN-3333

bakulf:


brizental (approved):
Do these condition watcher not have any tests? If not, would you mind filing a ticket to write them?

PR #5009, opened 2022-11-23
Custom code to retrieve the iOS language - VPN-3336

bakulf:
In theory QLocale::system()::uiLanguages() should return the list of ui-languages sorted. But, for some languages (es), on IOS, en-US is on top of the list even when it should not be.

### DIFF
--- a/src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
+++ b/src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
@@ -39,13 +39,8 @@ AddonConditionWatcherLocales::~AddonConditionWatcherLocales() {
 }
 
 bool AddonConditionWatcherLocales::conditionApplied() const {
-  QString code = SettingsHolder::instance()->languageCode();
-  if (code.isEmpty()) {
-    code = Localizer::systemLanguageCode();
-    if (code.isEmpty()) {
-      code = "en";
-    }
-  }
+  QString code = Localizer::instance()->languageCodeOrSystem();
+  Q_ASSERT(!code.isEmpty());
 
   code = Localizer::majorLanguageCode(code);
   return m_locales.contains(code.toLower());
--- a/src/addons/addon.cpp
+++ b/src/addons/addon.cpp
@@ -438,17 +438,12 @@ void Addon::retranslate() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
   Q_ASSERT(settingsHolder);
 
-  QString code = settingsHolder->languageCode();
-
-  QLocale locale = QLocale(code);
-  if (code.isEmpty()) {
-    locale = QLocale(Localizer::systemLanguageCode());
-  }
+  QLocale locale = Localizer::instance()->locale();
 
   if (!m_translator.load(
           locale, "locale", "_",
           QFileInfo(m_manifestFileName).dir().filePath("i18n"))) {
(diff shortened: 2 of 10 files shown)


mcleinman:
I'm unable to replicate the original bug, either on simulator or on device.

Logs from device (with Spanish above English):
```
[01.12.2022 11:01:06.713] Debug: (main - Localizer) System language codes: [es-US,es-Latn-US,en-US,en,en-Latn-US]
[01.12.2022 11:01:06.713] Debug: (main - Localizer) System language code: es-US
```

Is there a specific setup I can replicate the bug on?

bakulf:
See VPN-3336. How I reproduce it is to set the system language to `spanish (european)` - this can be called UUEE or something like that. The result is that the app will speak in English instead of Spanish. But all the other apps are in Spanish on my device.
Debugging this, it turned out that `uiLanguages` returns `en` as the first language instead of `es`.

src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
```diff
@@ -39,13 +39,8 @@ AddonConditionWatcherLocales::~AddonConditionWatcherLocales() {
...
 bool AddonConditionWatcherLocales::conditionApplied() const {
-  QString code = SettingsHolder::instance()->languageCode();
-  if (code.isEmpty()) {
-    code = Localizer::systemLanguageCode();
-    if (code.isEmpty()) {
-      code = "en";
-    }
-  }
+  QString code = Localizer::instance()->languageCodeOrSystem();
+  Q_ASSERT(!code.isEmpty());
```
brizental, line 43: What happens if the code is empty and we are in prod?

mcleinman (changes requested):
For two types of Spanish, my testing (iOS 16, iPhone 12 mini) shows our app in English.

I *think* it's because we're getting the langauges with two dashes rather than a dash and an underscore (`es-Latn_US`), but I could be a little off.

I confirmed the app was set to use the device language.

Logs:
```
[06.12.2022 14:22:29.531] Debug: (main - Localizer) Supported languages: [co,cy,de,dsb,el,en,en_CA,en_GB,es_AR,es_CL,es_ES,es_MX,fi,fr,fy_NL,hsb,hu,ia,id,is,it,ja,nl,pa_IN,pt_BR,pt_PT,ru,sk,sq,sv_SE,uk,zh_CN,zh_TW]
[06.12.2022 14:22:29.531] Debug: (main - Localizer) System language codes: [es-ES,es,es-Latn-ES,en-US,en,en-Latn-US,es-US,es-Latn-US]
[06.12.2022 14:22:29.531] Debug: (main - Localizer) System language code: en
[06.12.2022 14:22:29.531] Debug: (main - Localizer) Loading language:
```

Languages:
IMG_0014

src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
bakulf: languageCodeOrSystem returns the system language code which cannot be empty. Fallback is 'en'.

brizental: I ask because of the Q_ASSERT, if the code cannot be empty then it's unnecessary and confusing. Can we remove it?

bakulf:
New version with proper parsers and unit-tests.

mcleinman (approved):
Looks good to me. Thanks!

PR #5034, opened 2022-11-24
Fix regex for email validation - VPN-3275

Gela:
## Description

    - Address a bug where an input email such as "name@mail-server" was coming back as valid
    - Clean up the regex code for email validation

## Reference

    VPN-3275

## Checklist
    
- [x] I have added thorough tests where needed N/A


src/authenticationinapp/authenticationinapp.cpp
```diff
@@ -240,34 +240,17 @@ void AuthenticationInApp::requestErrorPropagation(
...
   if (emailAddress.isEmpty()) {
     return false;
   }
 
-  QStringList parts = emailAddress.split("@");
-  if (parts.length() != 2 || parts[1].length() > 255) {
-    return false;
-  }
-
-  static QRegularExpression emailRE("^[A-Z0-9.!#$%&'*+/=?^_`{|}~-]{1,64}$",
```
bakulf, line 254: I wonder why fxa supports so many chars in the first part of the email address. Do you mind ask them in the slack channel? If we apply your PR, we go out-of-sync with their impl. Which is fine, if we have good reasons.

oskirby (changes requested):
Can we add some tests to catch the example that validated incorrectly prior to this change, or any other difficult cases that you think might deserve a test vector?

The sources for the unit tests that ought to cover this are in `tests/auth/testemailvalidation.cpp`

src/authenticationinapp/authenticationinapp.cpp
Gela: Copied from our Slack conversation 

> Regarding the email authentication, I posted on Fxa this morning but no responses yet. I may be wrong but I believe their validator regex is buggy as well, for example in the Fxa website it is recognizing "g^[eg3.3@gmail.com](mailto:eg3.3@gmail.com)" as a valid email. I also don't see a benefit in splitting the input into 2 halves (post and pre @) and evaluating separately when it can be done in one pass, which for a short email it's minimal but it adds additional calculation overhead and complicates the logic. What are your thoughts? Personally I'm leaning towards opening an issue with them and have them update their logic rather than VPN using what is definitely a buggy

oskirby (approved):
Thanks, that addresses my concerns about unit testing.

bakulf (approved):

PR #5039, opened 2022-11-24
Start the transaction before the tutorials - VPN-3377

bakulf:


brizental (commented):
Was this a bug? Why was the change necessary? Would you mind adding tests in case this is fixing a bug?

bakulf:
I'll work on a test as a follow up.

brizental (approved):

PR #5040, opened 2022-11-24
Weight random server selection by number of servers.

oskirby:
## Description
Our current algorithm to pick a random location does so in a non-uniform manner by choosing a random country and then choosting a random city within it. This disproportionately causes the selection algorithm to bias unfairly towards locations in smaller countries with fewer cities. For example, `Skopje, Macedonia` with a single server is given an equal proportion of the generated selections as all of the United States combined.

We feel that it is fairer to select a location that is weighted by the number of servers available to it. This should result in a fairer distribution of load that better matches the capacity of a given location.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

bakulf (approved):
Good but it needs rebase

PR #5047, opened 2022-11-25
VPN-3382 - Do not use QVector to pass data to through the FFI boundary

brizental:
This is what the generated code looks like right now:

```cpp
struct AddonStateChangedExtra {
  QVariant _addonId = QVariant();
  QVariant _state = QVariant();

  FfiExtra ToFfiExtra() const {
    QVector<const char*> extraKeys;
    QVector<const char*> extraValues;
    if (_addonId.canConvert<QString>()) {
      extraKeys.append("addon_id");
      extraValues.append(_addonId.toString().toUtf8().constData());
    }
    if (_state.canConvert<QString>()) {
      extraKeys.append("state");
      extraValues.append(_state.toString().toUtf8().constData());
    }

    FfiExtra extras;
    extras.keys = std::move(extraKeys.begin());
    extras.values = std::move(extraValues.begin());
    extras.count = extraValues.count();
    return extras;
  };
};
```

And what it looks like with this change:

```cpp
struct AddonStateChangedExtra {
  QVariant _addonId = QVariant();
  QVariant _state = QVariant();

  FfiExtra ToFfiExtra() const {
    const char* extraKeys[2 + 1];
    const char* extraValues[2 + 1];
    int count = 0;
    if (_addonId.canConvert<QString>()) {
      extraKeys[count] = "addon_id";
      extraValues[count] = _addonId.toString().toUtf8().constData();

      count++;
    }
    if (_state.canConvert<QString>()) {
      extraKeys[count] = "state";
      extraValues[count] = _state.toString().toUtf8().constData();

      count++;
    }

    FfiExtra extras;
    extras.keys = std::move(extraKeys);
    extras.values = std::move(extraValues);
    extras.count = count;
    return extras;
  };
};
```

### DIFF
--- a/vpnglean/glean_parser_ext/templates/cpp.jinja2
+++ b/vpnglean/glean_parser_ext/templates/cpp.jinja2
@@ -41,20 +41,23 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
     const char* extraKeys[{{ obj|attr(name)|length }} + 1];
     const char* extraValues[{{ obj|attr(name)|length }} + 1];
     int count = 0;
+
+    QVector<QByteArray> _keepStringsAlive;
     {% for item, type in obj|attr(name) %}
     if (_{{item|camelize}}.canConvert<{{type|extra_type_name}}>()) {
       extraKeys[count] = "{{item}}";
       {% if type == "string" %}
-      extraValues[count] = _{{item|camelize}}.toString().toUtf8().constData();
+      _keepStringsAlive.append(_{{item|camelize}}.toString().toUtf8());
+      extraValues[count] = _keepStringsAlive[count].constData();
       {% elif type == "boolean" %}
-      extraValues[count] = _{{item|camelize}}.toBool() ? "true" : "false";
+      _keepStringsAlive.append(_{{item|camelize}}.toBool() ? "true" : "false");
+      extraValues[count] = _keepStringsAlive[count];
       {% elif type == "quantity" %}
-      std::string s = std::to_string(_{{item|camelize}}.toInt());
-      extraValues[count] = s.c_str();
+      _keepStringsAlive.append(QString::number(_{{item|camelize}}.toInt()).toUtf8());
+      extraValues[count] = _keepStringsAlive[count].constData();
       {% else %}
 #error "Glean: Invalid extra key type for metric {{obj.category}}.{{obj.name}}, defined in: {{obj.defined_in['filepath']}}:{{obj.defined_in['line']}})"
       {% endif %}
-
       count++;
     }
     {% endfor %}
--- a/vpnglean/glean_parser_ext/templates/cpp.jinja2
+++ b/vpnglean/glean_parser_ext/templates/cpp.jinja2
@@ -45,11 +45,12 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
     QVector<QByteArray> _keepStringsAlive;
     {% for item, type in obj|attr(name) %}
     if (_{{item|camelize}}.canConvert<{{type|extra_type_name}}>()) {
-      extraKeys[count] = "{{item}}";
       {% if type == "string" %}
       _keepStringsAlive.append(_{{item|camelize}}.toString().toUtf8());
       extraValues[count] = _keepStringsAlive[count].constData();
       {% elif type == "boolean" %}
+      // We don't need to actually keep these alive since they are literal strings.
+      // However we want to keep the indexes of keepStringsAlive in sync with the indexes of extraValues, so we keep this.
       _keepStringsAlive.append(_{{item|camelize}}.toBool() ? "true" : "false");
       extraValues[count] = _keepStringsAlive[count];
       {% elif type == "quantity" %}
@@ -58,6 +59,7 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
       {% else %}
 #error "Glean: Invalid extra key type for metric {{obj.category}}.{{obj.name}}, defined in: {{obj.defined_in['filepath']}}:{{obj.defined_in['line']}})"
       {% endif %}
+      extraKeys[count] = "{{item}}";
       count++;
     }
     {% endfor %}
--- a/vpnglean/glean_parser_ext/templates/cpp.jinja2
+++ b/vpnglean/glean_parser_ext/templates/cpp.jinja2
@@ -38,19 +38,20 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
   {% endfor %}
 
   FfiExtra ToFfiExtra() const {
+    m_keepStringsAlive.clear();
+
     const char* extraKeys[{{ obj|attr(name)|length }} + 1];
     const char* extraValues[{{ obj|attr(name)|length }} + 1];
     int count = 0;
 
-    QVector<QByteArray> _keepStringsAlive;
     {% for item, type in obj|attr(name) %}
     if (_{{item|camelize}}.canConvert<{{type|extra_type_name}}>()) {
       {% if type == "string" %}
       QByteArray s = _{{item|camelize}}.toString().toUtf8();
-      _keepStringsAlive.append(s);
+      m_keepStringsAlive.append(s);
       extraValues[count] = s.constData();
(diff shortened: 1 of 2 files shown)


vpnglean/glean_parser_ext/templates/cpp.jinja2
```diff
@@ -38,28 +38,31 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
   {% endfor %}
 
   FfiExtra ToFfiExtra() const {
-    QVector<const char*> extraKeys;
-    QVector<const char*> extraValues;
+    const char* extraKeys[{{ obj|attr(name)|length }} + 1];
+    const char* extraValues[{{ obj|attr(name)|length }} + 1];
```
brizental, line 42: Is it an issue that I create the array as large as it can be, but then calculate the actual count of elements and that is what I tell Rust is the actualy size of this array? (Sorry if this doesn't make sense)

vpnglean/glean_parser_ext/templates/cpp.jinja2
```diff
@@ -38,28 +38,31 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
...
+    int count = 0;
     {% for item, type in obj|attr(name) %}
     if (_{{item|camelize}}.canConvert<{{type|extra_type_name}}>()) {
-      extraKeys.append("{{item}}");
+      extraKeys[count] = "{{item}}";
       {% if type == "string" %}
-      extraValues.append(_{{item|camelize}}.toString().toUtf8().constData());
+      extraValues[count] = _{{item|camelize}}.toString().toUtf8().constData();
```
bakulf, line 48: I wonder when this string is deleted. What does it keep it alive? `toString` returns a temporary obj but you don't keep it alive, and probably the dtor runs. Same for the `toUtf8`.

bakulf (changes requested):

vpnglean/glean_parser_ext/templates/cpp.jinja2
brizental: Is this relevant given that I do a `std::move` below?

vpnglean/glean_parser_ext/templates/cpp.jinja2
```diff
@@ -38,28 +38,34 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
...
     if (_{{item|camelize}}.canConvert<{{type|extra_type_name}}>()) {
-      extraKeys.append("{{item}}");
+      extraKeys[count] = "{{item}}";
       {% if type == "string" %}
-      extraValues.append(_{{item|camelize}}.toString().toUtf8().constData());
+      _keepStringsAlive.append(_{{item|camelize}}.toString().toUtf8());
+      extraValues[count] = _keepStringsAlive[count].constData();
       {% elif type == "boolean" %}
-      extraValues.append(_{{item|camelize}}.toBool() ? "true" : "false");
+      _keepStringsAlive.append(_{{item|camelize}}.toBool() ? "true" : "false");
```
bakulf, line 53: because these are "static" strings, you don't need to keep them alive. YOu can directly do: `extraValues[count] = _{{item|camelize}}.toBool() ? "true" : "false"`

vpnglean/glean_parser_ext/templates/cpp.jinja2
```diff
@@ -38,28 +38,34 @@ struct {{ obj.name|Camelize }}{{ suffix }} {
...
     }
     {% endfor %}
 
     FfiExtra extras;
-    extras.keys = std::move(extraKeys.begin());
-    extras.values = std::move(extraValues.begin());
-    extras.count = extraValues.count();
+    extras.keys = std::move(extraKeys);
+    extras.values = std::move(extraValues);
+    extras.count = count;
```
bakulf, line 68: This will not work because at this point, `keepStringAlive` will be out-of-scope.

bakulf (changes requested):

vpnglean/glean_parser_ext/templates/cpp.jinja2
brizental: Duh, of course! Thanks for catching :)

vpnglean/glean_parser_ext/templates/cpp.jinja2
brizental: But this is happening at the same scope as keepStringsAlive :thinking:

brizental: Here is an example of the generated code:
```cpp
FfiExtra ToFfiExtra() const {
  const char* extraKeys[2 + 1];
  const char* extraValues[2 + 1];
  int count = 0;

  QVector<QByteArray> _keepStringsAlive;
  if (_addonId.canConvert<QString>()) {
    extraKeys[count] = "addon_id";
    _keepStringsAlive.append(_addonId.toString().toUtf8());
    extraValues[count] = _keepStringsAlive[count].constData();
    count++;
  }
  if (_state.canConvert<QString>()) {
    extraKeys[count] = "state";
    _keepStringsAlive.append(_state.toString().toUtf8());
    extraValues[count] = _keepStringsAlive[count].constData();
    count++;
  }

  FfiExtra extras;
  extras.keys = std::move(extraKeys);
  extras.values = std::move(extraValues);
  extras.count = count;
  return extras;
};
```

vpnglean/glean_parser_ext/templates/cpp.jinja2
brizental: Uh oh, actually I can't do this. I had this because I wanted to keep the index of keepStringsAlive in sync with the index of the `extraValues`.

brizental: I added a comment explaining this.

brizental: Hm, but this is going to fail since it is not a QByteArray right. I need to rethink this. The only reason why it did not fail is because we don't have boolean type extras right now.

brizental: Ok yeah now this is addressed. Please see: [9b9e861](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5047/commits/9b9e861d38206440019bab1b9fe2223a9167952c)

vpnglean/glean_parser_ext/templates/cpp.jinja2
bakulf: Right. The result is an array of char* that are not kept alive outside this method. At the end of the method, keepAlive is released and the pointers are freed.

brizental: Ugh, duh. Ok So I need to put that in the outer scope. Thanks for the patience in walking me through this one @bakulf :D

brizental: Tried again, this is what the generated code looks like now:
```cpp
   struct AddonMessageStateChangedExtra {
      QVariant _messageId = QVariant();
      QVariant _messageState = QVariant();

      FfiExtra ToFfiExtra() const {
        m_keepStringsAlive.clear();

        const char* extraKeys[2 + 1];
        const char* extraValues[2 + 1];
        int count = 0;

        if (_messageId.canConvert<QString>()) {
          QByteArray s = _messageId.toString().toUtf8();
          m_keepStringsAlive.append(s);
          extraValues[count] = s.constData();
          extraKeys[count] = "message_id";
          count++;
        }
        if (_messageState.canConvert<QString>()) {
          QByteArray s = _messageState.toString().toUtf8();
          m_keepStringsAlive.append(s);
          extraValues[count] = s.constData();
          extraKeys[count] = "message_state";
          count++;
        }

        FfiExtra extras;
        extras.keys = std::move(extraKeys);
        extras.values = std::move(extraValues);
        extras.count = count;
        return extras;
      };

      private:
        // Helper vector to extend the lifetime of the strings
        // which hold the extra key values until they are used.
        QVector<QByteArray> m_keepStringsAlive;
    };
```

bakulf (approved):

PR #5054, opened 2022-11-28
Taskcluster: Enable Sentry on Osx

strseb:
- This Pr splits our mac build - so we get a real debug and a real release build. 
- Release builds are forced-level-3 therefore we can have secrets. 
- Passing the sentry-keys into the relase build will activate the sentry-reporter :) 
- NIT: swapping the name mac/opt to mac_<cpu_arch> so it's consistent with windows and android.


taskcluster/ci/build/macos.yml
```diff
@@ -27,3 +27,21 @@ macos/opt:
...
+
+macos/opt:
+    description: "Mac Build"
+    treeherder:
+        symbol: B
+    requires-level: 3 
+    scopes:
+        - project:releng:services/tooltool/api/download/public
+        - project:releng:services/tooltool/api/download/internal
+        - secrets:get:project/mozillavpn/tokens
```
brizental, line 47: nit: newline.

taskcluster/scripts/build/macos_build.sh
```diff
@@ -24,6 +37,7 @@ print Y "Installing rust..."
 curl https://sh.rustup.rs -sSf | sh -s -- -y || die
 export PATH="$HOME/.cargo/bin:$PATH"
 
+print Y "Installing homebrew, cmake, ninja..."
```
brizental, line 40: Ninja?

strseb: https://ninja-build.org/ 
On taskcluster i'm using that as a generator  (`-GNinja`) instead of the default one - since it runs on all host's and we can use it for all targets, as it helped me get around msbuild and xcodebuild CI wierndess things :D

bakulf (approved):

mozrokafor (approved):
LGTM

PR #5058, opened 2022-11-28
Addon testing suite - ignore

bakulf:


mozrokafor (approved):
LGTM,

src/addons/manager/addonmanager.h
```diff
@@ -23,6 +23,16 @@ class AddonManager final : public QAbstractListModel {
...
   static QString addonServerAddress();
 
+  static bool signatureVerificationNeeded();
+
+  // This is mainly for testing. All the addons are unloaded and the cache is
+  // reset.
+  void reset();
+
+  // Force a fetch of the addon index manifest instead of waiting for the normal
+  // scheduling.
```
brizental, line 33: nit: Please use the doxygen format for docstring.

src/addons/manager/addonmanager.h
```diff
@@ -40,19 +50,12 @@ class AddonManager final : public QAbstractListModel {
...
   void retranslate();
 
   void forEach(std::function<void(Addon* addon)>&& callback);
 
-  void updateIndex(const QByteArray& index, const QByteArray& indexSignature);
-
-#ifdef UNIT_TEST
-  QStringList addonIds() const;
-#endif
+  void updateIndex(bool status, const QByteArray& index = QByteArray(),
```
brizental, line 57: Can we add a docstring for this one too? `status` is non obvious.

src/tasks/addonindex/taskaddonindex.cpp
```diff
@@ -43,15 +45,18 @@ void TaskAddonIndex::run() {
...
+        QString("%1manifest.json.sig").arg(AddonManager::addonServerAddress());
+    logger.debug() << "Download manifest signature URL:" << manifestSigUrl;
+
+    NetworkRequest* request =
+        NetworkRequest::createForGetUrl(this, manifestSigUrl, 200);
 
     connect(request, &NetworkRequest::requestFailed, this,
             [this](QNetworkReply::NetworkError error, const QByteArray&) {
               logger.error() << "Get addon index signature failed" << error;
+              AddonManager::instance()->updateIndex(false);
```
brizental, line 59: Why is this necessary if the index did not change?

tests/functional/testAddons.js
```diff
@@ -3,54 +3,115 @@
...
     await vpn.waitForElementProperty(
         'VPNAddonManager', 'loadCompleted', 'true');
+  });
 
-    let count = 0;
+  it('Empty addon index', async () => {
+    vpn.resetLastAddonLoadingCompleted();
+    await vpn.setSetting(
+        'addon/customServerAddress',
+        `http://localhost:${constants.ADDON_PORT}/01_empty_manifest/`);
```
brizental, line 23: nit: Let's put the whole URL in a constant.

tests/functional/testTutorials.js
```diff
@@ -27,8 +29,23 @@ describe('Tutorials', function () {
...
+    await vpn.waitForElementProperty(
+        'VPNAddonManager', 'loadCompleted', 'true');
+    vpn.resetLastAddonLoadingCompleted();
+    await vpn.setSetting(
+        'addon/customServerAddress',
+        `http://localhost:${constants.ADDON_PORT}/04_tutorials_basic/`);
+    await vpn.setSetting('addon/customServer', 'true');
+    await vpn.resetAddons();
+    await vpn.waitForCondition(async () => {
+      return await vpn.lastAddonLoadingCompleted();
```
brizental, line 42: No `await` needed this doesn't return a promise.

brizental: Can we just change `resetAddons` to return this condition everytime? This way that promise will only resolve once the reset is complete and also a bit of the complexity will be hidden from the caller.

tests/functional/testAddons.js
```diff
@@ -3,54 +3,115 @@
...
+                 await vpn.getElementProperty('VPNAddonManager', 'count'),
+                 10) === 0;
+    });
+
+    vpn.resetLastAddonLoadingCompleted();
+    await vpn.setSetting(
+        'addon/customServerAddress',
+        `http://localhost:${constants.ADDON_PORT}/03_single_addon/`);
+    await vpn.fetchAddons();
+    await vpn.waitForCondition(async () => {
```
brizental, line 107: Can we add this condition to the fetchAddons function too?

tests/functional/servers/addon.js
```diff
@@ -0,0 +1,110 @@
...
+    };
+    obj[`/${scenario}/manifest.json.sig`] = {
+      status: 404,
+      bodyRaw: '',
+    };
+
+    return obj;
+  },
+
+  createScenarioGenerated(scenario, generatedPath) {
```
brizental, line 81: If this function is not supposed to be called outside of this module, let's leave it our of the exported object.

tests/functional/servers/addon.js
```diff
@@ -0,0 +1,110 @@
...
+
+  set overrideEndpoints(value) {
+    server.overrideEndpoints = value;
+  },
+
+  throwExceptionsIfAny() {
+    server.throwExceptionsIfAny();
+  },
+
+  createScenario(scenario, addonPath) {
```
brizental, line 55: Would you mind addind a docstring here explaining what this function does exactly? It is not objvous what is meant by `scenario`.

brizental (commented):
Left some initial comments. I need to give this another pass later, it's dense. Overall I like the way we are going. This is a great improvement for addon testing capabilities!

src/tasks/addonindex/taskaddonindex.cpp
bakulf: to communicate the end of the operation to tests. If we don't do that, we will never know if the current operation succeeded or not.

tests/functional/testTipsAndTricksIntroModal.js
```diff
@@ -3,13 +3,26 @@
...
 const { homeScreen } = require('./elements.js');
 const vpn = require('./helper.js');
 
 describe('Tips and tricks intro modal', function () {
   this.timeout(60000);
 
   beforeEach(async () => {
+    await vpn.waitForElementProperty(
+        'VPNAddonManager', 'loadCompleted', 'true');
+    vpn.resetLastAddonLoadingCompleted();
```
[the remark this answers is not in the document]
bakulf: To be sure that the first loading was actually completed before starting tests. I moved it in the `resetAddons`

tests/functional/testAddons.js
```diff
@@ -8,49 +8,51 @@ const vpn = require('./helper.js');
 describe('Addons', function() {
   this.ctx.authenticationNeeded = true;
 
-  it('Addons are loaded', async () => {
-    if (!(await vpn.isFeatureFlippedOff('addonSignature'))) {
-      await vpn.flipFeatureOff('addonSignature');
-      assert(await vpn.isFeatureFlippedOff('addonSignature'));
-    }
+  it('Empty addon index', async () => {
+    await vpn.resetAddons('01_empty_manifest');
```
brizental, line 12: nit: I think it would be better if we reset to the empty manifest on the `beforeEach` block everytime and then in the  test cases we defaulted to `fetchAddons`. IIUC right now it is necessary to call reset as the first thing in each test case in order not to leak state from one test to the next. Is that correct? WDYT?

brizental: Maybe we even do that for ALL tests so that addon changes are never leaked no matter what people do? Not sure there a global beforeEach block, if not just ignore.

tests/functional/helper.js
```diff
@@ -419,6 +427,14 @@ module.exports = {
...
   },
 
+  lastAddonLoadingCompleted() {
+    return _lastAddonLoadingCompleted;
+  },
+
+  resetLastAddonLoadingCompleted() {
+    _lastAddonLoadingCompleted = false;
+  },
+
```
brizental, line 437: nit: These two are not necessary because lastAddonLoadingCompleted is not used outside of this module.

tests/functional/helper.js
```diff
@@ -484,6 +500,40 @@ module.exports = {
...
+    await this.setSetting(
+        'addon/customServerAddress', `${constants.ADDON_URL}/${addonPath}/`);
+    await this.setSetting('addon/customServer', 'true');
+
+    const json = await this._writeCommand('reset_addons');
+    assert(
+        json.type === 'reset_addons' && !('error' in json),
+        `Command failed: ${json.error}`);
+
+    await this.waitForCondition(async () => this.lastAddonLoadingCompleted());
```
brizental, line 517: I'm confused. Does the `_writeCommand('reset_addons')` end up having `connect` called again?

brizental (commented):
Ok, I get it now. I like how clean the addon functional tests are looking :100: Left just a few more nits and questions. Otherwise this is looking good.

tests/functional/helper.js
bakulf: no. `_writeCommand()` writes something on a established connected channel (websocket for functional tests, wasm JS for wasm).

tests/functional/testAddons.js
bakulf: At the end of any test we call `hardReset()` that deletes all the settings.

PR #5064, opened 2022-11-29
Improve the settings code via inspector

bakulf:
It turned out that we can cleanup the inspector + settings a lot using QT meta object.

### DIFF
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -82,163 +82,6 @@ struct InspectorSettingCommand {
   std::function<QJsonValue()> m_get;
 };
 
-// The list of settings exposed to the inspector.
-static QList<InspectorSettingCommand> s_settingCommands{
-    // Unsecured-network-alert
-    InspectorSettingCommand{
-        "unsecured-network-alert", InspectorSettingCommand::Boolean,
-        [](const QByteArray& value) {
-          SettingsHolder::instance()->setUnsecuredNetworkAlert(value == "true");
-        },
-        []() {
-          return SettingsHolder::instance()->unsecuredNetworkAlert() ? "true"
-                                                                     : "false";
-        }},
-
-    // Captive portal
-    InspectorSettingCommand{
-        "captive-portal-alert", InspectorSettingCommand::Boolean,
-        [](const QByteArray& value) {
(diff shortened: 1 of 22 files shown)


src/inspector/inspectorhandler.cpp
```diff
@@ -661,20 +511,66 @@ static QList<InspectorCommand> s_commands{
...
 
-          QStringList settings;
-          for (const InspectorSettingCommand& setting : s_settingCommands) {
-            settings.append(setting.m_settingName);
+          QMetaProperty property = metaObject->property(propertyId);
+          Q_ASSERT(property.isValid());
+
+          QJsonValue value;
+          switch (property.type()) {
+            case QVariant::Bool:
```
strseb, line 530: QVariant can handle quite a few cases out of the box if we don't want to manually limit the types: 

```
QMetaProperty property = metaObject->property(propertyId);
QVariant prop_value = property.read(settingsHolder);
if (prop_value.canConvert<QJsonValue>()) {
  obj["value"] = prop_value.toJsonValue();
  return obj;
}
else if (prop_value.canConvert<QJsonArray>()){
   obj["value"] = prop_value.toJsonArray();
  return obj;
}
switch (property.type()) {
            case QVariant::DateTime: // <- Really not sure if this is already supported out of the box.
            .....
}
```

src/inspector/inspectorhandler.cpp
```diff
@@ -610,49 +428,81 @@ static QList<InspectorCommand> s_commands{
...
+            return obj;
+          }
+
+          QMetaProperty property = metaObject->property(propertyId);
+          Q_ASSERT(property.isValid());
+
+          QVariant value;
+          switch (property.type()) {
+            case QVariant::Bool:
+              if (arguments[2] == "true") {
```
strseb, line 452: Nit: We could try to do: 
```
auto input = QVariant(arguments[2]);
if(input.canConvert(property.type()){
 if!(property.write(settingsHolder,input)){
 // FAILED TO WRITE
 }
}else{
 TYPE Mismatch
}
```

src/inspector/inspectorhandler.cpp
bakulf: oh! nice!!

brizental (approved):

PR #5065, opened 2022-11-29
Test settings rollback

bakulf:


tests/functional/servers/addon.js
```diff
@@ -16,8 +16,8 @@ function createScenario(scenario, addonPath) {
   if (!fs.existsSync(generatedPath)) {
     const manifestPath = path.join(addonPath, 'manifest.json');
     if (!fs.existsSync(manifestPath)) {
-      throw new Error(
-          `No generated and not manifest file! ${manifestPath} should exist!`);
+      throw new Error(`No generated and not manifest file! ${
```
brizental, line 19: This error message is confusing. What does it mean?

brizental (approved):
r+wc

tests/functional/servers/addon.js
bakulf: the addons for testing need to be generated. The CI does it automatically, but what about local execution? I want to have a message to communicate what the developer has to do.

PR #5079, opened 2022-11-29
Update via web for 2.10/2.11 windows clients

bakulf:
This is the change in the addon to force the update via web for windows if the range is [2.10.0, 2.11.0]

### DIFF
--- a/addons/message_update_v2.12/update.js
+++ b/addons/message_update_v2.12/update.js
@@ -1,3 +1,34 @@
 ((api) => {
-  api.navigator.requestScreen(api.navigator.ScreenUpdateRecommended);
+  // windows v2.10/v2.11 do require a web-based update.
+  if (api.env.platform != 'windows') {
+    api.navigator.requestScreen(api.navigator.ScreenUpdateRecommended);
+    return;
+  }
+
+  const parts = api.env.versionString.split('.');
+
+  // No idea which version we are in...
+  if (parts.length < 3) {
+    api.navigator.requestScreen(api.navigator.ScreenUpdateRecommended);
+    return;
+  }
+
+  const version = parts.map(a => parseInt(a, 10));
+
+  function versionCompare(a, b) {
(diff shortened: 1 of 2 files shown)


addons/message_update_v2.12/update.js
```diff
@@ -1,3 +1,34 @@
...
+      if (a[i] != b[i]) {
+        return a[i] > b[i] ? -1 : 1;
+      }
+    }
+    return 0;
+  }
+
+  if (versionCompare([2, 11, 1], version) >= 0 ||
+      versionCompare([2, 10, 0], version) < -1) {
+    api.navigator.requestScreen(api.navigator.ScreenUpdateRecommended);
```
oskirby, line 29: In this case, the balrog code should be working just fine. Shouldn't we simply doing nothing and leave the update recommendation screen to Balrog?

oskirby (commented):
I have mild trepidation that we are going to pop open a browser session directly to the 2.12 update page every time the client is opened. Is there a mechanism to dismiss this, users may find this to be somewhat jarring.

addons/message_update_v2.12/update.js
bakulf: > I have mild trepidation that we are going to pop open a browser session directly to the 2.12 update page every time the client is opened. Is there a mechanism to dismiss this, users may find this to be somewhat jarring.

this script runs only when the user clicks the button. This is the CTA for the button, see the manifest.json.

bakulf: > In this case, the balrog code should be working just fine. Shouldn't we simply doing nothing and leave the update recommendation screen to Balrog?

Correct. This is what that screen does. That screen triggers balrog and the self-update.

oskirby: Excellent, that addresses my concerns.

oskirby (approved):

Issue #4877, opened 2022-11-08
Update the build test client


data-sync-user:
From github action:
```
Build Test Client
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-3204)
┆Reporter: Andrea Marchesini

data-sync-user:
➤ Adrienne Davenport commented:

The deadline for making this update is May 31, 2023

PR #5080, opened 2022-11-29
Functional test cleanup and speedups

oskirby:
## Description
I have become frustrated at the functional tests taking 25 minutes to build. Let's see if we can do better.

Here is a quick breakdown of what I changed, and why it helped:
- Updating the github runners to `ubuntu-22.04` means that we no longer need to bring in a PPA for Qt6 support, and we can instead get those packages from the official repositories, which have much better caching and CDNs behind them.
- Caching `grcov` between runs saves us about 5 minuntes of compilation time.
- Minimizing the profiling data dramatically reduces the storage and bandwidth required to move the build artifacts around (3.5GB down to 250MB!).
- Now that the PPA is no longer necessary, copying the installed `.deb` files from build to test runners is actually slower than just installing fresh dependencies from the official repositories.
- Building the unit tests with Ninja does a much better job of building in parallel.

Some other room for improvement that I did not address:
- Building the addons could be done during the build stage. This would save CPU time, but not calendar time, and would be a more representative test.
- Ninja is way faster for Windows too, but it falls apart somewhere in the rust libraries.
- Do something with `ccache` between workflow runs? That would be outrageously fast!

## Reference
Github #4877 ([VPN-3204](https://mozilla-hub.atlassian.net/browse/VPN-3204))

## Checklist
    
### DIFF
--- a/.github/workflows/functional_tests.yaml
+++ b/.github/workflows/functional_tests.yaml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   build_test_app:
     name: Build Test Client
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     outputs:
       matrix: ${{ steps.testGen.outputs.tests }}
     steps:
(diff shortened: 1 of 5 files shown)


oskirby:
There are still some efficiency gains to be found in moving addon generation into the build stage, but that won't really affect the calendar time much and I'm a little bit intimidated by messing around with the addon build scripts for now.

oskirby:
I am also genuinely shocked at how long the unit tests take to build.

bakulf (approved):

PR #5083, opened 2022-11-29
Change LAN Bypass default value to true

MattLichtenstein:
## Description

- LAN Bypass will now be set on by default for Windows, macOS, Linux, and Android
- a) new users and b) existing users who have never manually toggled this setting, will have LAN bypass turned on by default
- Users who haved manually toggled this setting in the past, either to on or off, will not have their settings changed from their preference

## Reference

[VPN-3229: Make the LAN outgoing traffic be default-allowed on Windows, Linux, macOS and Android.](https://mozilla-hub.atlassian.net/browse/VPN-3229)

## Checklist
    
### DIFF
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -280,7 +280,7 @@ SETTING_BOOL(localNetworkAccess,        // getter
              removeLocalNetworkAccess,  // remover
              hasLocalNetworkAccess,     // has
              "localNetworkAccess",      // key
-             false,                     // default value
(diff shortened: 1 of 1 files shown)


MattLichtenstein:
@bakulf since it appears we are [getting rid of the the LAN Bypass setting altogether](https://mozilla-hub.atlassian.net/browse/VPN-3232), should we just remove the setting instead?

bakulf (approved):

PR #5085, opened 2022-11-30
Fix nits in v2.12 update/whasnew messages, update CODEOWNERS

flodolo:
## Description

I was once again ignored by CODEOWNERS in #5072, so I'm updating the order. From [GitHub docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners):

```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner
```

As for the review/fixes:
* `Opt in` is the verb, `opt-in` is the noun (like `log-in`/`log in`)
* `Multi-hop` is [always spelled](https://transvision.flod.org/?recherche=multi-hop&repo=vpn_client&sourcelocale=en&locale=en&search_type=strings_entities) with lowercase h in the product

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,8 +1,8 @@
-# Francesco Lodolo as main contact for string changes
-addons/**/manifest.json @flodolo
-translations/strings.yaml @flodolo
-
 # Beatriz Rizental as main contact for addon related changes
 /addons/ @brizental
 /src/addons/ @brizental
 /signature/ @brizental
+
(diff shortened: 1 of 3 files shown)


bakulf (approved):

bakulf:
I changed the configuration to require reviews from code owners before landing PRs.

PR #5086, opened 2022-11-30
Experiment: Merge functional test coverage results

oskirby:
## Description
I have a hypothesis: is the cause of our erratic codecov reports due to the parallel nature of our functional tests? What if we were to wait until the very end of the tests, gather all the profiling data together and uploading a single test report. Let's give it a try.

## Checklist
    

oskirby:
@mozrokafor I don't think this experiment was successful since I still observed a bunch of coverage variability even after merging the results. Is it okay if I close and discard this PR?

PR #5090, opened 2022-11-30
Fix the Android-Tile Service

strseb:
This is a 3part patch to fix the issues QA had. Most of those are caused of me merging a patch that was written a year ago 😅 

## Patch 1: Fix bitrotting and let the Tile Query connection status
Well the PR was quite old on merge and was still under the assumption the Service will push the Current State on connect. Now Service-Clients need to request this themselves. 

## Patch 2: Retrive the Config from Storage on a Fresh start
The new `reconnect` function used for the toggle did not account that the Service might have been killed between activations - It now get's the last config from storage, like always-on.

## Patch 3: Polish the Notification fallback messages
This is really Polish. Currently the client has quite a bit of dept around notifications. It depends on the Client beeing alive and pushing the right notification text. 
This PR put's some band-aid on that, so both Tile-Service and Always-On at least show the translated "connected/disconnected" messages, when we have a vpn-session without the client.

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
@@ -84,14 +84,13 @@ class NotificationUtil {
         val prefs = Prefs.get(context)
         prefs.edit()
             .putString("fallbackNotificationHeader", content.getString("productName"))
-            .putString("fallbackNotificationMessage", content.getString("idleText"))
+            .putString("fallbackNotificationMessage", content.getString("connectedText"))
+            .putString("fallbackHideNotificationMessage", content.getString("disconnectedText"))
             .apply()
 
         val channelName = content.getString("notification_group_name")
         val channelDescription = ""
         updateNotificationChannel(channelName, channelDescription)
-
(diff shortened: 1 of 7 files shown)


brizental (approved):

brizental:
```
[task 2022-11-30T12:53:29.986Z] Collecting redo==2.0.4
[task 2022-11-30T12:53:30.001Z]   Downloading redo-2.0.4-py2.py3-none-any.whl (12 kB)
[task 2022-11-30T12:53:30.003Z] ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
[task 2022-11-30T12:53:30.003Z]     redo==2.0.4 from https://files.pythonhosted.org/packages/06/c3/9a11ba6bd67be26e413e14a8ce765af72e493c384c894fc910e50f40692d/redo-2.0.4-py2.py3-none-any.whl (from -r taskcluster/requirements.txt (line 81)):
[task 2022-11-30T12:53:30.003Z]         Expected sha256 81066955041c853b0e6491eb65a0877dce45131c4cfa3d42d923fc2aa8f7a043
[task 2022-11-30T12:53:30.003Z]              Got        c76e4c23ab2f8840261736a851323cd98493710e7a9d36a1058535dca501f293
```

PR #5092, opened 2022-11-30
Update taskgraph?

strseb:
Used: 
```pip-compile --generate-hashes --output-file requirements.txt requirements.in```
To recompile the lockfile so that the file hash of redo is correct. See error in: 
https://firefox-ci-tc.services.mozilla.com/tasks/JHHifleAQ76g15aUbLRYYw


ahal (approved):
Lgtm!

PR #5094, opened 2022-11-30
Refactoring the recent-connection handling - VPN-3410

bakulf:


nebula/ui/components/VPNRecentConnections.qml
```diff
@@ -8,61 +8,18 @@ import QtQuick.Layouts 1.14
...
-                accessibleLabel
-            });
-        }
-        return recentConnections
+    VPNFilterProxyModel {
+        property int count: 0
+        id: recentConnectionsModel
+        source: VPNRecentConnectionModel
+        filterCallback: (obj) => { return obj.isMultiHop === showMultiHopRecentConnections && ++count < 3; }
     }
```
MattLichtenstein, line 23: This seems like it will return the first 2 items in the recent connections model, but I think we actually want the 2nd and 3rd items, otherwise we will be showing the currently selected server as a recent connection

MattLichtenstein (approved):

src/models/recentconnections.h
```diff
@@ -0,0 +1,127 @@
...
+  void serverChanged();
+
+  void readSettings();
+  void writeSettings();
+
+  bool migrate();
+
+ private:
+  RecentConnectionModel m_singleHopModel;
+  RecentConnectionModel m_multiHopModel;
```
MattLichtenstein, line 124: I like that these are separated now!

PR #5095, opened 2022-11-30
Update README.md for the functional tests + addons

bakulf:


README.md
```diff
@@ -506,14 +506,11 @@ ctest --test-dir build
 
 * Install node (if needed) and then `npm install` to install the testing
   dependencies
-* Install geckodriver and ensure it's on your path.
-  [Docs](https://www.selenium.dev/documentation/getting_started/installing_browser_drivers/)
+* Compile the testing addons: ./scripts/addon/generate_all_tests.py
```
brizental, line 509: Is it not possible to add this as a step to the cmake config?

bakulf: @oskirby what do you think? I guess so. Follow up?

brizental: Would you mind filing a ticket? Owen is on PTO, they might take a while to answer :)

brizental (approved):

PR #5096, opened 2022-11-30
Remove LAN Bypass setting

MattLichtenstein:
## Description

- Remove LAN Bypass UI
- Remove LAN Bypass setting
- Remove "Connecting to external devices" guide (from repo)
- [iOS and macOS] Change "Network settings" button text to "DNS settings" on settings menu, and link directly into the "Advanced DNS settings" view 

## Reference

[VPN-3229: Make the LAN outgoing traffic be default-allowed on Windows, Linux, macOS and Android.](https://mozilla-hub.atlassian.net/browse/VPN-3229)
[VPN-3232: Remove visibility to the LAN bypass feature currently shown to users on the Network Settings.](https://mozilla-hub.atlassian.net/browse/VPN-3232)
[VPN-3369: Remove the "Connecting to other devices while using the VPN" Guide](https://mozilla-hub.atlassian.net/browse/VPN-3369)

## Checklist
    

flodolo (approved):

src/dnshelper.cpp
```diff
@@ -126,13 +126,5 @@ bool DNSHelper::shouldExcludeDNS() {
     return false;
   }
 
-  if (isLocalDNS && settings->localNetworkAccess()) {
-    // DNS is lan, but we already excluded local-ip's, all good.
-    return false;
-  }
-
-  if (!isLocalDNS) {
```
bakulf, line 134: This should be: `return false;` Why true?

bakulf (changes requested):

src/dnshelper.cpp
MattLichtenstein: I've updated the function to never exclude local dns (return false for all local dns) because we're either a) on a platform that always excludes them elsewhere or b) on a platform that can't support excluding them (iOS)

Does that look right now?

src/dnshelper.cpp
```diff
@@ -120,19 +120,10 @@ bool DNSHelper::shouldExcludeDNS() {
...
-  if (isLocalDNS && settings->localNetworkAccess()) {
+  if (isLocalDNS) {
     // DNS is lan, but we already excluded local-ip's, all good.
     return false;
   }
 
-  if (!isLocalDNS) {
-    return false;
-  }
   return true;
```
bakulf, line 128: I'm talking about this. If we are here, the IP address is not part of LAN. And we should not exclude it. This should be `return false`. And at this point we can remove this method entirely.

bakulf (changes requested):

src/dnshelper.h
```diff
@@ -13,8 +13,6 @@ class DNSHelper final {
 
   static bool validateUserDNS(const QString& dns);
 
-  static bool shouldExcludeDNS();
-
  private:
   static bool isMullvadDNS(const QString& address);
```
bakulf, line 17: you can remove this too.

bakulf (approved):

PR #5125, opened 2022-12-05
Modularize the VPN client

bakulf:


src/moduleholder.cpp
```diff
@@ -0,0 +1,38 @@
...
+  MVPN_COUNT_CTOR(ModuleHolder);
+}
+
+ModuleHolder::~ModuleHolder() { MVPN_COUNT_DTOR(ModuleHolder); }
+
+void ModuleHolder::registerModule(const QString& moduleName,
+                                  Module* moduleObj) {
+  Q_ASSERT(!m_modules.contains(moduleName));
+  m_modules.insert(moduleName, moduleObj);
+  insert(moduleName, QVariant::fromValue(moduleObj));
```
strseb, line 30: Have you considered using [ dynamic properties? ](https://doc.qt.io/qt-6/qobject.html#setProperty) ?
That could enable writing
```MZModules.vpn.controller ```
instead of: 
```MZModules["vpn"].controller ```
which looks neat :D

Comment (bakulf) [reply src/moduleholder.cpp:30]: dynamic properties are ignored in QML. Do you know a way to expose them? In my experiments, this does not work.

**Comment by mozrokafor** [comment]
> _ No description provided. _

Its a pretty hefty, pr i wonder if a brief description may be useful here for when its ready for review

**Comment by oskirby** [comment]
One initial comment before I really dig into a review: the directory structure of `module/modulevpn/...` feels a little redundant can that just be `module/vpn/...` instead?

File: src/backendfailurewatcher.h [hunk 2]
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Please add docs to this file.

src/logserializewatcher.h
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Also missing docs.

src/module.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef MODULE_H
+#define MODULE_H
+
+#include "notificationhandler.h"
+
+#include <QObject>
+
+class QTextStream;
+
+class Module : public QObject {
```
brizental, line 14: Please use Doxygen format for docs. Also, it would be helper to have a docstring for the class itself.

src/moduleholder.h
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Missing docs.

src/modules/modulevpn.h
```diff
@@ -0,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Missing docs.

src/mozillavpn.h
```diff
@@ -101,8 +93,13 @@ class MozillaVPN final : public QObject {
   // MozillaVPN is null. It should be used rarely.
   static MozillaVPN* maybeInstance();
 
+  // This method initializes all the sub-objects.
   void initialize();
 
+  // This method starts the scheduling of tasks, the reading of the settings
+  // and the activation of the state machine.
```
brizental, line 100: Please use doxygen format for docs.

src/quitwatcher.h
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Missing docs.

src/updaterequiredwatcher.h
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Missing docs.

brizental (commented):
Just did a quick pass to ask for more docs. I limited the requests to all new class files :)

bakulf:
> One initial comment before I really dig into a review: the directory structure of `module/modulevpn/...` feels a little redundant can that just be `module/vpn/...` instead?

Done!

src/platforms/ios/ioscontroller.mm
```diff
@@ -4,14 +4,14 @@
...
 #include "ioscontroller.h"
 #include "Mozilla_VPN-Swift.h"
-#include "controller.h"
 #include "ipaddress.h"
 #include "leakdetector.h"
 #include "logger.h"
 #include "models/device.h"
 #include "models/keys.h"
 #include "models/feature.h"
 #include "models/server.h"
```
lesleyjanenorton, line 13: ios build wants: `#include "modules/vpn.h"` and `ModuleVPN::instance()->controller();` on L85

src/commands/commandui.cpp
```diff
@@ -544,25 +522,12 @@ int CommandUI::run(QStringList& tokens) {
 
 #if MVPN_IOS && QT_VERSION >= 0x060000 && QT_VERSION < 0x060300
     QObject::connect(qApp, &QCoreApplication::aboutToQuit, &vpn,
-                     &MozillaVPN::quit);
+                     &MozillaVPN::terminate);
```
lesleyjanenorton, line 525: some grumbling from xcode here: `terminate is a private member of MozillaVPN`

nebula/ui/components/VPNConnectionInfoScreen.qml
```diff
@@ -51,9 +51,9 @@ Rectangle {
         },
         State {
             name: "open-loading"
-            when: (VPNConnectionBenchmark.state !== VPNConnectionBenchmark.StateInitial
-                && VPNConnectionBenchmark.state !== VPNConnectionBenchmark.StateReady
-                && VPNConnectionBenchmark.state !== VPNConnectionBenchmark.StateError)
+            when: (MZModules["vpn"].connectionBenchmark.state !== MZModules["vpn"].connectionBenchmark.StateInitial
```
lesleyjanenorton, line 54: If this starts to feel verbose we could do something like `property var connectionBenchmark = MZModules["vpn"].connectionBenchmark` at the top then `connectionBenchmark.state` throughout.

src/backendfailurewatcher.h
```diff
@@ -0,0 +1,36 @@
...
+class BackendFailureWatcher final : public QObject {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(BackendFailureWatcher)
+
+ public:
+  explicit BackendFailureWatcher(QObject* parent);
+  ~BackendFailureWatcher();
+
+  /**
+   * @brief notify all the modules that the was a backend failure.
```
lesleyjanenorton, line 19: *there was

src/inspector/inspectorhandler.cpp
```diff
@@ -70,304 +69,269 @@ bool s_pickedItemsSet = false;
...
-                           ->captivePortalDetection()
-                           ->detectCaptivePortal();
-                       return QJsonObject();
-                     }},
-
-    InspectorCommand{
-        "force_captive_portal_detection", "Simulate a captive portal detection",
-        0,
+    InspectorHandler::InspectorCommand{
+        "mockFreeTrial", "Force the UI to show 7 day trial on 1 year plan", 0,
```
oskirby, line 303: I know this is really just whitespace changes, but this is also the only inspector command that uses camelCase. It's stands out.

src/inspector/inspectorhandler.cpp
```diff
@@ -616,44 +573,44 @@ static QList<InspectorCommand> s_commands{
...
-                       auto activity = AndroidVPNActivity::instance();
-                       Q_ASSERT(activity);
-                       auto type = QString(arguments[1]);
-                       auto json = QString(arguments[2]);
-
-                       ServiceAction a = (ServiceAction)type.toInt();
-                       AndroidVPNActivity::sendToService(a, json);
-                       return QJsonObject();
-                     }},
+    InspectorHandler::InspectorCommand{
```
oskirby, line 576: Now that we have the ability to register inspector commands from within modules, it may be cleaner to move this into `AndroidUtils`

addons/tutorial_04_split_tunneling/vpnIsOn.js
```diff
@@ -9,7 +9,9 @@ api.connectSignal(api.addon, 'playingChanged', () => {
     return;
   }
 
-  if (api.controller.state === api.controller.StateOff) {
+  const controller =
```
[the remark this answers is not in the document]
bakulf: correct.

birdsarah (changes requested):
Some initial thoughts ahead of our meeting. I haven't reviewed code changes just thinking about the new shape of the repo.

Going into this my assumption is that this is a code re-org that will allow us to add the app to add additional "modules" that hold entirely new pieces of functionality that are separate from the VPN. In the most extreme you could rebuild this app and be left with a blank canvas that lets you (for example) log in and all that's available is a profile page, settings and addons.

I'll start by saying that I'm supportive off this change.

Initial thoughts on the presented code structure:
* files where it looks like the separation of concerns is off:
  - app permissions
  - backend failure watcher
  - command
  - constants
  - error handler
  - externalophandler
  - mozillavpn
  - url opener
* files where the separation of concerns might be off:
  - ipaddress
  - ipaddress lookup
* files where the separation looks right but needs more work:
  - logout observer
  - profileflow
  - system tray notifications & status icon
* files that seem right even though there's top-level vpn logic:
  - featurelist
* what about the other top-level folders:
  - models
  - commands
  - daemon
  - platforms
  - rfc
  - tasks
  - tutorial
* what about the front-end architecture - how to keep this clean in a module world?

The above is my thoughts on the structure as presented but my bigger picture thought is about the approach here. I'd like to propose / discuss an approach that is more of a change but I think is more conservative. In brief:
* the first PR should move all of the code into a vpn module
* subsequent PRs pull up common code to the shared level. 

My thinking is this:
* Let's be pessimistic and assume a world where new non-vpn module development is slow and there's no big movement for, say, 6 months.
* During this time we want it to be easy for vpn developers to be clear where to put there code - the answer is easy because there's still only one place
* As the requirements of non-vpn modules becomes clear on an as-needed basis the work gets done by that module developers to extract and generalize, crucially, at the time the requirements are now defined. This is important to me because:
  - it appropriately puts the effort for the change into the hands of folks who have the new and well-defined need
  - premature generalization tends to cause problems. This PR has built into it a series of assumptions that we don't yet know to be true. For example, lets say it turns out one of the first requirements is an app that's free and has no login requirements. The first thing we'll need to do is change the structure that's baked into this idea and figure out how to re-split things up into an authenticated and unauthenticated view.

bakulf:
I wrote a document to describe more the whole approach and it's in the meeting invitation for the next conversation.

In the meantime, I want to reply to your last comment:

[...]
>     * During this time we want it to be easy for vpn developers to be clear where to put there code - the answer is easy because there's still only one place

1. First of all, we do not add new components so often. This happens rarely actually.
2.  There is a review process to help the VPN developers to add new components in the right place
3. Soon-ish the unit-tests will be compiled without modules. So, if the component is added in the wrong place, or a header file is included wrongly, we will be red in CI.

>     * As the requirements of non-vpn modules becomes clear on an as-needed basis the work gets done by that module developers to extract and generalize, crucially, at the time the requirements are now defined. This is important to me because:

I agree with you. This is why "shared" components (the untouched ones) are what we don't know yet how they will be used in the future development: authentication, localization, etc. I touched only what I am absolutely sure is unique for the VPN project (controller, server list, capitve-portal, etc).

>       * premature generalization tends to cause problems. This PR has built into it a series of assumptions that we don't yet know to be true. For example, lets say it turns out one of the first requirements is an app that's free and has no login requirements. The first thing we'll need to do is change the structure that's baked into this idea and figure out how to re-split things up into an authenticated and unauthenticated view.

We are 100% sure that the suite will require authentication. Maybe not subscription, but definitely authentication (mainly because P&S products are all behind auth). Plus, even if we did not need the authentication, this component is still cross-project because Relay, Monitor and VPN are all of them behind authentication. Plus, having a shared component, does not mean to use it: example, captive-portal code is compiled everywhere but on mobile it's not used.

PR #5128, opened 2022-12-05
Click outside a textfield to remove focus

MattLichtenstein:
## Description

- Remove focus from TextFields and TextAreas when clicking/tapping elsewhere on the screen
- Auto-focus Custom DNS textfield when it's radio box is selected ( also refactored DNSSettingsTabContent to use loaders instead of hiding UI components that are never used)

## Reference

[VPN-2975: Search fields from Mozilla VPN still active when performing other actions outside the field](https://mozilla-hub.atlassian.net/browse/VPN-2975)
[VPN-3343: Password strength balloon doesn’t show when clicking on the password field](https://mozilla-hub.atlassian.net/browse/VPN-3343)

## Checklist
    

nebula/ui/components/forms/VPNComboBox.qml
```diff
@@ -204,4 +204,15 @@ ComboBox {
             }
         }
     ]
+
+    Connections {
+        target: window
+        function onRemoveFocus(x, y) {
```
lesleyjanenorton, line 210: Comments:
- we could check whether the input has focus and exit out if not before doing the mapping. it seems like this could get called a lot on mobile and I'm wondering how expensive `mapToItem` will be.
- any reason not to consolidate the `onRemoveFocus()` logic into a single fn on main.qml?

lesleyjanenorton (approved):
Cool! Approved with a comment and a question.

nebula/ui/components/forms/VPNComboBox.qml
MattLichtenstein: Good ideas! added

PR #5148, opened 2022-12-07
VPN-3055 - Duplicate all data collection using vpnglean

brizental:
Still missing:
- [x] Tests

### DIFF
--- a/vpnglean/glean_parser_ext/templates/cpp.jinja2
+++ b/vpnglean/glean_parser_ext/templates/cpp.jinja2
@@ -30,7 +30,6 @@ namespace {{ category_name|snake_case }} {
 {% endfor %}
 } // namespace mozilla::glean
 
-namespace {
 {% for category_name, objs in all_objs.items() %}
   {% for obj in objs.values() %}
   {% if obj|attr("_generate_enums") %}
@@ -38,7 +37,6 @@ namespace {
   {%- endif %}
   {% endfor %}
 {% endfor %}
-} // namespace
 
 {% for category_name, objs in all_objs.items() %}
 class __DONOTUSE__{{ category_name|Camelize }} final {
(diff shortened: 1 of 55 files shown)


src/glean/private/event.h
```diff
@@ -62,7 +62,7 @@ class EventMetric final {
 
  public:
   explicit EventMetric(
-      int id, EventMetricExtraParser parser = EventMetricExtraParser());
+      int id, EventMetricExtraParser* parser = new EventMetricExtraParser());
```
brizental, line 65: TODO: Remove this default.

vpnglean/glean_parser_ext/templates/cpp.jinja2
```diff
@@ -30,15 +30,13 @@ namespace {{ category_name|snake_case }} {
...
 
-namespace {
 {% for category_name, objs in all_objs.items() %}
   {% for obj in objs.values() %}
   {% if obj|attr("_generate_enums") %}
 {{ macros.generate_extra_keys_parser(obj, category_name) }}
   {%- endif %}
   {% endfor %}
 {% endfor %}
-} // namespace
```
brizental, line 41: Had to remove the namespace because this is now a property on the GleanMetrics class. Added the `__DONOTUSE__` prefix, just in case people see it in the autocomplete and get confused.

bakulf (changes requested):
As discussed, i think record() should take a ref and not a ptr.

brizental:
Ok, so I was planning to write test for _all_ telemetry in this PR. After working on `[314e8dd](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5148/commits/314e8dd2ec5196d00686d1d9558bde91e377179c)` I realized that is unrealistic so I filed a ticket: https://mozilla-hub.atlassian.net/browse/VPN-3480

At least while working on that one test I made the testing API better, found some bugs in the Glean wrapper implementation and now we have at least ONE test that checks that all the telemetry is working (at least for the one metric that is being tested).

src/glean/private/event.h
brizental: Oh yeah, can't remove this because some events don't have extras.

brizental:
Ok just FYI, there is some bug and that is why all functional tests are failing. For reviewers: I will post here once that is fixed if you'd like to wait. Not sure if it's a minor or major one with the approach.

brizental:
> Ok just FYI, there is some bug and that is why all functional tests are failing. For reviewers: I will post here once that is fixed if you'd like to wait. Not sure if it's a minor or major one with the approach.

Ok that is fixed, something else is up with the Linux builds but it doesn't look major. Will look at it on Monday.

bakulf (approved):

lesleyjanenorton (approved):

brizental:
For now I am blocking this on a decision from #5193 . The CI failiures here would be fixed if we merge that. If we don't merge that I have to change the way I initialize structs here.

PR #5157, opened 2022-12-07
reset codecov coverage uploads

mozrokafor:
## Description

 in efforts to narrow down codecov issues, resetting coverage uploads back to unit tests only

### DIFF
--- a/.github/workflows/test_lottie.yaml
+++ b/.github/workflows/test_lottie.yaml
@@ -62,6 +62,8 @@ jobs:
 
           grcov $(pwd)/tests/unit/CMakeFiles/lottie_tests.dir -s ${{github.workspace}} \
               -t lcov --branch --ignore-not-existing > ${{github.workspace}}/lottie_lcov.info
+        ignore:
+          - "tests/*"
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
@@ -122,6 +124,8 @@ jobs:
 
           grcov $(pwd)/tests/unit/CMakeFiles/lottie_tests.dir -s ${{github.workspace}} \
               -t lcov --branch --ignore-not-existing > ${{github.workspace}}/lottie_lcov.info
+        ignore:
+          - "tests/*"
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/test_lottie.yaml
```diff
@@ -62,6 +62,8 @@ jobs:
 
           grcov $(pwd)/tests/unit/CMakeFiles/lottie_tests.dir -s ${{github.workspace}} \
               -t lcov --branch --ignore-not-existing > ${{github.workspace}}/lottie_lcov.info
+        ignore:
+          - "tests/*"
```
birdsarah, line 66: I'm thinking let's also not do codecov for the lottie tests and just have it for the unit tests so that we only have one workflow uploading to codecov. Once we understand that, we can look at what happens when two different workflows are uploading to codecov. What do you think?

.github/workflows/test_unit.yaml
```diff
@@ -64,19 +64,19 @@ jobs:
           grcov $(pwd)/tests/nativemessaging/CMakeFiles/nativemessaging_tests.dir -s ${{github.workspace}} \
               -t lcov --branch --ignore-not-existing > ${{github.workspace}}/nativemessaging_lcov.info
 
-          grcov $(pwd)/tests/qml/CMakeFiles/qml_tests.dir -s ${{github.workspace}} \
-              -t lcov --branch --ignore-not-existing > ${{github.workspace}}/qml_lcov.info
-
           grcov $(pwd)/tests/unit/CMakeFiles/unit_tests.dir -s ${{github.workspace}} \
               -t lcov --branch --ignore-not-existing > ${{github.workspace}}/unit_lcov.info
+        
+        ignore:
```
birdsarah, line 70: I'm surprised this works. I would expect a --ignore in the grcov calls

birdsarah (changes requested):

bakulf:
Have you tried locally to merge the grcov files and render some charts? Maybe the problem is in how we generate the file.
Plus, grcov is a Mozilla project: have you asked the developers to take a look?

strseb:
Instead of removing coverage i'd rather like to just disable sending and outputting this as an artifact so we can debug this in another workflow

strseb (changes requested):
Instead of removing coverage i'd rather like to just disable sending and outputting this as an artifact so we can debug this in another workflow

bakulf:
I suspect we don't want to land this PR. You can test your changes here and when there is a fix/solution we can land it. No needs to disable test coverage for all the tests in all the repo. Thoughts?

PR #5158, opened 2022-12-08
mhfix - ignore

mozrokafor:


tests/functional/testMultihop.js
```diff
@@ -37,600 +37,482 @@
      console.log(
          'Current city (localized):', currentCity,
          '| Current country code:', currentCountryCode);
-   });
-
-   it('opening the entry and exit server list', async () => {
```
bakulf, line 42: Why removing all these tests?

bakulf (changes requested):

PR #5163, opened 2022-12-08
VPN-3473 - Turn on sortincludes on clang-format

brizental:
This is ready for review, but it is blocked by #5125 as a request from @bakulf :smile: I am happy to take on any merge conflicts. 

The motivation here is: @bakulf always asks me to sort imports and automating that sounded like a good thing. Also, after I actually applied the sorting I thought it makes the import list look beautiful :nail_care: 

Since EVERYONE was going to be flagged for review here I thought it would be fine to have the discussion on whether or not we are all pro this change in here too!


.clang-format
```diff
@@ -25,7 +25,7 @@ NS_IMPL_CYCLE_COLLECTION_TRAVERSE_END_INHERITED|\
 NS_IMPL_CYCLE_COLLECTION_UNLINK_END_INHERITED|\
 NS_QUERYFRAME_TAIL.*$"
 
-SortIncludes: false
+SortIncludes: true
```
brizental, line 28: This is the change. All the rest is just applying the change.

strseb:
On windows we have quite a few where the order of imports is important - the the build errors :)

bakulf:
can you fix the linter? I think we can land this PR.

src/hacl-star/kremlin/minimal/FStar_UInt128.h
```diff
@@ -3,78 +3,81 @@
   Licensed under the Apache 2.0 License.
 */
 
-
 #ifndef __FStar_UInt128_H
 #define __FStar_UInt128_H
 #include <inttypes.h>
 #include <stdbool.h>
+
```
bakulf, line 10: revert these changes. half folder should not be included in clang-format.

bakulf (changes requested):

brizental:
> On windows we have quite a few where the order of imports is important - see the build errors :)

Thanks @strseb , it's possible to disable that only for `src/platform/windows` files. Would that be enough?

brizental:
Oops, closed by mistake.

strseb (approved):
> Thanks @strseb , it's possible to disable that only for src/platform/windows files. Would that be enough?

yes. :)

bakulf:
> > Thanks @strseb , it's possible to disable that only for src/platform/windows files. Would that be enough?
> 
> yes. :)

I think it's better to disable the sorting just on specific files via clang-format comments.

birdsarah (approved):

Gela (approved):

PR #5166, opened 2022-12-08
small pr to include generated macos unit tests in codecov upload

mozrokafor:
## Description

    include macos unit test reports to codecov upload

### DIFF
--- a/.github/workflows/test_unit.yaml
+++ b/.github/workflows/test_unit.yaml
@@ -74,7 +74,7 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           directory: .
-          flags: unit_tests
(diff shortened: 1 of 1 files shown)


oskirby (approved):
Good catch!

strseb (approved):
Screenshot 2022-12-08 at 17 48 48


How q__q

bakulf (approved):

oskirby:
> Screenshot 2022-12-08 at 17 48 48
>
> How q__q
I think the problem is that we have a spurious crash in the nativemessaging test that can cause the Linux test coverage do decrease at random. The problem seems to be that the rust bridge component segfaults. This feels like a separate issue though.

PR #5178, opened 2022-12-08
Experiment: MacOS functional testing

oskirby:
## Description
After a discussion with @mozrokafor it seems like MacOS functional testing might be feasible in CI. Let's give it an extremely quick and dirty try.

## Checklist
    

oskirby:
Well, the test build, and start. So that's cool. Unfortunately, it seems that when run in CI we are unable to connect the test runner to the inspector. So I think that's enough of an experiment for today.

PR #5180, opened 2022-12-09
Android crash fix on feedback

strseb:
Disclaimer: I don't quite understand the reason of this crash. 

Currently, when we want to have the logs, we ask the daemon. This reads the logfile and sends it over the binder. 
The controller passes that string into a QString and gets it to the client. 

The Client will feed that QString in a QTextStream into a QIODevice. Here's where the weirdness begins. 
- If that Device is a file or StdOut (like when we want to send the logfile) -> All good. 
- If that underlying Buffer is a QByteArray or QString, we will crash, if we add any more text after that backend-logfile has been appended.   

Anyway. This was a great reason to stop doing all of that. As the logfile just lives in the temp dir, we can also just have the client read that instead of having the deamon read and then IPC pass it. That works, no crashes.

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNLogger.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNLogger.kt
@@ -60,14 +60,6 @@ class Log {
             e(prefix + tag, message)
         }
 
-        fun getContent(): String? {
-            return try {
-                instance?.file?.readText()
-            } catch (e: Exception) {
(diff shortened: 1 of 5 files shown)


bakulf:
Can we drop the support for QString/QByteArray? And use file/stdout only?

strseb:
@bakulf yes, i think? 
We're reading this into a QString so that we can attach it into a network request. We could write this to a file and then reading it back .... but that sounds expensive :D

oskirby (approved):

bakulf (approved):

PR #5183, opened 2022-12-09
Ignore invalid QDateTime in VPNController - VPN-3421

bakulf:


lesleyjanenorton (approved):

strseb (approved):

strseb:
This makes sense, however this would mean VPN-3421 is not __really__ fixed, right? This would not fix that the Android Service does not provide the right connection timestamp. Can we have a followup bug for me to look at this?

bakulf:
Correct, I'm about to file a follow up to fix how android exposes the connection time.

PR #5192, opened 2022-12-12
Fix silent-server switch on iOS

strseb:
## Description

Currently we send ``ReasonNone`` to the iOS controller for a silent-server switch. This means it will try to activate the tunnel instead of switching. Therefore we never get a ```connected``` event back, timeout in the activation, try all servers until none is left and show the "no servers" error. 

## Reference

[    i.e Jira or Github issue URL](https://mozilla-hub.atlassian.net/jira/software/c/projects/VPN/boards/344?modal=detail&selectedIssue=VPN-3352&assignee=6047c4eeaddbcd006ba87be8)

### DIFF
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -55,6 +55,10 @@ namespace {
 Logger logger("Controller");
 
 ControllerImpl::Reason stateToReason(Controller::State state) {
+  if (state == Controller::StateOn) {
+    // This is a silent-server switch.
+    return ControllerImpl::ReasonSwitching;
+  }
   if (state == Controller::StateSwitching) {
     return ControllerImpl::ReasonSwitching;
   }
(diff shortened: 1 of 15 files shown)


src/controller.cpp
```diff
@@ -54,6 +54,10 @@ namespace {
 Logger logger(LOG_CONTROLLER, "Controller");
 
 ControllerImpl::Reason stateToReason(Controller::State state) {
+  if (state == Controller::StateOn) {
```
bakulf, line 57: This is an hack :/ Because it's not true that if the state is On we are switching.
Can we debug this more and find a better solution? Or, can you tell us more why you think this is the correct fix?

bakulf (changes requested):

src/controller.cpp
strseb: Right, did not think that we need this for deactivation aswell. 
New Approach: 
Have reason be passed explicitly and use `Reason::Switching` for a silent server switch and ```stateToReason()```for all other cases

bakulf (approved):
Nice!

mcleinman (commented):
Is there an easy way to force a silent server switch, to test this?

mcleinman (approved):
I'm unable to replicate the bug, which isn't unexpected as it looks like it's inconsistent. That said, this fix looks good to me so approving. Thanks!

PR #5200, opened 2022-12-13
Server view should switch location when the android back button is clicked - VPN-3474

bakulf:


src/ui/screens/home/ViewServers.qml
```diff
@@ -46,6 +46,16 @@ Item {
...
 
+    Connections {
+        function onGoBack(item) {
+            if (getStack() === item && root.visible) {
+                menu._menuOnBackClicked();
+            }
+        }
+
+        target: VPNNavigator
+    }
```
MattLichtenstein, line 57: Looks good. What are the benefits to checking those two conditions?

MattLichtenstein (approved):
Nice. Also just realized that clicking the android back button on the multi hop server list closes back to the home screen. Not caused by this so will file a separate bug

src/ui/screens/home/ViewServers.qml
bakulf: Well, getStack() returns the stackView, shared between all the ScreenHome views. But we want to have that view visible.

PR #5201, opened 2022-12-13
No ticket - Point glean submodule to release branch

brizental:


bakulf (approved):

brizental:
In order for the Taskcluster jobs to pass the cache needs to be purged. I don't have the permissios to do so.

scripts/linux/script.sh
```diff
@@ -132,7 +132,7 @@ print Y "Update the submodules..."
 git submodule init || die "Failed"
 git submodule update --remote --depth 1 i18n || die "Failed"
 git submodule update --remote --depth 1 3rdparty/wireguard-tools || die "Failed"
-git submodule update --remote --depth 1 3rdparty/glean || die "Failed"
+git submodule update --depth 1 3rdparty/glean || die "Failed"
```
brizental, line 135: This is not changed by mistake. Because the glean submodule is not pointing to "not the main branch" git was not finding the reference to the right commit with the `--remote` option. Which is totally weird, because documentation says that this option will use the branch set on the `.gitmodules` file and I did set it to the `release` branch there. Oh well.

PR #5203, opened 2022-12-13
Disable the split-tunneling and multi-hop features in 2.12 and earler versions - VPN-3491

bakulf:


brizental (approved):

flodolo:
I don't think this type of change really needs me ;-)

MattLichtenstein (approved):

PR #5208, opened 2022-12-13
GH-Pages: Auto Generate Doxygen Page

strseb:
Since we now agreed on doxygen, let's have a config and generate a Github-Subpage under ```/docs```


src/Doxyfile
```diff
@@ -0,0 +1,2723 @@
+# Doxyfile 1.9.5
+
+# This file describes the settings to be used by the documentation system
```
strseb, line 3: This is a super default one. 

Just set ```Generate HTML -> YES``` and our project name

brizental (approved):
Thanks for doing this @basti. Two things 1. Please assign yourself to https://mozilla-hub.atlassian.net/browse/VPN-3479 2. It's fine to use the default renderer for now, but please file a ticket for us to find a nicer one. That will be good especially once this documentation has enough in it that we are actively using it.

PR #5214, opened 2022-12-14
Pin Python dependencies

strseb:
## Description
CI broke again, as we did not pin our depdencies. Let's have a split requirements.in and a locked requirements.txt like in `/taskcluster`


bakulf (approved):

strseb:
Ugh. we have to many diffrent versions of python in our ci flying around. We need to address this first.

Issue #2362, opened 2021-11-27
Use libsecret to store password for the settings for Linux


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-1397)
┆Reporter: Andrea Marchesini

data-sync-user:
➤ Owen Kirby commented:

Oops, I implemented this a while ago but forgot to bring it into the sprint. IIRC it was actually done in January/December

PR #5219, opened 2022-12-14
Implement CryptoSettings for Linux using libsecret

oskirby:
## Description
Linux devices can make use of the `org.freedesktop.secret` D-Bus API to store and load keys and credentials in a way that protects them against readout without the user's login credentials.

## Reference
Github: #2362 ([VPN-1397](https://mozilla-hub.atlassian.net/browse/VPN-1397))

## Checklist
    

src/platforms/linux/linuxcryptosettings.cpp
```diff
@@ -3,15 +3,90 @@
...
+      
+      QString b64key(s_key.toBase64());
+      gboolean ok = secret_password_store_sync(cryptosettings_get_schema(),
+                                               SECRET_COLLECTION_DEFAULT,
+                                               "VPN settings encryption key",
+                                               qPrintable(b64key), nullptr,
+                                               &error, nullptr);
+      if (error != nullptr) {
+        logger.error() << "Key storage failed:" << error->message;
+        g_error_free(error);
```
bakulf, line 75: return false?

src/platforms/linux/linuxcryptosettings.cpp
```diff
@@ -3,15 +3,90 @@
...
+  }
+
+  logger.warning() << "Invalid key";
   return false;
 }
 
 // static
 CryptoSettings::Version CryptoSettings::getSupportedVersion() {
-  return CryptoSettings::NoEncryption;
+  return CryptoSettings::EncryptionChachaPolyV1;
```
bakulf, line 91: is there a way to check the libsecret status?

bakulf (commented):
The documentation of libsecret suggests using the _sync methods on non-ui threads.
Do you think it's doable?

src/platforms/linux/linuxcryptosettings.cpp
oskirby: Not in the official API as far as I can see, but we could check to see if the `org.freedesktop.secrets` name has been taken on the session D-Bus.

oskirby:
> The documentation of libsecret suggests using the _sync methods on non-ui threads. Do you think it's doable?

Unfortunately, I don't think so. The caller of `CryptoSettings::getKey()` is expecting the key to be ready upon return, which requires us to block until the D-Bus response arrives anyways. Which is precisely what the `_sync()` methods do.

src/platforms/linux/linuxcryptosettings.cpp
bakulf: yes, probably it's better. What do you think?

oskirby: I will add a check to ensure that `org.freedesktop.secrets` exists.

bakulf (approved):

PR #5226, opened 2022-12-15
VPN-3327 - Tutorials should not be broken when back button is clicked

bakulf:


birdsarah (commented):
This is new syntax to me. Can you tell me what the motivation is for using `[[nodiscard]]` and then casting to `void` in most places?

bakulf:
Discard is used to force the check of the return value. I ignore the return value only where the code does nothing after the calling of the method, just to avoid an empty if-stmt.
I use it mainly to have help from the compiler to find all the usage of that method. And to avoid future mistakes.

lesleyjanenorton (approved):

PR #5229, opened 2022-12-15
Remove VPNStatusIcon - VPN-3501

bakulf:


birdsarah (approved):

birdsarah:
For posterity. My understanding here is that it's used by our backend code just not the QML code which is why we're just removing this one piece.

PR #5237, opened 2022-12-15
Fix segfault on QML load error

Cimbali:
## Description

- When building from source, dependencies may be missing that cause QML to fail to load. 
   (There may be other situations but this is the scenario I ran into, as well as other users see #3749)
- The queued error handling code is never executed, because `QmlEngineHolder::instance()->window()` is dereferenced before the main loop (`app.qExec()`) is started
- This dereference of non-existing pointer causes a segfault
- Expected behaviour is to log `"Failed to load " << objUrl.toString()` and exit with `-1` as specified in error handling code
- Effectively the error handing code is currently dead code

This PR changes the delayed check for a check just after loading.

Alternately, all `QmlEngineHolder::instance()->window()` could be wrapped in a check of the window existing.
(This is a duplicate of #4121 but I mistakenly made the initial PR from my main branch. Sorry for the noise.)

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2403
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/3749

## Checklist
    
### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -567,16 +567,11 @@ int CommandUI::run(QStringList& tokens) {
 
     // Here is the main QML file.
     const QUrl url(QStringLiteral("qrc:/ui/main.qml"));
-    QObject::connect(
-        engine, &QQmlApplicationEngine::objectCreated, qApp,
-        [url](QObject* obj, const QUrl& objUrl) {
-          if (!obj && url == objUrl) {
-            logger.error() << "Failed to load " << objUrl.toString();
-            QGuiApplication::exit(-1);
-          }
-        },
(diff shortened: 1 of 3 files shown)


bakulf (approved):
Can you add `Q_ASSERT(hasWindows())` in `::window()` ? Thanks!
We are applying big changes in the structure of the repo. If you want, I can rebase this PR when ready.

Cimbali:
Added requested assert. You can rebase it when you feel like it I’m not sure when the structure changes are going through exactly.

bakulf:
I created a new PR, rebasing yours: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5378

PR #5243, opened 2022-12-16
Modules 02: move self-contained classes

bakulf:
Move logging classes, temporarydir, fontloader, taskscheduler into `shared`


oskirby (approved):

oskirby:
Any chance that the `shared` module is all platform independent? (eg: no `UNIT_TEST` or `MVPN_<PLATFORM>` definitions in that code?) It would make the tests a lot faster to build if we could compile it all as a static library and link it into the test applications that way.

bakulf:
No. This cannot be done. For instance the localization requires specific code per platform.

PR #5257, opened 2022-12-19
Modules 05: shared features and settings

bakulf:
VPN-3508 - part 5 - feature list and settings are shared
    
    *Settings changes*
    - shared and per-app settings
    - SettingsHolder application name from appconstants.h
    - cryptosettings as shared component
    - macos/ios crypto settings strings from appsconstants.h
    
    *Feature flags*
    - shared and per-app feature flags
    - apps/vpn/models/feature -> shared/feature
    
    *App template*
    - apps/template - the min set of files to create a new app
    
    *Platform specific code in shared*
    - WindowsUtils for the windowsVersion
    
    *Other changes*
    - Env as shared object
    - VPNSettings.placeholderUserDNS -> VPN.placeholderUserDNS
    - VersionApi.compareVersion -> VersionUtils.compareVersion (to be shareable)


src/shared/settingsholder.cpp
```diff
@@ -41,14 +40,14 @@ SettingsHolder* SettingsHolder::instance() {
   return s_instance;
 }
 
-SettingsHolder::SettingsHolder()
+SettingsHolder::SettingsHolder(const QString& appName)
```
oskirby, line 43: It seems like instead of needing to specify this explicitly, we could default to a sensible value derived from something like `qApp->applicationName()` or the name of the binary. That might cut down on the verbosity of specifying the settings name in the unit tests.

src/shared/platforms/linux/sources.cmake
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(polkit REQUIRED IMPORTED_TARGET polkit-gobject-1)
```
oskirby, line 6: `libsecret-1` is very much a shared dependency in order to realize the cryptosettings on Linux, but `polkit-gobject-1` is only required for the VPN module in order to communicate on the system D-Bus.

birdsarah:
Please make this PR more descriptive.

bakulf:
I'll add more description in the commit message.

src/shared/settingslist.h
```diff
@@ -0,0 +1,33 @@
...
+  SETTING(qint64, toLongLong, getter, __VA_ARGS__)
+
+#define SETTING_STRING(getter, ...) \
+  SETTING(QString, toString, getter, __VA_ARGS__)
+
+#define SETTING_STRINGLIST(getter, ...) \
+  SETTING(QStringList, toStringList, getter, __VA_ARGS__)
+
+// The app must implement its settings list file.
+#include "appsettingslist.h"
```
birdsarah, line 33: I don't see an appsettingslist.h .... I see one under src/apps/vpn

Maybe I'm getting confused about how naming works in this new structure.

src/apps/vpn/appsettingslist.h
```diff
@@ -2,32 +2,8 @@
...
-#define SETTING_INT64(getter, ...) \
-  SETTING(qint64, toLongLong, getter, __VA_ARGS__)
-
-#define SETTING_STRING(getter, ...) \
-  SETTING(QString, toString, getter, __VA_ARGS__)
-
-#define SETTING_STRINGLIST(getter, ...) \
-  SETTING(QStringList, toStringList, getter, __VA_ARGS__)
+// NOTE! This file is not meant to be used directly. Please use settingslist.h
+// instead.
```
birdsarah, line 6: Do you mean, don't #include appsettingslist.h but include settingslist.h instead. 

What if you want to add another setting?

I'm probably being over specific but is what you mean:
* you can update this file if needed
* but include settingslist.h

src/apps/vpn/appfeaturelist.h
```diff
@@ -2,21 +2,8 @@
...
-#if defined(_MSVC_TRADITIONAL) && _MSVC_TRADITIONAL
-#  error No supported
-#endif
-
-#define FEATURE_SIMPLE(id, name, releaseVersion, flippableOn, flippableOff, \
-                       dependencies, callback)                              \
-  FEATURE(id, name, false, L18nStrings::Empty, L18nStrings::Empty,          \
-          L18nStrings::Empty, "", "", "", releaseVersion, flippableOn,      \
-          flippableOff, dependencies, callback)
+// NOTE! This file is not meant to be used directly. Please use featurelist.h
```
birdsarah, line 5: this feels confusing...like the other similar pattern...but i'm happy to wait and see i'm probably missing context.

birdsarah (approved):
I have questions and would love answer but am approving as you might be runnign out of warm bodies.

src/CMakeLists.txt
```diff
@@ -41,6 +35,12 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten")
     set(MVPN_PLATFORM_NAME "wasm")
 endif()
 
+include(shared/sources.cmake)
+include(apps/vpn/cmake/sources.cmake)
+include(apps/vpn/cmake/sentry.cmake)
+
+target_link_libraries(mozillavpn PRIVATE shared-sources mozillavpn-sources glean lottie nebula translations)
```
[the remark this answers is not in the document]
bakulf: the order matter because I need to have the "platform" var set before including the shared/sources.cmake for: `include(${CMAKE_CURRENT_SOURCE_DIR}/shared/platforms/${MVPN_PLATFORM_NAME}/sources.cmake)`

src/shared/settingslist.h
bakulf: `appsettingslist.h` is the app-specific setting file. Each 'app' needs to provide it. You can find one in `apps/vpn` and one in `apps/template`.

src/apps/vpn/appsettingslist.h
bakulf: Correct. I'll update the comment. We have 2 level of settings:
- shared (settingslist.h)
- per-app (appsettingslist.h)
- per-module (not implemented yet)

The "include" diagram is the following:

1. settingsholder includes both settingslist.h
2. settingslist.h defines the macros, the shared-settings and includes the appsettingslist.h
3. appsettingslist.h defines the per-app settings.

src/apps/vpn/appfeaturelist.h
bakulf: Similar pattern. I update the comment here as well.

tests/dummyvpn/CMakeLists.txt
```diff
@@ -34,9 +34,17 @@ target_link_libraries(dummyvpn PRIVATE
 target_sources(dummyvpn PRIVATE
     ${CMAKE_SOURCE_DIR}/src/apps/vpn/platforms/dummy/dummycontroller.cpp
     ${CMAKE_SOURCE_DIR}/src/apps/vpn/platforms/dummy/dummycontroller.h
-    ${CMAKE_SOURCE_DIR}/src/apps/vpn/platforms/dummy/dummycryptosettings.cpp
 )
 
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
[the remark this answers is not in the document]
bakulf: before this PR, the dummyvpn client does not include linux-specific code. But now, the "shared-component" is independent from the targets. Because of this, for linux, we need dbus.

PR #5260, opened 2022-12-19
Add taskcluster jobs to build Ubuntu packages.

oskirby:
## Description
This PR adds Taskcluster jobs to produce packages for our supported Ubuntu releases (at present: `bionic`, `focal`, `jammy` and `kinetic`). This starts with an initial job to produce a source bundle, which is then shared by a separate build worker for each target release. The magic here is in producing a generic Docker image that matches an official Ubuntu build worker as closely as possible, but permits us to swap out the base container image.

Because we have split the build tasks up, the `build-linux/opt` task no longer really exists, and we will need to update Github's branch protection rules in order to merge this.

I haven't gone so far as to force offline builds like the official build workers do, so it is possible that a bug the golang and rust module vendoring could creep into the PPA.

So far this only produces build artifacts. We still need release tasks to publish them to archive.mozilla.org and/or push the source builds to a PPA. And there is some room for future improvement by targeting Debian and some of the RPM distributions too.

## Reference
JIRA issue [VPN-1706](https://mozilla-hub.atlassian.net/browse/VPN-1706)

## Checklist
    
### DIFF
--- a/taskcluster/ci/build/linux.yml
+++ b/taskcluster/ci/build/linux.yml
@@ -31,7 +31,7 @@ linux/source:
             ./scripts/linux/script.sh --source &&
             tar -C .tmp -zvcf /builds/worker/artifacts/mozillavpn-sources.tar.gz .
 
-linxu/bionic:
+linux/bionic:
     description: "Linux Build (Ubuntu/Bionic)"
     treeherder:
         platform: linux/bionic
--- a/taskcluster/ci/build/linux.yml
+++ b/taskcluster/ci/build/linux.yml
@@ -2,23 +2,95 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 ---
-linux/opt:
-    description: "Linux Build"
+task-defaults:
     treeherder:
-        symbol: B
-        kind: build
-        tier: 1
-        platform: linux/opt
+            symbol: B
+            kind: build
+            tier: 1
     worker-type: b-linux
+    requires-level: 1
+    worker:
+          max-run-time: 3600
+          chain-of-trust: true
+    run:
+          using: run-task
+          use-caches: true
+
+linux/source:
+    description: "Linux Source Build"
+    treeherder:
+        platform: linux/source
     worker:
         docker-image: {in-tree: build}
-        max-run-time: 3600
-        chain-of-trust: true
     release-artifacts:
         # release-artifacts expects file to be in /builds/worker/artifacts/
-        - target.tar.gz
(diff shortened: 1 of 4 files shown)


taskcluster/ci/build/linux.yml
```diff
@@ -2,23 +2,95 @@
...
     run:
-        using: run-task
-        use-caches: true
         cwd: '{checkout}'
-        command: ./taskcluster/scripts/build/linux.sh
+        command: >-
+            ./scripts/linux/script.sh --source &&
+            tar -C .tmp -zvcf /builds/worker/artifacts/mozillavpn-sources.tar.gz .
+
+linxu/bionic:
```
strseb, line 34: Nit: typo :)

taskcluster/ci/build/linux.yml
```diff
@@ -2,23 +2,95 @@
...
+        platform: linux/jammy
+    fetches:
+        build:
+            - artifact: mozillavpn-sources.tar.gz
+    dependencies:
+        build: build-linux/source
+    worker:
+        docker-image: {in-tree: linux-build-jammy}
+    release-artifacts:
+        - mozillavpn-jammy.tar.gz
```
strseb, line 78: Not sure that we need `release-artifacts` - If I understand correctly, the package files we're generating are for us, right? Like the github-ci ones. `release-artifacts` is only useful if we need to feed it downstream for signing. 

Maybe consider: 
```
worker:
        artifacts:
            - type: directory
              name: public/build
              path: /builds/worker/artifacts
```
And drop all the files there. This will make each file index-able, so QA (and anyone really) could get a `main.latest/ubuntu_jammy.deb` kind link :)

taskcluster/ci/build/linux.yml
```diff
@@ -2,23 +2,95 @@
...
     treeherder:
-        symbol: B
-        kind: build
-        tier: 1
-        platform: linux/opt
+            symbol: B
+            kind: build
+            tier: 1
     worker-type: b-linux
+    requires-level: 1
```
strseb, line 11: That is not needed.  `requires-level: 1` will always evaluate to true.

strseb (approved):
N-I-C-E! Thank you. 
Left some small nits :)

taskcluster/ci/build/linux.yml
oskirby: Well, at some point we are going to need a signing and release process for these build artifacts, but that isn't something I can work on without help from releng. So for now this only aims to produce unsigned builds for our own consumption.

oskirby: This is an extremely cool suggestion though, I like it.

PR #5270, opened 2022-12-20
Modules 06: renaming MVPN_* to MZ_*

bakulf:
This covers the shared folder. I'll continue the renaming when moving other components.


src/CMakeLists.txt
```diff
@@ -21,29 +21,29 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten"
...
-target_compile_definitions(mozillavpn PRIVATE "$<$<CONFIG:Debug>:MVPN_DEBUG>")
-target_compile_definitions(mozillavpn PRIVATE "MVPN_$<UPPER_CASE:${MVPN_PLATFORM_NAME}>")
-include(apps/vpn/cmake/${MVPN_PLATFORM_NAME}.cmake)
+include(shared/sources.cmake)
+include(apps/vpn/cmake/sources.cmake)
+include(apps/vpn/cmake/sentry.cmake)
+
+target_link_libraries(mozillavpn PRIVATE shared-sources mozillavpn-sources glean lottie nebula translations)
+
+target_compile_definitions(mozillavpn PRIVATE "$<$<CONFIG:Debug>:MZ_DEBUG>")
```
birdsarah, line 44: what is this mozillavpn? (apologies if this should be obvious i'm behind on reviewing this series of prs)

birdsarah (changes requested):
I haven't finished reviewing. I have one question and more broadly the PR in it's current state seems to be doing a lot more than renaming MVPN_* to MZ_* in the shared folder.

birdsarah:
> I haven't finished reviewing. I have one question and more broadly the PR in it's current state seems to be doing a lot more than renaming MVPN_* to MZ_* in the shared folder.

I think this is because it contains the 05 commit too.

bakulf:
Yes! my fault. I picked the wrong 'base' for this PR.

src/CMakeLists.txt
bakulf: This is the "mozillaVPN" target, the client.

birdsarah: ok. i think i'm going to need an overview of how this will work...will there be lots of target_compile_definitions at this level? ... like one for the tests etc.

birdsarah (approved):

PR #5273, opened 2022-12-20
VPN-2452 - Subscription Expiring Message

jps-moz:
## Description

-Implemented a message addon to notify users when their subscription is expiring.
-They also get an OS notification (until background notifications for mobile is implemented, this will only be for Desktop).
-The ‘Subscription expiring’ message doesn’t show anymore once the subscription end date > 7 days again.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2542

## Checklist
    
### DIFF
--- a/addons/message_subscription_expiring/isExpiring.js
+++ b/addons/message_subscription_expiring/isExpiring.js
@@ -1,0 +1,12 @@
+(function(api, condition) {
+  // Show message only if within 1 week of expiring.
+  let weekBeforeExpireMSecs = api.subscriptionData.expiresOn - 1000 * 60 * 60 * 24;
+
+  if (Date.now() < api.subscriptionData.expiresOn &&
+      Date.now() >= weekBeforeExpireMSecs) {
+    api.addon.date = weekBeforeExpireMSecs;
+    condition.enable();
+  } else {
+    condition.disable();
+  }
+})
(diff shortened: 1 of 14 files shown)


flodolo (approved):

addons/message_subscription_expiring/isExpiring.js
```diff
@@ -0,0 +1,11 @@
+(function(api, condition) {
+  //Show message only if within 1 week of expiring.
+  let weekBeforeExpireSecs = (api.subscriptionData.expiresOn / 1000) - 7*24*60*60;
```
bakulf, line 3: technically `expiresOn` can be 0. So, `weekBeforeExpireSecs` will be -604800, which is less than `Date.now()`.

bakulf (approved):

bakulf (changes requested):
Add a functional test for this addon to simulate its visibility. See the addon folder in tests/tunctional

addons/message_subscription_expiring/isExpiring.js
jps-moz: The other side of that && expression should handle that case :+1:

jps-moz:
> Add a functional test for this addon to simulate its visibility. See the addon folder in tests/tunctional

After exploring this folder and the surrounding test folders, I was unable to find a clear way to trigger the new message directly (for example, by editing the expiresOn property). It seems like these tests are for confirming new functionality or types of addons altogether, which this doesn't fit IMO. I would essentially be duplicating the code, but not testing the one seen by the user. Is there something I am misunderstanding here?

PR #5278, opened 2022-12-20
Android/TileService: Check Login state

strseb:
Let the Service communicate whether `reconnect()` can be called and let the tile service reflect that. In case no config is present we will switch to state `unknown` which will disable the button. 
Also let the client tell the service when a logout happened to clear out the old wireguard config.

https://mozilla-hub.atlassian.net/browse/VPN-3420


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -164,6 +164,16 @@ class VPNService : android.net.VpnService() {
         get() {
             return mConnectionTime
         }
+    var canActivate: Boolean = false
+        get() {
+            if (mConfig != null) {
+                // We have a config in the runtime, can do
+                return true
+            }
+            // Return what we have in the
```
brizental, line 173: This comment and the one above look incomplete.

src/apps/vpn/platforms/android/androidvpnactivity.cpp
```diff
@@ -149,3 +153,7 @@ void AndroidVPNActivity::startAtBootChanged() {
   QJsonDocument doc(args);
   sendToService(ServiceAction::ACTION_SET_START_ON_BOOT, doc.toJson());
 }
+
+void AndroidVPNActivity::onLogout() {
+  sendToService(ServiceAction::ACTION_CLEAR_STORAGE);
+}
```
brizental, line 159: nit: Missing newline.

src/apps/vpn/platforms/android/androidvpnactivity.h
```diff
@@ -48,6 +48,11 @@ enum ServiceAction {
   ACTION_GLEAN_SET_SOURCE_TAGS = 14,
   // Set startOnBoot pref
   ACTION_SET_START_ON_BOOT = 15,
+  // Reactivate the last connection (unusued from the client)
+  ACTION_REACTIVATE = 16,
```
brizental, line 52: Why do we need this if it is unused?

brizental (approved):
Do we have capabilities for unit testing in Android? I think it would be nice to have a test for this.

src/apps/vpn/platforms/android/androidvpnactivity.h
strseb: Completeness. it's part of our ipc protocol but not used in c++ (bot in kotlin-ipc).
I don't want to look in which file the magic integer is defined :D

strseb:
>Do we have capabilities for unit testing in Android? I think it would be nice to have a test for this. 
We do not! :D - We'd need a good restructuring of the android build process to be able to run android-tests (which is what we'd need here). 
Opened: https://mozilla-hub.atlassian.net/browse/VPN-3547
to make it happen.

PR #5285, opened 2022-12-21
VPN-3508 - part 9 - Introduce VPN GleanDeprecated

bakulf:
Deprecated because this will be removed when we get rid of glean-js. This is needed to remove the dependency between components and MozillaVPN.


src/shared/glean/gleandeprecated.h
```diff
@@ -0,0 +1,28 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef GLEANDEPRECATED_H
+#define GLEANDEPRECATED_H
+
+#include <QObject>
+#include <QVariant>
+
+class GleanDeprecated final : public QObject {
```
brizental, line 11: Please add a comment to this class on why it is deprecated and that it should not be used in new apps.

brizental (approved):

PR #5287, opened 2022-12-21
Modules 07: no VPN references in src/shared

bakulf:
I missed a few things.

### DIFF
--- a/src/apps/template/appconstants.h
+++ b/src/apps/template/appconstants.h
@@ -22,6 +22,33 @@ constexpr const char* MACOS_FALLBACK_APP_ID = "org.mozilla.macos.AppFooBar";
 constexpr const char* IOS_FALLBACK_APP_ID = "org.mozilla.ios.AppFooBar";
 #endif
 
+#if defined(MZ_WINDOWS)
+// Credential key for windows
+constexpr const wchar_t* WINDOWS_CRED_KEY = L"Foo bar";
+#endif
+
+#if defined(MZ_ANDROID)
+constexpr const char* ANDROID_LOG_NAME = "foobar";
+#endif
+
+#if defined(MZ_LINUX)
+constexpr const char* LINUX_CRYPTO_SETTINGS_KEY =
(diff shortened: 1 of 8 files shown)


src/apps/template/appconstants.h
```diff
@@ -22,6 +22,33 @@ constexpr const char* MACOS_FALLBACK_APP_ID = "org.mozilla.macos.AppFooBar";
 constexpr const char* IOS_FALLBACK_APP_ID = "org.mozilla.ios.AppFooBar";
 #endif
 
+#if defined(MZ_WINDOWS)
```
oskirby, line 25: I feel like these should not be in the template unless they are things we expect *all* applications to provide, and I don't think it's appropriate to assume that all applications are going to support addons and cryptosettings.

src/apps/template/appconstants.h
```diff
@@ -22,6 +22,33 @@ constexpr const char* MACOS_FALLBACK_APP_ID = "org.mozilla.macos.AppFooBar";
...
+// Credential key for windows
+constexpr const wchar_t* WINDOWS_CRED_KEY = L"Foo bar";
+#endif
+
+#if defined(MZ_ANDROID)
+constexpr const char* ANDROID_LOG_NAME = "mozillavpn";
```
oskirby, line 31: `constexpr const char* ANDROID_LOG_NAME = "foobar";` ?

src/apps/template/appconstants.h
bakulf: Why not? Why we should have not encrypted settings? Plus add-ons are a key component for content delivering. The suite app needs both components.

oskirby: I'm just thinking ahead for stuff like unit tests where we may not want to introduce platform-specific variability in the test runs.

bakulf: Why not? We run unit tests on all the platforms. In this way we increase the test coverage.

oskirby (approved):

PR #5288, opened 2022-12-21
Modules 08: language signature theme

bakulf:


src/shared/platforms/android/androidcommons.h
```diff
@@ -0,0 +1,23 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ANDROIDCOMMONS_H
+#define ANDROIDCOMMONS_H
+
+#include <QObject>
+
+// TODO: this needs to change!
+constexpr auto ANDROID_UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
```
strseb, line 11: This should be kept in androidutils.cpp

src/shared/platforms/android/androidcommons.h
```diff
@@ -0,0 +1,23 @@
...
+
+class AndroidCommons final : public QObject {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(AndroidCommons)
+
+ public:
+  static bool verifySignature(const QByteArray& publicKey,
```
strseb, line 18: If the function moves CPP classes it should also move classes in the JVM, otherwise we this can get messy  soon - I'd prefer we it to 1 JVM class<-> 1 CPP class bindings.  
So maybe this a new class a shared package.`org.mozilla.firefox.{something}` that we can use in both flavors.

strseb (approved):
r+ with comments :)

src/shared/signature.cmake
```diff
@@ -0,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
oskirby, line 1: This file was removed during the Apple universal build PR as it does not correctly handle multi-arch and cross compiled builds.

src/shared/platforms/linux/sources.cmake
```diff
@@ -9,3 +9,5 @@ target_include_directories(shared-sources INTERFACE ${libsecret_INCLUDE_DIRS})
 target_sources(shared-sources INTERFACE
      ${CMAKE_CURRENT_SOURCE_DIR}/shared/platforms/linux/linuxcryptosettings.cpp
 )
+
+include(${CMAKE_CURRENT_SOURCE_DIR}/shared/signature.cmake)
```
oskirby, line 13: `signature.cmake` was replaced by a generalized Rust helper, and you should now do something like the following:

```
include(${CMAKE_SOURCE_DIR}/scripts/cmake/rustlang.cmake)
add_rust_library(signature
    PACKAGE_DIR ${CMAKE_SOURCE_DIR}/signature
    BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}
    CRATE_NAME signature
)
target_compile_definitions(mozillavpn PRIVATE MZ_SIGNATURE)
target_link_libraries(mozillavpn PRIVATE signature)
```

oskirby (changes requested):

strseb (approved):

oskirby (approved):

PR #5292, opened 2022-12-21
VPN-3056 - Use vpnglean with the Glean Android SDK

brizental:
Still to-do:

- [x] Test (manually)


android/daemon/build.gradle
```diff
@@ -29,38 +30,41 @@ buildscript {
...
     defaultConfig {
         minSdkVersion Config.minSdkVersion
         targetSdkVersion Config.targetSdkVersion
-        
+
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
-        buildConfigField  "String", "VERSIONCODE" ,  '"' +System.getenv("VERSIONCODE") + '"'
+
+        buildConfigField  "String", "VERSIONCODE" ,  '"' + System.getenv("VERSIONCODE") + '"'
         buildConfigField  "String", "SHORTVERSION" ,  '"' + System.getenv("SHORTVERSION") +  '"'
```
brizental, line 50: I think the reason why build config is not generated by glean_parser in here, is because the daemon is a library. Since the build info is required for initializing Glean and there is [a strong recomendation from Glean](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/index.html#glean-integration-checklist) not to initialize Glean on libraries, maybe that step is skipped? cc @badboy for confirmation if this is a bug or intended behaviour.

README.md
```diff
@@ -420,26 +420,28 @@ it from the [official website](https://golang.org/dl/).
 
 2. Follow the [Getting started](https://doc.qt.io/qt-6/android-getting-started.html) page.
 
-3. Set the `QT_HOST_PATH` environment variable to point to the location of the `androiddeployqt` tool.
+3. Set the `QT_HOST_PATH` environment variable to point to the location of the `androiddeployqt` tool  -- minus the `/bin` suffix i.e. if `$(which androiddeployqt)` is `$HOME/Qt/6.2.4/gcc_64/bin/androiddeployqt`, `QT_HOST_PATH` is `$HOME/Qt/6.2.4/gcc_64/`.
 
-4. Set the `ANDROID_SDK_ROOT` and `ANDROID_NDK_ROOT` (required version: 21.0.6113669) environment variables,
-to point to the Android SDK and NDK installation directories.
+4. Set the `ANDROID_SDK_ROOT` and `ANDROID_NDK_ROOT` environment variables,
+to point to the Android SDK and NDK intallation directories. Required NDK versions: 23.1.7779620 and 21.0.6113669.
```
brizental, line 426: Sanity check from @strseb that both are really required. I was only able to build with both... But maybe it's something in my environment?

strseb: Uh! Yes, doc-rot. It should only require 23.1.7779620.

brizental: Ok, so the thing is. When I installed `23.1.7779620` I removed `21.0.6113669` and I that asked me for `21.0.6113669`.

brizental: Hm, if CI only has the 23 version I'll leave jsut that. Requiring two versions is weird.

strseb: Maybe QT has something that depends on 21.x? :/

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
```diff
@@ -32,10 +32,7 @@ class VPNServiceBinder(service: VPNService) : Binder() {
         const val setNotificationText = 8
         const val setStrings = 9
         const val recordEvent = 10
-        const val sendGleanPings = 11
-        const val gleanUploadEnabledChanged = 12
         const val getStatus = 13
-        const val gleanSetSourceTags = 14
```
brizental, line 38: Do we want to keep these in order?

strseb (approved):
r+ just 2 notes :) 


r+

android/daemon/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
     <application>
         <service android:name=".VPNService"
             android:permission="android.permission.BIND_VPN_SERVICE"
+            android:process=":QtOnlyProcess"
```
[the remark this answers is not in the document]
brizental: > This should be the daemon not qt :)

Do you mean like `android:process=":DaemonOnlyProcess"`?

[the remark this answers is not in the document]
strseb: Yes or: android:process=":Daemon"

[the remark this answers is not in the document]
brizental: Nice, ok. I didn't realize that was allowed :flushed:

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -19,29 +19,29 @@ import com.wireguard.config.Peer
 import com.wireguard.crypto.Key
 import org.json.JSONObject
 import java.util.*
+import org.mozilla.firefox.vpn.daemon.GleanMetrics.Sample
```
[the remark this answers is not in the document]
brizental: Yeah, that is a whole different issue. After this whole Glean migration is stable we will audit our data collection choices and the sample category will be gone first, for sure :D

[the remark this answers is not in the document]
brizental: > this is also going to start using a new app id and so would be the best opportunity to change it.

Wait, why do you say this will start using a new app id? I did not realize that would be the case.

PR #5294, opened 2022-12-21
caching whole geckodriver dir

mozrokafor:
## Description

    caching full geckodriver dir, as token is not authenticating the request

### DIFF
--- a/.github/workflows/wasm.yaml
+++ b/.github/workflows/wasm.yaml
@@ -45,6 +45,13 @@ jobs:
       - name: Setup emsdk
         uses: mymindstorm/setup-emsdk@v7
 
+      - name: Cache geckodriver
+        id: cache-lel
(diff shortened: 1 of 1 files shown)


.github/workflows/wasm.yaml
```diff
@@ -45,6 +45,13 @@ jobs:
...
         uses: mymindstorm/setup-emsdk@v7
 
+      - name: Cache geckodriver
+        id: cache-lel
+        uses: actions/cache@v3
+        with:
+          path: /opt/hostedtoolcache/geckodriver/
+          key: wasm_geckodriver
+
       - name: Set up gecko driver
```
oskirby, line 55: If the cache was a hit, do we need to skip the setup step to avoid overwriting the cache and hitting the API key again?

mozrokafor:
fixed

PR #5295, opened 2022-12-21
CI: remove geckodriver.

strseb:
## Description

In https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5264/files
geckodriver was actually removed from the running task. It's green. 
We certainly dont need it to build the client.


mozrokafor (approved):
LGTM, good catch

PR #5299, opened 2022-12-21
TC: Windows add Scopes for level-3 Builds

strseb:
## Description
In order for the sentry-stuff to actually work the build needs access to the tokens. Let's enable it for level-3 tasks :)


mozrokafor (approved):
LGTM

PR #5301, opened 2022-12-21
Fix VPN-3424: Unable to assign [undefined] to QString

lesleyjanenorton:
## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNLinkButton.qml
+++ b/nebula/ui/components/VPNLinkButton.qml
@@ -11,12 +11,13 @@ import Mozilla.VPN 1.0
 VPNButtonBase {
     id: root
 
-    property var labelText: ""
+
+    property string labelText
     property variant fontName: VPNTheme.theme.fontInterFamily
     property var baseColor: VPNTheme.theme.linkButton
     property var linkColor: VPNTheme.theme.blueButton
     property var fontSize: VPNTheme.theme.fontSize
-    property var textAlignment: Text.AlignHCenter
+    property real textAlignment: Text.AlignHCenter
     property var buttonPadding: VPNTheme.theme.hSpacing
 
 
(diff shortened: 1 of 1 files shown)


MattLichtenstein (approved):

bakulf (approved):

nebula/ui/components/VPNLinkButton.qml
```diff
@@ -11,7 +11,7 @@ import Mozilla.VPN 1.0
 VPNButtonBase {
     id: root
 
-    property var labelText
+    property var labelText: ""
```
bakulf, line 14: Can we use `string` as type instead?

bakulf (approved):

nebula/ui/components/VPNLinkButton.qml
MattLichtenstein: this is actually a good point, I believe using the correct data type not only provides a slight performance boost, but also allows you to not need to specify an `""` to denote that the string is defined, but empty

bakulf: Do you think it makes sense to audit the use of `var` and replace them with proper types? (not in this PR, not in this jira issue)

MattLichtenstein: It doesn't seem to be documented anywhere just how much of a performance degradation this causes, but it's enough for [the syntax checker look for them (id: M311)](https://doc.qt.io/qtcreator/creator-checking-code-syntax.html). That and we won't run into the QML warning that this PR sets out to fix as much. I tend to only use `var` when none of the other basic types would suffice

lesleyjanenorton: > allows you to not need to specify an `""` to denote that the string is defined, but empty

Nice. updated. +1 to audit.

PR #5307, opened 2022-12-22
VPN-3508: Modules 15: Unit test as app

bakulf:


src/apps/vpn/appsettingslist.h
```diff
@@ -157,26 +127,6 @@ SETTING_STRING(entryServerCountryCodeDeprecated,        // getter
                true                                     // remove when reset
 )
 
-SETTING_STRINGLIST(featuresFlippedOff,        // getter
```
strseb, line 160: Why are those removed?

src/shared/eventlistener.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "eventlistener.h"
 
+#include <QFileInfo>
```
birdsarah, line 7: Why is this needed?

src/apps/unit_tests/testfeature.cpp
```diff
@@ -0,0 +1,79 @@
...
+      "",                              // link URL
+      "1.0",                           // released
+      []() -> bool { return false; },  // Can be flipped on
+      []() -> bool { return false; },  // Can be flipped off
+      QStringList(),                   // feature dependencies
+      []() -> bool { return false; });
+  QVERIFY(!!Feature::get("testFeatureC"));
+  QVERIFY(!Feature::get("testFeatureC")->isSupported());
+}
+
```
birdsarah, line 78: I don't understand this test. It's not clear to me what it's testing and why.

birdsarah (approved):
Approve with comments.

Notes to self: this PR also includes:
- moving the licensemodel to shared
- moving some settings around
- adds a new test for urlQueryReplacement

Request: Can we add in the README some guidance about when unit tests should live under tests/ unit and when they should live under src/apps/unit_tests? My understanding from the PR:
* things in src/shared are tested in src/apps/unit_tests
* things in src/apps/vpn are tested in tests/unit

Note: I don't fully understand Feature and FeatureModel and the separation between them and the test in here related to it. I'm ok with not holding up this PR based on my lack of understanding. We can follow-up later.

birdsarah:
It looks to me that the failing windows unit test is a real failure and you need some extra cmake work, or something, so that you can successfully build the unit tests on windows. I'm going to try and get my windows box running.

src/apps/unit_tests/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+    target_link_libraries(app_unit_tests PRIVATE PkgConfig::libsecret)
+endif()
+
+target_compile_definitions(app_unit_tests PRIVATE MZ_DEBUG)
+target_compile_definitions(app_unit_tests PRIVATE MZ_DUMMY)
+
+# Unit test source files
+target_sources(app_unit_tests PRIVATE
+    main.cpp
+    helper.h
```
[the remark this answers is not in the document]
bakulf: They should be added here. Thanks!

src/apps/unit_tests/testfeature.cpp
bakulf: testFeatureC makes sense only for the model. Removing from this test.

PR #5311, opened 2022-12-22
ci: re-enable Linux signing tasks

ahal:
## Description

Re-enable Linux signing tasks without using the production key.

### DIFF
--- a/taskcluster/ci/build/source.yml
+++ b/taskcluster/ci/build/source.yml
@@ -22,6 +22,8 @@ source/vpn:
     add-index-routes:
         name: source
         type: build
+    release-artifacts:
+        - mozillavpn-sources.tar.gz
     run:
         using: run-task
         use-caches: true
--- a/taskcluster/ci/config.yml
+++ b/taskcluster/ci/config.yml
@@ -66,7 +66,7 @@ workers:
             provisioner: scriptworker-k8s
             implementation: scriptworker-signing
             os: scriptworker
-            worker-type: mozillavpn-t-signing
+            worker-type: mozillavpn-t-signing-dev
         signing:
             provisioner: scriptworker-k8s
             implementation: scriptworker-signing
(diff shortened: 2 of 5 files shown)


strseb (approved):

oskirby:
I can confirm that this produces signed source bundles, but I am struggling to verify that signature as I have no idea which public key to verify it against.

ahal:
@oskirby the public key should be attached to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1776071)

birdsarah:

strseb:
@ahal - Sorry for the ping, but what would be needed to get this green? Can we help? :)

ahal:
Actually, we might want to close this and remove any references to the Linux signing tasks :)

The reason is that these signing tasks were a launchpad thing and I don't think they'll be necessary once we switch over to Google Artifact Registry. We made the decision to not bother working on launchpad automation as a result.

strseb:
Tbf @oskirby and I just had a conversation yesterday, that it might be nice beetmoving a signed source bundle onto archive. 
We have a few bugs of maintainers that have issues building - as the source bundle github Publishes does not contain any third party code so people have to fetch that all themselves to build packages. 
So if we had a central place to get the same complete source bundle have been using to build the released client, that could be neat. If that also would be signed by Mozilla, that would be extra neat. 
If that is too much hassle to get working - no worries :)

ahal:
Ah ok, we could re-visit this then! I think the PR was mostly done (unless it's bitrotted again) and just needed testing.. Though I think the subkey that was being used to sign them might be expired now (I had told secops not to bother rotating it).. so I guess that might be a bit of a pain.

Though I guess I'm wondering what benefit signing would provide over just providing a hash or similar, as my understanding is that the signing was just a launchpad thing.

PR #5315, opened 2022-12-22
First pass QML custom property type cleanup 

lesleyjanenorton:
## Description

This fixes a few obviously incorrect custom property types in our QML files following this comment thread (https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5301#discussion_r1054825472). Full audit still needed.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNAlert.qml
+++ b/nebula/ui/components/VPNAlert.qml
@@ -13,26 +13,26 @@ Rectangle {
     id: alertBox
 
     // Defines what type of alert this is
-    property var alertType: stateNames.info
+    property string alertType: stateNames.info
     // Contains all accepted values for alertType
     readonly property var alertTypes: stateNames
     // Determines if the toast should fit into the layout or put itself on the top
-    property var isLayout: false
+    property bool isLayout: false
     // Callback when the action was pressed,
     // (does nothing if  no action text is defined)
     property var onActionPressed:function(){}
     // Callback when the close button was pressed
     property var onClosePressed:function(){}
     // Text of the toast
-    property var alertText: "Some Alert"
+    property string alertText: "Some Alert"
     // Text of the Action (in bold)
-    property var alertActionText: ""
+    property string alertActionText: ""
     // Auto Hide the Alert after X secounds (0 == infinte)
-    property var duration: 0
+    property real duration: 0
     // Delete itself after closing
-    property var destructive: false
+    property bool destructive: false
     // Fixes the Y on show() if the alert does not use Layout
-    property var setY: 0
+    property real setY: 0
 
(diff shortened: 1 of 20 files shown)


bakulf (approved):

nebula/ui/components/VPNAlert.qml
```diff
@@ -13,26 +13,26 @@ Rectangle {
     id: alertBox
 
     // Defines what type of alert this is
-    property var alertType: stateNames.info
+    property string alertType: stateNames.info
     // Contains all accepted values for alertType
     readonly property var alertTypes: stateNames
     // Determines if the toast should fit into the layout or put itself on the top
-    property var isLayout: false
+    property bool isLayout: false
```
MattLichtenstein, line 20: the default value of a QML bool is false, so technically this is redundant BUT I don't think thats totally intuitive and obvious so I think we should set this as a precedent that QML bools should always be defined when they are declared

nebula/ui/components/VPNAlert.qml
```diff
@@ -13,26 +13,26 @@ Rectangle {
...
     // Callback when the action was pressed,
     // (does nothing if  no action text is defined)
     property var onActionPressed:function(){}
     // Callback when the close button was pressed
     property var onClosePressed:function(){}
     // Text of the toast
-    property var alertText: "Some Alert"
+    property string alertText: "Some Alert"
```
MattLichtenstein, line 27: For properties like this, I think we should remove the default assigned value ("Some Alert") and utilize the `required` attribute so we don't ever accidentally end up with an unintended value.

nebula/ui/components/VPNAlert.qml
```diff
@@ -13,26 +13,26 @@ Rectangle {
...
     // Text of the Action (in bold)
-    property var alertActionText: ""
+    property string alertActionText: ""
```
MattLichtenstein, line 29: Don't need the empty string assignment

nebula/ui/components/VPNFocusOutline.qml
```diff
@@ -9,7 +9,7 @@ import Mozilla.VPN 1.0
 Rectangle {
     property variant focusedComponent
     property var focusColorScheme: VPNTheme.theme.blueButton
-    property var setMargins: -3
+    property real setMargins: -3
```
MattLichtenstein, line 12: I think we should use `int` in places where we realistically expect whole number, if for anything, for the sake of clarity so we know it would be unusual to set a decimal here

MattLichtenstein (approved):
Looks good, just some discussion points/precedent setting!

PR #5321, opened 2022-12-23
Introduce QmlPath - an XPath-like syntax for QML tree

bakulf:


flodolo:
On the localization side, at least for add-ons this seems to not create changes
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/3767063652

strings.yaml should probably be OK, not so sure about the strings still in the code.

lesleyjanenorton (approved):
Nice! Do we need to add some `conditions: {}` to the manifest so that folks on older versions don't get this?

src/shared/qmlpath.h
```diff
@@ -0,0 +1,103 @@
...
+ *   pick the one in position `number`.
+ * - '{' + propertyName + '}' - from all the items selected from the path
+ *   block, pick the ones that have a property with the name `propertyName`.
+ * - '{' + propertyName + '=' + propertyValue + '}' - from all the items
+ *   selected from the path block, pick the ones with the property with the name
+ *   `propertyName` and value set to `propertyValue`.
+ *
+ * Paths blocks can be concatenated: `/abc//foo[1]{p=42}/bar`
+ *
+ * Additionally, you can run a command to the final path with the syntax:
```
strseb, line 33: I am a BIG fan of this x-path style, **but**. 
Having statements to both query- and change state, that sounds too powerful - lot's of concerns here. 
I'd prefer this being restricted to output ref's that you can pass to an operator-function. 
This would allow us to make sure the context when Action X on Ref Y is performed is correct (see addon-manifest-question). 
Also there does not seem to be any user of the command-style right now, imho we should upgrade the old inspector actions to use QmlPaths instead of ::findName,

addons/tutorial_01_get_started/manifest.json
```diff
@@ -26,39 +27,39 @@
...
           "op": "signal",
-          "qml_emitter": "serverListButton-btn",
+          "qml_emitter": "//serverListButton-btn",
           "signal": "visibleChanged"
         }
       },
       {
         "id": "countryAU",
-        "element": "serverCountryList/serverCountry-au",
+        "element": "//serverCountryList/{className=VPNServerCountry}[0]",
```
strseb, line 36: Given this QT-Path will be evaluated, can this run the command part too? 
Is the scope sandboxed, or can any add-on theoretically change any property of any object?

strseb (changes requested):

src/shared/qmlpath.h
bakulf: I agree with you. Let's drop that part.

src/shared/qmlpath.cpp
```diff
@@ -0,0 +1,287 @@
...
+  if (obj.isEmpty()) {
+    return false;
+  }
+
+  qsizetype size = obj.size();
+  const QChar* input = obj.constData();
+
+  QList<Data> blocks;
+
+  while (true) {
```
oskirby, line 25: This took me a while to get my head around to check for correctness, but I think it's good now.

src/shared/qmlpath.cpp
```diff
@@ -0,0 +1,287 @@
...
+        filterByProperty(results, filter);
+        break;
+    }
+  }
+
+  if (results.isEmpty()) {
+    return nullptr;
+  }
+
+  return evaluateItems(results[0], collectChildItems(results[0]), i + 1);
```
oskirby, line 211: Perhaps I am overthinking this, but am getting the impression that this can prevent complicated matches from working as expected. For example, if we consider an example similar to the syntax blurb in the header file: `/abc//foo{p=42}/bar`

And give it the following QML:
```
Item {
  objectName: "abc"
  Item {
    objectName: "foo"
    property p: 42
    Item {
      objectName: "not-bar"
    }
  }
  Item {
    objectName: "example"
    Item {
      objectName: "foo"
      property p: 42
      Item {
       objectName: "bar"
      }
   }
}
```

Then presumeably, we will wind up with both `foo` objects in `results` but by only evaluating the first entry in the list we never actually find the one which satisfies the full path being filtered against. I think this needs to iterate on `results` until either a match is found or the list is exhausted.

oskirby (commented):
I like this, I am actually more excited to see what it can do for testing by giving us a means to reach any QML item without needing to assign unique names to literally everything.

src/shared/qmlpath.cpp
oskirby: And when considering that nesting can result in multiple matches, it's probably worth documenting somewhere exactly which item will be found (IIUC, I think we get the first match found in a pre-order traversal of the QML tree).

addons/tutorial_01_get_started/manifest.json
```diff
@@ -26,39 +27,39 @@
...
         "next": {
           "op": "signal",
-          "qml_emitter": "serverListButton-btn",
+          "qml_emitter": "//serverListButton-btn",
           "signal": "visibleChanged"
         }
       },
       {
         "id": "countryAU",
-        "element": "serverCountryList/serverCountry-au",
```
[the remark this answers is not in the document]
bakulf: the point here is to have a way to pick the first country without specifying its code/city. I dont want to change the IDs because I want to avoid the retranslation of the strings.

addons/tutorial_01_get_started/manifest.json
```diff
@@ -26,39 +27,39 @@
...
           "op": "signal",
-          "qml_emitter": "serverCountryList/serverCountry-au/serverCityList/serverCity-Melbourne/radioIndicator",
+          "qml_emitter": "//serverCountryList/{className=VPNServerCountry}[0]//serverCityList/{className=VPNRadioDelegate}[0]",
           "signal": "visibleChanged"
         }
       },
       {
         "id": "toggle",
-        "element": "controllerToggle",
+        "element": "//controllerToggle",
```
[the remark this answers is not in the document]
bakulf: yes, but i would not work with the country/city selection.

src/shared/qmlpath.h
```diff
@@ -0,0 +1,83 @@
...
+
+  struct Data {
+    bool m_nested = false;
+    QString m_key;
+
+    QList<Filter> m_filters;
+  };
+
+ public:
+  explicit QmlPath(QObject* parent = nullptr);
```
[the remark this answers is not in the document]
bakulf: but then I need to implement a `isValid()` for invalid syntax.

src/shared/qmlpath.h
```diff
@@ -0,0 +1,83 @@
...
+ *   pick the one in position `number`.
+ * - '{' + propertyName + '}' - from all the items selected from the path
+ *   block, pick the ones that have a property with the name `propertyName`.
+ * - '{' + propertyName + '=' + propertyValue + '}' - from all the items
+ *   selected from the path block, pick the ones with the property with the name
+ *   `propertyName` and value set to `propertyValue`.
+ *
+ * Paths blocks can be concatenated: `/abc//foo[1]{p=42}/bar`
+ */
+class QmlPath final : public QObject {
```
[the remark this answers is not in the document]
bakulf: No reasons. For the memory-management via parenting.

src/shared/qmlpath.cpp
bakulf: The first implementation was like you described. But this was incompatible with the 'range' implementation.
Let's use your QML tree. If I do: `/abc//foo[1]` I pick the nested foo. `/abc/foo[0]`, picks the first one. If you don't specify the index, the first one is evaluated. Basically, the current logic is that we have an implicit `[0]` at the end of each path.

bakulf: Actually, you are right. Fixed.

src/shared/qmlpath.h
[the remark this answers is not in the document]
oskirby: That should be pretty straightforeward, since the parser already does the work of erroring out on an invalid syntax: `bool isValid() const { return !m_blocks.isEmpty(); }`

src/shared/qmlpath.h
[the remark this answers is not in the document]
oskirby: I don't think it's necessary here and it just adds clutter.

oskirby (approved):

PR #5322, opened 2022-12-24
Vpnglean dependency cleanup

oskirby:
## Description
The new `vpnglean` module is fraught with some dependency issues as it pertains to the assorted bits of generated code. This PR attempts to minimize these dependencies, and improve the ability of CMake to detect when code needs to be rebuilt.

There are a few ways in which this is achieved:
- We renamed the `vpnglean` Rust library to `vpnglean_bindings` and create a new static `vpnglean` library in its place.
- The `vpnglean` static library now builds all the generated code plus any support classes. This solves many of the dependency issues that cause Qt's MOC to fail on generated sources by moving it all into one directory scope, and works around [this](https://gitlab.kitware.com/cmake/cmake/-/issues/18399) CMake bug.
- Move all the generated source files into `${CMAKE_CURRENT_BINARY_DIR}` so that we don't leave state in the source directory that can foul up later builds.
- As a general cleanup pass, try to minimize the amount of generated content in header files.

## Reference
Github: #5396  ([VPN-3627](https://mozilla-hub.atlassian.net/browse/VPN-3627))

## Checklist


bakulf (approved):

tests/unit/testaddon.cpp
```diff
@@ -1147,11 +1147,11 @@ void TestAddon::telemetry_state_change() {
   // Event metrics can only record InvalidValue or InvalidOverflow errors
   // https://mozilla.github.io/glean/book/reference/metrics/event.html#recorded-errors
 
-  QCOMPARE(mozilla::glean::sample::addon_state_changed.testGetNumRecordedErrors(
-               VPNGlean::InvalidValue),
+  QCOMPARE(mozilla::glean::sample::addon_state_changed.numRecordedErrors(
```
brizental, line 1150: Please change this back to `testGetNumRecordedErrors`. It is important we don't diverge from the Glean APIs (https://mozilla.github.io/glean/book/reference/metrics/event.html#testgetnumrecordederrors)

tests/unit/testaddon.cpp
```diff
@@ -1147,11 +1147,11 @@ void TestAddon::telemetry_state_change() {
...
+               EventMetric::InvalidValue),
```
brizental, line 1151: The ErrorType is not metric specific. It should remain in the VPNGlean class.

vpnglean/CMakeLists.txt
```diff
@@ -2,6 +2,31 @@
...
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_QUIET
+)
+
+## Add a static library for the generated C++ code.
+add_library(vpnglean STATIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/glean/event.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/glean/ping.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/event.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/ping.cpp
```
brizental, line 18: This is not really generated code though. The comment on L13 needs to be clarified.

vpnglean/CMakeLists.txt
```diff
@@ -48,73 +73,81 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten")
...
+        ${CMAKE_CURRENT_BINARY_DIR}/glean/generated/pings.cpp
+        ${CMAKE_CURRENT_BINARY_DIR}/glean/generated/pings.h
+        ${CMAKE_CURRENT_BINARY_DIR}/glean/generated/metrics.rs
+        ${CMAKE_CURRENT_BINARY_DIR}/glean/generated/pings.rs
     DEPENDS
         ${CMAKE_CURRENT_SOURCE_DIR}/glean_parser_ext/run_glean_parser.py
+        ${CMAKE_CURRENT_SOURCE_DIR}/glean_parser_ext/templates/cpp_metrics_header.jinja2
+        ${CMAKE_CURRENT_SOURCE_DIR}/glean_parser_ext/templates/cpp_metrics_source.jinja2
+        ${CMAKE_CURRENT_SOURCE_DIR}/glean_parser_ext/templates/cpp_pings_header.jinja2
+        ${CMAKE_CURRENT_SOURCE_DIR}/glean_parser_ext/templates/cpp_pings_source.jinja2
```
brizental, line 136: Hehe, you got me. I was too lazy to split that up into header and cpp file 😳

vpnglean/glean_parser_ext/templates/cpp_metrics_header.jinja2
```diff
@@ -22,21 +22,21 @@ namespace mozilla::glean {
...
-{% endfor %}
-} // namespace mozilla::glean
 
-{% for category_name, objs in all_objs.items() %}
-  {% for obj in objs.values() %}
-  {% if obj|attr("_generate_enums") %}
-{{ macros.generate_extra_keys_parser(obj, category_name) }}
-  {%- endif %}
   {% endfor %}
+} // namspace {{ category_name|snake_case }}
```
brizental, line 37:
```suggestion
} // namespace {{ category_name|snake_case }}
```

vpnglean/src/event.cpp
```diff
@@ -33,16 +26,45 @@ void EventMetric::record(const QJsonObject& extras) {
...
   if (!ffiExtras.keys.empty()) {
-#if not(defined(MZ_WASM) || defined(BUILD_QMAKE))
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
     glean_event_record(m_id, ffiExtras.keys.data(), ffiExtras.values.data(),
                        ffiExtras.keys.size());
 #endif
   } else {
-    logger.error() << "Attempted to record an event with extras, but no extras "
-                      "were provided. Ignoring.";
+    qWarning() << "Attempted to record an event with extras, but no extras "
```
brizental, line 66: Why switch from the logger?

brizental (changes requested):

vpnglean/src/event.cpp
oskirby: The general idea here is that `vpnglean` should be a standalone library, which means that it can't depend on anything in the VPN app or we risk creating an unsolvable dependency graph.

vpnglean/CMakeLists.txt
oskirby: It isn't really necessary, I just think it's a little cleaner this way.

tests/unit/testaddon.cpp
oskirby: I disagree, the `ErrorType` is only used as a part of the `EventMetric` class, so it should remain as a part of this library. Moving it back into `VPNGlean` would introduce circular dependencies between `vpnglean` and the VPN app.

oskirby: If we think this enum needs to be moved out of the `EventMetric` class, I would create a new header for it in the `vpnglean` directory.

brizental: This _will_ be moved out of the EventMetric class. Glean has multiple metric types. Although we currently only implement the Event metric, the plan is to implement more as we move on. All metric types share the ErrorType enum. As the comment says, we don't need to define this enum at all and instead import it through the FFI. Let's just leave it like this for now an work on https://mozilla-hub.atlassian.net/browse/VPN-3339 after.

vpnglean/CMakeLists.txt
brizental: Yeah I know and I agree it's cleaner.

vpnglean/src/event.cpp
brizental: Got it, hm. It would be overkill to turn the logger into a santdalone library as well so that we can use it here... However, would you mind adding a quick comment as to why we are going with a different logging method just here?

brizental (approved):

PR #5335, opened 2022-12-28
VPN-3485: Android: Fix background-connection timer

strseb:
## Description

In case the client is paused (not closed) because it went into the background and we have a non-client activation, the timestamp is off. Because the client will always use timestamp on when `connected()` is called. 
On android... or mobile in general, we can have a situation where `connected()`will be received way later. 

This pr add's an optional timestamp-argument to `connected()` for controllers to provide the right connection time.


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -51,14 +51,15 @@ class VPNService : android.net.VpnService() {
...
+                mConnectionTime = System.currentTimeMillis()
                 Log.i(tag, "Dispatch Daemon State -> connected")
                 mBinder.dispatchEvent(
                     VPNServiceBinder.EVENTS.connected,
                     JSONObject().apply {
+                        put("time", mConnectionTime)
                         put("city", mCityname)
                     }.toString()
                 )
-                mConnectionTime = System.currentTimeMillis()
```
bakulf, line 61: Why cannot we pass the connection time in the initialize method?

strseb: That should have been fixed. 
We get the time from initialize only when the app closed/killed. Are you suggesting we should reinitialize the controller anytime we move from background/stopped -> started?

bakulf (approved):

PR #5337, opened 2022-12-28
Rollback vpn controller for rollback tutorials

MattLichtenstein:
## Description

- Rollback the state of the VPN controller to what it was prior to starting a "needs rollback" tutorial

## Reference

[VPN-3125: Implement the Multi-hop Tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3125)
[VPN-3124: Implement the App Permissions Tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3124)

## Checklist
    
### DIFF
--- a/src/apps/vpn/addons/addontutorial.cpp
+++ b/src/apps/vpn/addons/addontutorial.cpp
@@ -13,6 +13,7 @@
 #include "itempicker.h"
 #include "leakdetector.h"
 #include "logger.h"
+#include "mozillavpn.h"
 #include "tutorial/tutorial.h"
 #include "tutorial/tutorialstep.h"
 
@@ -120,6 +121,12 @@ void AddonTutorial::play(const QStringList& allowedItems) {
   m_activeTransaction = false;
 
   if (settingsRollbackNeeded()) {
+
+    MozillaVPN* vpn = MozillaVPN::instance();
+    Controller* controller = vpn->controller();
+    Q_ASSERT(controller);
+    m_wasVPNOn = controller->state() == Controller::State::StateOn;
+
     m_activeTransaction = SettingsHolder::instance()->beginTransaction();
     if (!m_activeTransaction) {
       logger.warning() << "Unable to start a setting transaction";
@@ -148,14 +155,12 @@ void AddonTutorial::stop() {
     m_steps[m_currentStep]->stop();
   }
 
-  if (m_activeTransaction &&
-      !SettingsHolder::instance()->rollbackTransaction()) {
-    logger.warning() << "Unable to rollback a setting transaction";
+  if(m_settingsRollbackNeeded) {
+      beingRollbackSettings();
   }
 
   m_itemPicker->stop();
   m_currentStep = -1;
-  m_activeTransaction = false;
 
   if (m_navigatorReloader) {
     m_navigatorReloader->deleteLater();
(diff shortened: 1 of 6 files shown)


src/apps/vpn/addons/addontutorial.cpp
```diff
@@ -120,6 +121,12 @@ void AddonTutorial::play(const QStringList& allowedItems) {
   m_activeTransaction = false;
 
   if (settingsRollbackNeeded()) {
+
+    MozillaVPN* vpn = MozillaVPN::instance();
```
bakulf, line 125: This PR creates a strong connection between the tutorial engine and the controller. Instead, i would suggest using signals:
1. Creates signals when a transaction begins, finalizes, rollbacks.
2. In the controller, connect that signals to "something" that does the correct logic.
Would it work?

src/shared/settingsholder.h
```diff
@@ -110,6 +110,8 @@ class SettingsHolder final : public QObject {
  signals:
   void addonSettingsChanged();
   void inTransactionChanged();
+  void transactionBegan();
+  void transactionRolledBack();
```
bakulf, line 114: Add a transition committed signal even if not used.

src/shared/settingsholder.cpp
```diff
@@ -296,6 +297,8 @@ bool SettingsHolder::rollbackTransaction() {
     QMetaObject::invokeMethod(this, i.value().first, Qt::DirectConnection);
   }
 
+  emit transactionRolledBack();
```
bakulf, line 300: We should emit the signal even if `finalizeTransition` returns false to avoid inconsistent states.

src/apps/vpn/controller.cpp
```diff
@@ -128,6 +128,56 @@ void Controller::initialize() {
     logger.info() << "Canary Ping Succeeded";
   });
 
+  connect(SettingsHolder::instance(), &SettingsHolder::transactionBegan, this,
```
bakulf, line 131: This file is already too "polluted". What would if in the CTOR you create a new object `ControllerTransitionMonitor` (a better name, please ;) that does this logic ?

src/apps/vpn/controller.cpp
```diff
@@ -128,6 +128,56 @@ void Controller::initialize() {
...
+          // If vpn is already off, just turn it on
+          if (m_state == StateOff) {
+            activate();
+            return;
+          }
+
+          // run lambda function in a context so we can delete it after it runs
+          // once
+          QObject* context = new QObject(this);
+          connect(this, &Controller::stateChanged, context, [this, context]() {
```
bakulf, line 156: Instead of using a boolean, you can use an enum with values, not-in-transition, connected-before-transition, disconnected-before-transition. And you don't need this `context`.

src/apps/vpn/controller.cpp
```diff
@@ -128,6 +128,56 @@ void Controller::initialize() {
...
+            if (m_state == StateOff) {
+              activate();
+              delete context;
+            }
+          });
+        } else {
+          // If vpn is already on, just turn it off
+          if (m_state == StateOn) {
+            deactivate();
```
bakulf, line 165: What about context here?

bakulf (changes requested):

PR #5389, opened 2023-01-04
Check the iOS client version on itunes too - VPN-3621

bakulf:


src/apps/vpn/qmake/platforms/ios.pri
```diff
@@ -66,6 +66,7 @@ OBJECTIVE_SOURCES += \
     apps/vpn/platforms/ios/iosglue.mm \
     apps/vpn/platforms/ios/iosnetworkwatcher.mm \
     apps/vpn/platforms/ios/iosnotificationhandler.mm \
+    apps/vpn/platforms/ios/iosupdater.mm \
```
oskirby, line 69: Please add `iosupdater.mm` and `iosupdater.h` to CMake as well.

bakulf: I did, right?

src/apps/vpn/platforms/ios/iosupdater.mm
```diff
@@ -0,0 +1,101 @@
...
+  QJsonArray releases = json.object()["results"].toArray();
+  if (releases.isEmpty()) {
+    return false;
+  }
+
+  QString currentVersion = appVersion();
+
+  for (const QJsonValue& value : releases) {
+    QString version = value.toObject()["version"].toString();
+    if (VersionUtils::compareVersions(version, latestVersion) != 0) {
```
mcleinman, line 89:
```suggestion
    // version is Apple-reported (from iTunes request above), latestVersion is Guardian-reported (from VersionApi request). Only continue if Apple is able to provide the app version Guardian expects it to.
    if (VersionUtils::compareVersions(version, latestVersion) != 0) {
```

This looks great, but given how many different places we get version from, a comment giving some future wayfinding may be helpful.

mcleinman (commented):
Is there a recommended way to test this? I can figure out a way by hacking some variables in the code, but will QA need some steps as well?

oskirby (approved):

mcleinman (approved):
Caveat: I've tested this with hacking some variables in, but there has not been an opportunity to do end-to-end testing of this, given its reliance on version upgrades.

birdsarah:

PR #5398, opened 2023-01-04
Add task to perform server selection driven by GeoIP.

oskirby:
## Description
A significant number of our users will often install the VPN client, and use it in its default configuration indefinitely, meaning that they may never actually change their server away from the randomly-selected initial value. This can result in poor VPN performance if the random selection chooses a server on the opposite side of the internet.

A better approach for this initial selection would be to choose the nearest location and hope that it happens to correlate to the best network performance.

## Reference
Github #3682
JIRA [VPN-2333](https://mozilla-hub.atlassian.net/browse/VPN-2333) and [VPN-3111](https://mozilla-hub.atlassian.net/browse/VPN-3111)

## Checklist
    

src/apps/vpn/mozillavpn.cpp
```diff
@@ -422,8 +420,6 @@ void MozillaVPN::maybeStateMain() {
     return;
   }
 
-  Q_ASSERT(m_private->m_serverData.hasServerData());
```
bakulf, line 425: Why removing this? We must have a server selection before showing the main view. It seems that this is not guaranteed in this PR. 

I wonder if we should combine the serverlist task and this one in 1 single operation if the server data is empty.

We probably need a TaskSequence to wrap TaskServer and this new task, as part of the group of tasks to run during the authentication. 

And maybe we should keep this pickRandom if we arrive here without selection. 

Thought?

bakulf (changes requested):

src/apps/vpn/mozillavpn.cpp
oskirby: I kinda figured I this would have gotten flagged.

Yeah, I tried a few things on this one since we now do the server selection async, and there were a few things that came to mind:
- Add a call to get the geoIP as a part of the periodic refresh and cache it somewhere. Then server selection can be done synchronously, like with the `pickRandom()`. But then the geoIP gets very muddled once the user starts changing their location by activating the VPN. But this isn't really a big concern, and the GeoIP is already inaccurate if the VPN was active when launching the UI anyways.
- Schedule server selection as part of the periodic refresh, but just do a no-op if the server is already set, but this would work against my goal of adding an automatic server option.
- If the user does manage to get through the activation screens fast enough to arrive at the main screen before server selection, it doesn't actually break anything... they just get an empty server, and with [5812cf2](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5398/commits/5812cf2d9028edfb22a4c80d67324f398ecf8503) it just picks a server upon activation.
- My long term evil plan is to add an `Automatic` server option, which runs this task upon every activation in which case we can just leave the server unset intentionally via settings.

Ultimately, I felt that simply removing the assert was the cleanest option.

oskirby: After playing around a bit with an automatic server option, it seems that we kind of need a `Location` model anyways. So it actually better to put this assert back and deal with this another way.

src/apps/vpn/mozillavpn.cpp
```diff
@@ -1662,6 +1665,20 @@ void MozillaVPN::scheduleRefreshDataTasks(bool refreshProducts) {
           TaskGetSubscriptionDetails::NoAuthenticationFlow,
           ErrorHandler::PropagateError)};
 
+  // The VPN needs to be off in order to determine the client's real location.
+  // And it also needs to complete before TaskServers in case this triggers an
+  // automatic server selection.
+  //
+  // TODO: This ordering requirement can be relaxed in the future once automatic
```
bakulf, line 1672: do you want to file a jira issue to keep track of it?

bakulf (approved):

src/apps/vpn/models/servercountrymodel.cpp
```diff
@@ -269,6 +270,47 @@ QStringList ServerCountryModel::pickRandom() {
   Q_ASSERT(false);
 }
 
+// Select the city that we think is going to perform the best
+QStringList ServerCountryModel::pickBest(const Location& location) const {
+  double latitude = location.latitude();
+  double longitude = location.longitude();
+  double clientSin = qSin(latitude * M_PI / 180.0);
+  double clientCos = qCos(latitude * M_PI / 180.0);
+  if (qIsNaN(latitude) || qIsNaN(longitude)) {
```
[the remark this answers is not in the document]
oskirby: Good catch!

src/apps/vpn/models/servercountrymodel.h
```diff
@@ -41,7 +43,8 @@ class ServerCountryModel final : public QAbstractListModel {
 
   bool initialized() const { return !m_rawJson.isEmpty(); }
 
-  Q_INVOKABLE QStringList pickRandom();
+  Q_INVOKABLE QStringList pickRandom() const;
```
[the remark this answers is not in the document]
oskirby: Yeah, it is used to provide an initial value for the multihop view if the user has never set up multihop before. I think it's fine in this use case given that multihop is an advanced feature and we can expect the user to know what they're doing.

src/apps/vpn/mozillavpn.cpp
oskirby: I should!

Issue #5050, opened 2022-11-28
Segmentation fault with Qt 6.4.1 on Fedora 37


arnobl:
**Describe the bug**
After the system update that upgraded Qt from 6.3.1-2.fc37.x86_64 to 6.4.1-1.fc37.x86_64, the vpn crashes on start with a:

```
> mozillavpn             
QML debugging is enabled. Only use this in a safe environment.
[1]    8000 segmentation fault (core dumped)  mozillavpn
```

**VPN version:**
2.11.0 (2.202211100718)
Installed from the RPM package Github actions provide.

**Affected Platforms:**
Fedora 37

**Steps to reproduce:**
Start the app

**Expected result:**
Should not crash.

**Actual result:**
Crashes

**Additional notes:**

```
ldd $(which mozillavpn) 
	linux-vdso.so.1 (0x00007ffea1fb8000)
	libQt6Test.so.6 => /lib64/libQt6Test.so.6 (0x00007f768f5e4000)
	libQt6NetworkAuth.so.6 => /lib64/libQt6NetworkAuth.so.6 (0x00007f768ebcb000)
	libQt6DBus.so.6 => /lib64/libQt6DBus.so.6 (0x00007f768eafd000)
	libpolkit-gobject-1.so.0 => /lib64/libpolkit-gobject-1.so.0 (0x00007f768f5c3000)
	libgobject-2.0.so.0 => /lib64/libgobject-2.0.so.0 (0x00007f768ea9c000)
	libglib-2.0.so.0 => /lib64/libglib-2.0.so.0 (0x00007f768e95c000)
	libQt6WebSockets.so.6 => /lib64/libQt6WebSockets.so.6 (0x00007f768e928000)
	libQt6Widgets.so.6 => /lib64/libQt6Widgets.so.6 (0x00007f768e000000)
	libQt6Quick.so.6 => /lib64/libQt6Quick.so.6 (0x00007f768d800000)
	libQt6Qml.so.6 => /lib64/libQt6Qml.so.6 (0x00007f768d200000)
	libQt6Network.so.6 => /lib64/libQt6Network.so.6 (0x00007f768d03c000)
	libQt6Gui.so.6 => /lib64/libQt6Gui.so.6 (0x00007f768c600000)
	libQt6Core.so.6 => /lib64/libQt6Core.so.6 (0x00007f768be00000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f768ba00000)
[...]

PR #5433, opened 2023-01-05
Build RPM packages on Fedora 37

artfwo:
## Description

Update RPM package build configuration to use the current Fedora version 37. 

Additional dependencies on `qt6-qtbase-<driver>` packages are added as they are referenced by Qt6Sql cmake module and cmake configuration will fail without these packages.

## Reference

https://github.com/mozilla-mobile/mozilla-vpn-client/issues/5050

## Checklist
    

linux/mozillavpn.spec
```diff
@@ -28,6 +28,9 @@ BuildRequires: qt6-qtsvg-devel >= 6.0
 BuildRequires: qt6-qttools-devel >= 6.0
 BuildRequires: qt6-qtwebsockets-devel >= 6.0
 BuildRequires: qt6-qt5compat-devel >= 6.0
+BuildRequires: qt6-qtbase-mysql >= 6.0
```
bakulf, line 31: Why do we need these 3? We don't use them. Just SQLite.

artfwo: To use Qt6Sql cmake files these plugins have to be present during configuration stage, here is an example of an error that happens without these packages:

https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/3849926939/jobs/6560059259

I haven't found a single package that would install those drivers as dependencies, so I've added them to build dependencies one by one. There might be another way around this to avoid installing them, but I don't know of any.

oskirby: I've run into the same thing when playing around with RPM builds in taskcluster. I don't know why this became a problem in Fedora 37 and onwards, but so be it :)

oskirby (approved):
One concern that I have is whether the binary issue with Qt 6.4.1 would impact compatibility with users on Fedora 36 and earlier. (eg: if built with 6.4.1, does it run correctly for users with 6.2.x installed?).

However, since we don't officially support Fedora, we only really keep this workflow around as a smoke test to make sure we can build for RPM-based targets. Ultimately we want to produce official-looking builds with Taskcluster, but that is a far bigger task for another PR.

I am happy with merging this as-is.

linux/mozillavpn.spec
oskirby: One thing worth checking is whether we need to add a `Requires: qt6-qtbase-mysql` as well for Qt 6.4.x. You might want to check the logs to see if there are any warnings about the SQL driver not being found as this can affect how data is stored by QML.

artfwo: This doesn't seem to be a problem - I'm currently using the RPM build from most recent workflow run on this PR and I've only got sqlite driver installed on my system.

artfwo: Right, and also no warnings in the debug logs about any missing db drivers.

artfwo:
> One concern that I have is whether the binary issue with Qt 6.4.1 would impact compatibility with users on Fedora 36 and earlier. (eg: if built with 6.4.1, does it run correctly for users with 6.2.x installed?).

AFAIK, Fedora doesn't have LTS releases and the current version is always the supported version, so as long as previous stable builds are working I'd say it's ok to migrate to 37 as that's probably what the majority of the users have by now.

dannycolin:
> We say maintained for ''approximately 13 months'' because the supported period for releases is dependent on the date the release under development goes final. As a result, ''Release N'' is supported until four weeks after the release of ''Release N+2''.

This means that 36 will be supported up to 4 weeks after the release of 38.

PR #5435, opened 2023-01-05
Fix the get-started tutorial when a server is not available - VPN-3552

bakulf:
This is a fix in case:
- 1 city from the first country is unavailable
- all the cities from the first country are unavailable

It's not a fix in case, the first and the second countries do not have cities available. If we want to support this scenario, we need to implement the scrolling. And this becomes out-of-scope for 2.13/2.14.

### DIFF
--- a/addons/tutorial_01_get_started/manifest.json
+++ b/addons/tutorial_01_get_started/manifest.json
@@ -20,8 +20,8 @@
           "op": "vpn_off"
         },{
           "op": "vpn_location_set",
-          "exitCountryCode": "at",
-          "exitCity": "Vienna",
+          "exitCountryCode": "it",
+          "exitCity": "Milan",
           "entryCountryCode": "",
           "entryCity": ""
         }],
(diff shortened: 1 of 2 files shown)


addons/tutorial_01_get_started/manifest.json
```diff
@@ -20,8 +20,8 @@
           "op": "vpn_off"
         },{
           "op": "vpn_location_set",
-          "exitCountryCode": "at",
-          "exitCity": "Vienna",
```
MattLichtenstein, line 24: Why was this changed?

bakulf: because Austria is the second country in the list (at least in most of the languages). And if Australia is not usable, we switch to the second one, which must be:
- not opened
- not already selected

MattLichtenstein: would it be feasible to add a check to the `hasAvailableCities` property to make sure it has at least one unselected city, and then use that to query for the first country with `hasAvailableCities=true`?

bakulf: Yes, but we do not support scrolling in tutorial yet...

MattLichtenstein (approved):

PR #5419, opened 2023-01-05
Move troubleshooting info for multi-hop into list

flozia:
## Description

Moves the dedicated callout info into the bullet list. Instead of always showing this string for multi-hop connection we decided to only show it for speed test results that we mark as “slow”.

## Reference

- Jira [VPN-3218](https://mozilla-hub.atlassian.net/browse/VPN-3218): Improvements to the Speed test UI
- Figma: https://www.figma.com/file/3V7aahpQ096Qsp2L3SCT6h/Connection-Screen-Speedtest?node-id=1103%3A4728

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #5440, opened 2023-01-05
Remove multi-hop callout string

flozia:
## Description

In PR #5419 I introduced a variation of a string due to recent design changes. I was hesitant to remove the redundant string in case we’d need it again. I’m assuming they’ll be removed from Pontoon @flodolo? I’m not entirely sure what our current policy regarding unused strings is. Accumulating and having legacy strings floating around is definitely not ideal so I’m happy with whatever we decide on.

## Reference

- PR #5419

## Checklist
    
### DIFF
--- a/src/apps/vpn/ui/screens/devices/ViewDevices.qml
+++ b/src/apps/vpn/ui/screens/devices/ViewDevices.qml
@@ -153,6 +153,4 @@ VPNViewBase {
             }
         }
     ]
-
-    Component.onDestruction: menu.rightTitle = ""
 }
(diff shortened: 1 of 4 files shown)


MattLichtenstein (approved):

flodolo:
> I was hesitant to remove the redundant string in case we’d need it again. I’m assuming they’ll be removed from Pontoon @flodolo? I’m not entirely sure what our current policy regarding unused strings is. Accumulating and having legacy strings floating around is definitely not ideal so I’m happy with whatever we decide on.

We should always remove unused strings.

Note that they will stay around until there is a release branch that requires that string, so there is a bit of buffer in case we change idea.

flodolo (approved):

lesleyjanenorton (approved):

PR #5456, opened 2023-01-06
Rollback vpn controller for rollback tutorials

MattLichtenstein:
## Description

- Rollback the state of the VPN controller to what it was prior to starting a "needs rollback" tutorial

*Note: This relies on: #5450

## Reference

[VPN-3125: Implement the Multi-hop Tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3125)
[VPN-3124: Implement the App Permissions Tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3124)
[VPN-3593: VPN previous state not restored after interrupting the “How to exclude apps from VPN protection” tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3593)
[VPN-3677: The VPN state should be restored after a tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3677)
[VPN-3596: [iOS] VPN is turned OFF after completing the multi-hop tutorial](https://mozilla-hub.atlassian.net/browse/VPN-3596)

## Checklist
    
### DIFF
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -130,6 +130,17 @@ void Controller::initialize() {
     logger.info() << "Canary Ping Succeeded";
   });
 
+  connect(SettingsHolder::instance(), &SettingsHolder::transactionBegan, this,
+          [this]() { m_connectedBeforeTransaction = m_state == StateOn; });
+
+  connect(SettingsHolder::instance(),
+          &SettingsHolder::transactionAboutToRollBack, this, [this]() {
+            if (m_connectedBeforeTransaction)
+              MozillaVPN::instance()->activate();
+            else
+              MozillaVPN::instance()->deactivate();
+          });
+
   MozillaVPN* vpn = MozillaVPN::instance();
 
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -135,10 +135,14 @@ void Controller::initialize() {
 
   connect(SettingsHolder::instance(),
           &SettingsHolder::transactionAboutToRollBack, this, [this]() {
+            if (!m_connectedBeforeTransaction)
+              MozillaVPN::instance()->deactivate();
+          });
+
+  connect(SettingsHolder::instance(), &SettingsHolder::transactionRolledBack,
+          this, [this]() {
             if (m_connectedBeforeTransaction)
               MozillaVPN::instance()->activate();
-            else
-              MozillaVPN::instance()->deactivate();
           });
 
   MozillaVPN* vpn = MozillaVPN::instance();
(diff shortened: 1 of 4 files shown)


src/shared/settingsholder.h
```diff
@@ -110,6 +110,9 @@ class SettingsHolder final : public QObject {
  signals:
   void addonSettingsChanged();
   void inTransactionChanged();
+  void transactionBegan();
+  void transactionAboutToRollBack();
+  void transactionRolledBack();
```
birdsarah, line 115: I don't see `transactionRolledBack()` used anywhere. What am I missing?

src/apps/vpn/controller.cpp
```diff
@@ -128,6 +128,17 @@ void Controller::initialize() {
     logger.info() << "Canary Ping Succeeded";
   });
 
+  connect(SettingsHolder::instance(), &SettingsHolder::transactionBegan, this,
+          [this]() { m_connectedBeforeTransaction = m_state == StateOn; });
+
+  connect(SettingsHolder::instance(),
+          &SettingsHolder::transactionAboutToRollBack, this, [this]() {
```
birdsarah, line 135: I don't have enough background yet on the settings transaction setup. Why do we activate before rollback rather than after rollback? Based on nothing but naming I have thought we want to activate after rollback when our selected server is restored to it's original value. I'm sure this is just me not understanding I'm asking so I can get up to sped.

birdsarah (commented):
I have a question that I'd like to talk about for my learning.

I think you need to remove an extra signal....or again maybe I'm missing something.

src/apps/vpn/controller.cpp
bakulf: We want it here because, in case of a disconnect, if we do it after, first we end up doing a server switch, then a disconnect.

birdsarah:
Am I right in thinking this resolves VPN-3596 @MattLichtenstein?

src/apps/vpn/controller.cpp
bakulf: Matt and i are working on it. I can review this PR.

src/apps/vpn/controller.cpp
```diff
@@ -128,6 +128,17 @@ void Controller::initialize() {
...
+            if (m_connectedBeforeTransaction)
```
bakulf, line 136: We want to activate the VPN after the rollback.

bakulf (changes requested):

MattLichtenstein:
> Am I right in thinking this resolves [VPN-3596](https://mozilla-hub.atlassian.net/browse/VPN-3596) @MattLichtenstein?

@birdsarah from my testing, it looks like this is resolved. Added a reference to that ticket in the description

src/shared/settingsholder.cpp
```diff
@@ -296,6 +300,8 @@ bool SettingsHolder::rollbackTransaction() {
     QMetaObject::invokeMethod(this, i.value().first, Qt::DirectConnection);
   }
 
+  emit transactionRolledBack();
```
bakulf, line 303: Instead of this, can you do this: `qScopeGuard([&] { emit transactionRolledBack(); }` before `if (!finalizeTansaction()` ?

In this way, if something goes wrong, we still emit the transactionRolledBack.

bakulf (changes requested):

bakulf (approved):

PR #5457, opened 2023-01-06
Try to fix flaky functional tests

birdsarah:
Revert's @bakulf's parts 8, 9 and 10

Part 8 seems to have introduced instability in the functional tests....I could be wrong...


bakulf:
Which tests are you referring to? Strange. It was all green before landing. I'll take a look tonight

oskirby (approved):

PR #5459, opened 2023-01-06
Fix the wasm client functional test: testServers.js

bakulf:
The rollback of modules 8,9,10 was not correct. The real fix was related to the qmlPath porting of our tests.
Somehow there was a extra line which was "fine" for faster clients (dummyvpn), but not for wasm. The extra line was opening the server-list, closing the main view. On wasm, the main view is "released" too soon. That line needs to be removed.

### DIFF
--- a/tests/functional/testServers.js
+++ b/tests/functional/testServers.js
@@ -137,8 +137,6 @@ describe('Server list', function() {
 
     await vpn.waitForQueryAndClick(
         queries.screenHome.serverListView.BACK_BUTTON.visible());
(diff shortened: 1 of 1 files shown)


Gela (approved):

birdsarah:
I'm fine with being wrong about reverting 8,9,10. (To be clear I was thinking it was 8, the urlOpener, that was the problem. I had to revert 9 and 10 too because of the chain of dependencies.)

With regards to this PR, a good chunk of recent failures have been on testSubscription so I don't understand how this would address that.

PR #5503, opened 2023-01-10
VPN-3508 - part 14 - Cleanup networkRequests

bakulf:
In this commit I do the following:
    - implement a callback system to overwrite the default network behavior (unit-test, wasm)
    - move the single network request constructors in the corresponding code
    - move networkrequest.{cpp,h} into /src/shared
    - remove mocnetworkrequest from unit-tests

    Revising the network-request implementation I also found a bug (and fixed it) in
    the benchmark task logic where, if the upload is slower than the download, we
    do not show the proper error message.


src/apps/vpn/mozillavpn.cpp
```diff
@@ -1768,3 +1766,15 @@ void MozillaVPN::registerUrlOpenerLabels() {
                                        : AppConstants::API_STAGING_URL);
   });
 }
+
+// static
+QByteArray MozillaVPN::authorizationHeader() {
```
birdsarah, line 1771: Yay. This is easy to add a test for now...can you add it?

src/apps/vpn/update/versionapi.cpp
```diff
@@ -29,7 +30,9 @@ VersionApi::~VersionApi() {
 }
 
 void VersionApi::start(Task* task) {
-  NetworkRequest* request = NetworkRequest::createForVersions(task);
+  NetworkRequest* request = NetworkRequest::create(task, 200);
+  request->get(
+      QString("%1/api/v1/vpn/versions").arg(AppConstants::apiBaseUrl()));
```
birdsarah, line 35: I think it would be really helpful if all the Url Paths were in one place / file.

Ideally putting them all in one place:
* makes it easy to scan through the list
* means we're not repeating `.arg(AppConstants::apiBaseUrl())` everywhere.

birdsarah (commented):
This is a partial review....I'm working my way through it. But figured I'd share based on how far I'd gotten. I'll pick this up again tomorrow.

wasm/test.html
```diff
@@ -91,11 +91,16 @@ <h1>Mozilla VPN - <span class="web-viewer">WebAssembly Viewer</span></h1>
...
-     setTimeout(() => Module.mvpnNetworkResponse(id, JSON.stringify({ status: 200, body: btoa(String.fromCharCode.apply(null, new Uint8Array(1024))) })), 15000);
-      return;
-    }
+  if (url === "https://archive.mozilla.org/pub/vpn/speedtest/50m.data") {
+    // 50mb of data is too much to be handled in the browser.
+   setTimeout(() => Module.mvpnNetworkResponse(id, JSON.stringify({ status: 200, body: btoa(String.fromCharCode.apply(null, new Uint8Array(1024))) })), 1000);
+    return;
+  }
+
+  if (url == "https://dev.vpn-network-benchmark.nonprod.webservices.mozgcp.net/upload") {
```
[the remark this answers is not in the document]
bakulf: This is just for wasm tests. Actually, I would like to remove this all. Benchmark tests are the only reason why functional tests cannot run offline. We should implement something like what we have done for other services: mock all the responses.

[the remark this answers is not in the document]
bakulf: Btw: the reason why this is needed is because that endpoint does not support CORS.

src/apps/vpn/mozillavpn.cpp
bakulf: yes :)

src/apps/vpn/update/versionapi.cpp
bakulf: Good point!

src/apps/vpn/tasks/authenticate/taskauthenticate.cpp
```diff
@@ -49,13 +50,15 @@ void TaskAuthenticate::run() {
           this, &Task::completed);
 
   connect(m_authenticationListener, &AuthenticationListener::completed, this,
-          [this, pkceCodeVerifier](const QString& pkceCodeSucces) {
+          [this, pkceCodeVerifier](const QString& pkceCodeSuccess) {
             logger.debug() << "Authentication completed with code:"
-                           << logger.sensitive(pkceCodeSucces);
+                           << logger.sensitive(pkceCodeSuccess);
```
birdsarah, line 55: Success :D :D

src/shared/networkrequest.cpp
```diff
@@ -0,0 +1,487 @@
...
+#include <QUrl>
+
+#include "leakdetector.h"
+#include "logger.h"
+#include "networkmanager.h"
+#include "settingsholder.h"
+#include "task.h"
+
+// Timeout for the network requests.
+constexpr uint32_t REQUEST_TIMEOUT_MSEC = 15000;
```
birdsarah, line 23: Is this enough time for the delete devices request?

src/shared/networkrequest.cpp
```diff
@@ -0,0 +1,487 @@
...
+bool NetworkRequest::isRedirect() const {
+  int status = statusCode();
+  return status >= 300 && status < 400;
+}
+
+void NetworkRequest::handleHeaderReceived() {
+  // Suppress this signal if a redirect is about to happen.
+  int policy =
+      m_request.attribute(QNetworkRequest::RedirectPolicyAttribute).toInt();
+  if (isRedirect() && (policy != QNetworkRequest::ManualRedirectPolicy)) {
```
birdsarah, line 274: Just a question: Is this future proofing against wanting to use ManualRedirectPolicy at a later time?

src/shared/networkrequest.cpp
```diff
@@ -0,0 +1,487 @@
...
+      m_redirectedUrl = url().resolved(redirectUrl);
+    } else {
+      m_redirectedUrl = m_redirectedUrl.resolved(redirectUrl);
+    }
+  } else {
+    m_redirectedUrl = redirectUrl;
+  }
+  emit requestRedirected(this, m_redirectedUrl);
+#else
+  emit requestRedirected(this, redirectUrl);
```
birdsarah, line 304: Qt has a `redirected` signal, curious why we need this.

src/shared/networkrequest.cpp
birdsarah: I see we use disableTimeout for add device and balrog.

src/shared/networkrequest.cpp
```diff
@@ -0,0 +1,487 @@
...
+    logger.error() << error.errorString();
+    auto cert = error.certificate();
+    if (!cert.isNull()) {
+      logger.info() << "Related Cert:";
+      logger.info() << cert.toText();
+    }
+  }
+}
+
+void NetworkRequest::enableSSLIntervention() {
```
birdsarah, line 462: Would love a docstring about why sslIntervention is necesary.

birdsarah (commented):
Love this PR. The interface is much cleaner. I've left a few more comments many of which may just be things that I need to understand better.

My one ask is a follow-up from yesterday and I apologize for not being clear (I've sent you a longer version of this on slack).

Before I could go the network requests file and it was a little bit of scrolling but I could see all the api paths in one place. Now they're scattered throughout the code base.

Based on what I've seen in your PR it seems like AppConstants could hold this.

You currently have diffs like

Before:

```
NetworkRequest* request = NetworkRequest::createForVersions(task);
```

After:

```
  NetworkRequest* request = NetworkRequest::create(task, 200);
  request->get(AppConstants::apiUrl("/api/v1/vpn/versions"));
```

And I would love the After to be something like:

```
  NetworkRequest* request = NetworkRequest::create(task, 200);
  request->get(AppConstants:urlForVersions());
```

What do you think?

We already have `AppConstants::balrogUrl()` - we can have a whole url section of the AppConstants file to help us see all our external calls.

src/shared/networkrequest.cpp
bakulf: Right. Let's do it here as well.

src/shared/networkrequest.cpp
```diff
@@ -0,0 +1,487 @@
...
+  }
+
+  QNetworkAccessManager* manager =
+      NetworkManager::instance()->networkAccessManager();
+  handleReply(manager->post(m_request, body));
+
+  m_timer.start(REQUEST_TIMEOUT_MSEC);
+}
+
+void NetworkRequest::deleteRequest(const QUrl& url) {
```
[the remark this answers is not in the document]
bakulf: you renaming all the getRequest, postRequest, etc to getResource, postResource etc ?

src/shared/networkrequest.cpp
bakulf: This is a workaround for a horrible bug in Qt6, fixed in 6.2: https://bugreports.qt.io/browse/QTBUG-100651
This code should be removed as soon as we force qt6.2 as min version

src/shared/networkrequest.cpp
bakulf: Because we don't want to expose our internal QNetworkReply. The first argument (or the sender()) is the NetworkRequest object.

bakulf:
I don't like to have all those methods. Let's use an enum. PR updated.

birdsarah (approved):

birdsarah (changes requested):
I'm removing my Approve review because 

https://mozilla-hub.atlassian.net/browse/VPN-3882 / https://github.com/mozilla-mobile/mozilla-vpn-client/issues/5625

appears to be specific to this branch. See my testing notes on that issue.

birdsarah:
Is there a way to run the functional tests not in an offline mode, but really hitting guardian?

birdsarah:
One last thing. 

When I was running through debugging I noticed that there are loads of requests to the /versions end point. They always return 200.

When running this branch I got a few 502's before settling into 200s. Could be nothing, but mentioning just in case.

bakulf:
> appears to be specific to this branch. See my testing notes on that issue.

Good catch! Fixed.

bakulf:
> Is there a way to run the functional tests not in an offline mode, but really hitting guardian?

No, on purpose. The test simulates different account configuration and this is not possible if using guardian for real:
- multiple subscription types
- different device registrations
- FxA accounts with many scenarios (totp, blocked account, etc)

bakulf:
> When running this branch I got a few 502's before settling into 200s. Could be nothing, but mentioning just in case.

Version requests is a simple GET, with no custom headers (no auth). I wonder if we should debug this in guardian too.

bakulf:
To avoid similar bugs (a missing auth header), I added a `requiredHeaders` array + logic in the functional tests.

birdsarah (approved):

PR #5506, opened 2023-01-10
Fix VPN-3613: Close IP info panel on VPNController state changes

lesleyjanenorton:
## Description

Fixes VPN-3613

## Reference

VPN-3613

## Checklist
    

nebula/ui/components/VPNControllerView.qml
```diff
@@ -148,6 +148,10 @@ Item {
                 target: animatedRings
                 visible: false
             }
+            PropertyChanges {
+                target: ipInfoPanel
+                isOpen: false
+            }
```
MattLichtenstein, line 154: Why do we only close the ip info panel on certain state changes instead of any state changes?

lesleyjanenorton: For absolutely no good reason - thank you for the catch. Refactored for less verbosity. Also just realized this was missing a file and some other bits I probably lost jumping between branches. All here now.

nebula/ui/components/VPNControllerView.qml
```diff
@@ -594,18 +591,27 @@ Item {
...
         }
-        enabled: !connectionInfoScreenVisible && !ipInfoPanel.isOpen
+        enabled: !connectionInfoScreenVisible && !ipInfoPanel.visible
 
         Accessible.ignored: connectionInfoScreenVisible || ipInfoPanel.isOpen
     }
 
     VPNIPInfoPanel {
+        property var connectionStabilityWatcher: VPNConnectionHealth.stability
+        property var vpnStateWatcher: VPNController.state
```
MattLichtenstein, line 601: Is there any benefit to creating new properties / signal handlers instead of using `Connection` objects? If not, I think the latter is more intuitive

lesleyjanenorton: great point, i agree. updated.

nebula/ui/components/VPNControllerView.qml
```diff
@@ -649,6 +662,7 @@ Item {
                 height: iconSize
                 width: iconSize
             }
+            opacity: parent.enabled ? 1 : .6
```
MattLichtenstein, line 665: definitely not blocking, but I think we should get a tech debt bug written to standardize the opacity of a disabled icon button throughout the codebase (and maybe even standardize the way we add images to `VPNIconButton` - some instances use `Image`, others use `VPNIcon`, etc which opens the door for inconsistency).

For instance, [in VPNMenu](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/70df7ff2d46a652e4b8229f4e78bca8b3740cc84/nebula/ui/components/VPNMenu.qml#L59) we are using `.4` opacity for a disabled button

MattLichtenstein (approved):

nebula/ui/components/VPNControllerView.qml
lesleyjanenorton: +1

PR #5508, opened 2023-01-10
Fix VPN-3565: Don't reorder language list after selecting language

lesleyjanenorton:
## Description

I don't love this. Very open to other approaches.
This prevents the language list from reordering itself back into alphabetical itself when switching between some languages. 

## Reference

VPN-3565

## Checklist
    

bakulf:
Instead of doing this, let's sort the languages by Unicode char comparison. This is what Wikipedia does. No resorting.

Bonus goal is to show RTL languages correctly. See this list https://en.m.wikipedia.org/wiki/Foo_Fighters#/languages

PR #5579, opened 2023-01-11
If in multihop, then only show city names in the CI page

Gela:
## Description

If in multihop, then only show city names in the CI page. This change does not affect the UI in single hop mode.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-3639

## Checklist
    
### DIFF
--- a/nebula/ui/components/VPNConnectionInfoContent.qml
+++ b/nebula/ui/components/VPNConnectionInfoContent.qml
@@ -106,7 +106,7 @@ VPNFlickable {
                         VPNConnectionInfoItem {
                             id: entryServerLabel
                             title: serverLocations.isMultipHop
-                                ? VPNServerCountryModel.getLocalizedCountryName(VPNCurrentServer.entryCountryCode)
+                                ? VPNCurrentServer.localizedEntryCityName
                                 : ""
                             subtitle: ""
                             iconPath: serverLocations.isMultipHop
@@ -120,6 +120,7 @@ VPNFlickable {
 
                         VPNIcon {
                             id: arrowIcon
+                            anchors.horizontalCenter: parent.horizontalCenter
                             source: "qrc:/nebula/resources/arrow-forward-white.svg"
                             sourceSize {
                                 height: VPNTheme.theme.iconSize * 1.25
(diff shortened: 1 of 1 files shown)


nebula/ui/components/VPNConnectionInfoContent.qml
```diff
@@ -120,6 +120,7 @@ VPNFlickable {
 
                         VPNIcon {
                             id: arrowIcon
+                            anchors.horizontalCenter: parent.horizontalCenter
```
MattLichtenstein, line 123: using anchors inside a component that is in a layout will usually cause issues and unintended affects (and of course a qml warning 😂)

Screen Shot 2023-01-11 at 4 28 19 PM

as far as how to resolve this... it sort of depends on what the intended design is. 

- Do we want the arrow to always be dead center, splitting the two labels 50/50?
- Do we want the arrow to move in case we have a longer city on one side than the other?
- If we truncate one city, do we always want to truncate the other (for consistency) or do we want to try to always have at least one of the cities not truncated

@peterbenvenuto any thoughts?

Gela: Good point, I played around with some of the longest names I could find (NYC->NYC) and not seeing the same issue you mentioned. I wonder why in my case it adds "..." and in your it does not? 🤔 

I suppose I'm leaning towards not centering the arrow but let's see what @peterbenvenuto thinks.

Screenshot 2023-01-11 at 1 42 10 PM

nebula/ui/components/VPNConnectionInfoContent.qml
```diff
@@ -132,9 +133,9 @@ VPNFlickable {
                         }
 
                         VPNConnectionInfoItem {
-                            title: VPNServerCountryModel.getLocalizedCountryName(
-                                VPNCurrentServer.exitCountryCode
-                            )
+                            title: serverLocations.isMultipHop
+                            ? VPNCurrentServer.localizedCityName
+                            : VPNServerCountryModel.getLocalizedCountryName(VPNCurrentServer.exitCountryCode);
```
bakulf, line 138: This is the only place where we use: `VPNServerCountryModel.getLocalizedCountryName`.
Can we get rid of it? I would like to see: `VPNCurrentServer.localizedCountryName` instead.
Do you mind to do this? As part of this PR or a separate one.

bakulf (approved):
I reviewed only the use of the model. Please, wait for a review from front-end experts before merging.

nebula/ui/components/VPNConnectionInfoContent.qml
Gela: Jira issue: https://mozilla-hub.atlassian.net/browse/VPN-3884

PR #5610, opened 2023-01-12
Keep a copy of the current server data when activating the VPN

bakulf:
VPN-3734

The issue is described in a comment in controller.cpp, but for the reviewer:

* There is a gap between the request for activation and the actual controller operations because of tasks, JS, add-ons, etc. And Settings can change in the meantime. To avoid mistakes, we take a copy of the current location (MozillaVPN::serverData()) when the TaskControllerAction is created, and we propagate that object to the Controller.
* To improve the readability of the code, I renamed: MozillaVPN::currentServer to MozillaVPN::serverData - because it can be that it's not the current server if operations are still pending
* I also added a Controller::currentServer which is the _real_ ServerData - read-only, for obvious reasons.


bakulf:
I want to add a test. I'll update the PR tomorrow

src/apps/vpn/controller.cpp
```diff
@@ -186,14 +189,20 @@ void Controller::implInitialized(bool status, bool a_connected,
...
-bool Controller::activate() {
+bool Controller::activate(const ServerData& serverData) {
   logger.debug() << "Activation" << m_state;
 
   if (m_state != StateOff && m_state != StateSwitching) {
     logger.debug() << "Already connected";
     return false;
   }
 
+  m_serverData = serverData;
```
birdsarah, line 200: I'm finding it a little confusing about what is m_nextServerData and what is m_serverData.

Further down this file when we look at `switchServers`. I think the following happens:
* we request switchServers
* we set the property m_nextServerData
* we setState to `StateSwitching`
* we run deactivate, this leads to triggering of disconnected
* in disconnected we look to see if we're in StateSwitching and if yes we call activate with m_nextServerData which is what finally makes next the current server data.

1 - i'm finding it hard to reason about this state machine. maybe it's just me. i was looking for documentation like https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/authentication-in-app.md#the-whole-flow - does it exist?
2 - here in activate we set m_serverData but we haven't actually successfully done the activation yet. I'm feeling like I want a Controller::connected() slot that sets m_serverData after a successful activation

Can you help me with my understanding / mental model?

tests/functional/testConnection.js
```diff
@@ -90,4 +88,41 @@ describe('Connectivity', function() {
...
     assert(vpn.lastNotification().message.startsWith('Disconnected from '));
   });
+
+  it.only('Connect to VPN - race condition', async () => {
+    await vpn.setSetting(
+        'serverData',
+        '{"enter_city_name":"","enter_country_code":"","exit_city_name":"Melbourne","exit_country_code":"au"}');
+    await vpn.setSetting('connectionChangeNotification', 'true');
+
+    await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
```
birdsarah, line 98: Am I right in thinking this is trying to turn on the VPN?

This could be done at another time, but I think it would be helpful if this was wrapped in a helper function "activate VPN" (and it's sibling deactivate VPN") that assert the current state as on or off then hit click. This would make this test much easier to read.

birdsarah (commented):
Lots of questions, sorry.

src/apps/vpn/tasks/controlleraction/taskcontrolleraction.h
```diff
@@ -33,14 +34,15 @@ class TaskControllerAction final : public Task {
 
   virtual DeletePolicy deletePolicy() const { return NonDeletable; }
 
- private slots:
+ private:
```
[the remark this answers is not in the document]
bakulf: "Slots" key is not needed when we use function ptr in connect. See the latest connect-doc in qt.

src/apps/vpn/controller.h
```diff
@@ -179,6 +196,15 @@ class Controller final : public QObject {
 
   NextStep m_nextStep = None;
 
+  // There is a gap between the request for activation and the actual
+  // controller operations because of tasks, JS, add-ons, etc. And Settings can
+  // change in the meantime. To avoid mistakes, we take a copy of the current
+  // location when the TaskControllerAction is created, and we propagate that
+  // value to here in these two member variables. No
+  // MozillaVPN::serverData() should be used in the controller!
```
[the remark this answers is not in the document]
bakulf: I added more comments.

src/apps/vpn/controller.cpp
bakulf: Yes, I wrote a better (I hope so at least!) comment to describe the logic. 
To summarize it here:
- m_serverData is a copy of the location, taken when the user (or something else) triggers a VPN activation. This value is kept constant for the whole duration of the operation. If something happens, this and MozillaPVN::serverData are in sync.
- If there is a server-switch request while an activation is in progress, we start a disconnect operation and we store the new location in m_nextServerData.

See this flow:

```mermaid
sequenceDiagram
    User->>Client: Location X is set
    User->>Client: VPN activation request
    activate Client
    Client->>Client: Task scheduled (copy of location is taken: Xc = X)
    deactivate Client
    Client->>Controller: Task run: activate (Xc)
    activate Controller
    Controller->>Controller: m_serverData = Xc
    Controller->>Controller: Connecting...
    User->>Client: Location Y is set
    activate Client
    Client->>Client: Task scheduled (copy of location is taken: Yc = Y)
    deactivate Client
    Controller->>Controller: Conferming...
    Client->>Controller: Task run: serverSwitch (Yc)
    Controller->>Controller: m_nextServerData = Yc
    Controller->>Controller: Connected
    Controller->>Controller: deactivate...
    Controller->>Controller: ativate(m_nextServerData)
    Controller->>Controller: m_serverData = m_nextServerData
    deactivate Controller
```

tests/functional/testConnection.js
bakulf: Yes :) we have vpn.activate() but that goes through the inspector. Here I trigger the client on the toggle, which is cooler!

src/apps/vpn/tasks/controlleraction/taskcontrolleraction.h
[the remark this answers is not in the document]
birdsarah: Thank you!

tests/functional/testConnection.js
birdsarah: I 100% agree that we should be trigger the client with the toggle. I am just saying a helper function e.g.

pseudocode
```
testHelper.toggleOnVpn() {
 toggle = queries.screenHome.CONTROLLER_TOGGLE.visible()
 assert toggle.isOff()
 vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
}
```

So that that way the test code is much easier to read:

```
await vpn.setSetting('connectionChangeNotification', 'true');
await testHelper.toggleOnVpn()
```

src/apps/vpn/controller.cpp
birdsarah: Given that you've made this awesome diagram. Should it go in the docs folder?

src/apps/vpn/externalophandler.cpp
```diff
@@ -63,10 +65,14 @@ bool ExternalOpHandler::request(Op op) {
...
-      vpn->controller()->activate();
+      TaskScheduler::deleteTasks();
+      TaskScheduler::scheduleTask(
+          new TaskControllerAction(TaskControllerAction::eActivate));
       break;
     case OpDeactivate:
-      vpn->controller()->deactivate();
+      TaskScheduler::deleteTasks();
+      TaskScheduler::scheduleTask(
+          new TaskControllerAction(TaskControllerAction::eDeactivate));
```
Gela, line 75: Most of your changes are pretty clear but wondering about this, is this change related to keeping a copy of the server data?

bakulf: this is kind of unrelated. I just found that we are calling deactivate() directly instead of using tasks. This is bad because that could break the order of tasks.

strseb (approved):

PR #5621, opened 2023-01-13
Remove external references to http resources for benchmark tests - VPN-3879

bakulf:
Why: functional tests should run offline. The main blocker is the benchmark test because it uses external resources such as archive.m.o and httpstats.us. With this PR, I remove these dependencies.

### DIFF
--- a/src/apps/vpn/appconstants.h
+++ b/src/apps/vpn/appconstants.h
@@ -96,8 +96,6 @@ constexpr uint32_t BENCHMARK_MAX_DURATION_PING = 3000;
 constexpr uint32_t BENCHMARK_MAX_DURATION_TRANSFER = 15000;
 constexpr uint32_t BENCHMARK_THRESHOLD_SPEED_FAST = 25000000;    // 25 Megabit
 constexpr uint32_t BENCHMARK_THRESHOLD_SPEED_MEDIUM = 10000000;  // 10 Megabit
-constexpr const char* BENCHMARK_DOWNLOAD_URL =
-    "https://archive.mozilla.org/pub/vpn/speedtest/50m.data";
 
 #if defined(UNIT_TEST)
 #  define CONSTEXPR(type, functionName, releaseValue, debugValue, \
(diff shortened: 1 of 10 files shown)


bakulf:
The wasm failure is because we need the network-request cleanup PR.

src/apps/vpn/appconstants.h
```diff
@@ -168,9 +166,18 @@ PRODBETAEXPR(QString, addonBaseUrl,
...
+                 "MZ_BENCHMARK_DOWNLOAD_URL",
+                 "https://archive.mozilla.org/pub/vpn/speedtest/50m.data"));
+
 PRODBETAEXPR(
-    const char*, benchmarkUploadUrl, "https://benchmark.vpn.mozilla.org/upload",
-    "https://dev.vpn-network-benchmark.nonprod.webservices.mozgcp.net/upload");
+    QString, benchmarkUploadUrl, "https://benchmark.vpn.mozilla.org/upload",
+    Constants::envOrDefault(
+        "MZ_BENCHMARK_UPLOAD_URL",
+        "https://dev.vpn-network-benchmark.nonprod.webservices.mozgcp.net/"
```
flozia, line 179: Question (non-blocking): That reminds me that we also have an additional staging environment https://staging.vpn-network-benchmark.nonprod.webservices.mozgcp.net/ — would this be better to use or maybe it does not really matter?

tests/functional/testBenchmark.js
```diff
@@ -10,7 +10,22 @@ describe('Benchmark', function() {
...
+        status: 200,
+        bodyRaw: new Array(1024).join('a'),
+        callback: async req => await this.ctx.downloadUrlCallback(req)
+      },
+    },
+    POSTs: {},
+    DELETEs: {},
+  };
+
   it('Successful benchmark', async () => {
```
flozia, line 25: Praise: They run **so** fast now.

flozia (approved):

PR #5647, opened 2023-01-17
Fix a race condition in the Addon unit test - VPN-3904

bakulf:
If the unit-test runs slow, it can happen that the installation time is < 1 second. We need to reset the installation time before running the time-conditional addons

VPN-3904

### DIFF
--- a/tests/unit/testaddon.cpp
+++ b/tests/unit/testaddon.cpp
@@ -401,6 +401,8 @@ void TestAddon::conditionWatcher_featuresEnabled() {
 }
 
 void TestAddon::conditionWatcher_group() {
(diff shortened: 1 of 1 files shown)


tests/unit/testaddon.cpp
```diff
@@ -865,8 +867,6 @@ void TestAddon::message_notification_data() {
   TestHelper::resetLastSystemNotification();
   // Message is created for the first time,
   // but user is not logged in, no  message sent
-  static_cast<AddonMessage*>(
```
bakulf, line 868: I removed this because it's no-op

oskirby (approved):

PR #5648, opened 2023-01-17
VPN-3614: Tutorial content changes

MattLichtenstein:
## Description

- Change the completion text for the multi hop tutorial on the tutorial completion modal
- Change secondary button text on the tutorial completion modal

## Reference

[VPN-3614: Confusing text in the “Well done” Multi-hop tutorial modal](https://mozilla-hub.atlassian.net/browse/VPN-3614)

## Checklist
    
### DIFF
--- a/addons/tutorial_03_multi_hop/manifest.json
+++ b/addons/tutorial_03_multi_hop/manifest.json
@@ -13,7 +13,7 @@
     "title": "Using multi-hop VPN",
     "settings_rollback_needed": true,
     "subtitle": "Follow this walkthrough to learn how to use multi-hop VPN.",
-    "completion_message": "You’ve turned on multi-hop VPN and changed your exit location. Would you like to see more tips and tricks?",
+    "completion_message": "You’ve learned how to use multi-hop VPN and change your exit location. Would you like to learn more tips and tricks?",
     "steps": [
       {
         "id": "s1",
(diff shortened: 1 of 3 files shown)


addons/tutorial_03_multi_hop/manifest.json
```diff
@@ -13,7 +13,7 @@
     "title": "Using multi-hop VPN",
     "settings_rollback_needed": true,
     "subtitle": "Follow this walkthrough to learn how to use multi-hop VPN.",
-    "completion_message": "You’ve turned on multi-hop VPN and changed your exit location. Would you like to see more tips and tricks?",
+    "completion_message": "You’ve learned how to use multi-hop VPN and change your exit location. Would you like to learn more tips and tricks?",
```
flodolo, line 16: @bakulf 
Can we use a different ID in these manifests?

flodolo (changes requested):
Automation will remove translations if the en-US string changes, but we should still change IDs, because this impacts branches and previous releases.
https://mozilla-l10n.github.io/documentation/localization/making_string_changes.html

addons/tutorial_03_multi_hop/manifest.json
bakulf: oh no! Big problem here... not in the current implementation! Same for title, subtitle and so on. We need to find a solution. Follow up.

MattLichtenstein: In this scenario, since this tutorial isn't public yet, would keeping the id the same be okay?

flodolo: It's not important if it's public, it matters if it's been already exposed to localization (I think it is).

As I said, I build the automation to invalidate existing translations if the source string changes, but we should figure out a way to work around this for the future.

flodolo (approved):

bakulf (approved):

PR #5652, opened 2023-01-17
Revisit the language list

bakulf:
Issue: VPN-3565

With Santiago we decided the following approach:
1. each language has 2 "values": the native language name and the localized language name. Native language name is a fix string. The localized one follows the language settings.
2. the language list is sorted by native language name

Implementation-wise I suggest this:
1. similarly to the country/city localization, the localized language names are fetch from wikidata via the SPARQL endpoint
2. this is done via script (included in this PR)
3. the strings are stored in a languages.json file that is included as a resource in the app
4. LanguageI18N is a component that uses languages.json to translate language codes.

I want to have feedback from @flodolo, mainly.

image
image
image


flodolo:
> the language list is sorted by native language name

I assume the screenshots are from the current build, not this patch then?

flodolo:
https://mozilla-mobile.github.io/mozilla-vpn-client/?branch=languageSorting

It's not ideal, but we should probably capitalize the first letter of the native language name for consistency (when we store it in the JSON file).

Also, it doesn't look like the order changes when I switch language?

bakulf:
The screenshots are from the PR.

bakulf:
> Also, it doesn't look like the order changes when I switch language?

Correct. The language are sorted by native language name. So the order does not change.

flodolo:
> The screenshots are from the PR.

Thanks, somehow my brain got confused :-(

Ordering based on native language means that language names using other scripts will be relegated at the end. Is this acceptable? If it is, I'd still try to capitalize the first letter if the script is Latin, even if not ideal (unless there are functions in ICU that have a stand-alone version of the name? I know CLDR stores some of that information, [see example for Italian](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-misc-full/main/it/contextTransforms.json#L19-L22)).

For example, in Italian language names are lowercase, but if used at the beginning of the sentence (or stand-alone) they still need a capitalized first letter. 

You mentioned Wikipedia, which orders them under the hood by locale code, then displays the localized name. Why was that ruled out? The order is a bit awkward to me, but maybe folks are already used to that, and there is still a search field.

Screenshot 2023-01-19 alle 08 23 04

bakulf:
> For example, in Italian language names are lowercase, but if used at the beginning of the sentence (or stand-alone) they still need a capitalized first letter.

done.

> You mentioned Wikipedia, which orders them under the hood by locale code, then displays the localized name. Why was that ruled out? The order is a bit awkward to me, but maybe folks are already used to that, and there is still a search field.

Because we want to avoid the re-ordering of the list at runtime. It's confusing for users.

flodolo:
> Because we want to avoid the re-ordering of the list at runtime. It's confusing for users.

But wikipedia doesn't re-order languages at run-time, the order is fixed and determined by the locale code.

Check the home page for example for Italian
https://it.wikipedia.org/wiki/Pagina_principale

Bahasa Indonesian (id) is between Armenian (hy) and Japanese (ja). Arabic (ar) is right after Afrikaans (af).

bakulf:
> But Wikipedia doesn't re-order languages at run-time, the order is fixed and determined by the locale code.

Right, but not in the same view. The issue we see in the client is the following (macos/linux):
- open the language list
- pick Japanese: after Japanese, you see Chinese, Chinese Simplified, Greek, Russian, Ukrainian, Punjabi
- pick Ukrainian: the list ends now with Japanese, Chinese Simplified, Chinese.

This is reordering is the confusing part we want to fix.

bakulf:
Btw, because I fixed a bug in the sorting of the server list too, I included `servers.json`, correctly sorted. No changes.

tests/unit/languages/languages.json
```diff
@@ -0,0 +1,12 @@
+[
+ {
+  "wikiDataID": "Q10134",
+  "languageCode": "tlh",
+  "translations": {
+   "tlh": " ",
```
bakulf, line 6: What?!? Firefox does not have klingon simbols?!?

lesleyjanenorton (approved):

flodolo:
Just to confirm, this doesn't have the change we discussed yet (order set at build-time, not run-time)?

bakulf:
> Just to confirm, this doesn't have the change we discussed yet (order set at build-time, not run-time)?

It does:  https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5652/files#diff-3ac13b8fbb5436401fe111093c8b45f7cae6169edfdd0c1c30af119362947343R198

When the app reads all the languages, at startup time, it sorts the language using `LanguageI18N` which uses the `languages.json` as index.

flodolo (approved):

PR #5663, opened 2023-01-18
VPN-3099: Add recommended servers guide

lesleyjanenorton:
## Description

Adds manifest for the recommended servers guide

## Reference

VPN-3099

## Checklist
    

addons/guide_08_recommended_servers/manifest.json
```diff
@@ -0,0 +1,37 @@
+{
+  "api_version": "0.1",
+  "id": "guide_08_recommended_servers",
+  "name": "Guide: recommended servers",
+  "type": "guide",
```
bakulf, line 5: if we are not 100% sure about the strings, let's add `translatable: false`

MattLichtenstein (approved):

MattLichtenstein (changes requested):
The thumbnail/hero image is not correctly scaled

MattLichtenstein (approved):

addons/guide_08_recommended_servers/manifest.json
lesleyjanenorton: Added. I think this is meant to be final content but we don't need this guide immediately so I've added `translatable: false` for now and am following up with Santiago to verify.

addons/guide_08_recommended_servers/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+    "blocks": [
+      {
+          "id": "c_1",
+          "type": "text",
+          "content": "When you use Mozilla VPN, your internet activity is encrypted and routed through servers based in a location of your choice."
+      },
+      {
+          "id": "c_2",
+          "type": "text",
+          "content": "Sometimes, you might pick a location for a specific purpose; for example, if you're traveling internationally, you could choose a server location in your home country so that your banking apps function more smoothly."
```
flodolo, line 24:
```suggestion
          "content": "Sometimes, you might pick a location for a specific purpose; for example, if you’re traveling internationally, you could choose a server location in your home country so that your banking apps function more smoothly."
```

This should have tripped the linter, but it didn't (even before the add-on was marked as no-translate).
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/workflows/translations.yaml?query=branch%3Arec-servers-guide

I need to check if we are looking into add-ons.

Please replace all `'` with `’`.

addons/guide_08_recommended_servers/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+      },
+      {
+          "id": "c_3",
+          "type": "text",
+          "content": "But, for times when you don't have a specific location in mind, the VPN will recommend server locations. These recommendations are based on the number of servers available, as well as their distance from you."
```
flodolo, line 29:
```suggestion
          "content": "But, for times when you don’t have a specific location in mind, the VPN will recommend server locations. These recommendations are based on the number of servers available, as well as their distance from you."
```

addons/guide_08_recommended_servers/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+      },
+      {
+          "id": "c_4",
+          "type": "text",
+          "content": "These recommended locations will refresh occasionally when the VPN is turned off; if you'd like to use VPN with the best performance and speed, choose a recommended server location."
```
flodolo, line 34:
```suggestion
          "content": "These recommended locations will refresh occasionally when the VPN is turned off; if you’d like to use VPN with the best performance and speed, choose a recommended server location."
```

flodolo (changes requested):

flodolo (approved):

PR #5668, opened 2023-01-18
Localhost requests only in functional tests - VPN-3935

bakulf:
VPN-3935

### DIFF
--- a/src/shared/networkmanager.cpp
+++ b/src/shared/networkmanager.cpp
@@ -6,6 +6,8 @@
 
 #include "appconstants.h"
 #include "leakdetector.h"
+#include "networkrequest.h"
+#include "settingsholder.h"
 
 #if MZ_WINDOWS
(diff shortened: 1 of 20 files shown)


strseb:
Should this maybe be a setting instead of a feature flag?

bakulf:
> Should this maybe be a setting instead of a feature flag?

Done.

src/shared/networkmanager.cpp
```diff
@@ -15,13 +17,30 @@
 
 namespace {
 NetworkManager* s_instance = nullptr;
+
+bool localhostRequestCallback(NetworkRequest* request) {
+  QString host(request->url().host());
+  if (host == "localhost" || host == "127.0.0.1" || host == "::1") {
```
flozia, line 23: Praise: I saw the tests failing here before in the CI. Nice workaround with the captive portal! :)

flozia (approved):

PR #5698, opened 2023-01-19
Re-enable caches for mac/iOS builds

bhearsum:
While looking through the taskcluster code I happened to notice that caches were disabled for these builds. According to https://github.com/mozilla-mobile/mozilla-vpn-client/commit/5fe653860a0562651c7a60cb43cd3a5583dd4e40, the Mac one was disabled due to https://github.com/taskcluster/taskcluster/issues/5396, which I believe is now fixed and deployed? I assume it's off for iOS for the same reason, although those builds were added after the issue first came up so I'm not certain.

### DIFF
--- a/taskcluster/ci/build/ios.yml
+++ b/taskcluster/ci/build/ios.yml
@@ -21,6 +21,6 @@ ios/debug:
         - MozillaVPN.tar.gz
     run:
         using: run-task
-        use-caches: false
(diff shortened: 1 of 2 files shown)


bhearsum:
Nevermind, the Taskcluster fail shows this is not fixed (on our workers, at least) :laughing:

PR #5718, opened 2023-01-20
Take a copy of the authorization header in the CTOR of taskremovedevice in case we are removing the current device - VPN-3976

bakulf:
VPN-3976

### DIFF
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -889,17 +889,12 @@ void MozillaVPN::logout() {
   }
 
   if (m_private->m_deviceModel.hasCurrentDevice(keys())) {
-    TaskScheduler::scheduleTask(new TaskGroup(
-        {new TaskRemoveDevice(keys()->publicKey()),
-         // Immediately after the scheduling of the device removal, we want to
-         // delete the session token, so that, in case the app is terminated, at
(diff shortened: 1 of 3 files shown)


src/apps/vpn/mozillavpn.cpp
```diff
@@ -889,17 +889,12 @@ void MozillaVPN::logout() {
...
-    // In case the app is closed even before scheduling the previous TaskGroup,
-    // removing the key we will enforce a new authentication at the first
-    // TaskAccount execution.
-    m_private->m_keys.forgetKeys();
+    TaskScheduler::scheduleTask(new TaskRemoveDevice(keys()->publicKey()));
+
+    // Immediately after the scheduling of the device removal, we want to
+    // delete the session token, so that, in case the app is terminated, at
+    // the next execution we go back to the init screen.
+    reset(false);
```
brizental, line 897: Are you changing this just because reset is now possible because the task copied the auth header and there is not need to run them in sequence?

brizental (approved):

src/apps/vpn/mozillavpn.cpp
bakulf: Correct! Before this was not possible because of how NetworkRequest was written: the authorization header was computer in the CTOR. But now, the caller has full control of the headers and the authorization one can be set with a cached string value.

Issue #4966, opened 2022-11-18
Signal strength indicator not displayed for all locations


data-sync-user:
**Affected versions:**

* 2.12.0 (2.202211171941)

**Tested Platforms:**

* MacOS 12.5.1;

**Prerequisites:**

* Have a FxA account with an active subscription;
* Have VPN app open and be signed in;

**Steps to reproduce:**

1. Open server selection list;
1. Expand several countries;
1. Observe server location (USA for optimal number of locations);

**Expected result:**

* Signal strength indicator is displayed for all server locations;

**Actual result:**

* Signal strength indicator not displayed for all locations;

**Notes:**

* Some locations take a bit longer for the signal strength indicator to appear but for some it doesn’t even after several minutes;
* See attached video;

!Screen Recording 2022-11-18 at 16.49.08.mov|width=50%!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-3297)
┆Reporter: Robert Vass

data-sync-user:
➤ Sarah Bird commented:

Lesley Norton or Matt Lichtenstein do you have a S/M/L opinion on this?

data-sync-user:
➤ Owen Kirby commented:

I think the issue here is that we are opening the server list before the measurement has been completed. To resolve this I think we either need a mechanism to refresh the view once it’s done, or connect a signal to propagate changes to the indicator.

data-sync-user:
➤ Matt Lichtenstein commented:

Sarah Bird basically what Owen said, with preference towards the latter. Because I don’t know the solution off the top of my head, I would say an M

data-sync-user:
➤ Owen Kirby commented:

In https://mozilla-hub.atlassian.net/browse/VPN-2529 ( https://mozilla-hub.atlassian.net/browse/VPN-2529|smart-link ) we converted the connection score into a property, so the icons will now redraw themselves as data is gathered. The missing indicators may still occur but they should disappear quickly.

data-sync-user:
➤ Valentina Virlics commented:

Could we add a fixVersion for this ticket? Thank you! Owen Kirby Betty Fleming

data-sync-user:
➤ Valentina Virlics commented:

Naomi Kirby Thank you for adding the fixVersion!

Matt Lichtenstein 

[...]

Issue #3988, opened 2022-07-09
Add helper method to fetch server recommendation shortlist


data-sync-user:
It is probably asking too much of the QML to fetch the entire server list and then distill it down to a set of recommendations. We may want to implement this in C++ and return a shortlist. A naive implementation would simply sort by latency, and return the top few entries, but this is likely to result in strong grouping by country. We may want to pick the top entries while limiting them to one per country.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2529)
┆Reporter: Owen Kirby

data-sync-user:
➤ Betty Fleming commented:

Owen Kirby this one doesn’t have anyone assigned…shoudl it be you?  I”ll remove from 2.12,

data-sync-user:
➤ Valentina Virlics commented:

Santiago Andrigo Is this qa-actionable? Thank you!

data-sync-user:
➤ Santiago Andrigo commented:

I don’t think so. Adding qa-not-actionable. Owen Kirby please correct it if I’m wrong.

data-sync-user:
➤ Owen Kirby commented:

Santiago Andrigo we should defer QA until the UI is completed: https://mozilla-hub.atlassian.net/browse/VPN-3023 ( https://mozilla-hub.atlassian.net/browse/VPN-3023|smart-link )

PR #5724, opened 2023-01-21
VPN-2529: Fetch recommended servers

oskirby:
## Description
This implements a method to return a list of recommended servers, `ServerCountryModel::recommendedServers()` which returns a list of `ServerCity` objects. Along the way we also tackled some cleanup:
- `cityConnectionScore()` has been converted to a property of the `ServerCity` object, allowing QML to update itself as scores change.
- Move the great-circle distance algorithm into the `Location` model.
- Trigger server latency updates on changes to the underlying features.
- Sort the order of server latency fetches by geographic distance, to try and get the best-performing results first.
- Make `ServerCountryModel::pickBest()` a wrapper for `ServerCountryModel::recommendedServers(1)`

## Reference
Github #3988 ([VPN-2529](https://mozilla-hub.atlassian.net/browse/VPN-2529))
Github #4966 ([VPN-3297](https://mozilla-hub.atlassian.net/browse/VPN-3297))

## Checklist
    
[VPN-2529]: https://mozilla-hub.atlassian.net/browse/VPN-2529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPN-3297]: https://mozilla-hub.atlassian.net/browse/VPN-3297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/vpn/models/servercountrymodel.cpp
+++ b/src/apps/vpn/models/servercountrymodel.cpp
@@ -177,19 +177,6 @@ QVariant ServerCountryModel::data(const QModelIndex& index, int role) const {
       return QVariant();
   }
 }
-
-int ServerCountryModel::cityConnectionScore(const QString& countryCode,
-                                            const QString& cityCode) const {
-  for (const ServerCity& city : cities(countryCode)) {
-    if (city.code() == cityCode) {
-      return city.connectionScore();
-    }
-  }
-
-  // No such country was found.
-  return NoData;
-}
-
 QStringList ServerCountryModel::pickRandom() const {
   logger.debug() << "Choosing a random server";
   qsizetype index = QRandomGenerator::global()->generate() % m_servers.count();
--- a/src/apps/vpn/models/servercountrymodel.cpp
+++ b/src/apps/vpn/models/servercountrymodel.cpp
@@ -207,8 +207,8 @@ bool ServerCountryModel::exists(const QString& countryCode,
   return m_cities.contains(ServerCity::hashKey(countryCode, cityName));
 }
 
-const ServerCity& ServerCountryModel::findCity(
-    const QString& countryCode, const QString& cityName) const {
+const ServerCity& ServerCountryModel::findCity(const QString& countryCode,
+                                               const QString& cityName) const {
   auto index = m_cities.constFind(ServerCity::hashKey(countryCode, cityName));
   if (index == m_cities.end()) {
     static const ServerCity emptycity;
--- a/src/apps/vpn/models/servercountrymodel.cpp
+++ b/src/apps/vpn/models/servercountrymodel.cpp
@@ -218,23 +218,6 @@ const ServerCity& ServerCountryModel::findCity(
   return *index;
 }
 
-const QList<Server> ServerCountryModel::servers(const QString& countryCode,
-                                                const QString& cityName) const {
-  QList<Server> results;
-  const ServerCity& city = findCity(countryCode, cityName);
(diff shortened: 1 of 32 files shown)


nebula/mocks/Mozilla/VPN/MZFilterProxyModel.qml
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+
+Item {
+  property var filterCallback: () => {}
```
bakulf, line 8: This should not required. Why do we need to lock this object? Is it for qmllint?

src/apps/vpn/models/servercountrymodel.cpp
```diff
@@ -364,10 +302,54 @@ void ServerCountryModel::retranslate() {
   endResetModel();
 }
 
+unsigned int ServerCountryModel::avgLatency() const {
+  if (m_numLatencySamples == 0) {
+    return 0;
+  } else {
```
bakulf, line 308: No else here

src/apps/vpn/models/servercountrymodel.cpp
```diff
@@ -364,10 +302,54 @@ void ServerCountryModel::retranslate() {
...
+  if (server.latency() != 0) {
+    m_sumLatencyMsec -= server.latency();
+  } else {
+    m_numLatencySamples++;
+  }
+  m_sumLatencyMsec += msec;
+  server.setLatency(msec);
+
+  // Ugly iteration. Find the city for this server.
+  // TODO: Maybe it would be cleaner to move the latency data into ServerCity.
```
bakulf, line 329: Or move this servers into the cities.

bakulf (approved):

nebula/mocks/Mozilla/VPN/MZFilterProxyModel.qml
oskirby: Yes, this was an attempt to clean up some qmllint warnings. It was complaining that it couldn't determine the types of properties in an `MZFilterProxyModel`, so I simply attempted to add a mocked version. Unfortunately, there are a bunch of other QML linter warnings that are going to much harder to resolve. I'm debating if I should just back out of the linter work and deal with it in another PR.

src/apps/vpn/models/servercountrymodel.cpp
oskirby: The problem is that we also have a need to frequently look up servers by their public key, so moving the `Servers` back into the `ServerCity` model  would just cause some ugly iteration elsewhere. What we almost need is something like a database where both servers and cities can be searched via multiple criteria.

nebula/mocks/Mozilla/VPN/MZFilterProxyModel.qml
oskirby: @strseb  is at work on a much better solution with QML modules. So I am going to revert this.

lesleyjanenorton (approved):

PR #5726, opened 2023-01-21
Add -Wall -Werror to CMake

bakulf:


taskcluster/scripts/fetch/generate_vsstudio_bundle.ps1
```diff
@@ -35,7 +35,7 @@ if(!(Test-Path "C:\VisualStudio\VC\Auxiliary\Build\vcvars64.bat")){
         --add Microsoft.VisualStudio.Component.VC.Redist.MSM `
         --add Microsoft.VisualStudio.Component.VC.CoreBuildTools `
         --add Microsoft.VisualStudio.Component.VC.CoreIde `
-        --add Microsoft.VisualStudio.Component.Windows10SDK.19041 `
+        --add Microsoft.VisualStudio.Component.Windows10SDK.20348 `
```
strseb, line 38: This would require the vsstudio bundle to be updated aswell. it's currently not generated in ci.

taskcluster/scripts/fetch/enter_dev_shell.ps1
```diff
@@ -8,7 +8,7 @@
 
 $VS_STUDIO_LOCATION =resolve-path $PSScriptRoot
 $W10_SDK_PATH =  resolve-path "$PSScriptRoot\SDK"
-$W10_SDK_VERSION = "10.0.19041.0"
+$W10_SDK_VERSION = "10.0.20348.0"
```
strseb, line 11: Why does the SDK version need to change if 
```if(NOT MSVC)```
msvc is not touched?

File: src/apps/vpn/mozillavpn.cpp [hunk 3]
```diff
@@ -1585,16 +1585,7 @@ void MozillaVPN::exitForUnrecoverableError(const QString& reason) {
...
-  unsigned char* test = NULL;
-  test[1000] = 'a';  //<< here it should crash
-
-  // Interestingly this does not cause a "Signal" but a VC runtime exception
-  // and more interestingly, neither breakpad nor crashpad are catchting this on
-  // windows...
-  char* text = new char[100];
-  delete[] text;
-  delete[] text;
+  qFatal("Ready to crash!");
```
strseb, line 1588: Given we use this to test the crashreporter, i'd actually like to keep this a similar crash we'd want to detect. 
Does crashpad catch qFatals?

strseb (changes requested):

taskcluster/scripts/fetch/enter_dev_shell.ps1
bakulf: The version we use of win 10 SDK generates c++ warnings. We should use this version to enable WError on windows.
I would like to know the process for updating the SDK.

strseb: I see :) - 
This script generates the bundle, however given it requires root we can't run it in CI. 
Therefore we get the bundle in a fetch task. The next version we generate should live on tooltool - preferably :) 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/51876bf0021a9df290d984e48b16cd4fcc36143a/taskcluster/ci/fetch/kind.yml#L16-L25

src/apps/vpn/mozillavpn.cpp
bakulf: qFatail generates a core-dump, so yes, it should!

strseb (approved):

PR #5727, opened 2023-01-21
silent switch for settings changes - VPN-3601

bakulf:
Tech Doc:

- We want to use silent-switch functionality to apply setting changes when the VPN is on. Silent-switch is something we already have implemented for unstable servers.
- We need a `SettingsWatcher` obj that observers settings change. When the VPN is on, and there is a setting change, this component triggers a server-switch appending a TaskControllerAction task to the queue.
- We also need to remove a few glean events and the VPN-is-off warning alerts
- The best way to make the silent-switch code more readable, I introduce a StateSilentSwitching. In this way we can reduce the `stateToReason` calls.

### DIFF
--- a/src/apps/vpn/ui/screens/settings/ViewNotifications.qml
+++ b/src/apps/vpn/ui/screens/settings/ViewNotifications.qml
@@ -12,8 +12,6 @@ import components.forms 0.1
 
 
 VPNViewBase {
-    property bool vpnIsOff: (VPNController.state === VPNController.StateOff)
-
     id: vpnFlickable
     objectName: "settingsNotifications"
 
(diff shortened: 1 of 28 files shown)


src/apps/vpn/ui/screens/settings/ViewNotifications.qml
```diff
@@ -53,12 +33,10 @@ VPNViewBase {
             //% "Get notified if a guest Wi-Fi portal is blocked due to VPN connection"
             subLabelText: qsTrId("vpn.settings.guestWifiAlert.description")
             isChecked: (VPNSettings.captivePortalAlert)
-            enabled: vpnFlickable.vpnIsOff
+            enabled: true
```
lesleyjanenorton, line 36: We can just remove this line and the other instances of `enabled: true` since these are enabled by default.

lesleyjanenorton (approved):
Love it. We should have done this a hundred years ago. QML looks fine and runs without incident locally. A few thoughts/questions:
- A nice to have would be to disable the `vpn connected` os notification
- If we don't manage the switch successfully, what then? Maybe redirect to home?

bakulf:
> * A nice to have would be to disable the `vpn connected` os notification

Done. I had to introduce `StateSilentSwitching`. Please check if it works again with all the QML changes.

> * If we don't manage the switch successfully, what then? Maybe redirect to home?

I think this is an orthogonal problem: currently we do nothing if we app goes in no-signal state, or if the controller crashes, etc.: we don't change the screen and we don't show dialogs. Why this should be different? Note that we run silent switching in background often and nobody has complained so far.

But it would be nice to add a telemetry event to see how often a silent and non server switching fails.

strseb (approved):

PR #5745, opened 2023-01-23
Implement guide custom dns - VPN-3271

bakulf:


addons/guide_10_app_exclusions/manifest.json
```diff
@@ -0,0 +1,38 @@
+{
+    "api_version": "0.1",
+    "id": "guide_10_app_exclusions",
+    "name": "Guide: app exclusions",
```
MattLichtenstein, line 4: Would it make sense to somehow nest this into the same folder as app permissions, since it's essentially the same guide for a different version of the app?

bakulf: 2 manifests in the same folder? This is not supported.

MattLichtenstein: yea essentially. I was thinking more along the lines of

guide_07_app_permissions
↳ V2.13
  ↳ manifest.json
↳ V2.14
  ↳ manifest.json

bakulf: nice but we need to change the addon script to make it happen. Can we do it in a follow up?

MattLichtenstein (approved):

addons/guide_10_app_exclusions/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+    "guide": {
+        "id": "app_exclusions",
+        "title": "Why you should use app exclusions",
+        "advanced": true,
+        "image": "qrc:/addons/guide_10_app_exclusions/image.svg",
+        "blocks": [
+            {
+                "id": "c_2",
+                "type": "text",
+                "content": "By default, when you activate Mozilla VPN on your device, all the apps you have in it will have their traffic be automatically encrypted and routed by the VPN."
```
flodolo, line 19:
```suggestion
                "content": "By default, when you activate Mozilla VPN on your device, all the apps you have in it will have their traffic automatically encrypted and routed by the VPN."
```

@chrisrooneys 
That "be" feels unnecessary?

PR #5757, opened 2023-01-24
VPN-4001 - Update glean_parser_ext to be compatible with glean_parser 7.0.0

brizental:
With[ the breaking change on glean_parser](https://github.com/mozilla/glean_parser/releases/tag/v7.0.0) not to support untyped event extras, the `has_extra_types` property was removed.


bakulf (approved):

oskirby:
Does this require us to update our glean parser version as well? If so we should ensure the minimum version is tracked in `requirements.txt`

brizental:
> Does this require us to update our glean parser version as well? If so we should ensure the minimum version is tracked in requirements.txt

Not really. This was actually an unnecessary part of our code anyways because we don't have untyped extras, removing that check doesn't really change anything, but it does faul with the new glean_parser. Taskcluster seems to always use latest, that is why this was breaking CI.

Issue #5240, opened 2022-12-15
Clean-up git submodule usage


data-sync-user:
Proposal was discussed and agreed upon at Eng Meeting - Dec 13th - [https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/27034009/VPN+Engineering+Meeting+Notes](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/27034009/VPN+Engineering+Meeting+Notes|smart-link)

* We remove the unevenly applied “ignore=all”
  * This will have the effect of showing you better when using git if you have uncommitted changes in submodules - which you shouldn’t if you’re only running `git submodule update --init`
* In scripts throughout the codebase we only have `git submodule update --init`
  * Currently some scripts run `git submodule update --remote` which has the effect of pulling the latest submodule changes from the remote repo.
  * Everything and everyone should be building using the version of the submodule that are checked into the codebase
* We decide which, if any, submodules we do _not_ want kept in sync with the upstream repo (and we document it):
  * i18n
    * main
    * tbd on making this frictionless
  * 3rdparty/wireguard-tools
    * tag+dependabot
  * 3rdparty/wireguard-apple
    * tag+dependabot
  * 3rdparty/openSSL
    * tag+dependabot
  * 3rdparty/wireguard-go
    * tag+dependabot
  * 3rdparty/adjust-ios-sdk
    * tag+dependbot
  * 3rdparty/adjust-android-sdk
    * tag+dependabot
  * 3rdparty/glean
    * tag+dependabot
* We make a process for keeping all desired sub-modules up-to-date:
  * Option 1: We use dependabot to open submodule PRs.
    * This means that we stay up-to-date with our vendored-in code, without manual intervention, as soon as possible, but it could come at a bad time close to a fix freeze date.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-3507)
┆Reporter: Sarah Bird

data-sync-user:
➤ Betty Fleming commented:

Sarah Bird needs story points

PR #5784, opened 2023-01-26
VPN-4018 - Do not get latest submodule version on Windows CI

brizental:
We should remove this `--remote` everywhere else it is used.

### DIFF
--- a/taskcluster/scripts/build/windows.ps1
+++ b/taskcluster/scripts/build/windows.ps1
@@ -22,6 +22,7 @@ Remove-Item $FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/HostX64/x64
 python3 -m pip install -r requirements.txt --user
 python3 -m pip install -r taskcluster/scripts/requirements.txt --user
 git submodule update --init --force --recursive --depth=1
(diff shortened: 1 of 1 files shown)


strseb:
Curious if this works. TC also checks out submodules, hopefully it does not do remote 🤞

strseb (approved):

oskirby:
I think we might want to add `--remote` back for the `i18n` submodule. We are kind of relying on that to ensure we have updated strings in the absence of any kind of process for keeping the submodule current.

brizental:
> I think we might want to add --remote back for the i18n submodule. We are kind of relying on that to ensure we have updated strings in the absence of any kind of process for keeping the submodule current.

Good to know. Is there already a ticket for creating a process for this though? Happy to file one otherwise.

oskirby:
> Done and good to know. Is there already a ticket for creating a process for this though? Happy to file one otherwise.

I think the last time we got bitten by a submodule issue, we filed #5240 ([VPN-3507](https://mozilla-hub.atlassian.net/browse/VPN-3507)) to capture our thoughts on the matter.

oskirby (approved):

PR #5786, opened 2023-01-27
VPN-4021 - Update Glean to v52.2.0

brizental:
The Glean team has already fixed yesterday's ticket, so let's just update.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1812672


brizental:
Converting to draft, because the new update has some changes to the uniffi-bindgen script and we can't give it the `--config` flag anymore. Not sure whether we should just send a PR to the Glean repo or not use the script anymore. The latter would mean we would need to find a way to keep our uniffi version in sync with Glean's... Anyways. I'll come back to this another time.

oskirby (approved):

PR #5804, opened 2023-01-30
Chore: Upload Source Code with Sentry Symbols

strseb:
## Description

We can ask sentry to attach source-files to our symbols, allowing sentry to show where the error happens. Could be cool. Let's add that.


mozrokafor (approved):
LGTM

PR #5807, opened 2023-01-30
Chore: Flip Sentry On

strseb:
Everything Required from a Code-Perspective Landed, let's enable sentry on supported plattforms


brizental (approved):
YEAHHH WOOHOOO

PR #5820, opened 2023-01-31
VPN-Glean: Make Python a valid candidate for PYTHON_EXECUTABLE

strseb:
## Description
Currently only python3 is checked. In a conda env it seems we have python3 as the default `python` command.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.17)
     cmake_policy(SET CMP0099 OLD)
 endif()
 
-find_program(PYTHON_EXECUTABLE NAMES python3)
+find_program(PYTHON_EXECUTABLE NAMES python3 python)
 if(MSVC)
     include(src/apps/vpn/cmake/msvc.cmake)
 endif()


brizental (approved):

CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.17)
     cmake_policy(SET CMP0099 OLD)
 endif()
 
-find_program(PYTHON_EXECUTABLE NAMES python3)
+find_program(PYTHON_EXECUTABLE NAMES python3 python)
```
brizental, line 87: We should set this variable for all of cmake, not only vpnglean. Fine to do it as a follow up though.

oskirby: @brizental IIRC, that's exactly what we do - this is the top level `CMakeLists.txt` and it gets set for the entire project.

oskirby (approved):

CMakeLists.txt
brizental: LOL, duh! I was confused because the title of the PR says vpnglean.... Sorry!

PR #5840, opened 2023-02-01
WIP: Conda env for windows devs

strseb:
## Description
Same as we did for macos, but windows. 
The status is currently: "works for me", we need to update Tc and check in what to do about the msvc redist. 

# Todo: 
- [ ] Update taskcluster to use this toolchain.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,10 +11,6 @@
 	path = 3rdparty/wireguard-apple
 	url = https://github.com/mozilla/wireguard-apple
         ignore=all
-[submodule "3rdparty/openSSL"]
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -11,10 +11,6 @@
 	path = 3rdparty/wireguard-apple
 	url = https://github.com/mozilla/wireguard-apple
         ignore=all
-[submodule "3rdparty/openSSL"]
```
strseb, line 14: ooops: unrelated change. But it's actually valid to remove.

birdsarah:
Again will pick this up in the doc.

PR #5844, opened 2023-02-01
ci: use the proper beetmover destination on 'ship-client' phase

ahal:
Previously this was pointing at the candidates dir, even when running the ship-client phase.

## Description

    This change fixes the wrong build dirs being linked in the `release-notify` message that gets posted to Slack.

## Checklist
    
- [ x ] My code follows the style guidelines for this project
- [ x ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ x ] I have performed a self review of my own code
- [ x ] I have commented my code PARTICULARLY in hard to understand areas
- [ x ] I have added thorough tests where needed

### DIFF
--- a/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
@@ -66,7 +66,7 @@ def add_beetmover_worker_config(config, tasks):
                     build_id,
                 )
             )
-        else:
+        elif shipping_phase == "promote-client":
             destination_paths.append(
                 os.path.join(
                     "pub",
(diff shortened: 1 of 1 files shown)


gabrielBusta (approved):

gabrielBusta (changes requested):
The code is assuming a destination path here https://github.com/mozilla-mobile/mozilla-vpn-client/blob/d72b1c3f86f27302d25461df0f34701398c2fbed/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py#L148-L149

ahal:
Thanks! I think I'd prefer to be explicit, so I added a better exception if we fall off the end of the if statement.

birdsarah:

gabrielBusta (approved):

PR #5857, opened 2023-02-02
VPN-4063 - Add warning to settings views on iOS

lesleyjanenorton:
## Description

This PR adds `InformationCard.qml` and uses it to: 
1.) surface an advisory note on iOS, when the VPN is not off, that changing some settings will cause the VPN to briefly disconnect and reconnect
2.) Advise users that changing settings in `ViewPrivacy.qml` may impact some website functionality.

<table>
<tr>
<th colspan = "3" align="left">When VPN is ON</th>
</tr>
<tr>
<td>
Screen Shot 2023-02-02 at 10 16 27 AM

</td>
<td>
Screen Shot 2023-02-02 at 10 10 10 AM
</td>
</tr>
<tr>
<th colspan = "3" align="left">When VPN is OFF</th>
</tr>
<td>

Screen Shot 2023-02-02 at 6 38 01 PM

</td>
<td>
Screen Shot 2023-02-02 at 6 37 51 PM
</td>
</tr>

</table>
## Reference

VPN-4063

## Checklist
    

flodolo (approved):

src/apps/vpn/ui/screens/settings/InformationCard.qml
```diff
@@ -0,0 +1,59 @@
...
+import QtQuick.Window 2.12
+
+import Mozilla.VPN 1.0
+import compat 0.1
+import components 0.1
+
+import org.mozilla.Glean 0.30
+import telemetry 0.30
+
+Item {
```
MattLichtenstein, line 16: Are we able to remove this superfluous item and maybe just make the `Rectangle` the root?

src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
```diff
@@ -35,6 +35,25 @@ VPNViewBase {
         Layout.rightMargin: VPNTheme.theme.windowMargin
         Layout.topMargin: VPNTheme.theme.windowMargin / 2
 
+        InformationCard {
```
MattLichtenstein, line 38: Can we stick this in a loader (active on same conditions as visible, and remove visible)

src/apps/vpn/ui/screens/settings/ViewNotifications.qml
```diff
@@ -18,15 +18,33 @@ VPNViewBase {
     //% "Notifications"
     _menuTitle: qsTrId("vpn.settings.notifications")
 
-    _viewContentData: Column {
+    _viewContentData: ColumnLayout {
         Layout.preferredWidth: parent.width - VPNTheme.theme.windowMargin
         Layout.rightMargin: VPNTheme.theme.windowMargin
         spacing: VPNTheme.theme.windowMargin
 
+        InformationCard {
```
MattLichtenstein, line 26: Loader

src/apps/vpn/ui/screens/settings/ViewPrivacy.qml
```diff
@@ -27,53 +27,29 @@ VPNViewBase {
...
-                }
+            Layout.minimumHeight: textBlocks.implicitHeight
+            _infoContent: ColumnLayout {
+                id: textBlocks
                 VPNTextBlock {
                     Layout.fillWidth: true
                     width: undefined
                     text: VPNl18n.SettingsDnsSettingsWarning
                 }
+                VPNTextBlock {
```
MattLichtenstein, line 44: Loader

MattLichtenstein (changes requested):

src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
lesleyjanenorton: yep. better!

src/apps/vpn/ui/screens/settings/InformationCard.qml
lesleyjanenorton: We'd still need a seemingly superfluous Rectangle to serve as `source` for the drop shadow [though it's unclear why just setting this to parent is unsupported](https://doc.qt.io/qt-5/qml-qtgraphicaleffects-dropshadow.html#source-prop).

src/apps/vpn/ui/screens/settings/ViewPrivacy.qml
```diff
@@ -23,57 +23,35 @@ VPNViewBase {
...
-                    Layout.fillWidth: true
-                    width: undefined
-                    text: VPNl18n.SettingsDnsSettingsWarning
+        spacing: VPNTheme.theme.windowMargin * 1.5
+
+        Loader {
+            objectName: "privacySettingsViewInformationCardLoader"
+            Layout.alignment: Qt.AlignHCenter
+
+            sourceComponent: InformationCard {
```
MattLichtenstein, line 32: I don't think the whole card needs to be in a loader, since it's always going to be loaded - just the iOS specific text block

MattLichtenstein (changes requested):

src/apps/vpn/ui/ui.qrc
```diff
@@ -70,6 +70,7 @@
         <file>screens/ScreenViewLogs.qml</file>
         <file>screens/settings/appPermissions/AppPermissionsList.qml</file>
         <file>screens/settings/appPermissions/ViewAppPermissions.qml</file>
+        <file>screens/settings/InformationCard.qml</file>
```
MattLichtenstein, line 73: [non-blocking] What do you think about moving the InformationCard component to nebula? Seems like a re-usable component?

MattLichtenstein (approved):

src/apps/vpn/ui/ui.qrc
lesleyjanenorton: I also thought about his. I left it here since it's only being used in settings files but when/if we need to use it elsewhere, we should def bump it to nebula.

src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
```diff
@@ -29,18 +29,37 @@ VPNViewBase {
...
-        spacing: VPNTheme.theme.windowMargin * 2
-        Layout.fillWidth: true
+        spacing: VPNTheme.theme.windowMargin * 1.5
         Layout.leftMargin: VPNTheme.theme.windowMargin
         Layout.rightMargin: VPNTheme.theme.windowMargin
-        Layout.topMargin: VPNTheme.theme.windowMargin / 2
+
+        Loader {
+            objectName: "DNSSettingsViewInformationCardLoader"
+            active: Qt.platform.os === "ios" && VPNController.state !== VPNController.StateOff
```
bakulf, line 38: The best would be to expose the silent switch without disconnect property from the controller.

Basically: VPNController.disconnectNeededForSwitching (or a better name).

The impl should be done in controller.{cpp,h} and from there to the ControllerImpl via an override const method.

bakulf (approved):

bakulf (changes requested):
See my comment for the implementation

mcleinman (changes requested):
A couple small notes, and a few overall things:

1. Do we expect the lack of translation at this point in the process?
2. When the VPN is off, the box seems to have some extra padding.

IMG_0017
IMG_0016

Also, when rapidly toggling the DNS privacy options, I managed to get a crash. I think it's some sort of race condition, but I wasn't able to replicate this, so I'm not too worried - unless others have also experienced this.

translations/strings.yaml
```diff
@@ -314,6 +314,7 @@ settings:
   dnsSettingsCustomDNSError: Invalid DNS, please try a new one
   dnsSettingsWarning: These options may cause some websites to break or not display content correctly.
   dnsSettingsInputPlaceholder: Enter custom DNS
+  iOSDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
```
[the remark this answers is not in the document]
lesleyjanenorton: good point

src/apps/vpn/controller.h
```diff
@@ -109,6 +111,7 @@ class Controller final : public QObject {
   void captivePortalGone();
 
   const ServerData& currentServer() const { return m_serverData; }
+  bool silentServerSwitchingSupported() const;
```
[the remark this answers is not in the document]
lesleyjanenorton: Good question - I don't think this has ever been clearly defined and looking at `appfeaturelist.h` i'm confused by some of the overlap. For instance, why the `FEATURE(multiHop,` as well as the `multiHopSupported()`. 

My possibly incorrect understanding of `appfeaturelist.h` is that it's for:
- Building features behind feature flags - like `recommendedServers` currently
- Gating feature visibility in cases where we only want to release a feature on platform x, but not y. We delayed releasing in-app authentication on desktop for a long time, even though it was mostly functional and already released on mobile.
- Features we want to ship now but dynamically enable later via Guardian

lesleyjanenorton:
> 1. Do we expect the lack of translation at this point in the process?

Yes, though this is probably a build issue. Running `scripts/utils/import_language.py` and `scripts/utils/generate_strings.py` should fix.


> 2. When the VPN is off, the box seems to have some extra padding.

I'm glad it's not just me. Will tweak.

lesleyjanenorton:
Updated screenshots post-spacing fix. cc @mcleinman 
<table>
<tr>
<td>

Screen Shot 2023-02-06 at 7 07 49 PM 
</td>
<td>

Screen Shot 2023-02-06 at 7 07 42 PM
</td>

translations/strings.yaml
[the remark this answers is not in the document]
lesleyjanenorton: updated

bakulf (approved):

mcleinman (approved):
Thanks for those updates, looks great! Thank you!

PR #5875, opened 2023-02-06
Fix cmake-race with sentry

strseb:
## Description
Problem fixed in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5502 also applies for macos, let's fix it once and for all :3 
## Reference

>It seems the build of windows has gotten flaky since we started to build sentry.
The error seems to be that go-build tries to attach the state of the git repo into the binary.
My current guess it that we have a race between the git-submodule beeing checked out and the
Sentry-External Project messing with the files, therfore letting it fail only sometimes.
So i'd suggest we apply the suggestion of Use -buildvcs=false to disable VCS stamping. we're not using that, right?
Error Task: https://firefox-ci-tc.services.mozilla.com/tasks/A6ybTyseTgSR4vdm4xSOwg/runs/0/logs/public/logs/live.log


brizental (approved):

bakulf (approved):

oskirby (changes requested):
IIRC, the `-buildvcs` flag is only supported on go 1.18 and newer, and this breaks our Ubuntu Focal and Bionic builds which are stuck with go 1.16.

oskirby:
Actually, it seems that `golang-1.18` packages exist for both [focal-updates](https://packages.ubuntu.com/focal-updates/golang-1.18) and [bionic-updates](https://packages.ubuntu.com/bionic-updates/golang-1.18) so we could probably figure out a way to get Linux building again with go 1.18

oskirby (approved):

PR #5877, opened 2023-02-06
Update wg-go on osx

strseb:
## Description

We've seen failures of wg-go to build on OSX.
This was caused due to us updating to golang 1.18, this was however not noticed as the runner never "cleaned" properly after a run, and since we build in the `/3rdparty/wireguard-go` folder not the cmake_bin_dir, i never noticed too. 
Now we could rollback go... or upgrade :)


oskirby:
I think this is good to get the builds unstuck, but we might want to consider it as tech debt to revisit how `wireguard-go` gets built for MacOS in order to make it build in the binary directory instead.

PR #5879, opened 2023-02-06
Fix the string generation

bakulf:


flodolo (approved):
This one seems to work correctly both locally and in the action

PR #5880, opened 2023-02-06
Do not randomize server selection during a silent switch for settings - VPN-4071

bakulf:


src/apps/vpn/controller.cpp
```diff
@@ -218,19 +219,25 @@ bool Controller::activate(const ServerData& serverData) {
...
 
   clearConnectedTime();
   m_handshakeTimer.stop();
   m_activationQueue.clear();
 
-  Server exitServer = Server::weightChooser(m_serverData.exitServers());
+  QList<Server> exitServers = m_serverData.exitServers();
+  Server exitServer = serverSelectionPolicy == DoNotRandomizeServerSelection &&
+                              !m_serverData.exitServerPublicKey().isEmpty()
+                          ? Server::publicKeyChooser(
```
oskirby, line 238: This could also just be done with `ServerCountryModel::server(const QString& pubkey)` which will do an O(1) hash lookup instead of needing to iterate through the server list.

oskirby (commented):
Some general thoughts about the approach taken: We've kind of added a new action to the controller, namely to "reapply settings", but it has kind of been bodged into the `eActivate` action. For consistency with the rest of the actions that can be applied by `TaskControllerAction` maybe we should refactor this into something like:
- Add a new action enum, eg `TaskControllerAction::eUpdateConfig`
- Add a method to run it: `Controller::updateConfig()`

oskirby (approved):

Issue #3990, opened 2022-07-09
Add mechanism to refresh server latency data


data-sync-user:
The latency data a user sees is only relevant when the user remains in a fixed location. It may improve the users experience if we can detect that the user has changed location and invalidate the server latency data so that they are not making server choices based on old data.

For UX, it has also been requested to add a “refresh” button, which can trigger this directly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-2531)
┆Reporter: Naomi Kirby

data-sync-user:
➤ Santiago Andrigo commented:

Given we plan to have a refresh button, and that the server list would re-update anyway based on the app coming to the foreground, I think we can cancel this one. What do you think? Owen Kirby

data-sync-user:
➤ Beatriz Rizental Machado commented:

Removing the Data component as this one does not seem to be related to telemetry.

data-sync-user:
➤ Bianca Hidecuti commented:

Naomi Kirby Is there anything that QA can verify here, as the verification of the “Refresh” button will be covered in VPN-3023? Thank you.

data-sync-user:
➤ Bianca Hidecuti commented:

Marking this verified as fixed on Mozilla VPN 2.15 builds, across platforms, as the user is able to trigger the revaluation of the server latency data, by pressing the “Refresh” button. 

Naomi Kirby, if something else should be checked here, please let us know. Thanks.

PR #5882, opened 2023-02-06
VPN-2531: Expose method to refresh server connection scores.

oskirby:
## Description
Expose the `ServerLatency` object to QML, and provide a `refresh()` method to trigger a new ping sweep of the available servers, plus some useful properties about the progress of that ping sweep.

And, as some extra cleanup, I have moved the latency and cooldown properties out of the `Server` model and into the `ServerLatency` object.

## Reference
Github issue #3990 ([VPN-2531](https://mozilla-hub.atlassian.net/browse/VPN-2531))

## Checklist
    
[VPN-2531]: https://mozilla-hub.atlassian.net/browse/VPN-2531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/VPNServerList.qml
+++ b/nebula/ui/components/VPNServerList.qml
@@ -134,7 +134,7 @@ FocusScope {
                     accessibleName: VPNl18n.ServersViewRecommendedRefreshLabel
                     canGrowVertical: true
                     height: statusTitle.implicitHeight + VPNTheme.theme.vSpacingSmall
-                    rowShouldBeDisabled: !(VPNController.state === VPNController.StateOff)
+                    rowShouldBeDisabled: !(VPNController.state === VPNController.StateOff) || VPNServerLatency.isActive
 
                     onClicked: {
                         VPNServerLatency.refresh();
@@ -167,9 +167,9 @@ FocusScope {
                             horizontalAlignment: Text.AlignLeft
                             // TODO: Replace placeholder strings and generate
                             // values that will be set instead of `%1`
-                            text: VPNServerLatency.progress < 1.0
+                            text: VPNServerLatency.isActive
                                 ? "Checking... %1%".arg(Math.round(VPNServerLatency.progress * 100))
-                                : !statusComponent.rowShouldBeDisabled
+                                : (VPNController.state === VPNController.StateOff)
                                 ? "Last updated %1 ago.".arg(VPNServerLatency.lastUpdateTime)
                                 : "Last updated %1 ago. To update this list please first disconnect from the VPN."
                             wrapMode: Text.WordWrap
@@ -202,6 +202,7 @@ FocusScope {
                 Repeater {
                     id: recommendedRepeater
                     model: VPNServerCountryModel.recommendedLocations(5)
+
                     delegate: VPNClickableRow {
                         property bool isAvailable: modelData.connectionScore >= 0
                         id: recommendedServer
@@ -237,6 +238,19 @@ FocusScope {
                                 score: modelData.connectionScore
                             }
                         }
+                    }
+                }
+
+                Timer {
+                    id: recommendedLocationsRefreshTimer
+                    interval: 2000
+                    running: VPNServerLatency.isActive
+                    repeat: false
+                    onTriggered: {
+                        if (VPNServerLatency.isActive) {
+                            recommendedLocationsRefreshTimer.start();
+                        }
+                        recommendedRepeater.model = VPNServerCountryModel.recommendedLocations(5);
                     }
                 }
             }
--- a/nebula/ui/components/VPNServerList.qml
+++ b/nebula/ui/components/VPNServerList.qml
@@ -245,7 +245,6 @@ FocusScope {
                     id: recommendedLocationsRefreshTimer
                     interval: 2000
                     running: VPNServerLatency.isActive
-                    repeat: false
                     onTriggered: {
                         if (VPNServerLatency.isActive) {
                             recommendedLocationsRefreshTimer.start();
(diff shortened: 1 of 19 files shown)


oskirby:
There is a little bit of lingering weirdness with the server latency indicators in that changes in `ServerLatency::avgLatency()` don't trigger a property change. So it's not uncommon for the recommendations to fall out of sync with the displayed indicator. This could be addressed by emitting a singal on the `avgLatency()`, but I am worried about there being too many signals changing as the latency scan is in progress.

oskirby:
> There is a little bit of lingering weirdness with the server latency indicators in that changes in `ServerLatency::avgLatency()` don't trigger a property change. So it's not uncommon for the recommendations to fall out of sync with the displayed indicator. This could be addressed by emitting a signal on the `avgLatency()`, but I am worried about there being too many signals changing as the latency scan is in progress.

With any luck, this should now be fixed.

src/apps/vpn/models/servercity.cpp
```diff
@@ -108,15 +115,14 @@ const QString ServerCity::localizedName() const {
 }
 
 int ServerCity::connectionScore() const {
-  ServerCountryModel* scm = MozillaVPN::instance()->serverCountryModel();
+  ServerLatency* sl = MozillaVPN::instance()->serverLatency();
```
Gela, line 118: Hope this isn't too nit-picky, but can we rename all `sl` def/declarations to `ServerLatency* serverLatency` for better readability?

nebula/ui/components/VPNServerList.qml
```diff
@@ -167,8 +167,10 @@ FocusScope {
                             horizontalAlignment: Text.AlignLeft
                             // TODO: Replace placeholder strings and generate
                             // values that will be set instead of `%1`
-                            text: !statusComponent.rowShouldBeDisabled
-                                ? "Last updated %1 ago."
+                            text: VPNServerLatency.isActive
+                                ? "Checking... %1%".arg(Math.round(VPNServerLatency.progress * 100))
+                                : (VPNController.state === VPNController.StateOff)
+                                ? "Last updated %1 ago.".arg(VPNServerLatency.lastUpdateTime)
```
Gela, line 173: At least on MacOS due to the date format coming back, having "ago" at the end of the string doesn't seem right.

Screenshot 2023-02-09 at 10 00 31 AM

src/apps/vpn/serverlatency.cpp
```diff
@@ -225,3 +270,36 @@ void ServerLatency::recvPing(quint16 sequence) {
...
+    m_cooldown[publicKey] = QDateTime::currentSecsSinceEpoch() + timeout;
+  }
+
+  // Emit signals that the connection score may have changed.
+  ServerCountryModel* scm = MozillaVPN::instance()->serverCountryModel();
+  const Server& server = scm->server(publicKey);
+  const ServerCity& city =
+      scm->findCity(server.countryCode(), server.cityName());
+  if (city.initialized()) {
+    emit city.scoreChanged();
```
Gela, line 303: Sometimes instead of getting the "Last updated X ago" string, I get the default %1 placeholder. I am able to reproduce this if I hit refresh then turn the VPN on before the check completes. If user turns on the VPN mid-refresh we should probably just default back to the last "Last updated..." timestamp.

Screen Recording 2023-02-09 at 10 15 21 AM

Gela (changes requested):
Should there be a "cancel" button if the refresh is taking too long or if the user doesn't want to wait for it anymore?

src/apps/vpn/models/servercity.cpp
oskirby: That is a good suggestion, thank you.

oskirby:
I was hoping not to mess around with the UI too much and leave that work as a followup in [VPN-3023](https://mozilla-hub.atlassian.net/browse/VPN-3023) however, I have made a first pass to at least fix the grammar and string substitutions.

oskirby:
> Should there be a "cancel" button if the refresh is taking too long or if the user doesn't want to wait for it anymore?

I don't think a cancel button is necessary. The server list still work as normal while the refresh is running and the user can even choose a server or exit back to the main view without needing to wait for it to finish.

MattLichtenstein:
@oskirby you can roll back b317179 as it will be taken care of in [VPN-3023](https://mozilla-hub.atlassian.net/browse/VPN-3023) (we're actually going to re-utilize the same date/time format as [addon messages](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/02ae41a0d65f970c14153183ed98a0eccca58d58/src/apps/vpn/addons/addonmessage.cpp#L173-L209))

oskirby:
@MattLichtenstein sure thing!

nebula/ui/components/VPNServerList.qml
```diff
@@ -237,6 +240,19 @@ FocusScope {
                         }
                     }
                 }
+
+                Timer {
+                    id: recommendedLocationsRefreshTimer
+                    interval: 2000
+                    running: VPNServerLatency.isActive
+                    repeat: false
```
MattLichtenstein, line 248: `repeat` is set to false by default - no need to re-bind it here

MattLichtenstein (changes requested):
Reviewed the front-end. One small request

Gela (approved):

nebula/ui/components/VPNServerList.qml
oskirby: Done!

MattLichtenstein (approved):

PR #5889, opened 2023-02-07
Adding a glean extra key in case of handshake errors about the Controller operation - VPN-4083

bakulf:
@brizental can you review the code? I still need to ask for a data-review, but first I want to know if we are OK with this approach.

### DIFF
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -949,6 +949,11 @@ sample:
       transport:
         description: The Host network type (i.e wifi or 4g)
         type: string
+      operation_type:
+        description: |
+          The type of operation the controller was performing. Ex. none,
+          silent-switch, switch
+        type: string
 
   connectivity_stable:
     type: event
--- a/vpnglean/glean_parser_ext/templates/macros.jinja2
+++ b/vpnglean/glean_parser_ext/templates/macros.jinja2
@@ -29,7 +29,7 @@ struct {{ obj.name|Camelize }}ExtraParser : EventMetricExtraParser {
       keepStringsAlive.append(s);
       ffiExtras.values[count] = s.constData();
       {% elif type == "boolean" %}
-      ffiExtras.values[count] = parsedExtras._{{item|camelize}} ? "true" : "false";
+      ffiExtras.values[count] = parsedExtras._{{item|camelize}}.toBool() ? "true" : "false";
       {% elif type == "quantity" %}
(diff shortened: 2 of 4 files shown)


glean/metrics.yaml
```diff
@@ -949,6 +949,11 @@ sample:
       transport:
         description: The Host network type (i.e wifi or 4g)
         type: string
+      operation_type:
+        description: |
+          The type of operation the controller was performing. Ex. none,
+          silent-switch, switch
```
brizental, line 955: Can't we extract this same information from: https://dictionary.telemetry.mozilla.org/apps/mozilla_vpn/metrics/sample_controller_step?

What exactly is it we want to answer with this new collection?

vpnglean/glean_parser_ext/templates/macros.jinja2
```diff
@@ -29,7 +29,7 @@ struct {{ obj.name|Camelize }}ExtraParser : EventMetricExtraParser {
       keepStringsAlive.append(s);
       ffiExtras.values[count] = s.constData();
       {% elif type == "boolean" %}
-      ffiExtras.values[count] = parsedExtras._{{item|camelize}} ? "true" : "false";
+      ffiExtras.values[count] = parsedExtras._{{item|camelize}}.toBool() ? "true" : "false";
```
brizental, line 32: I am not sure the new collection is necessary. However, this definitely is :D Thanks for catching it!

glean/metrics.yaml
bakulf: We want to collect errors, right? Some errors are generated when we are in StateOn, which could come from StateSwitching.

vpnglean/glean_parser_ext/templates/macros.jinja2
bakulf: Initially I used a boolean :)

glean/metrics.yaml
brizental: Sure, but we can infer the state at the time the error occured by looking at the other event I linked to there.

bakulf: Yes, but it's not easy to compute I guess. The handshake timeout could happen in Connecting or Confirming state and we then need to go back in time to see the sequence of State changes per ping. Is it doable? Is it better to keep the current PR?

brizental: It is doable. It's not hard, it's medium difficulty. It requires some SQL or Looker calculations, but we should be able to do that. I am not very much in favour of adding redundant collection.

PR #5890, opened 2023-02-07
VPN-3880 and Nits:  Better Android compile experience

strseb:
## Description
 Small nits, see comments.

### DIFF
--- a/android/daemon/tunnel/libwg-go/Makefile
+++ b/android/daemon/tunnel/libwg-go/Makefile
@@ -45,6 +45,7 @@ $(BUILDDIR)/go-$(GO_VERSION)/.prepared: $(GRADLE_USER_HOME)/caches/golang/$(GO_T
 	patch -p1 -f -N -r- -d "$(dir $@)" < goruntime-boottime-over-monotonic.diff && \
 	touch "$@"'
 
+$(DESTDIR)/libwg-go.so: export GOROOT := $(BUILDDIR)/go-$(GO_VERSION)/
 $(DESTDIR)/libwg-go.so: export PATH := $(BUILDDIR)/go-$(GO_VERSION)/bin/:$(PATH)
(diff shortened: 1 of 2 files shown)


scripts/android/cmake.sh
```diff
@@ -117,8 +117,8 @@ fi
 printn Y "Cleaning the folder... "
 print G "done."
 
-rm -rf .tmp || die "Failed to remove the temporary directory"
-mkdir .tmp || die "Failed to create the temporary directory"
```
strseb, line 121: That was an Old thing we needed to to due to Qmake beeing wierd. 
We only need to remove android-build, so that on each compile the changes in `/android` get synced over. 
This should stop us pulling rust,sentry, etc everytime.

android/daemon/tunnel/libwg-go/Makefile
```diff
@@ -45,6 +45,7 @@ $(BUILDDIR)/go-$(GO_VERSION)/.prepared: $(GRADLE_USER_HOME)/caches/golang/$(GO_T
 	patch -p1 -f -N -r- -d "$(dir $@)" < goruntime-boottime-over-monotonic.diff && \
 	touch "$@"'
 
+$(DESTDIR)/libwg-go.so: export GOROOT := $(BUILDDIR)/go-$(GO_VERSION)/
```
strseb, line 48: This makes sure if you have another go env active... like conda, we still pickup the patched version this makefile creates :)

brizental (approved):

scripts/android/cmake.sh
brizental: Wait, but this is not in the PR anymore. Is that correct?

brizental: If this comes back, better to link the issue it fixes https://mozilla-hub.atlassian.net/browse/VPN-3880

strseb: oops- yes reverted too much of that file :D - will come back and i can link it.

PR #5819, opened 2023-01-31
VPN-3321 - Update to Qt 6.4.1

brizental:
Opening as a draft to see what explodes.

PR #5892, opened 2023-02-07
VPN-2458 - Update to Qt 6.3.2

brizental:
~Opening as a draft to see what explodes. Déjà vu https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5819~

This is ready for review. **All changes in this PR are necessary for all builds and tests to pass afte the Qt upgrade.**

- There are limitations with regards to using `environment` in WASM's QtLoaderw ith this upgrade that required a workaround. See [Hack around WASM environment issues](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892/commits/eb5fd21b2f302a1b9fe7dc8dcd3514d340a21110). 
- The WASM module generated is not capitalized anymore with the update, that was addressed by [Module -> module](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892/commits/a5e4cb03f5e357f2e9c968637916e3025f17ef39). 
- There are also issues with our TaskCluster setup that prevent it from rebuilding Qt after an update, which is addressed by [Catch misconfigs, remove qt-win-webengine](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892/commits/16f911b22044472ebf55d5578d62e5f9a6e7aa18) and [Hardcode Qt version in each taskcluster script](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892/commits/0b5cffb040d192b7f793bc1040b753a417525074).


brizental:
Alright, this is looking good as far as I can tell. No display issues like 6.4.

brizental:
@oskirby Owen, should I cherry pick that Apple universal build in here as well?

oskirby:
Looking at the tasks that ran in this PR, I don't think we actually triggered any of the `toolchain-qt-xxx` tasks after bumping the Qt version.

Note from the [build-windows/opt](https://firefox-ci-tc.services.mozilla.com/tasks/FKBghEXrSfm-y7d5NgNFnQ/runs/0/logs/public/logs/live.log) logs:
```
[task 2023-02-09T15:21:11.934Z] 
[task 2023-02-09T15:21:11.934Z] 
[task 2023-02-09T15:21:14.570Z] Using Qt version 6.2.4
[task 2023-02-09T15:21:14.663Z] Disabling Sentry, as params are not given
[task 2023-02-09T15:21:14.663Z] Sentry supported OS -> Windows;Darwin;Android
[task 2023-02-09T15:21:14.663Z] Cannot build sentry for Windows
```

@ahal shouldn't a change to the task definition in `taskcluster/ci/toolchain/qt.yml` have triggered a re-rebuild?

bakulf (changes requested):
A lot of unrelated changes. Can you revert/exclude the reformatting of tests?

taskcluster/ci/toolchain/qt.yml
```diff
@@ -5,7 +5,7 @@
 task-defaults:
     worker:
         env:
-            QT_VERSION: "6.2.4"
+            QT_VERSION: "6.3.2"
             QT_MAJOR: "6.2"
```
strseb, line 9: QT_MAJOS is 6.3 then?

strseb:
@oskirby - Changes in the Tasks env are not taken into account, only files. 
I've filed https://mozilla-hub.atlassian.net/browse/VPN-2572 ages ago because i thought this might bite us - well bea just stumbled about some tec-tebt. 
@brizental - Currently the tasks are not actually using that. Can you remove the TaskCluster "Qt" env thing and move it into the 3 toolchain scripts? That should do the trick for now. 
going forward i think we should attempt to have the toolchain tasks not fetch the stuff itselfs, but get a bundle from a fetch task in the future. We're just in this wierd state because the taskcluster runner just fails extracting the bundle - which is something we could also just fix. :)

strseb (changes requested):
We need to abandon the idea of setting the QT_Version in the toolchain tasks, via worker-env.
This should be inside the job's script so changes are picked up. Currently taskcluster is still using 6.2 for this PR.

taskcluster/ci/toolchain/qt.yml
brizental: Ok so, I added the QT vesions to the scripts themselves. The decision taks fails if I leave this environment empty though... Do you know why that is? The error is in the taskcluster code itself, but as far as I can tell, the env is not required in the worker section. Let me know.

src/apps/vpn/inspector/inspectorhandler.cpp
```diff
@@ -844,6 +845,14 @@ static QList<InspectorCommand> s_commands{
                        Constants::setVersionOverride(versionOverride);
                        return QJsonObject();
                      }},
+
+    InspectorCommand{"override_websocket_server_url",
```
birdsarah, line 849: Why is this in this PR? (Websocket test)

src/apps/vpn/websocket/websockethandler.cpp
```diff
@@ -71,12 +71,12 @@ QString WebSocketHandler::webSocketServerUrl() {
   return httpServerUrl.toLower().replace("http", "ws");
 }
 
-#ifdef UNIT_TEST
 // static
 void WebSocketHandler::testOverrideWebSocketServerUrl(const QString& url) {
   WebSocketHandler::s_customWebSocketServerUrl = url;
 }
 
+#ifdef UNIT_TEST
```
birdsarah, line 79: Why is this in this PR? (Websocket test)

src/apps/vpn/websocket/websockethandler.h
```diff
@@ -22,8 +22,10 @@ class WebSocketHandler final : public QObject {
 
   void initialize();
 
-#ifdef UNIT_TEST
+  // This is not in the UNIT_TEST block, because it is also used in the
```
birdsarah, line 25: Why is this in this PR? (Websocket test)

taskcluster/docker/wasm/Dockerfile
```diff
@@ -22,8 +22,8 @@ RUN apt-get -q update && apt-get install -y git ccache python3 python3-pip libgl
     cd /opt/ && \
     git clone https://github.com/emscripten-core/emsdk.git && \
     cd emsdk && \
-    ./emsdk install 2.0.14 && \
-    ./emsdk activate 2.0.14 && \
+    ./emsdk install 3.0.0 && \
+    ./emsdk activate 3.0.0 && \
```
birdsarah, line 26: Why is this in this PR? (wasm build updates)

tests/functional/helper.js
```diff
@@ -586,6 +586,13 @@ module.exports = {
     assert(
         json.type === 'set_version_override' && !('error' in json),
         `Command failed: ${json.error}`);
+  },
+
+  async overrideWebSocketServerUrl(url) {
```
birdsarah, line 591: Why is this in this PR? (Websocket test)

tests/functional/helperWasm.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
   },
 
   async send(msg) {
-    await driver.executeScript(`Module.inspectorCommand("${
+    await driver.executeScript(`module.inspectorCommand("${
```
birdsarah, line 63: Why is this in this PR? (wasm test)

tests/functional/setupWasm.js
```diff
@@ -110,6 +110,7 @@ exports.mochaHooks = {
     await startAndConnect();
     await vpn.setGleanAutomationHeader();
     await vpn.setSetting('tipsAndTricksIntroShown', 'true')
+    await vpn.flipFeatureOff('addonSignature');
```
birdsarah, line 113: Why is this in this PR? (wasm test)

wasm/controller.js
```diff
@@ -125,7 +125,7 @@ class Controller {
   _writeCommand(cmd) {
     return new Promise(resolve => {
       this._waitReadCallback = resolve;
-      Module.inspectorCommand(cmd);
+      module.inspectorCommand(cmd);
```
birdsarah, line 128: Why is this in this PR? (wasm test)

tests/functional/testPushMessageHandlers.js
```diff
@@ -72,6 +72,10 @@ describe('PushMessage.handlers', function() {
       }
     };
 
+    beforeEach(async () => {
```
birdsarah, line 75: Why is this in this PR? (websocket test)

wasm/main.js
```diff
@@ -74,7 +74,7 @@ class MVPNWasm {
         url == 'https://archive.mozilla.org/pub/vpn/speedtest/50m.data') {
       // 50mb of data is too much to be handled in the browser.
       setTimeout(
-          () => Module.mvpnNetworkResponse(id, JSON.stringify({
+          () => module.mvpnNetworkResponse(id, JSON.stringify({
```
birdsarah, line 77: Why is this in this PR? (wasm tweaks)

wasm/test.html
```diff
@@ -89,27 +79,49 @@ <h1>Mozilla VPN - <span class="web-viewer">WebAssembly Viewer</span></h1>
   // detector does not support cors.
   if (url.endsWith("/success.txt")) {
     setTimeout(() => {
-            Module.mvpnNetworkResponse(id, JSON.stringify({status: 200, body: btoa('success')})); }, 200);
+      module.mvpnNetworkResponse(id, JSON.stringify({ status: 200, body: btoa('success') }));
```
birdsarah, line 82: Why is this in this PR? (wasm test)

birdsarah (changes requested):
Typically squeaking in improvements as we go into a PR is fine. However, with a PR like this that: we may need to revert, that we will likely want to isolate it's impact, and that we will likely want to refer to in the future in order make the next update I'd like to ask for a really clean PR that only does the update.

At the least that definitely means moving out the wasm and websocket code that's in that PR unless it's a direct impact of the Qt upgrade. If it is a result of the Qt upgrade then the PR description should include that.

At the most that would mean first doing a PR that includes the extra fixes you've done like adding to `taskcluster/scripts/toolchain/compile_qt_6_mac.sh` to point to 6.2.4 and then a follow-on PR that only does 6.3 upgrade changes.

brizental:
> Typically squeaking in improvements as we go into a PR is fine. However, with a PR like this that: we may need to revert, that we will likely want to isolate it's impact, and that we will likely want to refer to in the future in order make the next update I'd like to ask for a really clean PR that only does the update.

> At the least that definitely means moving out the wasm and websocket code that's in that PR unless it's a direct impact of the Qt upgrade. If it is a result of the Qt upgrade then the PR description should include that.

The WASM test changes are all related to issues in the WASM tests after the Qt upgrade. I have specifically cleaned this PR up to not even include file re-formattings after Baku's review. Everything in it is necessary for the Qt upgrade to work. I have separated the WASM test changes in a specific commit  "[Hack around WASM environment issues](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892/commits/eb5fd21b2f302a1b9fe7dc8dcd3514d340a21110)" and added comments in the code that explain that the changes are necessary due to a Qt update, I thought that would be enough flagging for reviewers that the changes are there on purpose. I stand corrected and have updated the description.

> At the most that would mean first doing a PR that includes the extra fixes you've done like adding to taskcluster/scripts/toolchain/compile_qt_6_mac.sh to point to 6.2.4 and then a follow-on PR that only does 6.3 upgrade changes.

That fix is a bandaid. If we revert this PR for any reason I think that should go with it. As @strseb  pointed out there are tickets already filed to properly address those issues. I'd rather we prioritize those later instead of blocking this PR on that. WDYT?

birdsarah (approved):

birdsarah:
Thank you for updating the description and providing the extra clarity and context. When you merge this PR if you're able to ensure that the PR description makes it into the commit message that would be :rocket:.

brizental:
I am moving this back into a draft. Turns out the unit test jobs are not using the updated Qt. Also, WASM decided to stop working after the rebases, see error logs on the CI jobs. 🙃 🆘

brizental:
I am taking a step back from this to work on https://mozilla-hub.atlassian.net/browse/VPN-4225 first. It is too hard to reason about this PR as it is right now.

oskirby:
Given the complexity of this upgrade, perhaps we should instead think about splitting this work into smaller pieces and only try focusing on one platform at a time.

brizental:
> Given the complexity of this upgrade, perhaps we should instead think about splitting this work into smaller pieces and only try focusing on one platform at a time.

I think there is a opportunity to instead attempt to make the whole Qt upgrate task an easier one by standardizing how Qt is installed / fetched on CI + by centralizing the Qt version declaration. That is what the ticket I filed is about. Do you have concerns about such an approach @oskirby?

oskirby:
@brizental I think that standardizing the Qt installation method is going to be a similarly difficult task and I am concerned that it is going to get stuck as well. The smaller we can make these tasks the greater the likelyhood that we can land them.

In particular, I have some concerns about upgrading Qt on Linux as we have been relying mostly on the distributions' packages for us and we don't have much control over this ourselves unless we switch to a static build of Qt and manage it ourselves. I haven't been impressed with `aqt` in the past either; it has kind of worked for tests but breaks every now and then due to network outages, while the native packages are much more reliable.

oskirby:
We've had some discussions about switching to a static Qt build for Linux before to try and produce standalone `.deb` files that might have better portability (similar to how Zoom and VS Code are installed, for example). So, I do think it's a good idea to try and standardize on a static Qt build that we produce in taskcluster, but I know that it's going to be a lot of work.

birdsarah:
Will bring this back in the local build doc.

PR #5959, opened 2023-02-08
Enable addon per-language thresholds - VPN-4104

bakulf:


docs/add-on.md
```diff
@@ -49,6 +49,7 @@ Add-ons can enable and disable themselves using the `conditions` key in the mani
 | start_time | The epoch time that activates the current add-on | Integer | No | Yes |
 | end_time | The epoch time that deactivates the current add-on | Integer | No | Yes |
 | javascript | A script file is executed to change the condition status. See below | String | No | Yes | 
+| translation_threshold | The translation threshold to use. By default 0 (all is included) | Number | No | No |
```
flodolo, line 52: Should it be the other way around? i.e. default 1.0, given it seems like the most common case, and can be set to 0 as needed.

scripts/addon/build.py
```diff
@@ -377,6 +379,20 @@ def qtquery(qmake, propname):
                 os.system(f"{lconvert} -if xlf -i {xliff_path} -o {locale_file}")
                 os.system(f"{lrelease} -idbased {locale_file}")
 
+                xlifftool_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "utils", "xlifftool.py")
+                xlifftool_cmd = [sys.executable, xlifftool_path, "-C", f"--locale={locale}", xliff_path]
+                xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
+                completeness_output = xlifftool.stdout.decode("utf-8")
+                completeness.append(f"{locale}:{completeness_output}")
+
+        # If we don't have translations yet, we still have english at 100%.
```
flodolo, line 388:
```suggestion
        # If we don't have translations yet, we still have English at 100%.
```
nit

scripts/addon/build.py
```diff
@@ -377,6 +379,20 @@ def qtquery(qmake, propname):
                 os.system(f"{lconvert} -if xlf -i {xliff_path} -o {locale_file}")
                 os.system(f"{lrelease} -idbased {locale_file}")
 
+                xlifftool_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "utils", "xlifftool.py")
```
flodolo, line 382: I understand it's beyond the scope of this, but technically you could refactor that script a bit and import it here, instead of running it in a shell and capture the output.

brizental: I agree, that is one of the reason I opened https://mozilla-hub.atlassian.net/browse/VPN-3955

brizental (approved):

brizental:
r+wc. Let's address https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5959#discussion_r1100598921 before merge.

flodolo (approved):

PR #5969, opened 2023-02-09
VPN-4046 - Write functional test to catch multiple update messages

brizental:


src/apps/vpn/inspector/inspectorhandler.cpp
```diff
@@ -831,6 +831,13 @@ static QList<InspectorCommand> s_commands{
                        AddonManager::instance()->fetch();
                        return QJsonObject();
                      }},
+
+    InspectorCommand{"set_version_override", "Override the version string", 1,
```
bakulf, line 835: We already have something similar: force_update_check

src/shared/constants.cpp
```diff
@@ -13,13 +13,23 @@
...
 bool s_inProduction = true;
+QString s_versionOverride = "";
 }  // namespace
 
 bool Constants::inProduction() { return s_inProduction; }
 
 void Constants::setStaging() { s_inProduction = false; }
 
-QString Constants::versionString() { return QStringLiteral(APP_VERSION); }
+void Constants::setVersionOverride(const QString& versionOverride) {
```
bakulf, line 23: Revert this and see if we can use `InspectorHandler::appVersionForUpdate`

tests/functional/helper.js
```diff
@@ -568,17 +564,26 @@ module.exports = {
     _lastAddonLoadingCompleted = false;
 
     await this.setSetting(
-        'addonCustomServerAddress', `${addonServer.url}/${addonPath}/`);
+      'addonCustomServerAddress', `${addonServer.url}/${addonPath}/`);
```
bakulf, line 567: why all of these changes? I think the file is already indented via clang-format

bakulf (changes requested):

tests/functional/helper.js
brizental: This is a JavaScript file, so clang won't do anything... I have a "Format on Save" setup on VSCode and I didn't notice if had reformatted so much.

src/shared/constants.cpp
brizental: That doesn't work, because it doesn't override the actual version constant which is what the addon API uses.

src/apps/vpn/inspector/inspectorhandler.cpp
brizental: See my other comment. That doesn't work for this case.

src/shared/constants.cpp
bakulf: can we make it happen? I'm OK with your approach if we remove the other one.

tests/functional/helper.js
bakulf: This is not right. clang-format works with JS perfectly. See: `./scripts/apply-format -f -i tests/functional/helper.js`

src/shared/constants.cpp
brizental: Sure, the other one doesn't seem to be used anyways.

tests/functional/helper.js
brizental: Hm, TIL. I'll revert these changes, but we should be running these lints on CI.

src/shared/constants.cpp
brizental: Hm, so that is not used in tests but doesn't QA use it for something else? I can file a follow up to look into this, if you prefer. It is unrelated from this PR.

bakulf: They use it to test the update, forcing a lower version value.

brizental: Alright so that mean we can't remove it.

PR #5971, opened 2023-02-09
Fix tools after dependency bot updates - VPN-4140

bakulf:


strseb (approved):
Do we have a ci task that checks if those are not broken? 
If not can you open up a ticket? i can write one :)

PR #5975, opened 2023-02-09
VPN-4141 - Remove Glean Android daemon telemetry

brizental:
The Glean Migration work has caused this to crash on the Android client.

This is unfortunate, however it is temporary. We can reinstate daemon telemetry on https://mozilla-hub.atlassian.net/browse/VPN-3872.


strseb (approved):
:c

PR #5983, opened 2023-02-09
pre hook to make sure functional test dependencies are setup

mozrokafor:
## Description

    Test Chore: pre hook to setup test env

### DIFF
--- a/scripts/tests/prepare_tests.sh
+++ b/scripts/tests/prepare_tests.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+CHANGES=false
+if [[ ! -f ".env" ]]
+then
+    echo ".env file does not exist"
+    touch .env
+    cat << 'EOF' > .env
+        export PATH=$PATH:~/Qt/6.2.4/macos/bin:$PATH
+        export QT_MACOS_BIN=~/Qt/6.2.4/macos/bin
+        MVPN_API_BASE_URL=http://localhost:5000
+        MVPN_BIN=build/tests/dummyvpn/dummyvpn
+        ARTIFACT_DIR=tests/artifact
+EOF
+fi
+
+# check if build directory exists
+if [[ ! -d "./build/tests/dummyvpn" ]]
+then
+    echo "dummyvpn dir does not exist, run cmake command to build it"
+    CHANGES=true
+    cmake -B build .
+fi
+
+# rebuild if changes in the /src
+if [[ ! $(git status --porcelain --untracked-files=no | wc -l) -eq "0" ]]; then
+  echo "  🔴 Changes detected. Please rebuild dummyvpn."
+  CHANGES=true
+fi
+
+# create dummyvpn app if it doesnt exist
+if [[ ! -f "./build/tests/dummyvpn/dummyvpn" ]]
+then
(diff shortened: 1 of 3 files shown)


scripts/tests/prepare_tests.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+if [[ ! -f ".env" ]]
+then
+    echo ".env file does not exist"
+    touch .env
+    cat << 'EOF' > .env
+        export PATH=$PATH:~/Qt/6.2.4/macos/bin:$PATH
```
bakulf, line 7: this works on your computer only :) Not on mine where:
- I use a different version of QT
- it's not installed in that path
- and it's not mac...

scripts/tests/prepare_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+        export QT_MACOS_BIN=~/Qt/6.2.4/macos/bin
+        MVPN_API_BASE_URL=http://localhost:5000
+        MVPN_BIN=build/tests/dummyvpn/dummyvpn
+        ARTIFACT_DIR=tests/artifact
+EOF
+fi
+
+# check if build directory exists
+if [[ ! -d "./build/tests/dummyvpn" ]]
```
bakulf, line 16: note that 'build' is not a fix folder. Locally I don't use `build`.

scripts/tests/prepare_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+then
+    echo "dummyvpn dir does not exist, run cmake command to build it"
+    cmake -B build .
```
bakulf, line 19: I think the mocha beforeAll() function should check if the client exists and complain with a proper error message. Nothing more than this.

scripts/tests/prepare_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# install node modules to run functional tests if it doesn't exist
+if [[ ! -d "./node_modules" ]]
+then
+    echo "node modules does not exist, run npm install"
+    npm install
+fi
+
+# generate addons to run tests if they dont exist
+if [[ ! -d "./tests/functional/addons/01_empty_manifest" ]]
```
bakulf, line 42: this is a test specific file. It should not be given as granted

bakulf (changes requested):
I think all of this should (if we want - and I'm not 100% about it) in a `beforeAll()`

scripts/tests/prepare_tests.sh
mozrokafor: good catch, will remove

scripts/tests/prepare_tests.sh
mozrokafor: will remove the build command and put that in a string text

scripts/tests/prepare_tests.sh
mozrokafor: to my understanding this directory is a requirement to run the functional tests, if it does not exist should the dev trying to run tests be notified about the issue? better yet automatically generate the files for them?

mozrokafor:
team, will be advised to setup an alias(or best alternative) on their machine to take care of this.

PR #5985, opened 2023-02-09
CI-Blocker: update rustup

strseb:
## Description
We updated taskgraph, therefore invalidated all caches. 
During that time rustup.exe got updated, so we need to adjust size/shasum aswell.


mozrokafor (approved):
LGTM

PR #5987, opened 2023-02-10
Enable smarter fallback languages - VPN-4143

bakulf:


flodolo (approved):
Can't review the code, but r+ for the approach.

Could you add the JSON with fallback to codeowners, just in case, to make sure it doesn't get edited without review?

The other thing is making sure we have QA, or someone, testing the app to make sure this works as expected.

birdsarah (changes requested):
Looks good to me. Would like a test for the new logic.

PR #5993, opened 2023-02-10
Move testtasks in apps/unit_tests - VPN-4156

bakulf:
This PR does also the following:
    - remove tests for servers, accounts, devices - we have a great test coverage in functional tests
    - move functional and group tasks in the shared folder


src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -236,7 +236,7 @@ ColumnLayout {
             objectName: _objectName + "-relayUpsell-upgradeButton"
 
             onClicked: {
-                VPNGleanDeprecated.recordGleanEvent("bundle_upsell_upgrade_clicked");
+                MZGleanDeprecated.recordGleanEvent("bundle_upsell_upgrade_clicked");
```
brizental, line 239: If we are going to rename it to mzglean, maybe we should also rename the rust crate.

brizental (approved):
r+wc.

It's confusing to have two completely unrelated things in the same PR. No need to separate it now, but just something to keep in mind for future PRs.

PR #5994, opened 2023-02-10
Rename VPNL18n to VPNI18n

bakulf:
@flodolo how bad is it if I rename l18nstrings_p.cpp to i18nstrings_p.cpp from a Pontoon standpoint?


flodolo (approved):
It shouldn't create issues. What we have in the XLIFF are the qml filenames where the strings are (if they're not in `strings.yaml`)

lesleyjanenorton (approved):
"linternationalization"

PR #6006, opened 2023-02-13
VPN-4139 - Add back auth header to products requests

brizental:
This was accidentally removed by https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5503.

I wonder if we could have tests that would have caught this. I had two ideas 1. Running a local instance of Guardian alongside unit / functional tests instead of mocking everything. 2. Creating a suite of integration tests that run on merge or on the release branch that do not mock Guardian at all. WDYT?


bakulf (approved):

bakulf:
Can you update the tests too? Simply add `requiredHeaders: ['Authorization'],` in servers/guardian_endpoints.js+240

PR #6013, opened 2023-02-13
Fix the iOS crash - VPN-4112

bakulf:
Somehow iOS compile is unable to use '&' between enums. Let's use '+' to be consistent with the other DNS settings.


oskirby:
Hmm, neat, I bet this is also fixes [VPN-4048](https://mozilla-hub.atlassian.net/browse/VPN-4048).

oskirby (approved):

PR #6017, opened 2023-02-14
VPN-3023: Recommended server UI improvements

MattLichtenstein:
## Description

- Add a setting to keep track if the last selected server was a recommended server (`recommendedServerSelected`)
- Center selected server in "All" server list when going from "Recommended" -> "All"
- Update "Refresh recommended servers" button label strings and add them to strings.yaml
- Move date formatting function used by message addons to the localizer class

## Reference

[VPN-3023: Implement the Recommended Server UI](https://mozilla-hub.atlassian.net/browse/VPN-3023)
[VPN-4123: Collapsible card icon incorrect alignment](https://mozilla-hub.atlassian.net/browse/VPN-4123)


translations/strings.yaml
```diff
@@ -90,6 +90,15 @@ serversView:
     comment: Label for the “recommended” servers tab
   searchPlaceholder: Search countries
   searchNoResultsLabel: No results. Try a different search
+  recommendedRefreshlLoadingLabel:
+    value: Checking... %1%
```
flodolo, line 94: Please check the linter ;-)

Besides that, is it just me or that's a very strange structure? The ellipsis means that the action is in progress, they're never followed by more information

`Checking (%1%)…` would seem more natural to me.

translations/strings.yaml
```diff
@@ -90,6 +90,15 @@ serversView:
...
+    comment: Loading string while recommended servers are being calculated. Argument is the percentage of completion as an integer (i.e. 25)
+  recommendedRefreshLastUpdatedLabel:
+    value: Last updated %1.
+    comment: The time (or in extremely rare circumstances, the date, day of the week, or "yesterday") which the recommended servers list was last updated. Argument is the time or date/day
```
flodolo, line 98: How is the date formatted? Where is "yesterday" coming from?

Could we have a clearer split between sentence and date, i.e. `Last update(d): %1`?

flodolo (changes requested):

translations/strings.yaml
MattLichtenstein: > Besides that, is it just me or that's a very strange structure? The ellipsis means that the action is in progress, they're never followed by more information

> Checking (%1%)… would seem more natural to me.

@chrisrooneys - thoughts?

translations/strings.yaml
MattLichtenstein: The date is formatted identically to our in-app messages. A quick summary: 

- Timestamps can be “\<time\> “Yesterday”, “\<day of the week\>”, or “\<date\>”
  - Examples: “9:41 AM”, “Yesterday”, “Monday”, “Friday”, “7/5/22”
  -"Yesterday" is translated [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/082194b604b0b67c8d657b9d8252dc9636eaeb81/translations/strings.yaml#L743) (though now I am wondering if the capitalization is okay - what do you think?)
  -[Days of the week](https://doc.qt.io/qt-6.2/qlocale.html#dayName) and date formatting are handled by Qt on a locale basis

> Could we have a clearer split between sentence and date, i.e. Last update(d): %1?

@chrisrooneys - thoughts?

flodolo: >         -"Yesterday" is translated [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/082194b604b0b67c8d657b9d8252dc9636eaeb81/translations/strings.yaml#L743) (though now I am wondering if the capitalization is okay - what do you think?)

It's not, that's why I asked ;-) 

You need a new string that will be used exclusively in this context.

translations/strings.yaml
chrisrooneys: I'm good with that change

translations/strings.yaml
chrisrooneys: If I'm understanding, we need a separate string for "yesterday" as well as using a colon to divide the date from the rest of the sentence? I'm good with both of those. @MattLichtenstein let me know what you need.

oskirby: It's worth pointing out that the server latency data is refreshed on a 30 minute timer, so this string is almost always going to show "Today" except when the user has had their VPN turned on for more than a day (which blocks the refresh timer). It's not necessarily a problem, but it just makes this string a little less useful than you might expect.

src/shared/localizer.cpp
```diff
@@ -242,6 +243,45 @@ QMap<QString, double> Localizer::loadTranslationCompleteness(
   return result;
 }
 
+// static
+QString Localizer::formatDate(const QDateTime& nowDateTime,
+                                   const QDateTime& messageDateTime) {
+  qint64 diff = messageDateTime.secsTo(nowDateTime);
+  if (diff < 0) {
+    // The addon has a date set in the future...?
```
oskirby, line 251: Nit: this method is used for more than just addons now and the comment could get updated.

oskirby (approved):
Functionally, I think this looks like a solid implementation and I think we can move forward with this PR once the rest of the stakeholders are happy with it.

Aesthetically, I'm getting kinda worried that we are now loosing a lot of vertical real estate in the server selection view to all the tabs, labels and tooltips that come along for the ride. With 5 recommended locations, I have to scroll the window to see them all.

src/shared/localizer.cpp
peterbenvenuto: I hear you, but I think this is fine for now, the card is needed and was tested to be useful so we can't remove it. If we want to look at the screen again later down the road we can create a ticket for it, but for now this is what was tested so it we wont make any further changes to the UI. @oskirby

flodolo (approved):

flodolo (approved):

src/shared/localizer.cpp
lesleyjanenorton: Seconding Owen's aesthetic concerns. I'll file a ticket for exploring alternatives in future iterations. 
Screen Shot 2023-02-17 at 10 41 08 AM

lesleyjanenorton (approved):

flodolo (approved):

PR #6027, opened 2023-02-14
Do not show DNS-invalid warning if the DNS setting is not set to custom - VPN-4167

bakulf:


src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
```diff
@@ -209,7 +209,7 @@ VPNViewBase {
                     }
                     Layout.topMargin: VPNTheme.theme.listSpacing
 
-                    visible: ipInput.valueInvalid && ipInput.visible
+                    visible: ipInput.valueInvalid && ipInput.visible && VPNSettings.dnsProviderFlags === VPNSettings.Custom
```
MattLichtenstein, line 212: this should be put in a loader - though I am okay with doing that as a follow-up since it's not the premise of the issue

flozia (approved):

MattLichtenstein (approved):

lesleyjanenorton (approved):

src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
bakulf: @MattLichtenstein tell me more. I don't understand what you mean.

MattLichtenstein: @bakulf eventually we should move \~most\~ ("most" instead of "all" because some things take too long to load) components that we conditionally show/hide into [loaders](https://doc.qt.io/qt-6/qml-qtquick-loader.html) (we would essentially move the `visible` logic to the [active](https://doc.qt.io/qt-6/qml-qtquick-loader.html#active-prop) property to load/unload the component). It's a performance recommendation from the [qml docs](https://doc.qt.io/qt-6/qml-qtquick-loader.html#active-prop)

PR #6029, opened 2023-02-14
VPN-4150: Add multihop connection scoring

oskirby:
## Description
Recently, it was decided that the connection scoring indicators for the multihop view weren't a reliable indicator of expected connection performance, and as such we removed them in PR #5984 until they could be re-implemented in a way that's cognizant of the entry/exit server pairing. However, I **really** just want to get this featured closed out and finished, so let's implement it.

This refactors the `ServerCity` object just a little bit, and adds a QML-invokable method to compute a connection score for a multihop pairing. This method follows mostly the same algorithm as for single-hop, but uses geographic distance between the cities as a proxy for expected latency. The common components of the scoring algorithm have been split out into a private method `ServerCity::baseCityScore()` which can then be modified by either the single/entry/exit use cases.

## Reference
Jira issue [VPN-4150](https://mozilla-hub.atlassian.net/browse/VPN-4150)

## Checklist

[VPN-4150]: https://mozilla-hub.atlassian.net/browse/VPN-4150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/VPNServerCountry.qml
+++ b/nebula/ui/components/VPNServerCountry.qml
@@ -18,7 +18,10 @@ VPNClickableRow {
     property string _countryCode: code
     property var currentCityIndex
     property alias serverCountryName: countryName.text
-    property bool showConnectionScores: (focusScope.currentServer.whichHop === "singleHopServer")
+
+    // The city connection score can be used for every case except the multihop exit location,
+    // where we need to use the scoring between the entry and exit locations instead.
+    property bool useMultiHopScore: (focusScope.currentServer.whichHop === "multiHopExitServer")
 
     property bool hasAvailableCities: cities.reduce((initialValue, city) => (initialValue || city.connectionScore >= 0), false)
 
@@ -207,8 +210,7 @@ VPNClickableRow {
                         rightMargin: VPNTheme.theme.hSpacing
                         verticalCenter: parent.verticalCenter
                     }
-                    score: (focusScope.currentServer.whichHop !== "multiHopExitServer") ? modelData.connectionScore
-                           : modelData.multiHopScore(segmentedNav.multiHopEntryServer[0], segmentedNav.multiHopEntryServer[1])
+                    score: useMultiHopScore ? modelData.multiHopScore(segmentedNav.multiHopEntryServer[0], segmentedNav.multiHopEntryServer[1]) : modelData.connectionScore
                 }
             }
         }
--- a/nebula/ui/components/VPNServerCountry.qml
+++ b/nebula/ui/components/VPNServerCountry.qml
@@ -207,7 +207,8 @@ VPNClickableRow {
                         rightMargin: VPNTheme.theme.hSpacing
                         verticalCenter: parent.verticalCenter
                     }
-                    score: showConnectionScores ? modelData.connectionScore : (isAvailable ? VPNServerCountryModel.NoData : VPNServerCountryModel.Unavailable)
+                    score: (focusScope.currentServer.whichHop !== "multiHopExitServer") ? modelData.connectionScore
+                           : modelData.multiHopScore(segmentedNav.multiHopEntryServer[0], segmentedNav.multiHopEntryServer[1])
                 }
             }
         }
--- a/nebula/ui/components/VPNServerLatencyIndicator.qml
+++ b/nebula/ui/components/VPNServerLatencyIndicator.qml
@@ -15,7 +15,7 @@ VPNIcon {
         // Low latency
         State {
             when: (score === VPNServerLatency.Good ||
-                score == VPNServerLatency.Excellent)
+                score === VPNServerLatency.Excellent)
             PropertyChanges {
                 target: latencyIndicator
                 source: "qrc:/nebula/resources/server-latency-strong.svg"
--- a/nebula/ui/components/VPNServerLatencyIndicator.qml
+++ b/nebula/ui/components/VPNServerLatencyIndicator.qml
@@ -7,15 +7,15 @@ import QtQuick 2.5
 import Mozilla.VPN 1.0
 
 VPNIcon {
-    property int score: VPNServerCountryModel.NoData
+    property int score: VPNServerLatency.NoData
 
     id: latencyIndicator
 
     states: [
         // Low latency
         State {
-            when: (score === VPNServerCountryModel.Excellent ||
-                score === VPNServerCountryModel.Good)
+            when: (score === VPNServerLatency.Good ||
+                score == VPNServerLatency.Excellent)
             PropertyChanges {
                 target: latencyIndicator
                 source: "qrc:/nebula/resources/server-latency-strong.svg"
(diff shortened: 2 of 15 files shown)


src/apps/vpn/models/servercity.cpp
```diff
@@ -167,15 +150,70 @@ int ServerCity::connectionScore() const {
...
+    }
+  }
+
+  if (score > Excellent) {
+    score = Excellent;
+  }
+  return score;
+}
+
+int ServerCity::multiHopScore(const QString& country,
```
birdsarah, line 197: Seems like this and the above method would also benefit from tests. I'm not sure what our mocking capabilities are in order to be able to do this well. 

In particular, if I'm understanding the PR message correctly, given that we won't be using this code for a bit (...maybe never??) having tests will help us come back to it later and document then intended scoring.

birdsarah (changes requested):

src/apps/vpn/models/location.cpp
```diff
@@ -100,3 +100,27 @@ double Location::distance(double latitude, double longitude) const {
 
   return qAcos(m_latSin * otherSin + m_latCos * otherCos * diffCos);
 }
+
+// The same algorithm as above, but static and accepts any QObject with
+// latitude and longitude properties.
+double Location::distance(const QObject* a, const QObject* b) {
```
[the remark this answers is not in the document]
oskirby: I will add some unit tests.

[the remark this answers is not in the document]
oskirby: To answer your question, if either `QObject` is missing the latitude and longitude properties, then one or more of the booleans will be false, and this function returns false.

src/apps/vpn/models/servercity.cpp
oskirby: Yes, unit tests are good, I will add some.

This method is invoked for the multihop exit servers, so it is used whenever the feature is enabled.

nebula/ui/components/VPNServerCountry.qml
```diff
@@ -207,7 +207,8 @@ VPNClickableRow {
                         rightMargin: VPNTheme.theme.hSpacing
                         verticalCenter: parent.verticalCenter
                     }
-                    score: showConnectionScores ? modelData.connectionScore : (isAvailable ? VPNServerCountryModel.NoData : VPNServerCountryModel.Unavailable)
+                    score: (focusScope.currentServer.whichHop !== "multiHopExitServer") ? modelData.connectionScore
```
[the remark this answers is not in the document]
oskirby: Hmm, perhaps this would benefit from a comment to explain things better. The connection scores are trying to convey an idea of how good a location is when the user is browsing the server list, and it just so happens that the `modelData.connectionScore` property is valid for both the single-hop and multihop entry server selections, because we are representing the expected connection quality between the client and that chosen server.

It is only for the multihop exit where we need a special case, because we need to show the expected link performance between the entry and exit servers, and this has nothing to do with the user's location.

My preference was to put the common case first and then the multihop exit case second, but I can flip that around if you think it's better that way.

src/apps/vpn/models/servercity.cpp
oskirby: Hmm, the dependency tangle between `ServerCity` and `ServerLatency` classes is turning out to make this a pain in the butt to mock out for unit testing, but I will get this tested or die trying.

birdsarah: In case it helps, I am a proponent of (re-)architecting code for testability.

nebula/ui/components/VPNServerCountry.qml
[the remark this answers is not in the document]
birdsarah: > My preference was to put the common case first and then the multihop exit case second, but I can flip that around if you think it's better that way.

This makes sense. My comment was poorly worded. I think I would have better said: I find the current code a bit hard to read; can we think of ways of improving the readability? One though is keeping the test condition as a variable calculated elsewhere.  Second thought is that I find it easier to read non-negative boolean tests...but this may well just be me.

nebula/ui/components/VPNServerCountry.qml
```diff
@@ -18,7 +18,10 @@ VPNClickableRow {
     property string _countryCode: code
     property var currentCityIndex
     property alias serverCountryName: countryName.text
-    property bool showConnectionScores: (focusScope.currentServer.whichHop === "singleHopServer")
+
+    // The city connection score can be used for every case except the multihop exit location,
+    // where we need to use the scoring between the entry and exit locations instead.
+    property bool useMultiHopScore: (focusScope.currentServer.whichHop === "multiHopExitServer")
```
bakulf, line 24: Can we have an enum?

nebula/ui/components/VPNServerLatencyIndicator.qml
```diff
@@ -7,15 +7,15 @@ import QtQuick 2.5
...
 
     id: latencyIndicator
 
     states: [
         // Low latency
         State {
-            when: (score === VPNServerCountryModel.Excellent ||
-                score === VPNServerCountryModel.Good)
+            when: (score === VPNServerLatency.Good ||
+                score == VPNServerLatency.Excellent)
```
bakulf, line 18: ===

src/apps/vpn/models/location.cpp
```diff
@@ -100,3 +103,30 @@ double Location::distance(double latitude, double longitude) const {
...
+  bool aLatOkay;
+  bool aLongOkay;
```
bakulf, line 111: = false

src/apps/vpn/models/servercity.cpp
```diff
@@ -149,33 +155,33 @@ int ServerCity::connectionScore() const {
...
-      sumLatencyMsec += serverLatency->getLatency(pubkey);
-      activeServerCount++;
-    }
+  int score = serverLatency->baseCityScore(this, country);
+  if (score <= ServerLatency::Unavailable) {
+    return score;
   }
 
-  return (sumLatencyMsec + activeServerCount - 1) / activeServerCount;
+  // Increase the score if the distance betweens servers is less than 1/8th of
```
bakulf, line 172: Between?

bakulf (approved):

nebula/ui/components/VPNServerCountry.qml
oskirby: The `whichHop` string is used in a bunch of places, including our glean telemetry, so it's not a trivial refactor.

src/apps/vpn/serverlatency.cpp
```diff
@@ -29,6 +29,9 @@ constexpr const int SERVER_LATENCY_MAX_RETRIES = 2;
 // Delay the progressChanged() signal to rate-limit how often score changes.
 constexpr const uint32_t SERVER_LATENCY_PROGRESS_DELAY_MSEC = 500;
 
+// Minimum number of redundant servers we expect at a location.
+constexpr int SCORE_SERVER_REDUNDANCY_THRESHOLD = 3;
```
birdsarah, line 33: Random thought for the future. It would be fun to think about making this configurable by an add-on so we can then run experiments on what a good threshold is.

birdsarah (approved):

PR #6031, opened 2023-02-14
VPN-3684 Add boolean metric to glean

mcleinman:
## Description

Adding boolean metric to glean. From some local tests, it seems to work.

Additionally, add some debug features to the rust version of glean when it's built with staging. This is in `lib.rs`. Let me know if you'd like this broken out into its own PR, or you feel like any of it is too heavy-handed.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-3684

## Checklist
    

qtglean/src/ffi/string.rs
```diff
@@ -19,14 +19,14 @@ pub extern "C" fn glean_string_test_get_value(
     id: u32,
     ping_name: FfiStr
 ) -> *mut c_char {
-    let returnString = with_metric!(
+    let return_string = with_metric!(
```
mcleinman, line 22: This resolves a style-related build warning that I missed when I created this metric.

qtglean/src/lib.rs
```diff
@@ -57,6 +57,11 @@ pub extern "C" fn glean_initialize(is_telemetry_enabled: bool, data_path: FfiStr
     };
 
     register_pings();
+    if channel == "staging" {
+        glean::set_debug_view_tag("MozillaVPNRust");
+        glean::set_log_pings(true);
+    };
```
brizental, line 63: I am not pro adding this. These can be set using environment variables when necessary. There is also a ticket to add these options to the dev menu in the future. Setting the debug view tag for staging at all times will spam the debug viewer which is bad, we def should _not_  do it. Setting the log pings spams the VPN logs, which I might be persuaded to agree with though. WDYT?

brizental (changes requested):

qtglean/src/cpp/boolean.cpp
```diff
@@ -0,0 +1,33 @@
...
+  return glean_boolean_test_get_num_recorded_errors(m_id, errorType);
+#endif
+  return 0;
+}
+
+bool BooleanMetric::testGetValue(const QString& pingName) const {
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
+  return glean_boolean_test_get_value(m_id, pingName.toLocal8Bit());
+#endif
+  return true;
```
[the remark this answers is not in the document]
mcleinman: Thanks. I see your point(s). Updated.

qtglean/src/lib.rs
mcleinman: I 100% agree on the debug view tag; it's been removed.

I also agree that logging pings makes the logs dirtier, but our staging logs are somewhat busy already, and for testing that doesn't necessarily seem "wrong" to me; I end up using a lot of filtering and searching. I don't feel super strongly about this, though - just say the word and I'll pull it out.

brizental: I'd prefer we get into the habit of turning on log pings with environment variables / add the option to the dev menu. The logs are already verbose, true. However I don't see that as a good reason to add stuff to them when it's not necessary.

brizental (approved):
r+wc

qtglean/src/lib.rs
mcleinman: Fair enough. I removed it. Thanks!

PR #6036, opened 2023-02-15
VPN-3848: Remove ignore=all from .gitmodules

birdsarah:
## Description

Uniform handling of gitsubmodules when it comes to "ignore"

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-3848

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,19 +2,15 @@
 	path = i18n
 	url = https://github.com/mozilla-l10n/mozilla-vpn-client-l10n.git
 	branch = main
-	ignore=all
 [submodule "3rdparty/wireguard-tools"]
 	path = 3rdparty/wireguard-tools
 	url = https://github.com/WireGuard/wireguard-tools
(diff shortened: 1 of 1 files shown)


bakulf (approved):

.gitmodules
```diff
@@ -2,19 +2,15 @@
 	path = i18n
 	url = https://github.com/mozilla-l10n/mozilla-vpn-client-l10n.git
 	branch = main
-	ignore=all
```
strseb, line 5: I'd like to keep this. Given we update this __everytime__ we build, this should not be commited.

birdsarah: That is not what this line does.

If you set `ignore=all` then when you type `git status` you do not see that there have been changes to the submodule.

If you then type `git add .` you will still add the i18n module.

The intention of removing `ignore=all` is to try and help folks not accidentally commit i18n updates by making it clearer that that change has happened locally.

PR #6038, opened 2023-02-15
VPN-3849 Align git submodule usage

birdsarah:
## Description

Use `git submodule update --init` throughout codebase in order to:
* standardize usage
* not update submodules ahead of committed versions

The exception to this is i18n which we want to pull from remote. I've isolated this to just the `import_languages.py` script.

@oskirby In making these changes, I realized that `ppa_script.sh` did not include a call to `import_languages.py`. Can you review my change.

@strseb `taskcluster/scripts/push-addons/deploy.sh`. I haven't changed it but wanted to check in with you. Why is this file:
* doing a sync first - can we document why this is necessary
* doing --recursive (it's the only place in the repo we do)
* doing --force (again only place we do it)

### DIFF
--- a/scripts/linux/script.sh
+++ b/scripts/linux/script.sh
@@ -120,11 +120,8 @@ print G "${SHORTVERSION}"
 rm -rf .tmp || die "Failed to remove the temporary directory"
 mkdir .tmp || die "Failed to create the temporary directory"
 
-print Y "Update the submodules..."
-git submodule init || die "Failed"
-git submodule update --remote --depth 1 i18n || die "Failed"
-git submodule update --remote --depth 1 3rdparty/wireguard-tools || die "Failed"
-git submodule update --depth 1 3rdparty/glean || die "Failed"
+print Y "Get the submodules..."
+git submodule update --init --depth 1 || die "Failed"
 print G "done."
 
 print G "Creating the orig tarball"
@@ -139,6 +136,9 @@ mkdir -p .tmp/${WORKDIR} || die "Failed"
 rsync -a --exclude='.*' ${RSYNC_EXCLUDE_DIRS} . .tmp/${WORKDIR} || die "Failed"
 print G "done."
 cd .tmp
+
+print Y "Importing translation files..."
+(cd $WORKDIR && python3 scripts/utils/import_languages.py) || die "Failed to import languages"
 
 print Y "Generating glean samples..."
 (cd $WORKDIR && python3 scripts/utils/generate_glean.py) || die "Failed to generate glean samples"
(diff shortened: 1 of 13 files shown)


bakulf (approved):

strseb:
> taskcluster/scripts/push-addons/deploy.sh. I haven't changed it but wanted to check in with you.

Reason is "Copy Pasta", it should not need it :)

strseb:
@birdsarah - Should in this pr also maybe remove the "ignore=all" (except for i18n) so that we are deliberate about what changed?

strseb (approved):

taskcluster/scripts/push-addons/deploy.sh
```diff
@@ -5,15 +5,14 @@
 
 set -e
 
-git submodule sync --recursive
-git submodule update --init --force --recursive --depth=1
 # Reqs
+git submodule update --init --recursive --depth=1
```
birdsarah, line 9: @strseb for your review

strseb (approved):

oskirby:
@birdsarah we don't actually use `ppa_script.sh` for anything that I'm aware of anymore. We should probably remove it instead of trying to keep it up to date.

scripts/linux/script.sh
```diff
@@ -140,6 +137,9 @@ rsync -a --exclude='.*' ${RSYNC_EXCLUDE_DIRS} . .tmp/${WORKDIR} || die "Failed"
 print G "done."
 cd .tmp
 
+print Y "Importing translation files..."
```
oskirby, line 140: IMO, we shouldn't be calling this script directly anymore, it will dirty the source directory with the translation build artifacts and everything (except the submodule updates) are now handled by CMake at build time.

However, the script still needs to exist for iOS because it's still stuck on qmake.

taskcluster/scripts/build/android_build_debug.sh
```diff
@@ -5,12 +5,8 @@
 set -e
 
 # This script is used in the Android Debug (universal) build task
-git submodule init
-git submodule update
-# glean
```
birdsarah, line 10: @strseb please review again :)

Owen and I were discussing the upcoming ability to remove import_languages and that it's only used for qmake ios build and, maybe something to do with addons, but it's not needed for cmake builds.

So, I took the liberty of removing both generate_glean and import_languages from this and `android_build_release`.

Let me know if you object.

taskcluster/scripts/build/android_build_release.sh
```diff
@@ -5,13 +5,8 @@
 set -e
 
 # This script is used in the Android Build Release (universal) build task
-git submodule init
-git submodule update
-# glean
-./scripts/utils/generate_glean.py
```
birdsarah, line 11: @strseb please review again :)

Owen and I were discussing the upcoming ability to remove import_languages and that it's only used for qmake ios build and, maybe something to do with addons, but it's not needed for cmake builds.

So, I took the liberty of removing both generate_glean and import_languages from this and `android_build_debug`.

Let me know if you object.

strseb (approved):
lgtm

oskirby (approved):
Looks good

PR #6060, opened 2023-02-16
Chore: VPN-Service better handle errors

strseb:
Did i just work 2 Days on fixing the wrong bug? Yes. I should read bug descriptions. 
So - this a take on the `SIGABRT` crashes we're currently seeinng on android/wg-go.
I cannot reproduce that thing locally, so this is more a "what could go wrong, and how can we handle this better?" type patch. 

## Description
SIGABRT - that can be fired by the jni whenever a call to the jni cannot continue. This is a series of patches trying to stop us getting into that state.

- Patch 1: Make jni catch nullptr's

Fist line, if we call GetChars aginst a null jstring, that's a 100% sigabort. Let's return and throw a standard kotlin exception.

- Patch 2: TurnOn should catch config errors and deactivate the old tunnel after verification

The reasons we could call wgTurnOn with a null jString might be caused by a faulty config. Let's have turnOn check the json and the generated wg-config string before even turning the old tunnel off. This hopefully should cause a faulty call to turnOn to a no-op

- Patch 3: reconnect() should not write members and catch exceptions

turnOn can throw an exception, reconnect did not handle that. (possible crash) 
Also reconnect did write to the member, which should only keep the last positive activation. 
Meaning 2 calls to reconnect might have polluted the state. 

- Patch 4: Handle String? in VPNServiceBinder
That's a nit.


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -273,12 +279,23 @@ class VPNService : android.net.VpnService() {
...
+        try {
+            this.turnOn(config, forceFallBack)
+        } catch (e: Exception) {
+            Log.e(
+                tag,
+                "VPN service - Reconnect failed"
+            )
+            // TODO: If we end up here, we might have screwed up the connection.
+            // we should put out a notification that the user go into the app and does a manual
+            // connection.
```
brizental, line 297: Please file a ticket for this follow up and add the link to that here :)

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
```diff
@@ -56,7 +56,11 @@ class VPNServiceBinder(service: VPNService) : Binder() {
                     // [data] is here a json containing the wireguard conf
                     val buffer = data.createByteArray()
                     val json = buffer?.let { String(it) }
-                    val config = JSONObject(json)
+                    val config = json?.let { JSONObject(it) }
+                    if (config == null) {
+                        Log.e(tag, "Client provided config was not parsable")
+                        return true
```
brizental, line 62: How come we return true here if there was an error?

brizental (approved):
This looks good to me, however it still feels fragile. Did you consider using something like a Result i.e. a structure that may either hold a success or an error response in all of these functions? This way we would be forced to deal with all situations when a succesfull outcome didn't happen. Does that even make sense here?

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
strseb: Ha, it's documented: `returns true if the [code] was accepted` - the function code is accepted, just the payload is wront :)

brizental: Yeah I read that, but  I assumed if there was an error then the code had not been accepted... Clearly I don't know what is happening here :sos:

PR #6061, opened 2023-02-16
Removal LD override in openssl build script, which is not doing anything.

bhearsum:
We already did this for openssl, and although the others seem to work fine without it, we probably ought to make it explicit.

This also removes the original fix attempt of setting LD for openssl, which didn't work out.

### DIFF
--- a/taskcluster/scripts/toolchain/compile_openssl.ps1
+++ b/taskcluster/scripts/toolchain/compile_openssl.ps1
@@ -22,9 +22,6 @@ Set-Location openssl-OpenSSL_1_1_1m
 
 $NASM_PATH = resolve-path ../nasm-2.15.05
 $env:PATH = "$FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64;$env:PATH;$NASM_PATH"
-$env:LD = resolve-path "$FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64/link.exe"
-
-where link
 
 if(!(Test-Path ../SSL)){
     New-Item -Path ../SSL -ItemType "directory"


bhearsum:
(Retrying due to the QT build failing because of running out of disk space...)

bhearsum:
I think this is ready to go. (The disk space issues seem to have been a one off.)

taskcluster/scripts/build/windows.ps1
```diff
@@ -26,7 +26,7 @@ git submodule update --remote i18n
 
 # Fix: pip scripts are not on path by default on tc, so glean would fail
 $PYTHON_SCRIPTS =resolve-path "$env:APPDATA\Python\Python36\Scripts"
-$env:PATH ="$FETCHES_PATH;$QTPATH;$PYTHON_SCRIPTS;$env:PATH"
+$env:PATH ="$FETCHES_PATH;$FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64;$env:PATH;$QTPATH;$PYTHON_SCRIPTS;$env:PATH"
```
jcristau, line 29: I dislike that this means we hardcode the msvc version number in yet more places. Given this doesn't actually seems to be necessary for this script and the qt one, do we really want to?

taskcluster/scripts/toolchain/compile_openssl.ps1
```diff
@@ -22,9 +22,6 @@ Set-Location openssl-OpenSSL_1_1_1m
 
 $NASM_PATH = resolve-path ../nasm-2.15.05
 $env:PATH = "$FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64;$env:PATH;$NASM_PATH"
-$env:LD = resolve-path "$FETCHES_PATH/VisualStudio/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64/link.exe"
-
-where link
```
jcristau, line 27: Thanks for cleaning that up!

jcristau (approved):

taskcluster/scripts/build/windows.ps1
bhearsum: I don't feel too strongly. I mainly added this because it seems like it would make things more deterministic, rather than relying on whatever is set on the system already?

bhearsum:
I think this is ready to go. (The disk space issues seem to have been a one off.)

birdsarah:
@bhearsum the AuthenticationInBrowser test has been fixed. Would you like to rebase and push? The windows build failed last time the CI ran, would be nice to see everything green.

bhearsum:
> @bhearsum the AuthenticationInBrowser test has been fixed. Would you like to rebase and push? The windows build failed last time the CI ran, would be nice to see everything green.

Will do. The Windows build actually failed because the QT build ran out of disk space though :grimacing:. I've been working with RelOps on that, and we've just increased their disk size.

PR #6064, opened 2023-02-16
VPN-3706 recover orphaned wireguard process

strseb:
## Description

> VPN client on Android suddenly appears to be switched off without any notice. Users may notice an icon on the device status bar that the VPN is still On and confirm that the device is still protected by doing a Mullvad check https://mullvad.net/en/check/.  The issue has been recurring since V 2.10. Filing bug again. Original Jira bug Does it impact functionality or aesthetics? Yes, aesthetics.

My take, the VPNService was destroyed, but the process (with wireguard) lives on. This would cause the vpn to be active but the Service does not know about any active wireguard handles. 

Let's give the service the ability to check in if there are any active handles at the time of initialization


android/daemon/tunnel/libwg-go/api-android.go
```diff
@@ -224,4 +224,10 @@ func wgVersion() *C.char {
 	return C.CString("unknown")
 }
 
+//export wgGetLatestHandle
+func wgGetLatestHandle() int32 {
+	return int32(len(tunnelHandles) -1)
```
oskirby, line 229: What happens if `len(tunnelHandles)` is zero?

strseb: Any value below 0 is considered "off"" / "wireguard-returned-an-error"

oskirby (approved):

PR #6066, opened 2023-02-16
VPN-3292 & VPN-3284: Add "+ tax" to subscription plan pricing

MattLichtenstein:
## Description

Subscription management screen: 
- Add "+ tax" string to the subscription plan value in the subscription management view for US + Canada based users

Mobile IAP screen: 
- Update value props to match new content
- Add "+ tax" to US and Canadian pricing sublabels
- Update monthly plan content to match other pricing tiers

~*Note: [VPN-3284](https://mozilla-hub.atlassian.net/browse/VPN-3284) depends on https://github.com/mozilla-services/guardian-website/pull/1610 to call this API before showing the mobile IAP screen~ we now use currency code from the respective mobile app stores to determine whether or not a product is subject to tax (instead of location of account creation via Guardian ip lookup)

## Reference

[VPN-3292: Implement change in the in-app Subscription Management page to mention US and Canada prices are '+ tax'](https://mozilla-hub.atlassian.net/browse/VPN-3292)
[VPN-3284: Implement change in the in-app mobile pricing page to mention that US and Canada prices are '+ tax'.](https://mozilla-hub.atlassian.net/browse/VPN-3284)

[VPN-3284]: https://mozilla-hub.atlassian.net/browse/VPN-3284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -123,16 +123,9 @@ RadioDelegate {
             // TODO (maybe) - Do we want to add the subscription duration in months to the model?
             property var subscriptionDuration: getSubscriptionDuration(productType)
 
-            //% "Monthly plan"
-            property string productSingleMonth: qsTrId("vpn.subscription.monthlyPlan")
-
             //: %1 is replaced by the subscription duration in months. %2 is replaced by the total subscription cost.
             //% "%1-month plan: %2"
             property string productMultiMonth: qsTrId("vpn.subscription.multiMonthPlan").arg(col.subscriptionDuration).arg(productPrice)
-
-            //: “/month” stands for “per month”. %1 is replaced by the monthly cost (including currency).
-            //% "%1/month"
-            property string monthlyPrice: VPNI18n.SubscriptionManagementPlanMonthly.arg(productMonthlyPrice)
 
             property int trialDays: productTrialDays
 
@@ -144,7 +137,7 @@ RadioDelegate {
                 font.pixelSize: 16
                 lineHeight: VPNTheme.theme.labelLineHeight + 2
                 lineHeightMode: Text.FixedHeight
-                text: col.subscriptionDuration > 1 ? col.productMultiMonth : col.monthlyPrice
+                text: col.subscriptionDuration > 1 ? col.productMultiMonth : VPNI18n.SubscriptionManagementPlanMonthly.arg(productMonthlyPrice)
                 verticalAlignment: Text.AlignVCenter
                 wrapMode: Text.WordWrap
 
@@ -160,16 +153,11 @@ RadioDelegate {
                 lineHeight: 17.68
                 lineHeightMode: Text.FixedHeight
                 text: {
-                    const isUsOrCa = VPN.countryCode === "US" || VPN.countryCode === "CA"
-
-                    if(col.subscriptionDuration === -1)
-                        return ""
-                    else if(col.subscriptionDuration > 1) {
-                        return isUsOrCa ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : col.monthlyPrice
+                    if(col.subscriptionDuration > 0) {
+                        return VPNSubscriptionData.plusTax ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : VPNI18n.SubscriptionManagementPlanMonthlyWithoutTax.arg(productMonthlyPrice)
                     }
-                    //col.subscriptionDuration === 1
                     else {
-                        return isUsOrCa ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : col.productSingleMonth
+                        return ""
                     }
                 }
                 wrapMode: Text.WordWrap
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -156,9 +156,7 @@ RadioDelegate {
                     if(col.subscriptionDuration > 0) {
                         return VPNSubscriptionData.plusTax ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : VPNI18n.SubscriptionManagementPlanMonthlyWithoutTax.arg(productMonthlyPrice)
                     }
-                    else {
-                        return ""
-                    }
+                    return ""
                 }
                 wrapMode: Text.WordWrap
 
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -154,7 +154,7 @@ RadioDelegate {
                 lineHeightMode: Text.FixedHeight
                 text: {
                     if(col.subscriptionDuration > 0) {
-                        return VPNSubscriptionData.plusTax ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : VPNI18n.SubscriptionManagementPlanMonthlyWithoutTax.arg(productMonthlyPrice)
+                        return ["USD", "CAD"].includes(productCurrencyCode) ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : VPNI18n.SubscriptionManagementPlanMonthlyWithoutTax.arg(productMonthlyPrice)
                     }
                     return ""
                 }
--- a/nebula/ui/components/VPNSubscriptionOption.qml
+++ b/nebula/ui/components/VPNSubscriptionOption.qml
@@ -132,7 +132,7 @@ RadioDelegate {
 
             //: “/month” stands for “per month”. %1 is replaced by the monthly cost (including currency).
             //% "%1/month"
-            property string monthlyPrice: qsTrId("vpn.subscription.price").arg(productMonthlyPrice)
+            property string monthlyPrice: VPNI18n.SubscriptionManagementPlanMonthly.arg(productMonthlyPrice)
 
             property int trialDays: productTrialDays
 
@@ -159,7 +159,19 @@ RadioDelegate {
                 font.pixelSize: VPNTheme.theme.fontSize
                 lineHeight: 17.68
                 lineHeightMode: Text.FixedHeight
-                text: col.subscriptionDuration !== -1 ? (col.subscriptionDuration > 1 ? col.monthlyPrice : col.productSingleMonth) : ""
+                text: {
+                    const isUsOrCa = VPN.countryCode === "US" || VPN.countryCode === "CA"
+
+                    if(col.subscriptionDuration === -1)
+                        return ""
+                    else if(col.subscriptionDuration > 1) {
+                        return isUsOrCa ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : col.monthlyPrice
+                    }
+                    //col.subscriptionDuration === 1
+                    else {
+                        return isUsOrCa ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : col.productSingleMonth
+                    }
+                }
                 wrapMode: Text.WordWrap
 
                 Layout.fillWidth: true
--- a/src/apps/vpn/models/subscriptiondata.cpp
+++ b/src/apps/vpn/models/subscriptiondata.cpp
@@ -183,7 +183,7 @@ bool SubscriptionData::fromJsonInternal(const QByteArray& json) {
     return false;
   }
 
-  //Payments made using USD or CAD are subject to tax
+  // Payments made using USD or CAD are subject to tax
   m_planRequiresTax = QStringList({"USD", "CAD"}).contains(m_planCurrency);
 
   // Payment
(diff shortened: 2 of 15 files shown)


src/apps/vpn/mozillavpn.h
```diff
@@ -89,6 +89,7 @@ class MozillaVPN final : public QObject {
   Q_PROPERTY(bool updating READ updating NOTIFY updatingChanged)
   Q_PROPERTY(bool stagingMode READ stagingMode CONSTANT)
   Q_PROPERTY(bool debugMode READ debugMode CONSTANT)
+  Q_PROPERTY(QString countryCode READ countryCode CONSTANT)
```
MattLichtenstein, line 92: @bakulf does this seem like an appropriate place for this property?

translations/strings.yaml
```diff
@@ -671,6 +671,15 @@ subscriptionManagement:
   planValueYearly:
     value: "%1 Yearly"
     comment: The %1 represents a localized currency string.
+  planValueMonthlyPlusTax:
+    value: "%1 Monthly + tax"
+    comment: The %1 represents a localized currency string. This string is only used in US and Canada
```
flodolo, line 676: Is the comment accurate about US and Canada? That would be surprising.

Still, I would remove it, because we might want to use this elsewhere in the future.

flodolo (approved):

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
```diff
@@ -280,17 +280,21 @@ VPNViewBase {
     function getPlanText(currencyCode, amount) {
         const amountDisplay = (amount || 0) / 100;
         const localizedCurrency = VPNLocalizer.localizeCurrency(amountDisplay, currencyCode);
+        const isUsOrCa = VPN.countryCode === "US" || VPN.countryCode === "CA"
```
lesleyjanenorton, line 283: Can we use `VPNSubscriptionData.planCurrency` as we did for gating the bundle upsell UI? cc @birdsarah sanity checking that FxA communicates this for App Store and Play Store subs?
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/7301a22f69a1de190001275d645a2c941a70044e/src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml#L155-L160

src/apps/vpn/mozillavpn.h
oskirby: It would probably be better to expose the `Location` model to QML and then you can get the client location with `VPNLocation.countryCode` amongst other details.

translations/strings.yaml
```diff
@@ -671,6 +671,21 @@ subscriptionManagement:
...
+    value: "%1 Half-yearly + tax"
+    comment: The %1 represents a localized currency string
+  planValueYearlyPlusTax:
+    value: "%1 Yearly + tax"
+    comment: The %1 represents a localized currency string
+  planMonthly:
+    value: Monthly plan: %1
+    comment: Price for a monthly subscription. %1 represents the localized currency string
+  planMonthlyPlusTax:
+    value: %1/month + tax"
```
flodolo, line 687:
```suggestion
    value: "%1/month + tax"
```

translations/strings.yaml
```diff
@@ -671,6 +671,21 @@ subscriptionManagement:
...
+    value: "%1 Monthly + tax"
+    comment: The %1 represents a localized currency string
+  planValueHalfYearlyPlusTax:
+    value: "%1 Half-yearly + tax"
+    comment: The %1 represents a localized currency string
+  planValueYearlyPlusTax:
+    value: "%1 Yearly + tax"
+    comment: The %1 represents a localized currency string
+  planMonthly:
+    value: Monthly plan: %1
```
flodolo, line 684:
```suggestion
    value: "Monthly plan: %1"
```

flodolo (changes requested):

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
```diff
@@ -280,17 +280,21 @@ VPNViewBase {
...
 
         switch (VPNSubscriptionData.planBillingInterval) {
             case VPNSubscriptionData.BillingIntervalMonthly:
                 // {¤amount} Monthly
-                return VPNI18n.SubscriptionManagementPlanValueMonthly.arg(localizedCurrency);
+                return isUsOrCa ? VPNI18n.SubscriptionManagementPlanValueMonthlyPlusTax.arg(localizedCurrency)
```
bakulf, line 288: This is definitely a subscription-data property.

src/apps/vpn/mozillavpn.h
bakulf: Nope. This is a subscription data property. The +tax should be computed in VPNSubscriptionData

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
bakulf, line 283: You want to have: if(VPNSubscriptionData.plusTax) ...

bakulf (changes requested):

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
bakulf: Plus you want to check the location at the subscription time, not the current one. Because if i travel to Canada it don't start paying taxes with my European subscription. 

This bit must be exploded through Guardian.

translations/strings.yaml
```diff
@@ -709,6 +724,9 @@ subscriptionManagement:
   addFirefoxRelay:
     value: Add Firefox Relay
     comment: Block of text appearing above a link to learn more about the VPN / Relay bundle offering.
+  valueProp1: No logging of your network activity
```
flodolo, line 727: Please add comments for these 3 strings to explain how/where they're used.

Also, I assume these are not 2.14 at this point?

MattLichtenstein: These are 2.15 strings, correct

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
MattLichtenstein: We can use `VPNSubscriptionData.planCurrency` here, since the user already has a subscription, but we cannot use it on the IAP screen since the user does not have a subscription at that point.

Should we differ our methods based on whether the user has a subscription or not

CC: @jswinarton @bakulf @birdsarah can we confirm that we should be using currency codes (USD, CAD) instead of country codes (US, CA) for the subscription plan value and NOT the IAP screen?

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeededIAP.qml
```diff
@@ -192,4 +189,8 @@ VPNFlickable {
             }
         }
     }
+
+    Component.onCompleted: {
+        VPN.scheduleRefreshDataTasks(false)
+    }
```
MattLichtenstein, line 195: Looking into finding a better place to execute `TaskGetLocation` because if it takes a long time, this could result in the subscription option strings changing. Should probably do it after sign in, before the IAP screen is showed (during the loader)

translations/strings.yaml
```diff
@@ -709,6 +727,15 @@ subscriptionManagement:
   addFirefoxRelay:
     value: Add Firefox Relay
     comment: Block of text appearing above a link to learn more about the VPN / Relay bundle offering.
+  valueProp1:
+    value: No logging of your network activity
+    comment: Value proposition used to upsell the product
```
flodolo, line 732:
```suggestion
    comment: Value proposition used to promote the bundle with Relay
```

translations/strings.yaml
```diff
@@ -671,6 +671,24 @@ subscriptionManagement:
   planValueYearly:
     value: "%1 Yearly"
     comment: The %1 represents a localized currency string.
+  planValueMonthlyPlusTax:
+    value: "%1 Monthly + tax"
```
flodolo, line 675: One thing I'm missing: why "X Monthly" but also "X/month"? Is one of the 2 obsolete, is this a testing? I expect most locales to just translate them the same way.

nebula/ui/components/VPNSubscriptionOption.qml
```diff
@@ -159,7 +152,14 @@ RadioDelegate {
                 font.pixelSize: VPNTheme.theme.fontSize
                 lineHeight: 17.68
                 lineHeightMode: Text.FixedHeight
-                text: col.subscriptionDuration !== -1 ? (col.subscriptionDuration > 1 ? col.monthlyPrice : col.productSingleMonth) : ""
+                text: {
+                    if(col.subscriptionDuration > 0) {
+                        return VPNSubscriptionData.plusTax ? VPNI18n.SubscriptionManagementPlanMonthlyPlusTax.arg(productMonthlyPrice) : VPNI18n.SubscriptionManagementPlanMonthlyWithoutTax.arg(productMonthlyPrice)
+                    }
+                    else {
```
bakulf, line 159: No `else` after `return`

bakulf: Plus why this ""? Can you simply have the first if-stmt?

src/apps/vpn/models/subscriptiondata.cpp
```diff
@@ -222,6 +224,16 @@ void SubscriptionData::writeSettings() {
   SettingsHolder::instance()->setSubscriptionData(m_rawJson);
 }
 
+bool SubscriptionData::plusTax() {
+  MozillaVPN* vpn = MozillaVPN::instance();
+  QString countryCode = vpn->location()->countryCode();
```
bakulf, line 229: This is still wrong. The location should not be used here. As an European user i don't want to see +tax if i fly to Canada, right?

src/apps/vpn/models/subscriptiondata.cpp
```diff
@@ -222,6 +224,16 @@ void SubscriptionData::writeSettings() {
...
+
+  connect(vpn->location(), &Location::changed,
```
bakulf, line 231: This is exactly the bug I'm talking about. The +tax is based on the location at the subscription time.
Guardian should have this data.

bakulf: @brizental , in case Guardian does not have it can we use a global state prop set at the subscription time?

bakulf (changes requested):

src/apps/vpn/models/subscriptiondata.cpp
brizental: Yes we can. This feels like something Guardian should have though, if it doesn't maybe we should change it in Guardian to provide the "+ tax".

translations/strings.yaml
MattLichtenstein: "upsell" was probably a bad word choice here since this is unrelated to the relay bundle. Will replace with "promote"

translations/strings.yaml
MattLichtenstein: We use different verbiage in different parts of the app

<details>
<summary>Images</summary>

IMG_54B1CC17174D-1

Screen Shot 2023-02-22 at 10 18 14 AM

</details>

nebula/ui/components/VPNSubscriptionOption.qml
MattLichtenstein: returning nothing would result in a `Unable to assign [undefined] to QString`

src/apps/vpn/models/subscriptiondata.cpp
MattLichtenstein: This is only used before a subscription is purchased (on the IAP screen). What are you suggesting we use instead?

src/apps/vpn/ui/screens/settings/ViewSubscriptionManagement/ViewSubscriptionManagement.qml
birdsarah: No, fxa doesn't communicate currency. The app stores pick the currency for the user based on that user's signed in setting.

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeededIAP.qml
bakulf: remove that and add this task everywhere we spawn a TaskProducts in mozillavpn.cpp

bakulf (changes requested):

src/apps/vpn/mozillavpn.h
```diff
@@ -286,10 +286,12 @@ class MozillaVPN final : public QObject {
 
   bool checkCurrentDevice();
 
+  void scheduleTaskGetLocation();
+
  public slots:
   void requestAbout();
 
-  void scheduleRefreshDataTasks(bool refreshProducts);
+  Q_INVOKABLE void scheduleRefreshDataTasks(bool refreshProducts);
```
bakulf, line 294: Revert this

bakulf (approved):

src/apps/vpn/platforms/ios/iosiaphandler.mm
```diff
@@ -315,6 +315,7 @@ - (void)paymentQueueRestoreCompletedTransactionsFinished:(nonnull SKPaymentQueue
   productData->m_trialDays = discountDays;
   productData->m_monthlyPrice = monthlyPriceValue;
   productData->m_nonLocalizedMonthlyPrice = [monthlyPriceNS doubleValue];
+  productData->m_currencyCode = QString::fromNSString(product.priceLocale.currencyCode);
```
MattLichtenstein, line 318: Similar changes still need to be made to android iap handler

flodolo (approved):

PR #6074, opened 2023-02-17
Use 'en' country/city translations as final fallback - VPN-4201

bakulf:


src/apps/vpn/serveri18n.cpp
```diff
@@ -154,6 +154,13 @@ QString translateItemWithLanguage(const QString& languageCode,
     }
   }
 
+  // Let's use 'en' translation, which is still better than what we obtain from
+  // Guardian.
+  result = s_items.value(itemKey("en", countryCode, cityName));
+  if (!result.isEmpty()) {
```
brizental, line 160: If this is the final fallback how come it can still be empty?

bakulf: because we don't have the 'en' translation for all the city/countries.

brizental: What is wrong with what we obtain from Guardian?

bakulf: A lot of things are wrong. For instance they say: Osaka instead of Ōsaka.

brizental (approved):

src/apps/vpn/serveri18n.cpp
brizental: Should we file a follow up to fix that then?

bakulf: that is from Mullvad. There is nothing we can change...

PR #6081, opened 2023-02-17
VPN-4131 - Implement addon session state

brizental:
**This is a large PR, so I suggest going commit by commit during the review.**

One note about [DO NOT MERGE THIS ONE: Refactor split tunneling tutorial to use state](https://github.com/mozilla-mobile/mozilla-vpn-client/commit/afd6e095194e41a47762e146782ea20de5a5ecc2). Turns out we were too late to the party as this addon is actually disabled starting from 2.14... I added a commit with the refactor for illustration purposes, but we can't merge this change because this tutorial will not be enabled in versions that will have addon state support.

- [x] Oh I just realized there are leaks on the AddonSessionState. Gotta fix that before merging as well!

### DIFF
--- a/docs/add-on.md
+++ b/docs/add-on.md
@@ -65,7 +65,7 @@ Session state is _not_ persisted throughout user sessions i.e. it is wiped once 
 
 ### Local state
 
-> **Note**: Global state is not supported yet, if you need to implement an addon that requires it please refer to [VPN-3929](https://mozilla-hub.atlassian.net/browse/VPN-3929).
+> **Note**: Local state is not supported yet, if you need to implement an addon that requires it please refer to [VPN-3929](https://mozilla-hub.atlassian.net/browse/VPN-3929).
 
 Local state is persisted throughout user sessions i.e. not wiped after the application is killed. It is _not_ synced among the devices of a given user. It is local to each device.
 
--- a/docs/add-on.md
+++ b/docs/add-on.md
@@ -27,10 +27,54 @@ least a manifest.json file. The properties of this JSON file are:
 | name | The name of the add-on | String | Yes |
 | api_version | The version of the add-on framework | String | Yes |
 | type | One of the supported types (message, guide, tutorial, ...) | String | Yes |
-| conditions | List of conditions to met | Array of Condition objects | No |
+| conditions | List of conditions to meet | Array of Condition objects | No |
+| state | Object describing the state of the addon | Collection of state objects | No |
 
-Based on the add-on type, extra properties can be included. See the [tutorial](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/tutorials.md),
+Based on the add-on type, extra properties can be in added. See the [tutorial](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/tutorials.md),
 [guide](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/guides.md), and [message](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/message.md) documentation.
+
+## State Object
+
+An addon may have three different types of state: `session`, `local` or `global`.
+
+| Property | Description | Type | Required |
+| --- | --- | --- | --- |
+| session | An object describing the addons session state | State | No |
+| local | An object describing the addons local state | State | No |
+| global | An object describing the addons global state | State | No |
+
+The `State` interface is the same, regardless of the type of state.
+There is no limit of properties for a state, however the only supported state types are `string`, `number` and `boolean`.
+The state object may _not_ be nested.
+
+Following is the state interface, described using Typescript notation.
+
+```ts
+interface State {
+    [key: string]: {
+        type: “string” | “number” | “boolean”,
+        default: string | number | boolean,
+    }
+}
+
+```
+
+### Session state
+
+Session state is _not_ persisted throughout user sessions i.e. it is wiped once the application is killed.
+
+### Local state
+
+> **Note**: Global state is not supported yet, if you need to implement an addon that requires it please refer to [VPN-3929](https://mozilla-hub.atlassian.net/browse/VPN-3929).
+
+Local state is persisted throughout user sessions i.e. not wiped after the application is killed. It is _not_ synced among the devices of a given user. It is local to each device.
+
+### Global state
+
+> **Note**: Global state is not supported yet, if you need to implement an addon that requires it please refer to [VPN-2795](https://mozilla-hub.atlassian.net/browse/VPN-2795).
+
+Global state is persisted throughout user sessions and synced among user devices.
+
 
 ## Condition object
 
@@ -48,7 +92,7 @@ Add-ons can enable and disable themselves using the `conditions` key in the mani
 | trigger_time | A number identifying the number of seconds from the first execution of the client | Integer |  No | Yes |
 | start_time | The epoch time that activates the current add-on | Integer | No | Yes |
 | end_time | The epoch time that deactivates the current add-on | Integer | No | Yes |
-| javascript | A script file is executed to change the condition status. See below | String | No | Yes | 
+| javascript | A script file is executed to change the condition status. See below | String | No | Yes |
 | translation_threshold | The translation threshold to use. By default 1 (full translation required) | Number | No | No |
 
 Some conditions are dynamic. This means that the value can change their status dynamically during the app execution.
@@ -69,7 +113,7 @@ The list of setting keys can be found here: https://github.com/mozilla-mobile/mo
 
 ### Javascript conditions
 
-When the add-on manifest contains a `javascript` property in the `conditions` object, its value must be a javascript filename. 
+When the add-on manifest contains a `javascript` property in the `conditions` object, its value must be a javascript filename.
 
 The javascript file is executed when the add-on is loaded and it has to expose a function. For instance:
 
--- a/src/apps/vpn/addons/addon.cpp
+++ b/src/apps/vpn/addons/addon.cpp
@@ -382,7 +382,7 @@ Addon* Addon::create(QObject* parent, const QString& manifestFileName) {
     return nullptr;
   }
 
-  addon->m_state.setManifest(obj["state"].toObject());
+  addon->m_state = new AddonState(addon, obj["state"].toObject());
 
   QJsonObject javascript = obj["javascript"].toObject();
   if (!addon->evaluateJavascript(javascript)) {
--- a/src/apps/vpn/addons/addon.cpp
+++ b/src/apps/vpn/addons/addon.cpp
@@ -32,6 +32,7 @@
 #include "logger.h"
 #include "qmlengineholder.h"
 #include "settingsholder.h"
+#include "state/addonsessionstate.h"
 #include "telemetry/gleansample.h"
 #include "versionutils.h"
 
@@ -380,6 +381,8 @@ Addon* Addon::create(QObject* parent, const QString& manifestFileName) {
   if (!addon) {
     return nullptr;
   }
+
+  addon->m_state.setManifest(obj["state"].toObject());
 
   QJsonObject javascript = obj["javascript"].toObject();
   if (!addon->evaluateJavascript(javascript)) {
--- a/src/apps/vpn/addons/addon.cpp
+++ b/src/apps/vpn/addons/addon.cpp
@@ -423,49 +423,49 @@ Addon::Addon(QObject* parent, const QString& manifestFileName,
   SettingsHolder* settingsHolder = SettingsHolder::instance();
   Q_ASSERT(settingsHolder);
 
-  QString stateSetting = settingsHolder->getAddonSetting(StateQuery(id));
-  QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
+  QString statusSetting = settingsHolder->getAddonSetting(StatusQuery(id));
+  QMetaEnum statusMetaEnum = QMetaEnum::fromType<Status>();
 
-  bool isValidState = false;
-  int persistedState = stateMetaEnum.keyToValue(
-      stateSetting.toLocal8Bit().constData(), &isValidState);
+  bool isValidStatus = false;
+  int persistedStatus = statusMetaEnum.keyToValue(
+      statusSetting.toLocal8Bit().constData(), &isValidStatus);
 
-  if (isValidState) {
-    m_state = static_cast<State>(persistedState);
+  if (isValidStatus) {
+    m_status = static_cast<Status>(persistedStatus);
   }
 
-  if (m_state == Unknown) {
-    updateAddonState(Installed);
+  if (m_status == Unknown) {
+    updateAddonStatus(Installed);
   }
 }
 
 Addon::~Addon() { MZ_COUNT_DTOR(Addon); }
 
-void Addon::updateAddonState(State newState) {
-  Q_ASSERT(newState != Unknown);
+void Addon::updateAddonStatus(Status newStatus) {
+  Q_ASSERT(newStatus != Unknown);
 
-  if (m_state == newState) {
+  if (m_status == newStatus) {
     return;
   }
 
-  m_state = newState;
+  m_status = newStatus;
 
-  QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
-  QString newStateSetting = stateMetaEnum.valueToKey(newState);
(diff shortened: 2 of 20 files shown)


addons/tutorial_04_split_tunneling/vpnIsOn.js
```diff
@@ -1,22 +1,17 @@
...
+    }
 
-(function(api, condition) {
-api.connectSignal(api.addon, 'playingChanged', () => {
-  if (!api.addon.playing) {
-    return;
-  }
+    if (controller.state === controller.StateOff) {
+      condition.disable();
+      addon.state.session.set("splitTunnelTutorialVpnWasOn", false);
```
bakulf, line 9: Let's make this file back-compatible. Maybe with `if ("state" in addon) ...`

addons/tutorial_04_split_tunneling/vpnWasOn.js
```diff
@@ -1,15 +1,15 @@
...
+      return;
+    }
 
-  if (splitTunnelTutorialVpnWasOn) {
-    condition.enable();
-  } else {
-    condition.disable();
-  }
-});
+    if (addon.state.session.get("splitTunnelTutorialVpnWasOn")) {
```
bakulf, line 7: Same here.

docs/add-on.md
```diff
@@ -27,11 +27,55 @@ least a manifest.json file. The properties of this JSON file are:
...
+
+```
+
+### Session state
+
+Session state is _not_ persisted throughout user sessions i.e. it is wiped once the application is killed.
+
+### Local state
+
+> **Note**: Global state is not supported yet, if you need to implement an addon that requires it please refer to [VPN-3929](https://mozilla-hub.atlassian.net/browse/VPN-3929).
```
bakulf, line 68: Local?

src/apps/vpn/addons/addon.cpp
```diff
@@ -423,49 +426,49 @@ Addon::Addon(QObject* parent, const QString& manifestFileName,
   SettingsHolder* settingsHolder = SettingsHolder::instance();
   Q_ASSERT(settingsHolder);
 
-  QString stateSetting = settingsHolder->getAddonSetting(StateQuery(id));
-  QMetaEnum stateMetaEnum = QMetaEnum::fromType<State>();
+  QString statusSetting = settingsHolder->getAddonSetting(StatusQuery(id));
```
bakulf, line 429: Eventually all of this will be replaced with global states.

src/apps/vpn/addons/state/addonstate.h
```diff
@@ -0,0 +1,52 @@
...
+ * sessions.
+ *
+ */
+struct AddonState {
+  Q_GADGET
+  Q_PROPERTY(AddonSessionState* session READ session)
+
+ public:
+  AddonState() = default;
+  ~AddonState() { delete m_session; }
```
bakulf, line 27: Why a manual delete? Can you set the parent obj in m_session?

src/apps/vpn/addons/state/addonstate.h
```diff
@@ -0,0 +1,52 @@
...
+   * manifest.
+   */
+  void setManifest(const QJsonObject& manifest) {
+    m_session = AddonSessionState::fromManifest(manifest["session"].toObject());
+  }
+
+  AddonSessionState* session() const { return m_session; }
+
+ private:
+  AddonSessionState* m_session = nullptr;
```
bakulf, line 49: Maybe we don't need to allocate it. Just to avoid a nullptr exposed to the code by mistake.

src/apps/vpn/addons/state/addonstatebase.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return QJsonValue::String;
+  }
+
+  return QJsonValue::Undefined;
+}
+
+// static
+QHash<QString, QJsonValue> AddonStateBase::parseManifest(
+    const QJsonObject& manifest) {
+  QHash<QString, QJsonValue> spec;
```
bakulf, line 36: Here you are creating a hash table, then you populate it, and then you return a copy of it as return value. What about if you receive a reference as param to avoid the extra copy.

src/apps/vpn/addons/state/addonstatebase.cpp
```diff
@@ -0,0 +1,122 @@
...
+      continue;
+    }
+
+    spec.insert(key, def);
+  }
+
+  return spec;
+}
+
+AddonStateBase::AddonStateBase(QHash<QString, QJsonValue> spec)
```
bakulf, line 74: const &

src/apps/vpn/addons/state/addonstatebase.cpp
```diff
@@ -0,0 +1,122 @@
...
+                   << "to value of incorrect type. Ignoring.";
+    return;
+  }
+
+  setInternal(key, value);
+}
+
+void AddonStateBase::clear(const QString& key) {
+  if (key.isEmpty()) {
+    clearInternal();
```
bakulf, line 110: Return

src/apps/vpn/addons/state/addonstatebase.h
```diff
@@ -0,0 +1,88 @@
...
+ * values may be either boolean, numbers or strings. The keys are not dinamic.
+ * The values can be changed, but the kays are limited to whatever is defined in
+ * the addon manifest. The type of the value cannot be changed.
+ */
+class AddonStateBase : public QObject {
+  Q_OBJECT
+  Q_DISABLE_COPY_MOVE(AddonStateBase)
+
+ public:
+  ~AddonStateBase() = default;
```
bakulf, line 32: A CTOR with a parent?

src/apps/vpn/addons/state/addonstatebase.h
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * If the key has never been written to, this is a no-op.
+   *
+   * @param key The key to clear.
+   */
+  Q_INVOKABLE void clear(const QString& key = "");
+
+ protected:
+  AddonStateBase(StateHash spec);
+  StateHash m_defaults;
```
bakulf, line 75: Should this be private?

src/apps/vpn/addons/state/addonstatebase.cpp
```diff
@@ -0,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "addonstatebase.h"
+
+#include <QHash>
```
bakulf, line 7: Not needed

src/apps/vpn/addons/state/addonsessionstate.h
```diff
@@ -0,0 +1,56 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ADDONSESSIONSTATE_H
+#define ADDONSESSIONSTATE_H
+
+#include <QHash>
+#include <QJsonValue>
+#include <QObject>
+#include <QString>
```
bakulf, line 11: None of these headers are needed.

src/apps/vpn/addons/state/addonsessionstate.h
```diff
@@ -0,0 +1,56 @@
...
+#include "addonstatebase.h"
+
+/**
+ * @brief Addon State implementation,
+ * in which state does not persist throughout restarts.
+ */
+class AddonSessionState final : public AddonStateBase {
+ public:
+  ~AddonSessionState();
+  AddonSessionState(StateHash spec);
```
bakulf, line 22: A parent please :)

src/apps/vpn/addons/state/addonsessionstate.h
```diff
@@ -0,0 +1,56 @@
...
+   *
+   * Invalid properties on the manifest will be ignored. In the worst case if
+   * the manifest is completely wrong an empty state will be generated.
+   *
+   * @param manifest The addon's manifest.
+   * @return AddonState the generated state.
+   */
+  static AddonSessionState* fromManifest(const QJsonObject& manifest);
+
+ protected:
```
bakulf, line 47: Private.

src/apps/vpn/addons/state/addonsessionstate.cpp
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "addonsessionstate.h"
+
+#include <QHash>
+#include <QJsonValue>
+#include <QObject>
+#include <QString>
```
bakulf, line 10: Remove these headers

src/apps/vpn/addons/state/addonsessionstate.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "addonstatebase.h"
```
bakulf, line 12: Remove this too

src/apps/vpn/addons/state/addonsessionstate.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "leakdetector.h"
+
+// static
+AddonSessionState* AddonSessionState::fromManifest(
+    const QJsonObject& manifest) {
+  return new AddonSessionState(AddonStateBase::parseManifest(manifest));
+}
+
+AddonSessionState::AddonSessionState(QHash<QString, QJsonValue> spec)
```
bakulf, line 21: Const & + parent

src/apps/vpn/addons/state/addonsessionstate.cpp
```diff
@@ -0,0 +1,43 @@
...
+    : AddonStateBase(spec) {
+  MZ_COUNT_CTOR(AddonSessionState);
+}
+
+AddonSessionState::~AddonSessionState() { MZ_COUNT_DTOR(AddonSessionState); }
+
+QJsonValue AddonSessionState::getInternal(const QString& key) const {
+  return m_state[key];
```
bakulf, line 29: I wonder if m_state should be in the base class. You will need it also for the other implementations. The other impls will have a different propagation algorithm but m_state will probably remain.

tests/functional/testAddons.js
```diff
@@ -194,6 +194,9 @@ describe('Addons', function () {
 
     testCases.forEach(([_, shouldBeAvailable, testCase]) => {
       it(`message display is correct when subscription expiration ${testCase}`, async () => {
+        // This addon starts at version 2.15.
+        await vpn.setVersionOverride("2.15.0");
```
bakulf, line 198: Unrelated?

tests/unit/testaddonsessionstate.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "addons/state/addonsessionstate.h"
+
+void TestAddonSessionState::testGetAndSet() {
+  QJsonObject numProp;
+  numProp["type"] = "number";
+  numProp["default"] = 10;
+  QJsonObject manifest;
+  manifest["numProp"] = numProp;
+
+  AddonSessionState* state = AddonSessionState::fromManifest(manifest);
```
bakulf, line 16: Without a parent you leak this object.

bakulf (changes requested):
Wow! Well done! Just small comments about the parenting.

addons/tutorial_04_split_tunneling/vpnIsOn.js
brizental: We shouldn't merge this actually... I left this for illustration purposes, I am going to drop this commit before merging. This tutorial is disabled starting from v2.14, so I think there is no point.

docs/add-on.md
brizental: Oops, nice catch!

src/apps/vpn/addons/state/addonstate.h
brizental: Well, because of ignorance really :D Didn't realize that was the reason for setting the parent. Now that you say it, it makes sense :flushed:

src/apps/vpn/addons/state/addonstatebase.h
brizental: Just for testing purposes, no. I use it in the "SpyState" class I created for tests.

src/apps/vpn/addons/state/addonsessionstate.h
brizental: How do you know this? I always wonder haha

src/apps/vpn/addons/state/addonsessionstate.cpp
brizental: I had it in the bsae class initially. But I am not too sure it will be required for the other states. Like for local state, will we load from settings to memory on initialize? Or will we always just quuery the settings on `set`/`get`/`clear`? Ifwe do that then `m_state` is not necesssary. We can always move it back there if we realize all of them use it.

tests/functional/testAddons.js
brizental: Oh... Yes, unlreated. Oops.

tests/unit/testaddonsessionstate.cpp
brizental: Yes that is why I had the explicit `delete`... But I'll change that.

addons/tutorial_04_split_tunneling/vpnIsOn.js
brizental: Already discussed this offline with @bakulf and he agrees with the above comment :)

src/apps/vpn/addons/addon.cpp
```diff
@@ -357,19 +358,19 @@ Addon* Addon::create(QObject* parent, const QString& manifestFileName) {
   Addon* addon = nullptr;
 
   if (type == "i18n") {
-    addon = new AddonI18n(parent, manifestFileName, id, name);
```
bakulf, line 360: I don't like this too much. Because you are checking something here and then you use it elsewhere extracting the same keys, without checking them.

src/apps/vpn/addons/addon.cpp
```diff
@@ -412,60 +413,62 @@ Addon* Addon::create(QObject* parent, const QString& manifestFileName) {
...
+             const QJsonObject& manifest)
     : QObject(parent),
       m_manifestFileName(manifestFileName),
-      m_id(id),
-      m_name(name),
-      m_type(type) {
+      m_id(manifest["id"].toString()),
+      m_name(manifest["name"].toString()),
+      m_type(manifest["type"].toString()),
+      m_state(manifest["state"].toObject()) {
```
bakulf, line 422: For instance you don't check this object at all... This is fine, but inconsistent

src/apps/vpn/addons/addon.cpp
bakulf: What about if you create the addon state after this block? See the condition block. This way you can use addon as the parent.

bakulf (changes requested):
Nothing wrong. Let's talk about style. ;)

src/apps/vpn/addons/addon.cpp
brizental: Yes, I agree. I'll change it.

src/apps/vpn/addons/addon.h
```diff
@@ -56,6 +58,7 @@ class Addon : public QObject {
   const QString& id() const { return m_id; }
   const QString& type() const { return m_type; }
   const QString& manifestFileName() const { return m_manifestFileName; }
+  AddonState* state() const { return m_state; }
```
bakulf, line 61: Q_ASSERT(m_state);

src/apps/vpn/addons/state/addonstatebase.h
```diff
@@ -0,0 +1,86 @@
...
+ * practice, it is a dictionary of key/value pairs. The keys are strings and the
+ * values may be either boolean, numbers or strings. The keys are not dinamic.
+ * The values can be changed, but the kays are limited to whatever is defined in
+ * the addon manifest. The type of the value cannot be changed.
+ */
+class AddonStateBase {
+  Q_GADGET
+
+ public:
+  virtual ~AddonStateBase() {}
```
bakulf, line 30: = default;

bakulf (approved):

Issue #5395, opened 2023-01-04
Provide a build option for a "no glean" build


data-sync-user:
That would be useful in two use cases as of right now:

1. Linux builds that might not have access to submodules
1. WASM builds, which already do not include Glean because Glean Rust is not WASM compatible

Question: Is the data collection modal necessary in this type of build?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-3625)
┆Reporter: Beatriz Rizental Machado

data-sync-user:
➤ Owen Kirby commented:

To answer your question, the data collection modal doesn’t make sense in this build as we wouldn’t be collecting anything even when user accepts.

data-sync-user:
➤ Sebastian Streich commented:

The data collection modal would still indicate if sentry sessions can be started 😅  So that would still be valid.

PR #6083, opened 2023-02-17
VPN-3679 - Add missing python packages to linux/debian/control

brizental:
In the issue, the reporter says that without this packages the package fails to build from source. I added the packages, but I am skeptical it will be enough because we can't add glean-parser as a package there. WDYT @oskirby ?


oskirby:
@brizental I agree, it's kind of a mess and there is no good solution to the `glean_parser` conundrum.

A couple possible avenues of inquiry:
- A [glean-parser](https://packages.ubuntu.com/jammy/glean-parser) has existed since Ubuntu/22.04 (jammy) that we can use to satisfy this dependency. Unfortunately, that means we have to make do with a minimum version of 4.4.0, which predates rustlang support, and still leaves 20.04 (focal) and 18.04 (bionic) without viable build paths. Luckily for us, bionic is going to be deprecated in a month or two so that pain might not be as bad as we think.
- We follow through on #5395 ([VPN-3625](https://mozilla-hub.atlassian.net/browse/VPN-3625)) and provide a path to build without `glean_parser` at all.
- We start uploading our source build artifacts somewhere and direct users to consume those tarballs instead of downloading from git directly, since the tarballs will have the glean code prebuilt for them.

oskirby:
That being said, I have no objection adding `python3-jinja2` and `python3-click` as package dependencies, but they won't actually resolve the user's build problems.

oskirby (approved):

PR #6085, opened 2023-02-17
Fix VPN-4189 - UI issues in iOS disconnect warning

lesleyjanenorton:
## Description
This listens to VPNController.state and resets the warning height accordingly.

## Reference
VPN-4189

## Checklist
    
### DIFF
--- a/src/apps/vpn/ui/screens/settings/ViewPrivacy.qml
+++ b/src/apps/vpn/ui/screens/settings/ViewPrivacy.qml
@@ -44,6 +44,7 @@ VPNViewBase {
                     Accessible.name: text
                 }
                 Loader {
+                    id: disconnectWarningLoader
                     active: !VPNController.silentServerSwitchingSupported && VPNController.state !== VPNController.StateOff
                     Layout.fillWidth: true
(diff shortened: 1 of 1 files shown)


src/apps/vpn/ui/screens/settings/ViewPrivacy.qml
```diff
@@ -53,6 +54,12 @@ VPNViewBase {
                         Accessible.role: Accessible.StaticText
                         Accessible.name: text
                     }
+                    Connections {
+                        target: VPNController
+                        function onStateChanged() {
+                            disconnectWarningLoader.Layout.preferredHeight = disconnectWarningLoader.active ? disconnectWarningLoader.implicitHeight : 0;
+                        }
+                    }
```
MattLichtenstein, line 62: Hmmm, maybe putting this text block in a loader was a bad suggestion by me since this is kind of ugly

We can se `visible: active` on the loader, or just do away with the loader completely here. One day I'd like to profile if the former is actually any better

MattLichtenstein (changes requested):

MattLichtenstein (approved):

PR #6089, opened 2023-02-18
VPN-3423 Verify subscription before connecting

Gela:
## Description

This enhances the user experience when a subscription expires.

- Introduce a new state `StateCheckSubscription` which creates a network request to check user account data and make sure their subscription is still valid.
- Add a functional test which overrides the Guardian endpoint to mark a subscription as inactive __after__ the client has authenticated but before they toggle the VPN on. 
- Add a new `QmlQueryComposer` called `SUBSCRIPTION_NEEDED` in queries.js to enable testing this view (apparently we don't have any tests currently that test an expired subscription??)

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-3423

## Checklist
    
### DIFF
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -25,8 +25,8 @@
 #include "serveri18n.h"
 #include "serverlatency.h"
 #include "settingsholder.h"
-#include "tasks/function/taskfunction.h"
 #include "tasks/controlleraction/taskcontrolleraction.h"
+#include "tasks/function/taskfunction.h"
 #include "tasks/heartbeat/taskheartbeat.h"
 #include "taskscheduler.h"
 
@@ -221,11 +221,12 @@ bool Controller::activate(const ServerData& serverData,
       return true;
     }
 
-    // Before attempting to enable VPN connection we should check that the subscription is active.
+    // Before attempting to enable VPN connection we should check that the
+    // subscription is active.
     setState(StateCheckSubscription);
 
     // Set up a network request to check the subscription status.
-    // "task" is an empty task function which is being used to 
+    // "task" is an empty task function which is being used to
     // replicate the behavior of a TaskAccount.
     TaskFunction* task = new TaskFunction([]() {});
     NetworkRequest* request = new NetworkRequest(task, 200);
@@ -241,7 +242,7 @@ bool Controller::activate(const ServerData& serverData,
     connect(request, &NetworkRequest::requestCompleted, this,
             [](const QByteArray& data) {
               MozillaVPN::instance()->accountChecked(data);
-            });    
+            });
 
     setState(StateConnecting);
   }
@@ -572,7 +573,8 @@ void Controller::quit() {
   m_nextStep = Quit;
 
   if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting || m_state  == StateCheckSubscription) {
+      m_state == StateSilentSwitching || m_state == StateConnecting ||
+      m_state == StateCheckSubscription) {
     deactivate();
     return;
   }
@@ -589,8 +591,8 @@ void Controller::backendFailure() {
   m_nextStep = BackendFailure;
 
   if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting || m_state == StateCheckSubscription ||
-      m_state == StateConfirming) {
+      m_state == StateSilentSwitching || m_state == StateConnecting ||
+      m_state == StateCheckSubscription || m_state == StateConfirming) {
     deactivate();
     return;
   }
@@ -757,7 +759,7 @@ void Controller::statusUpdated(const QString& serverIpv4Gateway,
 
   list.swap(m_getStatusCallbacks);
   for (const std::function<void(
-           const QString&serverIpv4Gateway, const QString&deviceIpv4Address,
+           const QString& serverIpv4Gateway, const QString& deviceIpv4Address,
            uint64_t txBytes, uint64_t rxBytes)>&func : list) {
     func(serverIpv4Gateway, deviceIpv4Address, txBytes, rxBytes);
   }
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -236,7 +236,8 @@ bool Controller::activate(const ServerData& serverData,
     connect(request, &NetworkRequest::requestFailed, this,
             [](QNetworkReply::NetworkError error, const QByteArray&) {
               logger.error() << "Account request failed" << error;
-              REPORTNETWORKERROR(error, ErrorHandler::DoNotPropagateError, "PreActivationSubscriptionCheck");
+              REPORTNETWORKERROR(error, ErrorHandler::DoNotPropagateError,
+                                 "PreActivationSubscriptionCheck");
             });
 
     connect(request, &NetworkRequest::requestCompleted, this,
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -770,7 +770,7 @@ void Controller::statusUpdated(const QString& serverIpv4Gateway,
 
   list.swap(m_getStatusCallbacks);
   for (const std::function<void(
-           const QString& serverIpv4Gateway, const QString& deviceIpv4Address,
+           const QString&serverIpv4Gateway, const QString&deviceIpv4Address,
            uint64_t txBytes, uint64_t rxBytes)>&func : list) {
     func(serverIpv4Gateway, deviceIpv4Address, txBytes, rxBytes);
   }
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -238,6 +238,10 @@ bool Controller::activate(const ServerData& serverData,
               logger.error() << "Account request failed" << error;
               REPORTNETWORKERROR(error, ErrorHandler::DoNotPropagateError,
                                  "PreActivationSubscriptionCheck");
+
+              // Check if the error propagation has changed the Mozilla VPN
+              // state. Continue only if the user is still authenticated and
+              // subscribed.
               if (MozillaVPN::instance()->state() == MozillaVPN::StateMain) {
                 setState(StateConnecting);
               } else {
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -17,6 +17,7 @@
 #include "models/server.h"
 #include "models/servercountrymodel.h"
 #include "mozillavpn.h"
+#include "networkrequest.h"
 #include "rfc/rfc1112.h"
 #include "rfc/rfc1918.h"
 #include "rfc/rfc4193.h"
@@ -24,6 +25,8 @@
 #include "serveri18n.h"
 #include "serverlatency.h"
 #include "settingsholder.h"
+#include "task.h"
+#include "tasks/account/taskaccount.h"
 #include "tasks/controlleraction/taskcontrolleraction.h"
 #include "tasks/heartbeat/taskheartbeat.h"
 #include "taskscheduler.h"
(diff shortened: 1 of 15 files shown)


src/apps/vpn/controller.cpp
```diff
@@ -733,7 +759,7 @@ void Controller::statusUpdated(const QString& serverIpv4Gateway,
 
   list.swap(m_getStatusCallbacks);
   for (const std::function<void(
-           const QString&serverIpv4Gateway, const QString&deviceIpv4Address,
+           const QString& serverIpv4Gateway, const QString& deviceIpv4Address,
```
Gela, line 762: Automatic (and unrelated) change due to the Format hook

tests/functional/testSubscription.js
```diff
@@ -766,7 +843,7 @@ describe('Subscription view', function() {
               queries.screenSettings.subscriptionView.CANCELLED_LABEL.visible(),
               'text') === data.subscription.expected.label);
 
-      if (data.subscription.value._subscription_type == "web") {
+      if (data.subscription.value._subscription_type == 'web') {
```
Gela, line 846: Automatic (and unrelated) change due to the Format hook

tests/functional/testSubscription.js
```diff
@@ -1139,73 +1216,72 @@ describe('Subscription view', function() {
...
-          await vpn.flipFeatureOn('bundleUpgrade');
-        }
-
-        await clickSettingsIcon();
-        await openSubscriptionManagement();
-
-        await vpn.waitForQuery(queries.screenSettings.subscriptionView
-                                   .RELAY_UPSELL_STATUS.hidden());
-      });
+     async () => {
```
Gela, line 1219: Automatic (and unrelated) change due to the Format hook

src/apps/vpn/controller.h
```diff
@@ -186,6 +188,8 @@ class Controller final : public QObject {
   void serverDataChanged();
 
  private:
+  ErrorHandler::ErrorPropagationPolicy m_errorPropagationPolicy =
```
birdsarah, line 191: I don't think this is needed anymore.

src/apps/vpn/controller.cpp
```diff
@@ -219,6 +220,30 @@ bool Controller::activate(const ServerData& serverData,
...
+    // replicate the behavior of a TaskAccount.
+    TaskFunction* task = new TaskFunction([]() {});
+    NetworkRequest* request = new NetworkRequest(task, 200);
+    request->auth(MozillaVPN::authorizationHeader());
+    request->get(AppConstants::apiUrl(AppConstants::Account));
+
+    connect(request, &NetworkRequest::requestFailed, this,
+            [task, this](QNetworkReply::NetworkError error, const QByteArray&) {
+              logger.error() << "Account request failed" << error;
+              REPORTNETWORKERROR(error, m_errorPropagationPolicy, task->name());
```
birdsarah, line 239: I'm thinking this should be something like

```
REPORTNETWORKERROR(error, ErrorHandler::DoNotPropagateError, "PreActivationSubscriptionCheck");
```

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,82 @@ const SUBSCRIPTION_DETAILS = {
...
+
+    this.ctx.guardianOverrideEndpoints = {
+      GETs: {
+        '/api/v1/vpn/account': {
+          status: 200,
+          requiredHeaders: ['Authorization'],
+          body: userDataActive
+        }
+      },
+      POSTs: {
```
birdsarah, line 68: I think it's worth having a comment here about how the device data needs to match in all these cases.

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,82 @@ const SUBSCRIPTION_DETAILS = {
...
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].body =
+          userDataInactive;
+
+      // Step 3: Attempt to toggle the VPN on.
+      await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
+      await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
+
+      // Verify that user gets the "Subscribe to Mozilla VPN" screen.
+      await vpn.waitForQuery(queries.screenHome.SUBSCRIPTION_NEEDED.visible());
+    });
```
birdsarah, line 102: Please add a test that covers the case where the call to `/api/v1/vpn/account` fails. I think my expectation is that it does not affect the VPN continuing with activation.

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,82 @@ const SUBSCRIPTION_DETAILS = {
...
+      // "Subscribe to Mozilla VPN" screen.
+
+      // Step 1: User logs into the client successfully.
+      await vpn.authenticateInApp(true, true);
+
+      // Step 2: Override the Guardian endpoint to mock an expired subscription.
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].body =
+          userDataInactive;
+
+      // Step 3: Attempt to toggle the VPN on.
```
birdsarah, line 96: Let's use the vpn.activate() helper method unless you had a reason for not.

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,82 @@ const SUBSCRIPTION_DETAILS = {
...
+      }
+    };
+
+    it('Verify subscription before enabling VPN', async () => {
+      // This test verifies the case where a user is logged in
+      // but the VPN is off when their subscription expires.
+      // When they try to turn the VPN on, they get the
+      // "Subscribe to Mozilla VPN" screen.
+
+      // Step 1: User logs into the client successfully.
```
birdsarah, line 89: I love how beautifully commented this test is. 

However, the methods authenticateInApp and activate are very self explanatory. So I would only keep the Step 2 and 4 comments.

birdsarah (changes requested):

tests/functional/queries.js
```diff
@@ -176,7 +180,8 @@ const screenAuthenticationInApp = {
       new QmlQueryComposer('//authVerificationSessionByTotpNeeded-getHelpLink'),
   AUTH_TOTP_BACK_BUTTON:
       new QmlQueryComposer('//authVerificationSessionByTotpNeeded-backButton'),
-  AUTH_TOTP_PASTE_BUTTON: new QmlQueryComposer('//authVerificationSessionByTotpNeeded-inputPasteButton'),
+  AUTH_TOTP_PASTE_BUTTON: new QmlQueryComposer(
```
bakulf, line 183: Why this changes? We should use clang-format

bakulf (changes requested):
Well done. Consider deleting the task obj using the "destroyed" signal of the network request. And maybe move that block of code in a separate method.

tests/functional/testSubscription.js
Gela: `vpn.activate()` implementation explicitly checks that the VPN was enabled successfully (which doesn't/shouldn't happen in this case) so it ends up error-ing out if I use it here. Instead I need to explicitly attempt to turn the VPN on, then check that we land on the "Subscription required" screen.

tests/functional/queries.js
Gela: This is the result of running clang format on this file. I think it moved `//authVerificationSessionByTotpNeeded...` to a new line since it exceeds the line length

src/apps/vpn/controller.cpp
Gela: I think we should go back to `StateOff` but let me know if you disagree

src/apps/vpn/controller.cpp
```diff
@@ -219,7 +220,38 @@ bool Controller::activate(const ServerData& serverData,
...
+    NetworkRequest* request = new NetworkRequest(task, 200);
+    request->auth(MozillaVPN::authorizationHeader());
+    request->get(AppConstants::apiUrl(AppConstants::Account));
+
+    connect(request, &NetworkRequest::requestFailed, this,
+            [this](QNetworkReply::NetworkError error, const QByteArray&) {
+              logger.error() << "Account request failed" << error;
+              REPORTNETWORKERROR(error, ErrorHandler::DoNotPropagateError,
+                                 "PreActivationSubscriptionCheck");
+              if (MozillaVPN::instance()->state() == MozillaVPN::StateMain) {
```
bakulf, line 241: Write a comment saying that, we need to see if the error propagation has changed the Mozilla VPN state. We continue only if the user is still authenticated and subscribed.

tests/functional/queries.js
```diff
@@ -143,7 +144,8 @@ const screenAuthenticationInApp = {
   AUTH_SIGNIN_BUTTON: new QmlQueryComposer('//authSignIn-button'),
   AUTH_SIGNIN_GET_HELP_LINK: new QmlQueryComposer('//authSignIn-getHelpLink'),
   AUTH_SIGNIN_BACK_BUTTON: new QmlQueryComposer('//authSignIn-backButton'),
-  AUTH_SIGNIN_PASSWORD_PASTE_BUTTON: new QmlQueryComposer('//authSignIn-inputPasteButton'),
+  AUTH_SIGNIN_PASSWORD_PASTE_BUTTON:
```
bakulf, line 147: Are these changes from clang-format?

bakulf (approved):

tests/functional/queries.js
Gela: They are!

src/apps/vpn/controller.h
```diff
@@ -13,6 +13,7 @@
 #include <QTimer>
 #include <functional>
 
+#include "errorhandler.h"
```
birdsarah, line 16: Should this import be in the .cpp?

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,114 @@ const SUBSCRIPTION_DETAILS = {
...
+      // When they try to turn the VPN on, they get the
+      // "Subscribe to Mozilla VPN" screen.
+
+      await vpn.authenticateInApp(true, true);
+
+      // Step 1: Override the Guardian endpoint to mock an expired subscription.
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].body =
+          userDataInactive;
+
+      // Step 2: Attempt to toggle the VPN on. This should fail.
```
birdsarah, line 102: "This should fail" seems confusing. The next step contains the expectation.

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,114 @@ const SUBSCRIPTION_DETAILS = {
...
+      await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
+      await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
+
+      // Step 3: Verify that user gets the "Subscribe to Mozilla VPN" screen.
+      await vpn.waitForQuery(queries.screenHome.SUBSCRIPTION_NEEDED.visible());
+    });
+
+    it('Enable VPN with 500 error', async () => {
```
birdsarah, line 110:
```suggestion
    it('Continues to try connecting if call to check subscription status fails', async () => {
```

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,114 @@ const SUBSCRIPTION_DETAILS = {
...
+            device.name = req.body.name;
+            device.pubkey = req.body.pubkey;
+            device.unique_id = req.body.unique_id;
+          },
+          body: {}
+        },
+      }
+    };
+
+    it('Verify subscription before enabling VPN', async () => {
```
birdsarah, line 90:
```suggestion
    it('Prompts SubscriptionNeeded if a user's subscription is inactive on connection', async () => {
```
The test framework .describe( .it() ) are trying to nudge you to describe the expected behavior in the test name.

tests/functional/testSubscription.js
```diff
@@ -26,6 +27,114 @@ const SUBSCRIPTION_DETAILS = {
...
+
+      // Step 1: Override the Guardian endpoint to mock an expired subscription.
+      // Set the error status to 500.
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].body =
+          userDataInactive;
+      this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].status =
+          500;
+
+      await vpn.activate();
+      await vpn.waitForCondition(async () => {
```
birdsarah, line 130: I would prefer if this waitForCondition was for the `Connecting ...` text not the `VPN is on.` It's not expected that the VPN would successfully connect in this state.

birdsarah (approved):
Approve with some small changes requested.

tests/functional/testSubscription.js
Gela: From my conversation with @bakulf this morning, I believe the VPN should successfully connect in this case. Baku, can you confirm?

PR #6113, opened 2023-02-21
Chore - Print functional test logs if test didn't run at all

brizental:
Pulled out of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5892


tests/functional/setupVpn.js
```diff
@@ -150,7 +150,7 @@ exports.mochaHooks = {
 
   async afterEach() {
     // Collect errors on failure
-    if (this.currentTest.state === 'failed') {
+    if (!this.currentTest.state || this.currentTest.state === 'failed') {
```
brizental, line 153: If the test fails at setup no logs are printed because there is no state. This fixes it.

bakulf (approved):

PR #6130, opened 2023-02-22
vpn-4095 add ping

mcleinman:
## Description

Adds new ping. Have tested this by submitting the empty ping, can see on the debug ping viewer.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4095

## Checklist
    
### DIFF
--- a/qtglean/pings.yaml
+++ b/qtglean/pings.yaml
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # This file defines the pings that are recorded by the Glean SDK.
-# Their code APIs is automatically generated, at build time using,
+# Their code APIs is automatically generated at build time, using
 # the `glean_parser` PyPI package.
 
 ---
--- a/qtglean/pings.yaml
+++ b/qtglean/pings.yaml
@@ -1,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/2-0-0
+
+vpnsession:
+  description: |
+    Data for one VPN session (start/end time, session-based metrics, etc.)
+  include_client_id: true
+  send_if_empty: true
+  bugs:
+    - https://mozilla-hub.atlassian.net/browse/VPN-4095
+  data_reviews:
+    - N/A, will be done for metrics within
+  notification_emails:
+    - mcleinman@mozilla.com
--- a/qtglean/pings.yaml
+++ b/qtglean/pings.yaml
@@ -11,12 +11,15 @@ $schema: moz://mozilla.org/schemas/glean/pings/2-0-0
 
 vpnsession:
   description: |
-    Data for one VPN session (start/end time, session-based metrics, etc.)
+    Data for one VPN session, defined as a user turning on and eventually
+    turning off the VPN in the app. This ping will record start/end datestamp,
+    session-based metrics, etc. It will be sent after the user turns off the
+    VPN, and a end datestamp is recorded.
   include_client_id: true
-  send_if_empty: true
+  send_if_empty: false
   bugs:
     - https://mozilla-hub.atlassian.net/browse/VPN-4095
   data_reviews:
     - N/A, will be done for metrics within
   notification_emails:
-    - mcleinman@mozilla.com
+    - vpn@mozilla.com, mcleinman@mozilla.com
--- a/qtglean/glean_parser_ext/run_glean_parser.py
+++ b/qtglean/glean_parser_ext/run_glean_parser.py
@@ -116,7 +116,7 @@ if __name__ == "__main__":
 
         # Generate C++ header files
         for [ output, input ] in [
-            [os.path.join(args.outdir, "pings.h"), [os.path.join(yaml_files_path, "pings.yaml")]],
+            [os.path.join(args.outdir, "pings.h"), [os.path.join(yaml_files_path, "pings.yaml"), os.path.join(qtglean_yaml_files_path, "pings.yaml")]],
             [os.path.join(args.outdir, "metrics.h"), [os.path.join(yaml_files_path, "metrics.yaml"), os.path.join(qtglean_yaml_files_path, "metrics.yaml")]],
         ]:
             print("Generating {} from {}".format(output, input))
@@ -125,7 +125,7 @@ if __name__ == "__main__":
 
         # Generate C++ source files
         for [ output, input ] in [
-            [os.path.join(args.outdir, "pings.cpp"), [os.path.join(yaml_files_path, "pings.yaml")]],
+            [os.path.join(args.outdir, "pings.cpp"), [os.path.join(yaml_files_path, "pings.yaml"), os.path.join(qtglean_yaml_files_path, "pings.yaml")]],
             [os.path.join(args.outdir, "metrics.cpp"), [os.path.join(yaml_files_path, "metrics.yaml"), os.path.join(qtglean_yaml_files_path, "metrics.yaml")]],
         ]:
             print("Generating {} from {}".format(output, input))
@@ -134,7 +134,7 @@ if __name__ == "__main__":
 
         # Generate Rust files
         for [ output, input ] in [
-            [os.path.join(args.outdir, "pings.rs"), [os.path.join(yaml_files_path, "pings.yaml")]],
+            [os.path.join(args.outdir, "pings.rs"), [os.path.join(yaml_files_path, "pings.yaml"), os.path.join(qtglean_yaml_files_path, "pings.yaml")]],
             [os.path.join(args.outdir, "metrics.rs"), [os.path.join(yaml_files_path, "metrics.yaml"), os.path.join(qtglean_yaml_files_path, "metrics.yaml")]],
         ]:
             print("Generating {} from {}".format(output, input))


qtglean/pings.yaml
```diff
@@ -0,0 +1,22 @@
...
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/2-0-0
+
+vpnsession:
+  description: |
+    Data for one VPN session (start/end time, session-based metrics, etc.)
```
brizental, line 14: Please add to this description: 1. What do you mean by "VPN session" in practice; 2. What is the scheduling of this ping.

qtglean/pings.yaml
```diff
@@ -0,0 +1,22 @@
...
+  include_client_id: true
+  send_if_empty: true
```
brizental, line 16: Why?

qtglean/pings.yaml
```diff
@@ -0,0 +1,22 @@
...
+  bugs:
+    - https://mozilla-hub.atlassian.net/browse/VPN-4095
+  data_reviews:
+    - N/A, will be done for metrics within
```
brizental, line 20: AFAIK, new pings also require data-review.

qtglean/pings.yaml
```diff
@@ -0,0 +1,22 @@
...
+  notification_emails:
+    - mcleinman@mozilla.com
```
brizental, line 22: Also include vpn@mozilla.com here, just in case.

brizental (changes requested):
Are you planning on sending another PR under VPN-4095 where the ping scheduling is implemented? I would like to see that done as part of that ticket. It's hard to understand a ping without it's scheduling.

mcleinman:
# Request for data collection review form

1) What questions will you answer with this data?
This ping sets us up to add some metrics (which will have a different data review form): How long do users keep the VPN active for? Do they run into connectivity issues when they're using it?

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
We want basic data (how people use the VPN, and how often they run into issues) so that we can build a useful product for how people use it, and fix problems they have.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?
The alternative is continuing to not track this data. Right now, we don't have information like "how often do users switch VPN cities?" and "how long is the average VPN session?". This ping will allow us to add metrics to have rough guesses as to these fairly simple questions.

4) Can current instrumentation answer these questions?
No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Data_Collection) found on the Mozilla wiki.   
This request is simply for a ping, which in itself will collect no new data.

6) Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.
This collection will be documented in the Glean Dictionary at https://dictionary.telemetry.mozilla.org/

7) How long will this data be collected?  Choose one of the following:
* I want to permanently monitor this data. Matt Cleinman

8) What populations will you measure?
No filters - all channels, countries, and locales, unless the user has opted out of data collection on that device.

9) If this data collection is default on, what is the opt-out mechanism for users?
When launching the app for the first time, a user is given an option for whether to allow data collection. After this initial set up screen, a user can always toggle data collection permissions in the System Preferences screen.

10) Please provide a general description of how you will analyze this data.
Dashboards that product managers and others will consult on a regular basis.

11) Where do you intend to share the results of your analysis?
Within the Mozilla VPN team and Security and Privacy Products team.

12) Is there a third-party tool (i.e. not Glean or Telemetry) that you are proposing to use for this data collection? If so:
No

mcleinman:
Tagging @travis79 for data-review (my understanding is that a new ping requires data review)

qtglean/pings.yaml
mcleinman: Data review has been commenced, thanks.

qtglean/pings.yaml
mcleinman: This ping is going to be imperfect - it doesn't track a user turning the VPN on/off from mobile system settings (that is coming separately), and so there are situations where it may record 0 or 2+ start datestamps, when ideally it records just one. I initially set this as `true` as a safety valve - so that we can send it and try to unwind these sticky situations, if we wanted to.

In reality, the code is going to submit this right after we record an ending datestamp, and so it would effectively never actually be empty. And so I've updated this to `false`, the default.

qtglean/pings.yaml
mcleinman: Added, thanks.

qtglean/pings.yaml
mcleinman: Thanks, added.

travis79 (approved):
> # Request for data collection review form
> 
>     1. What questions will you answer with this data?
>        This ping sets us up to add some metrics (which will have a different data review form): How long do users keep the VPN active for? Do they run into connectivity issues when they're using it?
> 
>     2. Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
>        We want basic data (how people use the VPN, and how often they run into issues) so that we can build a useful product for how people use it, and fix problems they have.
> 
>     3. What alternative methods did you consider to answer these questions? Why were they not sufficient?
>        The alternative is continuing to not track this data. Right now, we don't have information like "how often do users switch VPN cities?" and "how long is the average VPN session?". This ping will allow us to add metrics to have rough guesses as to these fairly simple questions.
> 
>     4. Can current instrumentation answer these questions?
>        No.
> 
>     5. List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Data_Collection) found on the Mozilla wiki.
>        This request is simply for a ping, which in itself will collect no new data.
> 
>     6. Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.
>        This collection will be documented in the Glean Dictionary at [dictionary.telemetry.mozilla.org](https://dictionary.telemetry.mozilla.org/)
> 
>     7. How long will this data be collected?  Choose one of the following:
> 
> 
>     * I want to permanently monitor this data. Matt Cleinman
> 
> 
>     8. What populations will you measure?
>        No filters - all channels, countries, and locales, unless the user has opted out of data collection on that device.
> 
>     9. If this data collection is default on, what is the opt-out mechanism for users?
>        When launching the app for the first time, a user is given an option for whether to allow data collection. After this initial set up screen, a user can always toggle data collection permissions in the System Preferences screen.
> 
>     10. Please provide a general description of how you will analyze this data.
>         Dashboards that product managers and others will consult on a regular basis.
> 
>     11. Where do you intend to share the results of your analysis?
>         Within the Mozilla VPN team and Security and Privacy Products team.
> 
>     12. Is there a third-party tool (i.e. not Glean or Telemetry) that you are proposing to use for this data collection? If so:
>         No

# Data Review

1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way?

Yes, through the pings.yaml file and the [Glean Dictionary](https://dictionary.telemetry.mozilla.org/).

2) Is there a control mechanism that allows the user to turn the data collection on and off?

Yes, through the data preferences in the application settings.

3) If the request is for permanent data collection, is there someone who will monitor the data over time?

Pings don't automatically expire, so mcleinman and the VPN team will be responsible for this collection.

4) Using the **[category system of data types](https://wiki.mozilla.org/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Category 1 & 2

5) Is the data collection request for default-on or default-off?

Default-on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?

No

7) Is the data collection covered by the existing Firefox privacy notice?

Yes

8) Does the data collection use a third-party collection tool?

No

## Result

data-review+

qtglean/pings.yaml
```diff
@@ -0,0 +1,25 @@
...
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/2-0-0
+
+vpnsession:
+  description: |
+    Data for one VPN session, defined as a user turning on and eventually
+    turning off the VPN in the app. This ping will record start/end datestamp,
+    session-based metrics, etc. It will be sent after the user turns off the
```
brizental, line 16: > This ping will record start/end datestamp, session-based metrics, etc.

This is unecessary. The metrics added to the ping will be declared in the metrics.yaml file. If that ever changes we'd have to update here and it's likely we will forget and this will be stale.

qtglean/pings.yaml
```diff
@@ -0,0 +1,25 @@
...
+    VPN, and a end datestamp is recorded.
```
brizental, line 17:
```suggestion
    VPN.
```

qtglean/pings.yaml
brizental: > It will be sent after the user turns off the

I would like to understand the reasoning behind this scheduling choice. We had discussed a different design, previously along with @birdsarah. 

What if a user has an extremely long session? We would not hear from them for a long time and maybe be at risk of hitting Glean's 1MB ping payload limit. What if the user turns the VPN off by force killing the app, then this ping is just not sent? I think w need more triggers to this submit this ping. Please take a look at the `reasons` option for pings, we may want to use that to have a way of differenting pings sent at different time queues.

mcleinman: The scheduling was requested by Yeon Joo, who is doing the data engineering on this. Having to calculate the session length (as well as a couple other in-session metrics, like "was there an connection issue") she asked if we'd be able to package these session metrics all up in one ping.

We do *not* expect that this architecture captures all corner cases, and certainly this would miss an extremely long session whose payload was over 1MB*. Santiago is comfortable with this, as we believe it captures the vast majority of VPN usage, and the goal of the metrics is to be able to better understand user behavior to make decisions for the product. Additionally, I believe this architecture includes enough trip wires for us to discover if we're missing significant user behavior with this current approach (for instance, if the custom ping is missing a "session started" metric or contains multiple "session started" metrics, it implies something a little off), and thus we can change the approach down the line if we find that we're missing more than the trivial amount of user behavior we expect to.

If I'm understanding the thoughts behind your comment, it seems like you may want as precise an accounting as possible, for all situations. If that's true, then perhaps we should have a meeting w/ Santiago, as I agree with your conclusions of some missed corner cases, it's just something that we found acceptable for now (unless/until we get data to show it's a larger portion than we expect).

*Notably, this ping is *separate* from the work that will be done to capture session start/end at the daemon (or network extension layer), which is https://mozilla-hub.atlassian.net/browse/VPN-4038. This will serve as a secondary check to see roughly how many start/ends we're missing with the current planned approach, including those that start/end from mobile system settings.

brizental: > Santiago is comfortable with this, as we believe it captures the vast majority of VPN usage, and the goal of the metrics is to be able to better understand user behavior to make decisions for the product.

The thing is, we can't be sure of this. We don't have data to back the assumption.

---

 I am missing a request for comments on this. I would like to see a proposal / tech spec / document describing the scheduling of this ping and the reasoning for it. Specifically describing what were the discarded options and why they were discarded and documenting the discussions that happened around this design.

This ping, as I understand it, will be the transport for most of our user and health metrics. It is worth spending some time dwelling on its scheduling before moving ahead with implementation, to make sure all stakeholders agree on the design. That ideally would include engineering, product and data science because we will all consume the data collected by it.

See for example the discussion on https://mozilla-hub.atlassian.net/browse/VPN-3277 for an illustration of what I have in mind when I say I think this will be the transport of health metrics as well.

If any of my assumptions are not correct, then none of this is necessary. Let me know.

Note: this doesn't need to block this PR since it doesn't implement scheduling.

mcleinman: >>    Santiago is comfortable with this, as we believe it captures the vast majority of VPN usage, and the goal of the metrics is to be able to better understand user behavior to make decisions for the product.
> The thing is, we can't be sure of this. We don't have data to back the assumption.
I fully agree that we don't have data to back the assumption, but we won't have any data to know if this assumption (that several of us share, including product, who is the one requesting these metrics) is correct until we have some data coming in - it seems like a catch-22. This data work is being done with the assumption that we'll learn things that will lead us to soon do additional instrumentation, and this hunch feels like something that can be tested in the initial go of metrics.

----

I went back and found the source of "is it okay if we miss some corner cases, but track if it's a substantial amount", and it was a DM w/ Santiago about a month ago. The first public mention of this would have been the [related ticket](https://mozilla-hub.atlassian.net/browse/VPN-4095), created 3 weeks ago. 

While this has been approved by data science and product, I suppose the only engineer approval was mine, and I should have gotten wider buy in. I'll put this on hold, and write up a document for us to discuss.

I'm going to hold this PR on a resolution on this - because if we want to change this architecture, it's very likely that this ping becomes unnecessary. This ping may exist so that the VPN session start and VPN session end can be in the same ping (with the exception of corner cases), per request of data. If we're going to rearchitect it so that they may not be in the same custom ping, we likely don't need a custom ping at all.

qtglean/pings.yaml
mcleinman: The plan is to send it right after the end datestamp is recorded, so this should be accurate.

mcleinman:
@birdsarah From discussion w/ Adrienne, we're not going to hold this PR on Bea's return. Would you be able to give it a review?

mcleinman:
Adding @oskirby as a reviewer - Sarah is out on leave next week and finishing up a bunch of stuff today. Bea is out for a bit. Per comment above, we're going to move forward with this work even in her absence.

qtglean/pings.yaml
```diff
@@ -0,0 +1,25 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
```
Gela, line 6: nitpick:
```suggestion
# Their code APIs is automatically generated at build time, using
```

Gela (approved):

qtglean/pings.yaml
mcleinman: Thanks, good catch. (I did a copy paste from the other `pings.yml` file, and should have looked over the comments more carefully.)

PR #6133, opened 2023-02-23
VPN-4195 Improve Multi-Hop OS notifications

Gela:
## Description

Add a new string in strings.yaml to include both exit and entry server locations in the multihop VPN connected notification. 
Format: “Connected to X, through Y”. 

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4195

## Checklist
    
### DIFF
--- a/translations/strings.yaml
+++ b/translations/strings.yaml
@@ -273,7 +273,7 @@ notifications:
     comment: "%1 is the name of the city the VPN is connected to."
   VPNMultihopConnectedMessage:
     value: "Connected to %1, through %2"
-    comment: "%1 is the name of the entry server, %2 is the name of the exit server."
+    comment: "%1 is the name of the exit server, %2 is the name of the entry server."
   VPNDisconnectedTitle: VPN Disconnected
   VPNDisconnectedMessage:
     value: "Disconnected from %1"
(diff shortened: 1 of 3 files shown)


birdsarah:
@Gela it seems that you have your .js files setup for 4 spaces not 2 and that's causing a lot of the changes here. Can you unwind those?

birdsarah (commented):
Happy to approve once we revert the changes to testServerSearch

translations/strings.yaml
```diff
@@ -271,6 +271,9 @@ notifications:
   VPNConnectedMessage:
     value: "Connected to %1"
     comment: "%1 is the name of the city the VPN is connected to."
+  VPNMultihopConnectedMessage:
+    value: "Connected to %1, through %2"
+    comment: "%1 is the name of the entry server, %2 is the name of the exit server."
```
flodolo, line 276: On second thoughts, is this comment correct? 

I think the variables are swapped in the comment, "through %2" means that's the hop server, doesn't it?

brizental (approved):

translations/strings.yaml
Gela: Yes, nice catch. Updated the comment, thanks

flodolo (approved):

PR #6134, opened 2023-02-23
Set the authorize header in purchase requests - VPN-4204

bakulf:


src/apps/vpn/tasks/purchase/taskpurchase.cpp
```diff
@@ -56,6 +57,7 @@ TaskPurchase::~TaskPurchase() { MZ_COUNT_DTOR(TaskPurchase); }
 
 void TaskPurchase::run() {
   NetworkRequest* request = new NetworkRequest(this, 200);
+  request->auth(MozillaVPN::authorizationHeader());
```
brizental, line 60: Doe we need to add this to the functional test mocks as well?

bakulf: we do not have tests for the android/ios purchase...

brizental (approved):

PR #6141, opened 2023-02-23
VPN-4214 - Fix AuthenticationInBrowser test.

birdsarah:
After debugging with @mozrokafor we honed in on the fact that the mocked call to the DesktopAuthenticationListener is failing.

The DesktopAuthenticationListener is listening on QHostAddress::LocalHost. Equivalent to QHostAddress("127.0.0.1"). https://doc.qt.io/qt-6.2/qhostaddress.html#SpecialAddress-enum

I don't have confirmation of this, but I'm wondering if the Github Actions CI images changed and stopped routing "localhost" (that the test was using) to "127.0.0.1" and that's why the test suddenly broke.

### DIFF
--- a/tests/functional/helper.js
+++ b/tests/functional/helper.js
@@ -325,7 +325,7 @@ module.exports = {
       const urlObj = new URL(url);
 
       const options = {
-        hostname: urlObj.hostname,
+        hostname: '127.0.0.1',
         port: parseInt(urlObj.searchParams.get('port'), 10),
         path: '/?code=the_code',
         method: 'GET',
--- a/tests/functional/helper.js
+++ b/tests/functional/helper.js
@@ -335,9 +335,9 @@ module.exports = {
         const req = http.request(options, res => {});
         req.on('close', resolve);
         req.on('error', error => {
-          throw new error(
+          throw new Error(
               `Unable to connect to ${urlObj.hostname} to complete the
-              auth`);
+              auth. ${error.name}, ${error.message}, ${error.stack}`);
         });
         req.end();
       });


bakulf (commented):
This does not work for mobile env, where the services run on a different device.

birdsarah:
> This does not work for mobile env, where the services run on a different device.

Does this really apply. What we're doing in that line of code is making a callback to the DesktopAuthenticationListener.

My guess as to why this started failing is that GH actions has updated their CI images so localhost doesn't route to 127.0.0.1.

The DesktopAuthenticationListener is listening on QHostAddress::LocalHost. Equivalent to QHostAddress("127.0.0.1").

birdsarah:
> This does not work for mobile env, where the services run on a different device.

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/79bfbd643412d23eaa8633345a0b3ceed1e2ac7b/src/apps/vpn/authenticationlistener.cpp#L30

We don't do DesktopAuthenticationListener or anything equivalent on mobile.

mozrokafor (approved):

birdsarah:
Doing `cat /etc/hosts` on a GH Actions CI runner returns 

```
cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
10.1.0.27 fv-az283-835.kpokbfqxctbevcsxt0cqdblead.jx.internal.cloudapp.net fv-az283-835
```

We can see that both ipv4 and ipv6 are mapped to localhost.

birdsarah:
I can replicate this bug locally if I switch my `/etc/hosts` to match the one observed in the runner.

The proper fix for this is to make the DesktopAuthenticationListener respond to both ipv4 and ipv6 requests.

PR #6144, opened 2023-02-24
VPN-4214 Add comments to testAuthenticationInBrowser.js

birdsarah:
After discussing further with @bakulf we've agreed that the original fix in #6141 is correct and this PR now is just extra comments to help make this clear for our future selves.

====
# Old Description

In #6141 we hard-coded 127.0.0.1 to get the AuthenticationInBrowser.js test working again:

```patch
-        hostname: urlObj.hostname,
+        hostname: '127.0.0.1',
```

This PR reverses the above change and updates the `DesktopAuthenticationListener`.

In [d553d3f](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6144/commits/d553d3fabab9e7d43dafb084f31d4912612e8be4) I update `DesktopAuthenticaitonListener` to be listening on `QHostAddress::Any` which is, both IPv4 and IPv6 addresses https://doc.qt.io/qt-6.2/qhostaddress.html#SpecialAddress-enum. @oskirby posed the question "is this secure?" to me in slack. I'm not sure....I'm not exactly sure what `QHostAddress::Any` but I think it's "0.0.0.0" which seems to me like we don't want to open a server on. 

In ebfc80759 I update this so that we don't initialize the ReplyHandler with a host and I call listen to the two localhost addresses. 

I had a hard time finding documentation on `QOAuthHttpServerReplyHandler`, so here's the source: https://code.qt.io/cgit/qt/qtnetworkauth.git/tree/src/oauth/qoauthhttpserverreplyhandler.cpp?h=6.2.4

===

Some more info about the bug. I initially was able to repro on my machine by updated /etc/hosts. But that stopped working for me so I've only verified this fix in CI. 

I'll also note that in [this run](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/4258689859/jobs/7410278691#step:8:9) I changed the `/etc/hosts` file and not the `DesktopAuthenticationListener` and was also able to get the tests to pass continuing to point to some kind of local DNS issue.

### DIFF
--- a/src/apps/vpn/tasks/authenticate/desktopauthenticationlistener.cpp
+++ b/src/apps/vpn/tasks/authenticate/desktopauthenticationlistener.cpp
@@ -21,7 +21,7 @@ DesktopAuthenticationListener::DesktopAuthenticationListener(QObject* parent)
     : AuthenticationListener(parent) {
   MZ_COUNT_CTOR(DesktopAuthenticationListener);
 
-  m_server = new QOAuthHttpServerReplyHandler(QHostAddress::Any, this);
+  m_server = new QOAuthHttpServerReplyHandler(this);
   connect(m_server, &QAbstractOAuthReplyHandler::callbackReceived, this,
           [this](const QVariantMap& values) {
             logger.debug() << "DesktopAuthenticationListener data received";
@@ -53,7 +53,8 @@ void DesktopAuthenticationListener::start(Task* task,
                                    emailAddress));
 
   if (!m_server->isListening()) {
-    m_server->listen(QHostAddress::Any);
+    m_server->listen(QHostAddress::LocalHost);
+    m_server->listen(QHostAddress::LocalHostIPv6);
   }
 
   if (!m_server->isListening()) {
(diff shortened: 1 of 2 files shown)


src/apps/vpn/tasks/authenticate/desktopauthenticationlistener.cpp
```diff
@@ -21,7 +21,7 @@ DesktopAuthenticationListener::DesktopAuthenticationListener(QObject* parent)
     : AuthenticationListener(parent) {
   MZ_COUNT_CTOR(DesktopAuthenticationListener);
 
-  m_server = new QOAuthHttpServerReplyHandler(QHostAddress::LocalHost, this);
+  m_server = new QOAuthHttpServerReplyHandler(QHostAddress::Any, this);
```
birdsarah, line 24: Note that this isn't any different from just doing 

```
m_server = new QOAuthHttpServerReplyHandler(this);
```

tests/functional/helper.js
```diff
@@ -320,12 +320,14 @@ module.exports = {
...
-      // are mocking everything.
+      // are mocking everything. So this next chunk of code manually
+      // makes a call to the DesktopAuthenticationListener to mock
+      // a successful authentication in browser.
       const url = await this.getLastUrl();
       const urlObj = new URL(url);
 
       const options = {
-        hostname: '127.0.0.1',
+        hostname: urlObj.hostname,
```
birdsarah, line 330: Why not hardcode this as `localhost`? Maybe I don't understand `lastUrl` but it is probably a guardian url and that doesn't necessarily have the same host as the `DesktopAuthenticationListener.`

bakulf (approved):

PR #6145, opened 2023-02-24
Test Chore: updating asserts to provide actual values

mozrokafor:
## Description

   Currently, we have numerous asserts that check whether a condition is true or false. Updating these asserts to provide actual values vs expected values for debugging

### DIFF
--- a/tests/functional/testAuthenticationInBrowser.js
+++ b/tests/functional/testAuthenticationInBrowser.js
@@ -33,7 +33,7 @@ describe('User authentication in browser', function() {
     await vpn.waitForQuery(queries.screenInitialize.SIGN_UP_BUTTON.visible());
   });
 
-  it('Completes authentication', async () => {
+  it.only('Completes authentication', async () => {
     await vpn.waitForInitialView();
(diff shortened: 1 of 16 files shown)


tests/functional/testAuthenticationInBrowser.js
```diff
@@ -33,7 +33,7 @@ describe('User authentication in browser', function() {
     await vpn.waitForQuery(queries.screenInitialize.SIGN_UP_BUTTON.visible());
   });
 
-  it('Completes authentication', async () => {
+  it.only('Completes authentication', async () => {
```
bakulf, line 36: Revert this

bakulf (changes requested):

tests/functional/testAuthenticationInBrowser.js
mozrokafor: ah great catch, was doing to debugging previously, and this slipped through

bakulf (approved):

oskirby:
Thank you!

oskirby (approved):

PR #6152, opened 2023-02-24
VPN-2188 - Switch iOS build to cmake

brizental:
Draft because this works on my machine, but I am not sure it will work on CI.

If it works I'll go ahead with abolishing qmake and moving all iOS build scripts to cmake 🤞


glean/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 add_library(glean STATIC)
 
-if(NOT MSVC)
+if(NOT MSVC AND NOT XCODE)
```
bakulf, line 7: oh... why? XCODE means macos too.

bakulf (changes requested):

glean/CMakeLists.txt
brizental: Yes, yes. Right now I just want to see if this works on CI, the code is not ready for review and yes, this will need to change  to iOS only.

taskcluster/scripts/build/ios_build_no_adjust.sh
```diff
@@ -58,7 +58,7 @@ print Y "Configuring the build..."
 cat > xcode.xconfig << EOF
 APP_ID_MACOS = org.mozilla.macos.FirefoxVPN
 LOGIN_ID_MACOS = org.mozilla.macos.FirefoxVPN.login-item
-GROUP_ID_IOS = group.org.mozilla.ios.Guardian
+GROUP_ID_IOS = group.org.moziπmnlla.ios.Guardian
```
brizental, line 61: LOL, wtf

ci_scripts/ci_post_clone.sh
```diff
@@ -68,9 +31,48 @@ APP_ID_IOS = org.mozilla.ios.FirefoxVPN
 NETEXT_ID_IOS = org.mozilla.ios.FirefoxVPN.network-extension
 EOF
 
-if [ $CI_PRODUCT_PLATFORM == 'macOS' ]
-then
-  ./scripts/macos/apple_compile.sh macos
-else
-  ./scripts/macos/apple_compile.sh ios -q /Volumes/workspace/repository/qt_ios/$QTVERSION/macos/bin
-fi
+# TODO: REVERT HARDCODED LINK HERE
```
brizental, line 34: I'm going to send a follow up for this. We need to merge as is, and then on ce the toolchain job runs on CI we can change.

PR #6153, opened 2023-02-24
VPN-3929 - Implement local state

brizental:
Opening as a draft to get feedback on the implementation strategy. Will add tests before merging.


src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+      LocalState::getStorageIdentifierForKey(m_storageIdentifier, key), value);
+}
+
+void LocalState::clearInternal(const QString& key) {
+  settings()->beginGroup(s_localStateStorageIdentifier);
+
+  if (key.isEmpty()) {
+    settings()->remove(m_storageIdentifier);
+  } else {
+    settings()->remove(key);
```
brizental, line 66: This is missing an extra `settings()->beginGroup(m_storageIdentifier);`.

src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "settingsholder.h"
+
+namespace {
+Logger logger("LocalState");
+
+// Stores a list of identifiers for LocalState instances created during this
+// application run. Creating two that point to the same storage space is
+// likely a bug and will generated confusion from two places of the
+// code writing and reading from the same place in storage.
+QList<QString> s_registeredStorageIdentifiers;
```
bakulf, line 20: QStringList

src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+// static
+QString LocalState::getStorageIdentifierForKey(const QString& storageIdentifier,
+                                               const QString& key) {
+  return QString("%1/%2/%3")
+      .arg(s_localStateStorageIdentifier, storageIdentifier, key);
+}
+
+LocalState::LocalState(const QString& storageIdentifier,
+                       const QJsonObject& initialState)
```
bakulf, line 34: if this is a shared component, we should not ask to have a QJsonObject here. This is a implementation detail for addons.

src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+    : StateBase(initialState), m_storageIdentifier(storageIdentifier) {
+  if (s_registeredStorageIdentifiers.contains(storageIdentifier)) {
+    logger.warning() << "Creating a new LocalState instance with the same "
```
bakulf, line 37: maybe we should crash...

src/shared/state/statebase.cpp
```diff
@@ -30,35 +30,35 @@ QJsonValue::Type AddonStateBase::typeToQJsonValueType(QString type) {
 }
 
 // static
-StateHash AddonStateBase::parseManifest(const QJsonObject& manifest) {
+StateHash StateBase::parseAddonManifest(const QJsonObject& initialState) {
```
bakulf, line 33: well, this is wrong, right? Why having a "parseAddonManifest" if addons are not shared?

Maybe move it in Addon.

tests/unit/CMakeLists.txt
```diff
@@ -314,10 +314,10 @@ target_sources(unit_tests PRIVATE
     testaddonapi.h
     testaddonindex.cpp
     testaddonindex.h
-    testaddonstatebase.cpp
-    testaddonstatebase.h
-    testaddonsessionstate.cpp
-    testaddonsessionstate.h
+    teststatebase.cpp
```
bakulf, line 317: this should be moved in apps/unit_tests

bakulf (changes requested):

src/shared/state/statebase.cpp
brizental: Agree on moving it to addons.

brizental: Ok, so moving this to addons would require the validation code to move to addons as well and I did not like that. I instead just changed the wording on the doc comments and function names to not refer to addons specifically. Let me know what you think.

src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+Logger logger("LocalState");
+
+// Stores a list of identifiers for LocalState instances created during this
+// application run. Creating two that point to the same storage space is
+// likely a bug and will generated confusion from two places of the
+// code writing and reading from the same place in storage.
+QList<QString> s_registeredStorageIdentifiers;
+// The identifier of the settings group that will contain
+// ALL of the persisted state.
+const char* s_localStateStorageIdentifier = "mzLocalState";
```
bakulf, line 23: `constexpr const char * LOCALSTATE_STORAGE_IDENTIFIER = "mzLocalState";`

src/shared/state/localstate.cpp
```diff
@@ -0,0 +1,70 @@
...
+    : StateBase(initialState), m_storageIdentifier(storageIdentifier) {
+  if (s_registeredStorageIdentifiers.contains(storageIdentifier)) {
+    logger.warning() << "Creating a new LocalState instance with the same "
+                        "storageIdentifier as an existing LocalState "
+                        "instance. That is likely a bad idea.";
+  } else {
+    s_registeredStorageIdentifiers.append(storageIdentifier);
+  }
+
+  MZ_COUNT_CTOR(LocalState);
```
bakulf, line 44: 1. move this as first line of the CTOR.
2. `if (contains()) { warning(); return; }`

src/shared/state/localstate.h
```diff
@@ -0,0 +1,57 @@
...
+  void setInternal(const QString& key, const QJsonValue& value) override;
+  void clearInternal(const QString& key = "") override;
+
+  /**
+   * @brief Gets the settings from the settings holder. This class is a friend
+   * of that class, so it can access the internal QSettings directly.
+   *
+   * @return QSettings The application's internal settings.
+   */
+  static QSettings* settings() {
```
bakulf, line 47: Why do we need this? Can we remove it?
If by mistake we make it public, we expose the underlying QSettings obj to everybody.

src/shared/state/localstate.h
```diff
@@ -0,0 +1,57 @@
...
+    return &SettingsHolder::instance()->m_settings;
+  }
+
+  static QString getStorageIdentifierForKey(const QString& storageIdentifier,
+                                            const QString& key);
+
+  QString m_storageIdentifier;
```
bakulf, line 54: const QString

src/shared/state/sessionstate.cpp
```diff
@@ -0,0 +1,30 @@
...
+  return m_state[key];
+}
+
+void SessionState::setInternal(const QString& key, const QJsonValue& value) {
+  m_state[key] = value;
+}
+
+void SessionState::clearInternal(const QString& key) {
+  if (key.isEmpty()) {
+    m_state.clear();
```
bakulf, line 26: return :)

src/shared/state/sessionstate.h
```diff
@@ -0,0 +1,33 @@
...
+ public:
+  ~SessionState();
+
+  /**
+   * @brief Construct an SessionState object from a JSON object.
+   *
+   * @param initialState A JSON object specifying the state schema.
+   * @return State the generated state.
+   */
+  SessionState(const QJsonObject& initialState);
```
bakulf, line 24: explicit

src/shared/state/sessionstate.h
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ADDONSESSIONSTATE_H
```
bakulf, line 5: SESSIONSTATE_H

src/shared/state/localstate.h
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ADDONLOCALSTATE_H
```
bakulf, line 5: LOCALSTATE_H

src/shared/state/state.h
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ADDONSTATE_H
+#define ADDONSTATE_H
```
bakulf, line 6: STATE_H

src/shared/state/state.h
```diff
@@ -0,0 +1,43 @@
...
+struct State : public QObject {
+  Q_OBJECT
+  Q_PROPERTY(SessionState* session READ session)
+  Q_PROPERTY(LocalState* local READ local)
+
+ public:
+  State(QObject* parent, const SessionState& session, const LocalState& local)
+      : QObject(parent), m_session(session), m_local(local) {}
+
+  ~State() = default;
```
bakulf, line 33: `virtual ~State`

src/shared/state/state.h
```diff
@@ -0,0 +1,43 @@
...
+
+  SessionState* session() { return &m_session; }
```
bakulf, line 35: const ?

src/shared/state/state.h
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef ADDONSTATE_H
+#define ADDONSTATE_H
+
+#include <QJsonObject>
```
bakulf, line 8: you don't need this.

src/shared/state/sessionstate.h
```diff
@@ -0,0 +1,33 @@
...
+#define ADDONSESSIONSTATE_H
+
+#include "statebase.h"
+
+/**
+ * @brief Session State implementation,
+ * in which state does not persist throughout restarts.
+ */
```
bakulf, line 13: `class QJsonObject;`

src/shared/state/statebase.h
```diff
@@ -0,0 +1,109 @@
...
+  StateBase(const StateHash& spec) : m_defaults(spec) {}
+  StateHash m_defaults;
+
+  // Methods to override.
+  virtual QJsonValue getInternal(const QString& key) const = 0;
+  virtual void setInternal(const QString& key, const QJsonValue& value) = 0;
+  virtual void clearInternal(const QString& key = "") = 0;
+
+ private:
+  static QJsonValue::Type typeToQJsonValueType(QString type);
```
bakulf, line 84: const QString&

bakulf (changes requested):

src/shared/state/statebase.cpp
```diff
@@ -69,51 +69,60 @@ StateHash AddonStateBase::parseManifest(const QJsonObject& manifest) {
...
+  if (value.type() == m_defaults[key].type()) {
+    return value;
+  }
+
+  if (value.type() != QJsonValue::Null) {
+    logger.error()
+        << "Attempted to get key" << key
+        << "from state, but the stored value is of unexpected type. Clearing.";
+
+    clear(key);
```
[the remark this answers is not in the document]
brizental: What would you suggest we do instead?

src/shared/state/localstate.h
brizental: Ah, we need it :/ We need to grab the QSettings from the header file. The cpp file is not in the same scope and the `friend` thing doesn't work. Do you have a better suggestion?

src/shared/state/sessionstate.cpp
brizental: !!! Big deal. I'll see if I can add a test so we don't make this mistake again.

src/shared/state/statebase.cpp
```diff
@@ -30,35 +30,35 @@ QJsonValue::Type AddonStateBase::typeToQJsonValueType(QString type) {
 }
 
 // static
-StateHash AddonStateBase::parseManifest(const QJsonObject& manifest) {
+StateHash StateBase::parseStateFromJson(const QJsonObject& initialState) {
```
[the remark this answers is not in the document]
brizental: Can we do that even though we call this function only when initializing the StateBase?

src/shared/state/state.h
brizental: Are you sure? I get errors when I remove it. Note this file doesn't have a corresponding .cpp file. Should I add one?

src/shared/state/state.h
brizental: This is not an abstract class.

src/shared/state/state.h
brizental: Hum, we need to be able to call non-const methods with this. Would adding `const` here prevent us from doing that?

brizental:
Closing as this will not be implemented anymore.

PR #6157, opened 2023-02-25
Correct status code for IAP network requests - VPN-4204

bakulf:


src/apps/vpn/tasks/purchase/taskpurchase.cpp
```diff
@@ -56,7 +56,13 @@ TaskPurchase::TaskPurchase(Op op) : Task("TaskPurchase"), m_op(op) {
 TaskPurchase::~TaskPurchase() { MZ_COUNT_DTOR(TaskPurchase); }
 
 void TaskPurchase::run() {
-  NetworkRequest* request = new NetworkRequest(this, 200);
+  NetworkRequest* request = new NetworkRequest(this,
+#ifdef MZ_IOS
+                                               201
```
bakulf, line 61: maybe we should have a v2 endpoint for iOS to return 200. Follow up?

PR #6188, opened 2023-02-28
Silent switch when the list of apps to exclude changes - VPN-4254

bakulf:


lesleyjanenorton (approved):

lesleyjanenorton:
Is there any reason to keep `protectSelectedApps` ?

bakulf:
> Is there any reason to keep `protectSelectedApps` ?

It's currently used by `VPNToggleCard.qml`. We want to have it here because maybe the user changes the toggle state only without changing the list of Apps.

PR #6191, opened 2023-03-01
Do not use encrypted settings for auto-login linux env - VPN-3176

bakulf:


src/shared/platforms/linux/linuxcryptosettings.cpp
```diff
@@ -111,22 +111,55 @@ bool CryptoSettings::getKey(uint8_t key[CRYPTO_SETTINGS_KEY_SIZE]) {
...
+      }
     }
+
+    // Check if "org.freedesktop.Accounts" "AutomaticLoginUsers" is empty.
+    {
+      QDBusInterface iface(
+          "org.freedesktop.Accounts", "/org/freedesktop/Accounts",
+          "org.freedesktop.DBus.Properties", QDBusConnection::systemBus());
+
+      QDBusMessage reply =
```
oskirby, line 139: For correctness, we should check that the call succeeded before parsing the arguments, by examining `reply.type() == QDBusMessage::ReplyMessage`

oskirby (approved):

PR #6210, opened 2023-03-03
VPN-4270 - Hot reload qml

strseb:
## Description

I'll be out the next 2.5 weeks so i did not want to get this lost :) 

However this might make qml more sweet for people. 
This consists out of 2 things: 
-> hot.js: Is an fs_watcher and calls `live_reload <path-to-qml>` whenever something changes via inspector
-> Inspector_patchtes: new command `live_reload` -> will intercept all qml loads and replace a matching file with the new path, can be file://, http://, qrc://<file>. Will also reload the current screen.

https://user-images.githubusercontent.com/9611612/222613103-edca265c-f489-4762-afa7-22d344d62c19.mov

### DIFF
--- a/src/shared/frontend/navigator.cpp
+++ b/src/shared/frontend/navigator.cpp
@@ -408,50 +408,5 @@ void Navigator::registerScreen(int screenId, LoadPolicy loadPolicy,
 }
 
 void Navigator::reloadCurrentScreen() {
-  auto old_screen = m_currentScreen;
-
-  ScreenData* currentScreen = nullptr;
-  for (ScreenData& screen : s_screens) {
-    if (screen.m_screen == m_screenHistory.last()) {
-      currentScreen = &screen;
-      break;
-    }
-  }
-  Q_ASSERT(currentScreen);
-
-  for (int i = currentScreen->m_layers.length() - 1; i >= 0; --i) {
-    const Layer& layer = currentScreen->m_layers.at(i);
-
-    if (layer.m_type == Layer::eStackView) {
-      QVariant property = layer.m_layer->property("depth");
-      if (!property.isValid()) {
-        logger.warning() << "Invalid depth property!!";
-        continue;
-      }
-
-      int depth = property.toInt();
-      if (depth > 1) {
-        emit goBack(layer.m_layer);
-        return;
-      }
-
-      continue;
-    }
-
-    Q_ASSERT(layer.m_type == Layer::eView);
-    QVariant property = layer.m_layer->property("visible");
-    if (!property.isValid()) {
-      logger.warning() << "Invalid visible property!!";
-      continue;
-    }
-
-    bool visible = property.toBool();
-    if (visible) {
-      emit goBack(layer.m_layer);
-      return;
-    }
-
-    continue;
-  }
-  requestScreen(old_screen, ForceReloadAll);
+  requestScreen(m_currentScreen, ForceReloadAll);
 }
--- a/src/shared/frontend/navigator.cpp
+++ b/src/shared/frontend/navigator.cpp
@@ -406,3 +406,52 @@ void Navigator::registerScreen(int screenId, LoadPolicy loadPolicy,
   s_screens.append(ScreenData(screenId, loadPolicy, qmlComponentUrl,
                               requiresAppState, priorityGetter, quitBlocked));
 }
+
+void Navigator::reloadCurrentScreen() {
+  auto old_screen = m_currentScreen;
+
+  ScreenData* currentScreen = nullptr;
+  for (ScreenData& screen : s_screens) {
+    if (screen.m_screen == m_screenHistory.last()) {
+      currentScreen = &screen;
+      break;
+    }
+  }
+  Q_ASSERT(currentScreen);
+
+  for (int i = currentScreen->m_layers.length() - 1; i >= 0; --i) {
+    const Layer& layer = currentScreen->m_layers.at(i);
+
+    if (layer.m_type == Layer::eStackView) {
+      QVariant property = layer.m_layer->property("depth");
+      if (!property.isValid()) {
+        logger.warning() << "Invalid depth property!!";
+        continue;
+      }
+
+      int depth = property.toInt();
+      if (depth > 1) {
+        emit goBack(layer.m_layer);
+        return;
+      }
+
+      continue;
+    }
+
+    Q_ASSERT(layer.m_type == Layer::eView);
+    QVariant property = layer.m_layer->property("visible");
+    if (!property.isValid()) {
+      logger.warning() << "Invalid visible property!!";
+      continue;
+    }
+
+    bool visible = property.toBool();
+    if (visible) {
+      emit goBack(layer.m_layer);
+      return;
+    }
+
+    continue;
+  }
+  requestScreen(old_screen, ForceReloadAll);
+}
--- a/src/shared/frontend/navigator.h
+++ b/src/shared/frontend/navigator.h
@@ -62,6 +62,8 @@ class Navigator final : public QObject {
 
   Q_INVOKABLE bool eventHandled();
 
+  void reloadCurrentScreen();
+
   void registerReloader(NavigatorReloader* reloader);
   void unregisterReloader(NavigatorReloader* reloader);
 
--- a/src/shared/frontend/navigator.h
+++ b/src/shared/frontend/navigator.h
@@ -62,6 +62,11 @@ class Navigator final : public QObject {
 
   Q_INVOKABLE bool eventHandled();
 
+  /**
+   * @brief Request's the Load of the current screen with
+   * ForceReloadAll policy
+   *
+   */
   void reloadCurrentScreen();
 
   void registerReloader(NavigatorReloader* reloader);
--- a/src/shared/inspector/inspectorhotreloader.cpp
+++ b/src/shared/inspector/inspectorhotreloader.cpp
@@ -4,45 +4,41 @@
 
 #include "inspectorhotreloader.h"
 
-
+#include <QDir>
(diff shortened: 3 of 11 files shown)


src/apps/vpn/inspector/inspectorhotreloader.cpp
```diff
@@ -0,0 +1,33 @@
...
+  }
+  logger.debug() << "Not redirect: " << url.toString();
+  return url;
+}
+
+void InspectorHotreloader::annonceReplacedFile(const QUrl& path) {
+  logger.debug() << "Announced redirect! : " << path.fileName() << " as ->"
+                 << path.toString();
+  m_target->clearComponentCache();
+  m_announced_files.insert(path.fileName(), path);
```
bakulf, line 31: What about if you load again the same file? Why dup the insert?

src/apps/vpn/inspector/inspectorhandler.cpp
```diff
@@ -146,6 +148,16 @@ static QList<InspectorCommand> s_commands{
                        MozillaVPN::instance()->controller()->quit();
                        return QJsonObject();
                      }},
+    InspectorCommand{"live_reload", "Live reload file X", 1,
+                     [](InspectorHandler*, const QList<QByteArray>& args) {
+                       if (s_hotreloader == nullptr) {
+                         s_hotreloader = new InspectorHotreloader(
```
bakulf, line 154: In this way you create multiple obj for the same file, right?

tools/hot.js
```diff
@@ -0,0 +1,60 @@
...
+    ignoreInitial: true,
+    awaitWriteFinish: {stabilityThreshold: 200, pollInterval: 100}
+  });
+  watcher.on('all', function(event, path) {
+    if (event === 'add' || event === 'change') {
+      if (path.endsWith('.qml')) {
+        const relPath = path.replace(/\\/g, '/');
+
+        const url = `file://${currentPath}/${relPath}`;
+        websocket.send(`live_reload ${url}`);
```
bakulf, line 27: We need unload_reload too.

bakulf (changes requested):

bakulf:
What about this idea:

- let's create a new addon type: 'serviceworker' or something like that.
- each serviceworker addon must expose a JS file
- when the app is about to load something, it asks all the serviceworker addons if the input URL can be intercepted by them. The first one wins.
- if no addons are able to intercept the URL, the app loads it from QRC as usual
- if there is 1 addon able to intercept the URL, its content is used instead.

In this way we create:
- a language.json addon
- a server city/country name addon
- but we can also replace entire parts of the UI

Next step is to implement a hot-reload-addon, that does the intercept using local files. This can also be a feature for dev-menu. So basically, no inspector changes.

Thoughts?

tools/hot.js
strseb: Added! I've added an unload command that will revert all things.

src/apps/vpn/inspector/inspectorhotreloader.cpp
strseb: It's a QMap<string,string> with fileName as key, therefore we don't duplicate anything, we just update the path qml-engine can find that filename

src/apps/vpn/inspector/inspectorhandler.cpp
strseb: We always only update the path, not create new inserts.

strseb:
>What about this idea [...]

As we already talked offline 100% liking that. However this pr's focus is more to have a new DevTools ready. I've now kind of made http work, so calling `hot.js --remote` will also work with mobile.

src/shared/inspector/inspectorhotreloader.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include "inspectorhotreloader.h"
+
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+
+#include "networkmanager.h"
+
+namespace {
+Logger logger("hotttt");
```
bakulf, line 13: better name?

src/shared/inspector/inspectorhotreloader.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+}
+
+QUrl InspectorHotreloader::intercept(
+    const QUrl& url, QQmlAbstractUrlInterceptor::DataType type) {
+  if (type != DataType::QmlFile) {
```
bakulf, line 19: Do we need this?

src/shared/inspector/inspectorhotreloader.cpp
```diff
@@ -0,0 +1,63 @@
...
+                 << path.toString();
+  m_target->clearComponentCache();
+  m_announced_files.insert(path.fileName(), path);
+
+  if (path.scheme() == "file" || path.scheme() == "qrc") {
+    // is's a file, just load it!
+    Navigator::instance()->reloadCurrentScreen();
+    return;
+  }
+  if (path.scheme() != "http") {
```
bakulf, line 43: can you do this check before add it to the m_announced_files?

src/shared/inspector/inspectorhotreloader.cpp
```diff
@@ -0,0 +1,63 @@
...
+    logger.error() << "Unexpected File Scheme in: " << path.toString();
+  }
+  // Okay so let's have a request with the nm so the file is in the cache once
+  // we hit it!
+  QNetworkAccessManager* nm =
+      NetworkManager::instance()->networkAccessManager();
+  auto req = QNetworkRequest(path);
+  auto reply = nm->get(req);
+  QObject::connect(reply, &QNetworkReply::finished, [path] {
```
bakulf, line 52: how do you know that the entry has been stored in cache?

src/shared/inspector/inspectorhotreloader.h
```diff
@@ -0,0 +1,48 @@
+//
+// Created by Basti on 02/03/2023.
```
bakulf, line 2: header

src/shared/inspector/inspectorhotreloader.h
```diff
@@ -0,0 +1,48 @@
...
+//
+#ifndef MOZILLA_VPN_INSPECTORHOTRELOADER_H
+#define MOZILLA_VPN_INSPECTORHOTRELOADER_H
+
+#include <QApplication>
+#include <QObject>
```
bakulf, line 8: not needed

src/shared/inspector/inspectorhotreloader.h
```diff
@@ -0,0 +1,48 @@
...
+#include <QQmlAbstractUrlInterceptor>
+#include <QQmlEngine>
+
+#include "frontend/navigator.h"
+#include "frontend/navigatorreloader.h"
+#include "logger.h"
```
bakulf, line 14: remove these 3 headers.

src/shared/inspector/inspectorhotreloader.h
```diff
@@ -0,0 +1,48 @@
...
+   */
+  void annonceReplacedFile(const QUrl& path);
+
+  // Removes all injected paths
+  // Reload's the current screen with the internal
+  // Components
+  void resetAllFiles();
+
+ private:
+  QQmlEngine* m_target;
```
bakulf, line 44: = nullptr

bakulf (dismissed):

tools/qml_hot_reload/hot.js
```diff
@@ -0,0 +1,89 @@
...
+
+process.on('SIGINT', function() {
+  console.log("Unregistering all reloads...");
+  websocket.send(`reset_live_reload`);
+  process.exit();
+});
+
+// TODO: Theoretically we can serve this via http:
+// however i found that the first load was always screwed up lol
+//
```
Gela, line 70:
```suggestion
// TODO: Theoretically we can serve this via http:
// however I found that the first load never worked correctly
```

😅

Gela (changes requested):
This is really awesome 🥳

src/shared/inspector/inspectorhotreloader.cpp
strseb: Good Point. I was hoping NetworkManager's internal cache would have some "auto cache" enabled. But it seems that is not enabled. 
So i now have created a `fetchAndAnnounce` - Which will pull files locally and do replace to `file://<tmpFolder>/File.qml`

src/shared/frontend/navigator.cpp
```diff
@@ -406,3 +406,52 @@ void Navigator::registerScreen(int screenId, LoadPolicy loadPolicy,
...
+  ScreenData* currentScreen = nullptr;
+  for (ScreenData& screen : s_screens) {
+    if (screen.m_screen == m_screenHistory.last()) {
+      currentScreen = &screen;
+      break;
+    }
+  }
+  Q_ASSERT(currentScreen);
+
+  for (int i = currentScreen->m_layers.length() - 1; i >= 0; --i) {
```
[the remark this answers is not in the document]
strseb: Thank you for going though this -  removed that part. 
Turns out just doing `requestScreen(m_currentScreen, ForceReloadAll);` is enough to refresh the screens 😓

src/shared/frontend/navigator.h
```diff
@@ -62,6 +62,13 @@ class Navigator final : public QObject {
 
   Q_INVOKABLE bool eventHandled();
 
+  /**
+   * @brief Request's the Load of the current screen with
+   * ForceReloadAll policy
```
Gela, line 67:
```suggestion
   * @brief Requests reloading the current screen with
   * ForceReloadAll policy
```

Gela (approved):
This is awesome, thank you!

PR #6238, opened 2023-03-07
Expose the shared resources in src/shared - VPN-4305

bakulf:


src/apps/vpn/commands/commandui.cpp
```diff
@@ -305,307 +300,90 @@ int CommandUI::run(QStringList& tokens) {
     if (provider) {
       engine->addImageProvider(QString("app"), provider);
     }
-    qmlRegisterSingletonType<MozillaVPN>(
```
oskirby, line 308: This looks way cleaner, but can you share a word about why the call to `QQmlEngine::setObjectOwnership(obj, QQmlEngine::CppOwnership)` was removed? Was it never needed, or has something changed in how the C++/QML object lifecyle is garbage collected?

bakulf: it was needed with `qmlRegisterSingletonType<T>`. It's by default with the new syntax.

oskirby (approved):

PR #6257, opened 2023-03-08
Fix a crash in the current "subscription confirm" controller step - VPN-4314

bakulf:


src/apps/vpn/controller.cpp
```diff
@@ -234,32 +234,39 @@ bool Controller::activate(const ServerData& serverData,
...
-                setState(StateConnecting);
-              } else {
-                setState(StateOff);
+              if (MozillaVPN::instance()->state() != MozillaVPN::StateMain) {
+                return;
               }
+
+              setState(StateConnecting);
+
+              clearRetryCounter();
```
birdsarah, line 252: Can you explain about why this and the next line is needed?

src/apps/vpn/controller.cpp
```diff
@@ -234,32 +234,39 @@ bool Controller::activate(const ServerData& serverData,
...
 
               // Check if the error propagation has changed the Mozilla VPN
               // state. Continue only if the user is still authenticated and
               // subscribed.
-              if (MozillaVPN::instance()->state() == MozillaVPN::StateMain) {
-                setState(StateConnecting);
-              } else {
-                setState(StateOff);
+              if (MozillaVPN::instance()->state() != MozillaVPN::StateMain) {
+                return;
```
birdsarah, line 247: This is a `bool` method, why isn't this `return false;`?

birdsarah: This leaves us in state `StateCheckSubscription`, is that what we want?

birdsarah: Do we have docs for this state machine?

src/apps/vpn/controller.cpp
```diff
@@ -234,32 +234,39 @@ bool Controller::activate(const ServerData& serverData,
...
             });
 
     connect(request, &NetworkRequest::requestCompleted, this,
-            [this](const QByteArray& data) {
+            [this, serverSelectionPolicy](const QByteArray& data) {
               MozillaVPN::instance()->accountChecked(data);
               setState(StateConnecting);
+
+              clearRetryCounter();
+              activateInternal(DoNotForceDNSPort, serverSelectionPolicy);
```
birdsarah, line 262: Why is this call to `activateInternal` needed. It wasn't needed before this account check was added so I'm assuming it's being called elsewhere.

birdsarah (changes requested):
I have lots questions to help with my understanding.

Is there a way to write a test that replicates the crash scenario?

src/apps/vpn/controller.cpp
bakulf: 1. this return is for the nested function, which does not have a return value. So, `return` is correct
2. No because of this: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/apps/vpn/mozillavpn.cpp#L139-L146 - when MozillaVPN state id != StateMain, the controller is deactivated automatically
3. not enough. Basically not yet.

src/apps/vpn/controller.cpp
bakulf: This line should be moved elsewhere, but I did not want to apply big changes in this PR and just fix the crash.
This line resets the Controller timer (what we show in the UI). 
We don't want to move it in activateInternal() because that means it will be seen during silent-switch too. But still, it's horrible to have it around.

bakulf:
> Is there a way to write a test that replicates the crash scenario?

The test consists in connecting the VPN and wait more than a few seconds, in a debug build. If I want I can write such test but it will be just a big `sleep()`. But maybe we can write a better test to check the timer to work. We don't have tests for the timer string. Follow up?

src/apps/vpn/controller.cpp
bakulf: see the `return` after the network request.

src/apps/vpn/controller.cpp
birdsarah: > this return is for the nested function, which does not have a return value. So, return is correct

doh. thanks.

birdsarah (approved):

PR #6258, opened 2023-03-08
Stop depending on vpnFlickable id on focus change

MattLichtenstein:
## Description

- Instead of depending on a component specifically named `vpnFlickable` when ensuring a component is visible when gaining focus, iterate through it's view hierarchy to see if it is an ancestor of a scrollable view
  - Results in a fix for keyboard nav scrolling on the tips and tricks page
- Introduce MZUtils - a home for generic qml utility functions in nebula

https://user-images.githubusercontent.com/15353801/223855212-c2136daa-756b-4e24-8321-a57829c583ed.mov

## Reference

[VPN-2958: [a11y] Scroll is not working as expected using keyboard](https://mozilla-hub.atlassian.net/browse/VPN-2958)

### DIFF
--- a/nebula/ui/components/forms/MZRadioButton.qml
+++ b/nebula/ui/components/forms/MZRadioButton.qml
@@ -7,7 +7,7 @@ import QtQuick.Controls 2.14
 
 import Mozilla.Shared 1.0
 import components 0.1
-import util 0.1
+import utils 0.1
 
 RadioDelegate {
     id: radioControl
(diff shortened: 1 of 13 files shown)


nebula/ui/components/forms/MZRadioButton.qml
```diff
@@ -26,8 +27,8 @@ RadioDelegate {
     onFocusChanged: {
```
lesleyjanenorton, line 27: Should we change this one to also be `onActiveFocusChanged` ?

lesleyjanenorton (approved):
Nice!

nebula/ui/components/forms/MZRadioButton.qml
MattLichtenstein: Shouldn't make a difference here, but let's do it for consistency!

PR #6267, opened 2023-03-10
Refactoring log serializers - VPN-4326

bakulf:
Currently, all the log-serializing methods are in MozillaVPN and they are strongly VPN-based (controller, device, settings, etc).
             
             The new approach will be the following:
             
             - LogHandler.{cpp,h} contains all the serialization methods
             - LogHandler is exposed as `MZLog` in the `Mozilla.Shared` namespace
             - Any component that wants to report logs, can register itself using `LogHandler::registerLogSerializer()`
             - So far, the LogSerializers are `Controller`, `DeviceModel` (instead of `Device`), and `SettingsHolder`


bakulf:
Draft because I break android builds.

bakulf:
Waiting for our Android export, this is a good first step. We can remove VPNUtils.kt as a follow up.

oskirby (approved):

Issue #6207, opened 2023-03-02
Ctest support for functional testing


data-sync-user:
Wouldn’t it be nice if the functional tests could be automated and integrated with the `ctest` suite to create a one-stop-shop for testing that ties in with the build system.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-4292)
┆Reporter: Owen Kirby

Issue #6206, opened 2023-03-02
Cmake support for addon generation


data-sync-user:
Addons are currently built using a collection of python scripts that have to be run manually. Normally this can be ignored during client development work, but every now and then we need to generate an addon to run a test. It would be nice if these could be integrated with the cmake build process.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-4291)
┆Reporter: Naomi Kirby

data-sync-user:
➤ Beatriz Rizental Machado commented:

YES!!!

data-sync-user:
➤ Naomi Kirby commented:

This has been implemented by PR #6832 ( https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6832 ) which is now merged onto main

There are no specific QA tests around this work as it only affects the build processes for Addons, but it would be nice if we could verify that the generated addons continue to work as expected.

data-sync-user:
➤ Valentina Virlics commented:

Naomi Kirby Could we have a fixVersion for this ticket? Thank you!

data-sync-user:
➤ Santiago Andrigo commented:

Adding qa-not-actionable as per Naomi Kirby 's comment below.

PR #6270, opened 2023-03-10
CTest support for functional testing and addon generation

oskirby:
## Description
This started off as an experiment to try and make the functional tests runnable using CTest so that we can put all of the tests together in one suite. And then it wound up turning into an excercise in porting the Addon generation into CMake as well because it became impossible to separate the two.

For the time being, support for this is controlled with the `BUILD_FUNCTIONAL_TESTS` CMake option, otherwise it will get automatically run as a part of our unit test suite.

## Reference
Github issue #6207 ([VPN-4292](https://mozilla-hub.atlassian.net/browse/VPN-4292))
Github issue #6206 ([VPN-4291](https://mozilla-hub.atlassian.net/browse/VPN-4291))

## Checklist
    

scripts/cmake/addons.cmake
```diff
@@ -0,0 +1,63 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### Helper function to build a VPN Addon
```
bakulf, line 5: Does this mean we can use cmake target=addon to generate them? And get rid of the scripts?

bakulf (approved):

scripts/cmake/addons.cmake
oskirby: Yeah, that is one of my goals in this PR before I'll consider it ready for review.

PR #6271, opened 2023-03-11
Add metric for start on boot setting

mcleinman:
## Description

Adding another metric from Super Dooper metrics - this one tracking when we change the "start on boot" metric.

I did some local testing, and confirmed this is working as expected.

Data team will: Take last report from a given device (and a lack of report for an active device means opted in) to calculate overall population metric for this

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4069

## Checklist
    

bakulf (changes requested):
I don't like the idea of using glean in the codebase like this because it's hard to track the events and where they are generated. 
Plus it mixes frontend code and telemetry tasks. No code separation == spaghetti code.

I know that we already do this elsewhere but we can improve! 

What about having something in telemetry.cpp. you can connect the settingsholder::startAtBootChanged signal to glean. 

Thoughts?

mcleinman:
> I don't like the idea of using glean in the codebase like this because it's hard to track the events and where they are generated. Plus it mixes frontend code and telemetry tasks. No code separation == spaghetti code.
> 
> I know that we already do this elsewhere but we can improve!
> 
> What about having something in telemetry.cpp. you can connect the settingsholder::startAtBootChanged signal to glean.
> 
> Thoughts?

I like that a lot, totally agree with you that it's better code separation. I've updated this and pushed up new code.

scripts/macos/apple_compile.sh
```diff
@@ -109,8 +109,10 @@ python3 scripts/utils/import_languages.py $([[ $QTBINPATH ]] && echo "-q $QTBINP
...
 python3 scripts/utils/generate_glean.py || die "Failed to generate glean samples"
 
-print Y "Generating qtglean samples..."
-python3 qtglean/glean_parser_ext/run_glean_parser.py || die "Failed to generate qtglean samples"
+if [ "$OS" = "macos" ]; then
+  print Y "Generating qtglean samples..."
+  python3 qtglean/glean_parser_ext/run_glean_parser.py || die "Failed to generate qtglean samples"
+fi
 
 printn Y "Extract the project version... "
```
bakulf, line 117: This script is used only for iOS. Remove this block.

src/apps/vpn/telemetry.cpp
```diff
@@ -180,6 +181,15 @@ void Telemetry::initialize() {
         GleanSample::serverUnavailableError);
   });
 
+  connect(SettingsHolder::instance(), &SettingsHolder::startAtBootChanged, this,
+          []() {
+#if !defined(MZ_IOS)
```
bakulf, line 186: Start on boot is not a setting on iOS. This if-stmt can be removed

bakulf (changes requested):

scripts/macos/apple_compile.sh
mcleinman: Done - this was when I was trying to figure out why it failed, it's been removed

src/apps/vpn/telemetry.cpp
mcleinman: Done - this was when I was trying to figure out why it failed, it's been removed

bakulf (approved):

PR #6278, opened 2023-03-13
Chore: Ignore qtglean/Cargo.lock

birdsarah:


.gitignore
```diff
@@ -33,6 +33,7 @@ linux/netfilter/vendor/
 __pycache__
 lottie.mjs
 qml_lint_result.txt
+qtglean/Cargo.lock
```
bakulf, line 36: because it has been committed, you have to remove it too.

birdsarah: :doh:

bakulf (approved):

andersk:
Please don’t ignore `Cargo.lock` files. They’re required so that a given version of Mozilla VPN can be rebuilt from the same code. The NixOS package cannot be built without `Cargo.lock`, because NixOS enforces this kind of determinism using a sandbox that disables network access for builds that don’t have a fixed output hash.

PR #6294, opened 2023-03-14
Implement a share component for user authentication state and state machine - VPN-4338

bakulf:
This class will be extended in follow ups to move most of the state machine logic from MozillaVPN


src/apps/vpn/tasks/sendfeedback/tasksendfeedback.cpp
```diff
@@ -6,12 +6,12 @@
 
 #include <QJsonObject>
 
+#include "app.h"
 #include "appconstants.h"
+#include "env.h"
```
mcleinman, line 11: So I can learn, why do we need to add this line here?

src/apps/vpn/mozillavpn.h
```diff
@@ -38,7 +39,7 @@ class Telemetry;
 class Theme;
 class User;
 
-class MozillaVPN final : public QObject {
+class MozillaVPN final : public App {
```
mcleinman, line 42: What is the advantage of having `MozillaVPN` inherit from `App`?

src/apps/vpn/tasks/sendfeedback/tasksendfeedback.cpp
Gela: It appears to be used further down on line 48:

```
{"versionString", Env::versionString()},
```

Gela (approved):

src/apps/vpn/tasks/sendfeedback/tasksendfeedback.cpp
mcleinman: Thank you!

src/shared/app.h
```diff
@@ -0,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef APP_H
+#define APP_H
+
+#include <QObject>
+
+class App : public QObject {
```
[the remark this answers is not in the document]
bakulf: no because it will be much more than user-status. It will host the entire state machine.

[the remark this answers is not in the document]
mcleinman: While this ship has sailed, I still think something more descriptive than `app` - perhaps `StateMachine` - would make it more intuitive down the line. But I didn't block on this, and I'm okay that this merged.

PR #6297, opened 2023-03-14
Bump taskgraph decision image to upgrade taskgraph and pick up quiet unzip fix

bhearsum:


bhearsum:
Silly me, it turns out that we need to update `taskgraph` on the Windows builders to actually quiet down unzipping. There's no real point in merging this...

PR #6351, opened 2023-03-16
Implement the NavigatorBarModel + NavigatorBaricon - VPN-4375

bakulf:


src/shared/frontend/navigatorbaricon.h
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef NAVIGATORBARICON_H
+#define NAVIGATORBARICON_H
+
+#include <QObject>
+
+class NavigatorBarIcon final : public QObject {
```
MattLichtenstein, line 10: To more closely associate this with the actual NavigationBar UI component, and not the navigator class, do you think this class might be more appropriately named `navigationbaricon` (and same for the model)?

PR #6389, opened 2023-03-21
VPN-3680: Update controller view strings

lesleyjanenorton:
## Description

- Adds new accessible label strings for the speed test and connection information/ip panel buttons
- Removes old strings

## Reference

VPN-3680

## Checklist
    
### DIFF
--- a/src/apps/vpn/translations/strings.yaml
+++ b/src/apps/vpn/translations/strings.yaml
@@ -397,7 +397,6 @@ genericError:
 
 # Connection Info Screen
 connectionInfo:
-  openButton: Connection Info
   startSpeedTest:
     value: Start speed test
     comment: Accessible label for the button that opens the speed test view and initiates a speed test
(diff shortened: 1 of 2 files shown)


translations/strings.yaml
```diff
@@ -370,6 +369,9 @@ genericError:
 # Connection Info Screen
 connectionInfo:
   openButton: Connection Info
+  startSpeedTest:
+    value: Start speed test
+    comment: Accessible label for the button that opens the speed test view and initiates a speed test
```
flodolo, line 374: Accessible or Accessibility?

flodolo (approved):

translations/strings.yaml
```diff
@@ -340,7 +340,6 @@ dnsOverwriteDialog:
   primaryButton: Confirm
   secondaryButton: Maybe later
 
-aboutUs:
```
MattLichtenstein, line 343: probably need to add this back

MattLichtenstein (changes requested):
Also, should we remove the old strings that are being replaced?

flodolo:
> Leaves old strings put for now and can be deleted when no longer needed in release branch

This is not really necessary, automation takes care of keeping strings needed by release and release branches.

translations/strings.yaml
lesleyjanenorton: done, thanks!

MattLichtenstein (approved):

PR #6392, opened 2023-03-21
Save the DNS settings at focus - VPN-4403 & VPN-4405

bakulf:


src/apps/vpn/ui/screens/settings/ViewDNSSettings.qml
```diff
@@ -17,6 +17,7 @@ MZViewBase {
 
     _menuTitle: MZI18n.SettingsDnsSettings
 
+    property bool initialized: false
```
MattLichtenstein, line 20: no longer need the `initialized` property

bakulf: this is to prevent activeFocusChanged to trigger a save before onCompleted.

MattLichtenstein: why is it a problem if a save is made before `onCompleted`?

bakulf: because in onCompleted we call 'reset' that sets the correct values. If we save before that, we end up saving random things.

MattLichtenstein (approved):

PR #6401, opened 2023-03-21
Modularize i18n - VPN-4410

bakulf:
This PR modularize the app translations. The idea is the following:
- translations/CMakeLists.txt generates N libraries called `translations_<appName>`.
- the generated files are stored into `<buildFolder>/translations/generated/<appName>`
- Each app has a `translations` folder (ex: `src/apps/vpn/translations`) containing:
  - `i18n` - a git submodule or a folder containing translations
  - `template` - a file to define how the strings are called in the `i18n` folder. For instance, for the VPN project we use `mozillavpn`
  - `strings.yaml` - the famous strings yaml file
  - `extras` - anything here will be included in the `:/i18n/` QRC folder. For instance: `servers.json`, `languages.json`
- each app includes the `translation_<appName>` library, if it wants

This works with `qmake` too for iOS.

When we land this PR, we can create a new pontoon project for the relay app and include it as a submodule in `src/apps/relay/translations/i18n`. 

Note: the addons are still mozillaVPN-only. The addons for relay will be done in a follow up.


flodolo:
I need more time to go through this, but I noticed that the translations workflow now has 4 .ts files. Is that expected?

```
translations_auth_tests.ts
translations_template.ts
translations_unit_tests.ts
translations_vpn.ts
```

On a practical note: this has implications on the automation running in the l10n repository, which needs to deal with release branches on top of `main`.

Since I'm out the next week, I doubt that I will able to look into this, but I'd be happy to review a PR when I'm back. We shouldn't merge this PR until we have that part solved, because it would block any work on 2.15 strings.

flodolo:
> I need more time to go through this, but I noticed that the translations workflow now has 4 .ts files. Is that expected?

Even if this is expected, I assume there is something wrong, because these files shouldn't have the same content.

flodolo:
I tried to put together a [change](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/6205f0dd6270248dfe6bfdd8b0d7e0cb27001d9c) in the l10n automation, then run against this branch.
* The [log](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/actions/runs/4498642510/jobs/7915509710) for the extraction step moved from 11k lines to 40k, and it takes 3x the previous time. 
* The resulting XLIFF file is missing all the qml strings, it only has strings from `translations/generated/vpn/i18nstrings_p.cpp`.

So, there is something broken (running multiple times over?), or at least unclear to me.

flodolo:
> Note: the addons are still mozillaVPN-only. The addons for relay will be done in a follow up.

We're now generating ~30 addons, for 5 branches, for 4 projects. That's ~600 addons, when it should be ~150 (and maybe we could drop older branches).

We should run addon generation only for the vpn project for now.

scripts/utils/generate_ts.sh
```diff
@@ -21,82 +21,101 @@ mkdir -p cache || die
...
-        cp "addons/generated/addons/${ts_name}" addon_ts/
-      fi
-    done
-    rm addons/generated/addons/*.ts || die
+  for branch in $(git branch -r | grep origin/releases); do
+    git checkout $branch &>/dev/null || die
+
+    printn Y "Importing main strings from $branch..."
+    if [ -f translations/strings.yaml ]; then
+      python3 cache/generate_strings.py -o translations/generated/$project translations/strings.yaml || die
```
flodolo, line 70: One the branch the project doesn't exist, so you're extracting strings.yaml from VPN into all files.

flodolo (changes requested):
Still no clue why the XLIFF for vpn would be missing all the files outside of strings.yaml

scripts/utils/generate_ts.sh
```diff
@@ -21,82 +21,101 @@ mkdir -p cache || die
...
+  printn Y "$project - Generating a dummy PRO file... "
+  cat > translations/generated/$project/dummy_ts.pro << EOF
 HEADERS += i18nstrings.h
 HEADERS += \$\$files(../../src/shared/*.h, true)
-HEADERS += \$\$files(../../src/apps/vpn/*.h, true)
+HEADERS += \$\$files(../../src/apps/$project/*.h, true)
 HEADERS += \$\$files(../../nebula/*.h, true)
 
 SOURCES += i18nstrings_p.cpp
 SOURCES += ../i18nstrings.cpp
```
flodolo, line 41: Why do we need two for this file?

scripts/utils/generate_ts.sh
```diff
@@ -21,82 +21,101 @@ mkdir -p cache || die
...
-mkdir -p translations/generated || die
-cat > translations/generated/dummy_ts.pro << EOF
+  printn Y "$project - Generating strings... "
+  python3 cache/generate_strings.py src/apps/$project/translations/strings.yaml -o translations/generated/$project
+  print G "done."
+
+  printn Y "$project - Generating a dummy PRO file... "
+  cat > translations/generated/$project/dummy_ts.pro << EOF
 HEADERS += i18nstrings.h
 HEADERS += \$\$files(../../src/shared/*.h, true)
```
flodolo, line 36: Path should go 3 levels up, not 2? Also, this is not going to work on branches?

bakulf:
> I need more time to go through this, but I noticed that the translations workflow now has 4 .ts files. Is that expected?

Yes it is. We will use `translations_vpn.ts` for the VPN project instead of the `translations.ts` file we had before.

We can ignore the other files. Soon-ish we will have a `translations_relay.ts` for the Relay App.
I have a PR ready for the i18n mozillavpn repo to land immediately after this.

> On a practical note: this has implications on the automation running in the l10n repository, which needs to deal with release branches on top of `main`.

This is correct. We need to change the way we create branches to include the project name. But it can be done as a follow up. This PR is already complex enough.

scripts/utils/generate_ts.sh
bakulf: because `relay` and other "apps" will have their own folder.

scripts/utils/generate_ts.sh
bakulf: yes it does because shared/apps are not "old" enough. We support only 2.13 and 2.14 releases.

flodolo: I was talking about all paths, not just this one. You can check the .ts file generated, it's missing all strings but the ones defined in strings.yaml

scripts/utils/generate_ts.sh
flodolo: (I missed the names were slightly different, I thought it was the same file)

flodolo (approved):
Tested the latest version, everything seems to work as expected now.

birdsarah (approved):
r+wc

scripts/utils/generate_ts.sh
```diff
@@ -21,82 +21,110 @@ mkdir -p cache || die
...
-    for f in $ts_files
-    do
-      ts_name=$(basename "$f")
-      if [ -f "addon_ts/${ts_name}" ]; then
-        printn Y "File ${ts_name} exists, updating with branch strings..."
-        ${QT_HOST_BINS}/lconvert -i "addon_ts/${ts_name}" "addons/generated/addons/${ts_name}" -o tmp.ts || die
-        mv tmp.ts "addon_ts/${ts_name}"
+    printn Y "Importing main strings from $branch..."
+    if [ -f translations/strings.yaml ]; then
+      if [ $project = "vpn" ]; then
```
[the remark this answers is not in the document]
flodolo: This part of the script iterates over release branch.

If `translations/strings.yaml` exists, it means we're not on `main`, but on a release branch. In that case, we only need to extract the strings for the VPN project, otherwise the string for VPN 2.14 will end up in all other projects (e.g. Relay).

The `main` branch will have a `strings.yaml` file for each project, e.g. `translations/generated/vpn/translations/strings.yaml`.

Basically, we have a lot of code to deal with the existence of release branches, and we should probably clean it up at some point.

PR #6405, opened 2023-03-22
VPN-4276 - Implement 2.15 'What's new' and 'Update' message

brizental:


addons/message_whats_new_v2.15/manifest.json
```diff
@@ -0,0 +1,45 @@
...
+        "id": "c_1",
+        "type": "ulist",
+        "content": [
+          {
+            "id": "l_1",
+            "content": "We will now recommend a few of the highest-performing server locations when you’re on the Select Location screen."
+          },
+          {
+            "id": "l_2",
+            "content": "We’ve updated the App Exclusions screen to make it easier to choose which apps to leave out of VPN protection."
```
flodolo, line 30: This is not how it's spelled in the app (it's [App exclusions](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/translations/strings.yaml#L304)). I think I've noted this issue in review several times at this point, can we redirect it back to content so we can avoid it in the future?

flodolo, line 30:
```suggestion
            "content": "We’ve updated the App exclusions screen to make it easier to choose which apps to leave out of VPN protection."
```

This is not [how it's spelled in the app](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/translations/strings.yaml#L304). 

I think I've reported this issue in review a few times at this point, can we make a note about it (@chrisrooneys)?

flodolo (commented):
Is the content final or are we expecting more changes?

brizental:
> Is the content final or are we expecting more changes?

AFAIK it's final.

flodolo (approved):

addons/message_update_v2.15/manifest.json
```diff
@@ -22,18 +22,14 @@
         "content": [
           {
             "id": "l_1",
-            "content": "You can now choose to filter out malware, ads, and ad trackers, giving you more protection as you navigate the web. Find these features in “Privacy features” within the Settings menu."
+            "content": "We will now recommend a few of the highest-performing server locations when you’re on the Select Location screen."
           },
           {
             "id": "l_2",
-            "content": "When you first launch the app, the app will now automatically suggest the server location that is closest to you."
+            "content": "We’ve updated the App Exclusions screen to make it easier to choose which apps to leave out of VPN protection."
```
flodolo, line 29: Should have made it clear that this needed fixing also here.

brizental: Sheesh I am sorry flod, I should have paid more attention. Will send the fix in a separate PR.

PR #6413, opened 2023-03-22
VPN-4429 Move Android-Glean into android/common

strseb:
## Description
Qt/Glean is now Shared Code, therefore the binds should live in android/common

PR #6418, opened 2023-03-22
VPN-4428 Android VPNApplication to shared component

strseb:
Moves VPNApplication into the shared code folder - Also porting this to kotlin while at it :) 

Note: This is currently based on top of #6413 given they both touch the same files


bakulf (approved):

bakulf (approved):
I like this PR but it's red...

PR #6437, opened 2023-03-23
Relayapp - first commit

bakulf:
This is the first commit for the Relay app. It contains the following:

- a new Cmake target: `relayapp`
- CMake for the relay app
- constants for relay app (note that some of them point to VPN API - this will changed in the future)
- RelayApp class
- Navigation bar configuration
- An empty `main.qml` file

Running this app, there is nothing more than an empty blank window. But it's a starting point!


src/apps/relay/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+get_filename_component(MZ_SOURCE_DIR ${CMAKE_SOURCE_DIR}/src ABSOLUTE)
+get_filename_component(GENERATED_DIR ${CMAKE_CURRENT_BINARY_DIR}/generated ABSOLUTE)
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
strseb, line 8: Nit: Should this be somewhere "shared", given we use this logic for vpn aswell?

strseb:
Nit: is the i18n submodule update intended to be pushed aswell?

src/apps/relay/CMakeLists.txt
bakulf: Yes... but I don't know how yet!

bakulf:
> Nit: is the i18n submodule update intended to be pushed aswell?

No. Updated

src/apps/relay/CMakeLists.txt
oskirby: We could move it to the top-level `CMakeLists.txt` in which case these variables will be set for every target that gets built. It would be a lot cleaner, but it will probably require a bunch of refactoring in how we handle the dummy builds and unit tests (which are a bit of a hack anyways).

oskirby (approved):
Looks good

flodolo (commented):
Is it expected for the `translations_relay.ts` file to have strings in it? That seems wrong to me.

[translations_relay.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/11135309/translations_relay.txt)

bakulf:
> Is it expected for the `translations_relay.ts` file to have strings in it? That seems wrong to me.

This is correct. All the strings from shared components are included in each translation file by default. 
I have to rebase the PR. I know Lesley cleaned up the nebula folder a lot. You will see less strings soon.

flodolo:
> This is correct. All the strings from shared components are included in each translation file by default. I have to rebase the PR. I know Lesley cleaned up the nebula folder a lot. You will see less strings soon.

Can you take a look at the .ts file in automation, e.g. [here](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/4617444580?pr=6437)? I think the list of shared components needs more trimming, because there is stuff like `VPN is off/on`, which doesn't seem to be useful for other products.

One other issue, that might be fixed over time, is the fact that all strings have a `vpn.*` identified.

PR #6438, opened 2023-03-23
Addon shared - IGNORE

bakulf:


flodolo:
I think this is not extracting strings for addons, but the workflow only has them because of the branches?

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/scripts/addon/generate_all.py#L32-L35

Besides the fact that this seems a confusing way to get to parent directories (you can join a path to `os.path.pardir`), this will search for addons in `/addons`.

bakulf:
This PR is just to move the addon component to the shared folder. It's not about having cross-project addons.
This will be done in a follow up. No changes to the `scripts` folder.

flodolo:
> This PR is just to move the addon component to the shared folder. It's not about having cross-project addons.
> This will be done in a follow up. No changes to the `scripts` folder.

That's not the problem. If you merge this as is, and land a new add-on (or change a string), we won't expose those strings for translation.

src/apps/vpn/notificationhandler.cpp
```diff
@@ -333,10 +339,37 @@ void NotificationHandler::messageClickHandle() {
...
+
+  connect(addon, &Addon::conditionChanged, this, [this, addon](bool enabled) {
+    if (enabled) {
+      maybeAddonNotification(addon);
+    }
+  });
+}
+
+void NotificationHandler::maybeAddonNotification(Addon* addon) {
+  Q_ASSERT(addon->type() == "message");
```
brizental, line 367: nit: Can this function just receive an AddonMessage instead?

src/shared/addons/addonapi.cpp
```diff
@@ -0,0 +1,149 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: This is marked as a new file, but it's not really new, it's just moved. Can you fix this? Otherwise we lose the git history for it.

src/shared/externalophandler.cpp
```diff
@@ -60,30 +54,13 @@ bool ExternalOpHandler::request(Op op) {
...
-      vpn->controller()->quit();
-      break;
-
-    // List of no-op requests:
-    case OpNotificationClicked:
-      [[fallthrough]];
-    case OpCloseEvent:
-      break;
-  }
+  Q_ASSERT(m_ops.contains(op));
```
brizental, line 57: When would `m_ops` not contain `op`? Should we actually handle that?

src/apps/unit_tests/testaddon.h
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: This file is not new, just moved. If possible, please fix it.

brizental (changes requested):
I gave this a first pass. I have a suggestion to make it easier to review. One of the problems here is that it is one massive commit. Here is my suggestion to remediate that, note that this would mean commits with broken code, but we will squash it all before merging so I think it doesn't really matter:

1. Put all file moves and sources lists changes in one commit.
2. Put all renames and such in another commit.
3. Put whatever is left (which should be mostly logic changes) in another commit.

EXTRA: I added some notes about moved files that are marked as new. Marking them as moved instead would make it easier as well.

Would you mind making these changes? That would make it easier to review this PR. As of right now I am not very confident on the quality of the review I can make here.

src/apps/vpn/notificationhandler.cpp
bakulf: no because the AddonManager emits a signal when a new addon is created. We do not have the addon-type filtering. Plus addon-types will be removed soon, right?

src/shared/externalophandler.cpp
brizental: Sure, but Q_ASSERT only crashes in debug mode. So I production this will won't crash or will crash weirdly, no?

PR #6417, opened 2023-03-22
VPN-4208: New language view

MattLichtenstein:
## Description

- Abandon the `ToggleCard` design pattern for a more intuitive radio button used for "System language" selection
- Slight tweaks to `MZRadioDelegate` layout to be more robust (wrapping label, correct indicator size)

Screen Shot 2023-03-22 at 2 10 08 PM

## Reference

[VPN-4208: Implement new Language screen](https://mozilla-hub.atlassian.net/browse/VPN-4208)

PR #6440, opened 2023-03-24
Remove qsTrId() from Nebula and *some* soon to be shared screens

lesleyjanenorton:
## Description

This PR
- Removes instances of `qsTrId("...")` from Nebula files
  - Ignores MZToggle, which is up for relocation to vpn/ in #6354 
  - Ignores MZSystemAlert which needs a larger refactor that will likely make the switch to MZI18n unnecessary.
- Removes instances of `qsTrId("..")` from *some* shareable screens 
  - Pretends Give Feedback flow doesn't exist since soon it really won't 
  - Ignores Languages view since #6417 
  - Ignores views like settings/IAP where larger needed refactors will likely make the change unnecessary

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZSignOut.qml
+++ b/nebula/ui/components/MZSignOut.qml
@@ -6,8 +6,7 @@ import Mozilla.Shared 1.0
 import Mozilla.VPN 1.0
 
 MZFooterLink {
-    //% "Sign out"
-    labelText: qsTrId("vpn.main.signOut2")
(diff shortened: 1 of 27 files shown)


bakulf (approved):
R+ but let's see if @flodolo is ok with this change.

flodolo (approved):

PR #6451, opened 2023-03-24
Move the cmake openSSL for android configuration into src/shared

bakulf:


src/shared/platforms/android/sources.cmake
```diff
@@ -32,3 +32,71 @@ else()
         message( FATAL_ERROR "Adjust token cannot be empty for release builds")
     endif()
 endif()
+
+# This section integrates the OpenSSL-Prefab into our build system.  Google
+# Publishes builds of openSSL for android on maven. 
+
+include(ExternalProject)
+ExternalProject_Add(ndk_openssl
+  URL https://maven.google.com/com/android/ndk/thirdparty/openssl/1.1.1q-beta-1/openssl-1.1.1q-beta-1.aar
```
oskirby, line 41: Is there a reason why we're using a beta version of OpenSSL?

oskirby (approved):

src/shared/platforms/android/sources.cmake
strseb: Not openssl version used is beta; the "we compile it and ship it as gradle thing" is marked as beta, given they started offering those builds in 2020.
https://mvnrepository.com/artifact/com.android.ndk.thirdparty/openssl?repo=google

strseb (approved):

PR #6452, opened 2023-03-25
Make pushMessage + webSocket a shareable component - VPN-4442

bakulf:
* A registration push-message type approach
* move tests into the shared unit test app
* Code clean up


src/apps/vpn/mozillavpn.cpp
```diff
@@ -2236,3 +2238,27 @@ void MozillaVPN::registerExternalOperations() {
   eoh->registerExternalOperation(
       OpQuit, []() { MozillaVPN::instance()->controller()->quit(); });
 }
+
+void MozillaVPN::registerPushMessageTypes() {
+  PushMessage::registerPushMessageType(
+      "DEVICE_DELETED", [](const QJsonObject& payload) -> bool {
+        const QString& publicKey = payload["publicKey"].toString();
+        logger.debug() << publicKey;
```
brizental, line 2246: I think this is leftover from a long time ago when I was developing it. Would you mind removing this line?

src/shared/websocket/pushmessage.cpp
```diff
@@ -0,0 +1,78 @@
...
+ *     "as": "payload"
+ *   }
+ * }
+ * ```
+ *
+ * The `type` can be one of a set of known possibilities. See
+ * `PushMessage::registerPushMessageType`, to register new types.
+ *
+ * The payload format will depend on the `type` of the message and may be
+ * omiited. When present,  it will contain information necessary for executing
```
brizental, line 39: nit:
```suggestion
 * omited. When present,  it will contain information necessary for executing
```

src/shared/websocket/pushmessage.h
```diff
@@ -18,26 +17,13 @@ class PushMessage final : public QObject {
 
   bool executeAction();
 
-  enum MessageType {
```
brizental, line 21: With this change we need to update the description of the PushMessageReceived event metric. To make it generic. https://github.com/mozilla-mobile/mozilla-vpn-client/blob/c06ab8e1ce2bc8af81c85032bbeec59ffd776f07/glean/metrics.yaml#L1547

brizental (approved):
r+wc. Just some nits.

PR #6455, opened 2023-03-27
VPN-3690 - Add UUID metric type to qtglean

mcleinman:
## Description

Adding UUID metric type. Tested locally, and working.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-3690

## Checklist
    
### DIFF
--- a/qtglean/include/glean/uuid.h
+++ b/qtglean/include/glean/uuid.h
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef UUID_H
+#define UUID_H
+#include <QObject>
+
+#include "errortype.h"
+
+class UuidMetric final {
+  Q_GADGET
+
+ public:
+  // QML custom types require these three declarations.
+  // See: https://doc.qt.io/qt-6/custom-types.html#creating-a-custom-type
+  UuidMetric() = default;
+  UuidMetric(const UuidMetric&) = default;
+  UuidMetric& operator=(const UuidMetric&) = default;
+
+  explicit UuidMetric(int aId);
+  ~UuidMetric() = default;
+
+  Q_INVOKABLE void set(const QString& uuid) const;
+  Q_INVOKABLE QString generateAndSet() const;
+
+  // Test  only functions
+  Q_INVOKABLE QString testGetValue(const QString& pingName = "") const;
+  Q_INVOKABLE int32_t testGetNumRecordedErrors(ErrorType errorType) const;
+
+ private:
+  int m_id;
+};
+
+#endif  // UUID_H
--- a/qtglean/src/cpp/uuid.cpp
+++ b/qtglean/src/cpp/uuid.cpp
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "glean/uuid.h"
+
+#include <QDebug>
+
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
+#  include "qtglean.h"
+#endif
+
+UuidMetric::UuidMetric(int id) : m_id(id) {}
+
+void UuidMetric::set(const QString& uuid) const {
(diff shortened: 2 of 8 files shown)


qtglean/include/glean/uuid.h
```diff
@@ -0,0 +1,35 @@
...
+
+  Q_INVOKABLE void set(const QString& uuid) const;
+  Q_INVOKABLE QString generateAndSet() const;
+
+  // Test  only functions
+  Q_INVOKABLE QString testGetValue(const QString& pingName = "") const;
+  Q_INVOKABLE int32_t testGetNumRecordedErrors(ErrorType errorType) const;
+
+ private:
+  int m_id;
```
bakulf, line 32: Set a default value: `= 0;`

qtglean/src/cpp/uuid.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+int32_t UuidMetric::testGetNumRecordedErrors(ErrorType errorType) const {
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
+  return glean_uuid_test_get_num_recorded_errors(m_id, errorType);
+#endif
+  return 0;
+}
+
+QString UuidMetric::testGetValue(const QString& pingName) const {
```
bakulf, line 34: You are not using these test methods. Can you land tests together with this PR?

qtglean/include/glean/uuid.h
```diff
@@ -0,0 +1,35 @@
...
+#include "errortype.h"
+
+class UuidMetric final {
+  Q_GADGET
+
+ public:
+  // QML custom types require these three declarations.
+  // See: https://doc.qt.io/qt-6/custom-types.html#creating-a-custom-type
+  UuidMetric() = default;
+  UuidMetric(const UuidMetric&) = default;
```
bakulf, line 18: In the tests we need to check if m_id is correctly copied in this copy ctor.

qtglean/include/glean/uuid.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "errortype.h"
+
+class UuidMetric final {
+  Q_GADGET
+
+ public:
+  // QML custom types require these three declarations.
+  // See: https://doc.qt.io/qt-6/custom-types.html#creating-a-custom-type
+  UuidMetric() = default;
```
bakulf, line 17: Question for @brizental : can js/qml actually create these objects directly?

bakulf (changes requested):

qtglean/include/glean/uuid.h
brizental: Nope, this is not exposed to QML that way. This is required for the expansion of the Q_GADGET macro.

qtglean/src/cpp/uuid.cpp
brizental: These tests are for testing the actual collection of concrete metrics. We should ideally have tests with generic metrics, howeve due to bandwidth and other priorities we did not implement that. It is expected that when new collection is implemented it will have tests and by consequence will test that this API is working. All of this is to say: I would say it is out of scope for this PR, to test this, but we also should leave it here.

qtglean/include/glean/uuid.h
brizental: See my comments above and below, I do not think this is necessary.

qtglean/include/glean/uuid.h
brizental: I would argue we shouldn't do that. 1. Instances of UuidMetric objects should only be created by the glean_parser generated code through the constructor that require an id, so it will always havea value. 2. We are not doing this is any of the other metric implementations.

brizental (approved):

bakulf (approved):
All my comments can be applied as a follow up.

qtglean/src/cpp/uuid.cpp
```diff
@@ -0,0 +1,39 @@
...
+void UuidMetric::set(const QString& uuid) const {
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
+  return glean_uuid_set(m_id, uuid.toLocal8Bit());
+#endif
+}
+
+QString UuidMetric::generateAndSet() const {
+#if not(defined(__wasm__) || defined(BUILD_QMAKE))
+  return glean_uuid_generate_and_set(m_id);
+#endif
```
bakulf, line 24: `return QString();`

mcleinman: Good catch, thanks. Updated.

PR #6476, opened 2023-03-27
Clean up glean types - VPN-4455

bakulf:
There was nothing wrong in the previous implementation. Just 2 things we can do better:
- we exposed a CTOR for every glean type which left `m_id` not initialized. This CTOR can be removed.
- we duplicated/copied the glean types multiple times (in the QT world) every time we use them.

About this second point the issues are:
- why allocating more memory when we can use references/pointers?
- we have all these copy-CTORs without tests
- `m_id` cannot be `const` because of the copy-CTOR.
- all of this copying is because we use Q_GADGET with works with simple struct, copied all the times.

Q_GADGET does not work with pointers. I had to replace it with Q_OBJECT. I also add `Q_DISABLE_COPY_MOVE` to remove unwanted CTORs.

No changes in the API.


brizental:
Please just rebase to include the UUId metric here.

PR #6484, opened 2023-03-28
VPN-4274 Improve OS notifications

Gela:
## Description

Change the notification text for connecting and disconnecting the VPN to distinguish between singlehop and multihop settings. 
Please visit the original issue for the details. The only caveat is that because we don't have a way to track the previous state of the VPN, when we are switching from single to multihop and the other way around, we don't know anything about the previous state. For example, when going from multihop to single hop, we only know we are currently switching to single hop (without knowing if before it was single or multihop). For that reason there is no way to differentiate between multi to single hop and single to single hop scenarios when switching. So for now this PR only addresses changes other than "Switching".

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4274

## Checklist
    
### DIFF
--- a/src/apps/vpn/notificationhandler.cpp
+++ b/src/apps/vpn/notificationhandler.cpp
@@ -196,13 +196,31 @@ void NotificationHandler::showNotification() {
           return;
         }
         // "VPN Disconnected"
-        notifyInternal(None,
-                       I18nStrings::instance()->t(
-                           I18nStrings::NotificationsVPNDisconnectedTitle),
-                       I18nStrings::instance()
-                           ->t(I18nStrings::NotificationsVPNDisconnectedMessage)
-                           .arg(localizedCityName),
-                       NOTIFICATION_TIME_MSEC);
+        ServerData* serverData = vpn->serverData();
+        if (serverData->multihop())
+        {
+            QString localizedEntryCityName =
+                vpn->controller()->currentServer().localizedEntryCityName();
+
+            QString localizedExitCityName =
+                vpn->controller()->currentServer().localizedExitCityName();
+            
+            notifyInternal(None,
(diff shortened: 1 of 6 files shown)


src/apps/vpn/notificationhandler.cpp
```diff
@@ -196,13 +197,32 @@ void NotificationHandler::showNotification() {
...
-                       I18nStrings::instance()
-                           ->t(I18nStrings::NotificationsVPNDisconnectedMessage)
-                           .arg(localizedCityName),
-                       NOTIFICATION_TIME_MSEC);
+        ServerData* serverData = vpn->serverData();
+        if (serverData->multihop()) {
+          QString localizedEntryCityName =
+              vpn->controller()->currentServer().localizedEntryCityName();
+
+          QString localizedExitCityName =
```
bakulf, line 205: You already have this, and it's called `localizedCityName`. Use it instead of creating a new string.

If you dont like the name, you can rename it to `localizedExitCityName` at line 112.

tests/functional/testServerSearch.js
```diff
@@ -119,7 +119,7 @@ describe("Server list", function () {
     assert.strictEqual(vpn.lastNotification().title, "VPN Connected");
     assert.strictEqual(
       vpn.lastNotification().message,
-      `Connected to ${currentCity}, through ${currentCity}`
+      `Connected through ${currentCity} via ${currentCity}`
```
bakulf, line 122: Can you file a Jira issue to have a test for a multi hop notification with 2 different cities?

bakulf (changes requested):
Thanks! Just 1 string renaming.

flodolo (approved):

bakulf (approved):

PR #6486, opened 2023-03-28
VPN-3209: Simplify the 'Subscription Needed' UI

MattLichtenstein:
## Description

- Combine `ViewSubscriptionNeededIAP` and `ViewSubscriptionNeededWeb` into one view (`ViewSubscriptionNeeded`)

## Reference

[VPN-3209: Simplify the 'Subscription Needed' UI for mobile](https://mozilla-hub.atlassian.net/browse/VPN-3209)


src/apps/vpn/translations/strings.yaml
```diff
@@ -775,15 +775,6 @@ subscriptionManagement:
...
     comment: Block of text appearing above a link to learn more about the VPN / Relay bundle offering.
-  valueProp1:
-    value: No logging of your network activity
-    comment: Value proposition used to promote the product
-  valueProp2:
-    value: No one can track you
-    comment: Value proposition used to promote the product
-  valueProp3:
-    value: 500+ servers in 30+ countries
-    comment: Value proposition used to promote the product
```
MattLichtenstein, line 786: We may use these strings at some point in the future, does it make sense to remove them now, and only add them back if we need them again later? Would they have to be translated again if we remove them and add them back in the future?

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeeded.qml
```diff
@@ -0,0 +1,194 @@
...
+            onClicked: isMobile ? VPNPurchase.subscribe(subscriptionOptions.checkedButton.productId) : VPNPurchase.subscribe("web")
+        }
+
+
+        Loader {
+            Layout.fillWidth: true
+            Layout.topMargin: MZTheme.theme.vSpacingSmall
+            Layout.alignment: Qt.AlignHCenter
+
+            active: vpnFlickable.isMobile
```
MattLichtenstein, line 128: @peterbenvenuto the mocks show the ToS and Privacy Policy links as mobile only, but they are currently on desktop as well. Should they remain visible on desktop?

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeeded.qml
```diff
@@ -0,0 +1,194 @@
...
+        MZLinkButton {
+            id: restorePurchase
+
+            Layout.topMargin: 8
+            Layout.alignment: Qt.AlignHCenter
+
+            visible: Qt.platform.os === "ios"
+            labelText: MZI18n.RestorePurchaseRestorePurchaseButton
+            onClicked: VPNPurchase.restore()
+        }
```
MattLichtenstein, line 176: @peterbenvenuto The "Already a subscriber" restore purchase button has always been iOS only, but the mocks show it on desktop. Is this intended for desktop as well

src/apps/vpn/translations/strings.yaml
flodolo: I have a different question: were they ever used?

In general, we shouldn't land strings without code behind, because then we find ourselves in these situations (and already wasted time localizing them).

Please remove them, it's unlikely that we'll use them as is.

flodolo (approved):

src/apps/vpn/mozillavpn.cpp
```diff
@@ -1653,20 +1653,9 @@ void MozillaVPN::registerNavigatorScreens() {
...
-      []() -> bool { return false; });
-
-  Navigator::registerScreen(
-      MozillaVPN::ScreenSubscriptionNeededWeb,
-      Navigator::LoadPolicy::LoadTemporarily,
-      "qrc:/ui/screens/ScreenSubscriptionNeededWeb.qml",
+      "qrc:/ui/screens/ScreenSubscriptionNeeded.qml",
       QVector<int>{App::StateSubscriptionNeeded},
       [](int*) -> int8_t {
         return Feature::get(Feature::Feature_webPurchase)->isSupported() ? 99
```
bakulf, line 1661: You want to remove this part. Replace it with a `return 0`

bakulf (commented):
What about functional tests? Do they still work?

bakulf (changes requested):
See my previous comments.

src/apps/vpn/translations/strings.yaml
MattLichtenstein: They were being used in [ViewSubscriptionNeededIAP.qml](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6486/files#diff-f64cafb5944238cc6a0ddc325f30d14d6c4321eff3668730038097090eb43da8)

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeeded.qml
MattLichtenstein: Spoke with @peterbenvenuto - ToS and Privacy Policy links should be present on all platforms. Mocks have been update and `isMobile` requirement has been lifted

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeeded.qml
MattLichtenstein: Confirmed with @peterbenvenuto - "Already a subscriber" button is an iOS only feature to restore an app store purchase. Mocks have been updated

PR #6488, opened 2023-03-28
VPN-4318 VPN-4477 - Schedule TaskProduct before showing the subscription view, if not done before

strseb:
## Bug Description
This is a race condition on mobile - if the app was started while logged in, we are not going to fetch the products apart from "scheduleRefreshDataTasks" given the mobile app is probably in the background most of the time that is not guaranteed to have happened. 
This pr makes sure the account check will assert we have products.

## Fix: 
- [a108574](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6488/commits/a108574a006172f2c85c883d98030461a2e702fc) is the real fix. We organize all "taskProducts" creations in a single place and make sure to trigger it either when we start auth or randomly go into the needs subscription flow, i.e when the subscription expires. Also we change the UI so it show's a spinner if the Loading of the product's is not yet complete instead of showing an "empty" screen with a subscribe now button that is broken. 
- [64a32da](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6488/commits/64a32daf8a31563a357ab1d623fbfd20123cd566) Useful debugging thing, we should keep that.
- [dfebe11](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6488/commits/dfebe1160c1ce67b8ef1037903b1e7065b889917) this PR would crash as the iAP handler changes the QML model from the android thread, therefore QML will complain and crash. This worked as we so far never had qml bindings when the android thread touched the model. We need to make sure all state-changes are happening in the app's main thread. 

Since i touched UI - please enjoy this screen recording.

https://user-images.githubusercontent.com/9611612/228532525-31af8661-5ec9-437c-9fd7-706c89681c7c.mp4

### DIFF
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -757,14 +757,24 @@ void MozillaVPN::accountChecked(const QByteArray& json) {
   m_private->m_user.writeSettings();
   m_private->m_deviceModel.writeSettings();
 
-  if (m_private->m_user.subscriptionNeeded() && state() == StateMain) {
+  if (!m_private->m_user.subscriptionNeeded() || state() != StateMain) {
+    return;
+  }
+
+  ProductsHandler* products = ProductsHandler::instance();
+  // If the products are loaded, go to the subscription screen,
+  // otherwise scheudle the screen to show up after registration.
+  if (products->hasProductsRegistered()) {
     NotificationHandler::instance()->subscriptionNotFoundNotification();
     maybeStateMain();
     return;
   }
-
-  // To test the subscription needed view, comment out this line:
-  // m_private->m_controller.subscriptionNeeded();
+  logger.debug() << "Scheudling TaskProducts for Subsription View";
+  TaskScheduler::scheduleTask(new TaskProducts());
+  TaskScheduler::scheduleTask(new TaskFunction([this]() {
+    NotificationHandler::instance()->subscriptionNotFoundNotification();
+    maybeStateMain();
+  }));
 }
 
 void MozillaVPN::cancelAuthentication() {
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -572,7 +572,7 @@ void MozillaVPN::completeActivation() {
 
   // Here we add the current device.
   if (option != DeviceStillValid) {
-    addCurrentDeviceAndRefreshData(true);
+    addCurrentDeviceAndRefreshData();
   } else {
     // Let's fetch user and server data.
     scheduleRefreshDataTasks();
@@ -632,7 +632,7 @@ void MozillaVPN::removeDevice(const QString& publicKey, const QString& source) {
   Q_ASSERT(!m_private->m_deviceModel.hasCurrentDevice(keys()));
 
   // Here we add the current device.
-  addCurrentDeviceAndRefreshData(false);
+  addCurrentDeviceAndRefreshData();
 
   // Finally we are able to activate the client.
   TaskScheduler::scheduleTask(new TaskFunction([this]() {
@@ -1178,7 +1178,7 @@ void MozillaVPN::triggerHeartbeat() {
   TaskScheduler::scheduleTask(new TaskHeartbeat());
 }
 
-void MozillaVPN::addCurrentDeviceAndRefreshData(bool refreshProducts) {
+void MozillaVPN::addCurrentDeviceAndRefreshData() {
   TaskScheduler::scheduleTask(
       new TaskAddDevice(Device::currentDeviceName(), Device::uniqueDeviceId()));
   scheduleRefreshDataTasks();
@@ -1211,7 +1211,7 @@ void MozillaVPN::maybeRegenerateDeviceKey() {
 
   // We do not need to remove the current device! guardian-website "overwrites"
   // the current device key when we submit a new one.
-  addCurrentDeviceAndRefreshData(true);
+  addCurrentDeviceAndRefreshData();
   TaskScheduler::scheduleTask(new TaskFunction([this]() {
     if (!modelsInitialized()) {
       logger.error() << "Failed to complete the key regeneration";
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -151,15 +151,15 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
   });
 
   if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
-    // On iAP make sure the Products are loaded in time. 
-    // If we Move into any State adjecent to iAP - load the 
-    // products if we don't have that already. 
-    connect(this, &MozillaVPN::stateChanged, this, [this]() {  
-    if ( (state() == StateSubscriptionNeeded || 
-          state() == StateSubscriptionInProgress || 
-          state() == StateAuthenticating) 
-        && !ProductsHandler::instance()->hasProductsRegistered()) { 
-        TaskScheduler::scheduleTask(new TaskProducts()); 
+    // On iAP make sure the Products are loaded in time.
+    // If we Move into any State adjecent to iAP - load the
+    // products if we don't have that already.
+    connect(this, &MozillaVPN::stateChanged, this, [this]() {
+      if ((state() == StateSubscriptionNeeded ||
+           state() == StateSubscriptionInProgress ||
+           state() == StateAuthenticating) &&
+          !ProductsHandler::instance()->hasProductsRegistered()) {
+        TaskScheduler::scheduleTask(new TaskProducts());
       }
     });
   }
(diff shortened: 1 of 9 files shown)


bakulf (commented):
Let's talk about it tomorrow. I would not want to make this code more spaghetti code than what it is

strseb:
@bakulf - tldr; i kind of re-wrote that :)

src/shared/loghandler.cpp
```diff
@@ -240,10 +240,13 @@ void LogHandler::addLog(const Log& log,
 
   emit logEntryAdded(buffer);
 
-#if defined(MZ_ANDROID) && defined(MZ_DEBUG)
-  const char* str = buffer.constData();
-  if (str) {
-    __android_log_write(ANDROID_LOG_DEBUG, AppConstants::ANDROID_LOG_NAME, str);
+#if defined(MZ_ANDROID)
```
bakulf, line 243: Totally unrelated

src/apps/vpn/mozillavpn.cpp
```diff
@@ -149,6 +149,20 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
     }
   });
 
+  if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
```
bakulf, line 152: Move this into `ProductsHandler`.

src/apps/vpn/mozillavpn.cpp
```diff
@@ -755,14 +765,11 @@ void MozillaVPN::accountChecked(const QByteArray& json) {
...
-  if (m_private->m_user.subscriptionNeeded() && state() == StateMain) {
-    NotificationHandler::instance()->subscriptionNotFoundNotification();
-    maybeStateMain();
+  if (!m_private->m_user.subscriptionNeeded() || state() != StateMain) {
     return;
   }
-
-  // To test the subscription needed view, comment out this line:
-  // m_private->m_controller.subscriptionNeeded();
+  NotificationHandler::instance()->subscriptionNotFoundNotification();
```
bakulf, line 771: Also this can go to ProductsHandler. You can monitor the use subscription state and show the notification

src/apps/vpn/mozillavpn.cpp
```diff
@@ -1018,20 +1025,8 @@ void MozillaVPN::subscriptionStarted(const QString& productIdentifier) {
...
-      TaskScheduler::scheduleTask(new TaskProducts());
-      TaskScheduler::scheduleTask(new TaskFunction([this, productIdentifier]() {
-        subscriptionStarted(productIdentifier);
-      }));
-
-      return;
-    }
-  }
+  ProductsHandler* products = ProductsHandler::instance();
+  Q_ASSERT(products->hasProductsRegistered());
```
bakulf, line 1029: This assertion is not valid anymore. Move this logic in ProductsHandler maybe in a `maybeStartSubscription(productIdentifier)`

bakulf (commented):
Have you removed all the taskProducts from MozillaVpn.cpp? If yes, nice cleanup

bakulf (changes requested):
See my comments.

src/apps/vpn/mozillavpn.cpp
bakulf: What about this?

src/apps/vpn/mozillavpn.cpp
```diff
@@ -319,9 +319,6 @@ void MozillaVPN::initialize() {
   if (m_private->m_user.subscriptionNeeded()) {
     setUserState(UserAuthenticated);
     setState(StateAuthenticating);
-    if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
-      TaskScheduler::scheduleTask(new TaskProducts());
```
bakulf, line 323: we can remove the taskproducts header. Can we?

src/apps/vpn/productshandler.cpp
```diff
@@ -41,6 +45,30 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
...
+         state == App::StateSubscriptionInProgress ||
+         state == App::StateAuthenticating) &&
+        !this->hasProductsRegistered()) {
+      TaskScheduler::scheduleTask(new TaskProducts());
+    }
+  });
+
+  // If the user already has a Subscription
+  // Register so if that changes, we fire a notification.
+  if (!MozillaVPN::instance()->user()->subscriptionNeeded()) {
```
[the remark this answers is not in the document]
bakulf: What about if we have a m_userSubscribed boolean set to false and add a connect. 

Something like:

```
connect(MozillaVPN::instance()->user(), &User::changed, [this] {
   bool userWasSubscribed = m_userIsSubscribed;
   m_userIsSubscribed = MozillaVPN::instance()->user()->subscriptionNeeded();

      if (!MozillaVPN::instance()->user()->subscriptionNeeded()) {
        return;
      }
     
  if (userWasSubscribed && !m_userIsSubscribed) {      NotificationHandler::instance()->subscriptionNotFoundNotification();
    ...
```

bakulf (changes requested):

src/apps/vpn/mozillavpn.cpp
strseb: We still have one Task-Product creation in ``scheduleRefreshDataTasks(bool refreshProducts)``

bakulf: We can remove it. We don't need it right?

birdsarah:
Thank you for tackling this Basti. I haven't had a chance to look at the code. The only thing I would mention is that, probably due to my bad coding, this was a real nightmare when I was working on it. Make sure you've thoroughly run this through it's paces on iOS and on desktop (the web subscription flow).

src/apps/vpn/productshandler.cpp
```diff
@@ -41,6 +45,31 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
...
+    auto state = MozillaVPN::instance()->state();
+    if ((state == App::StateSubscriptionNeeded ||
+         state == App::StateSubscriptionInProgress ||
+         state == App::StateAuthenticating) &&
+        !this->hasProductsRegistered()) {
+      TaskScheduler::scheduleTask(new TaskProducts());
+    }
+  });
+
+  m_last_userSubscriptionNeeded =
```
bakulf, line 62: run this_only_ if state == StateMain othersise `subScriptionNeeded = true`

src/apps/vpn/productshandler.cpp
```diff
@@ -41,6 +45,31 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
...
+      MozillaVPN::instance()->user()->subscriptionNeeded();
+
+  connect(MozillaVPN::instance()->user(), &User::changed, [this] {
+    auto userSubscriptionNeeded =
```
bakulf, line 66: same here

src/apps/vpn/productshandler.cpp
```diff
@@ -41,6 +45,31 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
...
+        MozillaVPN::instance()->user()->subscriptionNeeded();
+    if (!m_last_userSubscriptionNeeded && userSubscriptionNeeded) {
+      NotificationHandler::instance()->subscriptionNotFoundNotification();
```
bakulf, line 69: only if the user is still subscribed. Check `isUserSubscribed`

src/apps/vpn/productshandler.h
```diff
@@ -83,6 +83,8 @@ class ProductsHandler : public QAbstractListModel {
   ProductsHandler(QObject* parent);
   ~ProductsHandler();
 
+  bool m_last_userSubscriptionNeeded = false;
```
bakulf, line 86: true :)

bakulf (changes requested):

src/apps/vpn/productshandler.cpp
```diff
@@ -41,6 +45,36 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
...
+  // products if we don't have that already.
+  connect(MozillaVPN::instance(), &MozillaVPN::stateChanged, this, [this]() {
+    auto state = MozillaVPN::instance()->state();
+    if ((state == App::StateSubscriptionNeeded ||
+         state == App::StateSubscriptionInProgress ||
+         state == App::StateAuthenticating) &&
+        !this->hasProductsRegistered()) {
+      TaskScheduler::scheduleTask(new TaskProducts());
+    }
+    if (state == App::StateMain) {
```
birdsarah, line 60: Totally agree with the previous `if` block. More than agree. It's clever and I really like it.

But, I don't see why the below `if` block is here. It doesn't seem like an obvious place to keep this....meaning I wouldn't know to look here for this logic. What about putting this at the end of `maybeStateMain`?

birdsarah (changes requested):
This looks great. Only one change I think I'd like to see. Let me know what you think.

src/apps/vpn/productshandler.cpp
strseb: I was asked to move that here :) - https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6488#discussion_r1152693162

I think the idea as a first step to move things out of the mozillvpn.cpp that could be reused, for you know that _thing_.
Given that is not happening - how about we move that piece back into it's original place in `void MozillaVPN::accountChecked` - Given that can only happen due to an account check?

birdsarah (approved):

PR #6498, opened 2023-03-28
Add multihop logging

mcleinman:
## Description

I was working a bug (https://mozilla-hub.atlassian.net/browse/VPN-4164) that I have a theory on - multihop is enabled. However, it doesn't seem like we log anything about multihop status. This fixes that moving forward.

## Reference

N/A

## Checklist
    
### DIFF
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -331,6 +331,7 @@ void Controller::activateInternal(DNSPortPolicy dnsPort,
   // For single-hop connections, exclude the entry server
   if (!Feature::get(Feature::Feature_multiHop)->isSupported() ||
       !m_serverData.multihop()) {
+    logger.info() << "Activating single hop";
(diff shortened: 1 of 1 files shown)


Gela (approved):
Thanks! My only question is where it says "not in controller", is that right given that it is still coming from the controller code?

bakulf (approved):

mcleinman:
> Thanks! My only question is where it says "not in controller", is that right given that it is still coming from the controller code?

Thanks for that! When digging in this week, I learned that we do multihop in 2 different ways. Some platforms - like Linux - can do it directly through their VPN controller, but in most we manually do it by setting it up w/ the VPN servers (and it looks the same to the platform's VPN settings). But it is in the controller class, which is confusing - I (hopefully) just made it clearer in the log line that we're talking about the platform's controller.

Gela:
That makes sense, thanks for the explanation and updating the log message 💯

Issue #6394, opened 2023-03-21
Split tunneling applications does not work on Linux


Cimbali:
**Describe the bug**
Applications marked as “excluded” from VPN traffic have no connectivity at all.

**VPN version:**
- Reproduced in 2.13.1 and current master (2.16)
- Also reproduced in 2.14.1 (from launchpad)

**Affected Platforms:**
Tested on:
- Linux OpenSUSE Tumbleweed (20230313)
- Ubuntu 20.04

With wireguard tools 1.0.20210914

**Prerequisites:**
N/A

**Steps to reproduce:**
1. Enable VPN
2. Whitelist an application

Note that whitelisting an application before enabling the VPN also does not work but for other reasons (the message `Unable to mark traffic: no fwmark` is displayed in the logs).

**Expected result:**
The whitelisted application can connect to the internet without its traffic being redirected through the VPN tunnel.

**Actual result:**
The whitelisted application has no connectivity at all.

**Additional notes:**
Here is a step-by-step way to reproduce from the command line, talking directly to the DBus service. Note that this works as the current console (Konsole) is recognised as an application by Mozilla VPN and has its own cgroup.

```sh
> curl -s https://ifconfig.me ; echo
# my public IP
> mozillavpn activate
> curl -s https://ifconfig.me ; echo
185.195.232.159
> dbus-send --system --print-reply --dest=org.mozilla.vpn.dbus / org.mozilla.vpn.dbus.cleanupLogs
method return time=1679404433.756883 sender=:1.1365 -> destination=:1.1921 serial=53 reply_serial=2
> dbus-send --system --print-reply --dest=org.mozilla.vpn.dbus / org.mozilla.vpn.dbus.firewallApp string:/usr/share/applications/org.kde.konsole.desktop string:excluded
method return time=1679404437.799670 sender=:1.1365 -> destination=:1.1922 serial=54 reply_serial=2
   boolean true
> curl --connect-timeout 30 https://ifconfig.me
curl: (28) Failed to connect to ifconfig.me port 443 after 30000 ms: Timeout was reached
> dbus-send --system --print-reply --dest=org.mozilla.vpn.dbus / org.mozilla.vpn.dbus.firewallApp string:/usr/share/applications/org.kde.konsole.desktop string:active
method return time=1679404481.963158 sender=:1.1365 -> destination=:1.1924 serial=55 reply_serial=2
   boolean true
> curl -s https://ifconfig.me ; echo
[...]

PR #6505, opened 2023-03-29
WIP: Fix linux split-tunnelling

Cimbali:
## Description

This fixes an issue identified on Linux, where split-tunnelling systematically fails with the message `Unable to mark traffic: no fwmark`.

More issues are likely outstanding and not yet fully identified, see #6394.

The linux daemon/DBus service activation ([`DBusService::activate()`](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b7ae35e8220dec558cc67c910a16a9c3da290228/src/apps/vpn/platforms/linux/daemon/dbusservice.cpp#L100-L131)) works as follows:
1. check polkit authorisation
2. load configuration
3. set disabled Apps (i.e. mark cgroup traffic)
4. bring up the wireguard interface and configure firewall rules in `Daemon::activate()`

Currently step 3 systematically fails in `NetfilterMarkCgroupV2` with message `Unable to mark traffic: no fwmark`, because the `fwmark` value is passed to netfilter as part of step 4:

https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b7ae35e8220dec558cc67c910a16a9c3da290228/linux/netfilter/netfilter.go#L465-L469

Since `fwmark` is a constant value I’ve modified the code to to pass it on the DBus instantiation. Alternately the 2 last steps of `DBusService::activate()` could be swapped around.

## Reference

#6394

## Checklist
    

oskirby (approved):
This looks good to me, nice find!

Cimbali:
It fixes an issue but as reported in #6394 the split tunnelling still doesn’t work for me even after this fix. Can you try app split-tunnelling with this PR @oskirby, check whether it’s just my setup? I’m a bit stuck at investigating it further myself.

bakulf (approved):

Cimbali:
I occasionally get `NetfilterGo: Netfiler commit failed: conn.Receive: netlink receive: no such file or directory` that prevent split-tunnelling altogether when disconnecting and reconnecting. Might be caused by this patch.

Cimbali:
- Used the listed alternate fix which seems to not trigger the bug I mentioned (to repro on original patch, quickly deactivate / activate VPN)
- Finished fixing #6394 by changing `raw` priority to `mangle` on rules named `mangle`, which means packets beyond the first also get the `fwmark` and connections work as expected

strseb (approved):

PR #6527, opened 2023-03-30
VPN-2991: make timer consistent for iOS

mcleinman:
## Description

`Controller::connected(const QString& pubkey, const QDateTime& connectionTimestamp)` sets the home screen's countup timer to the time since the `connectionTimestamp` (`m_connectedTimeInUTC = connectionTimestamp;`). If that timestamp isn't passed to it or isn't valid, it calls `Controller::resetConnectedTime`, which then sets the start time to the current time (`m_connectedTimeInUTC = QDateTime::currentDateTimeUtc();`).

We were using two different timers to determine the connection time to show on the countup for iOS:
- When creating a new connection, the timer’s start date was not provided, and thus it was set by `resetConnectedTime`.
- When re-launching the app with an active connection, the timer’s start date comes from the iOS system. We’ve saved the tunnel to preferences, and we use `tunnel.connection.connectedDate` for the start date. We don’t control that variable, from [Apple documentation](https://developer.apple.com/documentation/networkextension/nevpnconnection/1406140-connecteddate): “This property contains the date and time when the connection status changed to NEVPNStatusConnected after previously being set to NEVPNStatusDisconnected…”

This PR sets passes `tunnel.connection.connectedDate` upon initial connection, so the app now always uses the same time source for this timer.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2991

## Checklist
    

src/apps/vpn/platforms/ios/ioscontroller.mm
```diff
@@ -93,10 +93,11 @@
             return;
         }
       }
-      callback:^(BOOL a_connected) {
+      callback:^(BOOL a_connected, NSDate* date) {
+        QDateTime qtDate(QDateTime::fromNSDate(date));
```
bakulf, line 97: What about if date is null. The doc does not cover this scenario. Can you please add an if-stmt and use the current date-time?

bakulf (changes requested):

src/apps/vpn/platforms/ios/ioscontroller.mm
mcleinman: I tested this initially, and just re-checked it. It works as expected - when that Swift optional returns nil, we get null all the way up. 

Then in `Controller::connected`, we hit this forking logic:
```
if (connectionTimestamp.isValid()) {
    m_connectedTimeInUTC = connectionTimestamp;
    emit timeChanged();
    m_timer.start(TIMER_MSEC);
  } else {
    resetConnectedTime();
  }
```
And without a valid timestamp, we get the time from `resetConnectedTime()` - which is where it was coming from (on initial iOS connection) before this PR as well.

oskirby (approved):

bakulf (approved):

PR #6530, opened 2023-03-31
VPN-4491 - Create iOS build job that uses cmake

brizental:
Let's have duplicate jobs until qmake is removed for real.


brizental:
@bakulf @oskirby would one of you mind removing build-ios/debug as a required job and adding build-ios/debug/qmake and /cmake?

bakulf (approved):

PR #6532, opened 2023-03-31
VPN-4493 - Use cmake iOS build on XCode cloud

brizental:


bakulf (commented):
What's the reason to replace mozillavpn with the variable?

ci_scripts/ci_post_clone.sh
```diff
@@ -68,9 +31,48 @@ APP_ID_IOS = org.mozilla.ios.FirefoxVPN
...
+# In previous iterations "Mozilla VPN" didn't use to have a space in it,
+# to avoid the pains of migrations we will just rename here. 
+# This is harmless, but still very hacky.
+WORKSPACE_ROOT="$( cd "$(dirname "$0")/.." ; pwd -P )"
+sed -i.bak -E \
+    -e "s/Mozilla VPN/MozillaVPN/" \
+    "${WORKSPACE_ROOT}/CMakeLists.txt"
+sed -i.bak -E \
+    -e "s/MAIN_TARGET mozillavpn/MAIN_TARGET MozillaVPN/" \
+    "${WORKSPACE_ROOT}/CMakeLists.txt"
```
brizental, line 72: @bakulf this is the reason for replacing `mozillavpn` with the variable. I wanted to make this last replace easier.

bakulf: This can be achieved  using cmake flags.

src/CMakeLists.txt
```diff
@@ -43,30 +45,30 @@ endif()
 include(shared/sources.cmake)
 include(apps/vpn/cmake/sources.cmake)
```
bakulf, line 47: This can be called as you like.

src/apps/vpn/cmake/windows.cmake
```diff
@@ -4,25 +4,25 @@
 
 add_definitions(-DWIN32_LEAN_AND_MEAN)
 
-set_target_properties(mozillavpn PROPERTIES
+set_target_properties(${MAIN_TARGET} PROPERTIES
```
bakulf, line 7: This property values is what you want to change. Not the rest.

bakulf (changes requested):

ci_scripts/ci_post_clone.sh
brizental: I have attempted to achieve this by changing multiple variables listed in https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html and https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#target-properties without success in either changing the project name or the mozillavpn target name. 

The objective here is to generate an `.xcodeproj` file with the name `MozillaVPN.xcodeproj` and the main scheme to be also `MozillaVPN`. @oskirby @bakulf please let me know if you know how to do this.

src/CMakeLists.txt
brizental: @bakulf I don't understand what you mean by this comment. Would you mind expanding?

src/apps/vpn/cmake/windows.cmake
brizental: Pleaes see my previous comment. I have not had success on changing the XCode scheme name just by changing a target property.

oskirby:
Closing this as it has been superseeded by #6658.
I cannot thank you enough for your contributions on this @brizental, thank you so much!

PR #6534, opened 2023-03-31
Condition threshold should not ignore the full locale code if supported - VPN-4104

bakulf:


birdsarah (approved):
Am I understanding this correctly that the goal here is:

If the acceptable list of locales is "it" and "fo_BAR" then we should accept "it", "ir_RU", "it_DE", "it*" but only "fo_BAR" not "fo*"

If yes, then I'm happy with this PR. But I will say I don't understand what it has to do with VPN-4104 which is about per add-on threadholds. Looking at the rest of the code am I understanding this is one small part of per addon thresholds? (It would be helpful if you could narrate context in the PR).

PR #6537, opened 2023-04-02
Implement addon-replacer type - VPN-4260

bakulf:
This PR introduces a new addon type: "Replacer". This addon is able to intercept the loading of QRC resources and replace them with something included in its QRC storage. For instance, a manifest like this will replace the home view with the `fs/home.qml` file:

```
{
  ...
  "replacer": {
    "urls": [
      { "request": "qrc:/ui/screens/home/ViewHome.qml", "response": "fs/home.qml" }
    ]
  ...
}
```

But we can also replace/remove/add languages and city/country names. See the functional and unit tests.

### DIFF
--- a/docs/replacer.md
+++ b/docs/replacer.md
@@ -1,0 +1,34 @@
+# Replacer
+
+Replacer addons can be used to replace QRC resources at execution time. The addon must include all the required files in its bundle.
+
+The JSON files are validated via [JSON Schema](https://json-schema.org/) files:
+
+https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/scripts/ci/jsonSchemas
+
+On this page, we want to describe how to write a replacer addon.
+
+### Considerations
+
+ * Replacing QML files can be hazardous! The application will break if the QMLs are broken, or the dependencies are unsatisfied.
+
+ * Replacing an entire folder is even more dangerous! Image what happens if you replace the whole `i18n` folder with broken translation files.
+
+## JSON format
+
+The Replacer JSON files are addon JSON objects with the type "replacer" and a
+"replacer" property object with the following properties:
+
+| Property | Description  | Type | Required |
+| -- | -- | --| -- |
+| urls | An array of Replace objects | Array of objects | Yes |
+
+### Replace object
+
+This object must have the following structure:
+
+| Property | Description  | Type | Required |
+| -- | -- | --| -- |
+| request | The QRC URL to be replaced | string | Yes |
+| response | The relative path of the file/directory to use to replace the `request` | string | Yes |
+| type | The type of the request (`file` or `directory`). By default: `file` | string | No |
--- a/src/apps/unit_tests/resourceloader/languages.json
+++ b/src/apps/unit_tests/resourceloader/languages.json
@@ -1,0 +1,12 @@
+[
+ {
+  "languageCode": "tlh",
+  "languages": {}
+  },
+ {
+  "languageCode": "gr",
+  "languages": {
+   "tlh": "Κλίνγκον γλώσσα"
+  }
+ }
+]
--- a/src/shared/addons/addonreplacer.cpp
+++ b/src/shared/addons/addonreplacer.cpp
@@ -1,0 +1,154 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "addonreplacer.h"
+
+#include <QDir>
+#include <QFile>
+#include <QFileInfo>
+#include <QJsonArray>
+#include <QJsonObject>
+#include <QQmlEngine>
+#include <QScopeGuard>
+
+#include "leakdetector.h"
+#include "logger.h"
+#include "qmlengineholder.h"
+#include "resourceloader.h"
+
+namespace {
+Logger logger("AddonReplacer");
+
+QUrl normalizeUrlScheme(const QUrl& url, const QString& response) {
+  // QT is a strange framework. Sometimes the files are loaded using 'qrc:'
+  // scheme, sometimes, just ':'. Let's normalize the final URL based on the
+  // incoming one.
+
+  if (url.scheme() == "qrc" && response.startsWith(':')) {
+    return QString("qrc%1").arg(response);
+  }
+  return response;
+}
+}  // namespace
+
+// static
+Addon* AddonReplacer::create(QObject* parent, const QString& manifestFileName,
+                             const QString& id, const QString& name,
+                             const QJsonObject& obj) {
+  QJsonObject replacerObj = obj["replacer"].toObject();
+
+  QJsonArray urls = replacerObj["urls"].toArray();
+  if (urls.isEmpty()) {
+    logger.warning() << "Empty URLs";
+    return nullptr;
+  }
+
+  AddonReplacer* replacer =
+      new AddonReplacer(parent, manifestFileName, id, name);
+  auto guard = qScopeGuard([&] { replacer->deleteLater(); });
+
+  QFileInfo manifestFileInfo(manifestFileName);
+  QDir addonPath = manifestFileInfo.dir();
+
+  for (const QJsonValue& urlValue : urls) {
+    QJsonObject urlObj = urlValue.toObject();
+
+    QString urlRequest = urlObj["request"].toString();
+    QString urlResponse = urlObj["response"].toString();
+    QString urlResponseFile(addonPath.filePath(urlResponse));
+
+    if (urlRequest.isEmpty() || urlResponse.isEmpty()) {
+      logger.warning() << "Invalid URL content";
+      return nullptr;
+    }
+
+    if (!urlRequest.startsWith("qrc:")) {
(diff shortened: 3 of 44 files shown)


src/apps/unit_tests/resourceloader/languages.json
```diff
@@ -0,0 +1,9 @@
+[
+ {
+  "wikiDataID": "Q10134",
+  "languageCode": "tlh",
```
oskirby, line 4: Qapla`

bakulf: wo’ batlhvaD

src/shared/resourceloader.h
```diff
@@ -0,0 +1,38 @@
...
+  QString loadDir(const QString& dirName);
+
+  void addUrlInterceptor(QQmlAbstractUrlInterceptor* interceptor);
+  void removeUrlInterceptor(QQmlAbstractUrlInterceptor* interceptor);
+
+ private:
+  ResourceLoader(QObject* parent);
+
+ signals:
+  void cacheFlushNeeded();
```
brizental, line 32: This is a signal that needs to be implemented by users of the loaded resources IIUC. When is it necessary? Please add a docstring here with the explanation as well.

src/shared/resourceloader.h
```diff
@@ -0,0 +1,38 @@
...
+ public:
+  static ResourceLoader* instance();
+
+  ~ResourceLoader();
+
+  QString loadFile(const QString& fileName);
+  QString loadDir(const QString& dirName);
+
+  void addUrlInterceptor(QQmlAbstractUrlInterceptor* interceptor);
+  void removeUrlInterceptor(QQmlAbstractUrlInterceptor* interceptor);
```
brizental, line 26: Missing docs.

src/shared/models/licensemodel.cpp
```diff
@@ -25,6 +26,12 @@ LicenseModel* LicenseModel::instance() {
 
 LicenseModel::LicenseModel(QObject* parent) : QAbstractListModel(parent) {
   MZ_COUNT_CTOR(LicenseModel);
+
+  connect(ResourceLoader::instance(), &ResourceLoader::cacheFlushNeeded, this,
+          [this]() {
+            m_licenses.clear();
+            initialize();
+          });
```
brizental, line 34: It feels fragile that all consumers of the ResourceLoader singleton need to intrument this signal. Any way we can make this not necessary or obviously enforced?

src/shared/addons/addonreplacer.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (type.isEmpty() || type == "file") {
+      if (!QFile::exists(urlResponseFile)) {
+        logger.warning() << "The addon does not contain the file `"
+                         << urlResponseFile << "' required for `" << urlRequest
+                         << "'";
+        return nullptr;
+      }
+
+      replacer->m_replaces.append({eUrlFile, urlRequest, urlResponseFile});
+    } else if (type == "directory") {
```
brizental, line 76: Why not have a recursive check for files under the directory and subdirectories?

src/shared/addons/addonreplacer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  Addon::disable();
+  QmlEngineHolder::instance()->engine()->removeUrlInterceptor(this);
+  ResourceLoader::instance()->removeUrlInterceptor(this);
+}
+
+QUrl AddonReplacer::intercept(const QUrl& url,
+                              QQmlAbstractUrlInterceptor::DataType type) {
+  if (type != DataType::QmlFile) {
+    return url;
+  }
```
brizental, line 130: Why do we need this check?

src/shared/addons/addonreplacer.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  QString urlRequest = url.toString();
+
+  for (const Replace& replace : m_replaces) {
+    if (replace.m_type == eUrlFile) {
+      if (replace.m_request != urlRequest) continue;
+      return normalizeUrlScheme(url, replace.m_response);
+    }
+
+    Q_ASSERT(replace.m_type == eUrlDirectory);
```
brizental, line 140: How can the replacer be a directory, if we return in case `type` is not a QmlFile. Sorry if this doesn't make sense, I am a bit confused.

src/shared/addons/addonreplacer.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (!urlRequest.startsWith(replace.m_request)) continue;
```
brizental, line 141: Hm, what if we are replacing the folder `qrc:/components/mycomponent` and we are faced with the urlRequest `qrc:/components/mycomponent.qml`. Would that replace the qml file with a folder? Do we need to do a file extension check?

brizental (commented):
Since this is a very large PR, it would be easier to review if it were broken in smaller commits. From reading the code, I think there are fours separate things being done here.

- ResourceLoader changes
- Feature to add new i18n resources at execution time
- Replacer addon implementaion
- Docs

Is that correct?

---

I have given this a first pass, I have some questions and concerns. The replacer type is definitely something we want. I am not sure doing it as part of VPN-4260 is the best approach though, unless we are fine blocking that ticket on the broader discussion that the replacer addon type brings.

src/shared/addons/addonreplacer.cpp
bakulf: Because we need to know we are replacing a directory when we do `QDir(":/foo")`. It's not just about having a list of files. We need to support QDir too.

src/shared/addons/addonreplacer.cpp
bakulf: Because this object is also called by `ResourceLoader::loadDir()`.

src/shared/addons/addonreplacer.cpp
bakulf: no because `mycomponent.qml` is not included in `mycomponent`. There is a test for it.

src/shared/addons/addonreplacer.cpp
bakulf: Because so far I limit it to QML files and not to Javascript. We can remove this check.

src/shared/models/licensemodel.cpp
bakulf: nope. But the good thing is that, if we don't do it, the resource cannot be replaced. Not a big deal.

src/shared/resourceloader.h
bakulf: This is called when an interceptor is added or removed. We need to inform the entire app about this operation.

docs/replacer.md
```diff
@@ -0,0 +1,34 @@
...
+
+https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/scripts/ci/jsonSchemas
+
+On this page, we want to describe how to write a replacer addon.
+
+### Considerations
+
+ * Replacing QML files can be hazardous! The application will break if the QMLs are broken, or the dependencies are unsatisfied.
+
+ * Replacing an entire folder is even more dangerous! Image what happens if you replace the whole `i18n` folder with broken translation files.
```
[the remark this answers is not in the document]
bakulf: Definitely this is a powerful addon. But the addons are not open to external developers.
More likely we are not going to replace big sections of the app, but preventing now it is a odd move.

Think about:
- I want to add a new language
- I want to replace the contact us form
- I want to localize more countries and cities
- We need to update the license file as soon as possible for legal reasons.

We don't know how we will use this addon yet. The worst scenario is that we break the frontend. Testing will prevent it.

docs/replacer.md
```diff
@@ -0,0 +1,34 @@
...
+
+The JSON files are validated via [JSON Schema](https://json-schema.org/) files:
+
+https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/scripts/ci/jsonSchemas
+
+On this page, we want to describe how to write a replacer addon.
+
+### Considerations
+
+ * Replacing QML files can be hazardous! The application will break if the QMLs are broken, or the dependencies are unsatisfied.
```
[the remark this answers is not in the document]
strseb: +1 On this. Given that we currently have no way of sandboxing QML - this can be a huge code execution entry point. 

Maybe as a followup we should consider investigating if we can add a sec-context similar to principals, based on the `Q_INVOKABLE`'S `sender()`- so that i.e: 
A component in ://nebula// is not able to Access the VPNSettings Object. But a component in ://vpn can.

src/shared/addons/addonreplacer.cpp
```diff
@@ -0,0 +1,149 @@
...
+    QString urlResponseFile(addonPath.filePath(urlResponse));
+
+    if (urlRequest.isEmpty() || urlResponse.isEmpty()) {
+      logger.warning() << "Invalid URL content";
+      return nullptr;
+    }
+
+    QString type = urlObj["type"].toString();
+    if (type.isEmpty() || type == "file") {
+      if (!QFile::exists(urlResponseFile)) {
```
strseb, line 68: Should this also check that this in fact a `qrc://`? 
Otherwise what could stop me from doing `urlResponseFile="../../../../user_controlled_evil.qml"`

strseb (approved):
r+ wc

PR #6562, opened 2023-04-03
Chore: Correct location of App::State in glean.yaml

birdsarah:
@bakulf this probably isn't necessary given that it's a deprecated file. 

Just curious though. For the metric app_step, for example, where is that going to live? I'm searching in the code base and can't see another metrics.yaml with it in.


bakulf (approved):

lesleyjanenorton:
Closing since we'll soon not have the `shared` folder.

PR #6570, opened 2023-04-04
Import currency symbols from CLDR Unicode github repo - VPN-4306

bakulf:


flodolo (commented):
Is there a specific reason to write _ad hoc_ code on top of CLDR, and not using qt libraries directly?
https://doc.qt.io/qt-5/qml-qtqml-number.html#toLocaleCurrencyString-method

bakulf:
We use this method but we need to pass the symbol.

bakulf:
Note that we do this only as a fallback when the currency symbol of the current locale does not match with the requested currency.

src/apps/vpn/translations/i18n
flodolo, line 1: Not sure this change is expected

flodolo (approved):

strseb (approved):

PR #6575, opened 2023-04-04
VPN-4476 Fix share logs

strseb:
## Description

The jni signature was wrong.

### DIFF
--- a/src/shared/platforms/android/androidcommons.cpp
+++ b/src/shared/platforms/android/androidcommons.cpp
@@ -61,8 +61,7 @@ bool AndroidCommons::verifySignature(const QByteArray& publicKey,
 // static
 bool AndroidCommons::shareText(const QString& text) {
   return (bool)QJniObject::callStaticMethod<jboolean>(
-      UTILS_CLASS, "sharePlainText",
-      "(Ljava/lang/String;)Z",
+      UTILS_CLASS, "sharePlainText", "(Ljava/lang/String;)Z",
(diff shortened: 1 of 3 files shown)


bakulf (approved):

birdsarah (approved):
Sorry did not mean to wipe baku's r+.

bakulf (approved):

src/shared/platforms/android/androidcommons.cpp
```diff
@@ -61,8 +61,7 @@ bool AndroidCommons::verifySignature(const QByteArray& publicKey,
 // static
 bool AndroidCommons::shareText(const QString& text) {
   return (bool)QJniObject::callStaticMethod<jboolean>(
-      UTILS_CLASS, "sharePlainText",
-      "(Landroid/content/Context;Ljava/lang/String;)Z", getActivity().object(),
+      UTILS_CLASS, "sharePlainText", "(Ljava/lang/String;)Z",
```
bakulf, line 64: This is wrong. Shared plaintext should be in the shared android utils. Not in the VPN one. 

Can you move it there? And remove the UTILS macro on top.

bakulf (changes requested):

src/shared/platforms/android/androidcommons.cpp
bakulf: Also launch PlayStore should be in common utils.

PR #6581, opened 2023-04-04
VPN-2542 - Disable sub expiring addon

brizental:
This addon has a bug as described by QA on the ticket. Disabling it -- in a hacky way, we should just have a `enabled: false` field on the manifest.

### DIFF
--- a/addons/message_subscription_expiring/manifest.json
+++ b/addons/message_subscription_expiring/manifest.json
@@ -5,7 +5,7 @@
   "type": "message",
   "conditions": {
     "javascript": "isExpiring.js",
-    "min_client_version": "2.15.0"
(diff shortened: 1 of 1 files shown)


bakulf (approved):

bakulf (changes requested):
Actually we should simply fix the addon.
The bug is that we do not check `isCancelled`. The field `expireOn` has different meanings based on the cancellation state: it can be the date of the next renewal or the date of the subscription expiration.

Adding an if-stmt fixes the problem.

Plus, things can change! We should add a `connectSignal` to monitor the subscriptionData changes.

bakulf:
See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6633

PR #6609, opened 2023-04-06
Move common strings to a shared `strings.yaml`

bakulf:
There a few strings that have to be included in all the P&S products. Instead of duplicating them in all the `strings.yaml`, in this PR I introduce a shared strings.yaml which is "merged" with the product specific ones at build time.

### DIFF
--- a/scripts/utils/generate_strings.py
+++ b/scripts/utils/generate_strings.py
@@ -225,7 +225,7 @@ if __name__ == "__main__":
         metavar="SOURCE",
         type=str,
         action="store",
-        nargs="?",
+        nargs='+',
         help="YAML strings file to process",
     )
     parser.add_argument(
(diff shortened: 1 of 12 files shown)


scripts/utils/generate_strings.py
flodolo, line 1: Only did a quick pass, but this file still has an hard-coded parameter when generating the string ID
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/scripts/utils/generate_strings.py#L61

In fact, the string is generated as

```
    <message id="vpn.inAppMessaging.dateTimeYesterday">
        <location filename="translations/generated/vpn/i18nstrings_p.cpp" line="1122"/>
        <source>Yesterday</source>
    </message>
```

It should probably be a separate input parameter (the alternative is to make assumptions from the path).

strseb (approved):

flodolo (approved):

PR #6676, opened 2023-04-12
Implement the recruitment in-app message - VPN-4342

bakulf:


addons/message_paid_research/manifest.json
```diff
@@ -0,0 +1,58 @@
...
+    "badge": "survey",
+    "blocks": [
+      { "id": "c_1",
+        "type": "text",
+        "content": "We’re exploring ways to enhance the Mozilla VPN experience and want to hear your thoughts."
+      },
+      { "id": "c_2",
+        "type": "text",
+        "content": "This research can be completed right from your home computer."
+      },
```
MattLichtenstein, line 28: Looking at the mockup, it seems these two blocks should be combined into one. Is that correct?

flodolo (approved):
Is it safe to assume that this will only target en-US users (i.e. we can't follow up with folks who don't speak English)?

PR #6678, opened 2023-04-12
VPN-4605: Stop server refresh during tutorials

MattLichtenstein:
## Description

- When a tutorial is started, cancel ongoing server latency calculations and stop the refresh timer

## Reference

[VPN-4605: “Getting started with VPN tutorial” is broken when starting it before the recommended list is fully loaded ](https://mozilla-hub.atlassian.net/browse/VPN-4605)

### DIFF
--- a/src/apps/vpn/serverlatency.cpp
+++ b/src/apps/vpn/serverlatency.cpp
@@ -15,6 +15,7 @@
 #include "models/servercountrymodel.h"
 #include "mozillavpn.h"
 #include "pingsenderfactory.h"
+#include "tutorial/tutorial.h"
 
 constexpr const uint32_t SERVER_LATENCY_TIMEOUT_MSEC = 5000;
 
@@ -42,6 +43,7 @@ ServerLatency::~ServerLatency() { MZ_COUNT_DTOR(ServerLatency); }
 
 void ServerLatency::initialize() {
   MozillaVPN* vpn = MozillaVPN::instance();
+  Tutorial* tutorial = Tutorial::instance();
 
   connect(vpn->serverCountryModel(), &ServerCountryModel::changed, this,
           &ServerLatency::start);
@@ -51,6 +53,15 @@ void ServerLatency::initialize() {
 
   connect(&m_pingTimeout, &QTimer::timeout, this,
           &ServerLatency::maybeSendPings);
+
+  connect(tutorial, &Tutorial::playingChanged, this, [this, tutorial](){
+      if(tutorial->isPlaying()) {
+          stop();
+          m_refreshTimer.stop();
+          return;
+      }
+      m_refreshTimer.start();
+  });
 
   m_refreshTimer.setSingleShot(true);
   connect(&m_refreshTimer, &QTimer::timeout, this, &ServerLatency::start);


src/apps/vpn/serverlatency.cpp
```diff
@@ -52,6 +54,15 @@ void ServerLatency::initialize() {
   connect(&m_pingTimeout, &QTimer::timeout, this,
           &ServerLatency::maybeSendPings);
 
+  connect(tutorial, &Tutorial::playingChanged, this, [this, tutorial](){
+      if(tutorial->isPlaying()) {
+          stop();
+          m_refreshTimer.stop();
```
MattLichtenstein, line 60: @oskirby Decided to stop the timer since QTimer does not have built in pausing functionality, and with the refresh timer being frequent at 30 minutes, I feel it is okay to restart the timer here. Let me know what you think

bakulf (changes requested):
Can you describe what this issue is about? I don't see reasons to have this PR.

MattLichtenstein:
> Can you describe what this issue is about? I don't see reasons to have this PR.

@bakulf servers moving around or appearing/disappearing on the recommended server list while the tutorial is having the user select the first recommended server causes issues.

bakulf:
We should not reordering the list when is visible. It is bad for accessibility, but also in general. As a user, i want to select a server and the list changes while clicking.

We should fix this not just for tutorials. can you ping Santiago?

oskirby:
I would argue that the tutorial is too strict if it is written in a way that fails if the servers move around.

MattLichtenstein:
> We should not reordering the list when is visible. It is bad for accessibility, but also in general. As a user, i want to select a server and the list changes while clicking.

@bakulf I'm not sure that is a change worth making. We have a refresh button on the screen, and it would not make sense to require the user to click refresh and then have to exit and come back to the server list.

Also, the combination of 1) the odds of the user selecting a server while it is refreshing and 2) the impact of accidentally selecting the wrong server during a refresh (super low, they can just go back into the list and select again) and 3) the user ignoring the "checking" verbiage above the list during a refresh is pretty low, also with a low impact

> I would argue that the tutorial is too strict if it is written in a way that fails if the servers move around.

@oskirby this would require changes to the tutorial engine to re-determine what object is being targeted in the event that the current target is destroyed - @bakulf do you think this is reasonable?

albionx:
@bakulf We do allow the server list to refresh and move server location recommendations as the process updates. We thought about this and even had acceptance criteria (in the story) around the list not moving 'super fast'. In practice, the user experience doesn't seem to validate the concern that it would be too fast / jarring.

Also, +1 to Matt's argument that otherwise they user wouldn't see the results and need to hop out and hop into the view again.

@MattLichtenstein I understand your concern but - to keep things simple - what if we just said 'Click on any of the recommended locations' and pointed to the whole list? Items could then move but the target would remain constant (assuming there is a container of some kind for the 5 together). Wdyt?

bakulf:
Then i don't understand the issue. The button has not been clicked but the list is reordering. Why? Let's talk tomorrow to see how to address this issue.

bakulf:
Moving things around is not great and we should avoid it if it's not connected to an user action.


> ... what if we just said 'Click on any of the recommended locations' and pointed to the whole list? Items could then move but the target would remain constant (assuming there is a container of some kind for the 5 together). Wdyt?

This does not work with a moving list. The logic is the following:

1. The tutorial selects the first element in the list to be clicked.
2. The tooltip is positioned next to that element
3. The tutorial engine waits for a click on that element
4. The list is reordered. That element changes position...
5. All breaks.

We should remove 4. But not just for the tutorial engine, but also for accessibility.

I suggest:
1. The list is sorted when it's not visible or when the user clicks on the refresh button.

Thoughts?

albionx:
I think we are fine with the current behavior of self-triggering the refresh, Baku. If the user goes to the recommended tab, why would they need to click the tab and *then* the refresh button? It seems like the intent is obvious from the moment they went to the recommended tab.

Is there any other static element that we could use to anchor the tooltip? Maybe one from which we could then specify an offset, to correct for the misplacement? (prob a bad idea).

Matt's proposal is OK'ishhh to me - the hesitation lies in that this is the only tutorial that is not self-reverting. So getting the user to connect to a recommended server that is not quite our full recommendation is kinda meh. Especially since we plan to (in a future story) lead users to this tutorial from the onboarding flow.

MattLichtenstein:
Closing this PR as a resolution is provided by #6687

PR #6694, opened 2023-04-14
VPN-4600 - include APNs (Apple Push Notification service) in VPN

mcleinman:
## Description

The code is simple, but the risk is moderate - hoping to get at least a couple reviewers on this.

iOS 16.4 gave us some more granular control over `includeAllNetworks`, what we often call "killswitch". We can now exclude cellular services like MMS (which is very helpful), but also we can get granular on push notifications. There is [a new API](https://developer.apple.com/documentation/networkextension/nevpnprotocol/4140516-excludeapns) that defaults to excluding push notifications from the VPN tunnel. 

This PR does two things:
- It changes the default value for the new `excludeAPNs`, so we are now using the VPN for APNs
- It updates us to use Xcode 14.3 for building. This new API requires the iOS 16.4 SDK, which comes with Xcode 14.3 and later.

The risk for this PR is "prevent all of a user's notifications from reaching them". While I can't be sure what the behavior was prior to 16.4 - was APNs automatically excluded from the VPN by Apple, or included by default - the following ar a collection of smaller reasons I'm feeling comfortable with this change:
- issues with APNs, from a brief internet search, generally seem to be from a corporate VPN - not our style
- searched our historical issues for “ios notifications” and “apns” nothing has come up before 
- reading some apple docs - [1](https://support.apple.com/en-us/HT203609), [2](https://support.apple.com/guide/deployment/vpn-settings-overview-dep2d2adb35d/web), [3](https://support.apple.com/guide/deployment/configure-devices-to-work-with-apns-dep2de55389a/web)
- I just tested on my device, on both cellular and wifi. While that's a small test, it's a test.

How we'll know something went wrong - if we have users reporting issues receiving all notifications (not just from our app) on iOS.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4600

## Checklist
    

src/apps/vpn/platforms/ios/ioscontroller.swift
```diff
@@ -214,6 +214,10 @@ public class IOSControllerImpl : NSObject {
             Logger.global?.log(message: "Activating includeAllNetworks")
             proto!.includeAllNetworks = true
             proto!.excludeLocalNetworks = true
+
+            if #available(iOS 16.4, *) {
+                proto?.excludeAPNs = false
```
bakulf, line 219: 1. proto! 
2. Can you write a comment about "false"? I would have written "true".

We want to exclude APNs so...why false?

.xcode-version
```diff
@@ -1 +1 @@
-14.1
\ No newline at end of file
+14.3
```
strseb, line 1: Thank you for keeping this in sync 😅 - 
Currently this file is only for documentation purposes. Meaning we need to ask releng to update this - 
Or (as written in the xcode-cloud doc thing) we have tc auto bootstrap the xcode version noted in this file, which would be my preferred version. 
How quickly do you need this to merge 😅

src/apps/vpn/platforms/ios/ioscontroller.swift
mcleinman: 1. Why should we force unwrap here? I'd rather have us not set the flag than crash, if `proto` for some reason isn't found. Or gracefully handle the situation with a guard statement above. (I know it's a moot point here, as we force unwrap directly above this - but in general, I'd love to move towards having very few force unwraps in our code base.)

APNs exclusion or not is the big question of this PR, as this flag defaults to `true`. My thoughts: The VPN should capture as much traffic as possible. There may be some traffic we can't capture, and should be excluded. (I'm especially less worried about traffic going to the cell provider over the cellular network, as they already know your IP address and such.) In other words: My thought is that the default should be "all traffic in VPN", and we need a good reason as to why not to include something in the VPN. In this case, I can't find a reason to not include APNs traffic - It doesn't seem like including it has delivery reliability issues. In that case, why exclude it?

.xcode-version
mcleinman: Good to know - that is probably why iOS tests are failing, eh? There is no giant rush on this - and per above discussion with Baku, there is some chance this will just be closed out.

bakulf (changes requested):

src/apps/vpn/platforms/ios/ioscontroller.swift
mcleinman: 1. I'm in agreement that there are far better ways to handle this. I'm personally from the "force unwraps should be super-rare in a code base" camp (though I've worked in codebases with other philosophies, and can adapt), and so I'm always reluctant to add a force unwrap to code That said, as we both noted this is made irrelevant by the force unwraps directy above. I fully agree this should be better handled (perhaps with a `guard` above), but that seems like a larger lift than this PR. I made this a force unwrap for now.
 
2. Good call - added that comment.

bakulf (approved):

src/apps/vpn/platforms/ios/ioscontroller.swift
```diff
@@ -214,6 +214,11 @@ public class IOSControllerImpl : NSObject {
             Logger.global?.log(message: "Activating includeAllNetworks")
             proto!.includeAllNetworks = true
             proto!.excludeLocalNetworks = true
+
+            if #available(iOS 16.4, *) {
+                // By default, APNs is excluded from the VPN tunnel on 16.4 and later. We want to include it.
+                proto!.excludeAPNs = false
```
bakulf, line 220: Taskcluster logs say that this property does not exist.

PR #6695, opened 2023-04-15
Linux allow tunnel and vpn containers

Cimbali:
## Description

These changes build on (and require) #6505 and allow to take split-tunnelling further as requested in e.g. mozilla/multi-account-containers#2259

The effect of this PR is to allow mixing non-VPN tabs and VPN tabs in Firefox:
- Without proxy set for the container, tabs of that container bypass the VPN
- With a proxy for the container set to the [wireguard-default proxy](https://mullvad.net/en/help/socks5-proxy/) `socks://10.64.0.1:1080`, tabs access the internet through the VPN location used for the connection
- With a proxy for the container set by multi-account containers to a specific location, tabs access the internet through the specified location

In short, this is done by preventing the socks5 proxies used by mozillavpn/wireguard to be added to the split-tunnelling rules.

Changes are very lightweight (though only for Linux in this PR), and only affect private ranges of addresses (10.x). This means no side-effects even in the unlikely chance of users using the same range if IP addresses as mullvad in their LAN, as local routing rules are always checked before VPN routing (in which case the socks proxies are never reachable to start with).

## Reference

mozilla/multi-account-containers#2259

## Checklist
    

Cimbali:
- Rebased on master now #6505 is merged
- Restrained range of socks relays to 124.0.0.0/20 instead of initially used 124.0.0.0/16
- Moved table initialisation to `nftIfup` so it works after disconnect / reconnect

strseb:
Hey! Sorry for the late review for this - it dropped through the crack, if that happens again always feel free to either `@ping`  me or anyone else from the team :) 

The code looks great, no comments there but this is not the direction we decided to take to allow a selective bypass - there is currently in flight work to provide a localhost proxy only firefox can access, which will provide a non vpn route. 

See: https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/extension/socks5proxy/

Plan is to get that out either v2.25 or v2.26 - at which point mac out also be able to access that :)

Cimbali:
Thanks, I’ll keep an eye out for when that’s merged

PR #6704, opened 2023-04-17
VPN-4406 [Part 1] Add Inspector command to overwrite connection health stability

Gela:
## Description

Add a new inspector command to override connection health stability: stable, unstable and no_signal in mozillavpn.cpp which will be used in functional tests and by QA.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4406

## Checklist
    
### DIFF
--- a/src/apps/vpn/connectionhealth.h
+++ b/src/apps/vpn/connectionhealth.h
@@ -34,6 +34,10 @@ class ConnectionHealth final : public QObject {
   ConnectionStability stability() const { return m_stability; }
 
   void overwriteStabilityForInspector(ConnectionStability stability) {
+    // Only allow overriding the connection health stability in dev mode.
+    if (Constants::inProduction()) {
+      qFatal << "Illegal invokation";
+    }
     m_stabilityOverwritten = true;
     m_stability = stability;
     emit stabilityChanged();
--- a/src/apps/vpn/connectionhealth.h
+++ b/src/apps/vpn/connectionhealth.h
@@ -33,6 +33,12 @@ class ConnectionHealth final : public QObject {
 
   ConnectionStability stability() const { return m_stability; }
 
+  void overwriteStabilityForInspector(ConnectionStability stability) {
+    m_stabilityOverwritten = true;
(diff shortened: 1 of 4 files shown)


src/apps/vpn/connectionhealth.h
```diff
@@ -33,6 +33,12 @@ class ConnectionHealth final : public QObject {
 
   ConnectionStability stability() const { return m_stability; }
 
+  void overwriteStabilityForInspector(ConnectionStability stability) {
+    m_stabilityOverwritten = true;
```
strseb, line 37: Nit: What do you think about checking that this cannot be called outside of dev mode ? :) 
```
if(Constants::inProduction()){
 qFatal << "Illegal invokation" 
}
```

src/apps/vpn/mozillavpn.cpp
```diff
@@ -2134,6 +2134,35 @@ void MozillaVPN::registerInspectorCommands() {
...
+          MozillaVPN::instance()
+              ->connectionHealth()
+              ->overwriteStabilityForInspector(
+                  ConnectionHealth::ConnectionStability::Unstable);
+        } else if (stabilityMode == "nosignal") {
+          MozillaVPN::instance()
+              ->connectionHealth()
+              ->overwriteStabilityForInspector(
+                  ConnectionHealth::ConnectionStability::NoSignal);
+        } else {
```
strseb, line 2161: Nit: should we have something to "reset" the forced value?

strseb (approved):
r+wc

src/apps/vpn/mozillavpn.cpp
Gela: That's what "stable" does, are you thinking of a different way of doing this? I could add another accepted argument "reset" but under the hood it would do the same thing as setting it back to stable.

strseb: I meant like m_stabilityOverwritten=false so that i get "real" stablitty values on the next check

bakulf (approved):

PR #6709, opened 2023-04-17
VPN-4601 Add a functional test for subscription authentication completed

Gela:
## Description

    Add a functional test for subscription authentication completed

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4601

## Checklist
    

src/apps/vpn/ui/screens/subscriptionNeeded/ViewSubscriptionNeeded.qml
```diff
@@ -107,7 +108,7 @@ MZFlickable {
 
         MZButton {
             id: subscribeNow
-            objectName: "vpnSubscriptionNeededView"
+            objectName: "vpnSubscriptionNeededButton"
```
birdsarah, line 111: Much better!

tests/functional/testSubscription.js
```diff
@@ -87,6 +88,77 @@ describe('Subscription manager', function() {
...
+        await vpn.authenticateInApp(true, true);
+
+        // Mark the user subscription as inactive
+        this.ctx.guardianOverrideEndpoints.GETs['/api/v1/vpn/account'].body =
+            userDataInactive;
+
+        await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
+        await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
+
+        // Step 3: Verify that user gets the "Subscribe to Mozilla VPN" screen.
```
birdsarah, line 108: You have only a Step 3.

Clean-up the comments to be consistent.

tests/functional/testSubscription.js
```diff
@@ -87,6 +88,77 @@ describe('Subscription manager', function() {
       }
     };
 
+    it('Mock Desktop Authentication listener success', async () => {
```
birdsarah, line 91: The test name `Mock Descktop Authentication listener success` has two problems:
1) it is describing an implementation detail of the test not the functional feature we're trying to test (this all started with wanting to make sure authenticationCompleted was successfully triggered).
2) it's not following the style conventions of a test name.

Here's some more context so you can pick another name (please let me know if this doesn't make sense):

The mocha framework is trying to lead you in the grammar with `describe() it()`. For example

`describe('My Choir') it('sings the national anthem when the queen enters the room.')`

See the next tests' name  is `it(Prompts SubscriptionNeeded if a user\'s subscription is inactive on connection)` inside the describe('Subscription Manager') describe ('Expired Subscription')`.

So, firstly, I'm not sure this test is specific to expired subscriptions so you may want to move this out block out of inside "Expired Subscription"... I think the double nested describe is unnecessary so would be fine with removing it. We don't have so many tests in here that we need to get overly fussy about grouping them.

I would suggest a test name like `describe('Subscription Manger') it(`Returns user to Main Screen after user successfully completes Web Subscription flow`). There's lots of other possibilities hopefully you get the idea.

Long test names are a feature not a bug.

tests/functional/testSubscription.js
```diff
@@ -105,8 +177,8 @@ describe('Subscription manager', function() {
          await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
 
          // Step 3: Verify that user gets the "Subscribe to Mozilla VPN" screen.
-         await vpn.waitForQuery(
-             queries.screenHome.SUBSCRIPTION_NEEDED.visible());
+         await vpn.waitForQuery(queries.screenSubscriptionNeeded
```
birdsarah, line 180: I have noticed that this test is missing a Step 2!

Please clean this up while you're here.

I think the new test you've written would be best placed after this test because it's kinda adding more building blocks to the same starting point.

If there's a reason they're ordered this way, let me know.

birdsarah (changes requested):
Test itself looks great. Just documentation - naming and comments that need updating.

tests/functional/testSubscription.js
Gela: Thank you, I agree the current test name reflects the underlying implementation more than what the test is doing. I like your suggestion to rename it to `Returns user to Main Screen after user successfully completes Web Subscription flow`

birdsarah (approved):

PR #6716, opened 2023-04-18
VPN-4535: Add metric for web purchases

mcleinman:
## Description

A signal of `subscriptionCompleted` causes us to record this metric in the telemetry file. It was not being triggered by web purchases, in part because of the two different routes to calling `completeActivation`:
- For Android/iOS IAP, a completed IAP calls `MozillaVPN::subscriptionCompleted`, which in turn calls `completeActivation`. 
- Successful web subscriptions call `MozillaVPN::completeAuthentication`, which also calls `completeActivation`. 

Thus, we need to call `subscriptionCompleted` in the Telemetry file (which happens for any `PurchaseHandler`'s `subscriptionCompleted` without hitting `MozillaVPN::subscriptionCompleted`. The work in `mozillavpn.cpp` is to make this more specific for the IAP case.

I've tested this for both iOS and macOS (web):
- Signing into an account with an active subscription
- Fresh subscription
- Opening the app when an account is already signed in

In all cases, this seems to work as expected, with no unexpected side effects that I can find.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4535

## Checklist
    
### DIFF
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -37,6 +37,7 @@
 #include "productshandler.h"
 #include "profileflow.h"
 #include "purchasehandler.h"
+#include "purchaseiaphandler.h"
 #include "qmlengineholder.h"
 #include "releasemonitor.h"
 #include "serveri18n.h"
@@ -197,15 +198,20 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
   if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
     ProductsHandler::createInstance();
   }
+
+#if defined(MZ_ANDROID) || defined(MZ_IOS)
+  PurchaseHandler* purchaseHandler = PurchaseIAPHandler::createInstance();
+  connect(purchaseHandler, &PurchaseIAPHandler::subscriptionCompleted, this,
+          &MozillaVPN::subscriptionCompleted);
+#else
   PurchaseHandler* purchaseHandler = PurchaseHandler::createInstance();
+#endif
   connect(purchaseHandler, &PurchaseHandler::subscriptionStarted, this,
           &MozillaVPN::subscriptionStarted);
   connect(purchaseHandler, &PurchaseHandler::subscriptionFailed, this,
           &MozillaVPN::subscriptionFailed);
   connect(purchaseHandler, &PurchaseHandler::subscriptionCanceled, this,
           &MozillaVPN::subscriptionCanceled);
-  connect(purchaseHandler, &PurchaseHandler::subscriptionCompleted, this,
-          &MozillaVPN::subscriptionCompleted);
   connect(purchaseHandler, &PurchaseHandler::restoreSubscriptionStarted, this,
           &MozillaVPN::restoreSubscriptionStarted);
   connect(purchaseHandler, &PurchaseHandler::alreadySubscribed, this,
--- a/src/apps/vpn/purchasewebhandler.cpp
+++ b/src/apps/vpn/purchasewebhandler.cpp
@@ -44,6 +44,8 @@ void PurchaseWebHandler::startSubscription(const QString&) {
           &MozillaVPN::abortAuthentication);
   connect(taskAuthenticate, &TaskAuthenticate::authenticationCompleted, vpn,
           &MozillaVPN::completeAuthentication);
+  connect(taskAuthenticate, &TaskAuthenticate::authenticationCompleted, this,
+          &PurchaseWebHandler::subscriptionDone);
 
   TaskScheduler::scheduleTask(taskAuthenticate);
 }
@@ -67,4 +69,10 @@ void PurchaseWebHandler::nativeRegisterProducts() {
   logger.error()
       << "nativeRegisterProducts should not be called for PurchaseWebHandler";
   Q_ASSERT(false);
+}
+
+void PurchaseWebHandler::subscriptionDone(const QByteArray& json,
+                                          const QString& token) {
+  // This calls a telemetry event recording
+  emit subscriptionCompleted();
 }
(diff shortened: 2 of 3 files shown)


src/apps/vpn/purchasewebhandler.cpp
```diff
@@ -68,3 +70,9 @@ void PurchaseWebHandler::nativeRegisterProducts() {
       << "nativeRegisterProducts should not be called for PurchaseWebHandler";
   Q_ASSERT(false);
 }
+
+void PurchaseWebHandler::subscriptionDone(const QByteArray& json,
+                                          const QString& token) {
+  // This calls a telemetry event recording
+  emit subscriptionCompleted();
```
Gela, line 77: Would it be helpful to print a `logger.debug()` here?

Gela (approved):

src/apps/vpn/purchasewebhandler.cpp
```diff
@@ -44,6 +44,8 @@ void PurchaseWebHandler::startSubscription(const QString&) {
           &MozillaVPN::abortAuthentication);
   connect(taskAuthenticate, &TaskAuthenticate::authenticationCompleted, vpn,
           &MozillaVPN::completeAuthentication);
+  connect(taskAuthenticate, &TaskAuthenticate::authenticationCompleted, this,
+          &PurchaseWebHandler::subscriptionDone);
```
bakulf, line 48: Instead of calling this method, just emit the signal. Replace `subscriptionDone` with `subscriptionCompleted`

src/apps/vpn/mozillavpn.cpp
```diff
@@ -197,15 +198,20 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
...
+#else
   PurchaseHandler* purchaseHandler = PurchaseHandler::createInstance();
+#endif
   connect(purchaseHandler, &PurchaseHandler::subscriptionStarted, this,
           &MozillaVPN::subscriptionStarted);
   connect(purchaseHandler, &PurchaseHandler::subscriptionFailed, this,
           &MozillaVPN::subscriptionFailed);
   connect(purchaseHandler, &PurchaseHandler::subscriptionCanceled, this,
           &MozillaVPN::subscriptionCanceled);
-  connect(purchaseHandler, &PurchaseHandler::subscriptionCompleted, this,
```
bakulf, line 207: I don't understand how `subscriptionCompleted` arrives to `MozillaVPN` removing this line.

src/apps/vpn/mozillavpn.cpp
```diff
@@ -197,15 +198,20 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
   if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
     ProductsHandler::createInstance();
   }
+
+#if defined(MZ_ANDROID) || defined(MZ_IOS)
+  PurchaseHandler* purchaseHandler = PurchaseIAPHandler::createInstance();
```
bakulf, line 203: Doing this you break the abstraction.

bakulf (changes requested):

src/apps/vpn/mozillavpn.cpp
mcleinman: See longer comment below, but in this approach (which I'm very open to changing) we do not want the web purchase to trigger this.

src/apps/vpn/mozillavpn.cpp
mcleinman: There are a bunch of paths to take on this, all with tradeoffs. I'm definitely open to the idea that I didn't pick the best one for our codebase.

If we don't do something like this, I think we need to change the signal that the telemetry file uses to record IAP completion. From the description of this PR:
```
A signal of subscriptionCompleted causes us to record this metric in the telemetry file. It was not being triggered by web purchases, in part because of the two different routes to calling completeActivation:

- For Android/iOS IAP, a completed IAP calls MozillaVPN::subscriptionCompleted, which in turn calls completeActivation.
- Successful web subscriptions call MozillaVPN::completeAuthentication, which also calls completeActivation.
```
Prior to this PR, the web subscription path never called `MozillaVPN::subscriptionCompleted`, but that is what the telemetry file looked for in recording the metrics.

(It seems like `MozillaVPN::completeAuthentication` is called when starting the app when there is already an active subscription, so we can't use it as a sign that we just had a successful web subscription.)

We don't want web purchases to start triggering `MozillaVPN::subscriptionCompleted`. One possibility would be to create a new signal that the telemetry files uses to record the "subscription complete" metrics - and signal it from both the IAP and the web subscription paths. Does that seem cleaner?

birdsarah: > We don't want web purchases to start triggering MozillaVPN::subscriptionCompleted

Can you elaborate on why?

mcleinman: Apologies for lack of details.

Route for successful web purchases, from `PurchaseWebHandler`:
```
  connect(taskAuthenticate, &TaskAuthenticate::authenticationCompleted, vpn,
          &MozillaVPN::completeAuthentication);
```
Thus, once the web purchase flow starts, the app takes a successful authentication as a sign that the subscription is completed, and continues the flow.

Route for successful IAP, from `MozillaVPN`:
```
connect(purchaseHandler, &PurchaseHandler::subscriptionCompleted, this,
          &MozillaVPN::subscriptionCompleted);
```
WASM, iOS, and Android-specific IAP classes emit a `subscriptionCompleted` to start this path.

As of now, Web-based purchasing flows never emit `subscriptionCompleted`, and thus don't hit `MozillaVPN::subscriptionCompleted`. *Most* of the code within `subscriptionCompleted` is wrapped in `#ifdef`s and wouldn't cause side effects for when run after a successful web-based auth flow - however, both `subscriptionCompleted` and `completeAuthentication` next hit `completeActivation`, and I don't believe we want to run that code twice. 

`completeAuthentication` is called in two situations - when web-based subscriptions succeed, as well as when a typical, already-subscribed user opens the app. So the crux of the problem:
- For web subscriptions, we need to run `completeAuthentication`, which (unless we mess with the login process) will run `completeActivation`.
- For IAP subscriptions, we cannot run `completeAuthentication`, and thus (unless we mess with the login process) it needs to call `completeActivation` on its own path (which currently is `subscriptionCompleted`)

mcleinman:
Closing this in favor of new approach: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6760

PR #6722, opened 2023-04-19
Fix double leading / in mozillavpn.json path property

Cimbali:
## Description

Change `WEBEXT_INSTALL_LIBDIR` to be consistently an absolute path.

Currently uses disagree on whether it is meant to be absolute or relative to /

From linux/mozillavpn.spec:
```
%cmake -DWEBEXT_INSTALL_LIBDIR=/usr/lib -DCMAKE_INSTALL_SYSCONFDIR=/etc -DBUILD_TESTING=OFF
```

From linux/debian/rules:
```
	dh_auto_configure -- -DWEBEXT_INSTALL_LIBDIR=lib -DBUILD_ID=$(DEB_VERSION) -DBUILD_TESTING=OFF
```

This caused rpm to generate paths like `//usr/lib/mozillavpn/mozillavpnnp` which in turn cause Firefox to not find the mozillavpnnp binary.

Using an absolute path be default for `WEBEXT_INSTALL_LIBDIR`, as it is easier when variables point to install dirs in build processes, i.e. `/${lib}` is easier than stripping a leading / from `%{_libdir}`.

## Reference

None

## Checklist
    
### DIFF
--- a/extension/CMakeLists.txt
+++ b/extension/CMakeLists.txt
@@ -20,7 +20,7 @@ set_property(TARGET mozillavpnnp PROPERTY
 ## Handle installation by platform.
 if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     if(NOT DEFINED WEBEXT_INSTALL_LIBDIR)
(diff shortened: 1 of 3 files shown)


oskirby (approved):
Looks good to me

PR #6726, opened 2023-04-19
VPN-4406 [Part 2] Add functional test to verify subscription after a "No signal" state

Gela:
## Description

Add a test where user is logged in, their subscription is active and VPN is on. Then they enter No Signal state (mocked using the inspector command), turn off the VPN and should get the subscription not found screen.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4406

## Checklist
    
### DIFF
--- a/tests/functional/testSubscription.js
+++ b/tests/functional/testSubscription.js
@@ -136,6 +136,41 @@ describe('Subscription manager', function() {
                'VPN is on';
          });
        });
+
+    it('Go to "Subscribe to Mozilla VPN" screen once user toggles off VPN after subscription expires and they enter No Signal',
+       async () => {
+         // This test verifies the case where a user is logged in
+         // and the VPN is on when their subscription expires.
+         // They enter No Signal, and once they toggle the VPN off
+         // they get the "Subscribe to Mozilla VPN" screen.
+
+         await vpn.authenticateInApp(true, true);
+
+         // toggle on VPN here
+         await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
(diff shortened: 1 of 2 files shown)


bakulf (approved):

bakulf (commented):
It looks right but it fails. Do you know why?

Gela:
@bakulf It passes when it's the only test I'm running with `it.only` but fails otherwise. The assertion 

```
    Q_ASSERT(m_private->m_deviceModel.activeDevices() ==
             m_private->m_user.maxDevices());
```
fails in maybeStateMachine() in mozillavpn.cpp. I have a feeling it may be related to the `device` which is used to override subscription status but I'm not sure. I've seen something similar in the past and the solution was to explicitly declare/define the device. I tried declaring a new device with the same properties and it didn't work but I'll give it another try today

tests/functional/testSubscription.js
```diff
@@ -136,6 +136,41 @@ describe('Subscription manager', function() {
...
        });
+
+    it('Go to "Subscribe to Mozilla VPN" screen once user toggles off VPN after subscription expires and they enter No Signal',
+       async () => {
+         // This test verifies the case where a user is logged in
+         // and the VPN is on when their subscription expires.
+         // They enter No Signal, and once they toggle the VPN off
+         // they get the "Subscribe to Mozilla VPN" screen.
+
+         await vpn.authenticateInApp(true, true);
```
bakulf, line 147: We crash here because the previous test leaves the account endpoint set to status 500.
See line 127. 
Because of this, you are not subscribed here.

PR #6748, opened 2023-04-24
Linux improved desktop file search

Cimbali:
## Description

Improve the search for desktop files on Linux by following freedesktop specs closer

- correct user directories (from `$XDG_DATA_HOME` if defined before `$HOME/local/data`) 
- correct system fallbacks (both `/usr/locale/share/applications` and `/use/share/applications`)
- added missing desktop-file hiding mechanisms `OnlyShowIn`, `NotShowIn`, and `Hidden`
- added autostart desktop files to cover applications started at launch (because autostart is implemented by copied desktop files which means they are referenced as different application “ids”)

## Reference

None

## Checklist
    
### DIFF
--- a/src/apps/vpn/platforms/linux/linuxapplistprovider.cpp
+++ b/src/apps/vpn/platforms/linux/linuxapplistprovider.cpp
@@ -35,7 +35,8 @@ void LinuxAppListProvider::fetchEntries(const QString& dataDir,
                                         const QSet<QString>& desktopEnv) {
   logger.debug() << "Fetch Application list from" << dataDir;
 
-  QDirIterator iter(dataDir, QStringList() << "*.desktop", QDir::Files);
+  QDirIterator iter(dataDir, QStringList() << "*.desktop", QDir::Files,
+                    QDirIterator::Subdirectories);
   while (iter.hasNext()) {
     QFileInfo fileinfo(iter.next());
     QSettings entry(fileinfo.filePath(), QSettings::IniFormat);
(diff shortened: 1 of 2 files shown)


oskirby (approved):
This looks great!

Cimbali:
Actually [desktop file ids](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id) in the spec are only the basenames of the desktop files, and conflicting files in different directories should appear as a single application. Looking at auto-start desktop files only makes sense if we keep the full paths as ids. This has the drawback of generating more duplicate entries.

I could change the PR to use basenames instead.

Cimbali:
Small oversight: desktop files are allowed to be in subdirectories too.

This still currently duplicates (some) applications by using full paths as ids.

oskirby:
> Actually [desktop file ids](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id) in the spec are only the basenames of the desktop files, and conflicting files in different directories should appear as a single application. Looking at auto-start desktop files only makes sense if we keep the full paths as ids. This has the drawback of generating more duplicate entries.
> 
> I could change the PR to use basenames instead.

I like the idea of using the basename/desktop ID instead, that seems like the more correct approach and it probably cleans things up and removes ambiguity in the daemon. However, I would argue that we should tackle that in another PR.

Cimbali:
Thanks @oskirby. I’ve put the desktop ID stuff in PR #6782

PR #6751, opened 2023-04-25
Add message_paid_research addons to older branch

birdsarah:
This PR looks to merge two new addons `message_paid_research` and `message_paid_research_old` into a branch I created `addon_releases/2.14_mid_cycle_ship_paid_research`.

We want to immediately ship the `message_paid_research` addons. But we should use the most recent addons that were QA'd and are live in production.

I branched `addon_releases/2.14_mid_cycle_ship_paid_research` at commit 89989986a10522d60abf0f623a7d44704ccaad16. I then added the addons in this PR.

### DIFF
--- a/addons/message_paid_research/currencyCheck.js
+++ b/addons/message_paid_research/currencyCheck.js
@@ -1,0 +1,14 @@
+(function(api, condition) {
+function computeCondition() {
+  if (['USD', 'CAD'].includes(api.subscriptionData.planCurrency)) {
+    condition.enable();
+    return;
+  }
+
+  condition.disable();
+}
+
+api.connectSignal(api.subscriptionData, 'changed', () => computeCondition());
+
+computeCondition();
+});
--- a/addons/message_paid_research/manifest.json
+++ b/addons/message_paid_research/manifest.json
@@ -1,0 +1,55 @@
+{
+  "api_version": "0.1",
+  "id": "message_paid_research",
+  "name": "Participate in paid research to help make VPN better",
+  "type": "message",
+  "translatable": false,
+  "conditions": {
+    "min_client_version": "2.15.0",
+    "translation_threshold": 0,
+    "locales": ["en"],
+    "javascript": "currencyCheck.js"
+  },
+  "javascript": {
+    "enable": "setDate.js"
+  },
+  "message": {
+    "id": "message_paid_research",
+    "title": "Participate in paid research to help make VPN better",
+    "subtitle": "This study will take place in April and May 2023, and you will receive up to $100 USD (or the equivalent in your local currency) as our thanks for your time and feedback.",
+    "badge": "survey",
+    "blocks": [
+      { "id": "c_1",
+        "type": "text",
+        "content": "We’re exploring ways to enhance the Mozilla VPN experience and want to hear your thoughts. This research can be completed right from your home computer."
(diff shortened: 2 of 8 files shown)


addons/message_paid_research/manifest.json
```diff
@@ -0,0 +1,55 @@
+{
+  "api_version": "0.1",
+  "id": "message_paid_research",
+  "name": "Participate in paid research to help make VPN better",
+  "type": "message",
+  "translatable": false,
+  "conditions": {
+    "min_client_version": "2.15.0",
+    "translation_threshold": 0,
```
MattLichtenstein, line 9: is `translation_threshold` a 2.15+ feature? Is that why it's only on this message?
Also, are there any other differences between the messages besides `translation_threshold` and min/max versions?

MattLichtenstein: the 2.14 version will still get released, regardless of translation status as well, correct?

MattLichtenstein (approved):
Branching off at commit https://github.com/mozilla-mobile/mozilla-vpn-client/commit/89989986a10522d60abf0f623a7d44704ccaad16 seems logical to me ✅

addons/message_paid_research/currencyCheck.js
```diff
@@ -0,0 +1,14 @@
+(function(api, condition) {
+function computeCondition() {
+  if (['USD', 'CAD'].includes(api.subscriptionData.planCurrency)) {
```
MattLichtenstein, line 3: Acceptance criteria doesn't mention CAD - should it be included here?

addons/message_paid_research/manifest.json
```diff
@@ -0,0 +1,55 @@
...
+    "locales": ["en"],
+    "javascript": "currencyCheck.js"
+  },
+  "javascript": {
+    "enable": "setDate.js"
+  },
+  "message": {
+    "id": "message_paid_research",
+    "title": "Participate in paid research to help make VPN better",
+    "subtitle": "This study will take place in April and May 2023, and you will receive up to $100 USD (or the equivalent in your local currency) as our thanks for your time and feedback.",
```
MattLichtenstein, line 19: > (or the equivalent in your local currency)


If this is USD-only, is this part of the sentence necessary?

addons/message_paid_research/currencyCheck.js
albionx: It's OK to include Canadian folks, but we don't actually pricing in Canadian dollars, so wouldn't they all be rolling into USD anyway and CAD be an empty set?

addons/message_paid_research/manifest.json
albionx: translation_threshold is indeed only 2.15 and is the reason why there are two addOns for essentially the same thing. There shouldn't be any content differences.

And yes, both should go live and need no translation as they should both be targeting only the 'en' locale. Country disambiguation is happening on the survey itself.

addons/message_paid_research/manifest.json
albionx: Users in Canada and the UK could receive this message (since targeting is only 'en') and would be potentially accepted (only concern is timezone, which applies to UK), and then be paid the equivalent of 100 USD in their local currency.

PR #6760, opened 2023-04-25
VPN-4535 add metric for web purchases

mcleinman:
## Description

A signal of `subscriptionCompleted` causes us to record this metric in the telemetry file. It was not being triggered by web purchases, in part because of the two different routes to calling `completeActivation`:
- For Android/iOS IAP, a completed IAP calls `MozillaVPN::subscriptionCompleted`, which in turn calls `completeActivation`. 
- Successful web subscriptions call `MozillaVPN::completeAuthentication`, which also calls `completeActivation`. 

To have the metric recorded for both routes, we disconnect the telemetry file from `subscriptionCompleted`, and add a new signal that triggers it.

I've tested this for both iOS (IAP) and macOS (web):
- Signing into an account with an active subscription (no IAP completed metric expected)
- Fresh subscription (IAP metric expected with appropriate extra of "sku" whose value is "web" or the IAP ID)
- Opening the app when an account is already signed in (no IAP completed metric expected)

In all cases, this seems to work as expected, with no unexpected side effects that I can find.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4535

## Checklist
    
### DIFF
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -1045,6 +1045,7 @@ void MozillaVPN::restoreSubscriptionStarted() {
 }
 
 void MozillaVPN::subscriptionCompleted() {
+  emit logSubscriptionCompleted();
 #ifdef MZ_ANDROID
   // This is Android only
   // iOS can end up here if the subsciption get finished outside of the IAP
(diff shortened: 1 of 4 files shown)


birdsarah (approved):
Tested desktop purchase.

PR #6768, opened 2023-04-26
Fixes for release promotion on the staging repo

jcristau:
## Description

Release promotion on staging currently fails because:
- on-push decision tasks aren't indexed, so one needs to fill in `previous_graph_ids` manually
- only the prod repo has scopes to send slack notifications

This fixes those issues.

## Reference

N/A

## Checklist
    

taskcluster/mozillavpn_taskgraph/transforms/release_notify.py
```diff
@@ -76,7 +76,6 @@ def add_dependencies(config, tasks):
 
         task.setdefault("dependencies", {}).update(dependencies)
         task.setdefault("if-dependencies", []).extend(list(dependencies.keys()))
-        print(task["dependencies"])
```
ahal, line 79: Oops, thanks!

.taskcluster.yml
```diff
@@ -156,8 +156,8 @@ tasks:
                 routes:
                     $flattenDeep:
                         - checks
-                        - $if: 'level == "3"'
-                          then:
+                        - $if: 'isPullRequest'
+                          else:
```
ahal, line 160: Huh, does this work? TIL

Though I think it might be more readable to write:
```
$if: !isPullRequest
then:
```

.taskcluster.yml
```diff
@@ -156,8 +156,8 @@ tasks:
                 routes:
                     $flattenDeep:
                         - checks
-                        - $if: 'level == "3"'
```
ahal, line 159: I'm a bit concerned about this. For `cron` and `action`, the "project" value comes from build-decision, and we don't set a separate trust project for `staging-mozilla-vpn-client:
https://hg.mozilla.org/ci/ci-configuration/file/ee2e4aa3687f0de4eedd617ecca25494eca720bc/projects.yml#l826

I haven't looked into how build-decision is generating that value, but "project" is the only thing that would differentiate a bunch of the index routes below, e.g:
```
index.mozillavpn.v2.${project}.branch.${head_branch}.latest.taskgraph.decision-${cron.job_name}
```

ahal (changes requested):

.taskcluster.yml
ahal: I'm hoping that project will come from the alias in `projects.yml`, in which case we should be fine.

.taskcluster.yml
```diff
@@ -156,11 +156,11 @@ tasks:
...
+                        - $if: '!isPullRequest'
                           then:
                             - tc-treeherder.v2.${project}.${head_sha}
-                            # TODO Bug 1601928: Make this scope fork-friendly once ${project} is better defined. This will enable
-                            # staging release promotion on forks.
+                            # ${project} uniquely identifies the repo within this trust domain, and each project only has scopes
+                            # to its own index namespace to avoid collisions
```
jcristau, line 163: @ahal how does this look?

ahal: Looks good, thanks!

ahal (approved):

PR #6799, opened 2023-05-01
VPN-3389: Fix no-show verification code error message on ios

lesleyjanenorton:
## Description

https://user-images.githubusercontent.com/22355127/235540280-46109a56-c001-41b6-8313-43f8534d4daf.mov

## Reference

VPN-3389

## Checklist
    
### DIFF
--- a/nebula/ui/components/inAppAuth/MZInAppAuthenticationInputs.qml
+++ b/nebula/ui/components/inAppAuth/MZInAppAuthenticationInputs.qml
@@ -56,7 +56,7 @@ ColumnLayout {
                 Layout.fillWidth: true
                 _placeholderText: _inputPlaceholderText
                 Keys.onReturnPressed: col.submitInfo(textInput)
-                onDisplayTextChanged: if (hasError) hasError = false
+                onTextChanged: if (hasError) hasError = false
             }
 
             MZPasswordInput {


nebula/ui/components/inAppAuth/MZInAppAuthenticationInputs.qml
```diff
@@ -56,7 +56,7 @@ ColumnLayout {
                 Layout.fillWidth: true
                 _placeholderText: _inputPlaceholderText
                 Keys.onReturnPressed: col.submitInfo(textInput)
-                onDisplayTextChanged: if (hasError) hasError = false
+                onTextChanged: if (hasError) hasError = false
             }
```
MattLichtenstein, line 60: Interesting - I would expect these two signal handlers to be the same thing in this case. Do you know why this happens? Qt bug?

MattLichtenstein (approved):

nebula/ui/components/inAppAuth/MZInAppAuthenticationInputs.qml
lesleyjanenorton: I would too and it seems that they are handled the same in all cases apart from on ios. I'm not sure if this is a Qt "nuance" or something exacerbated by our[ special ios focus handling ](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/c647922f806b8b188b25e0d6163039fa689cca67/src/apps/vpn/ui/main.qml#L117).

PR #6809, opened 2023-05-03
VPN-4706, VPN-4709 - Initialize Glean instance from the Android Daemon

brizental:
I have tested this with a sample ping from the daemon and it works i.e. it is able to submit pings and record metrics.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,7 +14,7 @@ buildscript {
     }
     dependencies {
         classpath SharedDependencies.com_android_tools_build_gradle
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.8.10"
         classpath "org.jetbrains.kotlin:kotlin-serialization:1.4.30-M1"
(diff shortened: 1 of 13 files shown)


brizental:
Ok, moving this back to draft just because Is till need to figure out how to sync up the gleanEnabled status between the main app and the daemon, oops.

brizental (commented):
Follow up work: 

- https://mozilla-hub.atlassian.net/browse/VPN-4720
- I would also love to write tests for this. @strseb Do you think it will be possible to test this using the Android functional tests work that you are doing? Let me know and I'll file a ticket and block on your stuff :)

strseb (approved):

strseb:
Re: Testing. 
Instead of using functional test's we can write instrumented test cases. The `daemon project` is already setup to run those :) 
https://developer.android.com/training/testing/instrumented-tests
https://developer.android.com/training/testing/other-components/services

brizental:
> Instead of using functional test's we can write instrumented test cases. The daemon project is already setup to run those 

Nice. I'll write the tests in a follow-up just so I don't block this. https://mozilla-hub.atlassian.net/browse/VPN-4738

PR #6826, opened 2023-05-05
Chore: Truncate logs instead of clearing them

oskirby:
## Description
A common problem that I have encountered when analyzing log files is that the user will report an issue and when we ask for logs, we only receive a file that contains the launch of the client and the user clicking the export logs button. What I suspect is really occuring is that the VPN client is left to run for an extended period of time such that the log file exceeds `LOG_MAX_FILE_SIZE`. Then the next time the log file gets opened, its contents are simply erased and we lose the logs of the issue.

This is probably more common that we expect, since the client never really closes, it just goes into the background. Meaning that many users have the VPN client generating log data for days at a time.

To try and make an attempt to preserve logs, instead of clearing an oversize log file, we should instead make an attempt to truncate the file down to `LOG_MAX_FILE_SIZE / 2` while retaining the tail of the log.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/apps/auth_tests/main.cpp
+++ b/src/apps/auth_tests/main.cpp
@@ -41,7 +41,7 @@ int main(int argc, char* argv[]) {
   settingsHolder.setFeaturesFlippedOn(QStringList{
       "inAppAccountCreate", "inAppAuthentication", "accountDeletion"});
 
-  LogHandler::enableStderr();
+  LogHandler::setStderr(true);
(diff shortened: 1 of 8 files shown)


strseb (approved):
Nice - Can you open up a followup ticket to make sure the Android/iOS daemon are behaving consistent? 
Android at least currently behaves the same way.

PR #6863, opened 2023-05-10
VPN-4707 - Initialize Glean instance of the VPN network extension

brizental:
Opening this as a draft as there is still the piece of sharing Glean state between main app and net ext. I would like to get feedback on my approach to adding the Glean SDK to the network extension with cmake. I don't like the current approach, because it means we build the same library twice... But also other approaches did not work and this works... 👀


brizental:
Closing in favor of #7085.

The approach here was to statically link the Glean Rust library on both the netext and main app. Over on #7085 I took the approach of having a shared library linked to both binaries. I think that is a much better approach.

Issue #6820, opened 2023-05-04
VPN client crashes in staging mode 


data-sync-user:
**Reproduce:**

1. Open the client and go to the developer menu
1. Check staging server
1. Restart the client (it will crash on startup)

**Error:**

`Error: ASSERT failure in localizer: "Languages are out of sync with the translations", file /Users/mozilla/Desktop/mozilla-vpn-client/src/shared/localizer.cpp, line 437 (localizer.cpp:437)`

**Code:**

```
#ifndef UNIT_TEST
  if (!Constants::inProduction()) {
    Q_ASSERT_X(LanguageI18N::instance()->languageExists(code), "localizer",
               "Languages are out of sync with the translations");
  }
#endif
```

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-4718)
┆Reporter: Gela Malek Pour

data-sync-user:
➤ Gela Malek Pour commented:

lo is missing from the Localizer list, I also don’t see it listed in languages.json

data-sync-user:
➤ Gela Malek Pour commented:

This appears to have been a result of submodules being out of sync. Running git submodule update --init resolved the issue.

PR #6867, opened 2023-05-10
VPN-4718: Update languages.json to include Laotian (lo)

oskirby:
## Description
As a part of the localization code, we have a JSON database of languages, and their translations into every other language. This feeds into the languages menu. However, the list of languages is actually generated from the [mozilla-vpn-client-l10n](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n) project, and the two are only loosely coupled together. If the set of available languages falls out of sync then we trigger an assert and the client dies.

When this occurs, we get an error message along the lines of:
```
[09.05.2023 23:33:31.220] Error: ASSERT failure in localizer: "Languages are out of sync with the translations", file /home/runner/work/mozilla-vpn-client/mozilla-vpn-client/src/shared/localizer.cpp, line 436 (localizer.cpp:436)
```

The Laotian language was recently added to the l10n project, but without a corresponding update to the VPN client, this results in a failed assert whenever we are in staging mode. This pull request attempts to add Laotian to `languages.json` but in my opinion we should really move `languages.json` into the l10n project so that the two are always in sync.

## Reference
Github issue #6820 ([VPN-4718](https://mozilla-hub.atlassian.net/browse/VPN-4718))

## Checklist
    
[VPN-4718]: https://mozilla-hub.atlassian.net/browse/VPN-4718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/languagelocalizer/languageLocalizer.js
```diff
@@ -362,7 +362,7 @@ const LanguageLocalizer = {
 
       const cldrLanguages =
           (await this.retrieveJson(
-               `https://raw.githubusercontent.com/unicode-org/cldr-json/main/cldr-json/cldr-localenames-modern/main/${
+               `https://raw.githubusercontent.com/unicode-org/cldr-json/main/cldr-json/cldr-localenames-full/main/${
```
birdsarah, line 365: Curious about this change. Do we need a comment here about use `*-full` for future us.

birdsarah (approved):
Channeling my inner baku with an unsolicited review. This looks good. Agree on your idea for a more-correct solution....but also this won't happen that often and sometimes it's good to pick our battles if we can document "in case of this error, do y"

To that end one thing that would probably help us is if the commit that gets merged has the error that we were seeing in the client then a google search will likely lead us back to our own fix to our own problem.

tools/languagelocalizer/languageLocalizer.js
oskirby: I found that there were slightly more languages available in the `*-full` repository, and this resulted in a slightly more complete language database. This is why we gained some extra translations and currencies for Interlingua and Corsican in this PR.

Gela:
I noticed you added a few other entries such as "co" to languages. Is that because those are in the l10n but not our json file?

oskirby:
@Gela the `languages.json` file is mostly generated by the scripts in `tools/languageLocalizer` so these new translations for Interlingua and Corsican were added automatically by changing the language database from the `modern` to  the `full` URL.

To be honest, I didn't need to change the database URL, but it made the script run with significantly less user intervention.

src/apps/vpn/translations/i18n
flodolo, line 1: I don't think you want to include this change to the submodule.

flodolo (approved):

PR #6878, opened 2023-05-11
VPN-4579 add daemon ping

mcleinman:
## Description

Adds new ping, to set up for future metrics work.

Also updates description for `vpnsession` ping to match latest plans per the [tech spec](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/184419319/VPN+session+pings+-+Tech+spec).

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4579

## Checklist
    
### DIFF
--- a/src/apps/vpn/telemetry/pings.yaml
+++ b/src/apps/vpn/telemetry/pings.yaml
@@ -13,13 +13,27 @@ vpnsession:
   description: |
     Data for one VPN session, defined as a user turning on and eventually
     turning off the VPN in the app. This ping will record start/end datestamp,
-    session-based metrics, etc. It will be sent after the user turns off the
-    VPN, and a end datestamp is recorded.
+    session-based metrics, etc.
   include_client_id: true
   send_if_empty: false
   bugs:
     - https://mozilla-hub.atlassian.net/browse/VPN-4095
   data_reviews:
-    - N/A, will be done for metrics within
+    - https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6130#pullrequestreview-1312098471
+  notification_emails:
+    - vpn@mozilla.com, mcleinman@mozilla.com
+
+daemonsession:
+  description: |
+    Only on iOS and Android.
+    Data for VPN sessions, as recorded from the Android daemon and
+    iOS Network Extension. This ping will record start/end datestamp,
+    session-based metrics, etc.
+  include_client_id: true
+  send_if_empty: false
+  bugs:
+    - https://mozilla-hub.atlassian.net/browse/VPN-4579
+  data_reviews:
+    - ADD AFTER THIS IS DONE - DO NOT MERGE WITH THIS TEXT
   notification_emails:
     - vpn@mozilla.com, mcleinman@mozilla.com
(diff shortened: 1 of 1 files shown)


mcleinman:
# Request for data collection review form

1) What questions will you answer with this data?
This ping sets us up to add some metrics (which will have a different data review form): How long do users keep the VPN active for? Do they run into connectivity issues when they're using it? While there is a [similar ping already approved](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6130), it will not include data for VPN sessions started or ended from the mobile System Settings and a few other non-app situations. This ping will allow us to collect that additional data.

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
We want basic data (how people use the VPN, and how often they run into issues) so that we can build a useful product for how people use it, and fix problems they have.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?
The alternative is continuing to not track this data. Right now, we don't have information like "how often do users switch VPN cities?" and "how long is the average VPN session?". This ping will allow us to add metrics to have rough guesses as to these fairly simple questions.

4) Can current instrumentation answer these questions?
Not as completely. Without this ping, we get much of this data from the `vpnsession` ping - but we are missing an unknowable amount of data.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Data_Collection) found on the Mozilla wiki.   
This request is simply for a ping, which in itself will collect no new data.

6) Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.
This collection will be documented in the Glean Dictionary at https://dictionary.telemetry.mozilla.org/

7) How long will this data be collected?  Choose one of the following:
* I want to permanently monitor this data. Matt Cleinman

8) What populations will you measure?
No filters - all channels, countries, and locales, unless the user has opted out of data collection on that device.

9) If this data collection is default on, what is the opt-out mechanism for users?
When launching the app for the first time, a user is given an option for whether to allow data collection. After this initial set up screen, a user can always toggle data collection permissions in the System Preferences screen.

10) Please provide a general description of how you will analyze this data.
Dashboards that product managers and others will consult on a regular basis.

11) Where do you intend to share the results of your analysis?
Within the Mozilla VPN team and Security and Privacy Products team.

12) Is there a third-party tool (i.e. not Glean or Telemetry) that you are proposing to use for this data collection? If so:
No

mcleinman:
Tagging @travis79 for data-review (my understanding is that a new ping requires data review)

travis79:
> Tagging @travis79 for data-review (my understanding is that a new ping requires data review)

The ping itself shouldn't require a review if you aren't adding any additional metrics or expanding the population this will measure.  If there are metrics that will be sent using this ping, then the data-reviews for them should still be valid despite being sent in a different payload.

travis79:
Which is a good question as to "why" we have a data-review field for pings when it doesn't always apply.

mcleinman:
> Which is a good question as to "why" we have a data-review field for pings when it doesn't always apply.

Thanks @travis79 - removing you from this one.

MattLichtenstein (approved):

PR #6880, opened 2023-05-11
VPN-4761: Prevent toggling app exclusions when VPN is on

MattLichtenstein:
## Description

- Prevent changes to app exclusions while the VPN is on
- Moved InformationCard into Nebula and renamed to MZInformationCard (so it can be used in places outside of `screens/settings`
- Linux only

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/4a631b60-6d6a-4cc5-845b-746d5d9706cc

## Reference

[VPN-4761: [Windows/Linux] 2.15 stop-gap for modifying app exclusions while the VPN is on](https://mozilla-hub.atlassian.net/browse/VPN-4761)


MattLichtenstein:
Closing because the combination of [Fix Linux split tunnel tracking on server switch](https://github.com/mozilla-mobile/mozilla-vpn-client/commit/e3a1f511978f8e4a5763a3258ac7686090b94810) and [VPN-4725: Apply app exclusions when server switching](https://github.com/mozilla-mobile/mozilla-vpn-client/commit/df45eee11152e534d093d9b88b10db3ba1aed87b) remove the need for it

MattLichtenstein:
Re-opening since there are still app exclusion issues wrt silent server switching in 2.15

src/apps/vpn/ui/screens/settings/appPermissions/ViewAppPermissions.qml
```diff
@@ -26,15 +26,44 @@ MZViewBase {
     _menuTitle: MZI18n.SettingsAppExclusionSettings
     _viewContentData: ColumnLayout {
 
+        Loader {
+            Layout.leftMargin: MZTheme.theme.windowMargin
+            Layout.rightMargin: MZTheme.theme.windowMargin
+            Layout.bottomMargin: 24
+            Layout.fillWidth: true
+
+            active: Qt.platform.os === "windows" && VPNController.state !== VPNController.StateOff
```
MattLichtenstein, line 35:
```suggestion
            active: Qt.platform.os === "linux" && VPNController.state !== VPNController.StateOff
```

nebula/ui/components/MZCheckBox.qml
```diff
@@ -71,7 +71,7 @@ CheckBox {
 
             PropertyChanges {
                 target: checkBoxIndicator
-                border.color: checkBox.checked || checkBox.activeFocus ? MZTheme.theme.blueHovered : MZTheme.theme.fontColorDark
+                border.color: if(checBox.enabled) (checkBox.checked || checkBox.activeFocus ? MZTheme.theme.blueHovered : MZTheme.theme.fontColorDark)
```
lesleyjanenorton, line 74: typo in `if (checkBox.enabled)` and then`checkBox.checked || checkBox.activeFocus ? MZTheme.theme.blueHovered : MZTheme.theme.fontColorDark` should be wrapped in curly brackets instead of parenthesis.

MattLichtenstein: fixed the typo but left off the brackets to look consistent with the other property changes since its still a one-liner

flodolo:
A few questions:
* Why are we adding strings to a branch and not to the default branch? This came up in the l10n repository, and it took me a while to figure out its source (it didn't flag me for review, since it's on a branch).
* `VPN must be off to edit App Permissions`.  AFAIK, there is no `App Permissions`, there is `App preferences`?

MattLichtenstein:
> Why are we adding strings to a branch and not to the default branch? This came up in the l10n repository, and it took me a while to figure out its source (it didn't flag me for review, since it's on a branch).

This was meant to be a 2.15 only feature, and resolved in a different way in 2.16, but I will merge it into main as well for the time being

> `VPN must be off to edit App Permissions`. AFAIK, there is no App Permissions, there is `App preferences`?

`App Permissions` and `App preferences` are two different things. `App Permissions` is actually synonymous with `App Exclusions` and `Split Tunneling`. The string should actually say `VPN must be off to edit App Exclusions` but because this was a last minute change, we would not have time to get translations for a new string, so we went with something that we already had translations for.

MattLichtenstein:
> This doesn't match what I saw in the l10n repository though, as this string showed up as new.
>
> This is all the VPN must… strings we have
> https://transvision.flod.org/?recherche=VPN+must&repo=vpn_client&sourcelocale=en&locale=it&search_type=strings_entities

Hmm you're right, I am not seeing it on there, but I have translations for it in my local build 🤔

We definitely used to use the string - would it have anything to do with the fact that is uses the [old translations method](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/7c444fb97b9ad161d240db1db240f2c96a401eff/src/apps/vpn/ui/screens/settings/appPermissions/ViewAppPermissions.qml#L45-L46) (i.e. its not in strings.yaml) 

Screen Shot 2023-05-11 at 12 09 59 PM

PR #6915, opened 2023-05-16
Chore: Android-Cmake Improvements

strseb:
- Patch 1: Cleanup our NDK-Openssl file
- Patch 2: Infer NDK-Compiler from CMAKE_ANDROID_NDK

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    

scripts/cmake/rustlang.cmake
```diff
@@ -85,6 +85,21 @@ function(build_rust_archives)
...
     endif()
 
+    if(ANDROID)
+        if(CMAKE_HOST_SYSTEM_NAME STREQUAL Linux)
+            set(ANDROID_HOST_TAG linux-x86_64)
+        elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+            set(ANDROID_HOST_TAG darwin-x86_64)
+        elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
+            set(ANDROID_HOST_TAG windows-x86_64)
+        endif()
```
Gela, line 95: What if the system host name is not any of these 3, what happens in that case?

strseb: CMAKE_HOST_SYSTEM_NAME - is the "what os are you currently running cmake on?" 

If you have something that is not listed in here, you are running on a pc that should not have access to the android-ndk anyway. 😅 

https://developer.android.com/ndk/guides/other_build_systems#overview

Gela: Makes sense, should there be default `else` statement that errors out just in case?

strseb: Added :)

Gela:
What if the variable path on one's machine is different than what's being set in this change, would that break the build?

If that's the case, should we first check if `ANDROID_HOST_TAG` and other env variables are set first and if not, then attempt to figure it out automatically?

oskirby (approved):

scripts/cmake/rustlang.cmake
```diff
@@ -85,6 +85,31 @@ function(build_rust_archives)
...
+            set(ANDROID_HOST_TAG windows-x86_64)
+        else()
+            message("Unsupported OS?")
+        endif()  
+        # Toolchain.
+        set(ANDROID_TOOLCHAIN_ROOT
+            "${CMAKE_ANDROID_NDK}/toolchains/llvm/prebuilt/${ANDROID_HOST_TAG}")
+       
+        # Rust and Clang disagree about armv7, otherwise we can take the Rust build arch. 
+        if(${RUST_BUILD_ARCH} STREQUAL armv7-linux-androideabi )
```
brizental, line 103:
```suggestion
        if(${RUST_BUILD_ARCH} STREQUAL armv7-linux-androideabi)
```

brizental (approved):

scripts/cmake/rustlang.cmake
brizental: I'll commi this myself and attempt a merge. I think this might change the situation I am in local development right now.

brizental (changes requested):

strseb:
> What if the variable path on one's machine is different than what's being set in this change, would that break the build?
No - Right now if you have NDK_A in CMAKE_ANDROID_NDK and NDK_B in Path, stuff would horribly break with arcane error messages 😅 
So even on a current developer system with ndk on the path, given that we pin the clang-version to be used by setting `CC=` it shoul just reduce error surface. 

> If that's the case, should we first check if ANDROID_HOST_TAG and other env variables are set first and if not, then attempt to figure it out automatically?
Small note: those are non-cached cmake variables, so they are not inferred by your env - nor are they impacting something outside of cmake(as far as i know) - they are just for internal things.

scripts/cmake/rustlang.cmake
```diff
@@ -85,6 +85,31 @@ function(build_rust_archives)
...
+        elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+            set(ANDROID_HOST_TAG darwin-x86_64)
+        elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
+            set(ANDROID_HOST_TAG windows-x86_64)
+        else()
+            message("Unsupported OS?")
+        endif()  
+        # Toolchain.
+        set(ANDROID_TOOLCHAIN_ROOT
+            "${CMAKE_ANDROID_NDK}/toolchains/llvm/prebuilt/${ANDROID_HOST_TAG}")
```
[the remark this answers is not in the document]
strseb: WHAT

Gela (approved):
Either your current change or Bea's simplified code look good!

brizental (approved):

scripts/cmake/rustlang.cmake
```diff
@@ -85,6 +85,20 @@ function(build_rust_archives)
...
+        
+        # Rust and Clang disagree about armv7, otherwise we can take the Rust build arch. 
+        if(${RUST_BUILD_ARCH} STREQUAL armv7-linux-androideabi)
+            set(ANDROID_ARCH_NAME armv7a-linux-androideabi)
+        else()
+            set(ANDROID_ARCH_NAME ${RUST_BUILD_ARCH})
+        endif()
+
+        
+        list(APPEND RUST_BUILD_CARGO_ENV CC=${ANDROID_TOOLCHAIN_ROOT_BIN}/${ANDROID_ARCH_NAME}${ANDROID_NATIVE_API_LEVEL}-clang)
```
brizental, line 99: Wait! We need to change this to

```suggestion
        list(APPEND RUST_BUILD_CARGO_ENV RUSTFLAGS=-Clinker=${ANDROID_TOOLCHAIN_ROOT_BIN}/${ANDROID_ARCH_NAME}${ANDROID_NATIVE_API_LEVEL}-clang)
```

brizental (changes requested):
Sorry to request changes again! I forgot about this last thing

scripts/cmake/rustlang.cmake
strseb: Wait! I don't think they are exclusive. 
I think we should set both, so that if a rust crate calls out to a C compiler, it uses the right one.

brizental (approved):

PR #6916, opened 2023-05-16
Chore: do not search for strings in .ds_store

strseb:
## Description

This folder is a os-detail, we should not attempt to look for strings in there :)


Gela (approved):

bakulf (approved):

oskirby:
Could this be generalized to exclude all hidden directories (eg: those starting with a dot) just in case?

lesleyjanenorton (approved):

PR #6940, opened 2023-05-19
VPN-891 - Replace individual email addresses with team

birdsarah:
Fixes VPN-891

Use vpn-telemetry@mozilla.com that folks can subscribe to and the whole org can look at if needed as opposed to individual email addresses.

### DIFF
--- a/docs/qtglean.md
+++ b/docs/qtglean.md
@@ -11,7 +11,7 @@ Glean API docs on the book, choose the "Firefox Desktop" examples tab. The API w
 
 > **Note**: An understanding of the fundamentals of Glean such as metrics, pings and glean_parser is assumed.
 
-> **Note**: For a list of Glean features/metrics that are not yet implemented on the Mozilla VPN Glean integration,
+> **Note**: For a list of Glean features/metrics that are not yet implemented on the Mozilla VPN Glean integration, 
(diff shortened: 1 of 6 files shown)


bakulf (approved):

brizental (commented):
Note: It's [suggested](https://mozilla.github.io/glean/book/reference/yaml/metrics.html#notification_emails) that there is always an individual's email on the `notification_emails` list alongside a team.

birdsarah:
> Note: It's [suggested](https://mozilla.github.io/glean/book/reference/yaml/metrics.html#notification_emails) that there is always an individual's email on the `notification_emails` list alongside a team.

Thanks @brizental. I've updated to match this.

docs/qtglean.md
```diff
@@ -110,7 +110,7 @@ example:
     data_sensitivity:
       - technical
     notification_emails:
-      - brizental@mozilla.com
+      - brizental@mozilla.com, vpn-telemetry@mozilla.com
```
brizental, line 113: I don't think this is[ valid YAML list syntax](https://yaml.org/spec/1.2.2/#block-sequences). IIUC, this should be 
```suggestion
      - brizental@mozilla.com
      - vpn-telemetry@mozilla.com
```

brizental (changes requested):

brizental (approved):

PR #6948, opened 2023-05-19
Add some unit test coverage for connection health

csadilek:
## Description

This may seem a bit random :). I wanted to see what it "feels like" to add some unit test coverage to existing components, and picked the first function I came across that seemed interesting. I couldn't convince myself easily that the EWMA calculation works correctly just by looking at the code so I tried to write a test for it. (Spoiler: Yes, it all works correctly)

I chose the path of minimal refactoring and give the tests access to some internals, which I think is fine. The alternative would be to make these internals publicly accessible, which doesn't seem right as it would make the components fragile. The trade-off here is of course that the tests are then more fragile i.e., internal changes of the component can break the tests, but that seems better than exposing those internals via public API. I figured I put this up as a PR, because even in case we don't want to land it in this shape, it's a good discussion to have.

## Checklist
    
### DIFF
--- a/src/apps/vpn/connectionhealth.cpp
+++ b/src/apps/vpn/connectionhealth.cpp
@@ -17,30 +17,6 @@
 #include "mozillavpn.h"
 #include "telemetry/gleansample.h"
 
-// In seconds, the time between pings while the VPN is deactivated.
-constexpr uint32_t PING_INTERVAL_IDLE_SEC = 15;
-
-// In seconds, the timeout for unstable pings.
-constexpr uint32_t PING_TIME_UNSTABLE_SEC = 1;
-
-// In seconds, the timeout to detect no-signal pings.
-constexpr uint32_t PING_TIME_NOSIGNAL_SEC = 4;
-
-// Packet loss threshold for a connection to be considered unstable.
-constexpr double PING_LOSS_UNSTABLE_THRESHOLD = 0.10;
-
-// Destination address for latency measurements when the VPN is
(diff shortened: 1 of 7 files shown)


src/apps/vpn/connectionhealth.cpp
```diff
@@ -216,13 +193,17 @@ void ConnectionHealth::dnsPingReceived(quint16 sequence) {
...
   logger.debug() << "Received DNS ping:" << latency << "msec";
+  updateDnsPingLatency(latency);
+}
 
+void ConnectionHealth::updateDnsPingLatency(quint64 latency) {
   if (m_dnsPingInitialized) {
     m_dnsPingLatency *= (PING_BASELINE_EWMA_DIVISOR - 1);
     m_dnsPingLatency += latency;
-    m_dnsPingLatency /= PING_BASELINE_EWMA_DIVISOR;
+    m_dnsPingLatency = round(static_cast<float>(m_dnsPingLatency) / PING_BASELINE_EWMA_DIVISOR);
```
csadilek, line 206: This is a bit pendatic on my part. We were just truncating before, happy to reset this.

csadilek:
Looks like the test fails on Windows. 🎯 💯 :) Setting up a Window dev. environment too see what's going on.

csadilek:
Alright, unit tests are 🟢 here now.

bakulf (approved):

oskirby (approved):
Looks good! Thank you

PR #6949, opened 2023-05-19
VPN-4834: Disable free trial feature

MattLichtenstein:
## Description

- Disable the free trial feature (which was only currently supported on iOS) 
- Left code in tact in case we want to bring this feature back

## Reference

[VPN-4834: [iOS] 7-day free trial text still displayed in the onboarding screen](https://mozilla-hub.atlassian.net/browse/VPN-4834)


mcleinman (approved):
👍🏻  Thanks!

PR #6972, opened 2023-05-23
Chore: Full app Hot Reload

strseb:
## Description

Hot Reloading has a catch, it will currently only reload the current "screen" you are in. 
Meaning if you want to hot reload the Navigation Bar, because you are hunting down a crash - you cannot. 

This PR add's the ability to `hot.js` to press `r` and just reload the whole window. This will reload & recreate all objects inside the window. 

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/9611612/58929640-e497-4b9b-ab06-af263930a1cd

### DIFF
--- a/src/shared/inspector/inspectorhotreloader.cpp
+++ b/src/shared/inspector/inspectorhotreloader.cpp
@@ -138,7 +138,7 @@ void InspectorHotreloader::reloadWindow() {
   }
   logger.error() << "Closing and full reloading window";
 
-  int x, y = 0;
+  int x = 0, y = 0;
   // We may have multiple closed windows still in here.
   // So best to go over them all.
   QList<QObject*> rootObjects = engine->rootObjects();


src/shared/inspector/inspectorhotreloader.cpp
```diff
@@ -123,3 +126,39 @@ void InspectorHotreloader::resetAllFiles() {
...
+  // Here is the main QML file.
+  if (!engineHolder->hasWindow()) {
+    logger.error() << "No Window to reload";
+    return;
+  }
+  logger.error() << "Closing and full reloading window";
+
+  int x = 0, y = 0;
+  // We may have multiple closed windows still in here.
+  // So best to go over them all.
```
Gela, line 143: By saying we "may have multiple closed windows", do you mean lingering windows from other parts of the VPN client while clicking around?

Gela (approved):

src/shared/inspector/inspectorhotreloader.cpp
strseb: I mean from previous "reloads", as we will create a new Window and the last one may or may not have been deleted yet :)

PR #6973, opened 2023-05-23
VPN-4761: Prevent toggling app exclusions when VPN is on

MattLichtenstein:
## Description

- Cherry-picked from 2.15 (#6880)

## Reference

[VPN-4761: [Linux] 2.15 stop-gap for modifying app exclusions while the VPN is on](https://mozilla-hub.atlassian.net/browse/VPN-4761)


lesleyjanenorton (approved):

src/apps/vpn/translations/strings.yaml
```diff
@@ -155,6 +155,9 @@ splittunnel:
     comment:
       - This displays as informational text below split tunneling settings. The link directs to a sumo article which will help certain users who are having network issues using this feature.
   infoLinkText: Learn more
+  infoCardDescription: 
+    value: VPN must be off to edit App exclusions
```
flodolo, line 159: Since you pointed out that this is not an actual menu, do we need the uppercase `A`?

flodolo (approved):

PR #6979, opened 2023-05-23
Fix thinking CMakeLists.txt is an rcc addon

gabrielBusta:
## Description

Before we assumed only addon directories would be included under the `addons` path. This patch updates the logic to "this is an addon if it is a directory with a manifest.json file."

## Reference

N/A

## Checklist
    
### DIFF
--- a/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
@@ -17,7 +17,12 @@ def add_addons_release_artifacts(config, tasks):
             task["attributes"]["build-type"] == "addons/opt"
             and task["name"] == "addons-bundle"
         ):
-            addons = set(os.listdir("addons"))
+            addons = set(
+                name
+                for name in os.listdir("addons")
+                if os.path.isdir(os.path.join("addons", name))
+                and os.path.isfile(os.path.join("addons", name, "manifest.json"))
+            )
             for addon in addons:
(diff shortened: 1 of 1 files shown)


taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -141,9 +146,13 @@ def add_beetmover_worker_config(config, tasks):
             "shipping-phase": shipping_phase,
         }
```
gabrielBusta, line 148: This was `black`

taskcluster/mozillavpn_taskgraph/transforms/beetmover.py
```diff
@@ -141,9 +146,13 @@ def add_beetmover_worker_config(config, tasks):
...
 
-        dest = f"{archive_url}{destination_paths[0]}" if destination_paths else archive_url
+        dest = (
```
gabrielBusta, line 149: These are just formatting changes caused by running black on the files.

ahal (approved):
Thanks for taking this!

oskirby:
Thank you for catching this! And sorry for causing the issue with the build refactoring in the first place.

ahal:
The tasks keep getting canceled, but I can't merge unless they pass. This is blocking the addon release candidates, so if someone could merge for me I'd appreciate it!

bakulf (approved):

PR #6982, opened 2023-05-24
[RFC] Proposal to adopt an RFC process

lesleyjanenorton:
## Description

This PR proposes the adoption of an RFC process and submits an inaugural RFC describing the rationale for doing so. The goal here is to get alignment on adopting the process. We can follow up with an RFC template, additional notes on process, and modify any part of this as needed going forward. 

## Checklist
    
### DIFF
--- a/docs/rfcs/0001-rfc-process.md
+++ b/docs/rfcs/0001-rfc-process.md
@@ -1,6 +1,8 @@
 - Status: Proposed
-- Date: 2023-05-23
-- RFC PR: TODO
+- Date: 2023-05-29
+- Author: [@lesleyjanenorton](https://github.com/lesleyjanenorton)
+- RFC PR: [#6981](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6982)
+- Implementation GitHub issue: Not applicable. Merging [#6981](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6982) implements the proposal.
 
 
 ## Summary
@@ -31,24 +33,29 @@ RFCs will be submitted and reviewed as pull requests in this repository. If appr
 
 The high-level process to create an RFC is as follows:
 
-* Create an RFC document (like this one). (TODO: templates, maybe)
-* Open a pull request for the RFC document.
-* Ask for feedback on the pull request.
+* Create an RFC document (like this one, in Markdown).
+* Open a pull request for the RFC document and prefix the title of the PR with `[RFC]`.
+* File an issue requesting feedback on the pull request.
 
 During the lifetime of an RFC:
 
 * Discussions happen asynchronously on the pull request; anyone may engage.
-* Build consensus (TODO: agreement?) and integrate feedback.
+* Integrate feedback, build consensus, get stakeholder approval. All comments, not just those of the stakeholder, should be addressed and resolved.
 
 After the discussion phase has concluded:
 
-* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
-* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
+* If consensus is reached, the RFC is considered "accepted".
+* If consensus is not reached, the RFC is considered "rejected".
+* The status of the RFC is updated accordingly and is merged into the repository for documentation purposes.
+
+Note:
+* _Consensus_ in the bullets above should be understood as the general support of the team and stakeholder buy-in.
+* Rejected RFC proposals may be revived should the requirements change in the future.
 
 Authors may implement and submit the feature as a pull request once the RFC is accepted.
 
 ## Drawbacks
-- (?)
+- I can't think of any
 
 ## Questions
 - About nomenclature: Are we ok calling these 'RFCs'? Firefox Monitor and Firefox Relay refer to these as ADRs but 'RFC' strikes me as more to the point.
--- a/docs/rfcs/0001-rfc-process.md
+++ b/docs/rfcs/0001-rfc-process.md
@@ -28,7 +28,6 @@ RFCs will be submitted and reviewed as pull requests in this repository. If appr
  - Closer to the code, easier to find
  - In keeping with how other teams do it (TODO link evidence)
 
-
 ## Guide-level explanation
 
 The high-level process to create an RFC is as follows:
--- a/docs/rfcs/0001-rfc-process.md
+++ b/docs/rfcs/0001-rfc-process.md
@@ -1,0 +1,63 @@
+- Status: Proposed
+- Date: 2023-05-23
+- RFC PR: TODO
+
+
+## Summary
+Introducing a process for proposing, reviewing, and documenting decisions with high-impact implications for our codebase.
+
+## Motivation
+
+RFCs authored by this team are currently submitted and reviewed in a number of places, namely Confluence and Google Docs.
+
+Several challenges have been noted with this practice:
+- Docs are accessible to Mozillians only
+- Platform constraints make asynchronous discussions difficult and impossible to reference later (Confluence) 
+- Uneven awareness and ability for interested parties to participate in decisions as they're being made (Google Docs) 
+
+We need a more inclusive, more collaborative process for making decisions that impact our codebase.
+
+## Proposed Solution
+
+RFCs will be submitted and reviewed as pull requests in this repository. If approved, they'll be stored as documentation.
+
+**Salient points in favor**
+ - Viewable by all, not just Mozilla-internals
+ - Closer to the code, easier to find
+ - In keeping with how other teams do it (TODO link evidence)
+
+
+## Guide-level explanation
+
+The high-level process to create an RFC is as follows:
+
+* Create an RFC document (like this one). (TODO: templates, maybe)
+* Open a pull request for the RFC document.
+* Ask for feedback on the pull request.
+
+During the lifetime of an RFC:
+
+* Discussions happen asynchronously on the pull request; anyone may engage.
+* Build consensus (TODO: agreement?) and integrate feedback.
+
+After the discussion phase has concluded:
+
+* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
+* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
+
+Authors may implement and submit the feature as a pull request once the RFC is accepted.
+
+## Drawbacks
+- (?)
+
+## Questions
+- About nomenclature: Are we ok calling these 'RFCs'? Firefox Monitor and Firefox Relay refer to these as ADRs but 'RFC' strikes me as more to the point.
+- About templates: I'm in favor of including an RFC template as an example but I'd prefer we not be overly prescriptive about it, at least early on.
+- About documentation living elsewhere: What to do about design documents currently housed in Confluence, if anything? Are there documents we feel must be migrated over to GitHub and stored as RFCs? (Does not require a decision now) 
+
+## Prior art
+[Rust RFCs - Active RFC List](https://rust-lang.github.io/rfcs/)
+
+[Firefox Android Components RFCs](https://github.com/pocmo/android-components/tree/master/docs/rfcs)
+
+[Firefox Monitor ADRs](https://github.com/mozilla/blurts-server/tree/main/docs/adr)
--- a/docs/rfcs/0001-rfc-process.md
+++ b/docs/rfcs/0001-rfc-process.md
@@ -1,4 +1,4 @@
-- Status: Proposed
+- Status: Accepted
 - Date: 2023-05-29
 - Author: [@lesleyjanenorton](https://github.com/lesleyjanenorton)
 - RFC PR: [#6981](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6982)
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,5 @@
 personal_ws-1.1 en 211 utf-8
+ADRs
 APIs
 Adblock
 AdjustSdk
@@ -32,6 +33,7 @@ Lerche
 Linters
 LottieAnimation
 MERCHANTABILITY
+Mozillians
 MPL
 MSI
 Menubar


docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+The high-level process to create an RFC is as follows:
+
+* Create an RFC document (like this one). (TODO: templates, maybe)
+* Open a pull request for the RFC document.
+* Ask for feedback on the pull request.
+
+During the lifetime of an RFC:
+
+* Discussions happen asynchronously on the pull request; anyone may engage.
+* Build consensus (TODO: agreement?) and integrate feedback.
```
brizental, line 41: I think aiming for consensus is a trap here. There should be at least one key stakeholder for each RFC. We can base ourselves on the component owners and when an RFC doesn't have a clear owner fallback to the tech lead. The key stakeholder yes would need to sign off on the RFC, but overall consensus IMO should not be necessary.

Although redundant, for documentation purposes and future insurance I think it would be worth stating here that ALL comments should be addressed and resolved prior to approving an RFC, not only stakeholder comments. But when there is controversy, it should simply fallback to the key stakeholders and the RFC authors to reach an agreement.

brizental: There is a hole in this that even applies to this RFC, lol. Who to add as a key stakeholder when the tech lead is the author and there is no clear component owner? I guess in this case no stakeholders and the tech lead just feels the vibe? I'd be fine with that.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+
+After the discussion phase has concluded:
+
+* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
+* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
+
+Authors may implement and submit the feature as a pull request once the RFC is accepted.
+
+## Drawbacks
+- (?)
```
brizental, line 51: Personally, I don't see any drawbacks.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Questions
+- About nomenclature: Are we ok calling these 'RFCs'? Firefox Monitor and Firefox Relay refer to these as ADRs but 'RFC' strikes me as more to the point.
+- About templates: I'm in favor of including an RFC template as an example but I'd prefer we not be overly prescriptive about it, at least early on.
```
brizental, line 55: As I said above, I think this document itself is template enough at least for now.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+- About documentation living elsewhere: What to do about design documents currently housed in Confluence, if anything? Are there documents we feel must be migrated over to GitHub and stored as RFCs? (Does not require a decision now)
```
brizental, line 56: I think that would be unnecessary.

brizental (commented):
Very excited about this!

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
+- Status: Proposed
+- Date: 2023-05-23
+- RFC PR: TODO
```
bakulf, line 3: Add the author here.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+ - Viewable by all, not just Mozilla-internals
+ - Closer to the code, easier to find
+ - In keeping with how other teams do it (TODO link evidence)
+
+
+## Guide-level explanation
+
+The high-level process to create an RFC is as follows:
+
+* Create an RFC document (like this one). (TODO: templates, maybe)
```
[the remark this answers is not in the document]
bakulf: I vote to have a `0000-template.md`.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+
+* Discussions happen asynchronously on the pull request; anyone may engage.
+* Build consensus (TODO: agreement?) and integrate feedback.
+
+After the discussion phase has concluded:
+
+* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
+* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
+
+Authors may implement and submit the feature as a pull request once the RFC is accepted.
```
bakulf, line 48: I suggest to have the following "states":
- Proposed
- Rejected
- Accepted (which means that the RFC has not implemented)
- Applied/Implemented

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
+- Status: Proposed
+- Date: 2023-05-23
+- RFC PR: TODO
+
```
bakulf, line 4: Even if this is private, I wonder if we want to add a link to Jira/Github too

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
+* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
+
+Authors may implement and submit the feature as a pull request once the RFC is accepted.
+
+## Drawbacks
+- (?)
+
+## Questions
+- About nomenclature: Are we ok calling these 'RFCs'? Firefox Monitor and Firefox Relay refer to these as ADRs but 'RFC' strikes me as more to the point.
```
bakulf, line 54: +1 for RFC

bakulf (approved):

docs/rfcs/0001-rfc-process.md
Gela: We have 23 tech specs listed in Confluence, I like the idea of having everything in one place for future references. I can move them over if we decide to do that, it should be fairly quick

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
+- Status: Proposed
```
Gela, line 1: If the PR is active, I imagine we can assume it is in the "proposed" status. Once the reviewers approve, it'll become approved but then the author has to go back and create another commit to change this status. Does it make more sense to have the status in the body of the PR so it's easy to change?

brizental: You can change with a suggestion like:
```suggestion
- Status: Accepted
```
Which is easy enough IMO. There is always the possibility of an RFC being declined... Which would be nice to see right as you open it. 

One thing though, we should not run CI for PRs with the RFC prefix. Is that possible? cc @strseb We probably still want to run linters for spellcheck, but othewise it's all unnecessary.

strseb: Good callout. So we can stop the ci to run - however not by PR name but for example for pr's on a `rfc/*` branch. 
But we cannot remove the branch protection rules per PR. 
Meaning: 
- Only gh admins can merge those.
- We need a diffrent repo to store them.

brizental: Got it! I mean... We could always make every engineer in the team a GH admin @lesleyjanenorton 👀

docs/rfcs/0001-rfc-process.md
lesleyjanenorton: and maybe
- Deprecated ?

Gela:
Do we expect authors to still add their specs to [this chart](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/127369241/Tech+Specs)? Or is there going to be an index of everything somewhere?

docs/rfcs/0001-rfc-process.md
oskirby: Personally, I have always found the IETF's use of "Request for Comments" kind of odd as they sort of imply that the discussion is never really settled, but I am cool with calling these RFCs.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,63 @@
...
+
+During the lifetime of an RFC:
+
+* Discussions happen asynchronously on the pull request; anyone may engage.
+* Build consensus (TODO: agreement?) and integrate feedback.
+
+After the discussion phase has concluded:
+
+* If consensus(TODO: agreement?) is reached, the RFC is considered "accepted" and merged into the repository for documentation purposes.
+* If consensus(TODO: agreement?) is not reached, the RFC is considered "rejected" and the pull request is closed. Rejected RFC proposals may be revived should the requirements change in the future.
```
oskirby, line 46: There is value in committing the rejected RFCs to the repository too, they serve as a history of things that have already been brought up and decided against.

docs/rfcs/0001-rfc-process.md
csadilek: We don't necessarily need an explicit status in the RFC. It could be implicit which is how we use it on Android: 
- Opening the PR is PROPOSED 
- Merging the PR is ACCEPTED
- Closing the PR (unmerged, but with summary of outcome) is DECLINED

However, if we prefer having the DECLINED RFCs merged for the purpose of documentation and having it all in one place, I think that's fine too! We just have to remember to update the status before merging.

docs/rfcs/0001-rfc-process.md
csadilek: Good idea, I think a link to an issue would be nice to track the implementation of the RFC, if applicable.

lesleyjanenorton:
> 1. Each RFC should be tracked on Jira on a Spike ticket, so that we can allocate time for them properly on our sprints.

@brizental Sounds good. Given the access limitations with Jira though, I would prefer stipulating the creation of a GitHub issue and then just dealing with the Jira curation. What do you think?

docs/rfcs/0001-rfc-process.md
lesleyjanenorton: Agreed. Will update.

docs/rfcs/0001-rfc-process.md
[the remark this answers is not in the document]
csadilek: I also thought this one could act as a template and essentially be the 0000-rfc-*, but we can also refine this later.

docs/rfcs/0001-rfc-process.md
csadilek: I am not sure about the Applied/Implemented state, because that requires updating the RFC weeks, perhaps months later when the implementation lands. I think it's doable but also seems like something that will be forgotten :). How about we add a link to a Github issue instead to track the implementation for an accepted RFC? Then the current status can always be looked up and we don't have to remember updating the original RFC doc.

docs/rfcs/0001-rfc-process.md
csadilek: An advantage of calling it ADR would be that we use the same name as e.g., Monitor, Application Services, etc. 

A disadvantage of calling it ADRs is the confusion I've seen that it's not necessarily clear if the authors are just documenting a decision, or if they are actively seeking and accepting feedback. The name RFC doesn't seem to cause that confusion. I think we can make that clear either way though.

csadilek (approved):
Thank you, it's great to see this! I left a few comments for consideration too, but I think we can iterate on this as we go along and gain experience.

docs/rfcs/0001-rfc-process.md
csadilek: I think we want to aim at building consensus in the sense that all team members find the proposal acceptable enough to support it. That doesn't mean that we always have to unanimously agree that it's the best of all approaches. Just sharing, reviewing, and talking through will help a lot though in my experience, and making sure all feedback is heard and addressed.

RFC decisions can also be reversed in case the implementation doesn't pan out as expected.

My recommendation would be try it out and see how often we reach a point that requires a decision maker :). I'd expect this to be very rare.

brizental: I agree it would be sensitive to point out that consensus is something we aim for. I also agree controversial RFCs would be the outliers. I still think the document should be clear that it's not a blocker if consensus is not reached.

docs/rfcs/0001-rfc-process.md
brizental: I see value in merging declined RFCs, FWIW.

docs/rfcs/0001-rfc-process.md
lesleyjanenorton: I like templates too. I'd like to merge this as-is and then we can add a template in a follow-up as I imagine there would be additional discussion needed around which sections to include.

docs/rfcs/0001-rfc-process.md
lesleyjanenorton: I think we should consider these on an individual basis. If they are essential/helpful, I'd recommend we bring them over as documentation but perhaps not as RFCs. I do think it would be helpful to add links to future RFC documents to the list in Confluence, as we do for a number of the documents still living in Google Docs.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Proposed Solution
+
+RFCs will be submitted and reviewed as pull requests in this repository. If approved, they'll be stored as documentation.
+
+**Salient points in favor**
+ - Viewable by all, not just Mozilla-internals
+ - Closer to the code, easier to find
+ - In keeping with how other teams do it (TODO link evidence)
+
```
brizental, line 30: nit: 
```suggestion
```

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Guide-level explanation
+
+The high-level process to create an RFC is as follows:
+
+* Create an RFC document (like this one, in Markdown).
+* Open a pull request for the RFC document and prefix the title of the PR with `[RFC]`.
+* File an issue requesting feedback on the pull request.
```
brizental, line 38: This feels backwards. Filing the ticket shold be the first step. Especially for people inside the team, given this will need to be added to our sprints during spring planning. Also, it's worth specifying "file a public ticket".

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,70 @@
...
+
+* If consensus is reached, the RFC is considered "accepted".
+* If consensus is not reached, the RFC is considered "rejected".
+* The status of the RFC is updated accordingly and is merged into the repository for documentation purposes.
+
+Note:
+* _Consensus_ in the bullets above should be understood as the general support of the team and stakeholder buy-in.
+* Rejected RFC proposals may be revived should the requirements change in the future.
+
+Authors may implement and submit the feature as a pull request once the RFC is accepted.
```
brizental, line 55: This feels rushed. The next step after the RFC is approved should be to file a ticket / story / epic to do the work and not just file a pull request.

brizental (approved):
Minor comments, but overall r+.

docs/rfcs/0001-rfc-process.md
lesleyjanenorton: Aren't all tickets filed in Jira copied over to GitHub unless explicitly given the 'no-sync' label? I'm fine with adding this stipulation regardless.

Gela (approved):

strseb (approved):

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Drawbacks
+- I can't think of any
```
mcleinman, line 57: Only one I can think of is there are privacy/security concerns for some of it, as it's public.

Gela: The VPN codebase is public so I don't think we're really exposing anything that's not already out there, not sure how we feel about public designs that involve Guardian though.

PR #6989, opened 2023-05-25
VPN-4253: Give Daemon::activate a failure signal

strseb:
Right now, if `bool Daemon::activate` was called and fails, it will return false. 
This function during execution does change the system state: Device Statuses, Routing Tables, Firewall rules etc, so just aborting can introduce unwanted side effects. 
This pr adds a scope guard that unless we deactivate it (after a successful run) will emit a signal that we can use to roll back any transactional state. 
I'm using that to make sure the windows firewall rules cannot be active if the activation was not successful, that will address VPN-4253. 

We might want to consider rolling back other part's but i'm not that confident in the daemon space to make suggestions. :D


bakulf:
I would move this logic in the caller of `activate`. it seems that the bug is there.
Who call activate should be in charge of reverting the changes in case of a failure

PR #7000, opened 2023-05-27
Add Slovene and Turkish to languages.json / Fix CI

lesleyjanenorton:
This PR adds Slovene and Turkish ("sl" and "tr") to languages.json. ~~The changes here were generated by the [localization tool](https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/tools/languagelocalizer), which required a bit of duct taping before it would Just Work™. Ergo, I am suspicious of some of the deletions, particularly those around L81.~~

This should unblock the functional tests. See[ this PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6867) for an excellent summary of why.

## Reference

VPN-4874, VPN-4875
## Checklist
    

flodolo:
> which required a bit of duct taping before it would Just Work™

It would help to explain that the changes were, because this doesn't look right.

It's losing an entire language (Corsican), and losing several language names across different locales, but not everywhere (interestingly enough, all of them with underscores, like es_MX)

src/apps/vpn/translations/extras/languages.json
```diff
@@ -78,32 +78,7 @@
       "XXX": "¤"
     },
     "languages": {
-      "bg": "Lingua bulgara",
```
bakulf, line 81: This part is suspicious... I wonder why so many lines deleted.

bakulf (changes requested):
Usually what I did was to reintroduce deleted lines. No particular reasons, but just to have more language coverage.
The issue is that Wikipedia changes really often and there is not a good review process for these language details.

src/apps/vpn/translations/extras/languages.json
lesleyjanenorton: Agreed. Reverted these changes.

lesleyjanenorton:
@flodolo Deletions reverted. I'm not sure if they were caused by some hacking I did to get the script running locally and/or changes to wikidata or the unicode [CLDR Project repository](https://github.com/unicode-org/cldr-json). Agree about the documentation and filed VPN-4875 for clarification.

bakulf (approved):

PR #7021, opened 2023-05-30
Chore: Swap MSVC and Clang build

strseb:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    

taskcluster/ci/build/windows.yml
```diff
@@ -56,10 +56,10 @@ windows/opt:
                 - secrets:get:project/mozillavpn/tokens
             default: []
 
-windows/clang-cl:
+windows/opt:
```
ahal, line 59: This could potentially impact downstream release tasks, but I see they both have the `unsigned.zip` in the `release-artifacts` key.. so I think it should just work

PR #7022, opened 2023-05-30
Nit: Reduce Android builds created for PRs

strseb:
## Description
We are building 6x the Android Client - yet most of those builds are unused or checking the same thing. 
So let's trim the task's we are creating :)  

- Remove Android-Armv-7 Next ( never been used )
- Remove unused debug-intel builds
- Reduce the 'next' builds to only main (not critical if they break)
- Build a fake release build on pr's (so we know main does not explode)


hneiva (approved):
CI changes look good

PR #7028, opened 2023-05-30
Revert "Bump taskcluster-taskgraph from 5.1.1 to 5.2.0 in /taskcluster"

csadilek:
Reverts mozilla-mobile/mozilla-vpn-client#7010

This caused all builds to fail with: https://firefox-ci-tc.services.mozilla.com/tasks/CgsuoED1S8WWMUYsGf348A/runs/0/logs/live/public/logs/live.log


lesleyjanenorton (approved):

csadilek:
The [failure](https://github.com/mozilla-mobile/mozilla-vpn-client/runs/13526469355) in `build-ios/debug/cmake` has been around intermittently for a while so we can merge if nothing else shows up. `test-taskgraph-definition` passes again and fails consistently on main otherwise.

PR #7040, opened 2023-05-31
[gh-ci] Add a Task that will update i18n automatically

strseb:
## Description

Our handling of the i18n repo is a bit weird. 

We have removed the auto update-to-i18n/main from our build system - the CI will still do that, to make sure we have latest strings in there. 
Developers need to manually update the submodule- but apparently we are not commiting those updates upstream. 

So the last time we updated the submodule was 3 weeks ago
- https://github.com/mozilla-mobile/mozilla-vpn-client/tree/main/src/apps/vpn/translations

This can lead to weird scenarios outside of CI. It's hard to know what commit of the vpn is build with what version of the strings. 

This PR will add a workflow that: 
 1. Update the i18n repo.
 2. File a pr. 
 3. Approve the pr. 
 4. Mark is as auto merge, which means, if CI does pass, this will be merged automatically. 

Right now this will only run on a manual trigger but once we know it's working we can move this to a daily-cronjob.

### DIFF
--- a/.github/workflows/i18n_update.yaml
+++ b/.github/workflows/i18n_update.yaml
@@ -1,0 +1,54 @@
+name: i18n Updater
+# This Workflow will periodically go in: 
+# 1. update the i18n repo.
+# 2. file a pr. 
+# 3. approve the pr. 
+# 4. Mark is as auto merge, which means, if CI does pass, this will be merged automatically. 
+
+
+on:
+  # Triggers the workflow on pull request events but only for the main branch
+  # For now let's only have it manually, until we know this works! 
+  # schedule:
+  #  - cron: '0 5 * * *'
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  main:
+    name: Update Main Branch Languages
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


lesleyjanenorton (approved):

.github/workflows/i18n_update.yaml
```diff
@@ -0,0 +1,54 @@
...
+# 2. file a pr. 
+# 3. approve the pr. 
+# 4. Mark is as auto merge, which means, if CI does pass, this will be merged automatically. 
+
+
+on:
+  # Triggers the workflow on pull request events but only for the main branch
+  # For now let's only have it manually, until we know this works! 
+  # schedule:
+  #  - cron: '0 5 * * *'
```
brizental, line 13: Is there a ticket to validate the job and enable this? If there is, would you mind adding a link to it here?

brizental (approved):

.github/workflows/i18n_update.yaml
strseb: Ah i should have noted: i played with this on my fork. 
-> Task Run: https://github.com/strseb/mozilla-vpn-client/actions/runs/5135309815
-> Created and merged this PR: https://github.com/strseb/mozilla-vpn-client/pull/65 

But created VPN-4905 to test this and make it a cron once it works :)

PR #7042, opened 2023-05-31
ci: upgrade Taskgraph and dependencies

ahal:
## Description

  The `test-taskgraph-definitions` task is busted but has been passing due a bug in Taskgraph. Version 5.2.0 fixes the bug so the task should start properly failing. This PR will address the failures.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/ci/test/kind.yml
+++ b/taskcluster/ci/test/kind.yml
@@ -27,5 +27,7 @@ tasks:
             cwd: '{checkout}'
             command: >-
                 pip3 install -r taskcluster/requirements.txt &&
-                taskgraph full --p taskcluster/test/params &&
+                for param in `ls taskcluster/test/params`; do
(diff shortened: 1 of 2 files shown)


hneiva (approved):
Upgrades taskgraph from 5.1.1 to 5.2.1
LGTM

ahal:
I tracked down the failure to the fact that the generations are happening in parallel using an interactive task. I was unable to reproduce the issue when running the generations one at a time in a for loop. I'm also unable to reproduce the issue on my dev machine even when running in parallel.

I think for now I'm just going to adjust the task to run the generations one at a time, then file an issue in upstream Taskgraph for the parallelization bug.

ahal:
Looks like that did the trick, I'll re-request review for the task changes.

gabrielBusta (approved):

PR #7044, opened 2023-05-31
Bug 1774273 - Part (1/2) - Add signingscript macos notarization task

hneiva:
## Description

  MacOS notarization is moving from iscript to signingscript and requires a separate task.
    
  There are no ways of testing notarization outside production. So, this change only adds the notarization task, but *DOES NOT* affect any other tasks.
  The current signing still notarizes, and is treated as source/upstream by the beetmover task.

## Reference

[Bug 1774273](https://bugzilla.mozilla.org/show_bug.cgi?id=1774273)

## Checklist
    
### DIFF
--- a/taskcluster/ci/mac-notarization/kind.yml
+++ b/taskcluster/ci/mac-notarization/kind.yml
@@ -1,0 +1,39 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - mozillavpn_taskgraph.transforms.requires_level:transforms
+    - mozillavpn_taskgraph.transforms.release_index:transforms
+    - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - signing
+
+tasks:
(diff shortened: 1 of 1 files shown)


lesleyjanenorton (approved):

bhearsum (approved):

hneiva:
After some testing locally, realized the mac-notarization job would basically never run if it's "dangling".
Removing `run-on-tasks-for: []` makes it always run (given it's on level 3)

bhearsum:
> After some testing locally, realized the mac-notarization job would basically never run if it's "dangling". Removing `run-on-tasks-for: []` makes it always run (given it's on level 3)

Let's make sure this gets removed once we add the beetmover dependency. (I assume this is how the current signing is configured as well...)

hneiva:
> Let's make sure this gets removed once we add the beetmover dependency. (I assume this is how the current signing is configured as well...)

Actually signing happens on push, and beetmover happens on relpro/action
example: https://firefox-ci-tc.services.mozilla.com/tasks/HkjqMBWVQh-19iOJ3POBpw

hneiva:
@lesleyjanenorton I don't know why the unit tests are failing. I think it's unrelated. Are you able to help?

ahal:
Looks like there's something off with the commit history? The files changed only reflects one of the commits somehow

PR #7055, opened 2023-06-01
Linux: Remove the use of polkit

oskirby:
## Description
It seems that we no longer require the use of `polkit` in the VPN client, the daemon acquires sufficient privileges when launched via systemd, and is granted `CAP_NET_ADMIN` as well during installation. As such, we can simply remove all the code related to polkit from the codebase.

## Checklist
    

lesleyjanenorton (approved):
🫳
.
..
...
....
...
..
.
.
.


🎤

lesleyjanenorton:
Found whilst sanity-grepping: [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/653299c12a43671c5b8cb31ec159120490c8ac94/docs/dev-setup/linux.md?plain=1#L24) and[ here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/653299c12a43671c5b8cb31ec159120490c8ac94/.github/workflows/test_lottie.yaml#L27) and a few references in `.dictionary` 🧹

oskirby:
@bakulf I'm guessing this has never worked, because the `<allow_active>auth_admin</allow_active>` policy should have prompted the user with a password dialog when activating the VPN, and I don't think I have ever seen that behaviour out of the VPN client.

oskirby:
Okay, I think I understand why the polkit code does nothing. The call `polkit_unix_process_new_for_owner(getpid(), 0, -1)` means that we are checking the permissions of the daemon (not the entity that invoked the D-Bus method). Because the daemon starts as root, this means that the `allow_admin` permission will always be met.

For this to have had any effect, instead of `getpid()` we should have been using the PID of the D-Bus method's caller.

dreirund:
*[Side-Topic] for anyone interested: Some more context to this: https://www.openwall.com/lists/oss-security/2023/08/03/1.*

PR #7079, opened 2023-06-05
Mark OSX builds as universal in Treeherder.

strseb:
## Description

Looking at the task output of the Mac Builder - Since 2.16 this is a universal build - the taskcluster row should reflect that :) 

```
 ~/D/M/M/C/MacOS  file 'Mozilla VPN'                                                                    9  18:40 
Mozilla VPN: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
Mozilla VPN (for architecture x86_64):	Mach-O 64-bit executable x86_64
Mozilla VPN (for architecture arm64):	Mach-O 64-bit executable arm64
``?

### DIFF
--- a/taskcluster/ci/build/macos.yml
+++ b/taskcluster/ci/build/macos.yml
@@ -8,7 +8,7 @@ task-defaults:
             symbol: BD
             kind: build
             tier: 1
-            platform: macos/x86_64
+            platform: macos/universal
     requires-level: 1
     worker-type: b-macos
     fetches:


oskirby (approved):
I never had a good sense of what this field is used for, but I agree that we might as well keep it current.

Looks good to me.

gabrielBusta (approved):

strseb:
@oskirby that field is setting in which row treeherder will display the task https://treeherder.mozilla.org/jobs?repo=mozilla-vpn-client

PR #7096, opened 2023-06-06
VPN-4443: VPN session pings

MattLichtenstein:
## Description

- Add triggers for vpn session pings

## Reference

[VPN-4443: Session ping: Create session ping submissions triggers](https://mozilla-hub.atlassian.net/browse/VPN-4443)


MattLichtenstein:
Using the same data review as when the modified ping was introduced [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6130#issuecomment-1442357270)

src/apps/vpn/controller.cpp
```diff
@@ -55,6 +56,9 @@ constexpr const int CONNECTION_MAX_RETRY = 9;
 constexpr const uint32_t CONFIRMING_TIMOUT_SEC = 10;
 constexpr const uint32_t HANDSHAKE_TIMEOUT_SEC = 15;
 
+constexpr const uint32_t VPNSESSION_PING_TIMER_SEC = 10800;
+constexpr const uint32_t VPNSESSION_PING_TIMER_DEBUG_SEC = 10;
```
brizental, line 60: This is kinda excessive even for debugging IMO. To be honest, I am not really a big fan of having a different timer interval for debugging, unless you are explicitly debugging this ping it's just more noise in the logs. If you are explicitly debugging this ping you can just change the timer yourself for testing purposes. **tl;dr; What if we just remove this debugging interval?**

src/apps/vpn/telemetry/pings.yaml
```diff
@@ -16,6 +16,17 @@ vpnsession:
     session-based metrics, etc.
   include_client_id: true
   send_if_empty: false
+  reasons: 
+    flush: |
+      "Safety measure for when the vpn session does not end in an expected way"
```
brizental, line 21: Please add here exactly what triggers sending a ping with this reason.

src/apps/vpn/controller.cpp
MattLichtenstein: It was specifically written in the tech spec and I could see this potentially being useful for QA testing. @mcleinman thoughts?

Also, the debug timer is actually supposed to be 120 seconds - I had it set to 10 for... ultra debugging purposes (but this kind of proves your point 😂)

brizental: LOL "ultra debugging". Ok, if it's in the spec just leave it in. We can always file a ticket to remove it later if it becomes annyoing.

MattLichtenstein:
@mcleinman I have not included the "timeout" reason as I am not aware of any situations where the VPN could turn itself off due to a connection issue. Let me know what you think

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -126,6 +126,19 @@ MZViewBase {
...
+
+            Layout.rightMargin: MZTheme.theme.windowMargin
+            labelText: "VPNSession ping timeout debug mode"
+            subLabelText: "Shortens the VPNSession ping timeout from 3 hours to 2 minutes"
+            isChecked: MZSettings.vpnSessionPingTimeoutDebug
+            showDivider: false
+            onClicked: {
+                MZSettings.vpnSessionPingTimeoutDebug = !MZSettings.vpnSessionPingTimeoutDebug
+            }
+        }
```
brizental, line 140: Fancy, I like it.

src/apps/vpn/telemetry/pings.yaml
```diff
@@ -16,6 +16,15 @@ vpnsession:
     session-based metrics, etc.
   include_client_id: true
   send_if_empty: false
+  reasons: 
+    flush: |
+      Safety measure for when the vpn session does not end in an expected way, sent right before the start ping
+    start: |
+      When the vpn session starts
```
brizental, line 23: Can you be more specific here about what it means for a session to start i.e. "A session starts when a user turns on the VPN". This is documentation for data analyzers, so it's good to be very specific so they don't need to ask / look at the code.

src/apps/vpn/telemetry/pings.yaml
```diff
@@ -16,6 +16,15 @@ vpnsession:
...
+    end: |
+      When the vpn session ends
```
brizental, line 25: Same comment on documentation. I'd also add any caveats, such as "A VPN session never ends on a timeout" or "A VPN session doesn't end unless a user explicit turns the VPN off or force kills the VPN process". (Not sure these example are accurate, they are just examples).

src/apps/vpn/controller.cpp
```diff
@@ -582,10 +602,21 @@ void Controller::handshakeTimeout() {
   serverUnavailable();
 }
 
+void Controller::vpnSessionPingTimeout() {
```
brizental, line 605: Not sure this code should live in the controller, but instead in Telemetry. I'm really not sure though, so please argue with me :D

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
brizental: This requires a reset to be applied right?

brizental:
> @mcleinman I have not included the "timeout" reason as I am not aware of any situations where the VPN could turn itself off due to a connection issue. Let me know what you think

@oskirby Is this by design or should we file a ticket to timeout with the connections is lost?

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
MattLichtenstein: A reset of the VPN / controller, yes - not the whole app. Do you think it would be better to just make a note in the dev menu or make the adjustment while the controller is on?

src/apps/vpn/controller.cpp
MattLichtenstein: I can see that. We can really move all of this logic to the `telemetry` class and start/end timers or submit pings on `controller::stateChanged()`

src/apps/vpn/controller.cpp
```diff
@@ -55,6 +56,9 @@ constexpr const int CONNECTION_MAX_RETRY = 9;
 constexpr const uint32_t CONFIRMING_TIMOUT_SEC = 10;
 constexpr const uint32_t HANDSHAKE_TIMEOUT_SEC = 15;
 
+constexpr const uint32_t VPNSESSION_PING_TIMER_SEC = 10800; //3 hours
```
mcleinman, line 59: Nit - do this is math equation
```suggestion
constexpr const uint32_t VPNSESSION_PING_TIMER_SEC = 3 * 60 * 60; //3 hours
```

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -126,6 +126,19 @@ MZViewBase {
             }
         }
 
+        MZCheckBoxRow {
+            id: checkBoxRowVPNSessionPingTimeout
+
+            Layout.rightMargin: MZTheme.theme.windowMargin
+            labelText: "VPNSession ping timeout debug mode"
+            subLabelText: "Shortens the VPNSession ping timeout from 3 hours to 2 minutes"
```
mcleinman, line 134:
```suggestion
            subLabelText: "Shortens the VPNSession timer ping cadence from 3 hours to 2 minutes"
```

mcleinman (approved):
This looks great. Struggling with my build at the moment so haven't been able to test it yet, but if someone else confirms it works as expected then I'm a happy camper. Thanks for this!

brizental (approved):
Since the controller is fully mocked in unit tests, I believe the new architecture makes it possible to unit test these changes. The unit tests may require `test_before_next_submit` to be implemented though... Hum. Would you mind filing a ticket to add tests to these ping submission triggers and block that ticket on https://mozilla-hub.atlassian.net/browse/VPN-3338?

Off the top of my head, we need to add tests for 1. the session id is rotated when expected, 2. the expected context metrics are added to the start ping, 3. the pings with the expected reasons are sent at the expected triggers.

---

Tentatively approving this, but the positioning on the `session_id.generateAndSet` on a session end needs to be changed before merging, because that is a bug on the implementation.

brizental (approved):

PR #7110, opened 2023-06-07
Linux: Re-introduce D-Bus API Authorization

oskirby:
## Description
We recently found that the use of polkit in our code was incorrect in that it wasn't actually enforcing anything so we decided to simply remove the code around polkit. This PR attempts to restore proper authorization of the D-Bus API used to manipulate the controller and network interface. 

The authorization check goes as follows:
 - If the calling process has the `CAP_NET_ADMIN` capability, then access to the D-Bus API is granted.
 - If calling the `activate` method and the VPN was previously disconnected, access is granted and the UID that activated the connection is recorded for later.
 - If the caller matches the UID that activated the connection, then access is granted.
 - Otherwise, access is denied.

In other words: Any user can activate the VPN, and once activated only that user can manipulate the VPN connection. However, processes with `CAP_NET_ADMIN` bypass this check and are always authorized.

## Checklist


src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -263,3 +284,58 @@ bool DBusService::firewallClear() {
...
+  QDBusReply<uint> reply = iface->servicePid(message().service());
+  uint senderpid = reply.value();
+  if (!reply.isValid() || (senderpid == 0)) {
+    // Could not lookup the sender's PID. Rejected!
+    logger.warning() << "Failed to resolve sender PID";
+    return false;
+  }
+
+  // Allow the call if it was made by the same binary.
+  if (!m_daemonExePath.isEmpty() &&
```
bakulf, line 306: This can be an issue for local development. Often i use the daemon from the deb package with a compiled client app. 

Would that work?

oskirby: Unfortunately, this usecase does not work without a step that requires root priveleges. Either the client must be installed first (eg: `sudo cmake --install <build-dir>`) or we must grant it the `CAP_NET_ADMIN` permission (eg: `sudo setcap cap_net_admin+ep <build-dir>/src/mozillavpn`)

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+/* Drop root permissions from the daemon. */
+void DBusService::dropRootPermissions() {
+  logger.debug() << "Dropping root permissions";
+
+  cap_t caps = cap_get_proc();
+  if (caps == nullptr) {
+    logger.warning() << "Failed to retrieve process capabilities";
+    return;
+  }
+  auto guard = qScopeGuard([&] { cap_free(caps); });
```
Gela, line 314: Is this being used somewhere?

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+  // Acquire CAP_SETUID, we need it to masquerade as other users on their
+  // session busses for application tracking.
+  //
+  // NOTE: ptrace is a dangerous permission to hold. If it may be safer to
+  // relent on the executable check and grant CAP_NET_ADMIN to the client
+  // process during installation.
+  //
+  // Clear all other capabilities, effectively discarding our root permissions.
+  cap_value_t newcaps[] = {CAP_NET_ADMIN, CAP_SETUID};
+  int numcaps = sizeof(newcaps) / sizeof(cap_value_t);
```
Gela, line 331:
```suggestion
  const int numcaps = sizeof(newcaps) / sizeof(cap_value_t);
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+  }
+}
+
+/* Checks to see if the caller has sufficient authorization */
+bool DBusService::checkCallerAuthz() {
+  if (!calledFromDBus()) {
+    // If this is not a D-Bus call, it came from the daemon itself.
+    return true;
+  }
+  QDBusConnectionInterface* iface = QDBusConnection::systemBus().interface();
```
Gela, line 349:
```suggestion
  const QDBusConnectionInterface* iface = QDBusConnection::systemBus().interface();
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+
+  // If the VPN is active, and we know the UID that turned it on, as a special
+  // case we permit that user full access to the D-Bus API in order to manage
+  // the connection.
+  if (m_sessionUid != 0) {
+    QDBusReply<uint> reply = iface->serviceUid(message().service());
```
Gela, line 355:
```suggestion
    const QDBusReply<uint> reply = iface->serviceUid(message().service());
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+    if (reply.isValid() && m_sessionUid == reply.value()) {
+      return true;
+    }
+  }
+  // Otherwise, if this is the activate method, we permit any non-root user to
+  // activate the VPN, but we will remember their UID for later authentication.
+  else if ((message().type() == QDBusMessage::MethodCallMessage) &&
+           (message().member() == "activate")) {
+    QDBusReply<uint> reply = iface->serviceUid(message().service());
```
Gela, line 364:
```suggestion
    const QDBusReply<uint> reply = iface->serviceUid(message().service());
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+      m_sessionUid = senderuid;
+      return true;
+    }
+  }
+  // In all other cases, the use of this D-Bus API requires the CAP_NET_ADMIN
+  // permission, which we can check by examining the PID of the sender. Note
+  // that a zero UID (root) is used as a guard value to fall back to this case.
+
+  // Get the PID of the D-Bus message sender.
+  QDBusReply<uint> reply = iface->servicePid(message().service());
```
Gela, line 376:
```suggestion
  const QDBusReply<uint> reply = iface->servicePid(message().service());
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+  uint senderpid = reply.value();
```
Gela, line 377:
```suggestion
  const uint senderpid = reply.value();
```

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+    return false;
+  }
+
+  // Get the capabilties of the sender process.
+  cap_t caps = cap_get_pid(senderpid);
+  if (caps == nullptr) {
+    logger.warning() << "Failed to retrieve process capabilities";
+    return false;
+  }
+  auto guard = qScopeGuard([&] { cap_free(caps); });
```
Gela, line 390: If the lambda expression isn't being called further down, can we remove the `guard` assignment?

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -258,8 +291,126 @@ bool DBusService::firewallPid(int rootpid, const QString& state) {
...
+}
+
+uint DBusService::getCallerUid() {
+  if (!calledFromDBus()) {
+    // If this is not a D-Bus call, it came from the daemon itself.
+    return getuid();
+  }
+
+  // Get the UID of the D-Bus message sender.
+  QDBusConnectionInterface* iface = QDBusConnection::systemBus().interface();
```
Gela, line 408: Can these be marked as `const`? 
```suggestion
  const QDBusConnectionInterface* iface = QDBusConnection::systemBus().interface();
```

Gela (changes requested):

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
oskirby: This is just used for garbage collection. The scopeguard ensures that `cap_free()` is invoked regardless of how this function exits.

src/apps/vpn/platforms/linux/daemon/dbusservice.cpp
oskirby: The scope guard is used to ensure that `cap_free()` is called at function exit, regardless of which path we take to exit the function.

strseb (approved):
lgtm

Gela (approved):
Thank you!

PR #7112, opened 2023-06-07
[RFC] VPN-4736 - Sharing a session identifier between app and daemon

brizental:
One of the main concerns that prompted this RFC, was the concern on how to keep the identifiers in sync between main application and daemon and how to know if they are not in sync. 

I believe the first concern is addressed by the design itself and the second concern doesn't seem like a plausible scenario given the design. I am adding this note as a call for the reviewers  close attention to the synchronization aspect of the proposed solution. Thanks!

### DIFF
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -218,6 +218,5 @@ for cases such as iOS where the daemon process is only alive while a session is 
 ### Share the management burdens of the shared session id
 
 The first iteration of the proposed solution, included the main application storing and managing
-the `shared_session_id` and passing it along to the daemon when possible. That can be read on
-[7e20c2c933fc868b3ea50d6eb35651b904628658](https://github.com/mozilla-mobile/mozilla-vpn-client/tree/7e20c2c933fc868b3ea50d6eb35651b904628658).
+the `shared_session_id` and passing it along to the daemon when possible.
 Ultimately, that is an overly complex way to achieve the same end goal as the current proposal.
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -1,0 +1,226 @@
+- Status:
+- Date: 2023-06-07
+- Author: [@brizental](https://github.com/brizental)
+- RFC PR: [#TODO](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/TODO)
+- Implementation GitHub issue: TODO
+
+## Problem Statement
+
+The VPN application is divided in two processes: the main application and the daemon. The main application is where the user interface runs and the daemon is what manages the VPN configuration and tunnel.
+
+In desktop environments, even while backgrounded the main application process is still running, but in mobile environments it is not. Due to this caveat, the mobile daemon processes have a separate instance of Glean to be able to collect telemetry about the connection while the application is in the background.
+
+This architecture poses a problem for analyzers. It is not possible to easily join the data from the same session between daemon and main application. The session ids are different due to these being two different Glean instances.
+
+For more information on this topic, refer to [Mozilla VPN Telemetry Refactor](https://docs.google.com/document/d/1jyNZ_g_cUpZZsEr2hYwnwZgkxlvqTmoFUJKp_LAOPts/edit#heading=h.xararkvsnpss).
+
+## Proposed Solution
+
+> **Note** Honestly, I think for this proposed solution it's just easier to look at the pseudo-code in Appendix 1.
+
+A new metric will be added to both the daemonsession and the session pings: shared_session_id. This metric will have lifetime ping i.e. it is cleared from storage whenever a ping containing it is submitted. Thus it needs to be set again every time a new session/daemonsession ping is submitted.
+
+Both daemon and main application will store the value of this metric in their local storages for synchronization purposes. More about this in the sections below.
+
+A new reason will be added to the session pings “flush_init”. As better described in the sections below, the processes will be able to identify that either the VPN tunnel or main application did not gracefully hit the deactivation signal i.e. were force killed. The “flush_init” ping will be used to flush any data remaining from the previous session.
+
+There will be four key triggers that will handle the shared_session_id: initialization, on VPN activation, ping timer and on VPN deactivation.
+
+### ON VPN ACTIVATION
+
+#### MAIN APPLICATION
+
+In case the VPN is started from the main application, a session id will be created and shared through the activation signal with the daemon. The generated UUID will be stored in the main application local storage and the session ping will be submitted.
+
+#### DAEMON
+
+If the VPN is being activated through the main application, a session id will have been passed through the activation signal. Otherwise the Daemon will generate this value itself. The UUID will be stored in the daemon local storage and the session ping will be submitted.
+
+### ON PING TIMER
+
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage and recorded. The session ping is submitted.
+
+#### DAEMON
+
+The session id is retrieved from local storage and recorded. The session ping is submitted.
+
+### ON VPN DEACTIVATION
+
+The deactivation signal may not be hit in both the main application and the daemon, because both processes may be force killed. The local storage that contains the session id will be cleared after submitting the final session ping. This will be a flag at initialization that the final session ping was not submitted and there may be lingering data from a previous session in storage that needs to be submitted.
+
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage and recorded. The session ping is submitted. Finally local storage for the session id is cleared.
+
+#### DAEMON
+
+The session id is retrieved from local storage and recorded. The session ping is submitted. Finally local storage for the session id is cleared.
+
+### ON INITIALIZATION
+
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage. If it is not empty, there are four possibilities:
+
+1. The same session from a previous app run is still ongoing. No action is required
+2. Another session is ongoing in which case a flush ping is required to data from a previous session and start recording data for the new session
+3. No session is ongoing but there is a lingering session id indicating that there is still lingering data from a previous session in storage and a flush ping is required
+4. No session is ongoing and no lingering session id is in storage. No action is required
+
+#### DAEMON
+
+On the daemon there are only two possibilities:
+
+1. No session is ongoing but there is a lingering session id indicating that there is still lingering data from a previous session in storage and a flush ping is required
+2. No session is ongoing and no lingering session id is in storage. No action is required
+
+## APPENDIX
+
+### APPENDIX 1: PSEUDO CODE
+
+#### Main application
+
+```cpp
+MainApp::init() {
+  String lingeringSessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  // If this is not empty, the VPN was not deactivated from the app.
+  if (!lingeringSessionId.isEmpty()) {
+    if (Daemon::vpnIsOn()) {
+      try {
+        String sessionId = Daemon::getSessionId();
+        // Check if we are still in the same session
+        if (sessionId != lingeringSessionId) {
+          // Flush any lingering data from a previous session
+          pings::main_app::session.submit(INIT_FLUSH_REASON);
+
+          metrics::main_app::sessionId.set(sessionId);
+          // Record session::* context metrics e.g. app_exclusions.
+
+          LocalStorage::set(SESSION_ID_STORAGE_KEY, sessionId);
+        }
+
+        // Schedule the timer.
+        MainApp::schedulePingTimer();
+      } catch {
+        // Note: If communication with the daemon doesn't work,
+        // let's assume the VPN is off. This is very unlikely, given we
+        // have just communicated with the daemon through Daemon::vpnIsOn();
+
+        // Flush any lingering data from a previous session
+        pings::main_app::session.submit(INIT_FLUSH_REASON);
+        // Clear the sessionId storage
+        LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+      }
+    } else {
+      // Flush any lingering data from a previous session
+      pings::main_app::session.submit(INIT_FLUSH_REASON);
+      // Clear the sessionId storage
+      LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+    }
+  }
+}
+
+MainApp::onActivate() {
+  // Let's flush any lingering data for before activating the VPN.
+  // We want the active session pings to only include data from the current session.
+  //
+  // This ping should be always empty. This is a way to monitor if it isn't
+  pings::main_app::session.submit(FLUSH_REASON);
+
+  String sessionId = generateUUID();
+  try {
+    Daemon::activate(config, sessionId);
+    // Note: session::* metrics should only be recorded from now on,
+    // so that flush pings don't contain unexpected data.
+
+    metrics::main_app::sessionId.set(sessionId);
+    // Record session::* context metrics e.g. app_exclusions.
+
+    LocalStorage::set(SESSION_ID_STORAGE_KEY, sessionId);
+
+    // Submit the start ping
+    pings::main_app::session.submit(START_REASON);
+  } catch {
+    // TODO: Record error.
+    LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+  }
+}
+
+MainApp::onPingTimer() {
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::main_app::sessionId.set(sessionId);
+
+  // Submit the timer ping
+  pings::main_app::session.submit(TIMER_REASON);
+}
+
+MainApp::onDeactivate() {
+  // Stop the timer.
+  MainApp::stopPingTimer();
+
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::main_app::sessionId.set(sessionId);
+
+  // Submit the stop ping
+  pings::main_app::session.submit(TIMER_REASON);
+  LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+}
+```
+
+#### Daemon
+
+```cpp
+Daemon::init() {
+  String lingeringSessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+
+  // If this is not empty, the tunnel was force killed
+  if (!lingeringSessionId.isEmpty()) {
+    metrics::daemon::sessionId.set(lingeringSessionId);
+    // Flush lingering data
+    pings::daemonsession.submit(INIT_FLUSH_REASON);
+    LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+  }
+}
+
+Daemon::onActivate(Config config, String sessionId = "") {
+  // Let's flush any lingering data for before activating the VPN.
+  // We want the active session pings to only include data from the current session.
+  //
+  // This ping should be always empty. This is a way to monitor if it isn't
+  pings::daemonsession.submit(FLUSH_REASON);
+
+  auto sessionId = sessionId;
+  if (sessionId.isEmpty()) {
+    logger.info() << "Starting VPN from somewhere other than the main app";
+    sessionId = metrics::daemon::sessionId.generateAndSet();
+  } else {
+    metrics::daemon::sessionId.set(sessionId);
+  }
+  LocalStorage::set(SESSION_ID_STORAGE_KEY, sessionId);
+
+  // Submit the start ping
+  pings::daemonsession.submit(START_REASON);
+}
+
+Daemon::onPingTimer() {
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::daemon::sessionId.set(sessionId);
+
+  // Submit the timer ping
+  pings::daemonsession.submit(TIMER_REASON);
+}
+
+Daemon::onDeactivate() {
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::daemon::sessionId.set(sessionId);
+
+  // Submit the stop ping
+  pings::daemonsession.submit(STOP_REASON);
+
+  // Clear the local storage
+  LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+}
+
+```
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -2,7 +2,7 @@
 - Date: 2023-06-07
 - Author: [@brizental](https://github.com/brizental)
 - RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
-- Implementation GitHub issue: TODO
+- Implementation GitHub issue: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7588
 
 ## Problem Statement
 
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -146,14 +146,14 @@ If the main application is active during a session start it will,
 
 1. Signal to the daemon that a new session is to be started.
 2. It will then ask the daemon for the value of the shared session id it created.
-3. If the previous two communications where successful
+3. If the previous two communications were successful
    1. The `vpnsession` ping will be submitted with reason `flush`.
    2. The value of the retrieved `shared_session_id` is recorded.
    3. Other session metrics may be recorded from this point.
    4. If the communication with the daemon fails, an error metric is recorded.
 
 > **Note**: If the retrieved session id is the known inactive session id,
-> this is considered an unsuccessful retrieval and the error path is taken.
+> this is considered an unsuccessful retrieval and the error path -- i.e. step 4 of the above steps -- is taken.
 
 #### On session end
 
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -147,13 +147,13 @@ If the main application is active during a session start it will,
 1. Signal to the daemon that a new session is to be started.
 2. It will then ask the daemon for the value of the shared session id it created.
 3. If the previous two communications where successful
-   3.a. The `vpnsession` ping will be submitted with reason `flush`.
-   3.b. The value of the retrieved `shared_session_id` is recorded.
-   3.c. Other session metrics may be recorded from this point.
+   1. The `vpnsession` ping will be submitted with reason `flush`.
+   2. The value of the retrieved `shared_session_id` is recorded.
+   3. Other session metrics may be recorded from this point.
+   4. If the communication with the daemon fails, an error metric is recorded.
 
-> **Note**: If the communication with the daemon fails, an error metric is recorded. If the
-> retrieved session id is the known inactive session id, this is considered an unsuccessful
-> retrieval and the error path is taken.
+> **Note**: If the retrieved session id is the known inactive session id,
+> this is considered an unsuccessful retrieval and the error path is taken.
 
 #### On session end
 
--- a/docs/rfcs/0002-session-id-sharing.md
+++ b/docs/rfcs/0002-session-id-sharing.md
@@ -1,226 +1,223 @@
 - Status:
 - Date: 2023-06-07
 - Author: [@brizental](https://github.com/brizental)
-- RFC PR: [#TODO](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/TODO)
+- RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
 - Implementation GitHub issue: TODO
 
 ## Problem Statement
 
-The VPN application is divided in two processes: the main application and the daemon. The main application is where the user interface runs and the daemon is what manages the VPN configuration and tunnel.
+### The two processes: Main Application and Daemon
 
-In desktop environments, even while backgrounded the main application process is still running, but in mobile environments it is not. Due to this caveat, the mobile daemon processes have a separate instance of Glean to be able to collect telemetry about the connection while the application is in the background.
+The Mozilla VPN application is split in two processes: the main application and the daemon
+(a background process). The main application is where the user interface runs and the
+daemon is what manages the VPN configuration and tunnel.
 
-This architecture poses a problem for analyzers. It is not possible to easily join the data from the same session between daemon and main application. The session ids are different due to these being two different Glean instances.
+> **Note**: The daemon process may also be referred to as "network extension" on iOS
+> or "broker" in multiple platforms. Throughout this document the term "daemon" will be used.
 
-For more information on this topic, refer to [Mozilla VPN Telemetry Refactor](https://docs.google.com/document/d/1jyNZ_g_cUpZZsEr2hYwnwZgkxlvqTmoFUJKp_LAOPts/edit#heading=h.xararkvsnpss).
+#### The lifetime of each process varies between the supported platforms
+
+On Android the daemon process starts alongside the main application process once the application is
+launched and remains active until killed, regardless on whether the VPN is turned on or not.
+The main application process is only active while the application is in the foreground.
+iOS is similar, but the daemon process is only active while the VPN is turned ON.
+
+On Desktop environments the main application process is active while the application is active,
+independent of the application being foregrounded or backgrounded. Due to OS specific mechanisms
+such as [App Nap](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/AppNap.html)
+this process may go dormant intermittently after the application is backgrounded for a long time.
+
+#### The daemon process may be started independently from the main application on mobile platforms
+
+Both the operational system and the Mozilla VPN application expose features that allow the user
+to turn on the VPN without interacting with the main application. On mobile environments, when
+the VPN is turned on through any of these methods, the main application process is never launched
+while the daemon process is.
+
+### The issue with mobile session based telemetry
+
+When analyzing incoming telemetry from the Mozilla VPN application, it's a recurring requirement
+to be able to partition the incoming data by session. **A VPN session starts when the VPN is turned
+on through some user action and ends once it is turned off through some user action.**
+
+That is easily addressed on desktop environments, by having the main application manage a session_id
+metric and add that to all session pings. Data from a given session all have the same session_id
+and can be joined using that identifier.
+
+On mobile environments, due to the lack of guarantees on the main application process being alive
+throughout most or any of the VPN session, having the main application manage the session_id is unreliable.
+This identifier needs to be rotated on every session start/end and the main application process may
+not be running at these times.
+
+Moreover, on mobile platforms, the fact that the main application process is killed whenever
+the application goes to the background makes the main application telemetry insufficient for collecting
+all the telemetry required to understand the health of the Mozilla VPN application and its usage patterns.
+To address this issue, mobile daemons have a separate instance of Glean, used to collect telemetry
+while the application is backgrounded. In order to get a full picture of a mobile user's session,
+joining the daemon data with the main application is necessary. However, the telemetry clients in
+each process are independent so manually sharing an identifier is required for such joins.
+
+On [VPN session pings - Tech Spec](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/184419319/VPN+session+pings+-+Tech+spec)
+an installation_id metric is introduced that would allows joins between daemon and main application
+data. That still does not make it possible to join specific sessions i.e. we are able to infer that
+a given daemon session is coming from the same installation as a main application session, but we
+cannot reliably or easily infer that a given daemon session is the same as a given main application session.
+
+## Document Purpose
+
+Propose a solution to reliably join mobile session data between daemon and main application telemetry.
+
+## Requirements
+
+The main transport for session related data are the session pings, namely
+the [`vpnsession`](https://dictionary.telemetry.mozilla.org/apps/mozilla_vpn/pings/vpnsession) ping
+and the [`daemonsession`](https://dictionary.telemetry.mozilla.org/apps/mozilla_vpn/pings/daemonsession) ping.
+
+It is a requirement on the proposed solution, that the metrics included in each ping are all
+recorded during the session related to the shared session id in it i.e. the ping does not contain
+any metric from a previous session.
 
 ## Proposed Solution
 
-> **Note** Honestly, I think for this proposed solution it's just easier to look at the pseudo-code in Appendix 1.
+A new metric will be introduced to mobile platforms: `shared_session_id`. This metric will be the
+primary session identifier metric for Mozilla VPN mobile telemetry. This metric will have lifetime
+`User`, with this lifetime, once a metric is set it can only be overwritten but never cleared from
+the telemetry storage.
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
+- Status:
+- Date: 2023-06-07
+- Author: [@brizental](https://github.com/brizental)
+- RFC PR: [#TODO](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/TODO)
```
brizental, line 4:
```suggestion
- RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+### ON VPN ACTIVATION
+
+#### MAIN APPLICATION
+
+In case the VPN is started from the main application, a session id will be created and shared through the activation signal with the daemon. The generated UUID will be stored in the main application local storage and the session ping will be submitted.
+
+#### DAEMON
+
+If the VPN is being activated through the main application, a session id will have been passed through the activation signal. Otherwise the Daemon will generate this value itself. The UUID will be stored in the daemon local storage and the session ping will be submitted.
```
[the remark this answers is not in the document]
csadilek: Ah, I think it's answered below in the code snippet: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112/files#diff-0446585bf7fa18d37550784eb229ef4d13683272eda91aaa8d2c9134931c4c9cR92-R101

If the daemon has a different session ID, we will update the one in the main app when initialized later? If so, let's add that here too :).

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+- Author: [@brizental](https://github.com/brizental)
+- RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
+- Implementation GitHub issue: TODO
+
+## Problem Statement
+
+The VPN application is divided in two processes: the main application and the daemon. The main application is where the user interface runs and the daemon is what manages the VPN configuration and tunnel.
+
+In desktop environments, even while backgrounded the main application process is still running, but in mobile environments it is not. Due to this caveat, the mobile daemon processes have a separate instance of Glean to be able to collect telemetry about the connection while the application is in the background.
+
```
csadilek, line 12: Different topic: Sending telemetry "while the app is in the background" is, generally speaking, not ideal. Our use case here is different with the app rarely being opened/foregrounded once VPN is running, but we should work through how exactly (and as conservatively as possible) we're doing this. Do we perhaps already have a write-up of this aspect?

csadilek (commented):
❤️ 🎉 RFC#2

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+The session id is retrieved from local storage and recorded. The session ping is submitted.
+
+#### DAEMON
+
+The session id is retrieved from local storage and recorded. The session ping is submitted.
+
+### ON VPN DEACTIVATION
+
+The deactivation signal may not be hit in both the main application and the daemon, because both processes may be force killed. The local storage that contains the session id will be cleared after submitting the final session ping. This will be a flag at initialization that the final session ping was not submitted and there may be lingering data from a previous session in storage that needs to be submitted.
```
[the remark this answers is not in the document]
brizental: Correct. I'll try to clarify this.

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
brizental: Daemon -> Main App communication is only possible on Android and not in iOS. To keep implementation consistent between both environments, I considered it a limitation that the daemon can't send any data to the app. With that in mind, the app will ask for a session id when initialized in case the VPN is active at initialization time. That is described on the "On initialization" section. Do you think that is enough? Or should I still mention that here?

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
brizental:
```suggestion
The deactivation signal may not be hit in both the main application and the daemon, because both processes may be force killed. The local storage that contains the session id will be cleared after submitting the final session ping i.e. the ping sent upon VPN deactivation. This will be a flag at initialization that the final session ping was not submitted and there may be lingering data from a previous session in storage that needs to be submitted.
```

docs/rfcs/0002-session-id-sharing.md
brizental: There are three write ups related to this.
1. https://docs.google.com/document/d/1A2O_eACk0P3jDy0g2rRN5UTBOUj-iN-re6tVsRpxtDA/edit?usp=sharing
2. https://docs.google.com/document/d/1jyNZ_g_cUpZZsEr2hYwnwZgkxlvqTmoFUJKp_LAOPts/edit?usp=sharing
3. https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/184419319/Metric+pings+-+Tech+spec

Would you mind expanding on "Sending telemetry "while the app is in the background" is, generally speaking, not ideal (on mobile)". I know about the challenges of spawning background tasks for the sake of submitting telemetry. However in this case, I believe that doesn't apply since we will be sending telemetry from an already running process.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+- RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
+- Implementation GitHub issue: TODO
+
+## Problem Statement
+
+The VPN application is divided in two processes: the main application and the daemon. The main application is where the user interface runs and the daemon is what manages the VPN configuration and tunnel.
+
+In desktop environments, even while backgrounded the main application process is still running, but in mobile environments it is not. Due to this caveat, the mobile daemon processes have a separate instance of Glean to be able to collect telemetry about the connection while the application is in the background.
+
+This architecture poses a problem for analyzers. It is not possible to easily join the data from the same session between daemon and main application. The session ids are different due to these being two different Glean instances.
```
brizental, line 13: Note to self: expand a bit more here. The problem is unclear.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+The session id is retrieved from local storage and recorded. The session ping is submitted. Finally local storage for the session id is cleared.
+
+### ON INITIALIZATION
+
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage. If it is not empty, there are four possibilities:
+
+1. The same session from a previous app run is still ongoing. No action is required
+2. Another session is ongoing in which case a flush ping is required to data from a previous session and start recording data for the new session
```
brizental, line 68:
```suggestion
2. Another session is ongoing in which case a flush ping is required to submit data from a previous session and start recording data for the new session
```

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
csadilek: Thanks, yeah, I would mention it here, because it's early in the doc and when reading "Otherwise the Daemon will generate this value itself." I was immediately wondering how the main app would get to that new session ID.

docs/rfcs/0002-session-id-sharing.md
brizental: Got it, honestly performance is something I definitely overlooked when proposing and implementing the integration of Glean with the background processes. I have filed https://mozilla-hub.atlassian.net/browse/VPN-4965 so that we can deep dive on this.

brizental:
> I'm holding off on reviewing this for now. My understanding is that this ID sharing may or may not happen - it's dependent on how the data we get back from Super Dooper Phase 1 looks. (Essentially: Do we feel like we're missing an actionable amount of information by not connecting with a shared session identifier?) I'm going to wait on reviewing this until we're confident we're moving forward with future phases, as some of the underlying things could change before we get there.

This work is related to the work on the Super Dooper Dashboards epic in the sense that it proposes a slight change in scheduling of the session and daemonsession pings. It has been spun out of the that effort though for the reason you point out: it is unclear whether it is necessary for that or not.

This proposal is pre-work in preparation for [VPN-3874](https://mozilla-hub.atlassian.net/browse/VPN-3874). I will make sure to explictly block the ticket on this spike to make this clear. That ticket is about application health and performance monitoring and for that effort it will be necessary to be able to reliably partition data per VPN session. I will make sure to expand on that on this proposal.

I realize this document also doesn't contain a session on the impact this will have on current colletion i.e. whether it changes anything on the Super Dooper proposal. The short answer to that is: no. Although there is a new trigger proposed to send a session and daemon session ping, that should not impact the collection there. Other than the change in ping scheduling, there is also the proposal to add a new metric to the session and daemonsession pings. No removals, just an addition, thus also no impact.

I will make sure to make this clear in the proposal and address the other review comments. Meanwhile I would like to re-request that you review this proposal. Your opinion as the expert on the session and daemonsession pings is necessary to move forward here. I want to make sure the two lines of work are really not conflicting as stated above. cc @mcleinman

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+#### MAIN APPLICATION
+
+In case the VPN is started from the main application, a session id will be created and shared through the activation signal with the daemon. The generated UUID will be stored in the main application local storage and the session ping will be submitted.
+
+#### DAEMON
+
+If the VPN is being activated through the main application, a session id will have been passed through the activation signal. Otherwise the Daemon will generate this value itself. The UUID will be stored in the daemon local storage and the session ping will be submitted.
+
+### ON PING TIMER
```
Gela, line 39: Could this section be expanded to explain the reason we want to collect info during the ping?

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+### ON INITIALIZATION
+
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage. If it is not empty, there are four possibilities:
+
+1. The same session from a previous app run is still ongoing. No action is required
+2. Another session is ongoing in which case a flush ping is required to submit data from a previous session and start recording data for the new session
+3. No session is ongoing but there is a lingering session id indicating that there is still lingering data from a previous session in storage and a flush ping is required
```
mcleinman, line 69: Can you be explicit about how we determine which of these four situations we're in? (Perhaps having a section on communication between app and daemon would be helpful - currently the only mention in narrative form is the "when started from app, we send ID to daemon".

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::main_app::sessionId.set(sessionId);
+
+  // Submit the stop ping
+  pings::main_app::session.submit(TIMER_REASON);
+  LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+}
+```
+
+#### Daemon
```
mcleinman, line 172: Can you share some psuedocode for iOS? I don't think this will apply over there.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+    }
+  }
+}
+
+MainApp::onActivate() {
+  // Let's flush any lingering data for before activating the VPN.
+  // We want the active session pings to only include data from the current session.
+  //
+  // This ping should be always empty. This is a way to monitor if it isn't
+  pings::main_app::session.submit(FLUSH_REASON);
```
mcleinman, line 130: Is it worth a quick check here to ensure the local storage is clear? I don't currently see paths to allow that to happen, but seems like it could be a defensive double check if we wanted one.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+    pings::main_app::session.submit(START_REASON);
+  } catch {
+    // TODO: Record error.
+    LocalStorage::clear(SESSION_ID_STORAGE_KEY);
+  }
+}
+
+MainApp::onPingTimer() {
+  String sessionId = LocalStorage::get(SESSION_ID_STORAGE_KEY);
+  metrics::main_app::sessionId.set(sessionId);
```
mcleinman, line 153: What is this for? Wouldn't it already be set from the initialization?

mcleinman: I realized it, this is because we have a `ping` lifetime. What do you think about a `user` lifetime for this?

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+  // This ping should be always empty. This is a way to monitor if it isn't
+  pings::main_app::session.submit(FLUSH_REASON);
+
+  String sessionId = generateUUID();
+  try {
+    Daemon::activate(config, sessionId);
+    // Note: session::* metrics should only be recorded from now on,
+    // so that flush pings don't contain unexpected data.
+
+    metrics::main_app::sessionId.set(sessionId);
```
mcleinman, line 138: Nit: Any reason you decided not to set this metric before activating the daemon (line 134)? If we inverted, we could just pull the UUID from the metric creation.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+2. No session is ongoing and no lingering session id is in storage. No action is required
+
+## APPENDIX
+
+### APPENDIX 1: PSEUDO CODE
+
+#### Main application
+
+```cpp
+MainApp::init() {
```
mcleinman, line 86: What if it the app local storage is empty but there is an active VPN session in the daemon? I think we're missing that case.

mcleinman (changes requested):
I think we missed the "app turns on, no session ID in local storage, but there is an active session" case. And a few nits and questions to better understand this.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Proposed Solution
+
+> **Note** Honestly, I think for this proposed solution it's just easier to look at the pseudo-code in Appendix 1.
+
+A new metric will be added to both the daemonsession and the session pings: shared_session_id. This metric will have lifetime ping i.e. it is cleared from storage whenever a ping containing it is submitted. Thus it needs to be set again every time a new session/daemonsession ping is submitted.
+
+Both daemon and main application will store the value of this metric in their local storages for synchronization purposes. More about this in the sections below.
+
+A new reason will be added to the session pings “flush_init”. As better described in the sections below, the processes will be able to identify that either the VPN tunnel or main application did not gracefully hit the deactivation signal i.e. were force killed. The “flush_init” ping will be used to flush any data remaining from the previous session.
```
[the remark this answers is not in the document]
brizental: The main difference is that it's not sent right before a session start, but on initialize when there is a lingering session id.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+This architecture poses a problem for analyzers. It is not possible to easily join the data from the same session between daemon and main application. The session ids are different due to these being two different Glean instances.
+
+For more information on this topic, refer to [Mozilla VPN Telemetry Refactor](https://docs.google.com/document/d/1jyNZ_g_cUpZZsEr2hYwnwZgkxlvqTmoFUJKp_LAOPts/edit#heading=h.xararkvsnpss).
+
+## Proposed Solution
+
+> **Note** Honestly, I think for this proposed solution it's just easier to look at the pseudo-code in Appendix 1.
+
+A new metric will be added to both the daemonsession and the session pings: shared_session_id. This metric will have lifetime ping i.e. it is cleared from storage whenever a ping containing it is submitted. Thus it needs to be set again every time a new session/daemonsession ping is submitted.
```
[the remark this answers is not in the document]
brizental: The design doesn't really require it to be use lifetime.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+#### MAIN APPLICATION
+
+The session id is retrieved from local storage. If it is not empty, there are four possibilities:
+
+1. The same session from a previous app run is still ongoing. No action is required
+2. Another session is ongoing in which case a flush ping is required to submit data from a previous session and start recording data for the new session
+3. No session is ongoing but there is a lingering session id indicating that there is still lingering data from a previous session in storage and a flush ping is required
+4. No session is ongoing and no lingering session id is in storage. No action is required
+
+#### DAEMON
```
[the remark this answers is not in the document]
brizental: Hum, not sure I understand the question. Yes there are methods that get executed everytime the iOS netext is started and when the tunnel is started.

docs/rfcs/0002-session-id-sharing.md
brizental: I'd rather keep the pseudo-code generic. Do you have specific concerns for iOS? I'd rather go over them explicitly.

docs/rfcs/0002-session-id-sharing.md
brizental: Good catch, indeed that is missing.

docs/rfcs/0002-session-id-sharing.md
brizental: What would such a check be defending against?

docs/rfcs/0002-session-id-sharing.md
brizental: Not sure, probably no reason.

docs/rfcs/0002-session-id-sharing.md
brizental: Since we are already storing this value ourselves anyways I don't see the point of having a user lifetime.

oskirby:
So, I feel like the proposed implementation is overly complicated and we don't really need to rely on any local storage to achieve the desired result. We can note that the session identifier is attempting to track the lifetime of a connection attempt, which will by necessity always be a subset of the lifetime of the daemon. Therefore the storage of the session identifier can simply be moved into the daemon and it can live in-memory without needing persistent storage.

So, I would like to suggest the alternative approach:
1. Upon activation, the daemon creates a UUID to mark the beginning of a new session.
2. Upon server switch, and multihop, the daemon preserves the existing session UUID.
3. The session UUID is added to the various signals and statuses emanating from the daemon.
4. Upon deactivation the session UUID is set to an empty string.
5. The client, upon startup, queries the daemon's state and then learns of the session UUID and knows to attach it to any relevant telemetry pings.

This should also ensure that activation/deactivation attempts that are triggered through OS triggers (eg: directly via iOS network settings) also correctly track the UUID.

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
mcleinman: If we're having to manually set it before every ping - and often to the same thing it was already set to - that seems like a candidate for `user` lifetime.

I don't feel super-strongly about this, but it's how we handle `session_id`. It may make sense to align these two.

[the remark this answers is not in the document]
mcleinman: And thinking a bit more, is this metric taking over from `session_id`? Will we deprecate `session_id` after this?

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
mcleinman: Makes sense. Let's update the other one to be `flush_start` or something similar - so it's clear what the difference is. Can you add that to this doc.

docs/rfcs/0002-session-id-sharing.md
mcleinman: On iOS it's not a daemon, and so there isn't a concept of `init`. Where would that `init` block of psuedo code be on iOS? I think that information should be included here. (The other three daemon blocks have obvious analogous functions on iOS.) You're right, we don't need iOS-specific psuedocode overall, but it should be clear where the psuedocode will be implemented on both platforms - and I'm not sure it is for this bit.

docs/rfcs/0002-session-id-sharing.md
mcleinman: It defends against future regressions on `flush_init` checks. It's not needed, but it could be a good defensive check and shortcut future debugging in the case of a regression.

docs/rfcs/0002-session-id-sharing.md
mcleinman: `sessionID` currently has a `user` lifetime, so we don't need to keep re-setting it.

docs/rfcs/0002-session-id-sharing.md
mcleinman: Let me know when that case is added, and I'll re-review. Thanks.

mcleinman:
@oskirby that sounds fine to me. 

Bea, can you note this in the doc (which you likely  already thought of): We'll have to re-order some of the session activation Glean code in the app, so we don't send a "started session" ping from the app until after we get the new session UUID from the daemon.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+> **Note** Honestly, I think for this proposed solution it's just easier to look at the pseudo-code in Appendix 1.
+
+A new metric will be added to both the daemonsession and the session pings: shared_session_id. This metric will have lifetime ping i.e. it is cleared from storage whenever a ping containing it is submitted. Thus it needs to be set again every time a new session/daemonsession ping is submitted.
+
+Both daemon and main application will store the value of this metric in their local storages for synchronization purposes. More about this in the sections below.
+
+A new reason will be added to the session pings “flush_init”. As better described in the sections below, the processes will be able to identify that either the VPN tunnel or main application did not gracefully hit the deactivation signal i.e. were force killed. The “flush_init” ping will be used to flush any data remaining from the previous session.
+
+There will be four key triggers that will handle the shared_session_id: initialization, on VPN activation, ping timer and on VPN deactivation.
```
Gela, line 27: Is it possible that Daemon or the client could crash independently of each other and lose their session id? If that happens, I imagine a new session id is creating once it's restarted, how do we ensure that the other party knows about this and maintains the relevance somehow?

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,226 @@
...
+
+### ON VPN ACTIVATION
```
Gela, line 29: Nitpick: if there is a header, there should be a paragraph explaining the what/why even if it's pretty intuitive

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
brizental: Nope, this is a different metrics `shared_session_id`. After it's implemented and validated we can discuss maybe merging the two, but for now I'd rather have them separate.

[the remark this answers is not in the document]
brizental: I have rewritten the whole proposal and the rewrite has this metrics as `user` lifetime as suggested. Resolving this.

docs/rfcs/0002-session-id-sharing.md
brizental: The RFC has been rewritten and I am not sure this comment still applies. I'll resolve for now, but please reopen otherwise.

docs/rfcs/0002-session-id-sharing.md
brizental: Yes, there is a concept of init. See: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/apps/vpn/platforms/ios/iostunnel.swift#L17-L41. The terminology is different on iOS, but the network extension is just a separate process running some code.

docs/rfcs/0002-session-id-sharing.md
brizental: I'd rather we catch these types of regression on tests or worst case scenario on monitoring.

docs/rfcs/0002-session-id-sharing.md
brizental: This is applied in the rewrite.

docs/rfcs/0002-session-id-sharing.md
brizental: The rewrite makes this obsolete. Let me know if you agree, for now resolving.

docs/rfcs/0002-session-id-sharing.md
brizental: Hm, need to think if this still applied on the rewritten proposal.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+If the main application is active during a session end it will,
+
+1. Submit the session ping with reason `end`.
+2. Set `shared_session_id` metric is set to a known value. The same known value used for inactive
+   sessions on the `daemonsession` ping.
+
+#### On application start
+
+When the main application is started, just like in the daemon, there may be some lingering data
+from a previous session in storage. Om the main application, the session_id is not directly managed,
```
Gela, line 169:
```suggestion
from a previous session in storage. On the main application, the session_id is not directly managed,
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+  from the daemon?
+  Another core assumption here is that, if the daemon has an ongoing session the main application will,
+  most of the time, not run into any issues retrieving the shared_session_id from it. If there are too
+  many errors in communication that would mean that assumption was incorrect or there is a bug in the
+  implementation.
+
+## Considered Alternatives
+
+Following are the other considered alternatives to address the same issue as the proposed solution
+and the reasoning on why these were ultimately not the chosen alternatives.
```
Gela, line 203:
```suggestion
and the reasoning on why these were ultimately not the chosen solution.
```

docs/rfcs/0002-session-id-sharing.md
brizental: LOL, it's funny the spellchecker did not catch this.

docs/rfcs/0002-session-id-sharing.md
brizental: > Is it possible that Daemon or the client could crash independently of each other and lose their session id? If that happens, I imagine a new session id is created once it's restarted, how do we ensure that the other party knows about this and maintains the relevance somehow?

Ok so, rereading this. It is not possible. Not in the rewritten or in the original proposal. If either party crashes, the session is not lost, because it is in permanent storage.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+## Considered Alternatives
+
+Following are the other considered alternatives to address the same issue as the proposed solution
+and the reasoning on why these were ultimately not the chosen solution.
+
+### Collect all data on the daemon
+
+If all data were collected on the daemon, instead of having the two Glean architecture that is
+currently implemented that would not also fix the issue of joining daemon data with main application
+data, simply by not having main application data at all.
```
brizental, line 209: TODO: Rewrite this whole sentence. it makes no sense.

docs/rfcs/0002-session-id-sharing.md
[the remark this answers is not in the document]
brizental: Oops, forgot to include this comment on the rewrite. I'll do this tomorrow.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+When the main application is started, just like in the daemon, there may be some lingering data
+from a previous session in storage. On the main application, the session_id is not directly managed,
+so there is no way to know if there is lingering data or not.
+
+In this case, the main application will submit the `vpnsession` ping with reason `flush_init` on
+_every_ initialization. In case there is no lingering data, this is not really an issue. The ping
+will be sent with only the currently stored session id value.
+
+After sending the `flush_init` ping, the VPN will query the daemon for a session id.
+If a session id is received, that means there is an ongoing VPN session. In that case, the
```
mcleinman, line 177: Won't a session ID always be received? It's just whether it's the known inactive session value or not.

mcleinman (commented):
One small thing, otherwise this seems complete.

docs/rfcs/0002-session-id-sharing.md
brizental: Correct. But the daemon may not be running. In that case, no session id is received.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+1. The `daemonsession` ping is submitted with the reason `daemon_end`.
+2. The daemon's local storage for the `shared_session_id` is cleared.
+3. The `shared_session_id` metric is set to a known value. Due to its `User` lifetime,
+   this metric cannot be cleared from storage ever. To indicate a period of inactivity,
+   we will leverage a known UUID.
+
+#### On daemon process start
+
+The session end trigger may not be hit, if the daemon process is force killed. Therefore,
+whenever the daemon process is started the local storage will be searched for a lingering
```
mcleinman, line 129: We don't mean lingering anymore, right? We want "known inactive session ID"?

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+
+A new metric will be introduced to mobile platforms: `shared_session_id`. This metric will be the
+primary session identifier metric for Mozilla VPN mobile telemetry. This metric will have lifetime
+`User`, with this lifetime, once a metric is set it can only be overwritten but never cleared from
+the telemetry storage.
+
+The daemon process will be responsible for managing this metric. The daemon is the preferred
+place for this, because it is _guaranteed_ to be running throughout the whole VPN session.
+
+The main application will communicate with the daemon in order to get the correct value for
```
mcleinman, line 90: Can we add a couple lines about how the main application will communicate with the daemon and get the UUID back? I think this is already being used for both Android and iOS (and wouldn't require much new code), but would be good to call out in this doc.

mcleinman (commented):
This is looking great. Two small questions for clarity.

lesleyjanenorton (commented):
A few questions but otherwise this is looking really good Bea!

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+Both the operational system and the Mozilla VPN application expose features that allow the user
+to turn on the VPN without interacting with the main application. On mobile environments, when
+the VPN is turned on through any of these methods, the main application process is never launched
+while the daemon process is.
+
+### The issue with mobile session based telemetry
+
+When analyzing incoming telemetry from the Mozilla VPN application, it's a recurring requirement
+to be able to partition the incoming data by session. **A VPN session starts when the VPN is turned
+on through some user action and ends once it is turned off through some user action.**
```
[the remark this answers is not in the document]
brizental: Hum, it would still be a session. We would just not catch the `end` ping.

docs/rfcs/0002-session-id-sharing.md
brizental: I am not sure it is necessary to add such technical details to this document. Passing data back from daemon to main app is already implemented in both platforms we will just use the already existing infrastructure. Do you have a specific reason for this ask?

docs/rfcs/0002-session-id-sharing.md
brizental: It is lingering in this case. The daemon is still managing the UUID in it's own storage. A lingering session id for the daemon means the end event was not executed succesfully and there may be lingering data.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+The same three triggers which are important for the daemon and the `daemonsession` ping,
+are also important for the main application and the `vpnsession` ping.
+
+#### On session start
+
+If the main application is active during a session start it will,
+
+1. Signal to the daemon that a new session is to be started.
+2. It will then ask the daemon for the value of the shared session id it created.
+3. If the previous two communications where successful
```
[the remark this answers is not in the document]
brizental:
```suggestion
3. If the previous two communications were successful
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+   1. The `vpnsession` ping will be submitted with reason `flush`.
+   2. The value of the retrieved `shared_session_id` is recorded.
+   3. Other session metrics may be recorded from this point.
+   4. If the communication with the daemon fails, an error metric is recorded.
+
+> **Note**: If the retrieved session id is the known inactive session id,
+> this is considered an unsuccessful retrieval and the error path is taken.
```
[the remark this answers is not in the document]
brizental: What I mean by "error path" here is step 4 of the above steps. It does include recording an error metric.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+
+The same three triggers which are important for the daemon and the `daemonsession` ping,
+are also important for the main application and the `vpnsession` ping.
+
+#### On session start
+
+If the main application is active during a session start it will,
+
+1. Signal to the daemon that a new session is to be started.
+2. It will then ask the daemon for the value of the shared session id it created.
```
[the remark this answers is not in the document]
brizental: Ping submission is not a long running call. It just enqueues the ping, upload happens async in a separate thread / worker. So i'd say not risky.

There is no required recovery from the error path. On the validation step we will be able to tell if there an unexpectedly large number of sessions using the known inactive session id or if there are too many errors recorded. If there is, then that would mean we took the error path too much and we should consider addressing that. 

Also, there is always a flush ping before a new session for this reason too. Even if a given session is messed up, all the messed up data is flushed before a new session starts.

docs/rfcs/0002-session-id-sharing.md
mcleinman: These sorts of details can sometimes be things that add lots of complexity to an implementation (if it's not already solved or known how we'll do it, as I now know it is in this situation). While the document still lacks that level of detail, your comment puts my concerns to rest.

docs/rfcs/0002-session-id-sharing.md
mcleinman: Yup, total brainfart when I wrote that comment. Thanks.

mcleinman (approved):
This document seems technically sound to me, so I'm approving. 

I still have questions about whether the benefits of this work are worth the potential for future bugs, but Bea and I have discussed that point a lot. If we want to have a wider group discussion around this, I'm open to it - but I'm not blocking this moving forward.

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+If all data were collected on the daemon, instead of having the two Glean architecture that is
+currently implemented that would not also fix the issue of joining daemon data with main application
+data, simply by not having main application data at all.
+
+The issue with this approach, is that most of the telemetry collected on the Mozilla VPN application
+is in the main application. Button clicks, configuration, performance metrics and so on. If there were
+only a Glean instance of the daemon process, every time an event that requires telemetry happened on
+the main application that would need to be passed through to the daemon. Not only does that significantly
+increase the overhead of adding any new telemetry to the application, but it is also not possible
+for cases such as iOS where the daemon process is only alive while a session is active.
```
csadilek, line 216: Additionally, we'd also not be able to run experiments on the main application.

csadilek (approved):
Thank you for working through this!

As part of the intro and problem statement, it would great to call out use cases for data analysis we currently can't satisfy because we can't group by session.

I do have some questions about the communication between the main app the daemon to exchange the shared session ID, but I think those are best worked through on the implementation. I don't see them as blockers.

Just some nits and comments below, but this looks good to me! ✅ 🚢

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+iOS is similar, but the daemon process is only active while the VPN is turned ON.
+
+On Desktop environments the main application process is active while the application is active,
+independent of the application being foregrounded or backgrounded. Due to OS specific mechanisms
+such as [App Nap](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/AppNap.html)
+this process may go dormant intermittently after the application is backgrounded for a long time.
+
+#### The daemon process may be started independently from the main application on mobile platforms
+
+Both the operational system and the Mozilla VPN application expose features that allow the user
```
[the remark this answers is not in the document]
brizental:
```suggestion
Both the operating system and the Mozilla VPN application expose features that allow the user
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+That is easily addressed on desktop environments, by having the main application manage a session_id
+metric and add that to all session pings. Data from a given session all have the same session_id
+and can be joined using that identifier.
+
+On mobile environments, due to the lack of guarantees on the main application process being alive
+throughout most or any of the VPN session, having the main application manage the session_id is unreliable.
+This identifier needs to be rotated on every session start/end and the main application process may
+not be running at these times.
+
+Moreover, on mobile platforms, the fact that the main application process is killed whenever
```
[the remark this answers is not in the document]
brizental:
```suggestion
Moreover, on mobile platforms, the fact that the main application process may be killed whenever
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+
+There are three important triggers that require action on the shared session id and
+the `daemonsession` ping on the daemon process.
+
+#### On session start
+
+Once a new session is started,
+
+1. The `daemonsession` ping will be submitted with reason `flush`, in order to flush out
+   any lingering data before starting collection telemetry related to the new session.
```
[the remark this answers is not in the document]
brizental:
```suggestion
   any lingering data before starting to collect telemetry related to the new session.
```

lesleyjanenorton (approved):
Thank you for this, Bea. I'm looking forward to seeing this implemented and to what we learn from it!

lesleyjanenorton (approved):
Thank you for this, Bea. I'm looking forward to seeing this implemented and to what we learn from it!

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
+- Status:
+- Date: 2023-06-07
+- Author: [@brizental](https://github.com/brizental)
+- RFC PR: [#7112](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7112)
+- Implementation GitHub issue: TODO
```
brizental, line 5:
```suggestion
- Implementation GitHub issue: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7588
```

docs/rfcs/0002-session-id-sharing.md
```diff
@@ -0,0 +1,223 @@
...
+only a Glean instance of the daemon process, every time an event that requires telemetry happened on
+the main application that would need to be passed through to the daemon. Not only does that significantly
+increase the overhead of adding any new telemetry to the application, but it is also not possible
+for cases such as iOS where the daemon process is only alive while a session is active.
+
+### Share the management burdens of the shared session id
+
+The first iteration of the proposed solution, included the main application storing and managing
+the `shared_session_id` and passing it along to the daemon when possible. That can be read on
+[7e20c2c933fc868b3ea50d6eb35651b904628658](https://github.com/mozilla-mobile/mozilla-vpn-client/tree/7e20c2c933fc868b3ea50d6eb35651b904628658).
```
brizental, line 222:
```suggestion
the `shared_session_id` and passing it along to the daemon when possible.
```

PR #7124, opened 2023-06-08
VPN-4496 - Record health metrics on the Android daemon

brizental:
The connection health checks on the Daemon are different from the connection health checks on the client. This is a caveat worth noting at analysis time. I have added a longer description to the metrics to make this more explicit.

For this work I have considered that an unstable event happens when the connection is deemed worh a silent server switch while a no signal even happens when we have given up hope. Does that make sense?

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
@@ -10,6 +10,8 @@ import android.os.Build
 import android.os.CountDownTimer
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
+import org.mozilla.firefox.vpn.qt.GleanMetrics.Sample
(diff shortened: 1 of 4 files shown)


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -164,8 +166,8 @@ class ConnectionHealth(service: VPNService) {
     private val TaskCheckConnection = Runnable {
         kotlin.run {
             // capture stuff, as this the members will be set by another thread
-            val gateway = mGateway; // This lives inside the tunnel
-            val endpoint = mEndPoint; // This is the wg-endpoint (not in the tunnel)
+            val gateway = mGateway // This lives inside the tunnel
+            val endpoint = mEndPoint // This is the wg-endpoint (not in the tunnel)
```
brizental, line 170: This is not related. I just fixed it while I was fixing other lints.

strseb (approved):
giphy

PR #7176, opened 2023-06-13
VPN-3331 - WIP: Crash reporting for all! 

strseb:
- Fetch Crash Report info from Guardian
- Enable Crashreporting for Linux

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+bool TaskSentryConfig::handleNetworkResponse(const QByteArray& data) {
+  auto doc = QJsonDocument::fromJson(data);
+  if (!doc.isObject()) {
+    logger.error() << "Failed to get Sentry info";
+    return false;
+  }
+  // The only thing right now, we don't want to send
+  // crash-data, so we will only check for that.
```
brizental, line 60: I don't understand this comment.

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return false;
+  };
+
+  if (!obj.contains("endpoint")) {
+    logger.error() << "endpoint missing from Sentry info";
+    return false;
+  };
+  auto endpoint = obj["endpoint"].toString();
+  if (endpoint.isNull()) {
+    logger.error() << "endpoint missing from Sentry info";
```
brizental, line 78: nit:
```suggestion
    logger.error() << "Sentry provided endpoint is null";
```

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return false;
+  };
+  auto settings = SettingsHolder::instance();
+  settings->setSentryEndpoint(endpoint);
+  settings->setSentryDSN(dsn);
```
brizental, line 83: How come this is not constant

src/shared/sentry/sentryadapter.cpp
```diff
@@ -4,7 +4,10 @@
 
 #include "sentryadapter.h"
 
-#include <QtQml/private/qqmlengine_p.h>
+#ifndef UNIT_TEST
+// This is not found when building tests ... wierd.
```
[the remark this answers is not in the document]
strseb: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7176/commits/dfe8c8696fbfa2b322e5aced6b30d7a00c785a5d#diff-75169c8db799911a8b0af81ad1c17c24e41a868d224d756d249b7f8b070cee0c

tried that before, did not work :c `auth_test` did not build. In theory given sentry is part of shared-sources and that is an interface target, it "should" build. Also locally that builds too, so i blame something in github ci. 

>How come it's not necessary in tests?
Because i'm also "#ifndef UNIT_TEST" ing the code that calls this

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
strseb: Me neither, lol wtf

src/shared/sentry/sentryadapter.cpp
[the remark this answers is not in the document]
brizental: Got it. Hum, maybe because the code is not called ever it's being optimized away. Who know, looks correct to not import it in this case anyways.

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
strseb: Now currently you need to know our sentry creds to enable crash reporting. 
This is not good for linux, as when we build using PPA's we cannot have secrets :) 
But we also don't want that piece of info in our repo. 
That's why we moved that part to guardian and pull it from there.

brizental: Ah ok, so the whole reason for this is Linux essentially? I get not wanting it on our repo, but it could have been a build var otherwise.

strseb: > Ah ok, so the whole reason for this is Linux essentially? 
Yes. :D

brizental: Ok, would you mind adding a comment explaining this on the code? I guess the definition of TaskSentryConfig is the best place for it.

src/shared/cmake/sentry.cmake
```diff
@@ -81,29 +68,40 @@ if( ${_SUPPORTED} GREATER -1 )
...
                         -DANDROID_ABI=${ANDROID_ABI}
                         -DCMAKE_ANDROID_NDK=${ANDROID_NDK_ROOT}
                         -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK_ROOT}/build/cmake/android.toolchain.cmake
                         -DSENTRY_BACKEND=inproc
             )
+    
+    endif()
+    if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+        target_compile_definitions(shared-sources INTERFACE SENTRY_BUILD_STATIC)
+        # Compile Static for apple and link to libsentry.a
```
brizental, line 89: This comment is stale.

src/shared/sentry/sentryadapter.cpp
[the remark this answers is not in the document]
brizental: I guess this comment can go though.

src/shared/sentry/sentryadapter.cpp
```diff
@@ -37,17 +41,25 @@ SentryAdapter::SentryAdapter() { MZ_COUNT_CTOR(SentryAdapter); }
...
   if (!Feature::get(Feature::Feature_sentry)->isSupported()) {
     return;
   }
-  if (QString(Constants::SENTRY_DSN_ENDPOINT).isEmpty() ||
-      QString(Constants::SENTRY_ENVELOPE_INGESTION).isEmpty()) {
+  auto settings = SettingsHolder::instance();
+  if (!settings->hasSentryEndpoint() || !settings->hasSentryDSN()) {
+    // If we have no info on where to put crash info, let's
+    // query that and retry later.
     logger.error() << "Sentry failed to init, no sentry config present";
```
brizental, line 54: This is not really an error but an expected state when sentry has not been initialized before. Maybe reword and make it not an error?

brizental (approved):
Just nits. LGTM.

src/shared/sentry/sentryadapter.cpp
Gela, line 8:
```suggestion
// This is not found when building tests ... weird.
```

src/shared/sentry/sentryadapter.cpp
```diff
@@ -37,17 +41,25 @@ SentryAdapter::SentryAdapter() { MZ_COUNT_CTOR(SentryAdapter); }
...
+    auto t = new TaskSentryConfig();
```
Gela, line 55: In this instance it's not as prevalent but generally speaking I try to avoid one character var names
```suggestion
    auto sentryConfig = new TaskSentryConfig();
```

src/shared/tasks/sentry/tasksentry.cpp
```diff
@@ -76,12 +76,20 @@ void TaskSentry::run() {
 
 void TaskSentry::sendRequest() {
   NetworkRequest* request = new NetworkRequest(this, 200);
+  auto settings = SettingsHolder::instance();
+  auto endpoint = settings->sentryEndpoint();
+  auto dsn = settings->sentryDSN();
+  if (endpoint.isNull() || dsn.isNull()) {
+    // We could have not even been initializing sentry without that?!
```
Gela, line 83: Should we rephrase a comment like this to something a little more specific? maybe like:
```suggestion
    // We should not be able to initialize Sentry without an endpoint and DSN
```

Gela (approved):

.github/workflows/functional_tests.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v3
         with:
-          submodules: 'true'
+          submodules: 'recursive'
```
vinocher, line 27: Should docs\dev-setup\index.md be updated to add --recursive to 'git submodule update --init'?

src/shared/tasks/sentry/tasksentry.cpp
```diff
@@ -76,12 +76,20 @@ void TaskSentry::run() {
 
 void TaskSentry::sendRequest() {
   NetworkRequest* request = new NetworkRequest(this, 200);
```
vinocher, line 78: Won't this allocation be leaked if the code returns at line 86? Should this line be moved below the new code block?

src/shared/tasks/sentryconfig/tasksentryconfig.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef TASKSENTRYCONFIG_H
+#define TASKSENTRYCONFIG_H
+
+#include <QByteArray>
+#include <QObject>
+
+#include "task.h"
+
+class TaskSentryConfig final : public Task {
```
vinocher, line 14: In general, it would be very useful to add a class header comment that summarizes purpose and usage. But that pattern doesn't seem to be used in this code base.

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+void TaskSentryConfig::run() {
+  NetworkRequest* request = new NetworkRequest(this, 200);
+
+  QUrl url(AppConstants::apiBaseUrl());
+  url.setPath("/api/v1/vpn/crashreporting");
+
+  connect(request, &NetworkRequest::requestFailed, this,
+          [this](QNetworkReply::NetworkError error, const QByteArray&) {
+            Q_UNUSED(error);
```
vinocher, line 37: Nit: Instead of Q_UNUSED, couldn't the unused parameter name 'error' be removed?

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,83 @@
...
+              SentryAdapter::instance()->init();
+            }
+            emit completed();
+          });
+
+  request->get(url);
+}
+
+bool TaskSentryConfig::handleNetworkResponse(const QByteArray& data) {
+  auto doc = QJsonDocument::fromJson(data);
```
vinocher, line 54: Shouldn't doc be null checked to handle fromJson failure?

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,83 @@
...
+          [this](QNetworkReply::NetworkError error, const QByteArray&) {
+            Q_UNUSED(error);
+            logger.error() << "Failed to get Sentry info";
+            emit completed();
+          });
+  connect(request, &NetworkRequest::requestCompleted, this,
+          [this](const QByteArray& data) {
+            if (handleNetworkResponse(data)) {
+              // We now should have sentry info - let's init
+              SentryAdapter::instance()->init();
```
vinocher, line 45: Consider moving this into handleNetworkResponse for better cohesion, instead of depending on the side effect that handleNetworkResponse has set the sentry config info into SettingsHolder.

src/shared/tasks/sentryconfig/tasksentryconfig.h
```diff
@@ -0,0 +1,27 @@
...
+  Q_DISABLE_COPY_MOVE(TaskSentryConfig)
+
+ public:
+  TaskSentryConfig();
+  ~TaskSentryConfig();
+
+  void run() override;
+
```
vinocher, line 22: Should the DeletePolicy of this task be changed so it is not deleted?

src/shared/settingslist.h
```diff
@@ -224,5 +224,27 @@ SETTING_STRING(token,        // getter
...
 )
 
+SETTING_STRING(sentryDSN,        // getter
+               setSentryDSN,     // setter
+               removeSentryDSN,  // remover
+               hasSentryDSN,     // has
+               "sentry/dsn",     // key
+               "",               // default value
+               false,            // user setting
+               false,            // remove when reset
```
vinocher, line 234: Is it better to remove when reset?

linux/debian/control
```diff
@@ -27,7 +27,9 @@ Build-Depends: debhelper (>= 11.2),
                qmake6 (>=6.2.0~),
                qt6-base-dev (>=6.2.0~),
                qt6-base-dev-tools (>=6.2.0~),
+               qt6-base-private-dev (>=6.2.0~),
```
vinocher, line 30: For my info, could you please explain why these packages with private headers are needed?

vinocher:
Just checking, are the changes in vpn\translations\i18n needed for this change?

src/shared/tasks/sentry/tasksentry.cpp
```diff
@@ -76,12 +76,20 @@ void TaskSentry::run() {
...
+    return;
+  }
 
   request->requestInternal().setHeader(QNetworkRequest::ContentTypeHeader,
                                        "application/x-sentry-envelope");
-  request->requestInternal().setRawHeader("dsn",
-                                          Constants::SENTRY_DSN_ENDPOINT);
-  request->post(QUrl(Constants::SENTRY_ENVELOPE_INGESTION), m_envelope);
+  request->requestInternal().setRawHeader("dsn", dsn.toLocal8Bit());
+  request->post(QUrl(endpoint), m_envelope);
```
vinocher, line 92: Should QUrl be validated (StrictMode parsing + IsValid) before the post?

src/shared/sentry/sentryadapter.cpp
```diff
@@ -159,6 +171,7 @@ void SentryAdapter::declineCrashReporting() {
 }
 
 void SentryAdapter::captureQMLStacktrace(const char* description) {
+#ifndef UNIT_TEST
```
vinocher, line 174: Is this because we don't want to report crashes for unit tests?

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,83 @@
...
+    logger.error() << "DSN missing from Sentry info";
+    return false;
+  };
+
+  if (!obj.contains("endpoint")) {
+    logger.error() << "endpoint missing from Sentry info";
+    return false;
+  };
+  auto endpoint = obj["endpoint"].toString();
+  if (endpoint.isNull()) {
```
vinocher, line 75: Should QUrl be used here to parse the endpoint to check that the provided URL is valid?

.github/workflows/functional_tests.yaml
strseb: Good catch! added that.

linux/debian/control
strseb: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7203

We currently have an influx of crashes that are triggered from "something" called in the js-engine. 
So in sentryadapter we added a crashandler that can hook into the JS stack and produce a mini stacktrace - that will be attached in a sentry report. 
Those api's are not public. Since this pr will enable us to also ship sentry on linux we know have a dependency on both private qt and qml headers :)

src/shared/sentry/sentryadapter.cpp
strseb: No sadly this is me having a problem getting this to build on github CI. 
For an (for me) unknown reason, the compiler cannot find the private headers i.e  `include <QtQml/private/qqmlengine_p.h>` - which will cause this to fail if we build the tests. 
Given that we currently have no callers when running tests, i'd like to disable this and fix it once we actually need that.

src/shared/tasks/sentry/tasksentry.cpp
strseb: Not in this case - `NetworkRequest` is self-destructive after the request has finished :) 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/shared/networkrequest.h#L26

src/shared/tasks/sentryconfig/tasksentryconfig.h
strseb: I appreciate the callout :)

src/shared/tasks/sentryconfig/tasksentryconfig.h
strseb: Good catch! I've made it rescheudleable, given that this is not time critical :)

src/shared/settingslist.h
strseb: Oh right, yes let's do that :)

strseb:
> Just checking, are the changes in vpn\translations\i18n needed for this change?

Nope,  remove that :)

src/shared/tasks/sentry/tasksentry.cpp
vinocher: Thanks for pointing out the following in NetworkRequest will free the allocation:
connect(&m_timer, &QTimer::timeout, this, &NetworkRequest::maybeDeleteLater);
connect(m_reply, &QNetworkReply::finished, this,
          &NetworkRequest::maybeDeleteLater);

src/shared/tasks/sentryconfig/tasksentryconfig.h
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @brief TaskSentryConfig fetches Sentry Configuration
+ *
+ * This Task will schedule a Network request to Guardians
+ * `crashreporting` endpoint to get info on where to
+ * post crashreports to.
+ *
+ * After a successful fetch, it will store the config
+ * in the Settings an Call SentryAdapter::init to
```
vinocher, line 22: Nit: Typo 'an'

src/shared/tasks/sentryconfig/tasksentryconfig.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (!obj.contains("endpoint")) {
+    logger.error() << "endpoint missing from Sentry info";
+    return;
+  };
+  auto endpoint = obj["endpoint"].toString();
+  if (endpoint.isNull()) {
+    logger.error() << "Sentry provided endpoint is null";
+    return;
+  };
+  if (!QUrl(endpoint).isValid()) {
```
vinocher, line 75: Should strict mode be used here? Also, is it sufficient to rely on the check here and remove the URL validity check in  TaskSentry::sendRequest?

PR #7196, opened 2023-06-14
QOL: Conda/iOS - Auto Qt setup

strseb:
On Windows + Android we have a `conda-setup-qt.sh` that will download and setup qt for the current env. 

Since i'm too lazy to remember where qt-lives for every platform i like to have that setuped. 

- This does download + install QT of the .env's qt-version. 
- Add's activation scripts to set `QT_HOST_PATH` -> (so we can conda-pack the env for xcode-cloud) 
- add's a `qt-cmake` helper wrapper script that calls `<Qt unzipped path>/6.2.4/ios/bin/qt-cmake -S . -B build-ios -GXcode -DQT_HOST_PATH=<Qt unzipped path>/6.2.4/macos`, so i don't need to care about that.

### DIFF
--- a/scripts/cmake/golang.cmake
+++ b/scripts/cmake/golang.cmake
@@ -55,6 +55,7 @@ function(build_go_archive OUTPUT_NAME MODULE_FILE)
                     CGO_LDFLAGS="${GOBUILD_CGO_LDFLAGS}"
                     GOOS=${GOBUILD_GOOS}
                     GOARCH=${GOBUILD_GOARCH}
+                    GOROOT=$ENV{GOROOT}
                 ${GOLANG_BUILD_TOOL} build ${GOBUILD_ARGS} -o ${ABS_OUTPUT_NAME}
     )
 endfunction(build_go_archive)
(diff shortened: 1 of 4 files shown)


scripts/cmake/golang.cmake
```diff
@@ -55,6 +55,7 @@ function(build_go_archive OUTPUT_NAME MODULE_FILE)
                     CGO_LDFLAGS="${GOBUILD_CGO_LDFLAGS}"
                     GOOS=${GOBUILD_GOOS}
                     GOARCH=${GOBUILD_GOARCH}
+                    GOROOT=$ENV{GOROOT}
```
strseb, line 58: Unrelated nitpick, we should capture the GOROOT when configuring, i think.

oskirby: I would recommend that instead of capturing `GOROOT` I would point this to something like `${CMAKE_BUILD_DIR}/goroot` so that golang build stuff also gets cleared when we delete the build directory.

Gela (approved):

oskirby (approved):

Issue #7155, opened 2023-06-12
Support Qt 6.4/6.5 Builds


data-sync-user:
When trying to build using CMake with Qt 6.5.1, the following error is incurred: 

* `ld: can't link with bundle (MH_BUNDLE) only dylibs (MH_DYLIB) file '~/Qt/6.5.1/macos/./plugins/sqldrivers/libqsqlite.dylib' for architecture arm64`

`CMAKE_OSX_DEPLOYMENT_TARGET` also needs to be bumped to 10.15

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-4980)
┆Reporter: Matt Lichtenstein

data-sync-user:
➤ Valentina Virlics commented:

Matt Lichtenstein / Naomi Kirby Is there something QA needs to check here? Thank you!

data-sync-user:
➤ Matt Lichtenstein commented:

Qt 6.5 will likely require a regression test on the entire application, but I don’t think it makes sense to do that testing under this task.

Naomi Kirby would you agree?

data-sync-user:
➤ Valentina Virlics commented:

We will mark this as Done and label it “qa-verified“ after the regression period for 2.16 is finalized.

data-sync-user:
➤ Bianca Hidecuti commented:

Hey Matt Lichtenstein,

We were informed by Lesley Norton that the Qt version should be updated to 6.5.1 on MacOS, iOS, and Android, but we noticed that the Qt version for Android and iOS on the latest 2.17 builds is 6.2.4. 

For the Linux Jammy 2.17 build the Qt version is also 6.2.4, while on Linux Lunar the Qt version is 6.4.2. 

Could you please take a look at this? Thank you in advance!

data-sync-user:
➤ Paul Oiegas commented:

Hi team, I want to clearly understand what is required from QA for this request and maybe for further work unrelated to this request.

  1. Initially, we understood that this request needs Regression testing, but we got confirmation from Matt that it can wait for the RC week and be tested together with the normal Regression run.
However, Lesley Norton later told us on Slack that we shouldn't wait for the RC week to test this, as it needs Regression testing ASAP.
Does this request actually need specific Regression testing, or normal Feature testing would be enough? 

[...]

PR #7212, opened 2023-06-15
VPN-4980: Fix build on Qt 6.4+

oskirby:
## Description
At present, the VPN client fails to build on Qt 6.4 and beyond. This seems to be due to a few issues:
 - Plugins are now compiled as frameworks.
 - Qt 6.5 now has policies, and generates warnings if left unset.
 - Qt 6.5 requires a MacOS deployment target of 10.15 or later.

## Reference
Github issue: #7155 ([VPN-4980](https://mozilla-hub.atlassian.net/browse/VPN-4980))

## Checklist
    
[VPN-4980]: https://mozilla-hub.atlassian.net/browse/VPN-4980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,10 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten" AND
     )
 endif()
 
+if(QT_KNOWN_POLICY_QTP0001)
+    qt_policy(SET QTP0001 NEW)
+endif()
+
 message("Using Qt version ${Qt6_VERSION}")
 add_definitions(-DQT_DEPRECATED_WARNINGS)
 add_definitions(-DQT_DISABLE_DEPRECATED_BEFORE=0x050F00)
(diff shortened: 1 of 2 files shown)


MattLichtenstein (approved):
Awesome!

I also noticed the macOS target version had to be updated. For some reason, [the docs say qt 6.5 requires a macOS 11 target](https://doc.qt.io/qt-6/macos.html) but 10.15 works fine for me as well. Any idea why that might be?

oskirby:
> I also noticed the macOS target version had to be updated. For some reason, [the docs say qt 6.5 requires a macOS 11 target](https://doc.qt.io/qt-6/macos.html) but 10.15 works fine for me as well. Any idea why that might be?

The Qt documentation has been saying that the minimum MacOS version is 11 since they released Qt 6.0, so this isn't something new, and I don't really know how to test something that old anyways because all of our builds and testing is done on 12.x and later.

FWIW, I have only really verified that the builds succeed there may be runtime bugs that lurk in the resulting build yet to be found.

oskirby:
I also found that the gigantic pile of QML rendering bugs appear to have vanished on Qt 6.5.1.

PR #7216, opened 2023-06-15
VPN-4881: Fix 2.16 update message for old windows clients

MattLichtenstein:
## Description

- Update strings for 2.16 update message for Windows clients on 2.10-2.12 who need to upgrade via web

## Reference

[VPN-4881: Implement 'What's New' and 'Update' message for 2.16](https://mozilla-hub.atlassian.net/browse/VPN-4881)

### DIFF
--- a/addons/message_update_v2.16/enable.js
+++ b/addons/message_update_v2.16/enable.js
@@ -46,8 +46,8 @@
   api.addon.composer.remove('c_3');
 
   api.addon.setTitle(
-    'message.message_update_v2.15.block.extra_1', 'Download Mozilla VPN 2.15')
+    'message.message_update_v2.16.block.extra_1', 'Download Mozilla VPN 2.16')
   api.addon.setSubtitle(
-    'message.message_update_v2.15.block.extra_2',
-    'We’ve released an updated version of Mozilla VPN! Download it today to get the newest features and bug fixes:');
+    'message.message_update_v2.16.block.extra_2',
+    'We’ve released an updated version of Mozilla VPN! Download it today to get the newest features and bug fixes.');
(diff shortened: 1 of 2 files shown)


brizental (approved):

MattLichtenstein:
@flodolo fyi this is going to require translations

flodolo:
> @flodolo fyi this is going to require translations

How? This string is not changing the manifest, which is the only thing exposed to localization (and that's why I wasn't pinged for review).

flodolo:
AFAIK, you need to update this and use a new ID in the process
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/addons/message_update_v2.16/manifest.json#L56-L58

MattLichtenstein:
@flodolo the JS function conditionally replaces strings in the manifest. I saw translations for the old strings [here](https://pontoon.mozilla.org/nl/mozilla-vpn-client/addons/message_update_v2.16/strings.xliff/?string=281190)

flodolo:
> @flodolo the JS function conditionally replaces strings in the manifest. I saw a translation for one of the old strings [here](https://pontoon.mozilla.org/nl/mozilla-vpn-client/addons/message_update_v2.16/strings.xliff/?string=281190)

And those are extracted from the manifest file. What I'm saying is that the change in this PR won't show up for localization at any point, unless I'm very wrong in my understanding of the scripts extracting strings for addons.

MattLichtenstein:
@flodolo I've updated the manifest id's and text - these should be marked for localization now, right?

flodolo (approved):
Yes, now it will show up for localization.

PR #7221, opened 2023-06-15
fix iOS crash on fresh install

mcleinman:
## Description

On a fresh install (so no VPN profile already installed), I got a crash b/c we're force unwrapping this object after the user clicks the button in the onboarding "collect data?" screen, but this object doesn't exist yet. This fixes the crash.

This bug got uplifted to the 2.16 release branch, so this fix will also have to.

## Reference

N/A

## Checklist
    
### DIFF
--- a/src/shared/platforms/ios/iosgleanbridge.swift
+++ b/src/shared/platforms/ios/iosgleanbridge.swift
@@ -29,7 +29,7 @@ public class IOSGleanBridgeImpl : NSObject {
       
       do {
             Logger.global?.log(message: "Sending telemetry state change message.");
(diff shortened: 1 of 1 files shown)


mcleinman:
This isn't it.

If there is already a VPN Network Extension (you’ve previously installed the app and turned on the VPN), all works well with either ! or ? there.

If there is not a VPN Network Extension (fresh install for instance), the ! version crashes while logging in but the ? version doesn’t let you turn on the VPN from the homescreen (perhaps it may think it never finished initializing?). To be clear: Both of these fail, but in different ways.

PR #7230, opened 2023-06-16
VPN-4881: Fix 2.16 update/download message

MattLichtenstein:
## Description

- Remove unnecessary blocks from the 2.16 update / download message when showing for Windows 2.10, Windows 2.11, and Windows 2.12 (see [here](https://mozilla-hub.atlassian.net/browse/VPN-4034) for more info on why this is an issue)

## Reference

[VPN-4881: Implement 'What's New' and 'Update' message for 2.16](https://mozilla-hub.atlassian.net/browse/VPN-4881)

### DIFF
--- a/addons/message_update_v2.16/manifest.json
+++ b/addons/message_update_v2.16/manifest.json
@@ -22,11 +22,6 @@
         "content": "Be sure to update the app to make it even easier to protect your devices and your activity online!"
       },
       {
-        "id": "c_2b",
-        "type": "text",
-        "content": "Be sure to get the latest version to make it even easier to protect your devices and your activity online!"
-      },
-      {
         "id": "c_3",
         "type": "button",
         "style": "primary",
@@ -50,12 +45,7 @@
       {
         "id": "extra_1_216",
         "type": "text",
-        "content": "Download Mozilla VPN 2.16"
-      },
-      {
-        "id": "extra_2_216",
-        "type": "text",
-        "content": "We’ve released an updated version of Mozilla VPN! Download it today to get the newest features and bug fixes."
+        "content": "Download the new Mozilla VPN"
       }
     ]
   }
--- a/addons/message_update_v2.16/enable.js
+++ b/addons/message_update_v2.16/enable.js
@@ -1,7 +1,6 @@
 (function (api) {
   // Extra_1 and extra_2 are used only to have a localized string.
-  api.addon.composer.remove('extra_1');
-  api.addon.composer.remove('extra_2');
+  api.addon.composer.remove('extra_1_216');
 
   if (('updateTime' in api.settings)) {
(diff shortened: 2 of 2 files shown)


addons/message_update_v2.16/manifest.json
```diff
@@ -50,12 +45,7 @@
       {
         "id": "extra_1_216",
```
MattLichtenstein, line 46: @flodolo I [updated this ID for translations yesterday](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7216/commits/65a12d28565bc8fca8f5bd660a03863813532812), do I need to update it again

flodolo: Yes, and I'd like to understand how we avoid this level of churn.

This is the third request for this file that will get to localizers :-(

MattLichtenstein: This was a last minute string change, so, ideally, better planning 😅

I'm not sure how feasible this is, but it would be great if we could come up with a way to not push strings to localizers until they have been verified by QA possibly

flodolo: > This was a last minute string change, so, ideally, better planning 😅
> 
> I'm not sure how feasible this is, but it would be great if we could come up with a way to not push strings to localizers until they have been verified by QA possibly

You only need to let me know, that's a manual process (automation creates a pull request, I need to review it and merge it).

Note that we're already one week past string freeze, and we're still adding strings, so I assumed we needed that string out as soon as possible.

flodolo (approved):

addons/message_update_v2.16/manifest.json
MattLichtenstein: It was a good assumption, and if you had asked me yesterday, I would've definitely given you the go-ahead anyways. Because it's only a one day difference, in terms the of release timelines, I think it's probably fine to go ahead with these new strings. If this was going to push the release out by weeks, than yea, we could definitely push back on these string changes.

I'm guessing the bigger downside here is that we don't want our volunteer translators wasting there time translating things we aren't going to use. If they haven't already been translated by localizers, is there a way to remove the unused strings, or mark them somehow as "do not translate"?

flodolo: The strings will be removed once we merge this, assuming they're not used in any other branch.

There's no way around it.

MattLichtenstein:
@flodolo Can we hold off on translating these strings while I get a spot check done and approved by Santiago?

flodolo:
> @flodolo Can we hold off on translating these strings while I get a spot check done and approved by Santiago?

Yep. I'll wait for a green light here.

MattLichtenstein:
@flodolo we are good to go for translations!

PR #7234, opened 2023-06-16
VPN-2870 - reconnect About Us macOS menu item

mcleinman:
## Description

macOS includes an `About [app]` option in the menu, and I can't find a way to suppress this. Currently this menu option does nothing in our macOS app. 

Ideally this menu item would deep link into the About Us menu, but we can't deep link currently. Santiago is comfortable just going to the Settings screen for now (as it's better than having nothing happen). I spent some time w/ Matt L and we didn't find a hacky way to do deep links. And it didn't seem worth building that functionality into this PR.

Code path for this menu option:
- `MacOSMenuBar`'s initializer creates `m_aboutAction`, which connects the menu item to `MozillaVPN::OpAbout`
- In `MozillaVPN::registerExternalOperations` connects `OpAbout` with `MozillaVPN::requestAbout`
- `MozillaVPN::requestAbout` emits `aboutNeeded`
- And then nothing happened, as no function picked it up. This PR adds that back in.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2870

## Checklist
    
- [X]  I have performed a self review of my own code


bakulf (changes requested):
Move this to navigator.cpp

mcleinman:
> Move this to navigator.cpp

@bakulf good call to move it somewhere better, thanks. I've updated this. But isn't `Navigator.cpp` shared code? If we ever add another app to this codebase, this update will be problematic. Should it be moved to `mozillavpn.cpp`, or should we start some VPN-specific nav class (like we do with app features), or something else?

bakulf (approved):

PR #7235, opened 2023-06-16
Disable sign in button if input is empty

lesleyjanenorton:
## Description

This PR:
- Disables the 'Sign in' button when the password input has a text.length of 0. 
- Updates the functional test for this view

Note: The password length requirement to create an account in the client is 8 BUT I _think_, and will confirm, that it has historically possible to create an account with a password that is fewer than 8 characters. I'm therefore hesitant to stipulate that the 'Sign in' button stay disabled until the password input text.length === 8. 

## Checklist
    
### DIFF
--- a/src/apps/vpn/ui/authenticationInApp/ViewAuthenticationSignIn.qml
+++ b/src/apps/vpn/ui/authenticationInApp/ViewAuthenticationSignIn.qml
@@ -59,7 +59,7 @@ MZInAppAuthenticationBase {
         objectName: "authSignIn"
         id: authInputs
 
-        _buttonEnabled: MZAuthInApp.state === MZAuthInApp.StateSignIn && !activeInput().hasError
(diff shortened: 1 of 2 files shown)


MattLichtenstein:
It seems there is a bit more to this bug. From the bug description: 

> Have a Fx account without a password set (this can be achieved by subscribing to Mozilla VPN from [homepage](https://www-dev.allizom.org/en-US/products/vpn/));

It looks like signing-in with a password of any length using one of these password-less accounts just puts the UI in limbo

Still a good change, regardless 🙂

lesleyjanenorton:
@MattLichtenstein you're 100% correct. Removed the ticket refs and will follow up with Santiago and UX to hammer out what should happen here.

lesleyjanenorton:
Closing since this has gone quite stale

PR #7237, opened 2023-06-17
Update languages.json (also fix replacer test)

lesleyjanenorton:
## Description

We're releasing 2.15.1 specifically to introduce new locales. Since we don't want to _lose_ locales when updating from 2.15.1 -> 2.16.0, we'll need to keep 2.16.0 locales in sync with latest main, at least until 2.15.1 is out the door. 

Note to selves: When we cut the 2.15.1 branch, we'll need to address string IDs that have been deleted since 2.15.0. 

## Checklist
    

src/apps/vpn/translations/i18n
flodolo, line 1: Remove this?

flodolo (approved):

PR #7256, opened 2023-06-19
VPN-5035 - Always execute TunnelManager.initialize completion handler

brizental:
+ Two other minor bug fixes.

### DIFF
--- a/src/apps/vpn/commands/commandui.cpp
+++ b/src/apps/vpn/commands/commandui.cpp
@@ -236,21 +236,14 @@ int CommandUI::run(QStringList& tokens) {
     auto offlineStorageDirectory =
         QDir(QmlEngineHolder::instance()->engine()->offlineStoragePath() +
              "/Databases");
-    logger.debug()
-        << "HELLO"
-        << QmlEngineHolder::instance()->engine()->offlineStoragePath() +
-               "/Databases";
     if (offlineStorageDirectory.exists()) {
       QStringList files = offlineStorageDirectory.entryList();
       for (const QString& file : files) {
-        logger.debug() << "HELLO"
-                       << offlineStorageDirectory.absoluteFilePath(file);
         // Note: This is kinda dumb, it doesn't really know that this is
         // Glean.js' storage. Since Glean.js was the only thing using sqlite in
         // the app at the time of implementation this is fine. If we ever add
         // other SQLite using things, then we need to change this.
         if (file.endsWith(".sqlite")) {
-          logger.debug() << "HELLO" << file;
           QFile::remove(offlineStorageDirectory.absoluteFilePath(file));
         }
       }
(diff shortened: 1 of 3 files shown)


src/apps/vpn/platforms/ios/iostunnelmanager.swift
```diff
@@ -84,12 +91,12 @@ class TunnelManager {
             if tunnel == nil {
                 Logger.global?.log(message: "Creating the tunnel")
                 self!.tunnel = .success(NETunnelProviderManager())
+                completionHandler();
```
brizental, line 94: @mcleinman Not calling the handler to finalize initialization in all possible branches was what caused the VPN toggle not to work. Unit tests would have caught this. Do we have already a ticket to create Swift unit tests infra?

src/apps/vpn/commands/commandui.cpp
```diff
@@ -236,21 +236,14 @@ int CommandUI::run(QStringList& tokens) {
...
       QStringList files = offlineStorageDirectory.entryList();
       for (const QString& file : files) {
-        logger.debug() << "HELLO"
-                       << offlineStorageDirectory.absoluteFilePath(file);
         // Note: This is kinda dumb, it doesn't really know that this is
         // Glean.js' storage. Since Glean.js was the only thing using sqlite in
         // the app at the time of implementation this is fine. If we ever add
         // other SQLite using things, then we need to change this.
         if (file.endsWith(".sqlite")) {
-          logger.debug() << "HELLO" << file;
```
brizental, line 253: Unrelated changes, leftover from the removal of Glean.js.

src/apps/vpn/platforms/ios/iostunnelmanager.swift
brizental: I filed https://mozilla-hub.atlassian.net/browse/VPN-5048 to write the tests. I will block on the infra ticket, if it exists.

src/shared/platforms/ios/iosgleanbridge.swift
```diff
@@ -27,18 +27,20 @@ public class IOSGleanBridgeImpl : NSObject {
           defaults.set(isTelemetryEnabled, forKey: Constants.UserDefaultKeys.telemetryEnabled)
       }
       
-      do {
-            Logger.global?.log(message: "Sending telemetry state change message.");
+      if (TunnelManager.session != nil) {
+        do {
+            Logger.global?.log(message: "Attempting to send telemetry state change message.");
           try TunnelManager.session!.sendProviderMessage(
```
mcleinman, line 33:
```suggestion
      if let session = TunnelManager.session {
        do {
            Logger.global?.log(message: "Attempting to send telemetry state change message.");
          try session.sendProviderMessage(
```

`if let` checks if it's nil, and if not it enters the if block with `session` as a non-optional.

mcleinman (approved):
Looks great, thank you. Couple small nits, feel free to reject any/all of my feedback.

src/apps/vpn/platforms/ios/iostunnelmanager.swift
brizental: Thanks!

PR #7258, opened 2023-06-19
[Perf]: Enable QML cachegen on /nebula

strseb:
## Description

Right now we only bundle our qml stuff with the client. [However qt can pre-generate bytecode on compile: ](https://codereview.qt-project.org/c/qt/qtdeclarative/+/382748/12). Especially on iOS where we do not have access to a jit, should help us to get a bit more perf budget :) 

For that i've removed the `.qrc` files in nebula and replaced it with `qml_add_module` calls in cmake land.

You can see it's working, as we will now see cpp files generated by qmlcachegen - :) 
```
[506/959] Building CXX object nebula/ui/components/CMakeFiles/components.dir/.rcc/qmlcache/components_MZAvatar_qml.cpp.o
```

See also: https://doc.qt.io/qt-6/qtquick-deployment.html#ahead-of-time-compilation


nebula/nebula.cpp
```diff
@@ -4,15 +4,8 @@
...
 
 void Nebula::Initialize(QQmlEngine* engine) {
-  Q_INIT_RESOURCE(components);
-  Q_INIT_RESOURCE(themes);
-  Q_INIT_RESOURCE(nebula_resources);
-  Q_INIT_RESOURCE(compatQt6);
-  Q_INIT_RESOURCE(resourcesQt6);
-  Q_INIT_RESOURCE(utils);
-
   engine->addImportPath(QRC_ROOT);
```
oskirby, line 10: What's the impact of removing `addImportPath()`? It would be nice if we could just do away with the need to initialize the nebula library entirely.

strseb:
>Looks good, and this brings some much needed cleanup to the nebula QML module.
Any reason to go with qt_add_library over creating a [QML module](https://doc.qt.io/qt-6/cmake-build-reusable-qml-module.html)?

Nope! I did not know i can skip qt_add_library when using `qt_qml_add_module `:) - changed that

PR #7269, opened 2023-06-20
[RFC] VPN-4918 Refactor Controller and move connectivity logic to Connection Manager

Gela:
## Description

RFC to breakdown how we want to undertake the `Controller` refactor work.

## Feedback Request Until
Friday July 14th (7/14/2023)

## Reference

   https://mozilla-hub.atlassian.net/browse/VPN-4918

## Checklist
    
### DIFF
--- a/docs/rfcs/0003-rfc-controller_refactor.md
+++ b/docs/rfcs/0003-rfc-controller_refactor.md
@@ -1,0 +1,114 @@
+- Status: In review
+- Date: 2023-06-20
+- Author: [@gela](https://github.com/gela)
+- RFC PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7269
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+Introdoce a new `ConnectivityManager` component which sits between the frontend and controller to handle all connectivity related logic and keep the controller code lean and minimal.
+
+## Motivation
+The current controller code is complicated, lengthy and bug-prone. Over time we have continued to directly add more logic and checks for various scenarios to the controller code. 
+
+Because our current approach is not scalable, we’ve been facing numerous issues in the connectivity space, especially when it comes down to identifying exactly what is causing the issue. This means the behavior of our client is extremely inconsistent both in different platforms but also within the same environment. For example if a user loses network connectivity while the VPN is activated, sometimes they enter no signal, and other times they encounter the “server location unavailable” modal. This inconsistency is a result of racing checks and error reporting. The goal of this refactor is to combine all of these scenarios under one component and to get rid of the Connection Health entirely to resolve this issue.
+
+
+## Proposed Solution
+Move all connectivity related code out of the controller and into its own separate component which is referred to as the Connectivity Manager in this document. The controller object will continue to remain but its purpose will be drastically simplified to a few basic functionalities including `activate()` and `deactivate()`. The connectivity manager will primarily act as a liaison between the controller and the UI. 
+
+```mermaid
+graph TD
+    Controller --> ConnectivityManager
+    ConnectivityManager --> Frontend
+    Frontend --> Controller
+```
+
+The exisiting Connection Health object in VPN performs checks that act in the same space as the connectivity manager. For example in connection health we periodically send pings and collect some data upon receiving (or not receiving) a response which allows us to draw conslutions regarding network connectivity and server availability. Because there is so much overlap in the nature of both classes, the creation and implementation of the Connectivity Manager component allows us to entirely get rid of connection health and move any remaining logic to connectivity manager. We should be able entirely eliminate the `ConnectionHealth` class and its corresponding states (Stable, Unstable, NoSignal). The remainder of logic which sends and receives pings and measures latency and standard deviation can be all moved to the `ConnectivityManager`.
+
+Beyond the connectivity checks, the Controller also manages how much time has elapsed since the VPN was toggled on. Having this logic in controller bloats the code without any gains, so I would like to move that code out of `Controller` and instead introduce a simple `Timer` object that encapsulates the related logic.
+
+Because this is all part of a unified connectivity experience, we do not want to release bits and pieces of this until everything is implemented, test and ready to go, naturally this means that all implementation will live behind a feature flag until we are ready to test and release.
+
+
+## Implementation Plan
+Introduce a new ConnectivityManager component that is activated when the client is launched and connects to networkWatcher in case anything changes. This new component will manage the following work:
+- Internet probe
+- Firewall
+- Captive portal check
+- Server probe
+- Subscription status
+
+We will also add a new `Timer` component to keep track of how much time has elapsed since the VPN was activated. 
+
+
+The `ConnectivityManager` has various connection states to keep track of which connection related task the VPN is performing at any given point. This means that the `ConnectivityManager` will communicate with the `Controller` so that the controller knows which state to enter at every stage. Here is the proposed set of `ConnectivityManager` states and how they map to the `Controller` states:
+
+```c++
+  enum ConnectionState {
+    ConnectionStateOff,  // Corresponds to controller states: StateInitializing,
+                         // StateOff
+    ConnectionStateInternetProbe,      // Corresponds to controller states:
+                                       // StateConnecting
+    ConnectionStateServerProbe,        // Corresponds to controller states:
+                                       // StateConnecting
+    ConnectionStateFirewall,           // Corresponds to controller states:
+                                       // StateConnecting
+    ConnectionStateCaptivePortal,      // Corresponds to controller states:
+                                       // StateConnecting
+    ConnectionStateCheckSubscription,  // Corresponds to controller states:
+                                       // StateCheckSubscription
+    ConnectionStateOn,  // Corresponds to controller state: StateOn
+
+    // Mirrored from Controller states
+    ConnectionStateSilentSwitching,  // Corresponds to StateSilentSwitching
+    ConnectionStateSwitching,        // Corresponds to StateSwitching
+    ConnectionStateConfirming,       // Corresponds to controller states:
+                                     // StateConfirming
+    ConnectionStateDisconnecting     // Corresponds to controller state:
+                                     // StateDisconnecting
+  };
+```
+
+If at any point any of the checks in `ConnectivityManager::ConnectionStates` fails then we will emit a signal to the frontend explaining what exactly the issue is. For example if we are in `ConnectivityManager::ConnectionStateInternetProbe` and the internet probe fails, we will notify the frontend via `emit connectionStateInternetProbeFailed()` signal. This allows the UI to notify the user that the VPN activation failed due to their network connectivity issue and can possibly offer troubleshooting advice such as checking that their wifi/cellular is enabled. It is worth noting that in any given circumstance, we will not deactivate the VPN on our end regardless of the failure. 
+
+This is a big undertaking that will take multiple sprints to complete. This document does not quiet cover the extent of work required for creating new unit and functional tests as well as updating the existing ones that will be affected. I propose we break down the work like below:
+
+1. Create a new `ConnectivityManager` object with relevant states. Initially this will have a 1:1 map to the controller so we can debug and ensure the component works as intended. As we add the logic to probe for various scenarios and move further along the implementation work, we can remove them from the `Controller`.
+2. Add probe internet to ConnectivityManager
+3. Add firewall probe to ConnectivityManager
+4. Add captive portal probe to ConnectivityManager
+5. Add check subscription to ConnectivityManager
+6. Add server probing to ConnectivityManager
+7. Get rid of the connection health object
+8. Audit and cleanup `Controller::statechanged`. We need to audit all areas where code is listening to Controller::statechanged and see if it should be monitoring the `ConnectivityManager` instead, if that is the case, make the necessary changes. [Here](https://searchfox.org/mozilla-vpn-client/search?q=Controller%3A%3AstateChanged&path=&case=false&regexp=false) is a current list of every instance we should audit.
+9. Move timestamp code to its own class/object
+
+
+### Audit and cleanup `Controller::statechanged`
+Below is a list of all areas where we monitor when the controller state changes. For each usage, I have added a commment proposing if it should be changed to watch `ConnectivityManager::stateChanged` instead. My rule of thumb is that if the code is that if it is UI/Frontend related, it should monitor the `ConnectivityManager`. If it is error handling related, it should continue to monitor the `Controller`. 
+
+- `Commandactivate::run()`: ConnectivityManager
+- `Commanddeactivate::run()`: ConnectivityManager
+- `CommandStatus::run()`: ConnectivityManager
+- `CommandUI::run()` -> line #390 creates and shows notification, should be watching the connectivity manager. Line #401 changes the MacOS menu, should be watching connectivity manager. 
+- `ConnectionBenchmark::initialize()`: because it is doing error handling, it should continue monitoring the controller states.
+- `IpAddressLookup::initialize()`: Not sure about this.
+- `KeyRegenerator::KeyRegenerator()`: Controller I THINK.
+- `MozillaVPN::MozillaVPN()`: #167 Not sure.  #170 notifies the UI that a refresh is needed so it should be monitoring connectivity manager. #179 this most likely will be changed since we are also refactoring connection health stuff. #183 deals with captive portal stuff and should be monitoring the Connectivity Manager instead.
+- `IOSNetworkWatcher::initialize()`: I think this will continue to monitor the controller
+- `ServerConnection::ServerConnection()`: Controller
+- `ServerLatency::initialize()`: since for now the server latency code is staying in the controller, this will continue to monitor the controller states.
+- `SettingsWatcher::SettingsWatcher()`: controller
+- `SubscriptionMonitor::SubscriptionMonitor()`:` connectivity manager since it is the component that handles subscription moving forward.
+- `SystemTrayNotificationHandler::initialize()`: updates the context menu, should be watching connectivity manager
+- `TaskControllerAction::run()`: Queues connections, so it should continue watching the controller.
+- `Telemetry::initialize()`: It relies on the current state of the vpn to stop and start timer, and should be watching the connectivity manager. 
+
+
+## Questions
+- Please review the list of Controller::stateChanged audits and let me know if you have a different opinion.
+- Do we need a `ConnectionState::Unstable` in ConnectivityManager which we would enter if missing a bunch pf ping replies?
+- Thoughts on the component name `ConnectivityManager`? Speak now or forever hold your peace.
+
+
--- a/docs/rfcs/0003-rfc-controller_refactor.md
+++ b/docs/rfcs/0003-rfc-controller_refactor.md
@@ -19,10 +19,9 @@ Move all connectivity code out of the controller and into its separate component
 
 ```mermaid
 graph TD
-    Controller -->|activate| ConnectionManager
     ConnectionManager --> Frontend
     Frontend --> ConnectionManager
-    ConnectionManager -->|deactivate| Controller
+    ConnectionManager -->|activate, deactivate| Controller
 ```
 
 The existing Connection Health object in VPN performs checks that act in the same space as the `ConnectionManager`. For example, in connection health, we periodically send pings and collect some data upon receiving (or not receiving) a response which allows us to draw conclusions regarding network connectivity and server availability. Because there is so much overlap within both classes, the creation and implementation of the Connection Manager component allows us to entirely get rid of connection health and move any remaining logic to `ConnectionManager`. We should be able to entirely eliminate the `ConnectionHealth` class as well as the states `NoSignal` and `Stable` given that they will be handled within the new Connection Manager logic. For example, currently, we may enter the No Signal state due to several reasons such as losing network connectivity, expired subscription, or server location becoming unavailable, this means that No Signal is simply a side effect of something else going wrong, and because we now have explicit messaging and checks for those, No Signal does not serve us and can be removed.
@@ -82,11 +81,11 @@ This change also means that prior to the activation of the VPN, there are multip
 This is a big undertaking that will take multiple sprints to complete. This document does not cover the extent of work required for creating new unit and functional tests as well as updating the existing ones that will be affected. I propose we break down the work like below:
 
 1. Create a new `ConnectionManager` object with relevant states. Initially, this will have a 1:1 map to the controller so we can debug and ensure the component works as intended. As we add the logic to probe for various scenarios and move further along the implementation work, we can remove them from the `Controller`.
-2. Add probe internet to `ConnectionManager`
-3. Add firewall probe to `ConnectionManager`
-4. Add captive portal probe to `ConnectionManager`
-5. Add check subscription to `ConnectionManager`
-6. Add server probing to `ConnectionManager`
+2. Add probe internet to `ConnectionManager`: This will allow us to check if there is an active internet connection.
+3. Add firewall probe to `ConnectionManager`: To make sure there are no firewall issues blocking access to the internet.
+4. Add captive portal probe to `ConnectionManager`: This will ensure that the user is not stuck behind a captive portal, which can interfere with internet access.
+5. Add check subscription to `ConnectionManager`: This will allow us to verify that the user has an active subscription before proceeding.
+6. Add server probing to `ConnectionManager`: To ensure that the servers within the user's selected location are available and running.
 7. Get rid of the `ConnectionHealth` object
 8. Audit and cleanup `Controller::stateChanged`. We need to audit all areas where code is listening to `Controller::stateChanged` and see if it should be monitoring the `ConnectionManager` instead, if that is the case, make the necessary changes. [Here](https://searchfox.org/mozilla-vpn-client/search?q=Controller%3A%3AstateChanged&path=&case=false&regexp=false) is a current list of every instance we should audit.
 9. Move timestamp code to its own class/object
--- a/docs/rfcs/0003-rfc-controller_refactor.md
+++ b/docs/rfcs/0003-rfc-controller_refactor.md
@@ -65,7 +65,8 @@ Here is the proposed set of `ConnectionManager` states:
     ConnectionStateFirewall,
     ConnectionStateCaptivePortal,
     ConnectionStateCheckSubscription,
-    ConnectionStateUnstable,           
+    ConnectionStateUnstable,     
+    ConnectionStateIdle,       // Used when there are no active probes ongoing
   };
 ```
 
@@ -75,7 +76,9 @@ If at any point any of the checks in `ConnectionManager::ConnectionStates` fails
 
 This change also means that prior to the activation of the VPN, there are multiple checks and probes that need to succeed; this will take some time (hopefully no more than a couple of seconds), and as long as we communicate to the user via front end prompts about what is happening, this is acceptable. 
 
-To confidently determine that these checks do not delay the VPN activation by an unacceptable amount, we need to measure the duration of the _time elapsed from toggling on the VPN to activation_. This is the only relevant metric, given that in any other scenario the checks run in the background and would not affect the user experience. 
+To confidently determine that these checks do not delay the VPN activation by an unacceptable amount, we need to measure the duration of the _time elapsed from toggling on the VPN to activation_. We are interested in the VPN activation duration because it delays the VPN being toggled on.
+
+In addition to collecting metrics to assess the user experience, keeping track of the duration of each operation allows us to continually monitor the system health and identify any slow downs and changes.
 
 ## Task Breakdown
 
--- a/docs/rfcs/0003-rfc-controller_refactor.md
+++ b/docs/rfcs/0003-rfc-controller_refactor.md
@@ -32,7 +32,7 @@ The existing Connection Health object in VPN performs checks that act in the sam
 
 One exception to the rule is `ConnectionHealth::ConnectionStability::Unstable`. The connection health object periodically sends pings and measures the received responses to assess whether or not the network connectivity is stable. To continue notifying users about their unstable network, we will continue to use this state within `ConnectionManager`. 
 
-Beyond the connectivity checks, the Controller also manages how much time has elapsed since the VPN was toggled on. Having this logic in the controller bloats the code without any gains, so that code will be moved out of `Controller` and instead a simple `Timer` object will encapsulates the related logic. The timer starts when the VPN enters State On since doing the initial checks may take a few seconds that we don't need to account for.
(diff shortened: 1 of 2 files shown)


docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,112 @@
...
+- `SubscriptionMonitor::SubscriptionMonitor()`: connectivity manager since it is the component that handles subscription moving forward.
+- `SystemTrayNotificationHandler::initialize()`: updates the context menu, should be watching connectivity manager
+- `TaskControllerAction::run()`: Queues connections, so it should continue watching the controller.
+- `Telemetry::initialize()`: It relies on the current state of the vpn to stop and start timer, and should be watching the connectivity manager. 
+
+
+## Questions
+- Please review the list of `Controller::stateChanged` audits and let me know if you have a different opinion.
+- Do we need a `ConnectionState::Unstable` in `ConnectivityManager` which we would enter if missing a bunch pf ping replies?
+- Thoughts on the component name `ConnectivityManager`? Speak now or forever hold your peace.
```
mcleinman, line 112: What about `ConnectionManager`?

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,112 @@
...
+- Firewall
+- Captive portal check
+- Server probe
+- Subscription status
+
+We will also add a new `Timer` component to keep track of how much time has elapsed since the VPN was activated. 
+
+
+The `ConnectivityManager` has various connection states to keep track of which connection related task the VPN is performing at any given point. This means that the `ConnectivityManager` will communicate with the `Controller` so that the controller knows which state to enter at every stage. Here is the proposed set of `ConnectivityManager` states and how they map to the `Controller` states:
+
```
mcleinman, line 46: Is it possible to make the controller *only* take connection updates from this manager? Or even completely drop the controller state? I have some concern that even if we currently write this so that only the manager updates the controller state, at some point in the future someone will write code directly updating the controller state - causing the two to get out of sync and some weird bugs.

docs/rfcs/0003-rfc-controller_refactor.md
mcleinman, line 112: What about `ConnectionManager`?

mcleinman (commented):
I like this a lot! Thanks for taking the initiative on it.

docs/rfcs/0003-rfc-controller_refactor.md
Gela: I like `ConnectionManager` better too.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Proposed Solution
+Move all connectivity related code out of the controller and into its own separate component which is referred to as the Connectivity Manager in this document. The controller object will continue to remain but its purpose will be drastically simplified to a few basic functionalities including `activate()` and `deactivate()`. The connectivity manager will primarily act as a liaison between the controller and the UI. 
+
+```mermaid
+graph TD
+    Controller --> ConnectivityManager
+    ConnectivityManager --> Frontend
+    Frontend --> Controller
```
lesleyjanenorton, line 24: Above we say the connectivity manager `sits between the frontend and controller` but here it looks like there is still some communication from the frontend to the controller. Is this intentional?

docs/rfcs/0003-rfc-controller_refactor.md
Gela: Yes, good call out. I think we can entirely get rid of Controller states so we don't need to keep 2 different state machines in sync

lesleyjanenorton (commented):
Thank you @Gela ! I will continue going through this this afternoon!

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+Here is the proposed set of `ConnectionManager` states and how they map to the `Controller` states. Please note that some of these states may go away once we are further along the implementation (e.g. we may not need `ConnectionStateSilentSwitching` anymore):
+
+```c++
+  enum ConnectionState {
+    ConnectionStateOff,  // Corresponds to controller states: StateInitializing,
+                         // StateOff
+    ConnectionStateInternetProbe,      // Corresponds to controller states:
+                                       // StateConnecting
+    ConnectionStateServerProbe,        // Corresponds to controller states:
```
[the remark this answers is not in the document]
Gela: I don't think we can group them under one state, the advantage of having individual states is that we can communicate with the front end about what exactly is preventing them from activating VPN.

Additionally, some of those checks will run periodically outside the connecting state so in the backend we need to keep track of what state we are in.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
+- Status: In review
+- Date: 2023-06-20
+- Author: [@gela](https://github.com/gela)
+- RFC PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7269
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+Introduce a new `ConnectionManager` component which sits between the frontend and the controller to handle all connectivity related logic and keep the controller code lean and minimal.
```
brizental, line 9: nit:
```suggestion
Proposes a new `ConnectionManager` component which sits between the frontend and the controller to handle all connectivity related logic and keep the controller code lean and minimal.
```

brizental: "code lean and minimal". Can we be more concrete here? What does "lean and minimal" mean?

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Motivation
+The current controller code is complicated, lengthy, and bug-prone. Over time we have continued to directly add more logic and checks for various scenarios to the controller code, resulting in a Frankenstein monster that is hard to scale and keep under control.
```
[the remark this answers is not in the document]
brizental: I agree, I'd try to steer away from abstract generalizations such as "complicated, lengthy, and bug-prone" or "a Frankenstein monster that is hard to scale and keep under control".

I think it's worh pointing out the issues with our controller code, but if we are going to mention it's bug prone maybe link to a few bugs that the complexity of the controller has generated and so on.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Proposed Solution
+Move all connectivity code out of the controller and into its separate component which is referred to as the `ConnectionManager` in this document. The controller object will continue to remain but its purpose will be drastically simplified to a few basic functionalities including `activate()` and `deactivate()`. We will no longer have any states associated with the Controller and all states will live in the `ConnectionManager`. The `ConnectionManager` will primarily act as a liaison between the controller and the UI.
```
brizental, line 18: nit:
```suggestion
Move all connectivity code out of the controller and into its separate component which is referred to as the `ConnectionManager` in this document. 

The controller object will remain, but its purpose will be drastically simplified to a few basic functionalities including `activate()` and `deactivate()`. Ther will no longer be any states associated with the Controller and all states will live in the `ConnectionManager`. The `ConnectionManager` will primarily act as a liaison between the controller and the UI. 
```

brizental: Would you mind expanding here on what are the advantages of moving state management to the ConnectionManager?

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+    ConnectionManager -->|activate, deactivate| Controller
+```
+
+The existing Connection Health object in VPN performs checks that act in the same space as the `ConnectionManager`. For example, in connection health, we periodically send pings and collect some data upon receiving (or not receiving) a response which allows us to draw conclusions regarding network connectivity and server availability. Because there is so much overlap within both classes, the creation and implementation of the Connection Manager component allows us to entirely get rid of connection health and move any remaining logic to `ConnectionManager`. We should be able to entirely eliminate the `ConnectionHealth` class as well as the states `NoSignal` and `Stable` given that they will be handled within the new Connection Manager logic. For example, currently, we may enter the No Signal state due to several reasons such as losing network connectivity, expired subscription, or server location becoming unavailable, this means that No Signal is simply a side effect of something else going wrong, and because we now have explicit messaging and checks for those, No Signal does not serve us and can be removed.
```
brizental, line 27: This paragraph discusses the functionality of the ConnectionManager, but the only introduction to it is "Move all connectivity code out of the controller and into its separate component which is referred to as the `ConnectionManager`". What about including a section where the concrete features of the ConnectionManager are discussed? Nothing in depth necessary, but just a bridge to situate the reader.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+- Internet probe
+- Firewall
+- Captive portal check
+- Server probe
+- Subscription status
+
+We will also add a new `Timer` component to keep track of how much time has elapsed since the VPN was activated. 
+
+
+The `ConnectionManager` has various connection states to keep track of which connection related task the VPN is performing at any given point. This means that the `ConnectionManager` will communicate with the `Controller` so that the controller knows which state to enter at every stage. During one of the iterations of this proposal, someone asked if we can have signals instead of states in the `ConnectionManager`. The short answer is no. While emitting a signal to notify the UI about what we should be showing the user does the same thing, we still need to keep track of the current state internally so we know what to do next in case anything goes wrong at any other point.
```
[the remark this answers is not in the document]
brizental: The note seems to directly contradict a paragraph later on
> If at any point any of the checks in `ConnectionManager::ConnectionStates` fails then we will emit a signal to the frontend explaining what exactly the issue is. For example, if we are in `ConnectionManager::ConnectionStateInternetProbe` and the internet probe fails, we will notify the frontend via `emit connectionStateInternetProbeFailed()` signal. This allows the UI to notify the user that the VPN activation failed due to their network connectivity issue and may offer troubleshooting advice such as checking that their wifi/cellular is enabled. It is worth noting that in any given circumstance, we will not deactivate the VPN on our end regardless of the failure.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+                                     // StateConfirming
+    ConnectionStateDisconnecting     // Corresponds to controller state:
+                                     // StateDisconnecting
+  };
+```
+
+If at any point any of the checks in `ConnectionManager::ConnectionStates` fails then we will emit a signal to the frontend explaining what exactly the issue is. For example, if we are in `ConnectionManager::ConnectionStateInternetProbe` and the internet probe fails, we will notify the frontend via `emit connectionStateInternetProbeFailed()` signal. This allows the UI to notify the user that the VPN activation failed due to their network connectivity issue and may offer troubleshooting advice such as checking that their wifi/cellular is enabled. It is worth noting that in any given circumstance, we will not deactivate the VPN on our end regardless of the failure. 
+
+This change also means that prior to the activation of the VPN, there are multiple checks and probes that need to succeed; this will take some time (hopefully no more than a couple of seconds), and as long as we communicate to the user via front end prompts about what is happening, this is acceptable.
```
brizental, line 79: If the hope that this doesn't take a long time is informing the proposed solution here, we should have a plan to transform that hypothesis into a fact. Let's add a section here that specs out some telemetry that will be added to monitor that this assumption is correct.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+This is a big undertaking that will take multiple sprints to complete. This document does not cover the extent of work required for creating new unit and functional tests as well as updating the existing ones that will be affected. I propose we break down the work like below:
```
brizental, line 81: I think it's completely fine not to specify unit and function tests here, so maybe we can just remove the disclaimer. Unit and functional tests are just artifacts of the implementation.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+3. Add firewall probe to `ConnectionManager`: To make sure there are no firewall issues blocking access to the internet.
+4. Add captive portal probe to `ConnectionManager`: This will ensure that the user is not stuck behind a captive portal, which can interfere with internet access.
+5. Add check subscription to `ConnectionManager`: This will allow us to verify that the user has an active subscription before proceeding.
+6. Add server probing to `ConnectionManager`: To ensure that the servers within the user's selected location are available and running.
+7. Get rid of the `ConnectionHealth` object
+8. Audit and cleanup `Controller::stateChanged`. We need to audit all areas where code is listening to `Controller::stateChanged` and see if it should be monitoring the `ConnectionManager` instead, if that is the case, make the necessary changes. [Here](https://searchfox.org/mozilla-vpn-client/search?q=Controller%3A%3AstateChanged&path=&case=false&regexp=false) is a current list of every instance we should audit.
+9. Move timestamp code to its own class/object
+
+
+### Audit and cleanup `Controller::stateChanged`
```
brizental, line 94: Auditing is not described in this section. Let's maybe move that part out into a section before this one about monitoring and validation post-release using telemetry and/or Sentry.

brizental (commented):
First pass, a some nits and a couple comments. Not necessary for this RFC, but something to keep in mind for future RFCs: how about having a line length max lint to Markdown code to force people to break lines in paragraphs. When whole paragraphs are just one line it's hard to make specific comments and not overlap reviews.

docs/rfcs/0003-rfc-controller_refactor.md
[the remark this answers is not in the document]
Gela: I don't think they are contradicting each other. The note above is talking about only having signals to notify the front end and getting rid of ConnectionManager states, whereas the paragraph below it is saying if we are in a certain state and that probe fails then it will emit a signal to the UI. Does this clarify that?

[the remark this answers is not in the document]
brizental: Aaah. Ok I just needed read it another time. Ignore that one :)

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Implementation Plan
+Introduce a new `ConnectionManager` component that is activated when the client is launched and connects to `networkWatcher` in case anything changes. This new component will manage the following work:
+- Internet probe
+- Firewall
+- Captive portal check
+- Server probe
+- Subscription status
+
+We will also add a new `Timer` component to keep track of how much time has elapsed since the VPN was activated.
```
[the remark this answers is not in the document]
Gela: Right now the controller does the work of a timer, I agree that bit could have its own small RFC but I feel like I need to at the very least acknowledge that it's there in the controller. The RFC explains that the controller should only handle "activate" and "deactivate" so it makes sense to briefly mention that the timer will eventually be moved too.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,99 @@
...
+
+One exception to the rule is `ConnectionHealth::ConnectionStability::Unstable`. The connection health object periodically sends pings and measures the received responses to assess whether or not the network connectivity is stable. To continue notifying users about their unstable network, we will continue to use this state within `ConnectionManager`. 
+
+Beyond the connectivity checks, the Controller also manages how much time has elapsed since the VPN was toggled on. Having this logic in the controller bloats the code without any gains, so I would like to move that code out of `Controller` and instead introduce a simple `Timer` object that encapsulates the related logic. The timer starts when the VPN enters State On since doing the initial checks may take a few seconds that we don't need to account for.
```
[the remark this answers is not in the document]
Gela: I think when/if we get to the Timer work, we would need a separate RFC for that which would explore how we'd approach that work

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,115 @@
...
+- Captive portal check: determine uf the network is behind a captive portal.
+- Server probe: upon activation and then priodically check if there are available servers within user's selection location.
+- Subscription status: ensure that user has an active subscription, and if not, show the "No Subscription Found" screen.
+
+The logic to calculate elapsed time since the VPN was activated will also be moved out of the Controller and into a new `Timer` component. This bit of work will likely require a separate RFC. 
+
+
+The `ConnectionManager` will have two state machines:
+1. `VPNClientState` to determine if the VPN client is off, connecting, or on. 
+2. `ConnectionState` to keep track of which connection related task the VPN is performing at any given point.
```
mcleinman, line 53: Does this only correspond to active connection/disconnection states? If `VPNClientState` is on or off, what is the `ConnectionState`?

Gela: Good catch, updated the ConnectionState values

mcleinman: Thanks! And so `VPNClientState` is derived from `ConnectionState` right? It shouldn't be a variable that can be set, right? Don't know what it would be in C++ land, but in Swift I'd say we should be a computed property. Could be a function if needed.

Overall, having two state vars that need to stay in sync - with no guarantee that future maintainers will always remember that - seems like opening ourselves up to future bugs.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: With the latest edits I think this is much clearer. We can resolve this one.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+
+One exception to the rule is `ConnectionHealth::ConnectionStability::Unstable`. The connection health object periodically sends pings and measures the received responses to assess whether or not the network connectivity is stable. To continue notifying users about their unstable network, we will continue to use this state within `ConnectionManager`. 
+
+Beyond the connectivity checks, the Controller also manages how much time has elapsed since the VPN was toggled on. Having this logic in the controller bloats the code without any gains, so that code will be moved out of `Controller` and instead a simple `Timer` object will encapsulates the related logic. The timer starts when the VPN enters State On since doing the initial checks may take a few seconds that we don't need to account for.
```
brizental, line 35:
```suggestion
Beyond the connectivity checks, the Controller also manages how much time has elapsed since the VPN was toggled on. Having this logic in the controller bloats the code without any gains, so that code will be moved out of `Controller` and instead a simple `Timer` object will encapsulate the related logic. The timer starts when the VPN enters State On since doing the initial checks may take a few seconds that we don't need to account for.
```

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Implementation Plan
+Introduce a new `ConnectionManager` component that is activated when the client is launched and connects to `networkWatcher` in case anything changes. This new component will manage the following work:
+- Internet probe: assess if there is a network connection.
+- Firewall: assess if firewall is limiting network connectivity.
+- Captive portal check: determine uf the network is behind a captive portal.
+- Server probe: upon activation and then priodically check if there are available servers within user's selection location.
+- Subscription status: ensure that user has an active subscription, and if not, show the "No Subscription Found" screen.
+
+The logic to calculate elapsed time since the VPN was activated will also be moved out of the Controller and into a new `Timer` component. This bit of work will likely require a separate RFC.
```
brizental, line 48:
```suggestion
The logic to calculate elapsed time since the VPN was activated will also be moved out of the Controller and into a new `Timer` component. This bit of work is out of scope for this RFC. 
```

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+
+Before the VPN is toggled on, it is in `VPNClientStateOff`. When user toggles on the VPN, the client goes into `VPNClientStateConnecting`, the connection manager will perform the needed connection checks such as `ConnectionStateInternetProbe` and `ConnectionStateServerProbe`. If they all succeed, the VPN will enter `VPNClientStateOn` and call `Controller::activate()`.
+
+The `ConnectionManager` will communicate to the `Controller` to `activate()` and `deactivate()` the VPN. During one of the iterations of this proposal, someone asked if we can have signals instead of states in the `ConnectionManager`. The short answer is no. While emitting a signal to notify the UI about what we should be showing the user does the same thing, we still need to keep track of the current state internally so we know what to do next in case anything goes wrong at any other point.
+
+Here is the proposed set of `ConnectionManager` states and how they map to the `Controller` states. Please note that some of these states may go away once we are further along the implementation (e.g. we may not need `ConnectionStateSilentSwitching` anymore):
```
brizental, line 59:
```suggestion
Here is the proposed set of `ConnectionManager` states and how they map to the `Controller` states:
```

I would remove this sentence. Since that is not happening now and there is not much context on why that would go away it doesn't add much and makes things confusing.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+    ConnectionStateFirewall,           // Corresponds to client state:
+                                       // VPNClientStateConnecting
+    ConnectionStateCaptivePortal,      // Corresponds to client state:
+                                       // VPNClientStateConnecting
+    ConnectionStateCheckSubscription,  // Corresponds to client state:
+                                       // VPNClientStateConnecting
+    ConnectionStateOn,                 // Corresponds to client state:
+                                       // VPNClientStateOn
+  };
+```
```
brizental, line 85: Above there is some mention of keeping the Unstable state, but it's not listed here.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+The `ConnectionManager` will have two state machines:
+1. `VPNClientState` to determine if the VPN client is off, connecting, or on. 
+2. `ConnectionState` to keep track of which connection related task the VPN is performing at any given point. 
+
+Before the VPN is toggled on, it is in `VPNClientStateOff`. When user toggles on the VPN, the client goes into `VPNClientStateConnecting`, the connection manager will perform the needed connection checks such as `ConnectionStateInternetProbe` and `ConnectionStateServerProbe`. If they all succeed, the VPN will enter `VPNClientStateOn` and call `Controller::activate()`.
+
+The `ConnectionManager` will communicate to the `Controller` to `activate()` and `deactivate()` the VPN. During one of the iterations of this proposal, someone asked if we can have signals instead of states in the `ConnectionManager`. The short answer is no. While emitting a signal to notify the UI about what we should be showing the user does the same thing, we still need to keep track of the current state internally so we know what to do next in case anything goes wrong at any other point.
```
brizental, line 57: Why is it that we need to keep track of the state internally? How does it change what we are going to do next? That is unclear to me on the `ConnectionState` specifically.

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Metrics and Measurements
+
+This change also means that prior to the activation of the VPN, there are multiple checks and probes that need to succeed; this will take some time (hopefully no more than a couple of seconds), and as long as we communicate to the user via front end prompts about what is happening, this is acceptable. 
+
+To confidently determine that these checks do not delay the VPN activation by an unacceptable amount, we need to measure the duration of the _time elapsed from toggling on the VPN to activation_. This is the only relevant metric, given that in any other scenario the checks run in the background and would not affect the user experience.
```
brizental, line 93: Its worth noting that telemetry is not only useful for measuring metrics related to the user experience. It can be very useful for technical data as well to measure the health of this system. I am not saying we need telemetry for that here, but we should at least reword this sentence and give it a second thought if there are metrics we can add to this new system that would help us asses it's health after release.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: Above you also state that ConnectionState exists to "keep track of which connection related task the VPN is performing at any given point". "Off" and "On" are not really "tasks", as I understand it. Why are they part of this state?

docs/rfcs/0003-rfc-controller_refactor.md
```diff
@@ -0,0 +1,111 @@
...
+- Captive portal check: determine uf the network is behind a captive portal.
+- Server probe: upon activation and then priodically check if there are available servers within user's selection location.
+- Subscription status: ensure that user has an active subscription, and if not, show the "No Subscription Found" screen.
+
+The logic to calculate elapsed time since the VPN was activated will also be moved out of the Controller and into a new `Timer` component. This bit of work will likely require a separate RFC. 
+
+
+The `ConnectionManager` will have two state machines:
+1. `VPNClientState` to determine if the VPN client is off, connecting, or on. 
+2. `ConnectionState` to keep track of which connection related task the VPN is performing at any given point.
```
brizental, line 53: Bike shedding alert! `ConnectionState` doesn't really communicate "enumeration of connection related task the VPN is performing at any given point". What about `ConnectionManagerTask` / `ConnectionTask`?

docs/rfcs/0003-rfc-controller_refactor.md
brizental: I made a comment about this below, but I see this discussion is related. I'd say it's confusing to have On and Off be part of the ConnectionState because they are outside of the scope of that enum as it's described here. How about, when there are no connection tasks ongoing you either make that state a `nullptr` or introduce a `ConnectionState::Idle` option?

mcleinman: I don't feel strongly on whether we include an On/Off, but I am concerned about having two states that are closely connected, yet it's technically possible to modify just one.

docs/rfcs/0003-rfc-controller_refactor.md
Gela: > Above you also state that ConnectionState exists to "keep track of which connection related task the VPN is performing at any given point". "Off" and "On" are not really "tasks", as I understand it. Why are they part of this state?

I have those connections in `ConnectionState `to map to when `VPNClientState` is either on of off because otherwise when the `VPNClientState` is On or Off we weren't very clean on what state `ConnectionState` would be in

docs/rfcs/0003-rfc-controller_refactor.md
Gela: >To confidently determine that these checks do not delay the VPN activation by an unacceptable amount, we need to measure the duration

That's what this bit is referring to, what else in the context of system health would you wan to see in this section?

docs/rfcs/0003-rfc-controller_refactor.md
Gela: > when there are no connection tasks ongoing you either make that state a nullptr or introduce a ConnectionState::Idle option?

@brizental I don't think this will work given that a VPN could be toggled on (`ConnectionStateIdle`) and be in a secondary state such as `InternetProbe` or `ServerProbe`. I'm starting to think that it would make more sense for the VPN on/off to simply be a bool value when on is true and off is false.

docs/rfcs/0003-rfc-controller_refactor.md
Gela: These are asynchronous tasks that may trigger both upon activation && while the VPN is already active. For instance there is no point in probing servers or checking subscription if network probe is failing but we need a mechanism to keep track of it.

docs/rfcs/0003-rfc-controller_refactor.md
mcleinman: Are you suggesting that the VPNClientState would be a boolean? Or is that a different variable?

Gela: @mcleinman That would be a different variable (bool) in the controller, set to true when the VPN is toggled on, and false otherwise.

brizental: > @brizental I don't think this will work given that a VPN could be toggled on (ConnectionStateIdle) and be in a secondary state such as InternetProbe or ServerProbe.

Hm, but in this case it would not be idle, it would be performing internet/server probe task.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: > This is the only relevant metric, given that in any other scenario the checks run in the background and would not affect the user experience.

The proposed telemetry should help us understand system health. But this sentence is implying that unless something affects user experience we do not need to monitor it with telemetry and this is not correct. I am calling out that if the telemetry design was thought out with this in mind it should be revisited without this assumption. I did not have any extra telemetry in mind when I said it, it was more of a conceptual comment.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: Ok, but where do we store the state about whether network probe failed or not? The description of `ConnectionState` says

>  [...] `ConnectionState` to keep track of which connection related task the VPN is performing at any given point via the corresponding task manager `ConnectionTaskManager`. 

So this is just stating which task is being performed at each moment, not the result of the task itself. Is that incorrect?

docs/rfcs/0003-rfc-controller_refactor.md
Gela: @mcleinman No, I'm proposing we'd have one state machine `ConnectionState` in Connection Manager and one bool in the Controller to keep track of whether or not the VPN is on.

@brizental I see. I think it makes sense to have an Idle state for when we aren't running any checks in Connection Manager, and a bool to keep track of VPN on/off as explained above. Does that sound good?

docs/rfcs/0003-rfc-controller_refactor.md
Gela: @brizental There would be corresponding bool state variables which would be set to false if a certain probe fails, and we would then emit a signal to the front end stating the failure.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: I think it sounds good yes.

docs/rfcs/0003-rfc-controller_refactor.md
brizental: I see. In that case, it remains unclear why exactly the ConnectionState is necessary. We can work it out during implementation, you have thought about this longer than I have and it's likely just hard for me to visualize it.

brizental (approved):

docs/rfcs/0003-rfc-controller_refactor.md
mcleinman: Sounds good, thanks!

mcleinman (approved):
This is great, thank you!

PR #7294, opened 2023-06-22
VPN-5153 - Put replacer type addon behind feature flag

brizental:


src/shared/addons/addon.cpp
```diff
@@ -374,6 +374,14 @@ Addon* Addon::create(QObject* parent, const QString& manifestFileName) {
   }
 
   else if (type == "replacer") {
+    if (!Feature::get(Feature::Feature_replacerAddon)->isSupported()) {
+      // This addon type is blocked pending
+      // https://mozilla-hub.atlassian.net/browse/VPN-4858
+      logger.error() << "The replacer type addon is unsupported.";
+      Q_ASSERT(false);
```
bakulf, line 381: This is wrong. The client will crash for all the users with this version.

brizental: What's wrong about it?

strseb (approved):

PR #7304, opened 2023-06-22
Add a section to download VS for Windows dev setup

Gela:
## Description

Added a section to mention we need to download VS 2017 or later to get this to work. I didn't realize this was missing from the steps until I completely wiped my machine back to factory settings and noticed this was required.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/docs/dev-setup/windows.md
+++ b/docs/dev-setup/windows.md
@@ -8,6 +8,13 @@ https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozil
 
 Unzip the folder and remember the location for the configure step.
 
+## Get Visual Studio 2017 or Later
+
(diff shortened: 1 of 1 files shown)


strseb:
Can you specify, we are not using anything from vs2017.

strseb:
Ohno. We should not be using link.exe - that indicated that there is something wrong with the env. Because we certainly set the linker to be used. :/ 
```
conda env config vars set LD="lld-link"| Out-Null
...
conda env config vars set CMAKE_LINKER="lld-link" | Out-Null
```
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/scripts/windows/conda_setup_win_sdk.ps1#L53

What is the output of `get-ChildItem env:` inside your conda env?

PR #7336, opened 2023-06-26
[Chore] - Remove old msvc build target

strseb:
## Description

    We switched to using a clang-cl toolchain to build vpn. 
    The msvc task was only left as backup - in case something breaks. 
    It QA so far has not found anything related, to that - so for the next version we can remove it
    as it also is currently broken.


brizental (approved):

taskcluster/ci/build/windows.yml
```diff
@@ -40,22 +40,6 @@ task-defaults:
...
-        artifacts:
-            - type: file
-              name: public/build/Mozilla VPN.pdb
-              path: artifacts/Mozilla VPN.pdb
-    scopes:
-        by-level:
-            "3":
-                - secrets:get:project/mozillavpn/tokens
-            default: []
-
```
hneiva, line 58: Nit/Not a problem, just a cleanup: Since there's only 1 task type left in this kind, it would be good to merge `task-defaults` into`windows/opt`. Unless if we plan on adding another task?

PR #7341, opened 2023-06-26
Updates for Global Expansion dot release (v2.15.1)

lesleyjanenorton:
## Description
This PR 
- Bumps the client version number to 2.15.1
- Updates stage URL
- Updates languages.json
- Updates servers.json
- Bumps i18n pointer
- Cherry-picks #6465 because all wasm tests fail without it. cc @bakulf is this risky? 
- Does some one-off maneuvering in generate_ts.sh to avoid failing when looping to 2.16 release branch

Still routinely and frustratingly failing `build-windows/opt`, `build-ios/debug`, and `build-macos/universal`.
## Reference

VPN-5126

## Checklist
    

.github/workflows/translations.yaml
```diff
@@ -3,12 +3,10 @@ name: Translations
 on:
   push:
     branches:
-      - main
-      - 'releases/**'
+      - 'releases/2.15.*'
```
mcleinman, line 6: Why remove `main` here (and below)? Could we keep it all - make this run on `releases/**`, `releases/2.15.*`, and `main`?

scripts/utils/generate_ts.sh
```diff
@@ -58,7 +58,7 @@ mkdir -p addon_ts || die
 cp addons/generated/addons/*.ts addon_ts
 print G "done."
 
-for branch in $(git branch -r | grep origin/releases); do
+for branch in $(git branch -r | grep origin/releases/2.15.*); do
```
mcleinman, line 61: Do we need to add this in? Without it, we just get a longer list of release branches - including 2.15.*.

mcleinman (commented):
Couple questions, but this looks great. Thanks!

.github/workflows/translations.yaml
lesleyjanenorton: oops sorry, just seeing this comment otherwise I wouldn't have force pushed but this was find/replace pasta and is now reverted

scripts/utils/generate_ts.sh
lesleyjanenorton: It fails once we get to releases/2.16 when `strings.yaml` isn't found at `translations/strings.yaml` and (afaik) we don't need to import strings from that branch.

i18n
flodolo, line 1: Isn't this updated to the latest version by the build process?

scripts/utils/generate_ts.sh
flodolo: Just to confirm, this is a one-off problem, because of #6401? It's not a problem going forward to build from branches?

flodolo (approved):

scripts/utils/generate_ts.sh
lesleyjanenorton: That's correct. We have patched `generate_ts.sh` in 2.16 to handle the differences between branches.

PR #7267, opened 2023-06-20
conda-pack windows env

strseb:
## Description

We're currently fetching the whole conda-build env from the internet on every build. 
For android, we have moved that all into a toolchain task, allowing us to cache that, making stuff more stable :) 

Let's do that for the others too.

PR #7347, opened 2023-06-27
[CI] WIP fix 2.15.1

strseb:
mwefxcx3tux31

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ pathspec
 pyrsistent
 PyYAML
 yamllint
+conda-package-handling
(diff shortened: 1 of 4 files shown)


lesleyjanenorton:
duct-tape-this-is-the-end

strseb:
ELWH1HOUEAINo2R
@lesleyjanenorton  windows is green at least. 


this is done using the artifact i'm producing here #7267.
I'm zipping the conda-env and make this task use the rust installation from there instead of downloading this during the task run.

lesleyjanenorton (approved):
Yes! Let's let at least the queued up tasks run and then WFM.

Issue #6478, opened 2023-03-27
Handle an empty Staging Server URL gracefully (aka don't crash)


data-sync-user:
Right now in staging mode if I check “Use Staging Server” and leave the URL empty and close the VPN, the application crashes in the next startup due to the empty field QASSERT’ing. The only way to know what went wrong is to either read the logs or check the console which can be a little confusing. 

The only workaround is to delete the log file `vpn.moz` which will reset the URL to default. 

**Expectation:**

We should handle this gracefully. There are a few options:

1. If the url is blank on initialization, default back to `https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net`
1. **Do not allow the user to check “Use Staging Server” without providing a URL in the app (accepted proposal)**
1. Another alternative?

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-4456)
┆Reporter: Gela Malek Pour

data-sync-user:
➤ Gela Malek Pour commented:

In the same vein, if user sets up their local Guardian instance and change the URL to point to that instead (localhost:3000 for example) and then later stops the local instance of Guardian, the VPN will keep crashing (qFatal) due to “Non localhost reuqest detected” in networkmanager.cpp 

Again, this can be resolved by deleting vpn.moz

data-sync-user:
➤ Beatriz Rizental Machado commented:

{quote}In the same vein, if user sets up their local Guardian instance and change the URL to point to that instead (localhost:3000 for example) and then later stops the local instance of Guardian, the VPN will keep crashing (qFatal) due to “Non localhost reuqest detected” in networkmanager.cpp{quote}

This is related, but a seaprate issue. Would you mind filing a separate ticket Gela Malek Pour ?

data-sync-user:
➤ Vinoo Cherian commented:

Option 2 isn’t feasible because the user could have checked “Use Staging Server” before deleting the URL. And disabling the checkbox when the URL field is empty makes it difficult for a user to delete a URL and then replace it with a new one. 

Fixed as follows:
If the user did not specify a server address, change the empty address to the default server address on startup. 

Other solution that was considered:
Use Component.onDestruction in QML to uncheck staging server if an address was not specified. However onDestruction is not called when the window is closed. This may be a Qt issue.

data-sync-user:
➤ Santiago Andrigo commented:

Gela Malek Pour Vinoo Cherian 

[...]

PR #7362, opened 2023-06-29
VPN-4456: Don't use staging server if a server address is not specified

vinocher:
## Description

 If the user did not specify a server address, change the empty address to the default server address on startup. Otherwise the startup code path will assert that the server address is non-null, and the assert will crash the app in a debug build.

The assert is in AppConstants::setStaging:
Q_ASSERT(!s_stagingServerAddress.isEmpty());

Other solution that was considered:
Use Component.onDestruction in QML to uncheck staging server if an address was not specified. However onDestruction is not called when the window is closed. This may be a Qt issue.

## Reference

GitHub issue [6478](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/6478), [VPN-4456](https://mozilla-hub.atlassian.net/browse/VPN-4456)

## Checklist
    
[VPN-4456]: https://mozilla-hub.atlassian.net/browse/VPN-4456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/vpn/appconstants.cpp
+++ b/src/apps/vpn/appconstants.cpp
@@ -21,10 +21,10 @@ void AppConstants::setStaging() {
   Constants::setStaging();
 
   // Get staging server address. If it was set to empty by the user, remove it
-  // from SettingsHolder, and then query for it again to get the default address.
+  // from SettingsHolder, and then query for it again to get the default
+  // address.
   s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
-  if (s_stagingServerAddress.isEmpty())
-  {
+  if (s_stagingServerAddress.isEmpty()) {
     SettingsHolder::instance()->removeStagingServerAddress();
     s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
   }
--- a/src/apps/vpn/appconstants.cpp
+++ b/src/apps/vpn/appconstants.cpp
@@ -19,7 +19,15 @@ const QString& AppConstants::getStagingServerAddress() {
 
 void AppConstants::setStaging() {
   Constants::setStaging();
+
+  // Get staging server address. If it was set to empty by the user, remove it
+  // from SettingsHolder, and then query for it again to get the default address.
   s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
+  if (s_stagingServerAddress.isEmpty())
+  {
+    SettingsHolder::instance()->removeStagingServerAddress();
+    s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
+  }
   Q_ASSERT(!s_stagingServerAddress.isEmpty());
 }
 
--- a/src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
+++ b/src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
@@ -64,17 +64,19 @@ MZViewBase {
                 if (MZSettings.stagingServerAddress !== serverAddressInput.text) {
                     MZSettings.stagingServerAddress = serverAddressInput.text;
                 }
+
+                // If the server address has been deleted, uncheck checkBoxRowStagingServer to not use a
+                // staging server. Also reset the server address to the default, so the user can easily switch
+                // back to it by rechecking the checkbox.
+                if (serverAddressInput.text.length == 0) {
+                    MZSettings.stagingServer = false;
+                    serverAddressInput.text = "https://stage.guardian.nonprod.cloudops.mozgcp.net";
+                    MZSettings.stagingServerAddress = serverAddressInput.text;
+                }
             }
 
             Component.onCompleted: {
                 serverAddressInput.text = MZSettings.stagingServerAddress;
-            }
-
-            Component.onDestruction: {
-                // If a server address is not specified, don't use a staging server
-                if (serverAddressInput.text.length == 0) {
-                    MZSettings.stagingServer = false;
-                }
             }
         }
 
--- a/src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
+++ b/src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
@@ -64,15 +64,6 @@ MZViewBase {
                 if (MZSettings.stagingServerAddress !== serverAddressInput.text) {
                     MZSettings.stagingServerAddress = serverAddressInput.text;
                 }
-
-                // If the server address has been deleted, uncheck checkBoxRowStagingServer to not use a
-                // staging server. Also reset the server address to the default, so the user can easily switch
-                // back to it by rechecking the checkbox.
-                if (serverAddressInput.text.length == 0) {
-                    MZSettings.stagingServer = false;
-                    serverAddressInput.text = "https://stage.guardian.nonprod.cloudops.mozgcp.net";
-                    MZSettings.stagingServerAddress = serverAddressInput.text;
-                }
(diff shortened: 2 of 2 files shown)


src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -64,18 +64,20 @@ MZViewBase {
...
-            }
 
-            Component.onDestruction: {
-                // If a server address is not specified, don't use a staging server
+                // If the server address has been deleted, uncheck checkBoxRowStagingServer to not use a
+                // staging server. Also reset the server address to the default, so the user can easily switch
+                // back to it by rechecking the checkbox.
                 if (serverAddressInput.text.length == 0) {
                     MZSettings.stagingServer = false;
+                    serverAddressInput.text = "https://stage.guardian.nonprod.cloudops.mozgcp.net";
```
vinocher, line 73: Resetting to the default URL this way is a bit of a hack, because the string is being duplicated. Unfortunately, the MZSettings.StagingServerAddress property is not currently resettable, and supporting RESET in the Q_PROPERTY would be a larger change. Let me know if think this hack is not acceptable.

Gela: I think this is acceptable, this is primarily used during development only so we we shouldn't overcomplicate it.

Gela (approved):

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
MattLichtenstein: TIL about `RESET`! Agree, I don't think it's worth making changes to the settings macro for this one dev menu feature. Since our app constants aren't exposed to QML, do you think there are any good spots to add a CONST Q_PROPERTY somewhere to expose the default staging url?

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -64,18 +64,20 @@ MZViewBase {
...
-                serverAddressInput.text = MZSettings.stagingServerAddress;
-            }
 
-            Component.onDestruction: {
-                // If a server address is not specified, don't use a staging server
+                // If the server address has been deleted, uncheck checkBoxRowStagingServer to not use a
+                // staging server. Also reset the server address to the default, so the user can easily switch
+                // back to it by rechecking the checkbox.
                 if (serverAddressInput.text.length == 0) {
                     MZSettings.stagingServer = false;
```
MattLichtenstein, line 72: Again, since this is just the dev menu I think this is fine - but I'm not a huge fan of the pattern of disabling the text field right when the text length is 0. It's rare, but if I was looking to change this url, I would most likely start by double-clicking the current url to highlight it and hit backspace, which, in this case, would result in the field being disabled/reset.

Any issue with just doing this logic on destruction?

MattLichtenstein: Ahhh, I see where this would cause a crash when quitting with a blank staging URL. Is there any opportunity to mitigate within the C++ code on startup to revert an empty URL to the default staging address?

vinocher: >  I would most likely start by double-clicking the current url to highlight it and hit backspace, which, in this case, would result in the field being disabled/reset.

Good observation!

vinocher: The fix has been changed to handle this. When staging is set on app startup, an empty address is changed to the default address.

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
vinocher: Changed the solution to not require the default staging URL in QML.

Gela (approved):
I like the new flow much much better, thank you

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -69,13 +69,6 @@ MZViewBase {
             Component.onCompleted: {
                 serverAddressInput.text = MZSettings.stagingServerAddress;
             }
-
-            Component.onDestruction: {
-                // If a server address is not specified, don't use a staging server
-                if (serverAddressInput.text.length == 0) {
-                    MZSettings.stagingServer = false;
-                }
-            }
```
MattLichtenstein, line 78: Any reason not to keep the destruction code. It's a rare use case (since changing the server requires an app restart), but I don't see any harm in resetting it on destruction.

The only question I have is, if we destruct while `stagingServer === true` and `serverAddressInput.text === 0`, should we set them to production, or keep them on staging and just re-add default address

src/apps/vpn/appconstants.cpp
```diff
@@ -19,7 +19,15 @@ const QString& AppConstants::getStagingServerAddress() {
...
   Constants::setStaging();
+
+  // Get staging server address. If it was set to empty by the user, remove it
+  // from SettingsHolder, and then query for it again to get the default
+  // address.
   s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
+  if (s_stagingServerAddress.isEmpty()) {
+    SettingsHolder::instance()->removeStagingServerAddress();
+    s_stagingServerAddress = SettingsHolder::instance()->stagingServerAddress();
+  }
```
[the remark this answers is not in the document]
MattLichtenstein: That way, if we are on production and want to switch back to staging, the address will already be there and we don't have to hunt for it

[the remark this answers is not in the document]
vinocher: There are multiple calls to AppConstants::setStaging() in the code base, so I'd prefer not to make changes in all those call sites. I thought it is not possible for a user to change the server address when not staging?

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
vinocher: The destruction code runs when the view changes, but not when the window closes. (This may be a Qt problem .) Given, this inconsistency I think it may be better to handle it on app restart in AppConstants::setStaging().

src/apps/vpn/appconstants.cpp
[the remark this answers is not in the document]
MattLichtenstein: Ahhh, I see now - since we have to restart the app when we go from production -> staging anyways, there is no need to restore the URL when we are not on staging (eg deleting the address, then toggling staging off). This looks good to me 👍

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
MattLichtenstein: Same as above, since we have to restart the app anyways, no need to really do anything on destruction.

The only slightly confusing thing here is that if the URL in the textfield is blank, and staging is turned on, theres really no way for the user to know that we will revert back to the default staging URL on restart, but I don't think thats a huge deal.

MattLichtenstein (approved):
Looks good, nice job 🔥

PR #7367, opened 2023-06-29
VPN-5140,VPN-5193 - Remove the apps/ folder

brizental:
This is part 1 of VPN-5141. Part 2 will be about flattening the shared/ folder under src/.

~Opening this as a draft, just to see what breaks.~ Ready for review, the one failing job `glean probe-scraper` cannot be fixed before merging this. See comment below.


brizental:
The probe-scraper failure is something we need to fix over on the probe-scraper repository after we merge this in. PR is up, blocked on this one https://github.com/mozilla/probe-scraper/pull/606

flodolo:
PSA: I won't be around next week, and this is likely going to break string export (I remember breaking stuff when we moved the other way around).

brizental:
> PSA: I won't be around next week, and this is likely going to break string export (I remember breaking stuff when we moved the other way around).

I see https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/workflows/update.yaml needs to be updated for sure. I'll send a PR with that fix tomorrow to l10n. I think it's fine to defer merging this until you are back, if you prefer @flod.

flodolo:
(`@flod`  is another person on GitHub ;-) )

Yep, merging when I'm around to debug is probably for the best. We can also test automation through branches to see if it actually works before merging.

lesleyjanenorton:
Is this a new record? 
<table><tr><td>Screen Shot 2023-07-05 at 2 30 09 PM</td></tr></table>

scripts/utils/generate_strings.py
```diff
@@ -58,7 +58,7 @@ def parseTranslationStrings(project, yamlfile):
 
         for category in yaml_content:
             for key in yaml_content[category]:
-                string_id = f"{project}.{category}.{key}"
```
flodolo, line 61: This is going to break every single translation we have, and duplicate content until we have branches that use two different naming conventions.

Check the string IDs in the XLIFF file
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/en/mozillavpn.xliff#L827

The string IDs should NOT change as part of this, they didn't change as part of the previous move.

flodolo (changes requested):
I suggest to compare the outout of the translations workflow with main, there should be no differences in the generated files (attached to the workflow).

brizental:
> Is this a new record?

It looks scary, but it's mostly moving things around 🙏 🆘

scripts/utils/generate_strings.py
brizental: Brought the `vpn.` prefix back.

brizental:
> I suggest to compare the outout of the translations workflow with main, there should be no differences in the generated files (attached to the workflow).

I have followed this advice and with the current state of this branch, the generated files look _exactly_ the same.

flodolo:
This seems to be removing a lot of strings (check the diff on the `en` folder)?
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/371/

brizental:
> This seems to be removing a lot of strings (check the diff on the en folder)?

Huh, weird. When I diff'd the stringl18n_p.cpp file it was exactly the same. Thanks for double checking... I'll figure out what's up.

flodolo:
If you're not familiar with the XLIFF file, it looks like the strings from strings.yaml are fine, what is getting lost are strings using the old format (declared in the qml files).

brizental:
> If you're not familiar with the XLIFF file, it looks like the strings from strings.yaml are fine, what is getting lost are strings using the old format (declared in the qml files).

I am in fact not familiar with it. Thanks for the patience walking me through this @flodolo! I'll check why these strings are missing.

flodolo:
Getting closer: fewer strings missing, but now I have a bunch of **FooBar** 😨 (apparently extracted from `../src/apps/template/extrastrings.h`, previously they were `../src/apps/vpn/extrastrings.h`).

Also a few test strings from `../src/apps/unit_tests/testlocalizer.cpp`.

brizental:
> Getting closer: [fewer strings missing](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/372/files#diff-23c0d0ff5a7678a6d691121f6e2a4298bd8bbabb5e79f9b5049c14a880cd38e7), but now I have a bunch of FooBar 😨 (apparently extracted from ../src/apps/template/extrastrings.h, previously they were ../src/apps/vpn/extrastrings.h).

> Also a few test strings from ../src/apps/unit_tests/testlocalizer.cpp.

Ah, thanks for running the job. Turns out me running `python3 ./.github/scripts/extract_source_strings.py --input ../mozilla-vpn-client/translations.ts --output mozillavpn.xliff` locally for testing purposes gave me a [different generated xliff file](https://gist.github.com/brizental/f6c76c0d31ddee44218fc4a8693a692b). Clearly I am doing something wrong there. Anyways, indeed almost there 🤔

flodolo:
> Turns out me running `python3 ./.github/scripts/extract_source_strings.py --input ../mozilla-vpn-client/translations.ts --output mozillavpn.xliff` locally for testing purposes gave me a [different generated xliff file](https://gist.github.com/brizental/f6c76c0d31ddee44218fc4a8693a692b). Clearly I am doing something wrong there.

That should work. I wonder if the problem is local, untracked .ts files that are slightly different?

The [workflow](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/workflows/update.yaml#L54) in question only runs `generate_ts.sh`, then the command you already pointed out (first for `translations.ts`, then for addons).

brizental:
> That should work. I wonder if the problem is local, untracked .ts files that are slightly different?

The generate_ts assumes the remote will be named `origin` and mine is actually `upstream`. `origin` points to my fork which doesn't have the same release branches. I am able to repro the issues locally. This is tricky, because we need to be backwards compatible with previous structures of the repo. That is why we are getting the unit tests strings in there. Anyways, I have a good grasp of the issue now. Let's see what can be done.

flodolo:
Not sure if updating here is helpful, but with the last update:
* We lost the `vpn.` prefix in strings generate from `strings.yaml`
* The path to file changed, e.g. from `../src/apps/vpn/ui/screens/ScreenBackendFailure.qml` to `../src/ui/screens/ScreenBackendFailure.qml`. If needed, this is something I can work on in the [extraction script](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/scripts/extract_source_strings.py#L63-L74).

brizental:
> We lost the vpn. prefix in strings generated from strings.yaml

Yeah, this is likely just me messing with the git history.

> The path to file changed, e.g. from ../src/apps/vpn/ui/screens/ScreenBackendFailure.qml to ../src/ui/screens/ScreenBackendFailure.qml. If needed, this is something I can work on in the [extraction script](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/blob/main/.github/scripts/extract_source_strings.py#L63-L74).

Yeah, so this PR is about exactly changing the paths. Is it an issue that the paths have changed?

brizental:
> Not sure if updating here is helpful, but with the [last update](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/compare/5d4d7a3cd3039f5c14440d644aebfaf6d7652964..1a29ceb9544bee0b89e1256d94623b055c87b736):

It's been super helpful actually, thank you so much for watching and retrying so quickly. Can we have a new retry... I must have removed my `vpn.` prefix change during some rebase and now it's back. I see the linters have passed, so that is a good sign.

flodolo:
> Yeah, so this PR is about exactly changing the paths. Is it an issue that the paths have changed?

It is an issue, because all translations for those `file`  elements in the XLIFF get lost when the path changes.
But those paths are not relevant when importing translations back, and I can work around it (if you see the linked script, I'm already making changes to those paths) ;-)

[Remaining problems](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/372/files#diff-23c0d0ff5a7678a6d691121f6e2a4298bd8bbabb5e79f9b5049c14a880cd38e7) (ignoring the path changes):
* We seem to be losing a couple of files:
  * `../../nebula/ui/components/MZSystemAlert.qml`: these strings seem to be completely gone.
  * `./src/apps/vpn/ui/screens/ScreenCrashReporting.qml`: for some reason, these strings are showing up in a different file (`generated/l18nstrings_p.cpp`), but that's OK.

flodolo:
I created a branch in the l10n repository with the [path rewrite](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/compare/main...rewrite_path_single_app), then ran automation using that branch and this branch as sources.

This is the [PR created](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/373/). As you can see, the only diffs are about the two qml files I mentioned

brizental:
> ../../nebula/ui/components/MZSystemAlert.qml: these strings seem to be completely gone.

Hum, this is the nebula bug I thought could be done in a follow up. I know what it is about though, we have removed some .qrc files that need to be brought back apparently. I'll send a commit correcting this in a bit.

brizental:
@flodolo Ok, the last commit should bring back the strings from `../../nebula/ui/components/MZSystemAlert.qml`. Since you are saying `./src/apps/vpn/ui/screens/ScreenCrashReporting.qml` I'll just ignore that, sounds good?

flodolo:
> @flodolo Ok, the last commit should bring back the strings from `../../nebula/ui/components/MZSystemAlert.qml`. Since you are saying `./src/apps/vpn/ui/screens/ScreenCrashReporting.qml` I'll just ignore that, sounds good?

Correct. New diff
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/373/files#diff-15adf12b37d34327459e49e8e323a21ce35551fd9a8bbd1de076d47f0b1e75cb

I can recover translations for the file moving, since the ID remains the same, so we're good from a string-export point of view (as long as the app is localized when built :-) )

.taskcluster.yml
strseb, line 1: That is too much to review, can you disable the auto format thing ?

.taskcluster.yml
```diff
@@ -2,318 +2,323 @@
...
+      ownerEmail:
+        $if: 'tasks_for in ["cron", "action"]'
+        then: "${tasks_for}@noreply.mozilla.org"
+        else:
+          $if: 'tasks_for == "github-push"'
+          then: "${event.pusher.email}"
+          else:
             $if: 'tasks_for[:19] == "github-pull-request"'
-            then: ${event.pull_request.head.ref}
+            then: "${event.pull_request.user.login}@users.noreply.github.com"
```
strseb, line 19: Why

strseb (changes requested):
All good. But please revert [.taskcluster.yml](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7367/files#diff-a728f7e52d751b98eafa856e45594533339b44f229d7b83f930df335391e7e15)

.taskcluster.yml
brizental: Yeah, sorry!

.taskcluster.yml
brizental: Hum, I did not do this... Reverting.

strseb (approved):

brizental:
Just FYI, I will block merging this PR until next week when @flodolo is back. I believe [this](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/compare/main...rewrite_path_single_app) branch needs to be merged alongside this PR for l10n not to be broken.

brizental:
Oops, that was turned into an epic and I did not update the description https://mozilla-hub.atlassian.net/browse/VPN-5181

lesleyjanenorton (approved):
Ok! Let's update the ticket number in the description.

brizental:
Talked to @lesleyjanenorton and we have decided to create a feature branch for this epic and only merge these changes into main once 2.16 is out and we are about to cut the release branch for 2.17. The rationale there is that we do not want introduce this big structural change to the codebase that will make it harder to uplift commits from the main branch to the 2.16 branch.

PR #7377, opened 2023-06-30
VPN-4795: Tighter exclusion routing

oskirby:
## Description
Trying to do a bit of cleanup to the exclusion routing behaviour on Linux to try and tighten things up a bit to try and restrict access to the VPN servers and captive portal checks to the VPN client alone.

Along the way, this caused to refactor a few things:
- Combine the `HopConnection` and `InterfaceConfig` structures, they really do exactly the same thing just that one lives in the Daemon and one lives in the Controller.
- Convert the `hopindex` into a `hopType` enum to convey a bit more about the intent of what each Wireguard peer is trying to do.
- Move the configuration of exclusion routes down into the daemon to better take advantage of platform-specific features (such as throw routes on Linux).

## Checklist
    

Gela (commented):
Could this change be tested with functional and/or unit tests?

src/apps/vpn/interfaceconfig.h
```diff
@@ -10,8 +10,13 @@
...
 
-struct InterfaceConfig {
-  int m_hopindex = 0;
+class QJsonObject;
+
+class InterfaceConfig {
+ public:
+  InterfaceConfig() {}
+
+  QString m_hopType;  // One of: "single", "entry", or "exit"
```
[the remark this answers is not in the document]
oskirby: It has to get converted into JSON in order to be serialized over the IPC connection, which doesn't support enums, but you are right that this would be cleaner if we used enums in the C++ side of things.

oskirby:
> Could this change be tested with functional and/or unit tests?

Unfortunately, I don't think so, the interface configuration and bringup is pretty close to the platform and that isn't something that we can test in our unit and functional testing setup as it stands now.

lesleyjanenorton:
@oskirby  just merged your #7494 so `Add Wasm_Chrome` should pass after rebasing

src/apps/vpn/localsocketcontroller.cpp
```diff
@@ -114,49 +114,12 @@ void LocalSocketController::daemonConnected() {
...
 }
 
-void LocalSocketController::activate(const HopConnection& hop,
-                                     const Device* device, const Keys* keys,
+void LocalSocketController::activate(const InterfaceConfig& config,
                                      Controller::Reason reason) {
   Q_UNUSED(reason);
 
-  QJsonObject json;
+  QJsonObject json = config.toJson();
```
strseb, line 121: Love the cleanup!

src/apps/vpn/platforms/android/androidcontroller.cpp
```diff
@@ -106,56 +106,64 @@ AndroidController::~AndroidController() { MZ_COUNT_DTOR(AndroidController); }
...
+
   AndroidVPNActivity::connectService();
 }
 
-void AndroidController::activate(const HopConnection& hop, const Device* device,
-                                 const Keys* keys, Controller::Reason reason) {
-  Q_ASSERT(hop.m_hopindex == 0);
+void AndroidController::activate(const InterfaceConfig& config,
+                                 Controller::Reason reason) {
+  Q_ASSERT(config.m_hopType == InterfaceConfig::SingleHop);
```
strseb, line 120: We should consider also just doing `QJsonObject json = config.toJson();`
and adding the few extra things android requires to that obj. 
However that can be a followup that i gladly will take.

strseb (approved):

src/apps/vpn/platforms/android/androidcontroller.cpp
oskirby: I thought about it, but the format of the JSON used by the `AndroidController` is slightly different from what the desktop platforms use, so I opted to leave this method mostly untouched.

PR #7407, opened 2023-07-05
Chore: Remove obsolete addons

brizental:
Anything that was showing only for 2.13 or older was deleted + the 2.14 what's new message.


flodolo (approved):
Thanks for cleaning things up. If we could uplift this to [releases/2.16.0](https://github.com/mozilla-mobile/mozilla-vpn-client/tree/releases/2.16.0), we could effectively clean up strings.

PR #7418, opened 2023-07-05
update version

mcleinman:
Version bump for dot release


strseb (approved):

Gela (approved):

oskirby (approved):

lesleyjanenorton (approved):
Merging this despite the CI failures because:
- We aren't merging to main
- 2.15.2 is for Linux only and the failures don't impact release 
- The failures may not even be within our power to fix. To paraphrase Naomi: We suspect the macos workers have cached something they shouldn't have.

PR #7421, opened 2023-07-05
VPN-4238 Hide Speedtest restart button when VPN is disabled

janriokrause:
## Description

It does not make sense to restart the speedtest when you are not connected to the VPN. I would suggest hiding the restart button so you cannot enter an invalid state.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4238

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/1068037/661f30cf-bc05-45b1-92fd-fcd679154576

### DIFF
--- a/src/apps/vpn/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
+++ b/src/apps/vpn/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
@@ -146,7 +146,7 @@ Rectangle {
     MZIconButton {
         id: connectionInfoRestartButton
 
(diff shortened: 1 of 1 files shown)


MattLichtenstein (approved):
This looks good, especially since I'd assume if that button was clicked while disconnected we would get a crash in [ConnectionBenchmark::start()](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/efafaddc775f9e9683968be02805ee2933e7ec63/src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp#L90)

PR #7430, opened 2023-07-06
Ignore 2.15.* branches when extracting strings

flodolo:
## Description

Per conversation in Slack: the `2.15.0` and `2.15.1` branches were removed yesterday from this repository. That means that thousands of translations were removed from the l10n repository (won't link the PR, as it's large enough to make Firefox stall).

Reintroducing these branches exposes about 150 strings for translation, and that's not something we can afford (we can't recover previous translations). We also can't stop translating until these branches are removed again, so I'm skipping them in the generation process.

I will file an issue to make sure we remove them once there are no more `2.15.*` branches in the repository.

## Checklist
    
### DIFF
--- a/scripts/utils/generate_ts.sh
+++ b/scripts/utils/generate_ts.sh
@@ -65,6 +65,15 @@ EOF
   cp build-addons/*.ts addon_ts
 
   for branch in $(git branch -r | grep origin/releases); do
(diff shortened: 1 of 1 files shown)


strseb (approved):

flodolo:
Filed #7431 for the follow-up cleaning

PR #7432, opened 2023-07-06
i18n Update-Automation for Release branches 

strseb:
## tldr

On Main we already have a workflow that daily pulls in the newest language-submodule.
This PR add's a bit of automation to do that too for every branch in `release/*`that has no Github Release or Release Draft present. 

## WHY

We want to get rid of: 

```
for i in src/apps/*/translations/i18n; do
            git submodule update --remote $i
```
Because that makes it impossible to trace what commit has been built with which language and  makes it impossible for the CI to build an older release, once the strings have been removed on the head of the strings repo. Which i just may have caused 😓 

## Examples: 

Example Run 1: 
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/5475097381
Created PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7439

Example Run before i added 2.15.2 as a draft: (note i cancelled the 2.15.2 update ) 
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/5474990930

### DIFF
--- a/.github/workflows/i18n_update.yaml
+++ b/.github/workflows/i18n_update.yaml
@@ -94,7 +94,7 @@ jobs:
           commit-message: "[Bot] Update i18n"
           branch: i18n_automation_for_${{matrix.branch}}
           delete-branch: true
-          title: "[Bot]  Update i18n on ${{matrix.branch}}"
+          title: "[Bot] Update i18n on ${{matrix.branch}}"
           token: ${{ secrets.WIKI_TOKEN }}
       - name: Approve l10n_branch.
         run: gh pr review --approve "$PR_URL"
--- a/.github/workflows/i18n_update.yaml
+++ b/.github/workflows/i18n_update.yaml
@@ -108,3 +108,4 @@ jobs:
         env:
           PR_URL: ${{ steps.cpr.outputs.pull-request-url }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
--- a/.github/workflows/i18n_update.yaml
+++ b/.github/workflows/i18n_update.yaml
@@ -88,7 +88,7 @@ jobs:
       - uses: peter-evans/create-pull-request@v5
         id: cpr
         with:
-          # This may not be github, cuz gh actions cannot self approve
+          # This may not be GitHub, because GitHub actions cannot self-approve
           author: GitHub <noreply@github.com>
           committer: GitHub <noreply@github.com>
           commit-message: "[Bot] Update i18n"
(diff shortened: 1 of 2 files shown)


.github/workflows/i18n_update.yaml
```diff
@@ -50,4 +50,62 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+      - name: Generate Branch List
+        id: gen
+        shell: pwsh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ./.github/scripts/get_unreleased_branches.ps1  >> $env:GITHUB_OUTPUT
```
flodolo, line 69: I don't think the script is executable?

.github/workflows/i18n_update.yaml
```diff
@@ -50,4 +50,62 @@ jobs:
...
+        with:
+          submodules: 'true'
+          ref: 'releases/${{matrix.branch}}'
+      - name: Checkout mozilla-vpn-client-l10n Git
+        run: |
+          git submodule update --init --remote src/apps/vpn/translations/i18n/
+      - uses: peter-evans/create-pull-request@v5
+        id: cpr
+        with:
+          # This may not be github, cuz gh actions cannot self approve
```
flodolo, line 91:
```suggestion
          # This may not be GitHub, because GitHub actions cannot self-approve
```

.github/workflows/i18n_update.yaml
```diff
@@ -50,4 +50,62 @@ jobs:
...
+          author: GitHub <noreply@github.com>
+          committer: GitHub <noreply@github.com>
+          commit-message: "[Bot] Update i18n"
+          branch: i18n_automation_for_${{matrix.branch}}
+          delete-branch: true
+          title: "[Bot]  Update i18n on ${{matrix.branch}}"
```
flodolo, line 97:
```suggestion
          title: "[Bot] Update i18n on ${{matrix.branch}}"
```

flodolo (commented):
Not sure how I feel about adding dependency on PowerShell (but that's more for other devs to decide).

If you don't want to deal with bash (I get it, really), Python seems like a more portable and easier solution.

mcleinman (approved):
👍🏻

.github/workflows/i18n_update.yaml
strseb: Powershell does not really care about that 😅

strseb:
>Not sure how I feel about adding dependency on PowerShell (but that's more for other devs to decide).
If you don't want to deal with bash (I get it, really), Python seems like a more portable and easier solution.


I guess we are way past that point - most of our windows CI is completly running on top of that, however if you prefer python we can rewrite that too, i dont mind :)

PR #7454, opened 2023-07-06
VPN-5189: Pin i18n submodule for 2.15.2 (Linux only)

mcleinman:
## Description

We need to pin the i18n branch we're using for 2.15.2, which is only going to be available on Linux. I believe this does that. It:
- updates to the proper i18n commit 
- comments out the i18n update (not sure which of these two will be run, so commented out both)

While this is ugly, it's only going into the 2.15.2 branch, so I think this is okay.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5189

## Checklist
    
### DIFF
--- a/scripts/linux/script.sh
+++ b/scripts/linux/script.sh
@@ -122,7 +122,8 @@ mkdir .tmp || die "Failed to create the temporary directory"
 
 print Y "Get the submodules..."
 git submodule update --init --depth 1 || die "Failed to init submodules"
-git submodule update --remote i18n || die "Failed to pull latest i18n from remote"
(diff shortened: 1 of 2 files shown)


strseb (approved):
LGTM

oskirby (approved):

flodolo (approved):

lesleyjanenorton:
Merging this despite the CI failures because:
- We aren't merging to main
- 2.15.2 is for Linux only and the failures don't impact release
- The failures may not even be within our power to fix. To paraphrase Naomi: We suspect the macos workers have cached something they shouldn't have.

PR #7457, opened 2023-07-06
fixing aqt download issues by using apt

mozrokafor:
## Description
AQT has been brittle a bit lately. 
Let's switch to using the normal apt-packages, like we do on the unit-test tasks.


brizental (approved):

oskirby (approved):
Looks good!

PR #7495, opened 2023-07-11
version bump to 2.15.3

mcleinman:


oskirby (approved):
Looks good to me!

lesleyjanenorton (approved):
For posterity.  We are creating this dot release because: 

> Yesterday releng started the release process for 2.15.2, before we realized that cert issue was going to mess things up the release. Thus, version number 2.15.2 is burned for Linux. So I’m going to put up a new release branch called 2.15.3, which is identical to 2.15.2 (except a version bump).

mcleinman:
Merged w/o all tests passing because 2 remaining were iOS and macOS - and this release is Linux only. We decided to do same thing on 2.15.2, so felt okay doing it here.

PR #7535, opened 2023-07-17
[Sentry] Print where to find the crashreport

strseb:
## Description

Small nit: right now we only record the event id. This makes lookup of a sentry report possible, but you need to remember where to go. 
Let's just print the url for the report, so everony can quickly lookup what we collected.


lesleyjanenorton (approved):
Thanks @strseb, this is really helpful.

Gela (approved):

PR #7541, opened 2023-07-18
VPN-5152 - AndroidIAPHandler handle Q_Assert cases

strseb:
## Description

`AndroidIAPHandler::updateProductsInfo` is a microtask scheudled from the android main thread onto the qt main thread. 
However in our crashes we can see that before the crash we have [breadcrumbs](https://mozilla.sentry.io/issues/4264443753/) with 
```
[16.07.2023 13:30:23.081] (ProductsHandler) Debug: Stop products registration
``` 

This resets the model, causing:
```
    ProductsHandler::Product* productData =
        productsHandler->findProduct(productIdentifier);
```
To return `nullptr`, therefore we segfault. 
Our assert's are catching this, so let's also use them in release :)


src/shared/sentry/sentryadapter.cpp
```diff
@@ -70,6 +70,7 @@ void SentryAdapter::init() {
           &SentryAdapter::onLoglineAdded);
 
   sentry_options_t* options = sentry_options_new();
+  sentry_options_set_max_breadcrumbs(options,500);
```
Gela, line 73:
```suggestion
  sentry_options_set_max_breadcrumbs(options, 500);
```

Gela (approved):

src/shared/sentry/sentryadapter.cpp
```diff
@@ -59,6 +59,7 @@ void SentryAdapter::init() {
 
   sentry_options_t* options = sentry_options_new();
   sentry_options_set_dsn(options, Constants::SENTRY_DSN_ENDPOINT);
+  sentry_options_set_max_breadcrumbs(options,500);
```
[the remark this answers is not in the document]
strseb: That was waaay to less stuff to find out what is happening q_q

src/apps/vpn/platforms/android/androidiaphandler.cpp
```diff
@@ -270,14 +270,18 @@ void AndroidIAPHandler::onSubscriptionFailed(JNIEnv* env, jobject thiz,
...
+    return;
+  }
   QStringList productsUpdated;
 
   for (auto product : returnedProducts) {
     QString productIdentifier = product[QString("sku")].toString();
     ProductsHandler::Product* productData =
         productsHandler->findProduct(productIdentifier);
-    Q_ASSERT(productData);
+    if (productData == nullptr) {
```
vinocher, line 282: Nit: Typically 'if (!productData)' is preferred. See http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if

src/shared/sentry/sentryadapter.cpp
```diff
@@ -70,6 +70,7 @@ void SentryAdapter::init() {
           &SentryAdapter::onLoglineAdded);
 
   sentry_options_t* options = sentry_options_new();
+  sentry_options_set_max_breadcrumbs(options, 500);
```
vinocher, line 73: Looks like the default is 100. Is that too small? Just curious why this is being changed. Also, consider adding a comment explaning why the default is being changed.

vinocher:
There seem to be other calls in andriodiaphandler.cpp to  ProductsHandler::instance() where the returned value is used without a null check.

vinocher:
Does the similar code in AndroidIAPHandler::validatePurchase need to change the assert to a nullcheck?
  ProductsHandler::Product* productData =
      ProductsHandler::instance()->findProduct(sku);
  Q_ASSERT(productData);

PR #7546, opened 2023-07-18
[macOS]Mozilla VPN app crashes when entering specific symbols using Czech keyboard

Gela:
## Description

VPN client crashes if we enter a special character in various keyboards such as Czech and German on OSX only due to the check `if (Qt.platform.os === "osx")` here. This change was causing an infinite loop when QT autocompletes the special character, causing a crash.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5132

## Checklist
    
### DIFF
--- a/nebula/ui/components/forms/MZTextField.qml
+++ b/nebula/ui/components/forms/MZTextField.qml
@@ -56,8 +56,8 @@ TextField {
     onTextChanged: {
         if (Qt.platform.os === "osx") {
             textField.focusReasonA11y = true
-            textField.focus = false;
-            textField.forceActiveFocus();
+            Accessible.focused = false
+            Accessible.focused = true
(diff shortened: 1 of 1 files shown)


csadilek (approved):
Tested again locally to verify the crash/freeze is resolved and that voice over still works for all input fields while typing.

To clarify for others: It was the removal of focus from the input field in the original workaround that caused a freeze/crash while entering/completing specific symbols that need multiple key strokes e.g., á'bć. 

This workaround prevents that and specifically toggles `Accessible.focused` instead. There is likely still an underlying bug in QT which we originally filed as https://bugreports.qt.io/browse/QTBUG-108189. We should follow up there.

lesleyjanenorton:
@Gela when you have a minute, let's uplift this to 2.16. Thanks!

PR #7547, opened 2023-07-18
Reset accessibility focus (#7546)

Gela:
Cherry picking this commit from `main` to fix a crash with special characters: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7546

### DIFF
--- a/nebula/ui/components/forms/MZTextField.qml
+++ b/nebula/ui/components/forms/MZTextField.qml
@@ -56,8 +56,8 @@ TextField {
     onTextChanged: {
         if (Qt.platform.os === "osx") {
             textField.focusReasonA11y = true
-            textField.focus = false;
-            textField.forceActiveFocus();
+            Accessible.focused = false
+            Accessible.focused = true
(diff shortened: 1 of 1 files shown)


lesleyjanenorton (approved):
Thanks @Gela !

nebula/ui/components/forms/MZTextField.qml
```diff
@@ -56,8 +56,8 @@ TextField {
     onTextChanged: {
         if (Qt.platform.os === "osx") {
             textField.focusReasonA11y = true
-            textField.focus = false;
-            textField.forceActiveFocus();
+            Accessible.focused = false
```
lesleyjanenorton, line 59: Might be worthwhile to follow this up with a helpful code comment to our future selves explaining why we set `Accessible.focused` to false on L59, then to true on the next.

Gela: There is a comment by Florian right above `onTextChanged` that explains this weird workaround :D

PR #7564, opened 2023-07-20
VPN-5739: MacOS functional testing workflow

oskirby:
## Description
The functional test suite is platform agnostic, so there isn't really a good reason not to run them on every platform to improve our confidence that we aren't encountering some unexpected breakage. So let's do it on MacOS at least.

## Reference
Github #8334 ([VPN-5739](https://mozilla-hub.atlassian.net/browse/VPN-5739))

## Checklist
    
[VPN-5739]: https://mozilla-hub.atlassian.net/browse/VPN-5739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lesleyjanenorton:
oh my

oskirby:
So, it seems that the dummy client doesn't work with a static build because the QML import scanner is a bit of a mess and we have just never noticed because the CI and most developers use a dynamically linked build when testing. This will necessitate some refactoring of the dummy client to resolve before this PR can move forward.

oskirby:
This does enable the tests on MacOS, but they do exhibit a small degree of instability in CI. In particular the `Settings` and `Onboarding` tests seem to occasionally encounter a failure, but will typically pass on a re-run.

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+    outputs:
+      matrix: ${{ steps.testGen.outputs.tests }}
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+
+      - name: Install dependencies
```
Gela, line 30: Can we possibly move all dependencies under one singular job? Currently "build_test_app" and "functionaltests" jobs both install a number of dependencies, but these packages can be installed once and both "build_test_app" and "functionaltests" jobs can rely on it. Same comment for the Linux logic too.

Something like:

```c++
jobs:
  install-dependencies:
    name: Install Dependencies
    runs-on: macos-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3
        with:
          submodules: 'recursive'
      - name: Install Python and Homebrew
        run: |
          brew install python
          brew install ninja
      - name: Install Python packages
        run: |
          pip3 install -r requirements.txt
```

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 15
+          max_attempts: 3
+          command: |
+            export PATH=$GECKOWEBDRIVER:$(npm bin):$PATH
+            export HEADLESS=yes
+            export TZ=Europe/London
+            mkdir -p $ARTIFACT_DIR
+            npm run functionalTest --  --retries 3 ${{matrix.test.path}}
```
Gela, line 116:
```suggestion
            npm run functionalTest --  --retries 3 ${{ matrix.test.path }}
```

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+        env:
+          ARTIFACT_DIR: ${{ runner.temp }}/artifacts
+          MVPN_BIN: ./build/dummyvpn
+
+      - name: Uploading artifacts
+        uses: actions/upload-artifact@v3
+        if: ${{ always() }}
+        with:
+          name: ${{matrix.test.name}} Logs
```
Gela, line 125:
```suggestion
          name: ${{ matrix.test.name }} Logs
```

.github/workflows/wasm_tests.yaml
```diff
@@ -115,12 +115,7 @@ jobs:
...
           name: WebAssembly Build Qt6
-          # Destination path, WASM_BUILD_DIRECTORY
-          path: wasm_build
-
-      - name: Build addons
-        shell: bash
-        run: ./scripts/addon/generate_all_tests.py -q /opt/$QTVERSION/gcc_64/bin
+          path: build/
 
       - name: Running ${{matrix.test.name}} Tests
```
Gela, line 120:
```suggestion
      - name: Running ${{ matrix.test.name }} Tests
```

.github/workflows/wasm_tests.yaml
```diff
@@ -131,5 +126,6 @@ jobs:
           command: |
             export PATH=$GECKOWEBDRIVER:$(npm bin):$PATH
             export HEADLESS=yes
-            export WASM_BUILD_DIRECTORY=$(pwd)/wasm_build
+            export WASM_BUILD_DIRECTORY=$(pwd)/build/wasm_build
+            export MVPN_ADDONS_PATH=$(pwd)/build/addons
             xvfb-run -a npm run functionalTestWasm --  --retries 3 ${{matrix.test.path}}
```
Gela, line 131:
```suggestion
            xvfb-run -a npm run functionalTestWasm --  --retries 3 ${{ matrix.test.path }}
```

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+        shell: bash
+        run: |
+          pip3 install -r requirements.txt
+          brew install ninja
+
+      - name: Install Qt6
+        shell: bash
+        run: |
+          wget https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozillavpn.cache.level-3.toolchains.v3.qt-mac.latest/artifacts/public%2Fbuild%2Fqt6_mac.zip -O qt6_mac.zip
+          unzip -a -d ${{github.workspace}} qt6_mac.zip
```
Gela, line 40: Super nit, we seem to go back and forth between the 2 styles of formatting. I suggested we add a space between all `{{` for the sake of consistency but there are also plenty in the existing old code without any spaces so if you don't want to make this change, I'm totally fine with it as is. 

```suggestion
          unzip -a -d ${{ github.workspace }} qt6_mac.zip
```

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Compile test client
+        shell: bash
+        run: |
+          mkdir -p build/cmake
+          cmake -S $(pwd) -B build/cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DCMAKE_PREFIX_PATH=${{github.workspace}}/qt_dist/lib/cmake
```
Gela, line 47:
```suggestion
            -DCMAKE_PREFIX_PATH=${{ github.workspace }}/qt_dist/lib/cmake
```

.github/workflows/macos_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+          pip3 install -r requirements.txt
+          npm install
+
+      - name: Check build
+        shell: bash
+        run: |
+            chmod +x ./build/dummyvpn
+            ./build/dummyvpn -v
+
+      - name: Running ${{matrix.test.name}} Tests
```
Gela, line 105:
```suggestion
      - name: Running ${{ matrix.test.name }} Tests
```

Gela (commented):
Thanks for taking this on. I added some comments. Additionally on Slack we talked about the current intermittent failure rate on MacOS being too high (e.g. about 50% of CI jobs need a rerun to pass) which we agreed is a bit high. We should consider stabilizing the CI before merging this in to prevent PRs wrongfully getting blocked due to failed CI jobs.

Gela:
Out of curiosity, are there plans (or a Jira issue) to enable tests on Windows as well?

lesleyjanenorton (approved):
✨💫🤩

Do we still need `generate_all_tests.py`?

.github/workflows/macos_tests.yaml
oskirby: Not really, because the building of the client and the running of the test happen on separate workers and they both need to have dependencies installed. And worse yet, the dependencies needed by both of these tasks are slightly different; `build_test_app` needs the build dependencies while `functionaltests` needs the runtime dependencies.

We use multiple workers so that the tests can run in parallel, and it's also how Github's matrix feature works under the hood.

Gela: That makes sense, thanks for the additional context!

.github/workflows/linux_tests.yaml
```diff
@@ -1,4 +1,4 @@
-name: Functional tests
+name: Linux Tests
```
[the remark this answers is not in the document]
oskirby: I was seeking to make the names of the various functional test suites consistent, but I found that the "Functional" part was redundant, because the test jobs appear as `Linux Functional Tests / Functional test (xxxx, yyyy)` and this was getting a bit long. But I can certainly put it back if you think it helps.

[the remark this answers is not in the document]
Gela: Ah in that case please disregard this comment. As long as the CI job mentions "functional" so they don't get confused with unit tests, there's no need for duplicating

oskirby:
> Out of curiosity, are there plans (or a Jira issue) to enable tests on Windows as well?

There is now: #8388 [VPN-5768](https://mozilla-hub.atlassian.net/browse/VPN-5768)

oskirby:
> ✨💫🤩
> 
> Do we still need `generate_all_tests.py`?

The only place it was being used was in the WASM functional tests, as this is the only place that required manual test generation, but I felt it was kind of silly to keep that script around for this one use case.

Gela (approved):
wooo thank you!!!

PR #7565, opened 2023-07-20
VPN-5222: Fix signal for VPN connection metrics on iOS/Android

mcleinman:
## Description

We were sending an extra set of connection start pings when the mobile apps launched with an already-active VPN. This fixes the bug. See detailed diagnosis on the ticket.

To test, you'll need to add a line like `Glean.shared.setDebugViewTag("MattVPN")` in `iosgleanbridge.swift` after Glean is initialized (or before, I don't believe it matters). Alternatively, you could rely on an environment variable to test this, but then you must be sure to re-launch the app from xcode, and not on the device (as the re-run on the device won't have the environment var).

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5222

## Checklist
    
### DIFF
--- a/src/apps/vpn/controller.cpp
+++ b/src/apps/vpn/controller.cpp
@@ -545,6 +545,8 @@ void Controller::connected(const QString& pubkey,
   // We have succesfully completed all pending connections.
   logger.debug() << "Connected from state:" << m_state;
   setState(StateOn);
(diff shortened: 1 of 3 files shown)


MattLichtenstein (approved):
Nice catch on my oversight 😅

PR #7085, opened 2023-06-06
VPN-4707 - Initialize Glean on the iOS network extension

brizental:
~Draft as it is still missing the synchronization of uploadEnabled between netext and main app.~

This is ready for review.

Note: I suggest reviewing this commit by commit.

PR #7566, opened 2023-07-20
Refactor some Swift in iOS Tunnel Manager

mcleinman:
## Description

When the [IOSTunnelManager was added](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7085/), I talked with Bea about putting up a follow up PR to refactor it to be a bit more Swift-y. 

This PR:
- Removes a bunch of force unwraps
- Makes it a bit more Swift-y with optional chaining, additional guard checks, etc.
- Removes some unnecessary small things (semicolons, the `get` in the computed vars, the return in front of completion handler)
- It removes a fatal error. The caller deals with the lack of tunnel when an error is returned, so we don't need to crash the app - the caller can deal with this. 
- I didn't see where the failure state for the tunnel was ever used, so I removed it to simplify.

I considered removing the `withTunnel` function, but decided to save that for another time - As of now, it's only used in `IOSControllerImpl`.

## Reference

N/A

## Checklist
    
### DIFF
--- a/src/apps/vpn/platforms/ios/iostunnelmanager.swift
+++ b/src/apps/vpn/platforms/ios/iostunnelmanager.swift
@@ -5,6 +5,10 @@
 import Foundation
 import NetworkExtension
 
+class TunnelError: Error {
+
+}
+
 // Singleton that manages the VPN tunnel and exposes functions
 // for interacting with it.
 class TunnelManager {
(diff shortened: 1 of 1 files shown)


brizental (approved):
Not sure I am the best person to review this, but 🤷

Issue #7197, opened 2023-06-14
Significantly increase the rendering speed of the "Server list" screen


data-sync-user:
!image-20230614-141454.png|width=1512,height=670!

On my journey on finding out why iOS is “slow” I found this sweet blob of slowness. 

What you can see here is me on the Multihop screen opening one of the server lists. 

We take **1.11 SECOND** to create the QML objects in that view and blocking our Main thread while doing so. :sweat_smile: I guess we have too many servers now!

I’m pretty sure the offender is our Repeater - as I can bump that to 0.00s when I make the Proxy-Filter match nothing. 
i’m also sure it is not the Proxy-Filter being slow, as when i am not using it (pass `VPNServerCountryModel` directly) that makes no impact. 

Question to the QML people, I think repeater creates the object’s directly (right?), if so - can we maybe switch this to a list that only load’s elements as you scroll through? 

cc [~accountid:60482d76ceccdd006a1ce9dd] [~accountid:6237a3581c7f6a007047eb38]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-5021)

data-sync-user:
➤ Lesley Norton commented:

{quote}Question to the QML people, I think repeater creates the object’s directly (right?), if so - can we maybe switch this to a list that only load’s elements as you scroll through?{quote}

This. We should switch to a ListView {}. If I recall correctly we used that Repeater {} because it was easier to center the current server (clearly, no longer a good enough reason) and possibly because of some intricacies with our search bar. Either way, let’s fix this.

data-sync-user:
➤ Lesley Norton commented:

Same story for App Permissions I suspect.

PR #7571, opened 2023-07-21
Use `ListView` for server list to increase performance

janriokrause:
## Description

As more countries and servers are added, the server list screen becomes slower. The `Repeater` QML type creates all items at once. The `ListView` creates only a few visible items and loads more as needed. As a result, the performance increases enormously.

## Reference

* https://mozilla-hub.atlassian.net/browse/VPN-5021
* #7197

### DIFF
--- a/src/apps/vpn/models/servercountrymodel.cpp
+++ b/src/apps/vpn/models/servercountrymodel.cpp
@@ -239,14 +239,16 @@ const QString ServerCountryModel::countryName(
 
 const int ServerCountryModel::indexOfCountryCode(
     const QString& countryCode) const {
-  int i = 0;
-  for (const ServerCountry& country : m_countries) {
-    if (country.code() == countryCode) {
-      return i;
-    }
-    i++;
+  auto it = std::find_if(m_countries.begin(), m_countries.end(),
+                         [&countryCode](const ServerCountry& country) {
+                           return country.code() == countryCode;
+                         });
+
+  if (it != m_countries.end()) {
+    return std::distance(m_countries.begin(), it);
+  } else {
+    return -1;
   }
-  return -1;
 }
 
 void ServerCountryModel::retranslate() {
--- a/src/apps/vpn/models/servercountrymodel.cpp
+++ b/src/apps/vpn/models/servercountrymodel.cpp
@@ -237,6 +237,18 @@ const QString ServerCountryModel::countryName(
   return QString();
 }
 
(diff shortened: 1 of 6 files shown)


lesleyjanenorton (commented):
Thank you for the patch @janriokrause ! A few tabbing and spacing nits but overall this is looking good. I appreciate the cleanup and refactoring efforts here as I know these files can be a lot to contend with. 

A few things we'll need to do before we can merge:
- Restore 'scroll into view on tab or arrow' functionality to the server list items. This was previously handled by [this fn](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/34b6fae6ef1ef9bb9c6b3c6715a5ba744af7a824/nebula/ui/components/MZFlickable.qml#L90) which supports scrolling into view for a 'FlickView' but not within a 'ListView'.  Scrolling into view has occasionally been a real chore so feel free to reach out to @MattLichtenstein or I next week and we'll be happy to assist. 
- Run the sticky scrollbar changes by product and UX for blessing (I can own this)

MattLichtenstein:
Something else I observed was we are now scrolling the server list so that the country of the currently selected city is at the top of the list on open, whereas we used to center the selected city in the middle of the scroll view.

I don't dislike this new pattern, but we should clear it with UX. Also, there is the edge case when selecting, for example, Seattle under USA at the bottom of the list, a country with so many different server cities so that when re-opening the list after selection, the currently selected city is not visible

src/apps/vpn/models/servercountrymodel.cpp
```diff
@@ -237,6 +237,18 @@ const QString ServerCountryModel::countryName(
...
+    const QString& countryCode) const {
+  int i = 0;
+  for (const ServerCountry& country : m_countries) {
+    if (country.code() == countryCode) {
+      return i;
+    }
+    i++;
+  }
+  return -1;
+}
```
vinocher, line 250: Consider using STL for find. Something like:

int ServerCountryModel::indexOfCountryCode(
    const QString& countryCode) const {

    auto it = std::find_if(m_countries.begin(), m_countries.end(),
                           [&countryCode](const Country& country) {
                               return country.code() == countryCode;
                           });

    if (it != m_countries.end()) {
        return std::distance(m_countries.begin(), it);
    } else {
        return -1;
    }
}

strseb:
closing as stale.

PR #7578, opened 2023-07-24
VPN-5012 enforce https

strseb:
## Description

We are not using http anywhere, this was probably added so people 
can run against custom guardian instances :)


lesleyjanenorton (approved):

cure53:
:heavy_check_mark: LGTM

PR #7582, opened 2023-07-24
VPN-5027 - Disable Hardware rendering in ChromeOS vm's

strseb:
## Description
So i was pretty sure this is not a "Mozilla VPN issue" but i was supposed to look into that. So i did. 

```
SIGSEGV: Segfault
  ?, in rx::vk::ImageViewHelper::getLevelLayerDrawImageView

```
The segfault is actually not much to go after. The only online content i found related to that is us and ... chrome. 
Also this is actually in a system library, so something out of our control is going wrong. 

However, looking at google-play, we can clearly see the only devices encountering this crash are ChromeOS tablet's with mediatec chips... So maybe it's their driver? ( Also it would be neat if sentry told me that 😅  ) 

For my smol experiment, we can simply detect chrome-os android-vm's and put them into software rendering. If i am correct, this should stop that from happening :) 
i also added a fallback for the login screen, so it shows not the whole screen with a white background 😅, if we put the client into software rendering fallback.

### DIFF
--- a/src/apps/vpn/command.cpp
+++ b/src/apps/vpn/command.cpp
@@ -30,6 +30,13 @@
 #  include "platforms/windows/windowscommons.h"
 #endif
 
+#ifdef MZ_ANDROID
+#  include <QQuickWindow>
+#  include <QSGRendererInterface>
+
+#  include "platforms/android/androidutils.h"
+#endif
+
 #ifdef MZ_MACOS
 #  include "platforms/macos/macosutils.h"
 #endif
@@ -178,6 +185,12 @@ int Command::runQmlApp(std::function<int()>&& a_callback) {
 
 #ifdef MZ_WINDOWS
   if (WindowsCommons::requireSoftwareRendering()) {
+    QQuickWindow::setGraphicsApi(QSGRendererInterface::Software);
+  }
+#endif
+
+#ifdef MZ_ANDROID
+  if (AndroidUtils::isChromeOSContext()) {
     QQuickWindow::setGraphicsApi(QSGRendererInterface::Software);
   }
 #endif
--- a/src/apps/vpn/models/device.cpp
+++ b/src/apps/vpn/models/device.cpp
@@ -42,7 +42,7 @@ QString Device::currentDeviceName() {
       // status.
       MacOSUtils::computerName();
 #elif MZ_ANDROID
-      AndroidUtils::getDeviceModel();
+      AndroidUtils::getDeviceName();
 #elif MZ_WASM
       "WASM";
 #elif MZ_WINDOWS
--- a/src/apps/vpn/models/device.cpp
+++ b/src/apps/vpn/models/device.cpp
@@ -42,7 +42,7 @@ QString Device::currentDeviceName() {
       // status.
       MacOSUtils::computerName();
 #elif MZ_ANDROID
-      AndroidUtils::GetDeviceName();
+      AndroidUtils::getDeviceModel();
 #elif MZ_WASM
       "WASM";
 #elif MZ_WINDOWS
--- a/src/apps/vpn/platforms/android/androidutils.cpp
+++ b/src/apps/vpn/platforms/android/androidutils.cpp
@@ -30,20 +30,11 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
 
 // static
 QString AndroidUtils::getDeviceName() {
-  QJniEnvironment env;
-  jclass BUILD = env->FindClass("android/os/Build");
-  jfieldID model = env->GetStaticFieldID(BUILD, "DEVICE", "Ljava/lang/String;");
-  jstring value = (jstring)env->GetStaticObjectField(BUILD, model);
-  if (!value) {
-    return QString("Android Device");
+  auto name = readStaticString("android/os/Build", "DEVICE");
+  if(name.isNull()){
+    return QString("Unknown Device");
   }
-  const char* buffer = env->GetStringUTFChars(value, nullptr);
-  if (!buffer) {
-    return QString("Android Device");
-  }
-  QString res(buffer);
-  env->ReleaseStringUTFChars(value, buffer);
-  return res;
+  return name;
 };
 
 bool AndroidUtils::isChromeOSContext() {
@@ -58,21 +49,31 @@ bool AndroidUtils::isChromeOSContext() {
 
 // static
 QString AndroidUtils::getDeviceModel() {
-  QJniEnvironment env;
-  jclass BUILD = env->FindClass("android/os/Build");
-  jfieldID model = env->GetStaticFieldID(BUILD, "MODEL", "Ljava/lang/String;");
-  jstring value = (jstring)env->GetStaticObjectField(BUILD, model);
-  if (!value) {
-    return QString("Android Device");
+  auto model = readStaticString("android/os/Build", "MODEL");
+  if(model.isNull()){
+    return QString("Unknown Model");
   }
-  const char* buffer = env->GetStringUTFChars(value, nullptr);
-  if (!buffer) {
-    return QString("Android Device");
-  }
-  QString res(buffer);
-  env->ReleaseStringUTFChars(value, buffer);
-  return res;
+  return model;
 };
+
+
+// static
+QString AndroidUtils::readStaticString(const char* classname, const char* propertyName) {
+  QJniEnvironment env;
+  jclass targetClass = env->FindClass(classname);
+  jfieldID propertyID = env->GetStaticFieldID(targetClass, propertyName, "Ljava/lang/String;");
+  jstring propertyValue = (jstring)env->GetStaticObjectField(targetClass, propertyID);
+  if (!propertyValue) {
+    return QString();
+  }
+  const char* propertyValueBuffer = env->GetStringUTFChars(value, nullptr);
+  if (!propertyValueBuffer) {
+    return QString();
+  }
+  auto guard = qScopeGuard([&] { env->ReleaseStringUTFChars(propertyValue, propertyValueBuffer); });
+  QString res(propertyValueBuffer);
+  return res;
+}
 
 // static
 AndroidUtils* AndroidUtils::instance() {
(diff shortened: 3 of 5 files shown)


src/apps/vpn/ui/screens/initialize/ViewInitialize.qml
```diff
@@ -15,14 +15,25 @@ Item {
...
         }
+    }
+
+    MZRadialGradient {  
+        height: Screen.height
+        width: Screen.width
+        anchors {
+            top: parent.top
+            topMargin:  -window.safeAreaHeightByDevice()
+        }
```
lesleyjanenorton, line 36: If we slap an id on the new fallback rectangle this can just be `anchors.fill: idOfTheFallbackRectangle`.

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,34 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+  return res;
+};
+
+bool AndroidUtils::isChromeOSContext() {
+  /*
+   * If the device name end's with "_cheets" we're
+   * we are running in a Chrome OS / Android Emulator
+   * Situation.
+   */
+  return getDeviceName().endsWith("_cheets");
```
lesleyjanenorton, line 55: How sure are we about the `endsWith` and underscore placement? https://github.com/google/talkback/blame/9b1d5132b074174d02886faccf731e814d69363c/utils/src/main/java/FeatureSupport.java#L36

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,34 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+    return QString("Android Device");
+  }
+  QString res(buffer);
+  env->ReleaseStringUTFChars(value, buffer);
+  return res;
+};
+
+bool AndroidUtils::isChromeOSContext() {
+  /*
+   * If the device name end's with "_cheets" we're
```
lesleyjanenorton, line 51: *ends

lesleyjanenorton:
Thanks for digging in here, Basti!

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,35 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+  QString res(buffer);
+  env->ReleaseStringUTFChars(value, buffer);
+  return res;
+};
+
+bool AndroidUtils::isChromeOSContext() {
+  /*
+   * If the device name ends or starts with "cheets" we're
+   * we are running in a Chrome OS / Android Runtime Container
+   * Situation.
```
Gela, line 53:
```suggestion
   * If the device name ends or starts with "cheets" we're
   * running in a Chrome OS / Android Runtime Container
   * Situation.
```

Gela (approved):

src/apps/vpn/ui/screens/initialize/ViewInitialize.qml
```diff
@@ -15,14 +15,23 @@ Item {
...
 
         anchors {
             top: parent.top
             topMargin:  -window.safeAreaHeightByDevice()
         }
+    }
+
+    MZRadialGradient {  
+        height: Screen.height
+        width: Screen.width
```
lesleyjanenorton, line 33: can drop these lines too I am pretty sure

lesleyjanenorton (approved):

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,35 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+  jclass BUILD = env->FindClass("android/os/Build");
+  jfieldID model = env->GetStaticFieldID(BUILD, "DEVICE", "Ljava/lang/String;");
+  jstring value = (jstring)env->GetStaticObjectField(BUILD, model);
+  if (!value) {
+    return QString("Android Device");
+  }
+  const char* buffer = env->GetStringUTFChars(value, nullptr);
+  if (!buffer) {
+    return QString("Android Device");
+  }
```
vinocher, line 43: Consider using scope guards to release resources and place that line immediately after acquiring the resource. This can prevent leaks if the code that follows  is modified in the future to exit prior to releasing the resource. It looks like Qt supports something similar to C++ scope_exit:

auto guard = qScopeGuard([&] { env->ReleaseStringUTFChars(value, buffer); });

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,35 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
 }  // namespace
 
 // static
-QString AndroidUtils::GetDeviceName() {
+QString AndroidUtils::getDeviceName() {
+  QJniEnvironment env;
+  jclass BUILD = env->FindClass("android/os/Build");
+  jfieldID model = env->GetStaticFieldID(BUILD, "DEVICE", "Ljava/lang/String;");
+  jstring value = (jstring)env->GetStaticObjectField(BUILD, model);
+  if (!value) {
```
vinocher, line 37: Consider if value should be renamed to modelName and if buffer should be renamed to something like modelNameBuffer to avoid generic names.

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,7 +29,35 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
 }  // namespace
 
 // static
-QString AndroidUtils::GetDeviceName() {
+QString AndroidUtils::getDeviceName() {
```
vinocher, line 32: Isn't this new function the same as AndroidUtils::getDeviceModel?

strseb: Thanks! Splitted them up

src/apps/vpn/command.cpp
```diff
@@ -30,6 +30,13 @@
 #  include "platforms/windows/windowscommons.h"
 #endif
 
+#ifdef MZ_ANDROID
+#  include <QQuickWindow>
```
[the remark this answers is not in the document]
strseb: That's what clang-format is enforcing.  
I would assume our clang-format has `BasedOnStyle: Google` - might be originating from there. 😅 
I'm not aware this was a conscious decision from us when we introduced that :D

src/apps/vpn/platforms/android/androidutils.h
```diff
@@ -33,6 +55,9 @@ class AndroidUtils final : public QObject {
   static QJsonObject getQJsonObjectFromJString(JNIEnv* env, jstring data);
 
  private:
+  static QSting readStaticString(const char* classname,
```
vinocher, line 58: 'QSting' typo

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,22 +29,54 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+      env->GetStaticFieldID(targetClass, propertyName, "Ljava/lang/String;");
+  jstring propertyValue =
+      (jstring)env->GetStaticObjectField(targetClass, propertyID);
+  if (!propertyValue) {
+    return QString();
   }
-  const char* buffer = env->GetStringUTFChars(value, nullptr);
-  if (!buffer) {
-    return QString("Android Device");
+  const char* propertyValueBuffer = env->GetStringUTFChars(value, nullptr);
```
vinocher, line 71: 'value' needs to be 'propertyValue', I think.

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,22 +29,54 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
 }  // namespace
 
 // static
-QString AndroidUtils::GetDeviceName() {
+QString AndroidUtils::getDeviceName() {
+  auto name = readStaticString("android/os/Build", "DEVICE");
+  if (name.isNull()) {
+    return QString("Unknown Device");
```
vinocher, line 35: Previously this returned "Android Device" instead of "Unknown Device". It this difference OK?

src/apps/vpn/models/device.cpp
```diff
@@ -42,7 +42,7 @@ QString Device::currentDeviceName() {
       // status.
       MacOSUtils::computerName();
 #elif MZ_ANDROID
-      AndroidUtils::GetDeviceName();
+      AndroidUtils::getDeviceModel();
```
vinocher, line 45: The name of this method is inconistent with the corresponding names in the other utils. To make this consistent should AndroidUtils::getDeviceModel be renamed to AndroidUtils::getDeviceName, and name the getDeviceName to getDeviceCodename, because "Panther" etc. are code names.

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,23 +29,56 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
+   */
+  auto name = getDeviceCodename();
+  return name.endsWith("_cheets") || name.startsWith("cheets_");
+}
+
+// static
+QString AndroidUtils::getDeviceName() {
+  auto model = readStaticString("android/os/Build", "MODEL");
+  if (model.isNull()) {
+    return QString("Unknown Model");
```
vinocher, line 54: Shouldn't this be "Android Device" because the old code path returned that?

src/apps/vpn/platforms/android/androidutils.cpp
```diff
@@ -29,23 +29,56 @@ constexpr auto UTILS_CLASS = "org/mozilla/firefox/vpn/qt/VPNUtils";
...
   }
-  const char* buffer = env->GetStringUTFChars(value, nullptr);
-  if (!buffer) {
-    return QString("Android Device");
+  return name;
+};
+
+bool AndroidUtils::isChromeOSContext() {
+  /*
+   * If the device name ends or starts with "cheets" we're
```
vinocher, line 42: Nit: "device code name"

vinocher (approved):

PR #7591, opened 2023-07-25
VPN-4974 - Do not Store raw Pointers of ServerCountryModel

strseb:
## Description

I'm hunting the `QJSEngine::qt_metacall Segfault`
We now have a [QML trace:](
https://mozilla.sentry.io/discover/vpn-client:22dd932716d94b8b6a08d705d1b262d5/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=release&homepage=true&name=All+Events&project=4504197915607040&query=release%3A2.16.0&sort=-timestamp&statsPeriod=14d&yAxis=count%28%29)
it's crashing at: 
```
0: {column: -1, function: expression for score, line: 242, source: qrc:/ui/screens/home/servers/ServerList.qml}, 
```

So it's the an evaluated binding on `VPNRecommendedLocationModel`.

My current hunch, given that just before in the logs we can see the client fetched new servers`(ServerCountryModel) Debug: Reading from JSON`. 

So if the ServerCountryModel, calls `m_cities.clear()` - `VPNRecommendedLocationModel` can still hold pointers to those objects. Until Serverlatency refreshes those. 

This Pr Wraps the Raw pointers to `QPointers`, so if the underlying object is deleted, the pointer will be set to `nullptr`. 
Also `VPNRecommendedLocationModel` will now listen to updates on `ServerCountryModel` and update itself.

### DIFF
--- a/src/apps/vpn/models/recommendedlocationmodel.cpp
+++ b/src/apps/vpn/models/recommendedlocationmodel.cpp
@@ -55,36 +55,40 @@ void RecommendedLocationModel::initialize() {
 void RecommendedLocationModel::refreshModel() {
   logger.debug() << "Model refresh";
 
-  QList<const ServerCity*> cities = recommendedLocations(DEFAULT_ENTRIES);
+  QList<QPointer<ServerCity>> cities = recommendedLocations(DEFAULT_ENTRIES);
   if (m_recommendedCities.length() != cities.length()) {
     beginResetModel();
-    m_recommendedCities.clear();
-    for (auto& city : cities) {
-      // Wrap that pointers into a
-      // QPointer, so that we can check if the object has been deleted
-      m_recommendedCities.append(QPointer((ServerCity*)city));
-    }
+    m_recommendedCities.swap(cities);
     endResetModel();
     return;
   }
 
-  m_recommendedCities.clear();
-  for (auto& city : cities) {
-    m_recommendedCities.append(QPointer((ServerCity*)city));
-  }
+  m_recommendedCities.swap(cities);
   int numLocations = static_cast<int>(m_recommendedCities.length());
   emit dataChanged(createIndex(0, 0), createIndex(numLocations - 1, 0));
 }
+// static
+QList<const ServerCity*> RecommendedLocationModel::recommendedLocationsRaw(
+    unsigned int maxResults) {
+  QList<const ServerCity*> raw_list;
+  auto serverCityList = recommendedLocations(maxResults);
+  for (auto& cityPointer : serverCityList) {
+    if (!cityPointer.isNull()) {
+      raw_list.append(cityPointer.data());
+    }
+  }
+  return raw_list;
+}
 
 // static
-QList<const ServerCity*> RecommendedLocationModel::recommendedLocations(
+QList<QPointer<ServerCity>> RecommendedLocationModel::recommendedLocations(
     unsigned int maxResults) {
   double latencyScale = MozillaVPN::instance()->serverLatency()->avgLatency();
   if (latencyScale < 100.0) {
     latencyScale = 100.0;
   }
 
-  QVector<const ServerCity*> cityResults;
+  QVector<QPointer<ServerCity>> cityResults;
   QVector<double> rankResults;
   cityResults.reserve(maxResults + 1);
   rankResults.reserve(maxResults + 1);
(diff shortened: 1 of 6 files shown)


src/apps/vpn/models/recommendedlocationmodel.cpp
```diff
@@ -96,7 +100,7 @@ QList<const ServerCity*> RecommendedLocationModel::recommendedLocations(
     }
     if (i < static_cast<qsizetype>(maxResults)) {
       rankResults.insert(i, cityRanking);
-      cityResults.insert(i, &city);
+      cityResults.insert(i, (ServerCity*)&city);
```
strseb, line 103: Not great, i know but `QPointer` does not work with `const T*` q_q

brizental (approved):

vinocher:
This looks like the root cause for VPN-5208 too! That is also a crash while binding score.

src/apps/vpn/models/recommendedlocationmodel.cpp
```diff
@@ -54,12 +58,21 @@ void RecommendedLocationModel::refreshModel() {
   QList<const ServerCity*> cities = recommendedLocations(DEFAULT_ENTRIES);
   if (m_recommendedCities.length() != cities.length()) {
     beginResetModel();
-    m_recommendedCities.swap(cities);
+    m_recommendedCities.clear();
+    for (auto& city : cities) {
+      // Wrap that pointers into a
+      // QPointer, so that we are notified if they are
```
vinocher, line 64: Nit: We aren't really notified -- we can check if the object has been deleted using isNull

tests/dummyvpn/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 # finalization in order to pick up QML dependencies when built
 # statically.
 qt_add_executable(dummyvpn MANUAL_FINALIZATION)
-
+include(ExternalProject)
```
vinocher, line 9: Is this change needed in this PR?

src/apps/vpn/models/recommendedlocationmodel.cpp
```diff
@@ -54,12 +58,21 @@ void RecommendedLocationModel::refreshModel() {
...
+      // deleted.
+      m_recommendedCities.append(QPointer((ServerCity*)city));
```
vinocher, line 66: Is it possible to change the return type of recommendedLocations to QList<QPointer<ServerCity>>? If so, that would be a cleaner approach?

src/apps/vpn/models/recommendedlocationmodel.cpp
```diff
@@ -118,10 +131,13 @@ QVariant RecommendedLocationModel::data(const QModelIndex& index,
...
   }
-
+  QPointer<ServerCity> city = m_recommendedCities.at(index.row());
   switch (role) {
     case CityRole:
-      return QVariant::fromValue(m_recommendedCities.at(index.row()));
+      if (city.isNull()) {
+        return QVariant();
+      }
+      return QVariant::fromValue(city.data());
```
vinocher, line 140: Just checking that will be no reentrancy during the Qt binding evaluation after this line that can potentially cause a ServerCountryModel change signal to be emitted. If so, that can cause the same dangling pointer crash. I'd expect the Qt binding evaluation to occur without such re-entrancy, but I'm checking because I'm new to this framework.

tests/dummyvpn/CMakeLists.txt
```diff
@@ -47,6 +47,19 @@ endif()
...
+# at least we no longer invoke python scripts to 
+# invoke cmake. 
+#  ¯\_(ツ)_/¯
+ExternalProject_Add(functional_test_addons
+    SOURCE_DIR ${CMAKE_SOURCE_DIR}/tests/functional/addons
+    BINARY_DIR ${CMAKE_SOURCE_DIR}/tests/functional/addons/generated
+    CMAKE_CACHE_ARGS -DQt6_DIR:PATH=${Qt6_DIR} -DQt6QmlTools_DIR:PATH=${Qt6QmlTools_DIR} -DQt6CoreTools_DIR:PATH=${Qt6CoreTools_DIR}
+    INSTALL_COMMAND ""
+)
+add_dependencies(dummyvpn functional_test_addons)
```
vinocher, line 61: Is this change needed in this PR?

src/apps/vpn/models/recommendedlocationmodel.cpp
vinocher: Then how about introducing RecommendedLocationModel::recommendedLocationsRaw() for calling from QML, which would call RecommendedLocationModel::recommendedLocations and convert the returned QList<QPointer<ServerCity>> to QList<const ServerCity*>? Then C++ code would call recommendedLocations() and QML would call recommendedLocationsRaw()

src/apps/vpn/models/recommendedlocationmodel.cpp
vinocher: Thanks for confirming.

vinocher (approved):
Great job figuring this out!

vinocher (approved):

PR #7592, opened 2023-07-25
VPN-5261 - Bring back the 2.15 whats new message

brizental:
This was deleted by mistake, 2.15 is the current release.

### DIFF
--- a/addons/message_whats_new_v2.15/manifest.json
+++ b/addons/message_whats_new_v2.15/manifest.json
@@ -1,0 +1,45 @@
+{
+  "api_version": "0.1",
+  "id": "message_whats_new_v2.15",
+  "name": "What's new message",
+  "type": "message",
+  "conditions": {
+    "min_client_version": "2.15.0",
+    "max_client_version": "2.15.9"
+  },
+  "javascript": {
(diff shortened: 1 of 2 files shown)


flodolo:
Do we really need to reintroduce this in `main`? We are not going to update the 2.15.* branch (translations are frozen), so I feel like we're spending community bandwidth unnecessarily.

brizental:
I think this is a limitation on our addon infrastructure. The real fix here I guess would be to have different versions of the application fetch from different addon endpoints. Another option that came up on our meeting today is marking some strings as obsolete i.e. if they are old new langauges added do not need to translate them, but whatever is there stays there. Is that possible @flodolo ?

brizental:
> I might be missing what's the goal by adding these back (what's the behavior of users with or without, etc.)

The issue here is that 2.15 is the latest available version of the VPN and removing these from the main branch means they are not part of the addons RC -- yes, it is built from main :/. So users updating now to 2.15 are not getting this the "What's new" message.

flodolo (approved):
I'd prefer to keep these untranslated, but I can try to work around it to avoid retranslation if we really need to.

PR #7593, opened 2023-07-25
Sentry-Nit: Report Proper Plattform

strseb:
## Description

Sentry Native is not making a difference between Mobile OS'ses and Desktop: 
So Android and Linux are reported as Linux
Macos is "macOS" and iOS is Darwin 😅.

This PR add's a new tag, splitting this by `MZ_Platform`, making sure we can [quickly sort those](field=event.type&field=project&field=user.display&field=timestamp&field=replayId&name=All+Events&project=4504197915607040&query=platform%3A"MacOS"&sort=-timestamp&statsPeriod=14d&yAxis=count()) :)

### DIFF
--- a/src/shared/sentry/sentryadapter.cpp
+++ b/src/shared/sentry/sentryadapter.cpp
@@ -207,7 +207,7 @@ void SentryAdapter::captureQMLStacktrace(const char* description) {
 #endif
 }
 
-void SentryAdapter::setPlatformTags() const {
+void SentryAdapter::setPlatformTag() const {
 #if defined(MZ_LINUX)
   sentry_set_tag("platform", "Linux");
 #elif defined(MZ_MACOS)
--- a/src/shared/sentry/sentryadapter.cpp
+++ b/src/shared/sentry/sentryadapter.cpp
@@ -96,7 +96,7 @@ void SentryAdapter::init() {
     return;
   };
   m_initialized = true;
-  setPlatformTags();
+  setPlatformTag();
   logger.info() << "Sentry initialized";
 }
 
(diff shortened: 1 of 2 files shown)


src/shared/sentry/sentryadapter.cpp
```diff
@@ -205,3 +206,19 @@ void SentryAdapter::captureQMLStacktrace(const char* description) {
...
+#elif defined(MZ_WINDOWS)
+  sentry_set_tag("platform", "Windows");
+#elif defined(MZ_IOS)
+  sentry_set_tag("platform", "iOS");
+#elif defined(MZ_ANDROID)
+  sentry_set_tag("platform", "Android");
+#else
+  sentry_set_tag("platform", "Dummy");
+#endif
+}
```
brizental, line 224: nit: Newline 🧨

brizental (approved):

Gela (approved):

src/shared/sentry/sentryadapter.cpp
```diff
@@ -205,3 +206,19 @@ void SentryAdapter::captureQMLStacktrace(const char* description) {
   Q_UNUSED(description);
 #endif
 }
+
+void SentryAdapter::setPlatformTag() const {
```
vinocher, line 210: Just curious why this is needed, because the Sentry event dashboard already has the platform info on the right, in the Tag Details section.
For example, for Windows it has something like:
os Windows 10.0.22621
os.name Windows
release 2.16.0

src/shared/sentry/sentryadapter.h
```diff
@@ -145,6 +145,11 @@ class SentryAdapter final : public QObject {
   void needsCrashReportScreen() const;
 
  private:
+  /**
+   * @brief Set's Tags related to the Platform
```
vinocher, line 149: Nit: Typo 'Sets'

vinocher (approved):

vinocher (approved):

PR #7520, opened 2023-07-14
VPN-5181 - Unify the codebase under a single app

brizental:
Feature branch for epic VPN-5181. Draft until we 2.16 is closed and before cutting 2.17.

PR #7595, opened 2023-07-25
Chore - Schedule Sentry Tasks instantly

strseb:
## Description

I've been looking why we have so many "Abnormal" sessions in sentry and so little healthy ones. 
Problem is, session_end will attempt to send a request, if that fails we're abnormal. 

So let's try to do that immediatly instead of scheudling a new task. So we have better channces to get that out. 
This also makes the crash report screen pop instantly, which also is nice c:

### DIFF
--- a/src/apps/vpn/mozillavpn.cpp
+++ b/src/apps/vpn/mozillavpn.cpp
@@ -897,6 +897,8 @@ void MozillaVPN::mainWindowLoaded() {
 #endif
 #ifdef SENTRY_ENABLED
   SentryAdapter::instance()->init();
+  QObject::connect(controller(), &Controller::readyToQuit,
+                   SentryAdapter::instance(), &SentryAdapter::onBeforeShutdown);
 #endif
 }
 
--- a/src/shared/sentry/sentryadapter.cpp
+++ b/src/shared/sentry/sentryadapter.cpp
@@ -64,8 +64,6 @@ void SentryAdapter::init() {
       QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation));
   QString sentryFolder = dataDir.absoluteFilePath("sentry");
 
-  connect(qApp, &QCoreApplication::aboutToQuit, this,
-          &SentryAdapter::onBeforeShutdown);
   connect(log, &LogHandler::logEntryAdded, this,
           &SentryAdapter::onLoglineAdded);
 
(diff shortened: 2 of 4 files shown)


src/apps/vpn/mozillavpn.cpp
```diff
@@ -897,6 +897,8 @@ void MozillaVPN::mainWindowLoaded() {
 #endif
 #ifdef SENTRY_ENABLED
   SentryAdapter::instance()->init();
+  QObject::connect(controller(), &Controller::readyToQuit,
+                   SentryAdapter::instance(), &SentryAdapter::onBeforeShutdown);
```
vinocher, line 901: Isn't it possible to move this into SentryAdapter::init() for encapsulation? Also, a comment may be useful about why this is being done when Controller::readyToQuit is emitted.

src/shared/sentry/sentryadapter.h
```diff
@@ -61,7 +61,7 @@ class SentryAdapter final : public QObject {
    * or drisk writes.
    * After calling this, any call to sentry is UB.
    */
-  Q_SLOT void onBeforeShutdown();
+  static void onBeforeShutdown();
```
vinocher, line 64: Do the comments above this line need be changed because this now called a bit earlier?

src/apps/vpn/mozillavpn.cpp
strseb: Not right now, however this can wait until this is merged:
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7520

Right now Sentry-Adapter sit's in the "shared" part, therefore it cannot know about the Controller.

src/shared/sentry/sentryadapter.cpp
```diff
@@ -149,11 +150,10 @@ void SentryAdapter::transportEnvelope(sentry_envelope_t* envelope,
   sentry_free(sentry_buf);
 
   auto t = new TaskSentry(qt_owned_buffer);
-  TaskScheduler::scheduleTask(t);
+  TaskScheduler::scheduleTaskNow(t);
```
lesleyjanenorton, line 153: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/308692c7ee9d3bda41d8d6e4cd9af9dadc83c659/src/shared/taskscheduler.h#L20-L22
Footgun check: Is surety at 100%?

lesleyjanenorton (approved):

PR #7606, opened 2023-07-26
disable upgrade message

mcleinman:
## Description

Disable upgrade message. This is is so we can release addons. After it is released, this will be reverted.

If for some reason any clients out there are running version 1.0, they will see this message.

## Reference

N/A

## Checklist
    

lesleyjanenorton (approved):
👍

PR #7608, opened 2023-07-26
update log file extension

mcleinman:
## Description

Per discussion from several months ago, update the log file extension to make it easier to work with on macOS. The default program for `.log` files on mac is Console, which is far more powerful than TextEdit (which opens `.txt` files).

I've tested it on iOS and macOS (all desktop logging works the same). I can't find where the extension is on Android, but don't want to block this on Android. (It's somewhere in `VPNUtils.kt`'s `sharePlainText` - is it controlled by `put(MediaStore.Downloads.MIME_TYPE, "text/plain")`?)

## Reference

N/A

## Checklist
    

MattLichtenstein (approved):

vinocher:
Could the PR Description capture the reason why log extensions were decided to be easier to work with? (FYI, on Windows there is no default app association for .log files, so the user will need to create such an association.)

tools/inspector/elements/record-button.js
```diff
@@ -69,7 +69,7 @@ export class RecordButton extends LitElement {
     const blob = new Blob([entries.join('\n')], {type: 'octet/stream'});
     const url = window.URL.createObjectURL(blob);
     a.href = url;
-    a.download = 'mozillavpn.txt';
```
bakulf, line 72: Because we already have logs, I suggest removing the .txt file.

mcleinman:
> Could the PR Description capture the reason why log extensions were decided to be easier to work with? (FYI, on Windows there is no default app association for .log files, so the user will need to create such an association.)

I just updated it. Feel free to push back - when this was discussed several months ago, there were no downsides we could identify. This is one - so I'm now somewhat inclined to just close this out and keep the status quo.

vinocher:
> > Could the PR Description capture the reason why log extensions were decided to be easier to work with? (FYI, on Windows there is no default app association for .log files, so the user will need to create such an association.)
> 
> I just updated it. Feel free to push back - when this was discussed several months ago, there were no downsides we could identify. This is one - so I'm now somewhat inclined to just close this out and keep the status quo.

On Windows, there is no default association, but the user is prompted to choose an app when opening the file, so the user will be able to recover. Should we ask the support team if they have concerns? They may want to weigh in if they sometimes ask customers to open and read logs during a support incident. If they don't, this change seems OK.

mcleinman:
> > > Could the PR Description capture the reason why log extensions were decided to be easier to work with? (FYI, on Windows there is no default app association for .log files, so the user will need to create such an association.)
> > 
> > 
> > I just updated it. Feel free to push back - when this was discussed several months ago, there were no downsides we could identify. This is one - so I'm now somewhat inclined to just close this out and keep the status quo.
> 
> On Windows, there is no default association, but the user is prompted to choose an app when opening the file, so the user will be able to recover. Should we ask the support team if they have concerns? They may want to weigh in if they sometimes ask customers to open and read logs during a support incident. If they don't, this change seems OK.

The support team was fine with this change when we checked a few months ago. (I believe they mostly ask users to send us logs, though I may be wrong.) 

With what Vinoo just wrote, I'm going to get these tests passing, then plan on merging unless anyone speaks up with concerns.

vinocher (approved):

tools/inspector/elements/record-button.js
bakulf: So this comment has not been addressed. Now the user will have 2 files. This is bad. @mcleinman follow up?

PR #7616, opened 2023-07-27
VPN-5215 - Don't use zstd for addon compression

brizental:
Thanks @vinocher for figuring this out!


vinocher:
Just wondering why this started failing now because that build.py code is more than a year old. Maybe some Qt rcc change triggered this. Anyway, the build.py change looks good.

vinocher (approved):

PR #7623, opened 2023-07-28
NIT: Finalize Sentry release on GH Release step

strseb:
## Description
Sentry does not parse versions, but orders them by "newest" seen. 
This is annoying, because if we do a dot release, the timeline is fucked. 
And that ruin's regressions, i.e i mark a bug solved in 2.17 - it will instantly pop up again because sentry thinks 2.15.1 is newer .___. 

image

So let's mark a release as "done" whenever we publish them on gh, that will make sure the order is correct... or at least the same as our gh releases :)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,25 @@
+name: Post GH releases Actions
+
+on:
+  release:
+    types: [published] 
+  workflow_dispatch:
+
+jobs:
+  update_sentry_info:
+    name: Sentry - Mark as Released
+    runs-on: ubuntu-latest
+    steps:
+        - name: Clone repository
+          uses: actions/checkout@v2
+        - name: Get Files
+          env:
+            SENTRY_KEY: ${{ secrets.SENTRY_KEY }}
+          run: | 
+            TAG=${{github.ref_name}}
+            VERSION=${TAG:1}
+            npm install -g @sentry/cli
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,25 @@
...
+          env:
+            SENTRY_KEY: ${{ secrets.SENTRY_KEY }}
+          run: | 
+            TAG=${{github.ref_name}}
+            VERSION=${TAG:1}
+            npm install -g @sentry/cli
+            sentry-cli login --auth-token $SENTRY_KEY            
+            sentry-cli releases --org mozilla -p vpn-client finalize $VERSION
+
+
```
brizental, line 25: nit: Looks like there is rogue whitespace in these last two lines and GH thinks we are missing a newline at the end of the file.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,25 @@
...
+  workflow_dispatch:
+
+jobs:
+  update_sentry_info:
+    name: Sentry - Mark as Released
+    runs-on: ubuntu-latest
+    steps:
+        - name: Clone repository
+          uses: actions/checkout@v2
+        - name: Get Files
```
brizental, line 15: nit: This name doesn't seem to describe what the job is actually doing.

brizental (approved):
r+wc

PR #7657, opened 2023-08-01
VPN-5305: Fix edit button in devices view

MattLichtenstein:
## Description

- "Edit" button "enabled" state in devices view is now bound to active device count, so that if a second device is added or all devices (excluding the current device) are removed, the "Edit" button will enable/disable accordingly (without the need to reload the view)
- "Edit" / "Done" label text now updates to the correct state when a device is added/removed from the list by signing in/out from another device (without the need to reload the view)
-  Fix the Red Button theme to match the design system (it is now more accessible with better contrast)

## Reference

[VPN-5305: The “Edit” button from My devices screen remains disabled when new device is added](https://mozilla-hub.atlassian.net/browse/VPN-5305)

### DIFF
--- a/src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
+++ b/src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
@@ -115,7 +115,7 @@ ListView {
     Connections {
         target: VPNDeviceModel
         function onActiveDevicesChanged() {
-            if(!listView.anySwipesOpen()) listView.isEditing(false)
+            if (!listView.anySwipesOpen()) listView.isEditing(false)
         }
     }
 
--- a/src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
+++ b/src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
@@ -112,6 +112,13 @@ ListView {
         }
     }
 
+    Connections {
+        target: VPNDeviceModel
+        function onActiveDevicesChanged() {
+            if(!listView.anySwipesOpen()) listView.isEditing(false)
+        }
+    }
+
     Timer {
         id: delayTimer
         interval: 300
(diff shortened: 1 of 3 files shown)


src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
```diff
@@ -112,6 +112,13 @@ ListView {
         }
     }
 
+    Connections {
+        target: VPNDeviceModel
+        function onActiveDevicesChanged() {
+            if(!listView.anySwipesOpen()) listView.isEditing(false)
```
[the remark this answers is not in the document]
MattLichtenstein: We don't have a style guide in place for the project yet, but I am not opposed to using the google guidelines

src/apps/vpn/ui/screens/devices/VPNDeviceList.qml
```diff
@@ -112,6 +112,13 @@ ListView {
         }
     }
 
+    Connections {
+        target: VPNDeviceModel
+        function onActiveDevicesChanged() {
+            if (!listView.anySwipesOpen()) listView.isEditing(false)
```
vinocher, line 118: Assume that a swipe is open for Device1 and Device1 was just removed by signing out from another device, causing this code to execute. Will Device1 remain in listView (until the screen is reloaded)? Then, if the user selects delete in the swipe, will that fail silently because the device has already been deleted?

Or is it better to automatically update listView when Device1 is removed?

MattLichtenstein: `listView`'s model is bound to `VPNDeviceModel` so anytime there is a change to `VPNDeviceModel`, `listView` automatically updates. If the model is changed by something outside the UI (i.e. signing out from another device), the list will refresh and all swipes will be closed.

vinocher: Thanks for the explanation.

vinocher (approved):

PR #7660, opened 2023-08-01
Remove --remote from the scripts

Gela:
With the current setup any time the localization team adds a new language, it breaks our CI jobs and functional tests. Removing `--remote` ensures that new languages aren't automatically pulled in every pull request.

@strseb Can you confirm if `--remote` should be removed from the following files too? I was unsure so didn't remove them for now but let me know if they should be removed as well.

```
i18n_update.yaml --- line 73
> git submodule update --init --remote src/apps/vpn/translations/i18n/

import_languages.py --- line 83
> os.system(f"git submodule update --init --remote --depth 1 {i18ndir}")
```


strseb:
>i18n_update.yaml 
 No. that is the thing that will update the languages with a regular cadence :) 
 
 >import_languages.py --- line 83 
 That needs to go

.github/workflows/i18n_update.yaml
```diff
@@ -24,20 +24,6 @@ jobs:
...
-        id: cpr
-        with:
-          # This may not be github, cuz gh actions cannot self approve
-          author: GitHub <noreply@github.com>
-          committer: GitHub <noreply@github.com>
-          commit-message: "[Bot] Update i18n"
-          branch: i18n_automation
-          delete-branch: true
-          title: "[Bot] Update i18n"
-          token: ${{ secrets.WIKI_TOKEN }}
```
strseb, line 40: This need's to say, this is the bot that will daily try to update the languages :)

strseb (approved):
r+wc

PR #7667, opened 2023-08-02
Add a Dry Run mode for LanguageLocalizer

strseb:
## Description
We now have a bot that daily bumps i18n. 
For that it creates a PR. 
Let's add a check, that asserts `languages.json` is generated properly. aka - the script does not prompt for user input to generate that. 

So the PR cannot merge unless we manually fix that file on the branch :)


flodolo:
The idea makes sense to me, but I'm really useless when it comes to reviewing JavaScript code ;-)

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -128,36 +38,41 @@ const LanguageLocalizer = {
...
+                              .filter(f => f.isDirectory())
+                              .map(f=>f.name)
+                              .filter(f=>!f.startsWith(".")); // no
 
-    const languages = await this.sendMessageToClient('languages');
-    if (!Array.isArray(languages.value) || languages.value.length === 0) {
-      console.log('No languages! Are you authenticated already?');
-      this.quit();
-    }
+    this.languages = language_folders;
```
strseb, line 69: Small additional changes, instead of pulling the languages from a client built on that branch, we can also directly pull the languages from the checked out submodule.

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -128,36 +38,41 @@ const LanguageLocalizer = {
 
   async init() {
     program.version('0.0.1')
+    program.option("-c, --check", `
+      Check only, will fail if user input would be required to generate 
+      the file.
+    `);
 
-    program.option(
```
Gela, line 132: How come this was removed?

strseb: That was used to set where the `mozillavpn.exe` is that will be started to get it's languages. 
With this change we will instead check the `i18n` repo, so there is no need to know where the binary is :)

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -209,6 +128,12 @@ const LanguageLocalizer = {
     console.log(`  - Localizing language ${chalk.yellow(language)}...`);
     let languageData = this.oldData.find(s => s.languageCode === language);
     if (!languageData) {
+      if(this.options.check){
+        console.error(`New Unknown Language ${language} - cannot continue`);
+        console.error(`This is Likely a new Language check this branch out and run this`);
+        console.error(`tool locally again!`);
+        process.exit(1);
+      }
       console.log('    Unknown language!');
```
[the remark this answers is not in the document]
Gela: Ah right!

Gela (approved):

.github/workflows/linters.yaml
```diff
@@ -70,3 +70,20 @@ jobs:
           github_token: ${{ secrets.github_token }}
           reporter: github-pr-review # Change reporter
           android: true
+
+  check-language-json:
+    name: Check if languages.json is valid
+    runs-on: ubuntu-22.04
```
vinocher, line 76: Should this be ubuntu-latest?

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -128,36 +38,41 @@ const LanguageLocalizer = {
...
+    this.options = options;
+    if(options.check){
+      console.log("Running Checks Only");
+    }
     this.wbk = WBK({
       instance: 'https://www.wikidata.org',
       sparqlEndpoint: 'https://query.wikidata.org/sparql',
     })
+    const __filename = fileURLToPath(import.meta.url);
+    const __dirname = path.dirname(__filename);
```
vinocher, line 61: Just curious, why is the double underscore prefix naming pattern being used here?

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -209,6 +128,12 @@ const LanguageLocalizer = {
     console.log(`  - Localizing language ${chalk.yellow(language)}...`);
     let languageData = this.oldData.find(s => s.languageCode === language);
     if (!languageData) {
+      if(this.options.check){
+        console.error(`New unknown language ${language} - cannot continue`);
```
vinocher, line 132: Nit: "New" seems redundant here.

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -228,6 +153,12 @@ const LanguageLocalizer = {
     languageData.translations = {};
 
     if (!languageData.wikiDataID) {
+      if(this.options.check){
+        console.error(`${ languageData.languageCode } - No Wikidata ID found.`)
+        console.error(`This is likely a new language. Check this branch out and run this`)
+        console.error(`tool locally again!`)
+        process.exit(1);
+      }
```
vinocher, line 161: Should this block be moved to line 145?

tools/languagelocalizer/languageLocalizer.js
```diff
@@ -228,6 +153,12 @@ const LanguageLocalizer = {
     languageData.translations = {};
 
     if (!languageData.wikiDataID) {
+      if(this.options.check){
+        console.error(`${ languageData.languageCode } - No Wikidata ID found.`)
```
lesleyjanenorton, line 157: Let's only print this and exit out if `(!languageData.wikiDataID && !languageData.alertnativeDataWikiID)`. As is, this will fail (I think) on languages like `es_CL` and `fy_NL` which are both `wikiDataID: null` but have valid values for `alternativeDataWikiID` and are not new languages.

strseb: @lesleyjanenorton - Do you know what the point is of the alternativeDataWikiID? 

```
<http://www.wikidata.org/entity/${
        languageData.wikiDataID ||
        languageData.alternativeWikiDataID}>
``` the sparql query seems to or them either way. 

Looking at that, i think we should enforce 
```ts
{
 wikidataid:string,
 alternativeDataWikiID: string?
}

´´´

PR #7679, opened 2023-08-04
VPN-5246 - Turn off Sentry for iOS to prevent crashes

mcleinman:
## Description

VPN-5246's crash is caused by Sentry on iOS. Sentry creates Android and iOS specific SDKs, but we’re using the native (C/C++) one for all platforms. It seems like this Sentry native SDK [explicitly includes Qt support](https://docs.sentry.io/platforms/native/guides/qt/). However, from that documentation: `The Qt integration is part of the sentry-native SDK, which currently supports Windows, macOS, Linux, and Android.` Additionally, in the sentry-[native SDK readme](https://github.com/getsentry/sentry-native/#platform-and-feature-support), iOS support is nowhere to be found.

Thus, I believe we should not include use the Sentry native SDK on iOS. Per discussion on the ticket, we're feature flagging this for now so we can investigate this further. I've also added a new ticket to re-add Sentry back in: https://mozilla-hub.atlassian.net/browse/VPN-5324

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5324

## Checklist
    
### DIFF
--- a/src/shared/featurelist.h
+++ b/src/shared/featurelist.h
@@ -59,7 +59,7 @@ FEATURE(sentry,                     // Feature ID
         FeatureCallback_true,       // Can be flipped on
         FeatureCallback_true,       // Can be flipped off
         QStringList(),              // feature dependencies
-        FeatureCallback_true)
+        FeatureCallback_sentry)
 
 FEATURE(shareLogs,              // Feature ID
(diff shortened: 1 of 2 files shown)


strseb (approved):

strseb:
Thank you! :)

PR #7740, opened 2023-08-08
VPN-5328: Fix device list item text layout

MattLichtenstein:
## Description

- Elide device name and subtitle after 1 line

| Before  | After |
| ------------- | ------------- |
| image | image |

## Reference

[VPN-5328: Long device name is not wrapped in “My devices” screen](https://mozilla-hub.atlassian.net/browse/VPN-5328)

### DIFF
--- a/src/apps/vpn/ui/screens/devices/VPNDeviceListItem.qml
+++ b/src/apps/vpn/ui/screens/devices/VPNDeviceListItem.qml
@@ -74,7 +74,10 @@ MZSwipeDelegate {
                 spacing: 0
 
                 MZInterLabel {
+                    Layout.fillWidth: true
+
                     text: name
+                    maximumLineCount: 1
                     elide: Text.ElideRight
                     horizontalAlignment: Text.AlignLeft
                 }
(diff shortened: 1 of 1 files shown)


src/apps/vpn/ui/screens/devices/VPNDeviceListItem.qml
```diff
@@ -74,18 +74,21 @@ MZSwipeDelegate {
                 spacing: 0
 
                 MZInterLabel {
+                    Layout.fillWidth: true
+
                     text: name
+                    maximumLineCount: 1
                     elide: Text.ElideRight
```
lesleyjanenorton, line 81: Strange that this wasn't enough to just work. I guess it needed the `.fillWidth`?

lesleyjanenorton (approved):
👍

src/apps/vpn/ui/screens/devices/VPNDeviceListItem.qml
lesleyjanenorton: referring to the `Text.ElideRight` that was already in place

MattLichtenstein: Yep, `Layout.fillWidth` is needed here. It is worth noting that setting `maximumLineCount: 1` or setting `wrapMode: Text.NoWrap` have the same effect and will both cause the text to elide

PR #7747, opened 2023-08-09
QML Perf: Single-Hop - only load the current selected tab

strseb:
See also 🔗 https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7745

## Description
We are loading both the "recomended" part and the whole list - which consumes time and actually makes them both navigateable from talkback. 

Let's unload the part that we don't want to show


MattLichtenstein (approved):
Don't love that we are comparing using object names, but we can come back to this in the future since any better approach will require some refactoring. 

A very welcome change nonetheless. Nice call!

lesleyjanenorton (approved):

PR #7748, opened 2023-08-09
QML - Only load the Segmented nav screen

strseb:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    

src/apps/vpn/ui/screens/home/ViewServers.qml
```diff
@@ -132,7 +138,7 @@ Item {
             return;
         }
 
-        segmentedNav.stackContent.push(multiHopStackView);
+        segmentedNav.stackContent.push(multiHopLoader);
```
MattLichtenstein, line 141: Okay, so I know I am the one who implemented the segmented nav here, but can you think of any reason why we are manually pushing the multihop view/loader into the stacklayout inside of the segmented nav vs adding it via `stackContent` above?

`stackContent` is an alias for the StackLayout's `children` property which can take a list of views/items - I'm wonder if there was a reason as to why I chose to do this the "ugly" way, but if not, I'd like to do it the right way 🤔

src/apps/vpn/ui/screens/home/ViewServers.qml
```diff
@@ -84,16 +84,20 @@ Item {
...
         stackContent: [
-            ServerList {
-                id: singleHopServerList
-                currentServer: {
-                    "countryCode": VPNCurrentServer.exitCountryCode,
-                    "localizedCityName": VPNCurrentServer.localizedExitCityName,
-                    "cityName": VPNCurrentServer.exitCityName,
-                    "whichHop": "singleHopServer"
+            Loader {
+                active: segmentedNav.selectedIndex == 0
```
MattLichtenstein, line 88: Don't love the comparison using index since it requires us to keep the order of the segmented nav button and view in sync, but would require a refactor/usage of enums to fix. It's also unlikely that we modify this ordering in the future, so I think it's fine here.

src/apps/vpn/ui/screens/home/ViewServers.qml
strseb: I think that was because we did have a time where multi-hop was desktop only. 
So in seg-nav you wrote `visible: stack.children.length > 1` - so if we did not add that multi-hop after complete, the whole thing was invisible.

src/apps/vpn/ui/screens/home/ViewServers.qml
strseb: Removed that, we now check the tabID :)

MattLichtenstein (approved):

PR #7774, opened 2023-08-15
VPN-4986 Create a new Connection Manager Object with relevant states

Gela:
## Description

Create a new `ConnectionManager` class and move all functionalities from Controller to this new class. This component ConnectionManager sits between the frontend and the controller to handle all connectivity related logic and let the controller only handle activation and deactivation.
[This RFC](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/rfcs/0003-rfc-controller_refactor.md) goes into more details about the purpose and implementation details. 

In subsequent PRs, some of the functionality including activate() and deactivate() will be moved back to the Controller.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-4986

## Checklist
    

src/apps/vpn/connectionmanager.h
```diff
@@ -0,0 +1,244 @@
...
+  //
+  // At initialization time, these two member variables are set to
+  // MozillaVPN::serverData() to do not let not initialize.
+  //
+  // Please, do not use MozillaVPN::serverData() in the controller!
+  ServerData m_serverData;
+  ServerData m_nextServerData;
+
+  PingHelper m_ping_canary;
+  bool m_ping_received = false;
```
vinocher, line 232: Consider using lower camel casing for the names, for consistency with the C++ code base, now that it has moved to a new file.

src/apps/vpn/connectionmanager.h
```diff
@@ -0,0 +1,244 @@
...
+#include "ipaddress.h"
+#include "loghandler.h"
+#include "models/server.h"
+#include "models/serverdata.h"
+#include "pinghelper.h"
+
+class ControllerImpl;
+class MozillaVPN;
+
+class ConnectionManager : public QObject, public LogSerializer {
```
vinocher, line 23: Consider adding a header comment describing purpose and usage of this class.

src/apps/vpn/connectionmanager.h
vinocher, line 1: The doc mentioned that this change would be behind a feature flag.

src/apps/vpn/captiveportal/captiveportaldetection.cpp
vinocher, line 1: Consider expanding the PR description to describe why this change is being done.

src/apps/vpn/connectionmanager.h
```diff
@@ -0,0 +1,244 @@
...
+  //    ConnectionStateFirewall,
+  //    ConnectionStateCaptivePortal,
+  //    ConnectionStateCheckSubscription,
+  //    ConnectionStateUnstable,
+  //    ConnectionStateIdle,       // Used when there are no active probes
+  //    ongoing
+  //  };
+  //  Q_ENUM(ConnectionState)
+
+  // These are temporarily copied from the controller state to mirror them, but
```
vinocher, line 40: Do we use TODO flags along with tracking ticket numbers in the source code to track pending work? Or is this OK as is because we are sure that this change will be made soon?

src/apps/vpn/connectionmanager.cpp
```diff
@@ -0,0 +1,1000 @@
...
+  Q_ASSERT(m_state == StateOn);
+  emit timeChanged();
+}
+
+void ConnectionManager::quit() {
+  logger.debug() << "Quitting";
+
+  if (m_state == StateInitializing || m_state == StateOff) {
+    emit readyToQuit();
+    return;
```
vinocher, line 230: Is it OK that m_nextStep is not set to Quit here? (This is old code, so just checking.)

src/apps/vpn/connectionmanager.h
Gela: I'd like to keep anything that changes functionality behind a feature flag until the work is complete but because this is simply moving the code from one class to another without changing anything, I don't think a feature flag is necessary.

src/apps/vpn/connectionmanager.h
Gela: I'm not aware of a standard process for TODOs but I'd be happy to add a TODO flag for this. Let me know if the one I added looks okay or if there's a better way to do this

src/apps/vpn/connectionmanager.cpp
Gela: m_nextStep is set to quit below this if statement. Are you asking if it should also be set to Quit within this?

vinocher: Yes, because the app's slot calls App::quit on the readyToQuit signal, and ConnectionManager's m_state is not updated to Quit because of the early return on line 230.

A change may not be necessary because the app is t, but updating the state may help to help future debugging during app shutdown. Feel free to ignore this comment.

vinocher:
> I'd like to keep anything that changes functionality behind a feature flag until the work is complete but because this is simply moving the code from one class to another without changing anything, I don't think a feature flag is necessary. 

Yes, a feature flag for a large refactor like this is not easy and usually requires the feature check to be moved to a higher level, and for both the old and new code to be retained while the feature flag  remains in place. It's a trade-off between risk mitigation and source code clutter. If we feel that risk of this change is small because it mainly involves moving code, this seems OK. Thanks.

src/apps/vpn/connectionmanager.h
vinocher: Yes, a feature flag for a large refactor like this is not easy and usually requires the feature check to be moved to a igher, and for both the old and new code to be kept while the feature flag remains in place. It's a trade-off between risk mitigation and source code clutter. If we feel that risk of this change is small because it mainly involves moving code, this seems OK. Thanks.

src/apps/vpn/connectionmanager.h
vinocher: The code base has some instances of the following pattern which references a ticket that tracks remaining work:
// TODO: Implement this once we update to Qt6.3 (VPN-3511)

src/apps/vpn/connectionmanager.cpp
```diff
@@ -0,0 +1,999 @@
...
+
+  connect(&m_connectingTimer, &QTimer::timeout, this, [this]() {
+    m_enableDisconnectInConfirming = true;
+    emit enableDisconnectInConfirmingChanged();
+  });
+
+  connect(&m_handshakeTimer, &QTimer::timeout, this,
+          &ConnectionManager::handshakeTimeout);
+
+  LogHandler::instance()->registerLogSerializer(this);
```
mcleinman, line 99: do we also need to unregister the log serializer? This PR removes code doing this in controller.cpp.

src/apps/vpn/connectionmanager.cpp
```diff
@@ -0,0 +1,999 @@
...
+  connect(LogHandler::instance(), &LogHandler::cleanupLogsNeeded, this,
+          &ConnectionManager::cleanupBackendLogs);
+
+  connect(this, &ConnectionManager::readyToServerUnavailable,
+          Tutorial::instance(), &Tutorial::stop);
+}
+
+void ConnectionManager::implInitialized(bool status, bool a_connected,
+                                        const QDateTime& connectionDate) {
+  logger.debug() << "Controller initialized with status:" << status
```
mcleinman, line 183:
```suggestion
  logger.debug() << "Connect manager initialized with status:" << status
```

?

src/apps/vpn/connectionmanager.cpp
```diff
@@ -0,0 +1,999 @@
...
+  m_serverData.setExitServerPublicKey(
+      m_activationQueue.last().m_serverPublicKey);
+
+  m_pingReceived = false;
+  m_pingCanary.start(m_activationQueue.first().m_serverIpv4AddrIn, "0.0.0.0/0");
+  logger.info() << "Canary Ping Started";
+  activateNext();
+}
+
+void ConnectionManager::clearConnectedTime() {
```
mcleinman, line 471: Out of curiosity, why did you decide to reorder a bunch of the functions from how they were in `controller.cpp`? Totally fine, I'm just genuinely curious.

src/apps/vpn/connectionmanager.cpp
```diff
@@ -0,0 +1,999 @@
...
+}
+
+void ConnectionManager::captivePortalPresent() {
+  if (m_portalDetected) {
+    return;
+  }
+  m_portalDetected = true;
+  logger.info() << "Captive-Portal Present, next activation will show prompt";
+}
+/////
```
mcleinman, line 839: Is this an artifact that should be removed?

mcleinman (commented):
This is incredible, thank you! A few comments/questions.

vinocher (approved):

src/apps/vpn/connectionmanager.cpp
Gela: yes! good catch

src/apps/vpn/connectionmanager.cpp
Gela: I created the connection manager class and started moving functions over from the controller one by one and moved all dependencies for each function as needed so it ended up being a little different in ordering. If there's code that you think should be ordered differently let me know. In hindsight I could've just copied practically all of controller code over and worked backwards  🫠

src/apps/vpn/connectionmanager.cpp
Gela: @vinocher Your logic makes a lot of sense to me so I went ahead and added your suggestion to the if statement above. If 'll double check with Naomi to make sure we're not missing something.

src/apps/vpn/connectionmanager.cpp
mcleinman: Some of the ordering feels a bit less intuitive - activate/activateNext for example, but I'm certainly not going to block this on it. Thanks.

mcleinman (approved):
👍🏻

src/apps/vpn/connectionmanager.cpp
Gela: If it helps at all, the next issue that is in progress is moving activate/deactivate and the corresponding helper back into the controller so it will be a little more intuitive soon :)

PR #7513, opened 2023-07-13
VPN-4038: Add metrics to daemon

mcleinman:
## Description

This is the iOS-side work for VPN-4038. A separate PR will be put up for Android.

This adds in the iOS network extension metrics. 

Additionally, `activationAttemptId` was no longer sent to `startTunnel` as an option, and so we had an erroneous log line regarding the source of the VPN activation. This bug was filed as https://mozilla-hub.atlassian.net/browse/VPN-5053, and is fixed here.

To test this, at the end of the `init` of `iostunnel.swift`, add one or both of these lines:
```
Glean.shared.setDebugViewTag("VPNTest") // if you want to check in the Glean debug pings viewer
Glean.shared.setLogPings(true) // if you want to test in app logs
```

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4038

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #7777, opened 2023-08-15
VPN-4038: Android daemon metrics

mcleinman:
## Description

Adding daemon metrics for Android to match [tech spec](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/184419319/VPN+session+pings+-+Tech+spec#Pings-and-metrics.1). Analogous iOS work done in: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7513

To test, you likely want to add these lines to `VPNService.kt` between `Glean.registerPings(Pings)` and `        Glean.initialize(`:
```
Glean.setDebugViewTag("VPNTest")
Glean.setLogPings(true)
```   

You'll also need to have the Super Dooper metrics feature turned on in the dev menu, and allow data collection in the app.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4038

## Checklist
    

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
```diff
@@ -96,7 +96,7 @@ class VPNServiceBinder(service: VPNService) : Binder() {
                 // [data] is empty
                 // Activate the tunnel with the last config
                 try {
-                    this.mService.reconnect()
+                    this.mService.reconnect(source = "system")
```
brizental, line 99: I don't understand why this one is "system". Would you mind clarifying?

src/apps/vpn/telemetry/metrics.yaml
```diff
@@ -256,7 +256,7 @@ session:
   daemon_session_source:
     type: string
     description: |
-      (Mobile-only) Where the session started from. Should be "app", "system settings",
+      (Mobile-only) Where the session started from. Should be "app", "system",
       or "autostart on boot".
```
brizental, line 260: How about "autostart on boot" for Android?

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
mcleinman: It comes from `ACTIONS.reactivate`. The only place I see this used (from a global search on `reactivate`) is in `VPNTileService.kt`, where it's used when activating the VPN from the settings quickaction tile, which is outside the app.

src/apps/vpn/telemetry/metrics.yaml
mcleinman: Good call, done.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
brizental: Hum, this is an interesting case. We implemented the quicksettings tile, it's not really system setting IIUC. Should we maybe add a new key here e.g. "quicksettings"? IIRC there is a similar thing we can do on iOS so it may be necessary over there as well in the future.

src/apps/vpn/telemetry/metrics.yaml
mcleinman: I think I actually want to remove “autostart on boot” from a potential description. (If I recall my research from a month ago, it was nearly impossible to figure out the difference between “autostart on boot” and “started from system settings” programatically. It may have been possible with a lot of grossness, but with “anything outside the app” being the important piece, it seems like the “system”/“app” split is the most important piece here.)

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNServiceBinder.kt
mcleinman: The goal of the `daemon_session_source` is to figure out how many sessions are starting from the app, and how many are starting from somewhere outside the app (currently called “system”). I'm going to update the yaml file to make this more clear, and so I think keeping this just as "system" might be okay.

src/apps/vpn/telemetry/metrics.yaml
```diff
@@ -256,8 +256,9 @@ session:
   daemon_session_source:
     type: string
     description: |
-      (Mobile-only) Where the session started from. Should be "app", "system settings",
-      or "autostart on boot".
+      (Mobile-only) Where the VPN session started from, to determine how many sessions
+      start outside the app. Values should be "app" or "system". "System" covers
```
brizental, line 260: nit:
```suggestion
      start outside the app. Values should be "app" or "system". "system" covers
```

brizental (approved):

Issue #3168, opened 2022-03-25
[Linux] App exclusions not working


data-sync-user:
**VPN version:**

* 2.8~rc32 (2.8~rc32-focal1);

**Affected devices:**

* Ubuntu 20.04;

**Prerequisites:**

* Have a subscription made for Mozilla VPN;

* Have Mozilla VPN installed;

* Have Mozilla VPN opened and OFF;

**Steps to reproduce:**

1. Go to Settings ~~> Network settings~~ > App permissions;
1. Turn the App permissions feature off;
1. Check the Firefox browser;
1. Turn the VPN ON;
1. In Firefox access the[https://whatismyipaddress.com/](https://whatismyipaddress.com/|smart-link)  ;
1. Observe the displayed ip address and location;

**Expected result:**

* The real IP address is displayed;

**Actual result:**

* The VPN IP address is displayed even if the browser is checked in app permission;

**Additional notes:**

* see the attached screen record;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-1990)

data-sync-user:
➤ Betty Fleming commented:

Peter Benvenuto May want to look at design for this. Seems confusing to user.

data-sync-user:
➤ Maria Berlinger commented:

Hello Betty Fleming 
I’ll attach the logs here.

[^mozillavpn-2022-3-25_1.txt]

data-sync-user:
➤ Betty Fleming commented:

Owen Kirby logs attached…

data-sync-user:
➤ Valentina Virlics commented:

Might be fixed with this: https://mozilla-hub.atlassian.net/browse/VPN-2041 ( https://mozilla-hub.atlassian.net/browse/VPN-2041|smart-link ) .

data-sync-user:
➤ Owen Kirby commented:

So, the explanation of what’s going on in the logs is that there were no application launch events detected in the time between when the daemon was started and the activation of the VPN. Therefore, as far as the daemon was aware, Firefox was not running.

This can be caused by re-starting the daemon while the user has applications running, in which case we lose track of currently active programs. Normally users shouldn’t experience this unless they are upgrading or installing the client, as the daemon normally starts at boot.

[...]

PR #7779, opened 2023-08-16
VPN-1990: Split tunnel boot-time fixes

oskirby:
## Description
Currently, the split tunneling works great, but only if you launch the service sometime after the user login sessions are complete. If you let the daemon at boot (as is normally the case) then the we often detect user creation while systemd is still busy initializing things, often resulting in a failure to connect to the user's session bus because it doesn't exist yet. And without that connection to the user's session bus, we are unable to detect application launch events needed for split tunneling to work.

Unfortunately, there doesn't seem to be a signal that we can listen on to detect when the initialization is complete, so we work around the problem by polling the D-Bus user object until we get a state change out of the undocumented "opening" state.

This was also hampered by a bug in Qt, for which we needed to apply a workaround on certain Qt versions.

## Reference
Github issue #3168 ([VPN-1990](https://mozilla-hub.atlassian.net/browse/VPN-1990))
Qt issue [QTBUG-108822](https://bugreports.qt.io/browse/QTBUG-108822)

## Checklist
    
[VPN-1990]: https://mozilla-hub.atlassian.net/browse/VPN-1990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/daemon/dbusservice.cpp
+++ b/src/platforms/linux/daemon/dbusservice.cpp
@@ -31,6 +31,7 @@ constexpr const char* APP_STATE_BLOCKED = "blocked";
 constexpr const char* DBUS_LOGIN_SERVICE = "org.freedesktop.login1";
 constexpr const char* DBUS_LOGIN_PATH = "/org/freedesktop/login1";
 constexpr const char* DBUS_LOGIN_MANAGER = "org.freedesktop.login1.Manager";
+constexpr const char* DBUS_LOGIN_USER = "org.freedesktop.login1.User";
 
 DBusService::DBusService(QObject* parent) : Daemon(parent) {
   MZ_COUNT_CTOR(DBusService);
@@ -54,11 +55,11 @@ DBusService::DBusService(QObject* parent) : Daemon(parent) {
     logger.error() << "System bus is not connected?";
   }
   if (!bus.connect("", DBUS_LOGIN_PATH, DBUS_LOGIN_MANAGER, "UserNew", this,
-              SLOT(userCreated(uint, QDBusObjectPath)))) {
+                   SLOT(userCreated(uint, QDBusObjectPath)))) {
     logger.error() << "Failed to connect to UserNew signal";
   }
   if (!bus.connect("", DBUS_LOGIN_PATH, DBUS_LOGIN_MANAGER, "UserRemoved", this,
-              SLOT(userRemoved(uint, QDBusObjectPath)))) {
+                   SLOT(userRemoved(uint, QDBusObjectPath)))) {
     logger.error() << "Failed to connect to UserRemoved signal";
   }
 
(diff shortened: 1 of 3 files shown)


src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -50,10 +51,17 @@ DBusService::DBusService(QObject* parent) : Daemon(parent) {
...
+    logger.error() << "System bus is not connected?";
+  }
+  if (!bus.connect("", DBUS_LOGIN_PATH, DBUS_LOGIN_MANAGER, "UserNew", this,
+                   SLOT(userCreated(uint, QDBusObjectPath)))) {
+    logger.error() << "Failed to connect to UserNew signal";
+  }
+  if (!bus.connect("", DBUS_LOGIN_PATH, DBUS_LOGIN_MANAGER, "UserRemoved", this,
+                   SLOT(userRemoved(uint, QDBusObjectPath)))) {
+    logger.error() << "Failed to connect to UserRemoved signal";
+  }
```
Gela, line 64: Is it possible that more than one condition can occur (i.e. failing to connect to new user and failing to remove said user) within the same execution chain? Wondering if this should be a chain of else-ifs

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -172,19 +180,48 @@ void DBusService::userListCompleted(QDBusPendingCallWatcher* watcher) {
   if (reply.isValid()) {
     UserDataList list = reply.value();
     for (const auto& user : list) {
-      m_appTracker->userCreated(user.userid, user.path);
+      userCreated(user.userid, user.path);
```
Gela, line 183: Why is `m_appTracker->` removed here?

src/platforms/linux/daemon/dbusservice.cpp
oskirby: Realistically, none of these error cases should ever occur, this was just added for some extra verbosity while debugging in the event that we have a programming error of some kind. Basically, the system bus should always exist and the only reason for a `connect()` call to fail is because of a bug of some kind...  the bug in question that these logs pointed me towards was QTBUG-108822.

Hypothetically, if the system bus isn't connected then it is almost certain that the following `connect()` calls will fail too.

src/platforms/linux/daemon/dbusservice.cpp
oskirby: The gist of this PR is that we use a timer to poll the runtime state of the user's session and delay the app-tracker setup until the state settles, however, I only wanted to write that polling logic once so I put it into `DBusService::userCreated()` and made `DBusService::userListCompleted()` call that instead of invoking `m_appTracker->userCreated()` directly.

Note that we have two cases where we update the set of active users. When starting the daemon we fetch the list of running users by asking systemd via the `UserList` method on D-Bus, and then as new users are created and destroyed we get changes to the list of active users from the `UserNew` and `UserRemoved` signals.

Gela: Makes sense, thanks for explaining this

Gela (approved):

PR #7581, opened 2023-07-24
VPN-4712: add installation ID for iOS

mcleinman:
## Description

Android work will come in a separate PR.

To test, you likely want to add a debug tag (`Glean.shared.setDebugViewTag("VPNTest")`) after the two Glean initialization calls in Swift. Then, look at the pings [in Glean's tool](https://debug-ping-preview.firebaseapp.com/pings/VPNTest).

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4712
https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/184419319/VPN+session+pings+-+Tech+spec

## Checklist
    
- [X] My code follows the style guidelines for this project
- [X] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [X] I have performed a self review of my own code
- [X] I have commented my code PARTICULARLY in hard to understand areas
- [X] I have added thorough tests where needed

PR #7785, opened 2023-08-17
VPN-4712: Add installation ID on Android daemon

mcleinman:
## Description

iOS partner work done in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7581. This adds the Android daemon work.

Additionally, I realized that there was a bug when users removed Glean permissions - Glean would automatically clear the installation ID, but it wasn't cleared from settings. Then, if a user reactivated Glean permissions, the app Glean would report no installation ID but the daemon Glean would report the old installation ID - which would have allowed their old and new sessions to be linked, which we definitely do not want. I've added some code to fix that bug.

To test, you likely want to add these lines before the two instances of `Glean.initialize(`:
```
Glean.setDebugViewTag("VPNTest")
Glean.setLogPings(true)
```   
Then, look at the pings [in Glean's tool](https://debug-ping-preview.firebaseapp.com/pings/VPNTest).

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4712

## Checklist
    
### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
@@ -256,12 +256,13 @@ class VPNService : android.net.VpnService() {
 
         if (isSuperDooperMetricsActive) {
             val installationIdString = json.getString("installationId")
-            installationIdString?.let{
+            installationIdString?.let {
                 try {
                     val installationId = UUID.fromString(installationIdString)
                     Session.installationId.set(installationId)
                 } catch (e: Exception) {
-                    Log.e(tag, "Daemon installation ID string was not UUID")
+                    Log.e(tag, "Daemon installation ID string was not UUID:")
+                    Log.e(tag, e.toString())
                 }
             }
             Pings.daemonsession.submit(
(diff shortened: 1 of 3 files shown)


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -254,6 +255,15 @@ class VPNService : android.net.VpnService() {
         }
 
         if (isSuperDooperMetricsActive) {
+            val installationIdString = json.getString("installationId")
+            installationIdString?.let{
+                try {
+                    val installationId = UUID.fromString(installationIdString)
+                    Session.installationId.set(installationId)
+                } catch (e: Exception) {
+                    Log.e(tag, "Daemon installation ID string was not UUID")
```
brizental, line 264: nit: Let's also add the exception message to this log.

brizental (approved):

strseb (approved):

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -254,6 +255,15 @@ class VPNService : android.net.VpnService() {
         }
 
         if (isSuperDooperMetricsActive) {
+            val installationIdString = json.getString("installationId")
+            installationIdString?.let{
```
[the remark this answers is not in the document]
mcleinman: Done, thanks robot.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
mcleinman: Good call, done. Thanks.

PR #7788, opened 2023-08-17
VPN-5282 - test unauth feedback form

mozrokafor:
## Description

    automation test for VPN-5282 - unauth feedback form

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5282

### DIFF
--- a/src/apps/vpn/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
+++ b/src/apps/vpn/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
@@ -87,7 +87,7 @@ MZViewBase {
 
                     MZTextField {
                         id: emailInput
-                        objectName: "Email-Textfield"
+                        objectName: "emailTextfield"
                         verticalAlignment: Text.AlignVCenter
                         Layout.fillWidth: true
                         hasError: !MZAuthInApp.validateEmailAddress(emailInput.text)
@@ -145,7 +145,7 @@ MZViewBase {
 
             MZTextField {
                 id: subjectInput
-                objectName: "Subject-Textfield"
+                objectName: "subjectTextfield"
                 verticalAlignment: Text.AlignVCenter
                 Layout.alignment: Qt.AlignHCenter
                 Layout.fillWidth: true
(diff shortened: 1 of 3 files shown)


tests/functional/testContactUs.js
```diff
@@ -32,6 +32,26 @@ describe('Contact us view', function() {
         queries.screenGetHelp.contactSupportView.UNAUTH_USER_INPUTS.visible());
   });
 
+  it('Verify unauth feedback filling out', async () => {
```
strseb, line 35: First of: I do know this is how other test's have been written in the past - so in that context that would be an r+. Just some thoughts: 

I'm having a hard time just looking at the test understanding what the test is doing, when it passes and when it should fail. It seems to fill out text fields? Or is this test about that it's writiable if the user is not authenticated? Can we make this a bit more decreptive, so when we come back later we know what this thing is testing and what it's not. 

The only failing conditions are:
-  `USER_INFO:visible` -> That assert does not have text, why should it not be visible, (i suppose it's probably becuase are supposed to not be logged in? ) can we add some text here so we understand why that may not be visible? :) 
- Global Timeout: If we hit that, no other tests will be run, so we won't know if only this one is failing or all in the suite as you need to fix them one by one.  Should we maybe give each unit (or like only this one for start) it's own timeout, so we do not abort the whole suite?

strseb (approved):
r+wc

src/apps/vpn/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
```diff
@@ -87,7 +87,7 @@ MZViewBase {
 
                     MZTextField {
                         id: emailInput
-
+                        objectName: "Email-Textfield"
```
MattLichtenstein, line 90: nit: our object names are usually written in camelcase eg `emailTextField`

MattLichtenstein (approved):

tests/functional/testContactUs.js
mozrokafor: Will address this in a follow-up solution

PR #7792, opened 2023-08-17
VPN-4650: Recommended servers empty state

MattLichtenstein:
## Description

- Add an empty state for when there are no cached recommended servers to show when the VPN is on

## Reference

[VPN-4650: Implement empty state for recommended servers](https://mozilla-hub.atlassian.net/browse/VPN-4650)


src/apps/vpn/translations/strings.yaml
```diff
@@ -112,6 +112,12 @@ serversView:
   serverCityWithNoConnection:
     value: "%1. No connectivity"
     comment: Button label that reads out a city name whose connection is unavailable. Argument is the city name.
+  recommendedEmptyTitle: 
+    value: Turn off VPN to see list
```
flodolo, line 116: Was this shortened to fit in one line? It sound strange without articles, e.g. `Turn off VPN to see the list`.

MattLichtenstein: Thats certainly what it looks like. The strings are also seem redundant. What do you think of:

Title: Recommended servers unavailable
Description: To see recommended server locations, please turn the VPN off

flodolo (approved):

vinocher:
Isn't it confusing for the user to be told that they have to turn off the VPN to see recommended server locations? Why don't we just show the 'Checking' UI, where VPN updates the recommended server list?

MattLichtenstein:
> Isn't it confusing for the user to be told that they have to turn off the VPN to see recommended server locations? Why don't we just show the 'Checking ...' UI, where VPN updates the recommended server list?

Normally, a user does not have to turn off the VPN to see recommended servers, this is more of an edge case fix.

I'm not sure we ever got down to the root cause of why we see an [empty recommended server list](https://mozilla-hub.atlassian.net/browse/VPN-5290) because in theory there should always be some set of servers there (based on either server score or geo-IP) but we did notice that it always happened while the VPN was on, so I believe instead of investigating, we just slapped on an empty state as a catch all.

vinocher:
>Normally, a user does not have to turn off the VPN to see recommended servers, this is more of an edge case fix.

Is something like the following better, so the user understands that they are being asked to Toggle?

Label: "Toggle VPN to see list"
Sub Label:  "Turn off the VPN and then turn it on to see recommended server locations."

MattLichtenstein:
> > Normally, a user does not have to turn off the VPN to see recommended servers, this is more of an edge case fix.
> 
> Is something like the following better, so the user understands that they are being asked to Toggle?
> 
> Label: "Toggle VPN to see list" Sub Label: "Turn off the VPN and then turn it on to see recommended server locations."

If I understand correctly, you are saying we shouldn't suggest that the VPN needs to remain off to see the recommended servers (instead, that it is okay to do a quick toggle off and back on), but then I think your suggestion suffers the same since we would be implying (instead of recommending/suggesting) that the VPN needs to be turned off AND back on to see the recommended servers, which is also not true.

Side-note - I've also updated the PR description to reflect the most recent copy, though I don't think it implies anything different than the original.

vinocher (approved):

PR #7799, opened 2023-08-18
VPN-4226: Enable automatic restart of linuxdaemon

oskirby:
## Description
We are concerned that in the event that the Linux daemon crashes, it may leave the user in an unprotected state and the VPN broken. The first step in tightening this failure mode up is to enable the automatic restart of the daemon upon an abnormal exit condition.

## Reference
JIRA subtask [VPN-4226](https://mozilla-hub.atlassian.net/browse/VPN-4226)

## Checklist

[VPN-4226]: https://mozilla-hub.atlassian.net/browse/VPN-4226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/linux/mozillavpn.service.in
+++ b/linux/mozillavpn.service.in
@@ -6,6 +6,7 @@ After=modprobe@xt_cgroup.service
 [Service]
 Type=dbus
 BusName=org.mozilla.vpn.dbus
+Restart=on-failure
 ExecStart=@CMAKE_INSTALL_FULL_BINDIR@/mozillavpn linuxdaemon
 
 [Install]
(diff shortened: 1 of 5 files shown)


meepmeep22:
🦈

oskirby:
So after some testing, it seems that this works fine-ish. The only lingering concern has to do with how we clean up after ourselves with regards to DNS.

In particular, we find that when the VPN is active and a crash occurs, we don't fully restore the DNS search domains. However, the systemd-resolve daemon does a reasonably okay job of choosing a fallback server (I suspect it falls back to a hard-coded address like 8.8.8.8 or something). I think this is an acceptable compromise in the event of a crash, and it does not leave the system in a broken state (unlike MacOS, which required a watchdog process to fix DNS after a crash).

strseb (approved):
lgtm

PR #7800, opened 2023-08-18
New debug feature: Glean debug tag view

mcleinman:
## Description

This will help QA more easily test Super Dooper metrics. The classic way of doing this is hacking this into the code or using environment variables (only available on some platforms, and sometimes - like on iOS - only available when you're building from code). 

This activates the feature in the 5 instances of Glean that we use:
1. macOS/Windows/Linux
2. iOS main app
3. iOS Network Extension
4. Android main app
5. Android daemon

Steps:
1. Activating this featur 
2. Relaunching the app, making sure you're on a debug build
3. You can see the Glean pings in the [Glean Debug Ping Viewer](https://debug-ping-preview.firebaseapp.com/)

## Reference

This will help with https://mozilla-hub.atlassian.net/browse/VPN-5404

This resolves https://mozilla-hub.atlassian.net/browse/VPN-3030 and https://mozilla-hub.atlassian.net/browse/VPN-4720.

## Checklist
    
### DIFF
--- a/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
+++ b/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
@@ -44,10 +44,10 @@ object Utils {
     @SuppressLint("NewApi")
     @JvmStatic
     fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String,
-        isGleanDebugTagActive: Boolean) {
-        if (isGleanDebugTagActive) {
+        gleanDebugTag: String) {
+        if (!gleanDebugTag.isEmpty) {
             Log.i("VPNUtils", "Setting Glean debug tag.")
-            Glean.setDebugViewTag("VPNTest")
+            Glean.setDebugViewTag(gleanDebugTag)
         }
         Glean.initialize(
             applicationContext = ctx.applicationContext,
--- a/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
+++ b/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
@@ -43,7 +43,11 @@ object Utils {
     }
     @SuppressLint("NewApi")
     @JvmStatic
-    fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String) {
+    fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String, isGleanDebugTagActive: Boolean) {
+        if (isGleanDebugTagActive) {
+            Log.i("VPNUtils", "Setting Glean debug tag.")
+            Glean.setDebugViewTag("VPNTest")
+        }
         Glean.initialize(
             applicationContext = ctx.applicationContext,
             uploadEnabled = isTelemetryEnabled,
--- a/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
+++ b/android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
@@ -44,8 +44,9 @@ object Utils {
     @SuppressLint("NewApi")
     @JvmStatic
     fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String,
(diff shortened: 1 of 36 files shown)


MattLichtenstein (approved):
So much better than manually setting the environment variable 😌

Where in the code is it stipulated that you can only use this flag in a debug build?

brizental:
Note: This pull request would resolve both https://mozilla-hub.atlassian.net/browse/VPN-3030 and https://mozilla-hub.atlassian.net/browse/VPN-4720.

android/common/src/main/java/org/mozilla/firefox/qt/common/Utils.kt
```diff
@@ -43,7 +43,12 @@ object Utils {
     }
     @SuppressLint("NewApi")
     @JvmStatic
-    fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String) {
+    fun initializeGlean(ctx: Context, isTelemetryEnabled: Boolean, channel: String,
+        isGleanDebugTagActive: Boolean) {
+        if (isGleanDebugTagActive) {
+            Log.i("VPNUtils", "Setting Glean debug tag.")
+            Glean.setDebugViewTag("VPNTest")
```
brizental, line 50: I am not a big fan of pinning this to a single value. Since we have to pass a value anyways to the daemon, why not pass a string instead of a boolean?

src/apps/vpn/appfeaturelist.h
```diff
@@ -53,6 +53,13 @@ FEATURE(freeTrial,              // Feature ID
         QStringList(),          // feature dependencies
         FeatureCallback_false)  // Disabled while we rethink free trials
 
+FEATURE(gleanDebugViewTag,       // Feature ID
```
brizental, line 56: Since this is not a feature, but a configuration value it feels incorrect to use a feature flag to turn it on or off. In my opinion it makes more sense to have a setting value that holds this. And holds it as a string instead of a boolean so that each tester can set it differently.

brizental (changes requested):
I suggest a different approach here, where this value is a string input on the developer menu and doesn't require a re-initialize to be applied. That can be achieved by leveraging daemon messages both on Android and iOS. The infrastructure is already there and the `setUploadEnabled` functionality can be copied, so I believe that would not be high effort.

Having the ability to manipulate the value of the debug view tag is useful for test isolation and troubleshooting.

mcleinman:
@brizental this should be ready for another review. Thanks!

src/apps/vpn/ui/screens/getHelp/developerMenu/ViewDeveloperMenu.qml
```diff
@@ -126,6 +126,47 @@ MZViewBase {
             }
         }
 
+        MZCheckBoxRow {
+            id: checkBoxRowGleanDebugTag
+
+            Layout.fillWidth: true
+            Layout.rightMargin: MZTheme.theme.windowMargin
+            labelText: "Glean Debug View Tag"
+            subLabelText: "Glean debug view tag for testing. Must restart app to take effect."
```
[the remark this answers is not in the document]
mcleinman: It's fairly simple to make this happen for the main Glean instance (on all platforms), but we'd need to create a bunch more custom code to get the custom tag down to the iOS and Android daemon instantly, and that additional code didn't seem worthwhile to me (for future maintenance, fragility, etc. - as all we're getting out of it is a debug option that will likely be seldom used after this initial testing). With that in mind, I was balancing two options:
1) Instant implementation on main Glean instances but not daemon, or
2) Just ask for a restart

The pattern of "relaunch the app to make some debug options take effect" already exists (for staging server, etc.), it seemed in line with the rest of our debug menu to go with the second option. And it gives us the ability to put the same tag everywhere.

I'm a *big* believer that debug options should add an absolutely minimum amount of code. While Glean makes it possible to start this debug tag immediately, our existing VPN architecture doesn't - and I don't think we should change it to make a debug option start quicker.

brizental (approved):
Please assign yourself / link to the two tickets this addresses.

PR #7822, opened 2023-08-23
Remove Usage of Ad-ID

strseb:
## Description

We have introduced this for our filtered adjust - campain attribution. At v2.3
However we have not used that in 1+ years, so let's remove the permission, we can re-introduce at anytime if we ever need it.


mcleinman (approved):
👍🏻

PR #7839, opened 2023-08-25
Inspector Nit: Support Smol Screens

strseb:
## Description

Currently the detail sit's on a 3rd row, which might be tiny on small screens.
So let's have it float from the right, so it's visible like an overlay.
## Reference

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/9611612/c817aa76-782d-4e59-b71c-850857caa2ef

### DIFF
--- a/tools/inspector/views/view-ui.js
+++ b/tools/inspector/views/view-ui.js
@@ -28,18 +28,26 @@ export class ViewUi extends LitElement {
         width:100%;
     }
     aside{
-        height: 100%;
+        height: 70vh;
         max-width: 50%;
         flex-grow: 1;
         display: flex;
         flex-direction: column;
         padding: 20px;
-        border-left: solid 1px var(--lt-color-gray-900);
+        border: 1px solid var(--lt-color-gray-900);
         background: white;
-        position: sticky;
-        top: 10px;
         max-height: 92vh;
         overflow-y: auto;
+        overflow-x: auto;
+        flex: 4;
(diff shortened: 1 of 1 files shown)


tools/inspector/views/view-ui.js
```diff
@@ -28,18 +28,26 @@ export class ViewUi extends LitElement {
...
         padding: 20px;
-        border-left: solid 1px var(--lt-color-gray-900);
+        border: 1px solid var(--lt-color-gray-900);
         background: white;
-        position: sticky;
-        top: 10px;
         max-height: 92vh;
         overflow-y: auto;
+        overflow-x: auto;
+        flex: 4;
```
vinocher, line 42: Can't flex: 4; and flex-grow: 1; be removed?

tools/inspector/views/view-ui.js
```diff
@@ -28,18 +28,26 @@ export class ViewUi extends LitElement {
...
         flex-direction: column;
         padding: 20px;
-        border-left: solid 1px var(--lt-color-gray-900);
+        border: 1px solid var(--lt-color-gray-900);
         background: white;
-        position: sticky;
-        top: 10px;
         max-height: 92vh;
         overflow-y: auto;
+        overflow-x: auto;
```
vinocher, line 41: Should this be combined into overflow: auto?

lesleyjanenorton (approved):
🤩

PR #7843, opened 2023-08-27
VPN-2705 Users should have a way to opt out of default on data collection even before subscribing.

Gela:
## Description

Users should have a way to opt out of default on data collection even before subscribing. 
This PR adds the same "opt out" checkbox under App Preferences in the pre-login menu.

Screenshot 2023-08-26 at 6 38 22 PM

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-2705

## Checklist
    
### DIFF
--- a/nebula/ui/components/CMakeLists.txt
+++ b/nebula/ui/components/CMakeLists.txt
@@ -21,6 +21,7 @@ qt_add_qml_module(components
          MZCard.qml
          MZCheckBox.qml
          MZCheckBoxRow.qml
+         MZCheckBoxRowWithoutDivider.qml
          MZChevron.qml
          MZClickableRow.qml
          MZCollapsibleCard.qml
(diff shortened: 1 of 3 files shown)


MattLichtenstein (changes requested):
I agree with the concept of exposing this option pre-authentication, but I think we should get UX involved with what we want that to look like, and where we should put it

lesleyjanenorton (commented):
I support providing the opt out prior to authentication and appreciate you driving this forward but I agree with Matt's comments that we should loop in UX for some guidance here.
cc @peterbenvenuto Is this something you have bandwidth to support?

peterbenvenuto:
Hey! Thanks for the tag. I think this would be a good thing to bring up during our next design sync when Santiago is back! If this is somethigg we want to do, it would probably be a feature request and would require some design work + a ticket. Let's being it up next week, thanks!

Gela:
closing this until we finalize a design

PR #7844, opened 2023-08-27
Fix build error

Gela:


src/apps/vpn/connectionbenchmark/benchmarktasktransfer.cpp
```diff
@@ -44,7 +44,8 @@ void BenchmarkTaskTransfer::handleState(BenchmarkTask::State state) {
   if (state == BenchmarkTask::StateActive) {
 #if defined(MZ_DUMMY) || defined(MZ_ANDROID) || defined(MZ_WASM)
 #  if QT_VERSION >= 0x060500
-#    error Check if QT added support for QDnsLookup::lookup() on Android
+    logger.error()
+        << "Check if QT added support for QDnsLookup::lookup() on Android";
```
brizental, line 48: I think this was intentional to actually crash everything, @strseb ?? 

How about we instead check if Qt added support in Android and fix the code?

brizental (changes requested):

strseb (changes requested):
Keep the error but restrict it to
```
#if defined(MZ_ANDROID)
```
unless that was already checked :D

PR #7850, opened 2023-08-28
VPN-4831: Set Accessibility names of ViewServers tabs and fix other Select Location a11y issues

vinocher:
## Description

The ViewServers tabs (Recommended and All) did not have accessible names. Fixed by providing names in ViewServers.qml.
Some parent elements had names which were not useful, so have been marked as ignored.

## Reference

 [VPN-4831](https://mozilla-hub.atlassian.net/browse/VPN-4831)

## Checklist
    
[VPN-4831]: https://mozilla-hub.atlassian.net/browse/VPN-4831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/MZRadioDelegate.qml
+++ b/nebula/ui/components/MZRadioDelegate.qml
@@ -13,6 +13,7 @@ RadioDelegate {
     property bool isHoverable: true
     property var radioButtonLabelText
     property string accessibleName
+    property bool isRadioButtonLabelAccessible: true
     property var uiState: MZTheme.theme.uiState
     readonly property int labelX: radioButton.anchors.margins + radioButton.implicitWidth + radioButtonLabel.anchors.leftMargin
 
@@ -114,7 +115,7 @@ RadioDelegate {
 
     MZRadioButtonLabel {
         id: radioButtonLabel
-
+        Accessible.ignored: !isRadioButtonLabelAccessible
         anchors.verticalCenter: parent.verticalCenter
         anchors.left: radioButton.right
         anchors.right: parent.right
--- a/src/apps/vpn/ui/screens/home/ViewServers.qml
+++ b/src/apps/vpn/ui/screens/home/ViewServers.qml
@@ -16,7 +16,7 @@ import compat 0.1
 Item {
     id: root
     objectName: "viewServers"
-
+    Accessible.name: qsTrId("vpn.servers.selectLocation")
 
     MZMenu {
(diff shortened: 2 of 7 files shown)


nebula/ui/components/MZRadioDelegate.qml
```diff
@@ -13,6 +13,7 @@ RadioDelegate {
     property bool isHoverable: true
     property var radioButtonLabelText
     property string accessibleName
+    property bool isRadioButtonLabelAccessible: true
```
MattLichtenstein, line 16: Seems like a good spot to use an alias since we are just essentially exporting/exposing the `Accessible.ignored` property of the label

src/apps/vpn/ui/screens/home/ViewServers.qml
```diff
@@ -16,7 +16,7 @@ import compat 0.1
 Item {
     id: root
     objectName: "viewServers"
-
+    Accessible.name: qsTrId("vpn.servers.selectLocation")
```
MattLichtenstein, line 19: When does this get read?

MattLichtenstein (approved):

src/apps/vpn/ui/screens/home/ViewServers.qml
vinocher: Read when the screen reader shortcuts are used to move to the previous or next item, and that navigation reaches the 'Select Location' text.

src/apps/vpn/ui/screens/home/servers/ServerList.qml
```diff
@@ -77,8 +77,6 @@ FocusScope {
 
     Layout.fillWidth: true
     Layout.fillHeight: true
-    Accessible.name: menu.title
-    Accessible.role: Accessible.List
```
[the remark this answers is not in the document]
vinocher: It read  'Recommended' or 'All' list, but this is in the wrong place. It has been moved to MZTabNavigation.stack, where the list is actually implemented.

nebula/ui/components/MZRadioDelegate.qml
MattLichtenstein: Disregard this comment - you cannot alias to an attached property (like `Accessible.ignored`)

PR #7878, opened 2023-08-31
VPN-5452: Decrease desktop window height

MattLichtenstein:
## Description

- Decrease the height of the desktop window to 640

*Note: in draft while I go through app UI to ensure no regressions / layout issues

## Reference

[VPN-5452: Change size of desktop VPN dimensions to 360x640](https://mozilla-hub.atlassian.net/browse/VPN-5452)


lesleyjanenorton (commented):
@peterbenvenuto  The research cited in VPN-5452 is for websites and browser windows are easily resizable so I'm not sure the research applies evenly to applications, particularly ours since ours cannot be resized. I will obviously defer to your judgment here but this makes me nervous. 
@MattLichtenstein Mind adding a few screenshots of the views where we typically run into vertical visual real estate issues (ie the server lists)?

peterbenvenuto:
@lesleyjanenorton @MattLichtenstein I think the article was more pointing out the percentages of the population that user a certain screen size, I was intending to choose a window size that resonated with one of the higher percentages closest to the screen size we currently have. Right now the screen is way too tall, and support also mentioned people complained when they had a higher screen resolution that it was getting cut. We're also removing the expandable cards for some of these views with the implementation of the in context help popovers.

mcleinman (approved):
👍🏻

PR #7885, opened 2023-08-31
VPN-5321 follow up - show error screen if no connection at start of speed test

mcleinman:
## Description

Per comment on VPN-5321, the original work neglected to catch if the connection was already gone (but the VPN tunnel was active) when the user started their speed test.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5321

## Checklist
    
### DIFF
--- a/src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
+++ b/src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
@@ -129,7 +129,7 @@ void ConnectionBenchmark::start() {
 
   // Need this check to show error screen if there is no connection when
   // starting test.
-  checkStability();
+  showErrorIfNoSignal();
 }
 
 void ConnectionBenchmark::stop() {
@@ -223,10 +223,10 @@ void ConnectionBenchmark::handleStabilityChange() {
     return;
   }
   logger.debug() << "Handle stability change";
-  checkStability();
+  showErrorIfNoSignal();
 }
 
-void ConnectionBenchmark::checkStability() {
+void ConnectionBenchmark::showErrorIfNoSignal() {
   ConnectionHealth::ConnectionStability stability =
       MozillaVPN::instance()->connectionHealth()->stability();
   logger.debug() << "Current stability: " << stability;
--- a/src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
+++ b/src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
@@ -126,6 +126,10 @@ void ConnectionBenchmark::start() {
     m_benchmarkTasks.append(uploadTask);
     TaskScheduler::scheduleTask(uploadTask);
   }
+
+  // Need this check to show error screen if there is no connection when
+  // starting test.
+  checkStability();
 }
 
 void ConnectionBenchmark::stop() {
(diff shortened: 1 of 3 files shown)


src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -218,11 +222,14 @@ void ConnectionBenchmark::handleStabilityChange() {
   if (m_state == StateInitial || m_state == StateReady) {
     return;
   }
+  logger.debug() << "Handle stability change";
```
Gela, line 225: Move the log statement above the if/return

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
```diff
@@ -218,11 +222,14 @@ void ConnectionBenchmark::handleStabilityChange() {
...
+  checkStability();
+}
 
+void ConnectionBenchmark::checkStability() {
```
Gela, line 229: Could this be renamed to something different? Current naming of this function confuses me, I would expect a return value (such as `ConnectionHealth::ConnectionStability` or even `bool`) because it is checking something but it doesn't return anything.

Gela (changes requested):
Is there a functional test for this already? If not, could you add one?

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
mcleinman: This would be a change for this function, as it previously behaved similarly (before I broke it into two functions) - and breaks the pattern shared between it and `handleControllerState`. The function called a lot, and so I think I like keeping it as "only log if there is an update". Can you explain why you think we should change the logging here?

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
mcleinman: 100% great call out, thank you. What do you think about `showErrorIfNoSignal`? It's kinda clunky, open to suggestions or talking through options.

mcleinman:
> Is there a functional test for this already? If not, could you add one?

Thank you for the reminder! Working on this now.

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
Gela: My justification for moving the log statement to the first line would be that if the `if` statement returns then we wouldn't know that we entered that function. With that being said, I'm unaware of how much it is being called so if you feel like it would clutter up the logging, I'm totally fine with leaving it as is.

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
Gela: `showErrorIfNoSignal()` sounds great, thank you!

mcleinman:
> > Is there a functional test for this already? If not, could you add one?
> 
> Thank you for the reminder! Working on this now.

@Gela functional test now in here, thank you

src/apps/vpn/connectionbenchmark/connectionbenchmark.cpp
mcleinman: Thanks - I'm inclined to leave it as it was originally.

mcleinman:
I believe all comments (as of now) have been addressed.

Gela (approved):
Awesome, test looks great 🥳

Issue #286, opened 2020-11-23
Compile mozillaVPN for macOS supporting universal binaries (x86 + apple silicon)


bakulf:
https://bugreports.qt.io/browse/QTBUG-85279

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-82)
┆author: Data Sync

data-sync-user:
➤ Santiago Andrigo commented:

Is this really a bug or a feature request? What is obtained by this?

data-sync-user:
➤ Owen Kirby commented:

This would improve network performance on the modern Apple macbooks based on the M1 and M2 chips.

jensschaefer:
+1

strseb:
Done :)

yungchin:
Thanks for all the hard work on this! Can I ask: should this now just work when I fetch the installer from https://www.mozilla.org/en-GB/products/vpn/download/ -- if I understand correctly, all the PRs linked above should be included in the 2.16.0 tag, and I think that's the build I'm getting, but this still prompts me to install Rosetta?

strseb:
The binary is universal, however it seems we forgot to mark the installer as universal too 😓 - Sorry about that.

yungchin:
Ah super, thanks for looking into it so quickly, and for opening a new issue 🙏

PR #7913, opened 2023-09-05
VPN-5473 Mark the Macos PKG as universal

strseb:
## Description

We've been publishing a universal binary since 2.16 - However we totally forgot to update the installer.  This still causes rosetta to be prompted, even though it's not needed anymore q_q (See https://github.com/mozilla-mobile/mozilla-vpn-client/issues/286#issuecomment-1703937187) 
According to the docs [here](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW7) we need to mark the cpu_arches our pkg supports. And according to [this blog post](https://scriptingosx.com/2020/12/platform-support-in-macos-installer-packages-pkg/), we can add arm :)

### DIFF
--- a/macos/pkg/Distribution
+++ b/macos/pkg/Distribution
@@ -4,7 +4,7 @@
     <background mime-type="image/png" file="background.png" alignment="left"/>
     <welcome file="welcome.html" mime-type="text/html" />
     <conclusion file="conclusion.html" mime-type="text/html" />
-    <options customize="never" allow-external-scripts="no"/>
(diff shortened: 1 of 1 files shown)


mcleinman (approved):
👍🏻

mcleinman:
Can we get a ticket for this, mostly so we can ask QA to double check this?

PR #7921, opened 2023-09-06
Make the Inspector check a User Setting

strseb:
## Description
Right now we only enable the inspector when the Guardian API is anything but the "prod" one. 
This makes debugging prod issues annoying and also stops any attempt to build test's that actually test against prod :) 

Let's make this a setting in the dev menu - and have the ability to enable this via compiler flags, so i can build a client version that can test against prod servers :)


brizental (approved):
Fancy 💅

mcleinman (approved):
👍🏻

PR #7930, opened 2023-09-07
VPN-5097: Desktop onboarding flow

MattLichtenstein:
## Description

- Implements new onboarding flow for desktop platforms

*Note: This feature is hidden behind the "New onboarding" feature flag, which is off by default

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/5f0d2625-03b5-4f24-b5ce-3a30a3d2be07

## Reference

[VPN-5097: Implement Desktop onboarding flow](https://mozilla-hub.atlassian.net/browse/VPN-5097)


mcleinman:
I'm currently reviewing this, about halfway done. Will finish it up int he morning.

src/apps/vpn/translations/strings.yaml
```diff
@@ -561,6 +561,23 @@ onboarding:
   progressBarAccessibilityStepCurrent:
     value: "Onboarding step “%1” (%2 of %3). You are currently on this step."
     comment: Accessibility read out for the current step in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
+  dataSlideHeader: Data collection and use
+  dataSlideBody: We strive to provide you with choices and collect only the technical data we need to improve Mozilla VPN. Sharing data with Mozilla is optional.
+  dataSlideCheckboxLabel: Mozilla may collect technical data about my use of Mozilla VPN.
+  dataSlideLearnMoreCaption: Learn more about what data Mozilla collects and how it's used
```
flodolo, line 567: Please check the linter.

flodolo (approved):
(with the apostrophe fixed)

src/mozillavpn.cpp
```diff
@@ -394,13 +394,23 @@ void MozillaVPN::maybeStateMain() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
 
 #if !defined(MZ_ANDROID) && !defined(MZ_IOS)
-  if (!settingsHolder->postAuthenticationShown()) {
+  if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
+    if (!settingsHolder->onboardingCompleted()) {
+      if (!settingsHolder->onboardingStarted()) {
+        settingsHolder->setGleanEnabled(false);
```
vinocher, line 400: This line could use a comment that this is because the user hasn't opted into telemetry yet.

src/mozillavpn.cpp
```diff
@@ -394,13 +394,23 @@ void MozillaVPN::maybeStateMain() {
...
+    }
+  } else if (!settingsHolder->postAuthenticationShown()) {
     setState(StatePostAuthentication);
     return;
   }
 #endif
 
-  if (!settingsHolder->telemetryPolicyShown()) {
+  if (!settingsHolder->telemetryPolicyShown() &&
+      !Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
```
vinocher, line 413: May be more readable if this feature check was moved outside the settingsHolder->telemetryPolicyShown() check:

```if (!Feature::get(Feature::Feature_newOnboarding)->isSupported())
{
    Old code.
}```

File: src/mozillavpn.cpp [hunk 4]
```diff
@@ -429,6 +439,9 @@ void MozillaVPN::maybeStateMain() {
   maybeRegenerateDeviceKey();
 
   if (state() != StateUpdateRequired) {
+    // All users who get to StateMain (home screen) should never see onboarding
+    // in the future
+    settingsHolder->setOnboardingCompleted(true);
```
vinocher, line 444: Just wondering, is there a way to easily test the new onboarding flow without resetting the app?

src/mozillavpn.cpp
```diff
@@ -896,10 +909,20 @@ void MozillaVPN::mainWindowLoaded() {
...
-  logger.debug() << "telemetry policy completed";
-
+void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
+
+  if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
+#if !defined(MZ_ANDROID) && !defined(MZ_IOS)
+    logger.debug() << "onboarding completed";
+    // If new onboarding is turned off in the future, don't make them go through
+    // old onboarding (post auth + telemetry)
```
vinocher, line 919: This comment is a bit confusing to me. If the new onboarding is turned off, won't the feature check on line 915 prevent this block from being executed?

src/ui/screens/settings/ViewPrivacy.qml
```diff
@@ -54,61 +54,27 @@ MZViewBase {
...
-                    } else {
-                        dnsProviderFlags |= modelData.settingValue;
-                    }
 
-                    // We are not changing anything interesting for the privacy/dns dialog.
-                    if (MZSettings.dnsProviderFlags !== MZSettings.Custom) {
-                        MZSettings.dnsProviderFlags = dnsProviderFlags;
-                        return;
-                    }
+        PrivacyFeaturesList {
```
vinocher, line 58: Is this file still used, now that privacy/ViewPrivacy.qml has been introduced?

src/translations/strings.yaml
```diff
@@ -557,10 +557,29 @@ onboarding:
     comment: Label for onboarding progress bar step. It's the final step in the progress, and indicates that the user can now start using the app.
   progressBarAccessibilityStepComplete:
     value: "Onboarding step “%1” (%2 of %3). Select to go back to this step."
-    comment: Accessibility read out for the completed steps in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
+    comment: Accessibility read out for a completed steps in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
```
Gela, line 560:
```suggestion
    comment: Accessibility read out for completed steps in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
```

src/ui/screens/ScreenOnboarding.qml
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
+import QtQuick.Controls 2.14
```
Gela, line 6:
```suggestion
import QtQuick.Controls 2.15
```

Not sure how important it is but I think we can use 2.15

src/ui/screens/ScreenOnboarding.qml
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```
2.15?

src/ui/screens/ScreenOnboarding.qml
```diff
@@ -0,0 +1,43 @@
...
+        function onPlayingChanged() {
+           if (MZTutorial.playing) {
+               getStack().pop(null, StackView.Immediate)
+           }
+        }
+    }
+    Connections {
+        target: window
+
+        // TODO - This only works when we are already in ScreenHome
```
Gela, line 30: Is there a followup issue for this TODO? Or will it be implemented as part of this effort?

src/ui/screens/ScreenOnboarding.qml
```diff
@@ -0,0 +1,43 @@
...
+        function onShowServerList() {
+
+            // Don't push another server view if it's already open
+            if (getStack().currentItem.objectName === "viewServers") { return; }
```
Gela, line 34: Nit: readability

```suggestion
            const currentViewName = getStack().currentItem.objectName;
            if (currentViewName=== "viewServers") { 
              return; 
            }
```

src/ui/screens/onboarding/OnboardingDataSlide.qml
```diff
@@ -0,0 +1,105 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/onboarding/OnboardingDevicesSlide.qml
```diff
@@ -0,0 +1,208 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/onboarding/OnboardingDevicesSlide.qml
```diff
@@ -0,0 +1,208 @@
...
+                    segmentButtonId: "tabApple"
+                }
+            }
+
+            Component.onCompleted: {
+                switch(Qt.platform.os) {
+                    case "osx":
+                    case "ios":
+                        selectedIndex = 1
+                        break
```
Gela, line 89: Should `iPadOS` fall under the same SelectedIndex?

src/ui/screens/onboarding/OnboardingDevicesSlide.qml
```diff
@@ -0,0 +1,208 @@
...
+        Layout.leftMargin: MZTheme.theme.windowMargin * 2
+        Layout.rightMargin: MZTheme.theme.windowMargin * 2
+        Layout.bottomMargin: MZTheme.theme.windowMargin
+        Layout.fillWidth: true
+
+        implicitHeight: MZTheme.theme.rowHeight
+
+        labelText: MZI18n.GlobalGoBack
+
+        onClicked: root.backClicked()
```
Gela, line 206:
```suggestion
        onClicked: backClicked()
```

I think since `root` is in the same scope you shouldn't need to explicitly call it here

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
```diff
@@ -0,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
```diff
@@ -0,0 +1,78 @@
...
+        Layout.fillWidth: true
+    }
+
+    Item {
+        Layout.fillHeight: true
+        Layout.minimumHeight: 20
+    }
+
+    MZButton {
+        Layout.leftMargin: MZTheme.theme.windowMargin * 2
```
Gela, line 55:
```suggestion
        id: nextButton
        Layout.leftMargin: MZTheme.theme.windowMargin * 2
```

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
```diff
@@ -0,0 +1,78 @@
...
+        Layout.rightMargin: MZTheme.theme.windowMargin * 2
+        Layout.fillWidth: true
+
+        width: undefined
+        text: MZI18n.GlobalNext
+
+        onClicked: root.nextClicked()
```
Gela, line 62:
```suggestion
        onClicked: nextClicked()
```

src/ui/screens/onboarding/OnboardingStartSlide.qml
```diff
@@ -0,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/onboarding/OnboardingStartSlide.qml
```diff
@@ -0,0 +1,90 @@
...
+        Layout.leftMargin: MZTheme.theme.windowMargin * 2
+        Layout.rightMargin: MZTheme.theme.windowMargin * 2
+        Layout.bottomMargin: MZTheme.theme.windowMargin
+        Layout.fillWidth: true
+
+        implicitHeight: MZTheme.theme.rowHeight
+
+        labelText: MZI18n.GlobalGoBack
+
+        onClicked: root.backClicked()
```
Gela, line 88:
```suggestion
        onClicked: backClicked()
```

src/ui/screens/onboarding/ViewOnboarding.qml
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/onboarding/ViewOnboarding.qml
```diff
@@ -0,0 +1,54 @@
...
+import QtQuick.Controls 2.14
```
Gela, line 6:
```suggestion
import QtQuick.Controls 2.15
```

src/ui/screens/onboarding/ViewOnboarding.qml
```diff
@@ -0,0 +1,54 @@
...
+import components 0.1
+
+MZStepNavigation {
+    id: stepNav
+
+    anchors.top: parent.top
+    anchors.left: parent.left
+    anchors.right: parent.right
+    anchors.bottom: parent.bottom
+    anchors.topMargin: MZTheme.theme.vSpacing
```
Gela, line 20:
```suggestion
anchors {
        top: parent.top
        left: parent.left
        right: parent.right
        bottom: parent.bottom
        topMargin: MZTheme.theme.vSpacing
    }
```

src/ui/screens/settings/ViewPrivacy.qml
```diff
@@ -54,61 +54,27 @@ MZViewBase {
...
-                    settingDescription: MZI18n.SettingsPrivacyMalwareBody,
-                }
-            ];
+//            Connections {
+//                target: privacyFeatureList
+//                function onSettingClicked(dnsProviderFlags, active) {
+//                    privacyOverwriteLoader.dnsProviderValue = dnsProviderFlags;
+//                    privacyOverwriteLoader.active = active;
+//                }
+//            }
```
Gela, line 75: Remove commented out code?

src/ui/screens/settings/ViewPrivacy.qml
Gela: Wondering the same thing

src/ui/screens/settings/privacy/PrivacyFeaturesList.qml
```diff
@@ -0,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
```
Gela, line 5:
```suggestion
import QtQuick 2.15
```

src/ui/screens/settings/privacy/ViewPrivacy.qml
```diff
@@ -0,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
```
Gela, line 7:
```suggestion
import QtQuick 2.15
import QtQuick.Controls 2.15
import QtQuick.Layouts 1.15
```

src/ui/screens/settings/privacy/ViewPrivacy.qml
```diff
@@ -0,0 +1,115 @@
...
+            Layout.leftMargin: 16
+            Layout.rightMargin: 16
+            Layout.fillWidth: true
+
+            onSettingClicked: (dnsProviderFlags, active) => {
+                privacyOverwriteLoader.dnsProviderValue = dnsProviderFlags;
+                privacyOverwriteLoader.active = active;
+            }
+        }
+
```
Gela, line 69:
```suggestion
```

Gela (changes requested):

Gela (commented):
This is so cool, love the new onboarding flow!

src/translations/strings.yaml
```diff
@@ -955,6 +974,12 @@ global:
   clear: 
     value: Clear
     comment: Accessible label for clear button
+  next: 
+    value: Next
+    comment: Label for button used to proceed
+  getStarted:
+    value: Get started
+    comment: Label for a button use to begin something / proceed
```
flodolo, line 982:
```suggestion
    comment: Label for a button used to begin something / proceed
```

nebula/ui/components/MZCheckBoxRow.qml
```diff
@@ -60,10 +60,22 @@ RowLayout {
...
+            Layout.topMargin: label.visible || (!label.visible && lineCount > 1) ? 0 : 2 //accounts for the topMargin of the MZCheckbox
+
+            MZMouseArea {
+                enabled: !label.visible
+                anchors.fill: undefined
+                width: parent.implicitWidth
+                height: parent.implicitHeight
+                propagateClickToParent: false
+                onClicked: checkBox.clicked()
+            }
```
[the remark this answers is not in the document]
MattLichtenstein: Previously, all of our checkboxes had a label and a sublabel, but the checkboxes in the onboarding mocks only have one. Since the label of these checkboxes in onboarding look more akin to the sublabel of our regular checkboxes, I opted to hide the regular label and just show the sublabel, and added a mouse area so clicking on the sublabel (like clicking on the label of all other checkboxes) checks/unchecks the box. Normally our sublabels aren't clickable. This was discussed with @peterbenvenuto 

Alternatively, we could remove this extra mouse area and modify properties of the label to get the look of the sublabel (font size, color, line height, etc)

src/mozillavpn.cpp
MattLichtenstein: you can sign out and use the inspector command `set_setting onboardingCompleted false`

src/mozillavpn.cpp
MattLichtenstein: We're making sure that when the new onboarding is completed, we also set the old onboarding (aka `postAuthenticationShown`) to completed, that way if the feature is turned off at some point, the user doesn't see the old onboarding.

Does that make sense?

src/ui/screens/settings/ViewPrivacy.qml
MattLichtenstein: Nope! Rebase error, good catch.

nebula/ui/components/MZSegmentedToggle.qml
```diff
@@ -139,48 +140,68 @@ Rectangle {
...
+                    id: textLabel
+                    
+                    Text {
+                        id: text
+                        
+                        anchors.centerIn: parent
+                        text: MZI18n[segmentLabelStringId]
+                        horizontalAlignment: Text.AlignHCenter
+                        verticalAlignment: Text.AlignVCenter
+                        // Bug VPN-2158 - Apparently QText is not rendered on android
```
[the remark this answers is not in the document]
MattLichtenstein: This comment was added as part of the solution to [VPN-2158](https://mozilla-hub.atlassian.net/browse/VPN-2158), so it's just a reference

src/ui/screens/ScreenOnboarding.qml
MattLichtenstein: Ahhh, this was a bad copy and paste (from `ScreenHome`) this code is unnecessary in `ScreenOnboarding`

src/ui/screens/ScreenOnboarding.qml
MattLichtenstein: code was removed 🙂

src/ui/screens/onboarding/OnboardingDevicesSlide.qml
MattLichtenstein: unfortunately, Qt does not differentiate between iOS and iPadOS - [source](https://doc.qt.io/qt-6/qml-qtqml-qt.html#platform-prop)

But it's about time that they should!

src/ui/screens/onboarding/OnboardingDevicesSlide.qml
MattLichtenstein: Although we would be able to reference `backClicked` without using the parent's id, [QML coding conventions](https://doc.qt.io/qt-6/qml-codingconventions.html#unqualified-access) recommend that we do so

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
MattLichtenstein: I don't normally provide id's to components unless they need to be referenced, are you suggesting otherwise?

nebula/ui/components/MZCheckBoxRow.qml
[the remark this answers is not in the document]
vinocher: Thanks for the explanation. Since label already has the capability for visiblibility to be turned off (visible: !!labelText.length), this looks fine I think.

nebula/ui/components/MZCheckBoxRow.qml
```diff
@@ -60,10 +60,22 @@ RowLayout {
         MZTextBlock {
             id: subLabel
 
+            font.pixelSize: label.visible ? MZTheme.theme.fontSizeSmall : MZTheme.theme.fontSize
             Layout.fillWidth: true
             text: subLabelText
             visible: !!subLabelText.length
             wrapMode: Text.WordWrap
+            verticalAlignment: label.visible ? Text.AlignTop : Text.AlignVCenter
+            Layout.topMargin: label.visible || (!label.visible && lineCount > 1) ? 0 : 2 //accounts for the topMargin of the MZCheckbox
```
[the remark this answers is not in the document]
MattLichtenstein: > Or is it OK to use literals directly? 

We don't exactly have a convention for this yet. I am of the belief that it is okay to use literals if the property is not tightly coupled with another property, so in this case, I think it makes sense to add a property to theme.js. There's a lot of instances where we do things like use `windowMargins` anytime we need 16px of padding, which doesn't provide much value over using a literal, as in the event we need to change `windowMargins` everywhere, it's going to end up having unintended consequences.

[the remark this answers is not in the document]
vinocher: Sounds good for now. Would it be useful to switch to theme based settings for these in the future to handle scenarios like large fonts?

src/shared/app.h
```diff
@@ -50,6 +50,10 @@ class App : public QObject {
     // The user needs to accept the telemetry policy.
     StateTelemetryPolicy,
 
+    // The user was successfully authenticated and moves into the onboarding
+    // flow
+    StateOnboarding,
```
Gela, line 55: Does this need a corresponding state for the frontend? For example in ControllerImage.qml we account for other states and what the UI should do. This might be different so just checking

src/ui/screens/onboarding/OnboardingDataSlide.qml
```diff
@@ -0,0 +1,105 @@
...
+        Layout.leftMargin: MZTheme.theme.windowMargin * 2
+        Layout.rightMargin: MZTheme.theme.windowMargin * 2
+        Layout.fillWidth: true
+
+        text: MZI18n.OnboardingDataSlideHeader
+        horizontalAlignment: Text.AlignLeft
+    }
+
+    MZInterLabel {
+        Layout.topMargin: 8
```
Gela, line 29: Are absolute margin values okay or should these be relative? I see a few different values throughout this implementation.

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
Gela: From my past experiences I thought for better readability we should always add `id` but I just did some Googling and that definitely doesn't seem to be the standard so please disregard.

Gela (commented):
Thanks for addressing the comments. The changes look great. Can we include a functional test or two to cover this new feature?

Here are a few workflows we should consider testing:
- New user signs in/up for the first time and successfully goes through all onboarding steps and lands on the main controller screen
- New user moves a step through onboarding and goes back and forward again. Check to make sure the selections stay consistent (e.g. if user opts out of data tracking and goes forward and back, that selection stays consistent)

src/shared/app.h
MattLichtenstein: Hmm, not sure what you mean by "corresponding state for the frontend". We use this app state to tell the navigator to load `ScreenOnboarding.qml` once we are in `StateOnboarding`, but we aren't changing any UI elsewhere based on being in the onboarding state.

src/ui/screens/onboarding/OnboardingDataSlide.qml
MattLichtenstein: if by "relative", you mean "should we be using references to values in `theme.js`", then you can find some more info on this in a [discussion between Vinoo and I above](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7930#discussion_r1320401173) but in essence, I think literal values are fine if they aren't coupled/associated with anything else. This is just spacing between components on a page and isn't really re-usable in any way. 

If you meant something else, let me know!

nebula/ui/components/MZStepProgressBar.qml
```diff
@@ -70,10 +72,22 @@ Item {
...
 
-                onClicked: activeIndex = index
+                onClicked: {
+                    activeIndex = index
+                    buttonClicked(index)
+                }
 
-                accessibleName: currentState === MZStepProgressBarDelegate.State.Complete ? MZI18n.OnboardingProgressBarAccessibilityStepComplete.arg(labelText).arg(index + 1).arg(progressBar.model.count)
-                                                                                          : MZI18n.OnboardingProgressBarAccessibilityStepCurrent.arg(labelText).arg(index + 1).arg(progressBar.model.count)
+                accessibleName: {
```
vinocher, line 80: On Windows when I use the screen reader next command (CapsLock+RightArrow in Narrator), it doesn't go through the incomplete steps. Since a sighted user can see the incomplete steps, a screen reader user should also be able to access it using the next command. (The next command is different from TAB, which moves to the next focusable item.) This can be addressed in a future PR if needed.

src/mozillavpn.cpp
vinocher: Thanks for the explanation. Rephrasing the comment to something like the following may help the reader:
"Mark the old onboarding expereince as completed as well, ensuring that users do not have to go through it if the new onboaring feature is turned off"

vinocher (approved):
New experience looks great!

Gela (approved):

PR #7936, opened 2023-09-08
[PART 2] VPN-5181 - Flatten /shared folder

brizental:
This is a reaplication of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7936 on top of main and the second (and final) part of the large restructuring under VPN-5181.

Full disclosure I did change a bit the approach from that PR. Essentially I keep the shared-sources cmake file to make transition easier.


brizental:
Probe scraper failure needs to be fixed upstream as well. Sending a PR...

src/cmake/shared-sources.cmake
```diff
@@ -0,0 +1,268 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Library of sources shared between app and tests
+add_library(shared-sources INTERFACE)
```
Gela, line 6: Why do we have a shared cmake and sources.cmake as opposed to combining all in one `sources.cmake`?

src/telemetry/pings.yaml
```diff
@@ -9,6 +9,19 @@
...
+    A simple ping to send glean events generated by the Mozilla Cuddlefish apps.
+  include_client_id: true
+  send_if_empty: false
+  bugs:
+    - https://bugzilla.mozilla.org/show_bug.cgi?id=1708637
+  data_reviews:
+    - https://bugzilla.mozilla.org/show_bug.cgi?id=1711123
+  notification_emails:
+    - amarchesini@mozilla.com
+    - vpn-telemetry@mozilla.com
```
Gela, line 23: Why was this added?

src/translations/strings.yaml
```diff
@@ -852,6 +852,7 @@ inAppMessaging:
   searchBarPlaceholderText: Search messages
   emptyStateTitle: No messages
   emptyStateDescription: You’re up to date!
+  dateTimeYesterday: Yesterday
```
Gela, line 855: wondering also why this was added?

Gela (commented):
I have a few questions primarily to better my own understanding. Thanks for taking this on!

src/translations/strings.yaml
flodolo: It's a string from `src/shared/translations/strings.yaml`, which is removed in this PR.

src/cmake/shared-sources.cmake
brizental: This file is shared between tests and mozillavpn. If I merge the two files I'd have to include mozillavpn-sources in the tests instead. That would require a lot of changes, mostly band aids to get it to work. I think keeping this sources file separate for now is cleaner than that.

src/telemetry/pings.yaml
brizental: This is coming from src/shared/telemetry/pings.yaml

Gela (approved):

strseb (approved):

scripts/utils/generate_ts.sh
```diff
@@ -62,6 +62,7 @@ for branch in $(git branch -r | grep origin/releases); do
 
   UNFLATTENED=false
 
+  # TODO: Remove this once all branches have been updated to not have the shared/ folder.
```
lesleyjanenorton, line 65: Do we have a ticket for this? Let's create one if not and add the ticket number here please.

lesleyjanenorton (approved):

PR #7964, opened 2023-09-11
VPN-5405: turn on super dooper

mcleinman:
Ticket: https://mozilla-hub.atlassian.net/browse/VPN-5405

Per discussion w/ Santiago, we'll turn this on in advance of VPN-4578 being done. We'll do that testing during the QA period for 2.17. I'm leaving this as a feature flag in case we need to turn this back off.


oskirby (approved):
Looks good to me.

MattLichtenstein (approved):
👌

PR #7975, opened 2023-09-12
VPN-5413 - Fix Missing Notifications on SDK31

strseb:
Note: There are a lot of formatting changes, they will be gone once https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7977 has merged, so i can rebase 😅 

What is this fix?
On android sdk 31, we no longer implicit get a "allow vpn to send notifications" prompt when we create the notification channel. Therefore we need to add that to our permissions list. 
`NotificationUtil.kt` - now will check if permissions may be needed and if so will try to dispatch that to `VPNActivity`. 
It will fire the prompt and after sends a signal to`NotificationUtil`, so that it will remember that we have asked the user. 

In any case `NotificationUtil` will only prompt once, so that users are not asked repeatedly if they press no.
Note: Users can go into Settings->Notifications in the app to always change their mind.


brizental (commented):
I think you accidentally added the i18n submodule to the commit.

PR #7977, opened 2023-09-12
VPN-4161  - Kotlin: Only use one tool for lint and formatting

strseb:
## Description
We have been using 2 different tools for formatting and linting. Reviewdog's ktlint and our local ktlint job. 
They seem to now be completely diverge, making coding on android annyoing, not including the review spamming. 

i.e see https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7975 
It's formatted using `gradle ktlint` and review dog is not having it. 

So let's remove reviewdog and just use the repo-local `gradle ktlint` job. For that, I i'm also fixing the issues it's finding.


mcleinman (commented):
Do we want the "trailing comma" rule? Seems noisy, especially when there are nested parens like `InAppPurchase.kt`. IMO we could turn it off, but if you feel differently as the Android person, I'm fine with it as it currently is.

Thanks for taking this on and standardizing it.

PR #7993, opened 2023-09-12
VPN-5499: Remove unimplemented content from 2.17 messaging

MattLichtenstein:
## Description

    Describe your changes

## Reference

[VPN-5499: Remove unimplemented changes from 2.17 Update / What's new messages](https://mozilla-hub.atlassian.net/browse/VPN-5499)

### DIFF
--- a/addons/message_update_v2.17/manifest.json
+++ b/addons/message_update_v2.17/manifest.json
@@ -26,28 +26,8 @@
         "type": "ulist",
         "content": [
           {
-            "id": "l_1",
-            "content": "The ability to sign in using your Google or Apple account"
-          },
-          {
             "id": "l_2",
             "content": "A new “My Devices” screen that makes it easier to view and manage your devices"
-          },
-          {
-            "id": "l_3",
-            "content": "The ability to upgrade to an annual plan from within the app"
-          },
-          {
-            "id": "l_4",
-            "content": "Improved guidance for fixing VPN connection issues"
-          },
-          {
-            "id": "l_5",
-            "content": "Personalized suggestions for apps to exclude from VPN protection"
-          },
-          {
-            "id": "l_6",
-            "content": "More easily accessible help content throughout the app"
           }
         ]
       },    
--- a/addons/message_whats_new_v2.17/manifest.json
+++ b/addons/message_whats_new_v2.17/manifest.json
@@ -27,28 +27,8 @@
         "type": "ulist",
         "content": [
           {
-            "id": "l_1",
-            "content": "The ability to sign in using your Google or Apple account"
-          },
-          {
             "id": "l_2",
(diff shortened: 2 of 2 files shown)


flodolo:
This is really unfortunate, because we have already started translating this content, so we're going to throw away work from the community. Also, this change is happening after the expected string freeze.

How can we avoid repeating this going forward?

brizental (approved):

brizental:
> How can we avoid repeating this going forward?

If we see this is as a useful pattern going forward we could potentially implement a flag that prevent strings from getting extracted until we are sure they will make it into the release? Feels like overkill and like we should just _not_ do this going forward, but just in case.

flodolo:
> we could potentially implement a flag that prevent strings from getting extracted

Don't we already have that?

> "translatable": false

brizental:
> Don't we already have that?

Ah! 🤦‍♀️

flodolo (approved):
Please coordinate with content and make sure we use the exact same phrasing for the 2.18 update, if we end up including these features. That way, we can at least recover something from translation memories.

MattLichtenstein:
> This is really unfortunate, because we have already started translating this content, so we're going to throw away work from the community. Also, this change is happening after the expected string freeze.

This was the plan all along, but we hand't considered the wasteful community contributions. We are experimenting with our release process in a way that puts our feature/fix freeze and string freeze on the same day, therefore it wasn't known what features would make it into the release so we "over-prepared" (by an unreasonable margin at that... since most of those features are unstarted/nowhere near completion) with the message content, knowing we would have to take some of the strings out. 

> Please coordinate with content and make sure we use the exact same phrasing for the 2.18 update, if we end up including these features. That way, we can at least recover something from translation memories.

This is the current plan - are there any other downsides to this?

flodolo:
@MattLichtenstein 
Yet another problem of this approach :-( 

We need this fix on the release branch
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/releases/2.17.0/addons/message_whats_new_v2.17/manifest.json

MattLichtenstein:
> @MattLichtenstein 
> 
> Yet another problem of this approach :-( 
> 
> 
> 
> We need this fix on the release branch
> 
> https://github.com/mozilla-mobile/mozilla-vpn-client/blob/releases/2.17.0/addons/message_whats_new_v2.17/manifest.json

This is fine, since we build addons from the main branch.

flodolo:
> > @MattLichtenstein
> > Yet another problem of this approach :-(
> > We need this fix on the release branch
> > https://github.com/mozilla-mobile/mozilla-vpn-client/blob/releases/2.17.0/addons/message_whats_new_v2.17/manifest.json
> 
> This is fine, since we build addons from the main branch.

That's not the problem: localization extracts strings from all the branches, which means these strings that are not used are still being translated, until we removed them from all release branches.

PR #8009, opened 2023-09-13
another taskcluster run

mcleinman:
We need a new android build, as Taskcluster busyness meant we got a set of release builds that were out of order. This causes errors when uploading to the Google Play Console.

Re-running those taskcluster jobs did not result in artifacts with new version numbers, so the problem continues. This PR is just to trigger a new run for Android artifacts.

Screenshot 7
(image is from last time we had this issue, but it's the same message)


MattLichtenstein (approved):
Looks great, nice job 😆

mcleinman:
Merged w/o waiting for tests because this is a 0 line change. Feel free to let me know if you think this was not a good practice, even in this case.

PR #8022, opened 2023-09-14
VPN-5540 Remove obsolete permission

strseb:
## Description

Apparently we've had this since ... [forever](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/v2.0.1/android/AndroidManifest.xml#L10) ? So this is probably [a thing we just got by default from qt](https://sourcegraph.com/github.com/qt/qtbase@6.2.4/-/blob/src/corelib/CMakeLists.txt?L364) when we first generated this file. 

Just checked, the only thing we export to an external storage are the logfiles (so we can get a share intent without giving other apps access to our fs ) - that still works, given we use mediastore which does not need additional permissions, not surprising.


brizental (approved):

Gela (approved):

mcleinman (approved):
Thanks!

PR #8026, opened 2023-09-14
[Uplift] Fix 2.17 messaging

MattLichtenstein:
## Description

- Content parity with main branch for the 2.17 update and what's new messages

*Note: Translations consider all branches, so even though addons are built off of `main`, these strings would still be unnecessarily translated in the `releases/2.17.0` branch if they aren't removed. 

## Reference

N/A


flodolo (approved):

mcleinman (approved):
👍🏻

PR #8073, opened 2023-09-19
Fix Wireguard-Go Goroot

strseb:
## Description
To properly work in xcode - we should capture the env variable during configuration, cache it and then use it once we build.


Gela (approved):

scripts/cmake/golang.cmake
```diff
@@ -5,6 +5,12 @@
 ## Find the absolute path to the go build tool.
 find_program(GOLANG_BUILD_TOOL NAMES go REQUIRED)
 
+# Capture this during configuration and re-use that 
+# during build, as we might not have the same env variables.
+# for example when building in XCode. 
+set(GOLANG_GOROOT $ENV{GOROOT} CACHE STRING "Path to GOROOT variable")
+
+
```
MattLichtenstein, line 13:
```suggestion

```

MattLichtenstein (approved):

PR #8104, opened 2023-09-22
Fix VPN-5587: Prompt web download for mac users on v2.16

lesleyjanenorton:
## Description

This PR updates `message_update_v2.17` to prompt mac users on v2.16 to update via the web-based download/install flow, instead of the traditional Balrog/in-app flow. 

## Reference

VPN-5587 and related VPN-5537, VPN-5538

## Checklist
    
### DIFF
--- a/addons/message_update_v2.17/enable.js
+++ b/addons/message_update_v2.17/enable.js
@@ -6,6 +6,15 @@
     api.addon.date = (api.settings.updateTime.getTime() / 1000);
   }
 
+
+  // Macos v2.16.0 requires a web-based update.
+  if (api.env.platform === 'macos' && api.env.versionString === '2.16.0') {
+    api.addon.setTitle(
+        'message.message_update_v2.17.block.extra_1_217',
+        'Download the new Mozilla VPN');
+    return api.addon.composer.remove('c_3');
+  }
+
   // Windows v2.10 to v2.12 do require a web-based update,
   // with the exception on v2.11.1.
   if (api.env.platform !== 'windows' || api.env.versionString === '2.11.1') {
@@ -41,5 +50,6 @@
   api.addon.composer.remove('c_3');
 
   api.addon.setTitle(
-    'message.message_update_v2.17.block.extra_1_217', 'Download the new Mozilla VPN')
+      'message.message_update_v2.17.block.extra_1_217',
+      'Download the new Mozilla VPN');
 })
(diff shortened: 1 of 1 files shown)


addons/message_update_v2.17/enable.js
```diff
@@ -6,6 +6,15 @@
     api.addon.date = (api.settings.updateTime.getTime() / 1000);
   }
 
+
+  // Macos v2.16.0 requires a web-based update.
+  if (api.env.platform === 'macos' && api.env.versionString === '2.16.0') {
+    api.addon.setTitle(
+        'message.message_update_v2.17.block.extra_1_217',
+        'Download the new Mozilla VPN');
+    return api.addon.composer.remove('c_3');
```
MattLichtenstein, line 15: should we just return nothing here to be consistent with the rest of the file?

MattLichtenstein (approved):

addons/message_update_v2.17/enable.js
```diff
@@ -6,6 +6,16 @@
     api.addon.date = (api.settings.updateTime.getTime() / 1000);
   }
 
+
+  // Macos v2.16.0 requires a web-based update.
+  if (api.env.platform === 'macos' && api.env.versionString === '2.16.0') {
+    api.addon.setTitle(
+        'message.message_update_v2.17.block.extra_1_217',
+        'Download the new Mozilla VPN');
```
vinocher, line 14: Doesn't this 'Download the new Mozilla VPN' string need to be localized?

addons/message_update_v2.17/enable.js
```diff
@@ -6,6 +6,16 @@
     api.addon.date = (api.settings.updateTime.getTime() / 1000);
   }
 
+
+  // Macos v2.16.0 requires a web-based update.
```
vinocher, line 10: Explaining the reason for this in the comment may be useful to the reader.

addons/message_update_v2.17/enable.js
MattLichtenstein: It's translated via the manifest [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/b65a21c4761369e790b275fc85a2a28e6eec85dc/addons/message_update_v2.17/manifest.json#L58) (also see the comment at the top of `enable.js`)

addons/message_update_v2.17/enable.js
lesleyjanenorton: I'm sorry @vinocher! I'm saw this right as I merged. Will send up a follow up.

PR #8133, opened 2023-09-26
VPN-5312 Introduce a on/off boolean in Controller to replace State::On and State::Off

Gela:
## Description

- Replace `State::StateOn` and `State::StateOff` with a function `isVPNActive()` that returns a bool indicating whether or not the VPN is active.
- Repurpose `StateOn` to `State::StateIdle`. This is the "resting" state the VPN is in if it is On and isn't any particular state such to perform tasks such as checking subscription, checking network connectivity, captive portal, etc. This is the "I'm on but I'm not doing any checks" state.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5312

## Checklist
    
### DIFF
--- a/src/connectionmanager.cpp
+++ b/src/connectionmanager.cpp
@@ -150,8 +150,6 @@ void ConnectionManager::initialize() {
 
   connect(SettingsHolder::instance(), &SettingsHolder::transactionBegan, this,
           [this]() {
-            //            m_connectedBeforeTransaction =
-            //                m_state == ConnectionManager::StateOn;
             m_connectedBeforeTransaction =
                 m_state ==
                 MozillaVPN::instance()->connectionManager()->isVPNActive();
@@ -194,18 +192,15 @@ void ConnectionManager::implInitialized(bool status, bool a_connected,
 
   if (!status) {
     REPORTERROR(ErrorHandler::ControllerError, "controller");
-    //    setState(StateOff);
     deactivateVPN();
     return;
   }
 
   if (processNextStep()) {
-    //    setState(StateOff);
     deactivateVPN();
     return;
   }
 
-  //  setState(a_connected ? StateOn : StateOff);
   if (a_connected) {
     setState(StateIdle);
     activateVPN();
@@ -233,7 +228,6 @@ qint64 ConnectionManager::time() const {
 }
 
 void ConnectionManager::timerTimeout() {
-//  Q_ASSERT(m_state == StateIdle);
 #ifdef MZ_IOS
   // When locking an iOS device with an app in the foreground, the app's JS
   // runtime is stopped pretty quick by the system. For this VPN app, that
@@ -246,7 +240,6 @@ void ConnectionManager::timerTimeout() {
   }
 #else
 
-  //    Q_ASSERT(m_state == StateIdle);
   ///@TODO This call to activateVPN should be removed.
   activateVPN();
   //  Q_ASSERT(isVPNActive());
@@ -258,7 +251,6 @@ void ConnectionManager::timerTimeout() {
 void ConnectionManager::quit() {
   logger.debug() << "Quitting";
 
-  //  if (m_state == StateInitializing || m_state == StateOff) {
   if (m_state == StateInitializing || !isVPNActive()) {
     m_nextStep = Quit;
     emit readyToQuit();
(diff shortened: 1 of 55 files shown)


src/connectionmanager.cpp
```diff
@@ -1010,6 +1030,11 @@ bool ConnectionManager::deactivate() {
...
 
   Q_ASSERT(m_impl);
+  deactivateVPN();
   m_impl->deactivate(stateToReason(m_state));
   return true;
 }
+
+bool ConnectionManager::isVPNActive() { return m_VPNActive; }
+void ConnectionManager::activateVPN() { m_VPNActive = true; }
+void ConnectionManager::deactivateVPN() { m_VPNActive = false; }
```
MattLichtenstein, line 1040: we'll want to emit `isVPNActiveChanged()` anywhere `m_VPNActive` is mutated

src/connectionmanager.h
```diff
@@ -56,7 +56,7 @@ class ConnectionManager : public QObject, public LogSerializer {
     StateCheckSubscription,
     StateConnecting,
     StateConfirming,
-    StateOn,
+    StateIdle,
```
vinocher, line 59: StateIdle seems like a confusing name because it doesn't indicate that the VPN is on. Isn't the original StateOn name better?

tests/functional/testConnection.js
```diff
@@ -30,16 +30,19 @@ describe('Connectivity', function() {
...
-          queries.screenHome.CONTROLLER_TITLE, 'text');
-      return connectingMsg === 'Connecting…';
-    });
-
-    assert.equal(
-        await vpn.getQueryProperty(
-            queries.screenHome.CONTROLLER_SUBTITLE, 'text'),
-        'Masking connection and location');
+    // This bit is commented out because the UI has slightly changed 
+    // with the work in VPN-5312 and as a result the Connecting screen doesn't show.
```
vinocher, line 34: Is VPN-5312 the correct ticket? VPN-5312 is the ticket that tracks the change in this PR.

src/connectionmanager.cpp
```diff
@@ -233,24 +238,30 @@ void ConnectionManager::timerTimeout() {
     emit timeChanged();
   }
 #else
+
+  ///@TODO This call to activateVPN should be removed.
+//  activateVPN();
+  //  Q_ASSERT(isVPNActive());
+
```
vinocher, line 245: Should this commented code be removed?

src/ui/screens/home/controller/ControllerImage.qml
```diff
@@ -176,9 +174,7 @@ Rectangle {
         },
         State {
             name: "noSignalOn"
-            when: (VPNController.state === VPNController.StateOn ||
-                   VPNController.state === VPNController.StateSilentSwitching) &&
-                VPNConnectionHealth.stability === VPNConnectionHealth.NoSignal
+            when: VPNController.isVPNActive && VPNController.state != VPNController.StateSilentSwitching && VPNConnectionHealth.stability === VPNConnectionHealth.NoSignal
```
MattLichtenstein, line 177: why did we go from needing to be on OR silent switching to needing to be on AND NOT silent switching

src/connectionmanager.cpp
vinocher, line 1040: Can't activateVPN and deactivateVPN be removed isVPNActive be changed to return the active state based on the value of m_state? Otherwise we are storing the state in two different places (in m_state and in m_VPNActive) with the risk of inconsistency in the future when the code evolves.

MattLichtenstein:
> Repurpose StateOn to State::StateIdle. This is the "resting" state the VPN is in if it is On and isn't any particular state such to perform tasks such as checking subscription, checking network connectivity, captive portal, etc. This is the "I'm on but I'm not doing any checks" state.

Question: Am I understanding correctly that we plan to remove `StateOff` at some point? If so, what state will we be in when the VPN is off?

MattLichtenstein (changes requested):

tests/functional/testConnection.js
Gela: Yes, it's referring to the work in the same PR. The plan is to have a front end engineer tidy up the transitions and then we can re-enable the 'connecting' bits

Gela:
Closing this as we are going a different route

PR #8141, opened 2023-09-27
VPN-5145: Implement Mobile onboarding flow

MattLichtenstein:
## Description

- Implements the onboarding flow for iOS and Android
  - Different "Start" slide than desktop which prompts the user to allow/disallow adding the VPN configuration to the system
  - Various UI layout changes to support the other cross-platform slides
- Moves onboarding after device limit screen ([see comment](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8141/files#diff-a131ac478dadc38997760e67ddb60a6657adc7fb3790e77542c3bce63b659268R405))
- Replace old onboarding QR codes with styled QR codes

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/0a572fea-2889-4a50-ac7d-ffe15782ed16

## Reference

[VPN-5145: Implement Mobile onboarding flow](https://mozilla-hub.atlassian.net/browse/VPN-5145)
[VPN-5631: [a11y] Users are able to move forward through the new onboarding flow while using the progress bar when the Screen reader is enabled ](https://mozilla-hub.atlassian.net/browse/VPN-5631)


src/translations/strings.yaml
```diff
@@ -555,8 +555,8 @@ onboarding:
   progressBarAddDevices:
     value: Devices
     comment: Label for onboarding progress bar step
-  progressBarGetStarted:
-    value: Start
+  progressBarGetStarted2:
+    value: Startup
```
flodolo, line 559: The new label feels quite confusing: any explanation from content about the reason for this change?

src/translations/strings.yaml
```diff
@@ -576,13 +576,17 @@ onboarding:
   privacySlideBody: Use these features for more protection. They may cause issues on some sites, so you can turn them off anytime in settings.
   devicesSlideHeader: Install Mozilla VPN on up to 5 devices
   devicesSlideBody: Scan this QR code using a tablet or mobile device’s camera to install the VPN on another device.
+  devicesSlideBodyMobile: Scan this QR code using a tablet or mobile device’s camera to install the VPN on another device. To install on a desktop computer, visit the Download page on the Mozilla VPN website.
```
flodolo, line 579: > tablet or mobile device’s camera 

Non-native speaker: shouldn't both have the genitive?

src/translations/strings.yaml
traceyvantyghem: @flodolo Agreed. I think this was copied over from the desktop onboarding, which has as the last step "Startup" and prompts users to turn on "Connect VPN on startup". 

This is a totally different step, though. It seems like a better label here might be "Connect". Thoughts? (If @flodolo is ok with this change, @MattLichtenstein is there an issue with making this update?)

src/translations/strings.yaml
traceyvantyghem: @flodolo Well, the genitives look ok to me, but I think this sentence could be a bit more direct.

What about "To install the VPN on another tablet or mobile device, scan the QR code."?

src/translations/strings.yaml
flodolo: `Connect` works for me, there's no space for interpretation

src/translations/strings.yaml
flodolo: Works fine for me 👍🏼

src/translations/strings.yaml
MattLichtenstein: No strong opinions here, but for a bit of background: 

The step was originally labelled "Get started" to denote it was the last step. This eventually got shortened to "Start" and it seems like that meaning may have gotten lost in translation when we switched the final slide on desktops content to being about connecting on startup (instead of quick access), leading to the final step being called "Startup." This now has a different meaning than the intended "Get started" and I agree it doesn't make much sense on mobile. I prefer "Connect" but also don't mind going back to "Start" if we wanted to be consistent across platforms (which I don't think is absolutely necessary, since the content IS different, just another option)

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
```diff
@@ -41,15 +41,14 @@ ColumnLayout {
     }
 
     PrivacyFeaturesList {
-        Layout.topMargin: 24
+        Layout.topMargin: Qt.platform.os === "android" || Qt.platform.os === "ios" ? 32 : 24
```
mcleinman, line 44: Is there a utility function available for `Qt.platform.isMobile`? If not, is it worth making one?

mcleinman, line 44: Is there a utility function available for `Qt.platform.isMobile`? If not, is it worth making one?

src/mozillavpn.cpp
```diff
@@ -407,14 +416,14 @@ void MozillaVPN::maybeStateMain() {
         settingsHolder->setGleanEnabled(false);
```
mcleinman, line 416: Is this bug being resolved in another ticket?

src/mozillavpn.cpp
```diff
@@ -396,7 +395,17 @@ void MozillaVPN::maybeStateMain() {
...
 
-#if !defined(MZ_ANDROID) && !defined(MZ_IOS)
+  if (!m_private->m_deviceModel.hasCurrentDevice(keys())) {
+    Q_ASSERT(m_private->m_deviceModel.activeDevices() ==
+             m_private->m_user.maxDevices());
+    setState(StateDeviceLimit);
+    return;
+  }
+
+  // Onboarding needs to come after device limit because on mobile users can
```
mcleinman, line 405: Thanks for this comment, super helpful for future maintenance.

src/mozillavpn.cpp
```diff
@@ -134,7 +133,7 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
   connect(this, &MozillaVPN::stateChanged, [this]() {
     // If we are activating the app, let's initialize the controller and the
     // periodic tasks.
-    if (state() == StateMain) {
+    if (state() == StateMain || state() == StateOnboarding) {
```
mcleinman, line 136: Why this update? Because we need to be able to start the connection during onboarding now? Why didn't we need this for desktop, out of curiousity?

src/platforms/ios/ioscontroller.swift
```diff
@@ -188,10 +188,21 @@ public class IOSControllerImpl : NSObject {
                 if let error = saveError {
                     IOSControllerImpl.logger.error(message: "Connect Tunnel Save Error: \(error)")
                     failureCallback()
+                    if isOnboarding {
+                        IOSControllerImpl.logger.error(message: "Finishing onboarding...")
```
mcleinman, line 192:
```suggestion
                        IOSControllerImpl.logger.error(message: "Finishing onboarding, but saving tunnel resulted in error...")
```

src/platforms/ios/ioscontroller.swift
```diff
@@ -188,10 +188,21 @@ public class IOSControllerImpl : NSObject {
...
+                        onboardingCompletedCallback()
+                    }
                     return
                 }
 
                 IOSControllerImpl.logger.info(message: "Saving the tunnel succeeded")
+                
+                //Used to create a VPN configuration without connecting during onboarding
+                if isOnboarding {
+                    IOSControllerImpl.logger.error(message: "Finishing onboarding... do not turn on the VPN after gaining permission")
```
mcleinman, line 202: Should this (and the above one) be logged as an error or another level?

android/vpnClient/src/main/java/org/mozilla/firefox/vpn/qt/VPNActivity.java
```diff
@@ -162,6 +162,7 @@ private void initServiceConnection(){
   private final int ACTION_NOTIFICATION_PROMPT_SENT = 19;
   private final int EVENT_PERMISSION_REQURED = 6;
   private final int EVENT_DISCONNECTED = 2;
+  private final int EVENT_CANCELLED = 9;
```
mcleinman, line 165: `cancelled` for this (in several files) almost makes it sound like the user cancelled something. Did you consider something like `EVENT_ONBOARDING_COMPLETE` (that's a little long, but hopefully you get my gist)?

mcleinman (commented):
Just a few questions, mostly so I better understand this. Thanks! It's looking great!

src/mozillavpn.cpp
MattLichtenstein: [VPN-5520: All pre-onboarding glean data being deleted on first sign in](https://mozilla-hub.atlassian.net/browse/VPN-5520)

src/mozillavpn.cpp
MattLichtenstein: With the current implementation, we need the connection manager initialized in onboarding as a means to present the "Add VPN configuration" system modal (starting with `ConnectionManager::activate()`). We do not have this requirement on desktop.

src/platforms/ios/ioscontroller.swift
MattLichtenstein: Do you think this clarification is still needed after the error log 3 lines above that says "Connect Tunnel Save Error"?

android/vpnClient/src/main/java/org/mozilla/firefox/vpn/qt/VPNActivity.java
MattLichtenstein: my original reasoning was to make the naming for all of this generic enough so that it made sense if the reasoning was something other than onboarding in the future. as you can see, I eventually went against that on the iOS side, so I will make it more consistent on Android as well

src/translations/strings.yaml
traceyvantyghem: I think, since this really is a different step from the desktop onboarding, it's better to be more specific here than consistent across platforms. (My understanding is that these labels are acting as navigation labels for screen readers, and "Connect" is definitely a more clear and accurate label for this step than "Start").

src/translations/strings.yaml
traceyvantyghem: @MattLichtenstein sorry for the delay - yes, let's update on desktop too. Thank you!

src/mozillavpn.cpp
mcleinman: We decided not to block on this, right?

src/ui/screens/onboarding/OnboardingPrivacySlide.qml
mcleinman: Thanks!

src/platforms/ios/ioscontroller.swift
mcleinman: Personally, I think it makes it even clearer when a future person is skimming logs, but I don't feel strongly about that at all - I'm okay if you want to keep it as is.

src/mozillavpn.cpp
mcleinman: Makes sense, thanks.

mcleinman (changes requested):
After logging in on iOS, this is the first thing I see:
IMG_D30F33317E6B-1

MattLichtenstein:
> After logging in on iOS, this is the first thing I see:
> 
> IMG_D30F33317E6B-1
> 
> 

This onboarding experience is still behind the `newOnboarding` feature flag

src/mozillavpn.cpp
MattLichtenstein: We are blocking turning the feature flag on, on that ticket

mcleinman:
> > After logging in on iOS, this is the first thing I see:
> > [removed big image]
> 
> This onboarding experience is behind the `newOnboarding` feature flag, which is currently off by default

But with or without the feature flag, this isn't something that should be seen on mobile, right?

MattLichtenstein:
> > > After logging in on iOS, this is the first thing I see:
> > > [removed big image]
> > 
> > 
> > This onboarding experience is behind the `newOnboarding` feature flag, which is currently off by default
> 
> But with or without the feature flag, this isn't something that should be seen on mobile, right?

🤦‍♂️ yup, totally misinterpreted the comment. now fixed

mcleinman (approved):
👍🏻 Looks great, tests great on iOS. (I had to turn on the feature manually; I believe that is expected.) Thanks!

src/translations/strings.yaml
flodolo: I think this is unclear also on Desktop. Does it mean "Start the VPN" or something else? I would have no clue how to translate this without additional context; without it, we're going to get very confusing translations.

src/translations/strings.yaml
```diff
@@ -575,14 +578,18 @@ onboarding:
   privacySlideHeader: Get more privacy
   privacySlideBody: Use these features for more protection. They may cause issues on some sites, so you can turn them off anytime in settings.
   devicesSlideHeader: Install Mozilla VPN on up to 5 devices
-  devicesSlideBody: Scan this QR code using a tablet or mobile device’s camera to install the VPN on another device.
+  devicesSlideBody2: To install the VPN on another tablet or mobile device, scan the QR code.
+  devicesSlideBodyMobile: To install the VPN on another tablet or mobile device, scan the QR code. To install on a desktop computer, visit the Download page on the Mozilla VPN website.
```
flodolo, line 582: Nitpicking, but there is no Mozilla VPN website, there's a mozilla.org website with VPN pages.

src/translations/strings.yaml
```diff
@@ -575,14 +578,18 @@ onboarding:
...
   devicesSlideDeviceTypeLabel: "Select device type:"
   devicesSlideDeviceTypeAndroid: Android
   devicesSlideDeviceTypeApple: Apple
   devicesSlideQRCodeAndroid: QR Code that links to the Mozilla VPN product page on the Google Play Store
   devicesSlideQRCodeApple: QR Code that links to the Mozilla VPN product page on the iOS App Store
   startSlideHeader: Connect Mozilla VPN on startup
   startSlideBody: Turn on to launch the VPN whenever you start your device. We’ll automatically reconnect you to the same server location you last used.
+  startSlideMobileHeader: Allow Mozilla VPN to create a VPN connection
+  startSlideMobileBody: Your mobile device will now ask you to allow us to create a VPN connection. Please allow this so that Mozilla VPN works correctly on your device.
+  startSlideApproveButtonLabel: Approve Mozilla VPN
```
flodolo, line 592: Why the switch to approve, if the terminology has been "allow" so far (and probably is in the phone)?

> Your mobile device will now ask you to allow us to create a VPN connection. Please allow this so that Mozilla VPN works correctly on your device.

I find it very confusing when software uses "we" to refer to itself. We're already repeating "allow" a lot, it doesn't seem so bad to repeat "Mozilla VPN".

> Your mobile device will now ask you to allow Mozilla VPN to create a VPN connection. Please allow this so that Mozilla VPN works correctly on your device.

flodolo (changes requested):
Mostly nitpicking on the language, but the "Startup" string needs a clarification also for desktop.

src/translations/strings.yaml
MattLichtenstein: agree with all of this! @traceyvantyghem thoughts?

src/translations/strings.yaml
traceyvantyghem: Ok, I would leave this one, because you can go to Google and search "Mozilla VPN", and you do get what looks like a dedicated site as the first hit. I know it's technically pages within the Mozilla site, but being more accurate here doesn't really seem like it'll add any value, imo.

flodolo:
Could you provide a screenshot? I feel like I'm getting more confused at this point :-(

traceyvantyghem:
> Could you provide a screenshot? I feel like I'm getting more confused at this point :-(

Sorry @flodolo - I totally misinterpreted your comment. My mistake. Matt reached out in Slack to let me know about the confusion and said we should be good on the Startup/Connect question.

flodolo:
> Sorry @flodolo - I totally misinterpreted your comment. My mistake. Matt reached out in Slack to let me know about the confusion and said we should be good on the Startup/Connect question.

No worry, the confusion is actually unrelated to your last comments. I feel like I'm not fully understanding how this label is used, and that's necessary to evaluate the comment.

MattLichtenstein:
> No worry, the confusion is actually unrelated to your last comments. I feel like I'm not fully understanding how this label is used, and that's necessary to evaluate the comment.

@flodolo I will be adding a better comments to these labels in an upcoming commit

src/translations/strings.yaml
traceyvantyghem: Agree about "Allow".

I think we could get away from using "us" here, but maybe we could avoid a little of the repetition. What about this:

**Allow Mozilla VPN to create a connection**

Your mobile device will now ask you to allow Mozilla VPN to create a connection. Please allow this so that the VPN works correctly on your device.

[Connect Mozilla VPN]

flodolo (approved):

Issue #7626, opened 2023-07-28
WindowsPingSender::pingEventReady


data-sync-user:
Sentry Issue: [VPN-CLIENT-4F](https://mozilla.sentry.io/issues/4263093822/?referrer=jira_integration)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0x0: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0x0
  File "windowspingsender.cpp", line 130, in WindowsPingSender::pingEventReady
  ?, in free_base
  File "qarraydata.h", line 160, in QTypedArrayData<T>::deallocate
  File "qarraydatapointer.h", line 131, in QArrayDataPointer<T>::{dtor}
  File "qstring.h", line 1302, in QString::{dtor}
...
(8 additional frame(s) were not displayed)
```

This ticket was automatically created by Sentry via [Open A Jira Bug for new Sentry-Crashes](https://sentry.io/organizations/mozilla/alerts/rules/vpn-client/13654141/)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5288)
┆Reporter: Basti

data-sync-user:
➤ Santiago Andrigo commented:

Basti Is this correctly a high? what’s the user impact?

data-sync-user:
➤ Basti commented:

Its our nr 4 crash on the leaderboard. But given we now know what a critical crash looks now… this is certainly not one 🙂

data-sync-user:
➤ Santiago Andrigo commented:

Ok, moving this to Medium. Feel free to adjust!

data-sync-user:
➤ Santiago Andrigo commented:

Upping this to High as it’s currently considered the #1 crash issue on Windows, as per Vinoo Cherian

PR #8161, opened 2023-09-29
VPN-5288: Delay deletion of PingSender to avoid the risk of pending events being delivered after deletion

vinocher:
## Description

Sentry reports a NULL AV in WindowsPingSender::pingEventReady around the following code, but the absence of some stack frames have made it difficult to determine the exact cause.

```
memcpy(&sequence, replies[i].Data, sizeof(quint16));
emit recvPing(sequence);
```

The pingEventReady slot  is called asynchronously when IcmpSendEcho2Ex's event has been signaled to indicate that the ping has been received. A hunch is that this event was pending in the queue when WindowsPingSender was deleted and was delivered after deletion and while the icmpCleanupHelper thread was concurrently executing. [~QObject destructor docs](https://doc.qt.io/qt-6/qobject.html#dtor.QObject) say that that all signals to the object are disconnected and pending posted events for the objects are removed from the event queue. However, it also recommends that is often safer to use deleteLater in this scenario. 

So this fix uses deleteLater to delete the PingSender, to guarantee deletion after all pending events have been delivered to it. 

This also fixes the size of the replyBuffer, which needs an additional sizeof(IO_STATUS_BLOCK), as specified in the [IcmpSendEcho2Ex docs](https://learn.microsoft.com/en-us/windows/win32/api/icmpapi/nf-icmpapi-icmpsendecho2ex).

## Reference
[VPN-5288](https://mozilla-hub.atlassian.net/browse/VPN-5288), [GitHub 7626](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7626)
## Checklist
    
[VPN-5288]: https://mozilla-hub.atlassian.net/browse/VPN-5288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pinghelper.cpp
+++ b/src/pinghelper.cpp
@@ -47,7 +47,7 @@ void PingHelper::start(const QString& serverIpv4Gateway,
   // we happen to be on one of these unlucky devices, create a DnsPingSender
   // instead.
   if (!m_pingSender->isValid()) {
-    delete m_pingSender;
+    m_pingSender->deleteLater();
     m_pingSender = new DnsPingSender(m_source, this);
   }
 
@@ -71,7 +71,7 @@ void PingHelper::stop() {
   logger.debug() << "PingHelper deactivated";
 
   if (m_pingSender) {
-    delete m_pingSender;
+    m_pingSender->deleteLater();
     m_pingSender = nullptr;
   }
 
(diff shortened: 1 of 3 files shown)


oskirby (approved):
I think this PR looks good.

Great catch on the `IO_STATUS_BLOCK`!

PR #8162, opened 2023-09-29
Only check for captive portal upon activation and deactivation

Gela:
## Description
With the current implementation, we check for captive portal in 2 instances and show the "Captive Portal Detected" :

a. When user tries to activate the VPN
b. We continue polling for the captive portal while the VPN is on

This PR changes the behavior so we no longer poll for captive portal on a regular cadence. Instead only check for captive portal and show the "Captive Portal Detected" screen in the following scenarios:

a. When user tries to activate the VPN
b. When user tries to deactivate the VPN __if__ we detect a captive portal
c. When the network changes and a captive portal is detected.

This PR makes the following changes:
1. Remove the code to exclude IP addresses.
2. Rather than continuously polling for captive portal, we only check this upon VPN activation and deactivation.
3. If user switches from a working network to one with captive portal (ex: going from cellular to a Starbucks network) which requires authentication, VPN will enter No Signal until user deactivates, in which case we perform a captive portal check and if that ends up being the issue, we will prompt the user with "CaptivePortalScreen".

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5651
    https://mozilla-hub.atlassian.net/browse/VPN-4271

## Checklist
    
- [ ] I have added thorough tests where needed -- This will be done in a subsequent PR.


src/platforms/android/androidcontroller.cpp
```diff
@@ -149,19 +149,6 @@ void AndroidController::activate(const InterfaceConfig& config,
   foreach (auto item, config.m_allowedIPAddressRanges) {
     allowedIPs.append(IPAddress(item.toString()));
   }
-  foreach (auto addr, config.m_excludedAddresses) {
```
lesleyjanenorton, line 152: Just digging in here and curious: why this change? I didn't think captive portal was enabled on Android. 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/89c56f10db8b11d0045b9da01f734929b2d19275/src/featurelistcallback.h#L59C1-L78

Gela: We are getting rid of `m_excludedAddresses` entirely and keeping this breaks the Android build

src/daemon/daemon.cpp
```diff
@@ -121,11 +121,6 @@ bool Daemon::activate(const InterfaceConfig& config) {
     }
   }
 
-  // Configure routing for excluded addresses.
-  for (const QString& i : config.m_excludedAddresses) {
-    addExclusionRoute(IPAddress(i));
```
vinocher, line 126: Should Daemon.addExclusionRoute, delExclusionRoute and m_excludedAddrSet also be removed?

src/mozillavpn.cpp
```diff
@@ -298,7 +298,6 @@ void MozillaVPN::initialize() {
   AndroidUtils::instance();
 #endif
 
-  m_private->m_captivePortalDetection.initialize();
```
vinocher, line 301: Should all the CaptivePortalDetection related code be removed?

src/daemon/daemon.cpp
Gela: I had the same question but Naomi confirmed that for now those should stay. From @oskirby:

> I think the long term goal would be to remove the exclusion routing code completely, but I think this particular function is still needed on MacOS for the Wireguard server address. The goal here is to addres issue with the smallest amount of change

vinocher:
Could you please explain how the user experience will differ before and after the change?

src/daemon/daemon.cpp
vinocher: Sure. Perhaps open a ticket to remove additional code later?

src/mozillavpn.cpp
Gela: Hmm, are you referring to the following 3?

```
  connect(&m_private->m_connectionManager, &ConnectionManager::stateChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::stateChanged);

  connect(&m_private->m_connectionHealth, &ConnectionHealth::stabilityChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::stateChanged);

  connect(SettingsHolder::instance(),
          &SettingsHolder::captivePortalAlertChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::settingsChanged);
```

src/daemon/daemon.cpp
vinocher: BTW, I think macOS and Windows use MacosRouteMonitor and WindowsRouteMontitor's addExclusionRoute for excluding the VPN server addresss, which is different from the the Daemon's addExclusionRoute.

src/mozillavpn.cpp
vinocher: Yes, and the CaptivePortalDetection class.

Gela:
> Could you please explain how the user experience will differ before and after the change?

Updated the PR description with this info, good call out!

src/mozillavpn.cpp
vinocher: Never mind, we still need these based on the PR description.

vinocher: However, can these 3 be moved into CaptivePortalDetection.initialize() and remove the call to set up a slot for NetworkWatcher::networkChange? This call can be restored then, which seems cleaner than not calling m_captivePortalDetection.initialize when CaptivePortalDetection is still used.

vinocher:
Won't the 'Guest Wi-Fi portal alert' checkbox in the AppPreferences/Notifications screen be affected by this change?

src/mozillavpn.cpp
vinocher: For my understanding, could you please explain why this line was removed? It seems like CaptivePortalDetection still needs the code that is currently in m_captivePortalDetection.initialize().

src/daemon/daemon.cpp
Gela: The following calls from `MacosRouteMonitor` 

```c++
  bool addExclusionRoute(const IPAddress& prefix);
  bool deleteExclusionRoute(const IPAddress& prefix);
  void flushExclusionRoutes();
```
are referenced in `wireguardutilsmacos` so I'm inclined to leave it as is for now. I have a follow up issue here: https://mozilla-hub.atlassian.net/browse/VPN-5649

oskirby: To shed a bit of light on this question. The exclusion routing code in `MacosRouteMonitor` and `WindowsRouteMonitor` are the platform specific logic to interface with the routing table, and the Daemon's methods are wrappers around them which provide reference counting.

The reference counting was important for doing server switching. The gist of how server switching is that we start a connection to a new peer and then drop the connection to the old peer afterwards. If the client specifies an exclusion route then the sequence of events the route monitor sees would look like:
```
Exclude captive portal (old peer)
... wireguard handshake with old peer ...
... some time passes ...
... old peer starts to fail ...
Exclude captive portal exclusion (new peer)
... wireguard handshake with new peer ...
Remove captive portal exclusion (old peer)
```

However, if we are removing the ability for the client to send an exclusion route, then I don't think the reference counting is necessary anymore, and the platform's `WireguardUtils::updatePeer()` methods can just invoke the route monitor directly as needed.

So, the long TLDR is that I think we can remove the daemon's exclusion routing methods in this PR.

src/mozillavpn.cpp
oskirby: IIUC, this is the meaty change that intends to disable the captive portal, but I don't think this is sufficient, digging into the code we find in `src/captiveportal/captiveportaldetection.cpp`:
```
void CaptivePortalDetection::initialize() {
  m_active = SettingsHolder::instance()->captivePortalAlert();
  const auto networkWatcher = MozillaVPN::instance()->networkWatcher();
  connect(networkWatcher, &NetworkWatcher::networkChange, this,
          &CaptivePortalDetection::networkChanged);
}
```

What I think we really desire is the removal of the `CaptivePortalMonitor` class which generates periodic checks.

vinocher:
How is the user's choice in _App Preferences/Notifications/Guest Wi-Fi Portal Alert_ affected after this change?

Gela:
@vinocher that checkbox is removed because it is no longer applicable. Just added this to the PR description

src/ui/screens/settings/ViewNotifications.qml
```diff
@@ -20,22 +20,6 @@ MZViewBase {
...
-            objectName: "settingCaptivePortalAlert"
-            visible: MZFeatureList.get("captivePortal").isSupported
-            //% "Guest Wi-Fi portal alert"
-            labelText: qsTrId("vpn.settings.guestWifiAlert")
-            //% "Get notified if a guest Wi-Fi portal is blocked due to VPN connection"
-            subLabelText: qsTrId("vpn.settings.guestWifiAlert.description")
-            isChecked: (MZSettings.captivePortalAlert)
-            showDivider: false
-            onClicked: {
-                MZSettings.captivePortalAlert = !MZSettings.captivePortalAlert
```
vinocher, line 34: Should the captivePortalAlert related Settings code be removed too?

vinocher:
Captive Portal is also detected when the network changes (see CaptivePortalDetection::networkChanged). Consider updating the PR description to mention that.

vinocher:
Unless there are some other dependencies, moving the following into CaptivePortalDetection::initialize improves encapsulation. But this can also be done in another change.

  ```
connect(&m_private->m_connectionManager, &ConnectionManager::stateChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::stateChanged);

  connect(&m_private->m_connectionHealth, &ConnectionHealth::stabilityChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::stateChanged);

  connect(SettingsHolder::instance(),
          &SettingsHolder::captivePortalAlertChanged,
          &m_private->m_captivePortalDetection,
          &CaptivePortalDetection::settingsChanged);
```

File: src/ui/screens/settings/ViewNotifications.qml (hunk 4 above)
Comment (vinocher) [reply src/ui/screens/settings/ViewNotifications.qml:34]: Another option is to keep this UI which controls if the ScreenCaptivePortal screen should be shown when the captive portal detection is run when the Connection manager state changes or when the network changes.

File: src/platforms/windows/daemon/wireguardutilswindows.cpp [hunk 5]
```diff
@@ -177,12 +171,6 @@ bool WireguardUtilsWindows::deletePeer(const InterfaceConfig& config) {
   QByteArray publicKey =
       QByteArray::fromBase64(qPrintable(config.m_serverPublicKey));
 
-  // Clear exclustion routes for this peer.
-  if (config.m_hopType != InterfaceConfig::MultiHopExit) {
-    m_routeMonitor.deleteExclusionRoute(IPAddress(config.m_serverIpv4AddrIn));
-    m_routeMonitor.deleteExclusionRoute(IPAddress(config.m_serverIpv6AddrIn));
-  }
-
```
vinocher, line 185: This block is needed because it excludes the VPN server from the tunnel. Same comment for the other change in this file.

src/platforms/macos/daemon/wireguardutilsmacos.cpp
```diff
@@ -180,13 +173,6 @@ bool WireguardUtilsMacos::deletePeer(const InterfaceConfig& config) {
   QByteArray publicKey =
       QByteArray::fromBase64(qPrintable(config.m_serverPublicKey));
 
-  // Clear exclustion routes for this peer.
-  if ((config.m_hopType != InterfaceConfig::MultiHopExit) &&
-      (m_rtmonitor != nullptr)) {
-    m_rtmonitor->deleteExclusionRoute(IPAddress(config.m_serverIpv4AddrIn));
-    m_rtmonitor->deleteExclusionRoute(IPAddress(config.m_serverIpv6AddrIn));
-  }
-
```
vinocher, line 189: This block is needed because it excludes the VPN server from the tunnel. Same comment for the other change in this file.

src/platforms/android/androidcontroller.cpp
```diff
@@ -149,19 +149,6 @@ void AndroidController::activate(const InterfaceConfig& config,
...
-    excludedIPs.append(IPAddress(addr));
-  }
-  if (!config.m_serverIpv4AddrIn.isEmpty()) {
-    excludedIPs.append(IPAddress(config.m_serverIpv4AddrIn));
-  }
-  if (!config.m_serverIpv6AddrIn.isEmpty()) {
-    excludedIPs.append(IPAddress(config.m_serverIpv6AddrIn));
-  }
```
vinocher, line 160: This block is needed because it excludes the VPN server from the tunnel.

src/platforms/android/androidcontroller.cpp
```diff
@@ -149,19 +149,6 @@ void AndroidController::activate(const InterfaceConfig& config,
...
-  foreach (auto item, IPAddress::excludeAddresses(allowedIPs, excludedIPs)) {
-    fullAllowedIPs.append(QJsonValue(item.toString()));
-  }
```
vinocher, line 163: This block is needed because it handles allowed IPs that go through the tunnel.

src/daemon/daemon.cpp
vinocher: Exclusion routes are needed for the VPN server, so those can't be removed.

src/daemon/daemon.cpp
```diff
@@ -210,28 +205,6 @@ bool Daemon::parseStringList(const QJsonObject& obj, const QString& name,
   return true;
 }
 
-bool Daemon::addExclusionRoute(const IPAddress& prefix) {
```
vinocher, line 213: These need to be removed from the daemon.h header too.

src/platforms/windows/daemon/windowsroutemonitor.cpp
```diff
@@ -67,7 +67,6 @@ WindowsRouteMonitor::WindowsRouteMonitor(QObject* parent) : QObject(parent) {
 WindowsRouteMonitor::~WindowsRouteMonitor() {
   MZ_COUNT_DTOR(WindowsRouteMonitor);
   CancelMibChangeNotify2(m_routeHandle);
-  flushExclusionRoutes();
```
vinocher, line 70: This line is needed because the VPN server is in the exclusion route.

src/captiveportal/captiveportaldetection.cpp
```diff
@@ -34,18 +33,23 @@ void CaptivePortalDetection::initialize() {
...
+          &ConnectionManager::stateChanged, this,
+          &CaptivePortalDetection::stateChanged);
+
+  connect(MozillaVPN::instance()->connectionHealth(),
+          &ConnectionHealth::stabilityChanged, this,
+          &CaptivePortalDetection::stateChanged);
+
+  connect(SettingsHolder::instance(),
+          &SettingsHolder::captivePortalAlertChanged, this,
+          &CaptivePortalDetection::settingsChanged);
```
vinocher, line 47: These also need to be removed from MozillaVPN::MozillaVPN

vinocher (approved):
Looks good. QA may need to test this to check if there are concerns with the new user experience.

lesleyjanenorton:
@Gela can you please sync with QA to let them know about the changes made here and help them revise their test cases?

Gela:
@lesleyjanenorton yes definitely, I will reach out to Bianca and Valentina

cure53:
✔️ LGTM

PR #8199, opened 2023-10-04
VPN-3360 Add ability to upgrade monthly plans from subscription management view

lesleyjanenorton:
## Description

This PR
- Removes VPN / Relay Bundle upgrade ephemera
- Introduces 'Upgrade to Annual' UI in the subscription management view
  - This UI should only be visible to subscribers when: 
    - The 'annualUpgrade' feature flag is enabled
    - The use has a monthly subscription
    - The user purchased their subscription through the subscription platform (not through one of the app stores)
    - MAYBE only on desktop (see open question 1) 

Waiting on the following decisions, but they do not need to block given this is behind a feature flag.

1. Can we show this upgrade UI on mobile or would that be construed as an in-app purchase?
2. What UTM parameters should be attached to the outbound upgrade URL? 

Telemetry to come in subsequent PR. 

## Reference

VPN-3360

## Checklist
    

flodolo (approved):

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
-                onClicked: {
-                    Glean.sample.bundleUpsellLinkClicked.record();
-                    MZUrlOpener.openUrlLabel("relayPremium");
-                }
-            }
+            elide: Text.ElideRight
         }
 
+
         MZButtonBase {
```
MattLichtenstein, line 181: Would it be preferable to use an `MZButton` here?

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
         Layout.bottomMargin: MZTheme.theme.windowMargin
 
-        Component.onCompleted: {
-            if (visible) {
-                Glean.sample.bundleUpsellViewed.record();
-            }
-        }
-
-        ColumnLayout {
+        MZBoldInterLabel {
```
MattLichtenstein, line 168: Label should be center aligned vertically

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -29,7 +29,7 @@ ColumnLayout {
                 visible: false
             }
             PropertyChanges {
-                target: relayUpsell
+                target: upgradeToAnnualSub
```
MattLichtenstein, line 32: I know you're just following the pattern of the file (and the pattern that is commonly used elsewhere in the app) of adding property changes for each state, but what do you think about removing these, and just adding `visible: type === "text-upgrade"` to the `upgradeToAnnualSub` RowLayout?

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
-        Component.onCompleted: {
-            if (visible) {
-                Glean.sample.bundleUpsellViewed.record();
-            }
-        }
-
-        ColumnLayout {
+        MZBoldInterLabel {
+            color: MZTheme.colors.purple70
+            horizontalAlignment: Text.AlignLeft
```
MattLichtenstein, line 170: `horizontalAlignment` of `MZBoldInterLabel` is `Text.AlignLeft` by default, no need to re-bind

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
-                Layout.alignment: Qt.AlignLeft
-                padding: 4
-                Layout.leftMargin: -4
-
-                onClicked: {
-                    Glean.sample.bundleUpsellLinkClicked.record();
-                    MZUrlOpener.openUrlLabel("relayPremium");
-                }
-            }
+            elide: Text.ElideRight
```
MattLichtenstein, line 177: No need for the `elide` property if we are expecting this label to wrap with no `maximumLineCount`

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
     }
 
     RowLayout {
```
MattLichtenstein, line 159: This whole thing should really be inside a loader, because right now we are instantiating this row for every single subscription management line item in the list. this is a bad pattern throughout this whole file

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
-
-        ColumnLayout {
+        MZBoldInterLabel {
+            color: MZTheme.colors.purple70
+            horizontalAlignment: Text.AlignLeft
+            font.pixelSize: MZTheme.theme.fontSizeSmall
+            text: MZI18n.SubscriptionManagementUpgradeToAnnual // "Save 50%..."
+            wrapMode: Text.WordWrap
+            lineHeight: 13
+            Layout.alignment: Qt.AlignLeft
```
MattLichtenstein, line 175: `Layout.alignment: Qt.AlignLeft` is not needed here

MattLichtenstein (changes requested):

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
```diff
@@ -157,71 +157,48 @@ ColumnLayout {
...
-        }
-
-        ColumnLayout {
+        MZBoldInterLabel {
+            color: MZTheme.colors.purple70
+            horizontalAlignment: Text.AlignLeft
+            font.pixelSize: MZTheme.theme.fontSizeSmall
+            text: MZI18n.SubscriptionManagementUpgradeToAnnual // "Save 50%..."
+            wrapMode: Text.WordWrap
+            lineHeight: 13
```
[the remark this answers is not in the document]
lesleyjanenorton: Yes, good catch.

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
lesleyjanenorton: We seem to be getting this for free - perhaps something to do with RowLayout {}. Do you think it's worth stipulating it anyway in case there is some discrepancy in handling between platforms?

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
lesleyjanenorton: I don't _think_ so - but mainly because we'd have to override  a lot of baked in styling in MZButton.

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
lesleyjanenorton: I'd rather leave it here for now to keep the pattern the same. Agree this file needs a trip to the cleaners though.

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
MattLichtenstein: thats because `font.pixelSize` and `lineHeight` are currently the same, so the label component is only as big as the text inside of it (no need to align), but once you [change the lineHeight ](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8199#discussion_r1348821231) to be bigger than the font, the label becomes bigger than just the text inside of it, making it necessary to vertically align

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
lesleyjanenorton: The properties of MZButton that we'd need to change would be `height`, `width`, `Layout.preferredWidth`, `Layout.preferredHeight`, the font family of the button label, and `setMargins` in MZUiStates. We also don't need the button loader.

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
lesleyjanenorton: Agree but would prefer to do any structural refactoring separately.

src/ui/screens/settings/ViewSubscriptionManagement/SubscriptionManagementItem.qml
MattLichtenstein: If you just set a `Layout.preferredWidth` and `Layout.preferredHeight` on this button instance, then `height` and `width`  will be ignored (these properties shouldn't be defined inside `MZButton` anyways)

We don't need the button loader in most of primary buttons, but it is there. It itself should be in a `Loader` (outside the scope here)

For the focus outline margins (`setMargins`), where does it say they are supposed to differ from all other primary buttons? 

On the subject of **font family** for our primary buttons - all figma mockups are using Inter Semi-Bold, whereas in the application, they are all using Metropolis Semi-Bold; [VPN-5603: Change all button labels font to Inter](https://mozilla-hub.atlassian.net/browse/VPN-5603) exists to do this work, so it being the wrong font for the time being is better, because it will at least be consistently wrong with the rest of the UI 😅 

You would have to change the button label's font pixel size (from 15 -> 13) by either exposing that property (preferred), or passing in a new `Label` for the `contentItem`

If you do decide to go the `MZButtonBase` route (which again, I don't think is the right path), we should fix:
- `font.family` - it should be `fontInterSemiBoldFamily`
-  Label `horizontalAlignment` - should be `Text.AlignHCenter
- `Accessible.name` - button label is not currently read aloud by screen readers
- I don't think the button should have `Layout.fillWidth: true` - think about if the "Save 50%" label is really short in another localization, this button will become really long
- There is a typo in the `Layout.preferredHeight` binding

lesleyjanenorton: Recap of our convo for posterity:
We have moved to MZButton but this leaves us with a few unfavorable outcomes...
  - We need to stipulate `Layout.preferredWidth` to override the default. There does not appear to be a good way to do this other than to set it to `undefined`, which produces benign console complaining and feels gross. Other explored options produced polish loops, which we definitely don't want. Matt to do a quick exploration of alternatives.
  - The move to MZButton means the button label will now elide. This seems undesirable and, because we definitely don't want to make any sweeping changes in MZButton (because it's everywhere, because it can't be feature flagged), this _may_ mean a change back to MZButtonBase.

MattLichtenstein (approved):

PR #8201, opened 2023-10-04
VPN-5559: Scroll to off screen components via keyboard nav

MattLichtenstein:
## Description

- When navigating keyboard focus to a component that is either partially or completely off screen in the downward direction, scroll the view downward so that component is fully visible (even when the navbar is not visible)

## Reference

[VPN-5559: [a11y] Some screens from the app are not scrollable when navigating via keyboard](https://mozilla-hub.atlassian.net/browse/VPN-5559)


lesleyjanenorton (approved):
💫

PR #8206, opened 2023-10-05
VPN-2859: Prevent connection benchmark results flicker

MattLichtenstein:
## Description

- Prevent flickering of the connection benchmark results when closing an in-progress connection benchmark

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/600daef4-8294-4be5-8df3-fab68cbcaabf

## Reference

[VPN-2859: Connection info results screen flashes when cancelling/closing a benchmark in progress](https://mozilla-hub.atlassian.net/browse/VPN-2859)

### DIFF
--- a/src/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
+++ b/src/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
@@ -120,7 +120,7 @@ Rectangle {
         id: connectionInfoContent
 
         opacity: visible && root.state !== "closing" ? 1 : 0
-        visible: root.state === "open-ready" /*|| root.state === "closing"*/
+        visible: root.state === "open-ready"
 
         Behavior on opacity {
             NumberAnimation {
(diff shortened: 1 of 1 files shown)


src/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
```diff
@@ -124,7 +120,7 @@ Rectangle {
         id: connectionInfoContent
 
         opacity: visible && root.state !== "closing" ? 1 : 0
-        visible: root.state === "open-ready" || root.state === "closing"
+        visible: root.state === "open-ready" /*|| root.state === "closing"*/
```
vinocher, line 123: Should the commented out code be removed?

src/ui/screens/home/controller/connectionInfo/ConnectionInfoScreen.qml
```diff
@@ -28,7 +28,7 @@ Rectangle {
             PropertyChanges {
                 target: root
                 opacity: 0
-                visible: false
+                restoreEntryValues: false
```
vinocher, line 31: If visible is not set to false, like in the old code, will this cause screen readers to read this content even when this is in closed state?

MattLichtenstein: For this particular instance, no because the children components containing screen-reader accessible items (eg `ConnectionInfoContent`, `ConnectionInfoError`) will not be visible, even though the root (`Rectangle`) is

vinocher (approved):

PR #8207, opened 2023-10-05
Miscellaneous onboarding changes

MattLichtenstein:
## Description

- [Deskop only] Don't show post authentication "quick access" screen on every login when the new onboarding experience is not enabled
- Add some padding between content and buttons on privacy slide

## Reference

[VPN-5660: [l10n] “Get more privacy” screen from new onboarding has no padding between the “Next” button and the description above](https://mozilla-hub.atlassian.net/browse/VPN-5660)

### DIFF
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -408,7 +408,7 @@ SETTING_BOOL(postAuthenticationShown,        // getter
              "postAuthenticationShown",      // key
              false,                          // default value
              true,                           // user setting
-             true,                           // remove when reset
+             false,                           // remove when reset
              false                           // sensitive (do not log)
 )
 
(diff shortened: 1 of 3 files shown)


lesleyjanenorton (approved):

src/settingslist.h
```diff
@@ -408,7 +408,7 @@ SETTING_BOOL(postAuthenticationShown,        // getter
              "postAuthenticationShown",      // key
              false,                          // default value
              true,                           // user setting
-             true,                           // remove when reset
+             false,                           // remove when reset
```
lesleyjanenorton, line 411:
```suggestion
             false,                         // remove when reset
```

lesleyjanenorton: Linter appeasement.

PR #8266, opened 2023-10-11
VPN-5643 - Ensure we always use an English backup language

mcleinman:
## Description

We were showing strings like `vpn.inAppAuth.signInSubtitle2` in some languages, but not others. Description of why follows, but the ultimate fix (done in this PR) is to always fall back to English.

- When we load a language's fallback languages, we always load English. But we don't load backup languages when the build believes the translation completeness is 1.0. Which was happening here.
- This completeness comes from a `translations.completeness` file, which we build ourselves.
- The translations.completeness counts the targets (translated string) and sources (English string) in each language's .xliff file (which comes from the translation repo), and uses that ratio to determine the completeness.
- The .xliff file usually has unstranslated strings  as a source.
- However, at the point the bug was filed .xliff didn't have the brand new strings (aka sources), and so it erroneously calculated some languages as 100% translated. (Which explains why the bug was happening only for languages that were fully translated, except for those new strings - as they reported in as 100% translated. Languages that already weren’t fully translated load fallbacks, including the final English fallback.)
- Thus, we should ensure we always fall back to English.

## To test this
1. Checkout the commit that merged the new strings (at this point, there hadn't been a roundtrip to pontoon, so the bug exists): f84b5af2bd5ba57afb1fb71f3b4278e47c697c13
2. Clean your repo (if you don't clean, it will retain the cached newer translations)
3.  run `git submodule update --recursive -f`(make sure you have the translations repo that is missing those strings)
4. Run the app
5. Set app language is Chilean Spanish
6. Most of the new strings are in the onboarding. The easiest place I found to view this problem (and the fix) is in the delete account flow (don't worry, you won't actually delete the account), on the screen where you put in the account password.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5643

## Checklist
    
### DIFF
--- a/scripts/addon/build.py
+++ b/scripts/addon/build.py
@@ -438,6 +438,11 @@ with open(args.source, "r", encoding="utf-8") as file:
                 xlifftool_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "utils", "xlifftool.py")
                 xlifftool_cmd = [sys.executable, xlifftool_path, "-C", f"--locale={locale}", xliff_path]
                 xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
+                # This completeness metric can be out-of-date, sometimes reporting a higher-than-actual completeness after a new string is
+                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denomenator is used.
+                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into pontoon (via the `extract new strings`
+                # job being run and merged into the translation repo) and then the VPN repo merging in a new translations commit.
+                # We protect against this by always loading English as a fallback language.
                 completeness_output = xlifftool.stdout.decode("utf-8")
                 completeness.append(f"{locale}:{completeness_output}")
 
--- a/scripts/addon/build.py
+++ b/scripts/addon/build.py
@@ -439,7 +439,7 @@ with open(args.source, "r", encoding="utf-8") as file:
                 xlifftool_cmd = [sys.executable, xlifftool_path, "-C", f"--locale={locale}", xliff_path]
                 xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
                 # This completeness metric can be out-of-date, sometimes reporting a higher-than-actual completeness after a new string is
-                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denomenator is used.
+                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denominator is used.
                 # To show up in the .xliff, there must be a complete cycle of a new string being pulled into pontoon (via the `extract new strings`
                 # job being run and merged into the translation repo) and then the VPN repo merging in a new translations commit.
                 # We protect against this by always loading English as a fallback language.
--- a/scripts/addon/build.py
+++ b/scripts/addon/build.py
@@ -440,7 +440,7 @@ with open(args.source, "r", encoding="utf-8") as file:
                 xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
                 # This completeness metric can be out-of-date, sometimes reporting a higher-than-actual completeness after a new string is
                 # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denominator is used.
-                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into pontoon (via the `extract new strings`
+                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into Pontoon (via the `extract new strings`
                 # job being run and merged into the translation repo) and then the VPN repo merging in a new translations commit.
                 # We protect against this by always loading English as a fallback language.
                 completeness_output = xlifftool.stdout.decode("utf-8")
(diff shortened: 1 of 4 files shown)


lesleyjanenorton (approved):
Thanks @mcleinman !

scripts/addon/build.py
```diff
@@ -438,6 +438,11 @@ def qtquery(qmake, propname):
                 xlifftool_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "utils", "xlifftool.py")
                 xlifftool_cmd = [sys.executable, xlifftool_path, "-C", f"--locale={locale}", xliff_path]
                 xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
+                # This completeness metric can be out-of-date, sometimes reporting a higher-than-actual completeness after a new string is
+                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denomenator is used.
```
flodolo, line 442:
```suggestion
                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denominator is used.
```

scripts/addon/build.py
```diff
@@ -438,6 +438,11 @@ def qtquery(qmake, propname):
...
+                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into pontoon (via the `extract new strings`
```
flodolo, line 443:
```suggestion
                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into Pontoon (via the `extract new strings`
```

flodolo (approved):
Not sure I'm really suited for this, only noting a couple of typos.

src/localizer.cpp
```diff
@@ -344,12 +344,9 @@ bool Localizer::loadLanguage(const QString& requestedLocalCode) {
     localeCode = systemLanguageCode();
   }
 
-  double completeness = m_translationCompleteness.value(localeCode, 0);
-  if (completeness < 1) {
```
MattLichtenstein, line 348: Am I understanding correctly that we are removing the check for 100% language completion and always loading our fallback languages no matter what? 

Also are there any drawbacks to loading a fallback language even when a language (i.e. english.. do we even generate a fallback language for english) is truly at 100% (including new strings)?

mcleinman: That is a correct understanding.

I think it would be pretty ugly to try to detect if there were new strings causing us to falsely report 100% - and this false 100% report was why this bug existed. I'm not aware of downsides of loading backup languages, but I'm also not well-versed in this. Do you (or anyone else looking at this ticket) know of any potential downsides that I could investigate?

scripts/addon/build.py
```diff
@@ -438,6 +438,11 @@ def qtquery(qmake, propname):
                 xlifftool_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "utils", "xlifftool.py")
                 xlifftool_cmd = [sys.executable, xlifftool_path, "-C", f"--locale={locale}", xliff_path]
                 xlifftool = subprocess.run(xlifftool_cmd, capture_output=True)
+                # This completeness metric can be out-of-date, sometimes reporting a higher-than-actual completeness after a new string is
+                # added to the app. This happens because the .xliff file doesn't have all the source strings, so an incorrect denominator is used.
+                # To show up in the .xliff, there must be a complete cycle of a new string being pulled into Pontoon (via the `extract new strings`
```
bakulf, line 443: It seems a timing issue. Does this happen with new strings only?

Does it mean that the release cycle is faster than the pontoon string merging task? How can it be?

I would prefer to detect this and throw errors at build time.

bakulf (changes requested):

scripts/addon/build.py
mcleinman: Yes, it's only for new strings, and only on languages that were previously 100% translated. It can mean the testing/RC is moving quicker than the strings do, as they can take up to 5 days or so - there are two PRs that need to happen to get everything working on our side:
1. We add strings into our codebase (in this case, Sept 29)
2. [Extract new strings and update all locales](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/9c0b512c2ea0f5710133c72cf3a07210381e2a3e) job is run, and merged into the translations repo (happened Oct 2 in this case, it’s set to run Mon-Thurs at 7am)
3. We update our codebase with the new translations (in this situation, [Oct 3rd](https://github.com/mozilla-mobile/mozilla-vpn-client/commit/bc47ff046c4a436ad8be7a6bf23690a2fee86d36))

One alternative I considered is changing how we count translation strings, which would fix the issue of miscalculated translation completeness. We'd have to write a new script for this, which seemed more code, more fragile, and more time consuming than always falling back to English.

Another alternative I considered is to run the extract new strings job on our repo before building (while using the same method of counting strings) - Essentially duplicate the Pontoon repo's ["Extract new strings" job](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/actions/runs/6530122985/workflow#L54), which runs scripts that live in our repo. This would put the sources in the `.xilff` files, preventing the bug where we misreport languages as 100% translated, causing this bug.

I didn't take this route for a couple reasons:
- It seems fragile. It needs to keep in sync w/ the GH action in the other repo, and tweaks to the script could break this or the other one. I don't expect it to fail regularly, but there is a brittleness that means I'd anticipate some issue over the next many years.
- Always falling back to English is more defensive code - it means this bug will never crop up, and could protect against other issues.

Are there downsides to always falling back to English? If there aren't any, I think my preference is to not add more build complexity.

bakulf (approved):

Issue #7952, opened 2023-09-11
[a11y] Elements in Select Location screen are not correctly narrated by the Screen Reader


data-sync-user:
**Important details:**

* “Last updated” is not correctly narrated when the VPN is OFF
* “Select entry location/Select exit location” is not correctly narrated when the VPN is OFF and ON.

**Affected versions:**

* 2.17.0 (2.202309082148);

**Tested Platforms:**

* Windows 10;
* MacOS 13;

**Prerequisites:**

* Have an active account with a subscription made for Mozilla VPN
* Have Mozilla VPN app installed on your device and turned off
* Have a screen reader activated (Narrator, VoiceOver)

**Steps to reproduce:**

1. Go to Select location ~~> Single~~hop -> Recommended tab.
1. Navigate to Last updated section using the keyboard  
1. Observe how the elements are narrated
1. Switch to Multi-hop
1. Navigate to Select entry location/Select exit location using the keyboard
1. Observe how the elements are narrated

**Expected result:**

1. “Last updated“ string is correctly narrated when the VPN is off;
1. The servers selected from both entry and exit location are correctly narrated by the screen reader

**Actual result:**

1. “Last updated“ string is not narrated when the VPN is off; instead, “Refresh list button blank” is narrated. When navigating from bottom to top (using CAPS +U keys), the “Last updated” string is correctly narrated only on Windows; on Mac, the string is not narrated when navigating from bottom to top.
1. The servers selected from both entry and exit location are not narrated by the screen reader; instead “Select entry location button”/”Select exit location button” is narrated.

**Notes:**

* The issue is also reproducible on 2.16.0.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5492)
┆Reporter: Raluca Monica Ilban

data-sync-user:
➤ Raluca Monica Ilban commented:

Verified as fixed on the latest 2.18 builds, using Windows 10 and MacOS 12.6.6.

PR #8269, opened 2023-10-11
VPN-5492: Some elements in 'Select Location' screen are not read by screen reader

vinocher:
## Description

Some elements in the 'Select Location' screen did not have the correct Accessible names.

## Reference

[VPN-5492](https://mozilla-hub.atlassian.net/browse/VPN-5492), [GitHub 7952](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7952)

## Checklist
    
[VPN-5492]: https://mozilla-hub.atlassian.net/browse/VPN-5492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/screens/home/ViewMultiHop.qml
+++ b/src/ui/screens/home/ViewMultiHop.qml
@@ -76,7 +76,8 @@ StackView {
 
                 objectName: "buttonSelectEntry"
                 titleText: MZI18n.MultiHopFeatureMultiHopEntryLocationHeader
-                descriptionText: VPNCurrentServer.localizedEntryCityName
+                // The entry server may not be set in VPNCurrentServer.localizedEntryCityName if the user has just switched to the MultiHop screen, so use multiHopEntryServer
+                descriptionText: segmentedNav.multiHopEntryServer[2]
                 contentChildren: [
                     ServerLabel {
                         id: entryLabel
@@ -110,7 +111,7 @@ StackView {
                     objectName: "buttonSelectExit"
                     btnObjectName: "buttonSelectExit-btn"
                     titleText: MZI18n.MultiHopFeatureMultiHopExitLocationHeader
-                    descriptionText: VPNCurrentServer.localizedExitCityName
+                    descriptionText: segmentedNav.multiHopExitServer[2]
                     contentChildren: [
 
                         ServerLabel {
--- a/src/ui/screens/home/ViewMultiHop.qml
+++ b/src/ui/screens/home/ViewMultiHop.qml
@@ -76,7 +76,7 @@ StackView {
 
                 objectName: "buttonSelectEntry"
                 titleText: MZI18n.MultiHopFeatureMultiHopEntryLocationHeader
-                descriptionText: titleText
+                descriptionText: VPNCurrentServer.localizedEntryCityName
                 contentChildren: [
                     ServerLabel {
                         id: entryLabel
@@ -110,7 +110,7 @@ StackView {
                     objectName: "buttonSelectExit"
                     btnObjectName: "buttonSelectExit-btn"
                     titleText: MZI18n.MultiHopFeatureMultiHopExitLocationHeader
-                    descriptionText: titleText
+                    descriptionText: VPNCurrentServer.localizedExitCityName
                     contentChildren: [
 
                         ServerLabel {
--- a/src/ui/screens/home/ViewMultiHop.qml
+++ b/src/ui/screens/home/ViewMultiHop.qml
@@ -76,7 +76,7 @@ StackView {
 
                 objectName: "buttonSelectEntry"
                 titleText: MZI18n.MultiHopFeatureMultiHopEntryLocationHeader
-                // The entry server may not be set in VPNCurrentServer.localizedEntryCityName if the user has just switched to the MultiHop screen, so use multiHopEntryServer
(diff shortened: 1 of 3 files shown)


Gela (approved):

MattLichtenstein (changes requested):
When accessing the multi-hop view, but still being set to single-hop, the entry server city name is not read out. It works fine if multi-hop is enabled. My guess is that there is no Entry server set when we are set to single-hop

setting `descriptionText: segmentedNav.multiHopEntryServer[2]` would likely do the trick (as that is what is used for the actual label text)... not sure why though since it is bound to the same property (`VPNCurrentServer.localizedEntryCityName`) - probably something to do with binding updates

vinocher:
> When accessing the multi-hop view, but still being set to single-hop, the entry server city name is not read out. It works fine if multi-hop is enabled. My guess is that there is no Entry server set when we are set to single-hop
> 
> setting `descriptionText: segmentedNav.multiHopEntryServer[2]` would likely do the trick (as that is what is used for the actual label text)... not sure why though since it is bound to the same property (`VPNCurrentServer.localizedEntryCityName`) - probably something to do with binding updates

Good catch! Yes, It looks like this occurs because multiHopEntryServer is set in ViewServers.qml when the user switches to the multi-hop view, and VPNCurrentServer.localizedEntryCityName  has not been set yet. Replaced the etnry server's descriptionText with multiHopEntryServer[2]. Also changed the exist server's descriptionText to multiHopExitServer[2] for consistency.

MattLichtenstein:
> > When accessing the multi-hop view, but still being set to single-hop, the entry server city name is not read out. It works fine if multi-hop is enabled. My guess is that there is no Entry server set when we are set to single-hop
> > setting `descriptionText: segmentedNav.multiHopEntryServer[2]` would likely do the trick (as that is what is used for the actual label text)... not sure why though since it is bound to the same property (`VPNCurrentServer.localizedEntryCityName`) - probably something to do with binding updates
> 
> Good catch! Yes, It looks like this occurs because multiHopEntryServer is set in ViewServers.qml when the user switches to the multi-hop view, and VPNCurrentServer.localizedEntryCityName has not been set yet. Replaced the entry server's descriptionText with multiHopEntryServer[2]. Also changed the exist server's descriptionText to multiHopExitServer[2] for consistency.

Yep, so from my understanding this definition in `ViewServers.qml` isn't actually ever useful:

`property var multiHopEntryServer: [VPNCurrentServer.entryCountryCode, VPNCurrentServer.entryCityName, VPNCurrentServer.localizedEntryCityName]`

Maybe we just define it to be an empty list `property var multiHopEntryServer: []` since we never use the default value

would also need to change [this line](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3206e02e69e43eb0c556a8ca0335ca0d558fe5ab/src/ui/screens/home/ViewServers.qml#L112) to check for `multiHopEntryServer.length === 0`

what do you think?

vinocher:
> > > When accessing the multi-hop view, but still being set to single-hop, the entry server city name is not read out. It works fine if multi-hop is enabled. My guess is that there is no Entry server set when we are set to single-hop
> > > setting `descriptionText: segmentedNav.multiHopEntryServer[2]` would likely do the trick (as that is what is used for the actual label text)... not sure why though since it is bound to the same property (`VPNCurrentServer.localizedEntryCityName`) - probably something to do with binding updates
> > 
> > 
> > Good catch! Yes, It looks like this occurs because multiHopEntryServer is set in ViewServers.qml when the user switches to the multi-hop view, and VPNCurrentServer.localizedEntryCityName has not been set yet. Replaced the entry server's descriptionText with multiHopEntryServer[2]. Also changed the exist server's descriptionText to multiHopExitServer[2] for consistency.
> 
> Yep, so from my understanding this definition in `ViewServers.qml` isn't actually ever useful:
> 
> `property var multiHopEntryServer: [VPNCurrentServer.entryCountryCode, VPNCurrentServer.entryCityName, VPNCurrentServer.localizedEntryCityName]`
> 
> Maybe we just define it to be an empty list `property var multiHopEntryServer: []` since we never use the default value
> 
> would also need to change [this line](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3206e02e69e43eb0c556a8ca0335ca0d558fe5ab/src/ui/screens/home/ViewServers.qml#L112) to check for `multiHopEntryServer.length === 0`
> 
> what do you think?

That would break the case when the user is already in multi-hop and then goes into Select Location's multi-hop view directly, because multiHopEntryServer would not be initialized. It looks like multiHopEntryServer does need to be initialized to handle that case. And `multiHopEntryServer.length === 0` check is needed to handle the case when the user in is single-hop.

MattLichtenstein:
> That would break the case when the user is already in multi-hop and then goes into Select Location's multi-hop view directly, because multiHopEntryServer would not be initialized. It looks like multiHopEntryServer does need to be initialized to handle that case. And the `multiHopEntryServer.length === 0` check is needed to handle the case when the user is in single-hop and goes into Select Location's multi-hop view .

you're totally right. okay, I will settle for a comment above `multiHopEntryServer` declaration stating that these are defined in `handleSegmentClick` if we're not already set to multi-hop

MattLichtenstein (approved):

PR #8270, opened 2023-10-12
VPN-5707 Remove the "Messages - REMOVE ME" option from the Developer menu

Gela:
## Description

Santiago said he hates it and we must keep Santiago happy so this has got to go. Sorry I don't make the rules! 😁 

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5705

## Checklist
    

src/ui/screens/getHelp/developerMenu/ViewMessages.qml
```diff
@@ -1,80 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: Is this not useful though? Why did that menu option even have the `REMOVE ME` in it?

Gela: @MattLichtenstein Looks like this was added by you September of last year. Can you confirm that this isn't needed?

MattLichtenstein: this was actually a Baku thing - and I don't ever use this so delete away!

MattLichtenstein (approved):

src/ui/screens/getHelp/developerMenu/ViewMessages.qml
mcleinman: My gut reaction was that we should keep the menu, but I just opened it for the first time and it looks like it just displays the same messages that are in the inbox. I may be missing some functionality, but if that's all it is, I can't think of how it would be useful. If there is some additional functionality, I agree we should keep it.

mcleinman (approved):
👍🏻

src/ui/screens/getHelp/developerMenu/ViewMessages.qml
MattLichtenstein: this view preceeded the message inbox UI. it was a way to test message addons (and before that, all addons) before we had proper UI

PR #8276, opened 2023-10-12
VPN-5688: Prevent VPN activation when completing onboarding

MattLichtenstein:
## Description

- After allowing the system to create a VPN configuration at the end of the mobile onboarding flow, prevent the controller from re-attempting to connect
- iOS and Android only

## Reference

[VPN-5688: Client is turned ON when the user allows to create a VPN connection](https://mozilla-hub.atlassian.net/browse/VPN-5688)

### DIFF
--- a/src/platforms/ios/ioscontroller.swift
+++ b/src/platforms/ios/ioscontroller.swift
@@ -116,7 +116,7 @@ public class IOSControllerImpl : NSObject {
         }
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, isOnboarding: Bool, failureCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void) {
+    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, isOnboarding: Bool, disconnectCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void) {
         IOSControllerImpl.logger.debug(message: "Connecting")
 
         let _ = TunnelManager.withTunnel { tunnel in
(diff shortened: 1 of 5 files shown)


src/platforms/ios/ioscontroller.swift
```diff
@@ -200,6 +200,7 @@ public class IOSControllerImpl : NSObject {
                 //Used to create a VPN configuration without connecting during onboarding
                 if isOnboarding {
                     IOSControllerImpl.logger.info(message: "Finishing onboarding... do not turn on the VPN after gaining permission")
+                    failureCallback()
```
mcleinman, line 203: Since this is now being intentionally used as a `disconnectCallback`, what do you think about renaming it? And updating the log line in the callback?

mcleinman (changes requested):
After onboarding (testing on iOS), the first time I hit the VPN toggle in the app it doesn't turn on. I believe this is related to this PR.

Also, this records glean metrics for turning off the tunnel, right? Is that something that is okay here? (I'm not sure what the answer is.)

MattLichtenstein:
> After onboarding (testing on iOS), the first time I hit the VPN toggle in the app it doesn't turn on. I believe this is related to this PR.

Actually, this exists outside this change, albeit less noticeable as there is the whole "grant system permissions / add configuration" flow in between the user clicking the toggle and the VPN turning on

> Also, this records glean metrics for turning off the tunnel, right? Is that something that is okay here? (I'm not sure what the answer is.)

it does, and I had not considered telemetry impact... I doubt we'd want actual disconnected telemetry

mcleinman:
> > After onboarding (testing on iOS), the first time I hit the VPN toggle in the app it doesn't turn on. I believe this is related to this PR.
> 
> Actually, this exists outside this change, albeit less noticeable as there is the whole "grant system permissions / add configuration" flow in between the user clicking the toggle and the VPN turning on

Do you mean "outside this change" as in "current state in prod" or "current state with new onboarding"? Currently on prod, doesn't it default to being on when you come back from the permission, so if there is something weird here it's not noticeable to users?

MattLichtenstein:
> Do you mean "outside this change" as in "current state in prod" or "current state with new onboarding"? Currently on prod, doesn't it default to being on when you come back from the permission, so if there is something weird here it's not noticeable to users?

I mean currently in prod, heres a video of the flicker (connecting -> off -> on)


https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/0642718e-85a5-4b36-8fc2-d9ab0320c54c

src/platforms/ios/ioscontroller.swift
MattLichtenstein: Agree - updated.

Gela (commented):
c++ changes look reasonable to me, I'd defer to Matt C for the swift/iOS specific changes :)

mcleinman:
> > Do you mean "outside this change" as in "current state in prod" or "current state with new onboarding"? Currently on prod, doesn't it default to being on when you come back from the permission, so if there is something weird here it's not noticeable to users?
> 
> I mean currently in prod, heres a video of the flicker (connecting -> off -> on)
> RPReplay_Final1697223015.MP4

My concern is that this makes the flicker bug much worse, from a user's perspective:

Currently:
1. User tries to turn on the VPN
2. User is sent to Settings to approve VPN
3. User comes back to app, VPN flicks off
4. User turns on the VPN

With this PR:
1. VPN is setup in oboarding
2. User tries to turn on the VPN
3. VPN flicks off
4. User must turn on the VPN again

The current setup arguably seems reasonable (as they come back to our app), but the second setup very much seems buggy (as the VPN just doesn't turn on), and this is the user's first experience with our product.

I believe this bug is now being tracked as https://mozilla-hub.atlassian.net/browse/VPN-5681. If we're going to merge this before fixing VPN-5681, I'd want approval from Design and/or Product. Does that seem reasonable? (Also, if we need someone to look into 5681, I'm happy to pick that up.)

MattLichtenstein:
> My concern is that this makes the flicker bug much worse, from a user's perspective:
> 
> Currently:
> 
> 1. User tries to turn on the VPN
> 2. User is sent to Settings to approve VPN
> 3. User comes back to app, VPN flicks off
> 4. User turns on the VPN
> 
> With this PR:
> 
> 1. VPN is setup in oboarding
> 2. User tries to turn on the VPN
> 3. VPN flicks off
> 4. User must turn on the VPN again
> 
> The current setup arguably seems reasonable (as they come back to our app), but the second setup very much seems buggy (as the VPN just doesn't turn on), and this is the user's first experience with our product.

I agree, this bug is exacerbated due to the change in flow (since there is no settings flow to break up the vpn toggle press and then flicker off back on), but in both the current production and in this PR, the VPN DOES turn on after only being tapped once after a second or two of "confusion" (or at least that has been my experience). Can you verify that you are definitely needing to click the vpn toggle twice? attaching a video of what I am seeing with this patch:

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/e979ba80-6887-4312-b5e8-19bafb6f23f0

> I believe this bug is now being tracked as https://mozilla-hub.atlassian.net/browse/VPN-5681. If we're going to merge this before fixing [VPN-5681](https://mozilla-hub.atlassian.net/browse/VPN-5681), I'd want approval from Design and/or Product. Does that seem reasonable? (Also, if we need someone to look into 5681, I'm happy to pick that up.)

I am not entirely sure how related they are as the bug I believe we are discussing happens in prod 😳 so hopefully not very related. 

Definitely agree that if we should get opinions from prod/design before proceeding with the "flicker"

mcleinman:
My bad on two points:
- It may not be the same as VPN-5681, good call
- It does connect after a second or so, I've just been impatient when testing apparently. 🙃 

That said, I still think we need a sign off before merging - or to work on that delay. I'm happy to look into it if we decide to block this PR on that fix.

mcleinman (approved):
👍🏻

src/telemetry.cpp
```diff
@@ -219,7 +219,8 @@ void Telemetry::initialize() {
       connectionManager, &ConnectionManager::controllerDisconnected, this,
       [this, connectionManager]() {
         if (Feature::get(Feature::Feature_superDooperMetrics)->isSupported()) {
-          if (connectionManager->state() == ConnectionManager::StateOff) {
+          if (connectionManager->state() == ConnectionManager::StateOff &&
+              SettingsHolder::instance()->onboardingCompleted()) {
```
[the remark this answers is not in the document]
MattLichtenstein: 😅

PR #8279, opened 2023-10-12
VPN-5697 - Allow account deletion on Android

mcleinman:
## Description

This was already implemented on iOS, so this was essentially just turning on the feature for Android. However, there were also some requests to improve the formatting and colors and spacing, so that is the bulk of this PR.

- The formatting changes: Better spacing, more consistent font sizes, blue button color for the "Cancel" button.
- Hiding elements w/ no content is to remove the additional spacing they were providing.
- The addition to `android.cmake` (`${Qt6_DIR}/../../libQt6Svg_${ANDROID_ABI}.so`) is due to a note a few lines above what I added (along with a build error I experienced): `android-deploy-qt is bad and randomly decides to not deploy some lib's we are linking to. So as long as qt is behaving bad, let's force the deployment. Feel free to add libs that are breaking on your build locally.`

Mocks: https://www.figma.com/file/CIelhzmFVyDqn6gNvMXolc/Subscription-Management-In-App?type=design&node-id=2375-1618&mode=design&t=xmGujHUgHm1ldZDW-0

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5697

## Checklist
    
### DIFF
--- a/src/ui/deleteAccount/ViewDeleteAccountRequest.qml
+++ b/src/ui/deleteAccount/ViewDeleteAccountRequest.qml
@@ -64,7 +64,7 @@ MZInAppAuthenticationBase {
             font.pixelSize: MZTheme.theme.fontSize
 
             Layout.fillWidth: true
-            Layout.bottomMargin: MZTheme.theme.vSpacingSmall
+            Layout.bottomMargin: MZTheme.theme.listSpacing
         }
 
         Repeater {
@@ -103,7 +103,6 @@ MZInAppAuthenticationBase {
         MZLinkButton {
             Layout.fillWidth: true
 
-            fontName: MZTheme.theme.fontFamily
             // Cancel
             labelText: MZI18n.InAppSupportWorkflowSupportSecondaryActionText
             onClicked: {
--- a/src/ui/deleteAccount/ViewDeleteAccountRequest.qml
+++ b/src/ui/deleteAccount/ViewDeleteAccountRequest.qml
@@ -39,6 +39,7 @@ MZInAppAuthenticationBase {
     property bool allowAccountDeletion: false
 
     _changeEmailLinkVisible: false
+    _disclaimersVisible: false
     _viewObjectName: "authDeleteAccountRequest"
     _menuButtonAccessibleName: MZI18n.GlobalGoBack
     _menuButtonImageSource: "qrc:/nebula/resources/back.svg"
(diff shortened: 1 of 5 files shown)


src/ui/deleteAccount/ViewDeleteAccountRequest.qml
```diff
@@ -59,9 +61,10 @@ MZInAppAuthenticationBase {
                 .arg("<b style='color:" + MZTheme.theme.fontColorDark + ";'>"
                     + MZAuthInApp.emailAddress + "</b>")
             textFormat: Text.RichText
+            font.pixelSize: MZTheme.theme.fontSize
 
             Layout.fillWidth: true
-            Layout.bottomMargin: MZTheme.theme.vSpacing
+            Layout.bottomMargin: MZTheme.theme.vSpacingSmall
```
MattLichtenstein, line 67: mocks show 24px of spacing between the subtitle and the checkbox section. here we have the `bottomMargin` (16px) combined with the `spacing` (16px) totaling 32px of spacing

src/ui/deleteAccount/ViewDeleteAccountRequest.qml
```diff
@@ -100,10 +103,9 @@ MZInAppAuthenticationBase {
         MZLinkButton {
             Layout.fillWidth: true
 
-            fontName: MZTheme.theme.fontBoldFamily
+            fontName: MZTheme.theme.fontFamily
```
MattLichtenstein, line 106: this is actually still the wrong font family. the default one set in [MZLinkButton.qml](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3206e02e69e43eb0c556a8ca0335ca0d558fe5ab/nebula/ui/components/MZLinkButton.qml#L15) is correct, so you can remove this line altogether

MattLichtenstein (changes requested):

src/ui/deleteAccount/ViewDeleteAccountRequest.qml
mcleinman: Thanks, good catch. Updated.

src/ui/deleteAccount/ViewDeleteAccountRequest.qml
mcleinman: Thanks, good catch. Updated.

MattLichtenstein (approved):

PR #8283, opened 2023-10-13
VPN-5520: Decouple data collection checkbox from setting during onboarding

MattLichtenstein:
## Description

- Prevent data collection checkbox in onboarding from modifying internal `gleanEnabled` setting until onboarding is completed
  - The reasoning for this is because we want the initial state of the checkbox to be unchecked so that no changes are necessary to be made by the user to opt out of data collection (aside from completing onboarding), but coupling this with the internal `gleanEnabled` setting would wipe all authentication telemetry for the user.

## Reference

[VPN-5520: All pre-onboarding glean data being deleted upon reaching the Data Collection Consent screen](https://mozilla-hub.atlassian.net/browse/VPN-5520)

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -919,14 +919,6 @@ void MozillaVPN::mainWindowLoaded() {
 void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
 
-  // Set glean based on data collection checkbox checked state
-  // Extracting this property from QML seems like the best approach to avoid
-  // creating a disposable member variable or setting
-  QObject* dataCollectionCheckbox =
-      InspectorUtils::queryObject("//dataCollectionCheckBox");
-  settingsHolder->setGleanEnabled(
-      dataCollectionCheckbox->property("isChecked").toBool());
-
   if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
     logger.debug() << "onboarding completed";
     settingsHolder->setOnboardingCompleted(true);
@@ -936,7 +928,12 @@ void MozillaVPN::onboardingCompleted() {
     // will never see onboarding again
     settingsHolder->setOnboardingStep(0);
 
-    // Mark the old onboarding expereince as completed as well, ensuring that
+    // Toggle glean on or off at the end of onboarding, depending on what the
+    // user selected
+    settingsHolder->setGleanEnabled(
+        settingsHolder->onboardingDataCollectionEnabled());
+
+    // Mark the old onboarding experience as completed as well, ensuring that
     // users do not have to go through it if the new onboaring feature is turned
     // off
     settingsHolder->setPostAuthenticationShown(true);
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -823,7 +823,6 @@ bool MozillaVPN::checkCurrentDevice() {
 }
 
 void MozillaVPN::logout() {
-  SettingsHolder::instance()->setOnboardingCompleted(false);
   logger.debug() << "Logout";
 
   ErrorHandler::instance()->requestAlert(ErrorHandler::LogoutAlert);
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -920,9 +920,13 @@ void MozillaVPN::mainWindowLoaded() {
 void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
 
-  //Set glean based on data collection checkbox checked state
-  QObject* dataCollectionCheckbox = InspectorUtils::queryObject("//dataCollectionCheckBox");
-  settingsHolder->setGleanEnabled(dataCollectionCheckbox->property("isChecked").toBool());
+  // Set glean based on data collection checkbox checked state
+  // Extracting this property from QML seems like the best approach to avoid
+  // creating a disposable member variable or setting
+  QObject* dataCollectionCheckbox =
+      InspectorUtils::queryObject("//dataCollectionCheckBox");
+  settingsHolder->setGleanEnabled(
+      dataCollectionCheckbox->property("isChecked").toBool());
 
   if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
     logger.debug() << "onboarding completed";
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -21,6 +21,7 @@
 #include "glean/mzglean.h"
 #include "i18nstrings.h"
 #include "inspector/inspectorhandler.h"
+#include "inspector/inspectorutils.h"
 #include "leakdetector.h"
 #include "logger.h"
 #include "loghandler.h"
@@ -401,15 +402,6 @@ void MozillaVPN::maybeStateMain() {
   // getting to the home screen
   if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
     if (!settingsHolder->onboardingCompleted()) {
-      // We turn telemetry off when onboarding starts so that the user has to
-      // specifically opt in to all future telemetry *Note: This is needed
-      // because telemetry is on prior to onboarding to record metrics regarding
-      // auth
-      if (!settingsHolder->onboardingStarted()) {
-        settingsHolder->setGleanEnabled(false);
-        settingsHolder->setOnboardingStarted(true);
-      }
-
       setState(StateOnboarding);
       return;
     }
@@ -831,6 +823,7 @@ bool MozillaVPN::checkCurrentDevice() {
 }
 
 void MozillaVPN::logout() {
+  SettingsHolder::instance()->setOnboardingCompleted(false);
   logger.debug() << "Logout";
 
   ErrorHandler::instance()->requestAlert(ErrorHandler::LogoutAlert);
@@ -926,6 +919,10 @@ void MozillaVPN::mainWindowLoaded() {
 
 void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
+
+  //Set glean based on data collection checkbox checked state
+  QObject* dataCollectionCheckbox = InspectorUtils::queryObject("//dataCollectionCheckBox");
+  settingsHolder->setGleanEnabled(dataCollectionCheckbox->property("isChecked").toBool());
 
   if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
     logger.debug() << "onboarding completed";
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -21,7 +21,6 @@
 #include "glean/mzglean.h"
 #include "i18nstrings.h"
 #include "inspector/inspectorhandler.h"
-#include "inspector/inspectorutils.h"
 #include "leakdetector.h"
 #include "logger.h"
 #include "loghandler.h"
@@ -923,11 +922,6 @@ void MozillaVPN::onboardingCompleted() {
     logger.debug() << "onboarding completed";
     settingsHolder->setOnboardingCompleted(true);
 
-    // Resetting for the benefit of testing so that we only have to reset one
-    // setting (onboardingCompleted) manually. No real affect on user since they
-    // will never see onboarding again
-    settingsHolder->setOnboardingStep(0);
-
     // Toggle glean on or off at the end of onboarding, depending on what the
     // user selected
     settingsHolder->setGleanEnabled(
@@ -937,6 +931,12 @@ void MozillaVPN::onboardingCompleted() {
     // users do not have to go through it if the new onboaring feature is turned
     // off
     settingsHolder->setPostAuthenticationShown(true);
+
+    // Resetting for the benefit of testing so that we only have to reset one
+    // setting (onboardingCompleted) manually. No real affect on user since they
+    // *should* never see onboarding more than once
+    settingsHolder->setOnboardingStep(0);
+    settingsHolder->setOnboardingDataCollectionEnabled(false);
 
   } else {
     logger.debug() << "telemetry policy completed";
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -379,6 +379,17 @@ SETTING_BOOL(onboardingCompleted,        // getter
              false                       // sensitive (do not log)
 )
 
+SETTING_BOOL(onboardingDataCollectionEnabled,        // getter
+             setOnboardingDataCollectionEnabled,     // setter
+             removeOnboardingDataCollectionEnabled,  // remover
+             hasOnboardingDataCollectionEnabled,     // has
+             "onboardingDataCollectionEnabled",      // key
+             false,                                  // default value
+             false,                                  // user setting
+             false,                                  // remove when reset
+             false                                   // sensitive (do not log)
+)
+
 SETTING_INT(onboardingStep,        // getter
             setOnboardingStep,     // setter
             removeOnboardingStep,  // remover
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -368,17 +368,6 @@ SETTING_STRINGLIST(missingApps,        // getter
                    false               // sensitive (do not log)
 )
 
-SETTING_BOOL(onboardingStarted,        // getter
-             setOnboardingStarted,     // setter
-             removeOnboardingStarted,  // remover
-             hasOnboardingStarted,     // has
-             "onboardingStarted",      // key
-             false,                    // default value
-             false,                    // user setting
-             false,                    // remove when reset
-             false                     // sensitive (do not log)
-)
-
 SETTING_BOOL(onboardingCompleted,        // getter
              setOnboardingCompleted,     // setter
              removeOnboardingCompleted,  // remover
--- a/src/ui/screens/onboarding/OnboardingDataSlide.qml
+++ b/src/ui/screens/onboarding/OnboardingDataSlide.qml
@@ -65,8 +65,9 @@ ColumnLayout {
         leftMargin: 0
         subLabelText: MZI18n.OnboardingDataSlideCheckboxLabel
         showDivider: false
+        isChecked: MZSettings.onboardingDataCollectionEnabled
 
-        onClicked: isChecked = !isChecked
+        onClicked: MZSettings.onboardingDataCollectionEnabled = !MZSettings.onboardingDataCollectionEnabled
     }
 
     Item {
--- a/src/ui/screens/onboarding/OnboardingDataSlide.qml
+++ b/src/ui/screens/onboarding/OnboardingDataSlide.qml
@@ -55,19 +55,18 @@ ColumnLayout {
     }
 
     MZCheckBoxRow {
+        id: dataCollectionCheckBox
         objectName: "dataCollectionCheckBox"
 
         Layout.leftMargin: MZTheme.theme.windowMargin * 2
         Layout.rightMargin: MZTheme.theme.windowMargin * 2
         Layout.fillWidth: true
 
-        subLabelText: MZI18n.OnboardingDataSlideCheckboxLabel
         leftMargin: 0
-        isChecked: MZSettings.gleanEnabled
+        subLabelText: MZI18n.OnboardingDataSlideCheckboxLabel
         showDivider: false
-        onClicked: {
-            MZSettings.gleanEnabled = !MZSettings.gleanEnabled
-       }
+
+        onClicked: isChecked = !isChecked
     }
 
     Item {
--- a/src/ui/screens/onboarding/OnboardingDataSlide.qml
+++ b/src/ui/screens/onboarding/OnboardingDataSlide.qml
@@ -55,7 +55,6 @@ ColumnLayout {
     }
 
     MZCheckBoxRow {
-        id: dataCollectionCheckBox
         objectName: "dataCollectionCheckBox"
 
         Layout.leftMargin: MZTheme.theme.windowMargin * 2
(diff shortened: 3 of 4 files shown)


src/mozillavpn.cpp
```diff
@@ -927,6 +919,14 @@ void MozillaVPN::mainWindowLoaded() {
 void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
 
+  // Set glean based on data collection checkbox checked state
+  // Extracting this property from QML seems like the best approach to avoid
+  // creating a disposable member variable or setting
+  QObject* dataCollectionCheckbox =
```
bakulf, line 925: Nothing guarantees that this object is still there. I would use a setting flag, a signal, or something else here instead of qmlpath.

src/mozillavpn.cpp
```diff
@@ -401,15 +402,6 @@ void MozillaVPN::maybeStateMain() {
   // getting to the home screen
   if (Feature::get(Feature::Feature_newOnboarding)->isSupported()) {
     if (!settingsHolder->onboardingCompleted()) {
-      // We turn telemetry off when onboarding starts so that the user has to
-      // specifically opt in to all future telemetry *Note: This is needed
-      // because telemetry is on prior to onboarding to record metrics regarding
-      // auth
-      if (!settingsHolder->onboardingStarted()) {
-        settingsHolder->setGleanEnabled(false);
-        settingsHolder->setOnboardingStarted(true);
```
Gela, line 410: What was the line `settingsHolder->setOnboardingStarted(true);` doing before? Wondering if that should still stay?

tests/functional/testOnboarding.js
```diff
@@ -12,24 +12,23 @@ describe('Onboarding', function() {
...
-    assert.equal(await vpn.getSetting('onboardingStarted'), true);
 
     //DATA SLIDE
     //Do some checks on the initial state of the data slide
     await vpn.waitForQuery(queries.screenOnboarding.DATA_SLIDE.visible());
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.ONBOARDING_VIEW, 'currentIndex'), 0);
     assert.equal(await vpn.getSetting('onboardingStep'), 0);
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'false');
-    assert.equal(await vpn.getSetting('gleanEnabled'), false);
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
```
Gela, line 31: So right off the bat, we are collecting telemetry and then once the user gets to the consent page on Onboarding they can opt out if they choose to?

tests/functional/testOnboarding.js
```diff
@@ -205,13 +203,17 @@ describe('Onboarding', function() {
     assert.equal(await vpn.getQueryProperty(queries.screenSettings.privacyView.BLOCK_ADS_CHECKBOX, 'checked'), 'true');
     assert.equal(await vpn.getQueryProperty(queries.screenSettings.privacyView.BLOCK_TRACKERS_CHECKBOX, 'checked'), 'true');
     assert.equal(await vpn.getQueryProperty(queries.screenSettings.privacyView.BLOCK_MALWARE_CHECKBOX, 'checked'), 'true');
+    assert.equal(await vpn.getSetting('dnsProviderFlags'), 14);
```
Gela, line 206: What is 14 here? I see DNS Provider flags defined as following in the code. They would be 0, 1, 2, 4, 8 🤔 

```c++
  enum DNSProviderFlags {
    Gateway = 0x00,
    Custom = 0x01,
    BlockAds = 0x02,
    BlockTrackers = 0x04,
    BlockMalware = 0x08,
  };
```

tests/functional/testOnboarding.js
```diff
@@ -277,4 +279,65 @@ describe('Onboarding', function() {
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.ONBOARDING_VIEW, 'currentIndex'), 3);
     assert.equal(await vpn.getSetting('onboardingStep'), 3);
   });
+
+  it.only('Verify data collection checkbox and setting are decoupled', async () => {
```
Gela, line 283: I've made this same mistake so many times!

```suggestion
  it('Verify data collection checkbox and setting are decoupled', async () => {
```

tests/functional/testOnboarding.js
```diff
@@ -277,4 +279,65 @@ describe('Onboarding', function() {
...
+
+    //This function completes onboarding, and checks that the gleanEnabled setting remains enabled throughout the flow
```
Gela, line 285: For the sake of consistency, could you move the test description right above the test (so right above `it(...)`)?


also THANK YOU for adding tests, love to see it :)

tests/functional/testOnboarding.js
```diff
@@ -277,4 +279,65 @@ describe('Onboarding', function() {
...
+    //Ensure we are in onboarding
+    await vpn.waitForQuery(queries.screenOnboarding.DATA_SLIDE.visible());
+
+    //Ensure that the data collection checkbox and internal setting are decoupled
+    //We will finish onboarding with the data collection checkbox being unchecked
+    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'false');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.DATA_CHECKBOX.visible());
+    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'true');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
```
Gela, line 310: Is this back and forth checkbox toggle necessary here for the sake of the test flow? Asking because we are testing checking this on line 30 in a different test.

Gela (changes requested):

src/mozillavpn.cpp
```diff
@@ -927,6 +920,10 @@ void MozillaVPN::mainWindowLoaded() {
 void MozillaVPN::onboardingCompleted() {
   SettingsHolder* settingsHolder = SettingsHolder::instance();
 
+  //Set glean based on data collection checkbox checked state
+  QObject* dataCollectionCheckbox = InspectorUtils::queryObject("//dataCollectionCheckBox");
```
brizental: This feels fragile. What if the user restart the application while in the middle of onboarding? Won't they directly go to the last step? What is the state of the data collection checkbox then? It might be safer to use a setting for the state of this checkbox.

tests/functional/testOnboarding.js
```diff
@@ -12,24 +12,23 @@ describe('Onboarding', function() {
 
   beforeEach(async () => {
     await vpn.flipFeatureOn("newOnboarding");
-    assert.equal(await vpn.getSetting('onboardingStarted'), false);
     assert.equal(await vpn.getSetting('onboardingCompleted'), false);
+    assert.equal(await vpn.getSetting('onboardingStep'), 0);
     await vpn.authenticateInApp();
   });
 
   //Tests pretty much all of the new onboarding, including all slides, controls, navigation, and ensuring settings persist into StateMain
```
brizental, line 20: I know this is not part of this PR, but why use a single test to test so many different things? If possible, it would be ideal to make the tests more specific. It makes it easier to debug breakage and understand the test.

tests/functional/testOnboarding.js
brizental: Why is this part of this PR?

tests/functional/testOnboarding.js
```diff
@@ -12,24 +12,23 @@ describe('Onboarding', function() {
...
     //DATA SLIDE
     //Do some checks on the initial state of the data slide
     await vpn.waitForQuery(queries.screenOnboarding.DATA_SLIDE.visible());
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.ONBOARDING_VIEW, 'currentIndex'), 0);
     assert.equal(await vpn.getSetting('onboardingStep'), 0);
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'false');
-    assert.equal(await vpn.getSetting('gleanEnabled'), false);
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
 
     //Check data collection checkbox
```
brizental, line 33: The check below `assert.equal(await vpn.getSetting('gleanEnabled'), true);` is now outdated, since the setting is decoupled from the checkbox.

tests/functional/testOnboarding.js
brizental: If `it.only` is being used for debugging reasons, we have the option to use mocha's `--grep` option to run a single test case instead, no code changes required. Just do `npm run functionTest path/to/test.js -- --grep "decoupled"` for example.

tests/functional/testOnboarding.js
```diff
@@ -277,4 +280,69 @@ describe('Onboarding', function() {
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.ONBOARDING_VIEW, 'currentIndex'), 3);
     assert.equal(await vpn.getSetting('onboardingStep'), 3);
   });
+
+  it.only('Verify data collection checkbox and setting are decoupled', async () => {
```
brizental, line 283: This test case should be broken into multiple too.

tests/functional/testOnboarding.js
brizental: Note: This also should not be necessary by ending the test case here and starting new ones for the other checks.

tests/functional/testOnboarding.js
```diff
@@ -277,4 +280,69 @@ describe('Onboarding', function() {
...
+    await vpn.waitForQuery(queries.screenOnboarding.STEP_NAV_STACK_VIEW.ready());
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.START_NEXT_BUTTON.visible());
+    await vpn.waitForQuery(queries.screenHome.SCREEN.visible());
+    assert.equal(await vpn.getSetting('onboardingCompleted'), true);
+
+    //Now that onboarding is complete, make sure the gleanEnabled setting is now disabled
+    assert.equal(await vpn.getSetting('gleanEnabled'), false);
+
+    //Now repeat the process, but we will finish onboarding with the data collection checkbox being checked
```
brizental, line 321: New test case here.

tests/functional/testOnboarding.js
```diff
@@ -277,4 +280,69 @@ describe('Onboarding', function() {
...
+    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'false');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.DATA_CHECKBOX.visible());
+    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'true');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.DATA_CHECKBOX.visible());
+    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'false');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+
+    //Complete onboarding, and check that the gleanEnabled setting does not change between steps until onboarding is completed
```
brizental: New test case here.

tests/functional/testOnboarding.js
```diff
@@ -277,4 +280,69 @@ describe('Onboarding', function() {
...
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.DEVICES_NEXT_BUTTON.visible());
+    await vpn.waitForQuery(queries.screenOnboarding.STEP_NAV_STACK_VIEW.ready());
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+    await vpn.waitForQueryAndClick(queries.screenOnboarding.START_NEXT_BUTTON.visible());
+    await vpn.waitForQuery(queries.screenHome.SCREEN.visible());
+    assert.equal(await vpn.getSetting('onboardingCompleted'), true);
+
+    //Now that onboarding is complete, make sure the gleanEnabled setting is still enabled
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
+  });
```
brizental, line 342: Please add a test case for when the application is closed and reopened during onboarding.

tests/functional/testOnboarding.js
```diff
@@ -295,19 +312,8 @@ describe('Onboarding', function() {
...
-    await vpn.waitForQueryAndClick(queries.screenOnboarding.PRIVACY_NEXT_BUTTON.visible());
-    await vpn.waitForQuery(queries.screenOnboarding.STEP_NAV_STACK_VIEW.ready());
-    assert.equal(await vpn.getSetting('gleanEnabled'), true);
-    await vpn.waitForQueryAndClick(queries.screenOnboarding.DEVICES_NEXT_BUTTON.visible());
-    await vpn.waitForQuery(queries.screenOnboarding.STEP_NAV_STACK_VIEW.ready());
-    assert.equal(await vpn.getSetting('gleanEnabled'), true);
-    await vpn.waitForQueryAndClick(queries.screenOnboarding.START_NEXT_BUTTON.visible());
-    await vpn.waitForQuery(queries.screenHome.SCREEN.visible());
-    assert.equal(await vpn.getSetting('onboardingCompleted'), true);
+    //Complete onboarding, and check that the gleanEnabled setting remains enabled throughout onboarding
```
brizental, line 315: nit: Redundant and incorrect comment. I think we can just remove it.

tests/functional/testOnboarding.js
```diff
@@ -328,19 +334,8 @@ describe('Onboarding', function() {
...
```
brizental, line 337: nit: Redundant comment too.

tests/functional/testOnboarding.js
```diff
@@ -205,13 +203,17 @@ describe('Onboarding', function() {
...
+    assert.equal(await vpn.getSetting('dnsProviderFlags'), 14);
     await vpn.waitForQueryAndClick(queries.screenSettings.BACK.visible());
     await vpn.waitForQuery(queries.screenSettings.STACKVIEW.ready());
 
     //Check data collection and start at boot
     await vpn.waitForQueryAndClick(queries.screenSettings.APP_PREFERENCES.visible());
     assert.equal(await vpn.getQueryProperty(queries.screenSettings.appPreferencesView.DATA_COLLECTION, 'isChecked'), 'true');
+    assert.equal(await vpn.getSetting('gleanEnabled'), true);
     assert.equal(await vpn.getQueryProperty(queries.screenSettings.appPreferencesView.START_AT_BOOT, 'isChecked'), 'true');
+    assert.equal(await vpn.getSetting('startAtBoot'), true);
```
brizental, line 215: Why is this part of this PR?

tests/functional/testOnboarding.js
```diff
@@ -277,4 +279,65 @@ describe('Onboarding', function() {
...
+    //Now that onboarding is complete, make sure the gleanEnabled setting is now disabled
+    assert.equal(await vpn.getSetting('gleanEnabled'), false);
+
+    //Now repeat the process, but we will finish onboarding with the data collection checkbox being checked
+
+    //Reset settings, logout, and authenticate
+    await vpn.setSetting('onboardingCompleted', false)
+    await vpn.setSetting('gleanEnabled', true);
+    await vpn.logout();
+    await vpn.authenticateInApp();
```
brizental, line 327: This whole block can be removed by breaking this test case into mulitple.

brizental (changes requested):

src/mozillavpn.cpp
MattLichtenstein: `onboardingStarted` was a way to know that onboarding was shown for the first time (with consecutive shows being possible via quitting and reopening the app). It was meant to disable glean when the onboarding flow is shown for the first time

tests/functional/testOnboarding.js
MattLichtenstein: yep. their selection will take affect once onboarding is completed

tests/functional/testOnboarding.js
MattLichtenstein: 🫠

tests/functional/testOnboarding.js
MattLichtenstein: another thing I was cleaning up while in the file

tests/functional/testOnboarding.js
Gela: Can you add a comment to explain the math/logic for that line?

src/settingslist.h
```diff
@@ -379,6 +379,17 @@ SETTING_BOOL(onboardingCompleted,        // getter
...
 )
 
+SETTING_BOOL(onboardingDataCollectionEnabled,        // getter
+             setOnboardingDataCollectionEnabled,     // setter
+             removeOnboardingDataCollectionEnabled,  // remover
+             hasOnboardingDataCollectionEnabled,     // has
+             "onboardingDataCollectionEnabled",      // key
+             false,                                  // default value
+             false,                                  // user setting
+             false,                                  // remove when reset
```
brizental, line 389: This should be `true`, no?

src/ui/screens/onboarding/OnboardingDataSlide.qml
```diff
@@ -65,8 +65,9 @@ ColumnLayout {
         leftMargin: 0
         subLabelText: MZI18n.OnboardingDataSlideCheckboxLabel
         showDivider: false
+        isChecked: MZSettings.onboardingDataCollectionEnabled
 
-        onClicked: isChecked = !isChecked
+        onClicked: MZSettings.onboardingDataCollectionEnabled = !MZSettings.onboardingDataCollectionEnabled
```
brizental, line 70: I think we should clear this setting after onboarding is complete, just to be sure. In the unlikely event that the user is put through onboarding again after going through it once. This must always be `false` at the start of onboarding, right?

tests/functional/testOnboarding.js
brizental: Totally fine not to do this on this PR, but in that case please file a ticket to fo it as a follow up.

tests/functional/testOnboarding.js
```diff
@@ -280,64 +299,72 @@ describe('Onboarding', function() {
...
+
+    //Complete onboarding and get to the home screen
+    await completeOnboarding(await vpn.getSetting('onboardingStep'));
+
+    //Ensure glean is disabled
     assert.equal(await vpn.getSetting('gleanEnabled'), false);
+  });
 
-    //Now repeat the process, but we will finish onboarding with the data collection checkbox being checked
+  //Tests completing onboarding with the data collection checkbox enabled
```
brizental, line 332: nit: Redundant comment.

tests/functional/testOnboarding.js
```diff
@@ -280,64 +299,72 @@ describe('Onboarding', function() {
...
-    assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.DATA_CHECKBOX, 'isChecked'), 'true');
-    assert.equal(await vpn.getSetting('gleanEnabled'), true);
 
-    //Complete onboarding, and check that the gleanEnabled setting remains enabled throughout onboarding
-    await completeOnboarding();
+    //Complete onboarding and get to the home screen
+    await completeOnboarding(await vpn.getSetting('onboardingStep'));
 
-    //Now that onboarding is complete, make sure the gleanEnabled setting is still enabled
+    //Ensure glean is disabled
```
brizental, line 344: enabled*

tests/functional/testOnboarding.js
```diff
@@ -280,64 +299,72 @@ describe('Onboarding', function() {
...
     assert.equal(await vpn.getSetting('gleanEnabled'), true);
   });
+
+  //Tests onboardingDataCollectionEnabled setting persists across app sessions
```
brizental, line 348: nit: Redundant comment.

tests/functional/testOnboarding.js
```diff
@@ -280,64 +299,72 @@ describe('Onboarding', function() {
...
+
+    //Quit and relaunch the app
+    await vpn.quit();
+    await setup.startAndConnect();
+
+    //Ensure we are still in onboarding
+    await vpn.waitForQuery(queries.screenOnboarding.DATA_SLIDE.visible());
+
+    //Ensure onboardingDataCollectionEnabled is set to true
+    assert.equal(await vpn.getSetting('onboardingDataCollectionEnabled'), true);
```
brizental, line 368: nit: Also check the checkbox is checked.

src/mozillavpn.cpp
```diff
@@ -21,6 +21,7 @@
 #include "glean/mzglean.h"
 #include "i18nstrings.h"
 #include "inspector/inspectorhandler.h"
+#include "inspector/inspectorutils.h"
```
brizental, line 24: This can be removed.

tests/functional/testOnboarding.js
```diff
@@ -277,4 +298,73 @@ describe('Onboarding', function() {
     assert.equal(await vpn.getQueryProperty(queries.screenOnboarding.ONBOARDING_VIEW, 'currentIndex'), 3);
     assert.equal(await vpn.getSetting('onboardingStep'), 3);
   });
+
+  //Tests that the data collection checkbox is bound to onboardingDataCollectionEnabled, and it's default state
+  it('Data collection checkbox default and binding', async () => {
```
brizental, line 303: nit: Instead of having this separate test case, I'd check the default and binding in all the tests related to the checkbox. That approach will catch possible issues with the resetting of the checkbox.

brizental (approved):
r+wc. Left some nits, nothing major. Thanks for working on this, Matt!

src/settingslist.h
MattLichtenstein: we don't need to remove this setting on reset since it gets set back to it's default setting

Gela (approved):

PR #8297, opened 2023-10-16
[RFC] VPN-4935 - Cirrus Experiments

brizental:
This RFC was first shared with product and the Nimbus team on Google Docs. It has been aproved by that crowd and I am now sharing it here to get technical feedback from engineers that also work on the MozVPN.

[Rendered](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/340996ebf399d39708dd4d596546cf292c1f0469/docs/rfcs/0004-cirrus-experiments.md).

### DIFF
--- a/docs/rfcs/0004-cirrus-experiments.md
+++ b/docs/rfcs/0004-cirrus-experiments.md
@@ -66,7 +66,7 @@ Cirrus uses Glean internally. Therefore it requires a Glean application id. The 
 
 Documentation of metrics and pings sent by the Cirrus service can be found at <https://github.com/mozilla/experimenter/blob/main/cirrus/server/telemetry/docs/metrics.md>.
 
-Note Cirrus also  requires a Remote Settings URL.  That will be provided by the Nimbus team and there is no action on our side required.
+> **Note** Cirrus also  requires a Remote Settings URL.  That will be provided by the Nimbus team and there is no action on our side required.
 
 #### 2\. Deploy Cirrus as a sidecar to Guardian
 
@@ -111,7 +111,7 @@ This endpoint will be changed to only accept POST requests. Requests will be exp
 }
 ```
 
-Note Throughout this document, client_id will be referred to as "experimenter id" so as not to confuse it with the Glean client_id.
+> **Note** Throughout this document, client_id will be referred to as "experimenter id" so as not to confuse it with the Glean client_id.
 
 The Cirrus reply will be added to the body of the response object, like so:
 ```
@@ -138,7 +138,7 @@ The Cirrus reply will be added to the body of the response object, like so:
 }
 ```
 
-Note This request does not need to be authenticated. It only needs to be made using HTTPS.
+> **Note** This request does not need to be authenticated. It only needs to be made using HTTPS.
 
 If "featuresOverride" conflicts with "experimentalFeatures", "experimentalFeatures" will always have higher priority.
 
@@ -162,7 +162,7 @@ Timing: The experimenter id must be set as soon as possible on the application l
 
 The /featurelist endpoint is already queried by the application  on initialization and once every hour. Other than these triggers, a feature list request will be made whenever the experiment id is set.
 
-Note The list of active features will be added to each VPN log summary, to aid in debugging branch specific bugs.
+> **Note** The list of active features will be added to each VPN log summary, to aid in debugging branch specific bugs.
 
 #### 3\. Apply features
 
@@ -199,7 +199,7 @@ EXPERIMENTAL_FEATURE(
 
 The EXPERIMENTAL_FEATURE macro, intentionally doesn't contain the field for declaring if the feature can be flipped on or off. Experimental features must always be "toggleable".
 
-Note Mozilla VPN features may or may not be "toggleable", meaning they may not be allowed to be turned on or off. Experimental features must always be toggleable, otherwise they cannot be applied dynamically. This has consequences for the implementation of the treatment itself. This needs to be considered on a case by case basis.
+> **Note** Mozilla VPN features may or may not be "toggleable", meaning they may not be allowed to be turned on or off. Experimental features must always be toggleable, otherwise they cannot be applied dynamically. This has consequences for the implementation of the treatment itself. This needs to be considered on a case by case basis.
 
 It also doesn't contain the feature callback field. Experimental features always default to false unless they have been turned on through Cirrus. 
 
@@ -219,7 +219,7 @@ MZSettings.onboardingCtaStyleExperiment.fontFamily
 
 Experimental feature settings will only have public getters. They may only be set or removed by the TaskGetFeatureList.
 
-Note In order for this setup to work, experimental features will have to be declared in a separate feature list file e.g. experimentalfeaturelist.h. This will make it possible to import the experimental features only on the settings holder module and create the settings API necessary.
+> **Note** In order for this setup to work, experimental features will have to be declared in a separate feature list file e.g. experimentalfeaturelist.h. This will make it possible to import the experimental features only on the settings holder module and create the settings API necessary.
 
 It will be the responsibility of the TaskGetFeatureList at the time of parsing the featurelist response, to set each of these settings to the received value. Turning the feature on will happen only after all settings are set.
 
--- a/docs/rfcs/0004-cirrus-experiments.md
+++ b/docs/rfcs/0004-cirrus-experiments.md
@@ -3,13 +3,9 @@
 - Author: [@brizental](https://github.com/brizental)
 - RFC PR: [#TODO](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/TODO)
 - Implementation GitHub issue: TODO
-  
+
 Mozilla VPN and Nimbus Integration
 ==================================
-
-Author: Beatriz Rizental (brizental)
-
-Ticket: [VPN-5247](https://mozilla-hub.atlassian.net/browse/VPN-5247)
 
 TIMELINE
 ========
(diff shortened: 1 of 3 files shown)


docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+Server side will be responsible for hosting the Cirrus service and exposing its functionality to the VPN client.
+
+#### 1\. Register the Cirrus application id with Glean
+
+Cirrus uses Glean internally. Therefore it requires a Glean application id. The process for requesting an application id can be found on[  Enabling data to be ingested by the data platform](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html). We would want something like "mozillavpn_cirrus".
+
+Documentation of metrics and pings sent by the Cirrus service can be found at <https://github.com/mozilla/experimenter/blob/main/cirrus/server/telemetry/docs/metrics.md>.
+
+Note Cirrus also  requires a Remote Settings URL.  That will be provided by the Nimbus team and there is no action on our side required.
```
brizental, line 69:
```suggestion
> **Note** Cirrus also  requires a Remote Settings URL.  That will be provided by the Nimbus team and there is no action on our side required.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+    "key1":  "value1",
+    "key2":  {
+      "key2.1":  "value2",
+      "key2.2":  "value3"
+    }
+  }
+}
+```
+
+Note Throughout this document, client_id will be referred to as "experimenter id" so as not to confuse it with the Glean client_id.
```
brizental, line 114:
```suggestion
> **Note** Throughout this document, client_id will be referred to as "experimenter id" so as not to confuse it with the Glean client_id.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+    },
+    "FeatureN":  {
+      "VariableN.1":  "valueX",
+      "VariableN.2":  "valueY"
+    }
+  }
+}
+```
+
+Note This request does not need to be authenticated. It only needs to be made using HTTPS.
```
brizental, line 141:
```suggestion
> **Note** This request does not need to be authenticated. It only needs to be made using HTTPS.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+Nimbus relies on Glean data for experiment analysis. Therefore the generated experimenter id needs to be included in all Glean pings. The Glean team is working on a new API that will facilitate setting this id, see[  Bug 1848201](https://bugzilla.mozilla.org/show_bug.cgi?id=1848201). The Mozilla VPN application will rely on this API to provide the experiment id to Glean.
+
+Timing: The experimenter id must be set as soon as possible on the application lifecycle, since it is required for all Cirrus requests and should be included in all Glean pings.
+
+#### 2\. Query Nimbus for active features
+
+The /featurelist endpoint is already queried by the application  on initialization and once every hour. Other than these triggers, a feature list request will be made whenever the experiment id is set.
+
+Note The list of active features will be added to each VPN log summary, to aid in debugging branch specific bugs.
```
brizental, line 165:
```suggestion
> **Note** The list of active features will be added to each VPN log summary, to aid in debugging branch specific bugs.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+EXPERIMENTAL_FEATURE(
+      onboardingCtaStyleExperiment, // Feature ID
+      "Experimentations on the onboarding CTA", // Feature name
+      QStringList("color", "fontFamily"),  // Feature variables
+)
+```
+
+The EXPERIMENTAL_FEATURE macro, intentionally doesn't contain the field for declaring if the feature can be flipped on or off. Experimental features must always be "toggleable".
+
+Note Mozilla VPN features may or may not be "toggleable", meaning they may not be allowed to be turned on or off. Experimental features must always be toggleable, otherwise they cannot be applied dynamically. This has consequences for the implementation of the treatment itself. This needs to be considered on a case by case basis.
```
brizental, line 202:
```suggestion
> **Note** Mozilla VPN features may or may not be "toggleable", meaning they may not be allowed to be turned on or off. Experimental features must always be toggleable, otherwise they cannot be applied dynamically. This has consequences for the implementation of the treatment itself. This needs to be considered on a case by case basis.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+Or from QML like so:
+
+```
+MZSettings.onboardingCtaStyleExperiment.color
+MZSettings.onboardingCtaStyleExperiment.fontFamily
+```
+
+Experimental feature settings will only have public getters. They may only be set or removed by the TaskGetFeatureList.
+
+Note In order for this setup to work, experimental features will have to be declared in a separate feature list file e.g. experimentalfeaturelist.h. This will make it possible to import the experimental features only on the settings holder module and create the settings API necessary.
```
brizental, line 222:
```suggestion
> **Note** In order for this setup to work, experimental features will have to be declared in a separate feature list file e.g. experimentalfeaturelist.h. This will make it possible to import the experimental features only on the settings holder module and create the settings API necessary.
```

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+This document details the integration of Nimbus on the Mozilla VPN.
+
+As described on[  "Nimbus for Mozilla VPN"](https://docs.google.com/document/d/12oRkct-l-n9Sq8fzRH1YiiA8CtbEqDxASSpVqC2AKI4/edit#heading=h.lo57q9b2r8iy) Nimbus integration on the Mozilla VPN will be done through a server side integration with[  Cirrus](https://github.com/mozilla/experimenter/tree/main/cirrus).
+
+Nimbus
+------
+
+Nimbus is the experimentation platform built by Mozilla and used by Mozilla products.
+
+Through a web application called "Experimenter", experiments are defined. Through the Nimbus SDKs or Cirrus, applications query Nimbus for the active experiment branches, called "features" on the Nimbus ecosystem. Nimbus takes care of all the experiment targeting logic to serve the application with the correct features active for each given application instance.
```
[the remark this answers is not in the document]
mcleinman: Update: From reading later on in the doc, I believe this to be true.

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+    1.  [Beatriz Rizental](https://github.com/brizental)  My hunch here is that users that have telemetry turned off will not participate in experiments. Wouldn't that break the targeting logic for example when it's based on percentage of users? However if a user is already enrolled in some experiment, we will keep them in that branch until the experiment is over. Otherwise turning data collection off would have unexpected behavior in some cases.
+
+    2.  [Jared Lockhart](https://github.com/jaredlockhart)  Our existing process in other applications is that at the time the user opts out of telemetry and/or experimentation (opting out of telemetry also opts out of experiments, opting out of experiments does not opt out of telemetry), they are unenrolled from the experiment. This can not be managed by the Nimbus infrastructure, it must be implemented by the integration application.  Ie there must be some UI that allows a user to opt out of telemetry and/or experimentation, the application must store that preference, and any application that is opted out must no longer send requests to Cirrus.
```
mcleinman, line 260: I think this should be sketched out in this doc, as it add some complexity.

mcleinman (commented):
This is awesome, thank you! A couple things I'm curious to hear more about.

lesleyjanenorton (approved):
Thank you for all of your hard work on this Bea ✨🧪✨

docs/rfcs/0004-cirrus-experiments.md
brizental: I agree. I will just talk to Santiago about what he feels is the best approach, because this is a product question IMO.

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+That poses an issue for non-logged in users. In these cases, a random UUID will be generated for the experimenter id. (See also question 2. on the "Open questions" section of this document).
+
+Nimbus relies on Glean data for experiment analysis. Therefore the generated experimenter id needs to be included in all Glean pings. The Glean team is working on a new API that will facilitate setting this id, see[  Bug 1848201](https://bugzilla.mozilla.org/show_bug.cgi?id=1848201). The Mozilla VPN application will rely on this API to provide the experiment id to Glean.
+
+Timing: The experimenter id must be set as soon as possible on the application lifecycle, since it is required for all Cirrus requests and should be included in all Glean pings.
+
+#### 2\. Query Nimbus for active features
+
+The /featurelist endpoint is already queried by the application  on initialization and once every hour. Other than these triggers, a feature list request will be made whenever the experiment id is set.
```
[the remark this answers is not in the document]
brizental: Good callouts. For the first question, yes that will be a non-fatal error and whatever are the currently active features will remain. I will update the document with this caveat.

For the second question, that is a great point and I believe this will be feature dependant. If the feature is something like "change the design of the vpn toggle button" or something very disruptive like that we probably don't want to apply it while the user is interacting with the app... But if it's something like extra servers in the servers list we may not care as much. We are able to attach metadata to features, so this could be some boolean flag that features have that state whether it can be applied dinamically or not. I will add this caveat as well. 

Finally for the last question, I will double check with Santiago, but I do not think we should notify users for a restart to apply experiments. If features require a restart they will be applied on next restart and that is all. The feature list is persisted, if this is not documented in the RFC I will include that as well.

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+> **Note** The list of active features will be added to each VPN log summary, to aid in debugging branch specific bugs.
+
+#### 3\. Apply features
+
+Feature flags will be used to apply features dynamically. This binds experiments to the Mozilla VPN release cycle, new feature flags cannot be dynamically set on the client  -- even if the experiment treatment is implemented through addons. Experiments may be designed and implemented in Nimbus/Experimenter before a given feature has been released, however users will only actively see treatments being applied once they are on a version of the application that contains the treatment.
+
+Once the feature list is received from Cirrus, it will be parsed and known features will be turned on while unknown features will be discarded.
```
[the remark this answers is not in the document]
brizental: Answer to this in my previous comment.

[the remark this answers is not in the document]
vinocher: FYI, in this comment I was mainly asking if there is a standard way to mark such features, so the client app can recognize them as restart-required. It looks like the answer is that some metadata will be be attached. Do other products that use Cirrus have standard metadata for this?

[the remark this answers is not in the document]
brizental: It doesn't seem they have a standard mechanism for it, no. As I think about this, I think we do not need a flag for it. If a feature can only be applied on restart, that caveat will simply be implemented in the feature itself e.g. it is applied on initialization only or something of the sort.

docs/rfcs/0004-cirrus-experiments.md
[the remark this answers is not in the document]
brizental: Coming back here after a long time: 1. I talked ot Santiago and he agrees we do not need to notify users for a restart. 2. I think the features themselves will change in the way they are implemented and therefore be applied or not on restart. I think that is not something that neeeds to come as a flag from Cirrus.

docs/rfcs/0004-cirrus-experiments.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Client side implementation
+
+Client side will be responsible for querying the server for active features and applying treatments based on the obtained data.
+
+#### 1\. Generate an experimenter id
+
+In order for Nimbus to run its targeting logic, it requires an experimenter id to be attached to each request.
+
+The experimenter id must be generated per user. It's important that this is user based to ensure users are assigned to the same features in all of their logged in devices.
```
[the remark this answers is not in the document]
brizental: Correct, we are going to use the FxA id here. That will make sure the experimenter id is stable across devices for the same user. I will update the document with this information.

mcleinman (approved):
👍🏻

PR #8300, opened 2023-10-16
VPN-5612 Do not deactivate VPN upon encountering backendfailure

Gela:
## Description

A failed Guardian heartbeat will lead to a back-end failure in the client code, but we should not be ever deactivating the VPN without the user specifically attempting to do so as it can go unnoticed by the user.
This PR removes the call to `backendFailure()` in the connection manager when user is already authenticated if the heartbeat completes with an error code. 

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5612

## Checklist
    
### DIFF
--- a/src/connectionmanager.cpp
+++ b/src/connectionmanager.cpp
@@ -863,12 +863,18 @@ bool ConnectionManager::switchServers(const ServerData& serverData) {
 void ConnectionManager::backendFailure() {
   logger.error() << "backend failure";
 
-  if (m_state == StateInitializing || m_state == StateOff) {
+  if (m_state == StateInitializing || m_state == StateOff ||
+      m_state == StateOn) {
     emit readyToBackendFailure();
     return;
   }
 
   m_nextStep = BackendFailure;
+
+  if (m_state == StateDisconnecting) {
+    deactivate();
+    return;
+  }
 }
 
 #ifdef MZ_DUMMY
--- a/src/connectionmanager.cpp
+++ b/src/connectionmanager.cpp
@@ -872,8 +872,7 @@ void ConnectionManager::backendFailure() {
 
   if (m_state == StateOn || m_state == StateSwitching ||
       m_state == StateSilentSwitching || m_state == StateConnecting ||
-      m_state == StateCheckSubscription || m_state == StateConfirming ||
-      m_state == StateDisconnecting) {
+      m_state == StateCheckSubscription || m_state == StateConfirming) {
     deactivate();
     return;
   }
--- a/src/connectionmanager.cpp
+++ b/src/connectionmanager.cpp
@@ -869,13 +869,6 @@ void ConnectionManager::backendFailure() {
   }
 
   m_nextStep = BackendFailure;
-
-  if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting ||
-      m_state == StateCheckSubscription || m_state == StateConfirming) {
-    deactivate();
-    return;
-  }
 }
 
 #ifdef MZ_DUMMY
--- a/src/connectionmanager.cpp
+++ b/src/connectionmanager.cpp
@@ -863,15 +863,17 @@ bool ConnectionManager::switchServers(const ServerData& serverData) {
 void ConnectionManager::backendFailure() {
   logger.error() << "backend failure";
 
-  if (m_state == StateInitializing || m_state == StateOff ||
-      m_state == StateOn) {
+  if (m_state == StateInitializing || m_state == StateOff) {
     emit readyToBackendFailure();
     return;
   }
 
   m_nextStep = BackendFailure;
 
-  if (m_state == StateDisconnecting) {
+  if (m_state == StateOn || m_state == StateSwitching ||
+      m_state == StateSilentSwitching || m_state == StateConnecting ||
+      m_state == StateCheckSubscription || m_state == StateConfirming ||
+      m_state == StateDisconnecting) {
     deactivate();
     return;
   }
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -1217,7 +1217,8 @@ void MozillaVPN::heartbeatCompleted(bool success) {
   logger.debug() << "Server-side check done:" << success;
 
   if (!success) {
-    m_private->m_connectionManager.backendFailure();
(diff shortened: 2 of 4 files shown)


src/connectionmanager.cpp
```diff
@@ -863,16 +863,15 @@ bool ConnectionManager::switchServers(const ServerData& serverData) {
...
     emit readyToBackendFailure();
     return;
   }
 
   m_nextStep = BackendFailure;
 
-  if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting ||
-      m_state == StateCheckSubscription || m_state == StateConfirming) {
+  if (m_state == StateDisconnecting) {
```
vinocher, line 874: Don't we need to deactivate in the case of real backend failures and health check failures, so we transition to a well defined state? For example backend failure is emitted because of failure to launch the tunnel, handshake with VPN server, any other failures in the daemon. Changing the condition here seems risky because we may end up in an undefined state (activated but not working).

Should we instead stop deactivation only on the Guardian heartbeat failure, because that should not affect the VPN?

src/connectionmanager.cpp
```diff
@@ -872,7 +872,8 @@ void ConnectionManager::backendFailure() {
 
   if (m_state == StateOn || m_state == StateSwitching ||
       m_state == StateSilentSwitching || m_state == StateConnecting ||
-      m_state == StateCheckSubscription || m_state == StateConfirming) {
+      m_state == StateCheckSubscription || m_state == StateConfirming ||
+      m_state == StateDisconnecting) {
```
vinocher, line 876: Is this change needed?

tests/functional/testHeartbeat.js
```diff
@@ -54,21 +49,30 @@ describe('Backend failure', function() {
...
+    await vpn.forceHeartbeatFailure();
+
+    // TODO: without call to click on cancel this test fails because 
+    // we get stuck on the loading spinner 
+    // after authenticating following the changes in mozillavpn.cpp
+    // to remove the call to backend failure in heartbeatCompleted().
+    // Is clicking on "cancel" here okay?
+    await vpn.waitForQueryAndClick(
+      queries.screenAuthenticating.CANCEL_FOOTER_LINK.visible());
+
```
vinocher, line 61: Maybe the solution is to raise the backend failure only on StateAuthenticating, because it does make sense to raise a Guardian failure during authentication. So, something like:
void MozillaVPN::heartbeatCompleted(bool success) {
  logger.debug() << "Server-side check done:" << success;

  if (!success) {
    if (state() == StateAuthenticating) {
        m_private->m_connectionManager.backendFailure();
    }
    return;
  }

src/tasks/heartbeat/taskheartbeat.cpp
```diff
@@ -36,12 +36,16 @@ void TaskHeartbeat::run() {
 
             int statusCode = request->statusCode();
 
+            logger.debug() << "Guardian network request encountered an issue. "
```
vinocher, line 39: Couldn't this be merged with the existing logging on line 32 by addind the status code there?

src/tasks/heartbeat/taskheartbeat.cpp
```diff
@@ -36,12 +36,16 @@ void TaskHeartbeat::run() {
...
+                              "HTTP status code: "
+                           << statusCode;
+
             // Internal server errors.
             if (statusCode >= 500 && statusCode <= 509) {
               vpn->heartbeatCompleted(false);
             }
 
-            // Request failure ((?!?)
+            // Client errors.
```
vinocher, line 48: Is the previous comment better?

src/mozillavpn.cpp
```diff
@@ -1216,7 +1216,9 @@ void MozillaVPN::backendServiceRestore() {
 void MozillaVPN::heartbeatCompleted(bool success) {
   logger.debug() << "Server-side check done:" << success;
 
-  if (!success) {
+  // In the event of a Guardian error during authentication,
+  // the "something went wrong" screen is displayed to the user.
```
vinocher, line 1220: Should the comment explain why (because the service that is used for authentication is unavialble)? Should the comment also say that other heartbeat errors are ignored to prevent unnecessary VPN interruptions?

tests/functional/testHeartbeat.js
```diff
@@ -54,21 +49,26 @@ describe('Backend failure', function() {
     await vpn.waitForQuery(
         queries.screenInitialize.AUTHENTICATE_VIEW.visible());
 
-    await backendFailureAndRestore();
+    // Ensure that in the event of a Guardian error during authentication,
+    // the "something went wrong" screen is displayed to the user.
```
vinocher, line 53: Super nit: extra white spaces at the end of this line.

tests/functional/testHeartbeat.js
```diff
@@ -122,16 +124,15 @@ describe('Backend failure', function() {
...
 
-      await backendFailureAndRestore();
+      await vpn.forceHeartbeatFailure();
 
       await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
       assert.equal(
           await vpn.getQueryProperty(
               queries.screenHome.CONTROLLER_TITLE.visible(), 'text'),
-          'VPN is off');
+          'VPN is on');
```
vinocher, line 133: Should a comment be added here that the VPN should not be disconnected because of heart beat failure?

tests/functional/testHeartbeat.js
```diff
@@ -105,13 +105,15 @@ describe('Backend failure', function() {
...
-      await backendFailureAndRestore();
+      await vpn.forceHeartbeatFailure();
 
       await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
-      assert.equal(
-          await vpn.getQueryProperty(
-              queries.screenHome.CONTROLLER_TITLE.visible(), 'text'),
-          'VPN is off');
+
+      await vpn.waitForCondition(async () => {
```
vinocher, line 112: Should a comment be added here that the VPN should not be disconnected because of heart beat failure?

src/mozillavpn.cpp
Gela: Definitely worth adding a bit more context!

lesleyjanenorton (approved):
✨
Thank you @vinocher @Gela !

vinocher (approved):

PR #8301, opened 2023-10-16
VPN-5697 follow up

mcleinman:
## Description

This comes from additional QA on the ticket.

Per chat on ticket and 1:1 w/ Peter, we're going to remove the "X" button from the 2FA screen, as we also have the "cancel" button to duck out of the flow.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5697

## Checklist
    
### DIFF
--- a/nebula/ui/components/inAppAuth/MZInAppAuthenticationBase.qml
+++ b/nebula/ui/components/inAppAuth/MZInAppAuthenticationBase.qml
@@ -15,6 +15,7 @@ MZFlickable {
     property var _menuButtonOnClick
     property bool _changeEmailLinkVisible: false
     property bool _disclaimersVisible: true
+    property bool _backButtonVisible: true
     property string _viewObjectName: ""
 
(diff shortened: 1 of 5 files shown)


nebula/ui/components/inAppAuth/MZInAppAuthenticationBase.qml
```diff
@@ -59,6 +60,8 @@ MZFlickable {
                 Layout.preferredHeight: MZTheme.theme.rowHeight
                 Layout.preferredWidth: MZTheme.theme.rowHeight
                 Layout.leftMargin: MZTheme.theme.windowMargin / 2
+                enabled: _backButtonVisible
```
MattLichtenstein, line 63: any reason we need to disable AND hide vs just hide?

mcleinman: We absolutely do not, good catch. Thanks, updated.

MattLichtenstein (approved):

Issue #7995, opened 2023-09-12
Narrator's Move Next command loops in new onboarding's StepProgressBar


data-sync-user:
Use the Move Next command (Caps Lock + Right Arrow) of the Windows screen reader, Narrator, to navigate through the new onboarding flow. When it reaches a disabled step in the StepProgressBar, it moves to the view and then loops back to the enabled steps:

!Mozilla VPN 2023-09-11 13-41-25.mp4|width=33.333333333333336%!

This doesn’t repro on Mac.

Focus changes to the view when the disable step is reached, so Narrator follows the focus. The Accessibility Inspect tool shows that the disabled step is enabled for Accessibility on Windows, and may be the cause for this issue.

Tabbing through works fine with the screen reader on Windows.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-5524)
┆Reporter: Vinoo Cherian

data-sync-user:
➤ Raluca Monica Ilban commented:

Hi Vinoo Cherian , could you please add the fixVersion in which this issue was fixed? Thank you!

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this on Mozilla VPN 2.18.0 (2.202310192143) using Windows 10/11 and when using the Caps Lock + Right Arrow keys while the Narrator is enabled, the disabled steps from the progress bar gain focus and are narrated by the screen reader, without changing to another view on the post-login onboarding flow (this also happens on macOS while using the right key) - I will attach a video as well.

Vinoo Cherian, is there something else that we should check here? Should the client behave differently? I am unable to reproduce the reported issue on previous builds as well. Thanks in advance.

cc: Valentina Virlics 

!bandicam 2023-10-20 12-06-45-079.mp4|width=840,height=996!

data-sync-user:
➤ Vinoo Cherian commented:

Bianca Hidecuti The video looks good.

data-sync-user:
➤ Bianca Hidecuti commented:

Thanks, Vinoo. 

Based on previous comments, I am marking this verified as fixed.

PR #8302, opened 2023-10-16
VPN-5524: Mark StepProgressBar's button as not focusable when disabled, Fix a11y name

vinocher:
## Description

1. StepProgressBar's disabled button was mistakenly identified as focusable by the Windows Narrator screen reader, causing Narrator to attempt focusing to it during the Move Next command. This focus attempt failed inside the Qt framework due to the button's disabled state, causing Qt to fall back to setting focus on the parent Focus Scope (StackView). As Narrator follows focus, this resulted in the navigation looping issue described in the bug. A disabled button is expected to be non-focusable for Accessibility, but Qt 6.2.4 has a bug where it used only activeFocusOnTab to determine focusability. (See QAccessibleQuickItem::state.) The fix also sets activeFocusOnTab to workaround this. This issue has been resolved in Qt 6.5.1.

2. The incomplete (disabled) step in StepProgressBar contained a period at the end of the name, "Onboarding step “%1” (%2 of %3)." The Narrator screen reader reads it as 'Dot', likely due to confusion with the  preceding ')'. This issue doesn't occur without a preceding parentheses. Fixed by removing the period.

## Reference

[VPN-5524](https://mozilla-hub.atlassian.net/browse/VPN-5524), [Github 7995](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7995)

## Checklist
    
[VPN-5524]: https://mozilla-hub.atlassian.net/browse/VPN-5524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -568,7 +568,7 @@ onboarding:
     value: "Onboarding step “%1” (%2 of %3). You are currently on this step."
     comment: Accessibility read out for the current step in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
   progressBarAccessibilityStepIncomplete:
-    value: "Onboarding step “%1” (%2 of %3)."
+    value: "Onboarding step “%1” (%2 of %3)"
     comment: Accessibility read out for incomplete future steps in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
   dataSlideHeader: Data collection and use
   dataSlideBody: We strive to provide you with choices and collect only the technical data we need to improve Mozilla VPN. Sharing data with Mozilla is optional.
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -567,9 +567,6 @@ onboarding:
   progressBarAccessibilityStepCurrent:
     value: "Onboarding step “%1” (%2 of %3). You are currently on this step."
     comment: Accessibility read out for the current step in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
-  progressBarAccessibilityStepIncomplete:
-    value: "Onboarding step “%1” (%2 of %3)."
-    comment: Obsolete string, replaced by progressBarAccessibilityStepIncomplete2.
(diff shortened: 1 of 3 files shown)


src/translations/strings.yaml
```diff
@@ -568,7 +568,7 @@ onboarding:
     value: "Onboarding step “%1” (%2 of %3). You are currently on this step."
     comment: Accessibility read out for the current step in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
   progressBarAccessibilityStepIncomplete:
-    value: "Onboarding step “%1” (%2 of %3)."
```
flodolo, line 571: Do we need this change, given it's an accessibility label? This will force retranslation in all locales, and it doesn't seem useful.

vinocher:
> Do we need this change, given it's an accessibility label? This will force retranslation in all locales, and it doesn't seem useful. 

This is the fix for issue 2 in the PR description. The Windows screen reader reads the period as "Dot" because it is confused by the preceding ")". So, I feel that we need this change because it affects the Acessibility experience.

src/translations/strings.yaml
```diff
@@ -568,7 +568,7 @@ onboarding:
     value: "Onboarding step “%1” (%2 of %3). You are currently on this step."
     comment: Accessibility read out for the current step in the step progress bar. %1 is the name of the step. %2 is the index of the current step. %3 is the total number of steps.
   progressBarAccessibilityStepIncomplete:
```
MattLichtenstein, line 570: Modifying an already-translated string requires a new string id. I usually just tack on "2" at the end

src/translations/strings.yaml
vinocher: This is the fix for issue 2 in the PR description. The Windows screen reader reads the period as "Dot" because it is confused by the preceding ")". So, I feel that we need this change because it affects the Acessibility experience.

MattLichtenstein (approved):

flodolo (approved):

PR #8313, opened 2023-10-17
VPN-5005  - Bring MacOS daemon connection implmentation up to date with Linux

brizental:
Note: This is a re-implementation of https://github.com/mozilla-mobile/mozilla-vpn-client/blob/78c485286423def31a624b2943e91c562d17a39e/src/platforms/linux/daemon/dbusservice.cpp#L381 for MacOS platform.

~Opening as draft to get initial feedback as I attempt to write some tests for this.~

[EDIT] I'll add tests in a follow up PR if I am able to write them, just to move this along more quickly. Therefore, taking this out of draft mode.

Two differences to the Linux impl:

1. Check status is authorized to any user when disconnected
2. There is no concept of capabilities on MacOS AFAICT. Therefore only root or the user that activated the VPN can send commands to the daemon, no exception based on user capabilities.


src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,60 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+    // Case: VPN is currently active and caller is **not** one who activated it.
+    // Caller **can't** access the APIs.
+    logger.error() << "Command issued by unexpected user" << uid;
+    return false;
+  }
+
+  // Case: VPN is not connected and we got a request to check the status, fine.
+  if (command == "status") {
+    return true;
+  }
```
brizental, line 211: Allowing this might not actually be necessary if we change how things are done on the local socket controller. As things stand right now the local socket controller has a bunch of checkStatus calls while the VPN is not connected. Allowing this feels harmless, so I left it like this... 🤷‍♀️

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,60 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
   m_socket->write(QJsonDocument(obj).toJson(QJsonDocument::Compact));
   m_socket->write("\n");
 }
+
+bool DaemonLocalServerConnection::isCallerAuthorized(const QString& command) {
+#if defined(MZ_MACOS)
```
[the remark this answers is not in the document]
brizental: Nevermind they jsut stub it. https://github.com/postgres/postgres/blob/2406c4e34ccca697bd5a221f8375f335b5841dea/doc/src/sgml/client-auth.sgml#L1726-L1732.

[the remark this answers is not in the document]
brizental: Calling on @vinocher . Should we file a follow up to do this on Windows? Or do you think this is not even possible?

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,60 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+    return false;
+  }
+
+  if (uid == 0) {
+    logger.debug() << "Command issued by root.";
+    // Case: This is root calling, root is always authorized.
+    return true;
+  }
+
+  QJsonObject status = Daemon::instance()->getStatus();
```
oskirby, line 192: There's a bit of a performance hit, and a race condition waiting to bite us here. In the case where the daemon has been requested to start a connection, but the Wireguard handshake hasn't completed yet, `Daemon::getStatus()` will return false which might result in an unexpected flow through this function.

In the case where at least one connection request has been made, then `Daemon::getStatus()` is going to make an IPC call down to the tunnel process (MacOS/Windows) or Kernel (Linux), which will add unnecessary latency to this check. It would be better to check if a connection has been requested, regardless of the status of that connection. For example, we could test if `Daemon::m_connections.isEmpty()` or by setting a bool on `activate()` and clearing it on `deactivate()`.

src/daemon/daemonlocalserverconnection.cpp
oskirby: I think this is fine, checking the status is idempotent and it doesn't divulge any information that a user wouldn't be able to gather from inspecting the system anyways.

src/daemon/daemonlocalserverconnection.cpp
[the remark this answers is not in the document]
Gela: can we use [this code](https://github.com/postgres/postgres/blob/master/src/port/getpeereid.c) for Windows/platforms without `peereid()`? I think it's unix specific but just in case I'm missing something

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,60 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
   m_socket->write("\n");
 }
+
+bool DaemonLocalServerConnection::isCallerAuthorized(const QString& command) {
+#if defined(MZ_MACOS)
+  uid_t uid;
+  gid_t gid;
+  int result = getpeereid(m_socket->socketDescriptor(), &uid, &gid);
+  if (result == -1) {
+    logger.error() << "Unable to determine if caller is authorized."
```
Gela, line 181: should we log `errno` in this case? http://www.qnx.com/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fe%2Ferrno.html

src/daemon/daemonlocalserverconnection.cpp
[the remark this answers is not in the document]
brizental: That code would fallback to the "no implementation available" branch on Windows.

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,60 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+    }
+
+    // Case: VPN is currently active and caller is **not** one who activated it.
+    // Caller **can't** access the APIs.
+    logger.error() << "Command issued by unexpected user" << uid;
+    return false;
+  }
+
+  // Case: VPN is not connected and we got a request to check the status, fine.
+  if (command == "status") {
```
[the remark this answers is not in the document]
brizental: The no connection check returns an object with the only preoperty `connection`:`false`, while the connected check returns an object with a bit more information about the connection (see: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/daemon/daemon.cpp#L445-L474). I am not really sure if that is a concern, however I am going with the principle of least priviledge possible here just in case.

src/daemon/daemonlocalserverconnection.cpp
brizental: Addressed in [24c36ad](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8313/commits/24c36adb3194aabc3af117c9007f2ccec0b4bfdb). LMK.

src/daemon/daemon.cpp
```diff
@@ -335,6 +341,8 @@ bool Daemon::parseConfig(const QJsonObject& obj, InterfaceConfig& config) {
 bool Daemon::deactivate(bool emitSignals) {
   Q_ASSERT(wgutils() != nullptr);
 
+  m_state = Inactive;
```
brizental, line 344: I am little bit worried about the accuracy of setting this state here. There are multiple paths here where deactivation will fail, when that happens does mean the VPN remain active? Or do we actually go into an error state? Would it make sense to introduce `Error` and `Deactivating` states? @oskirby

src/daemon/daemon.cpp
```diff
@@ -69,7 +71,10 @@ bool Daemon::activate(const InterfaceConfig& config) {
   // If the activation abort's for any reason `the `activationFailure` signal is
   // emitted.
   logger.debug() << "Activating interface";
-  auto emit_failure_guard = qScopeGuard([this] { emit activationFailure(); });
+  auto emit_failure_guard = qScopeGuard([this] {
+    m_state = Inactive;
+    emit activationFailure();
```
Gela, line 76: Could you add a log statements here for activation failure?

src/daemon/daemon.cpp
```diff
@@ -69,7 +71,10 @@ bool Daemon::activate(const InterfaceConfig& config) {
   // If the activation abort's for any reason `the `activationFailure` signal is
   // emitted.
   logger.debug() << "Activating interface";
```
Gela, line 73:
```suggestion
  logger.debug() << "Activating interface. Current state:" << m_state;
```

src/daemon/daemon.cpp
```diff
@@ -57,6 +57,8 @@ Daemon* Daemon::instance() {
 bool Daemon::activate(const InterfaceConfig& config) {
   Q_ASSERT(wgutils() != nullptr);
 
+  m_state = Activating;
```
Gela, line 60: Is it worth checking `m_state` first? I assume we only ever go from inactive->activating so other states could hint at an underlying issue

```c++
if (m_state != Inactive) {
    return false;
}
```

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,59 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+  }
+
+  if (uid == 0) {
+    logger.debug() << "Command issued by root.";
+    // Case: This is root calling, root is always authorized.
+    return true;
+  }
+
+  Daemon::State state = Daemon::instance()->state();
+  if (state != Daemon::Inactive) {
```
Gela, line 194: Why do we check for `if (state != Daemon::Inactive)` here? Regardless of the Daemon state, shouldn't we check whether or not a caller is authorized?

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,59 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+    if (m_sessionUid == uid) {
+      // Case: VPN is currently active and caller is the one who activated it.
+      // Caller can access the APIs.
+      return true;
+    }
+
+    // Case: VPN is currently active and caller is **not** one who activated it.
```
Gela, line 201:
```suggestion
    // Case: VPN is currently active and caller is **not** the one who activated it.
```

src/daemon/daemon.cpp
brizental: It's possible this method will be called when active in the case of server switching.

src/daemon/daemon.cpp
```diff
@@ -69,7 +71,10 @@ bool Daemon::activate(const InterfaceConfig& config) {
   // If the activation abort's for any reason `the `activationFailure` signal is
   // emitted.
   logger.debug() << "Activating interface";
-  auto emit_failure_guard = qScopeGuard([this] { emit activationFailure(); });
+  auto emit_failure_guard = qScopeGuard([this] {
+    m_state = Inactive;
```
brizental, line 75:
```suggestion
    m_state = Inactive;
    logger.error() << "Activation failed";
```

src/daemon/daemonlocalserverconnection.cpp
brizental: Not in the case of `status` or `activate`.

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -160,3 +171,59 @@ void DaemonLocalServerConnection::write(const QJsonObject& obj) {
...
+  if (command == "status") {
+    return true;
+  }
+
+  // Case: VPN is not connected and we just got an activate command.
+  // Anyone can activate and we hold on to the activator's id
+  // for subsequent authorizations.
+  if (command == "activate") {
+    logger.debug() << "New session started by user" << uid;
+    m_sessionUid = uid;
```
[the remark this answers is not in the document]
brizental: I tried to address this comment on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8313/commits/56cc06a8b2ca62dd24f026f005184b6199e03e51. Let me know if the solution is satisfactory!

src/daemon/daemon.h
```diff
@@ -41,6 +42,8 @@ class Daemon : public QObject {
   QString logs();
   void cleanLogs();
 
+  DaemonSession* session() { return &m_session; }
```
vinocher, line 45: Should this move into DaemonLocalServer?

src/daemon/daemonsession.h
```diff
@@ -0,0 +1,36 @@
...
+class QLocalSocket;
+
+class DaemonSession final {
+ public:
+  explicit DaemonSession();
+  ~DaemonSession();
+
+  bool isActive() const { return m_sessionOwner != -1; }
+
+  bool start(QLocalSocket* m_socket = nullptr);
```
vinocher, line 17: Rename m_socket to socket?

src/daemon/daemonsession.h
```diff
@@ -0,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef DAEMONSESSION_H
+#define DAEMONSESSION_H
+
+class QLocalSocket;
+
+class DaemonSession final {
```
vinocher, line 10: Consider if something like DaemonAccessControl or PeerAuthorizationCheckerare better names.

src/daemon/daemonsession.cpp
```diff
@@ -0,0 +1,93 @@
...
+  return m_sessionOwner == uid;
+}
+
+// static
+int DaemonSession::getPeerId(QLocalSocket* socket) {
+#if defined(MZ_MACOS)
+  Q_ASSERT(socket);
+
+  uid_t uid;
+  gid_t gid;
```
vinocher, line 80: Nit: Init these for safety.

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -91,6 +95,24 @@ void DaemonLocalServerConnection::parseCommand(const QByteArray& data) {
...
+      logger.error() << "Attempted to send command:" << type
+                     << "but there is no ongoing session. Ignoring.";
+      return;
+    }
+  } else {
+    if (!session->isPeerAuthorized(m_socket)) {
+      logger.warning() << "Caller is unauthorized. Ignoring command.";
+      return;
+    }
+  }
```
vinocher, line 115: Consider extracting a method for this block.

src/daemon/daemonlocalserverconnection.cpp
```diff
@@ -91,6 +95,24 @@ void DaemonLocalServerConnection::parseCommand(const QByteArray& data) {
   QString type = typeValue.toString();
 
   logger.debug() << "Command received:" << type;
+  auto session = Daemon::instance()->session();
+  if (!session->isActive()) {
+    if (type == "activate" && !session->start(m_socket)) {
+      logger.error() << "Unable to start new session."
+                        "Ignoring activate command.";
+      return;
+      // Other than "activate", only "status" command is available for any user.
```
vinocher, line 104: Should 'status' be allowed for unauthorized users?

src/daemon/daemon.h
brizental: I put it here mostly for convenience. The connection has access to the Daemon already and not to the DaemonLocalServer.

src/daemon/daemonlocalserverconnection.cpp
brizental: It is only allowed when the connection is turned off. I also added this caveat for convenience due to the current state of the code. Some more discussion on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8313#discussion_r1362162963

src/daemon/daemonlocalserverconnection.cpp
brizental: I moved this to the DaemonAccessControl class into it's own method.

src/daemon/daemonaccesscontrol.cpp
```diff
@@ -0,0 +1,135 @@
...
+bool DaemonAccessControl::authorizeCommandForPeer(const QString& command,
+                                                  QLocalSocket* socket) {
+  if (!isSessionActive()) {
+    if (command == "activate") {
+      if (!startSession(socket)) {
+        logger.error() << "Unable to start new session."
+                          "Can't authorize `activate` command.";
+
+        return false;
+      }
```
brizental, line 66:
```suggestion
      }
      
      return true;
```

src/daemon/daemonaccesscontrol.h
```diff
@@ -0,0 +1,42 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef DAEMONSESSION_H
+#define DAEMONSESSION_H
+
+#include <QString>
+
+class QLocalSocket;
+
+class DaemonAccessControl final {
```
Gela, line 12: Can you add a comment to document the purpose of this class?

src/daemon/daemonaccesscontrol.h
```diff
@@ -0,0 +1,42 @@
...
+   * @brief Only the peer that started the session or a peer that has
+   * special permissions e.g. root, may modify the currently ongoing
+   * session.
+   */
+  bool isPeerAuthorized(QLocalSocket* socket = nullptr) const;
+
+ private:
+  // The id of the user who started the currently active session.
+  //
+  // A value of -1 means there is no currently ongoing session.
```
Gela, line 38:
```suggestion
  // A value of -1 means there is currently no ongoing session.
```

src/daemon/daemonaccesscontrol.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  logger.debug() << "Session started by user" << peerId;
+  m_sessionOwner = peerId;
+  return true;
+}
+
+void DaemonAccessControl::resetSession() {
+  if (!isSessionActive()) {
+    return;
+  }
```
Gela, line 51:
```suggestion
  if (!isSessionActive()) {
    logger.debug() << "No active sessions found";
    return;
  }
```

src/daemon/daemonaccesscontrol.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+bool DaemonAccessControl::isSessionActive() const {
+  return m_sessionOwner != -1;
+}
+
+bool DaemonAccessControl::startSession(QLocalSocket* socket) {
+  if (isSessionActive()) {
+    return false;
+  }
```
Gela, line 36:
```suggestion
  if (isSessionActive()) {
    logger.debug() << "New session failed. There is already an active session.";
    return false;
  }
```

src/daemon/daemonaccesscontrol.h
```diff
@@ -0,0 +1,42 @@
...
+ public:
+  explicit DaemonAccessControl();
+  ~DaemonAccessControl();
+
+  bool isSessionActive() const;
+
+  bool startSession(QLocalSocket* socket = nullptr);
+  void resetSession();
+
+  bool authorizeCommandForPeer(const QString& command,
```
vinocher, line 22: Since this returns a bool, is something like isCommandAuthorizedForPeer a better name?

src/daemon/daemonaccesscontrol.h
```diff
@@ -0,0 +1,42 @@
...
+class QLocalSocket;
+
+class DaemonAccessControl final {
+ public:
+  explicit DaemonAccessControl();
+  ~DaemonAccessControl();
+
+  bool isSessionActive() const;
+
+  bool startSession(QLocalSocket* socket = nullptr);
```
vinocher, line 19: Should  QLocalSocket* in this file be changed to const QLocalSocket*? Passing as a const reference would be even better, but not needed because the caller may need to be changed.

vinocher: Shouldn't the null default argument be removed in this file because the implementation expects a non-null argument?

src/daemon/daemonaccesscontrol.cpp
```diff
@@ -0,0 +1,137 @@
...
+        logger.error() << "Unable to start new session."
+                          "Can't authorize `activate` command.";
+
+        return false;
+      }
+
+      return true;
+    } else if (command == "status") {
+      // Other than "activate",
+      // only "status" command is available when there is no ongoing session.
```
vinocher, line 71: Could the comment be expanded to explain why "status" is allow while inactive? i.e. the client app asks for status when the vpn is not activated.

src/daemon/daemonaccesscontrol.h
```diff
@@ -0,0 +1,42 @@
...
+#include <QString>
+
+class QLocalSocket;
+
+class DaemonAccessControl final {
+ public:
+  explicit DaemonAccessControl();
+  ~DaemonAccessControl();
+
+  bool isSessionActive() const;
```
vinocher, line 17: Should isSessionActive and startSession be made private because they aren't called ouside this class?

src/daemon/daemon.h
vinocher: Putting it in Daemon exposes it to the Linux, which doesn't use it (and requires the linux.cmake change) . A potential solution may be to add it to DaemonLocalServer as follows (either return a reference or pointer to the singleton):

```
DaemonAccessControl& DaemonLocalServer::accessControl() {
  static DaemonAccessControl s_accessControl;
  return s_accessControl;
}
```

Then the caller can get it as :

```
  auto accessControl = DaemonLocalServer::accessControl();
  accessControl.resetSession();

```

brizental: Woops, forgot to do this one.

brizental: Done.

vinocher (approved):
Looks good!

cure53:
✔️ LGTM

PR #8320, opened 2023-10-17
Load the tunnel before leaving flow

mcleinman:
## Description

We were leaving the tunnel setup flow after saving the tunnel, but not after loading it. It seems like without loading it, Apple briefly disconnects the tunnel on the next save, which is sending a disconnected message, which we pass on until we turn off the toggle. 

I'm not yet sure if this fixed VPN-5681, but I believe it solves the problem when the app is first loaded.

## Reference

N/A

## Checklist
    
### DIFF
--- a/src/platforms/ios/ioscontroller.swift
+++ b/src/platforms/ios/ioscontroller.swift
@@ -195,15 +195,7 @@ public class IOSControllerImpl : NSObject {
                     return
                 }
 
-                IOSControllerImpl.logger.info(message: "Saving the tunnel succeeded")
-                
-                //Used to create a VPN configuration without connecting during onboarding
-                if isOnboarding {
-                    IOSControllerImpl.logger.info(message: "Finishing onboarding... do not turn on the VPN after gaining permission")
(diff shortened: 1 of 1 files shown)


MattLichtenstein (approved):
nice find!

MattLichtenstein:
have you been able to determine why (though much less apparent and concerning) this happens even when the new onboarding experience is not enabled?

mcleinman:
> have you been able to determine why (though much less apparent and concerning) this happens even when the new onboarding experience is not enabled?

Not as satisfactorily as I'd like. I believe - but don't have enough research done to swear to this - that when we save/load a new VPN config, iOS flashes a `vpnStatusDidChange` to `disconnected`. In this case, we catch this and turn off the toggle. But we're still watching for a successful connection, and so `monitorConnection` sends a signal when it finds the VPN connected - which then turns on the toggle.

If this is accurate, your work in the new onboarding to pre-setup the VPN config without turning it on essentially prevents this bug, as we've already setup the config before turning on the VPN.

PR #8326, opened 2023-10-18
VPN-5733: Fix account deletion word wrapping

MattLichtenstein:
## Description

- For our `MZCheckBoxRow` sublabels, let wrap on a word ending if possible, otherwise, wrap on character (instead of only wrapping on word ending)

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-10-18 at 11 59 50 AM | Screenshot 2023-10-18 at 12 00 04 PM |

## Reference

[VPN-5733: Delete  Mozilla account screen has UI issues for some languages](https://mozilla-hub.atlassian.net/browse/VPN-5733)

### DIFF
--- a/nebula/ui/components/MZCheckBoxRow.qml
+++ b/nebula/ui/components/MZCheckBoxRow.qml
@@ -64,7 +64,7 @@ RowLayout {
             Layout.fillWidth: true
             text: subLabelText
             visible: !!subLabelText.length
-            wrapMode: Text.WordWrap
+            wrapMode: Text.Wrap
             verticalAlignment: label.visible ? Text.AlignTop : Text.AlignVCenter
             Layout.topMargin: label.visible || (!label.visible && lineCount > 1) ? 0 : MZTheme.theme.checkBoxRowSubLabelTopMargin //accounts for the topMargin of the MZCheckbox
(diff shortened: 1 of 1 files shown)


nebula/ui/components/MZCheckBoxRow.qml
```diff
@@ -64,7 +64,7 @@ RowLayout {
             Layout.fillWidth: true
             text: subLabelText
             visible: !!subLabelText.length
-            wrapMode: Text.WordWrap
+            wrapMode: Text.Wrap
```
mcleinman, line 67: Why not `WrapAtWordBoundaryOrAnywhere`?

MattLichtenstein: They appear to be the same thing, though `WrapAtWordBoundaryOrAnywhere` is [not documented](https://doc.qt.io/qt-6.2/qml-qtquick-text.html#wrapMode-prop) for use in QML

mcleinman: Thanks, sorry for missing that documentation.

mcleinman (approved):
🚀

PR #8329, opened 2023-10-18
VPN-5681 - remove iOS VPN profile when logging out or resetting

mcleinman:
## Description

For several reasons (to make "reset and quit" cleaner, to clearly cut someone off after their subscription ends or they log out) we want to remove the VPN config from iOS System Settings. This does that. 

We may want to add this functionality to non-iOS platforms, but I'm leaving that for the future.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5681
This also fulfills https://mozilla-hub.atlassian.net/browse/VPN-4216

## Checklist
    

src/mozillavpn.cpp
```diff
@@ -1290,6 +1292,7 @@ void MozillaVPN::hardReset() {
 void MozillaVPN::hardResetAndQuit() {
   logger.debug() << "Hard reset and quit";
   hardReset();
+  connectionManager()->deleteTunnel();
```
Gela, line 1295: Should the inspector command be updated to do delete the tunnel?

```c++
  InspectorHandler::registerCommand(
      "quit", "Quit the app", 0,
      [](InspectorHandler*, const QList<QByteArray>&) {
        MozillaVPN::instance()->connectionManager()->quit();
        return QJsonObject();
      });
```

src/connectionmanager.cpp
```diff
@@ -256,6 +256,12 @@ void ConnectionManager::quit() {
   }
 }
 
+void ConnectionManager::deleteTunnel() {
+  if (m_impl) {
+    m_impl->deleteTunnel();
+  }
+}
+
```
Gela, line 264: Should we also check and attempt to delete the tunnel in `connectionManager::quit()`?

MattLichtenstein (changes requested):
long overdue safety net here, nice job!

I think we'll also want to delete the tunnel any time we arrive in `App::State::StateInitialize` or `App::UserState::UserNotAuthenticated`. The most notable case I'm thinking of is when a device is deleted from another device and sets those states in `MozillaVPN::reset()`

src/mozillavpn.cpp
mcleinman: This is when the app closes, right? If so, I don't think we want this here. 

On iOS and Android, removing the tunnel config means the user is going to get a system prompt the next time they turn on the VPN. I believe we want to only remove this when a user is signed out. But perhaps I'm misunderstanding what the quit action does here?

src/connectionmanager.cpp
mcleinman: Same comment as above.

src/mozillavpn.cpp
Gela:
```c++
void ConnectionManager::quit() {
  logger.debug() << "Quitting";

  if (m_state == StateInitializing || m_state == StateOff) {
    m_nextStep = Quit;
    emit readyToQuit();
    return;
  }

  m_nextStep = Quit;

  if (m_state == StateOn || m_state == StateSwitching ||
      m_state == StateSilentSwitching || m_state == StateConnecting ||
      m_state == StateCheckSubscription) {
    deactivate();
    return;
  }
}
```

We are deactivating the VPN on quit(), I guess I assumed we would want to remove any lingering process upon deactivation but if that assumption is incorrect we can discard this comment

mcleinman: Also, I'm open to better names then `deleteTunnel`. Maybe `deleteTunnelSystemConfig`? That's too wordy maybe... I first went with `removeTunnel`, but I was worried that sounded too much like breaking down the tunnel - and I wanted to make it clearly destructive/rare. `deleteOSTunnelConfig`?

mcleinman: I'd hope we're killing any lingering processes elsewhere in the code. 😬  This PR is about what is saved to disk for the VPN config, so I wasn't focused on the active VPN breakdown (though I did test to make sure the VPN status didn't affect this deletion).

Gela: I agree `deleteTunnel` and `removeTunnel` are both a bit misleading, it definitely confused me 😁 I don't think `deleteTunnelSystemConfig` is too wordy. Both that and `deleteOSTunnelConfig` are reasonable names IMO

src/mozillavpn.cpp
```diff
@@ -1290,6 +1292,7 @@ void MozillaVPN::hardReset() {
 void MozillaVPN::hardResetAndQuit() {
   logger.debug() << "Hard reset and quit";
   hardReset();
```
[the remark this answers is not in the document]
mcleinman: We start the app in`StateInitialize`, right? If that's the case, wouldn't connecting `StateInitialize` to removing the config end up with some bad side effects? (We could mitigate by only destroying it when we *move into* StateInitialize from another state, but that seems like an antipattern, and is fragile if we ever change the app to start in StatePreInit (and then go into StateInitialize or something like that).

mcleinman:
> long overdue safety net here, nice job!
> 
> I think we'll also want to delete the tunnel any time we arrive in `App::State::StateInitialize` or `App::UserState::UserNotAuthenticated`. The most notable case I'm thinking of is when a device is deleted from another device and sets those states in `MozillaVPN::reset()`

We start the app inStateInitialize, right? If that's the case, wouldn't connecting StateInitialize to removing the config end up with some bad side effects? (We could mitigate by only destroying it when we move into StateInitialize from another state, but that seems like an antipattern, and is fragile if we ever change the app to start in StatePreInit (and then go into StateInitialize or something like that).

src/mozillavpn.cpp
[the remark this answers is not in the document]
mcleinman: Good call. Should we also add it to `MozillaVPN::reset`? I don't see a connection between `hardReset` and `reset`, though I may be missing it.

MattLichtenstein:
> We start the app inStateInitialize, right? If that's the case, wouldn't connecting StateInitialize to removing the config end up with some bad side effects?

haha yea definitely don't want to remove the config on every app launch. what about `App::UserState::UserNotAuthenticated`? that seems appropriately related to when we want to remove the config. I wonder if we unauthenticate before quitting on a reset and quit... doubt it though. maybe there is no one size fits all here and we just need to remove the config on a couple of different occasions

src/mozillavpn.cpp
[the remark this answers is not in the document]
mcleinman: I've added it to `hardReset`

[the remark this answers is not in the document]
MattLichtenstein: I'd say so, since thats the codepath that is taken when a device is removed (providing we can't find a one size fits all solution as mentioned in the comments below)

src/platforms/ios/ioscontroller.swift
```diff
@@ -237,10 +237,21 @@ public class IOSControllerImpl : NSObject {
         TunnelManager.session?.stopTunnel()
     }
 
+    @objc func removeTunnel() {
```
Gela, line 240: I think we're renaming this function too, right? I just wanted to point out I see both `deleteTunnel` and `removeTunnel` in the changes

src/mozillavpn.cpp
mcleinman: I'm going with `deleteOSTunnelConfig` for now - about to update this. Let me know if you prefer something else.

src/platforms/ios/ioscontroller.swift
mcleinman: Yes, thanks. I just updated both `removeTunnel` and `deleteTunnel` to `deleteOSTunnelConfig`.

Gela:
Can we add a functional test for this change? Not sure if there's a way to test toggling on the VPN from systems setting but I think we can write a test for the other scenario `VPN cannot be turned on on a first attempt after a reset and quit action `

mcleinman:
> > We start the app inStateInitialize, right? If that's the case, wouldn't connecting StateInitialize to removing the config end up with some bad side effects?
> 
> haha yea definitely don't want to remove the config on every app launch. what about `App::UserState::UserNotAuthenticated`? that seems appropriately related to when we want to remove the config. I wonder if we unauthenticate before quitting on a reset and quit... doubt it though. maybe there is no one size fits all here and we just need to remove the config on a couple of different occasions

Just looked at this a bit, and it seems like every use of `UserState::UserNotAuthenticated` leads to logging out - it seems unneeded here, but if you want to be extra defensive I can add a listener for the `userStateChanged` signal in `ConnectionManager` and run the "remove profile" code there.

src/platforms/ios/iostunnelmanager.swift
```diff
@@ -47,7 +47,7 @@ class TunnelManager {
         }
     }
     
-    static func withTunnel(_ f: (_ tunnel: NETunnelProviderManager) throws -> Any) -> Any? {
+    @discardableResult static func withTunnel(_ f: (_ tunnel: NETunnelProviderManager) throws -> Any) -> Any? {
```
mcleinman, line 50: This allows us to ignore the result, which gives some cleaner code where we were using this.

mcleinman:
> Can we add a functional test for this change? Not sure if there's a way to test toggling on the VPN from systems setting but I think we can write a test for the other scenario `VPN cannot be turned on on a first attempt after a reset and quit action `

Good call, working on test now.

src/mozillavpn.cpp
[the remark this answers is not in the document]
mcleinman: I think I'm on board, but want to call this out - We have a few error paths that bounce someone back to re-authenticate. In these cases, they log back in and (as of now) they can turn on the VPN by flicking the switch. If we add this iOS system config removal to `reset`, they'll now need to re-add the config (w/ the system dialog box) when they re-activate the VPN after logging back in. On the other hand, it's more defensive code, and prevents future potential issues. Thoughts on those tradeoffs?

tests/functional/helper.js
```diff
@@ -189,6 +193,14 @@ module.exports = {
         `Command failed: ${json.error}`);
   },
 
+  // This is used when hitting the "Reset and Quit" button
+  async clickOnQueryAndAcceptAnyResults(id) {
+    assert(await this.query(id), 'Clicking on an non-existing element?!?');
```
Gela, line 198:
```suggestion
    assert(await this.query(id), 'Element does not exist.');
```

Gela (changes requested):

tests/functional/helper.js
```diff
@@ -189,6 +193,14 @@ module.exports = {
         `Command failed: ${json.error}`);
   },
 
+  // This is used when hitting the "Reset and Quit" button
```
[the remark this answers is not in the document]
mcleinman: Thanks. Done - let me know if you think we need some more here.

tests/functional/helper.js
```diff
@@ -76,16 +76,20 @@ module.exports = {
...
-            queries.screenHome.CONTROLLER_TITLE.visible(), 'text');
-        let unsettled = await this.getMozillaProperty(
-            'Mozilla.VPN', 'VPNConnectionHealth', 'unsettled');
-        return (title == 'VPN is on') && (unsettled == 'false');
-      });
+      awaitConnectionOkay();
     }
   },
 
+  async awaitConnectionOkay() {
```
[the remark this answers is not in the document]
mcleinman: Thanks. Done and done - let me know if you think we need some more here.

Gela (approved):
Wooo! Thanks for figuring out a way to add a test for this too 👌🏽

MattLichtenstein (approved):
nice job!

question though: our functional test suite doesn't currently run on mobile clients, and these changes are iOS specific. is there any benefit to the added functional test?

tests/functional/testSettings.js
```diff
@@ -808,4 +809,49 @@ describe('Settings', function() {
...
+    // can't use waitForQueryAndClick for final click because it returns an
+    // error - as expected, we crashed the app - but that causes test to fail
+    await vpn.clickOnQueryAndAcceptAnyResults(
+        queries.screenDeveloperMenu.RESET_AND_QUIT_BUTTON.visible());
+
+    // Confirm the app quit
+    assert.equal(setup.vpnIsInactive(), true);
+
+    // relaunch app
+    await setup.startAndConnect();
```
MattLichtenstein, line 847: wasm seems to struggle a bit with restarting the client, so I've been excluding these kinds of tests via

``` JS
    if (this.ctx.wasm) {
      return;
    }
```

mcleinman: Thanks for saving me some research on this one. Just added this in.

mcleinman:
> question though: our functional test suite doesn't currently run on mobile clients, and these changes are iOS specific. is there any benefit to the added functional test?

Correct - this is something I mentioned when Gela and I paired on this yesterday. This doesn't actually run on iOS, but in my opinion the benefits are:
- At some point hopefully we'll run functional tests there, and then we'll have this test ready to go.
- It helps us stay true to our "always add tests" mission.

Gela:
@MattLichtenstein In addition to what Matt C said, we are also testing the "Reset and Quit" functionality on Desktop in general which I believe wasn't being done previously.

PR #8336, opened 2023-10-19
Chore - Remove annual plan content from update addons

brizental:
Annual plan feature is shipping in 2.18 behind a feature flag still. Therefore it needs to be gone from the update messages.

### DIFF
--- a/addons/message_update_v2.18/manifest.json
+++ b/addons/message_update_v2.18/manifest.json
@@ -22,16 +22,6 @@
         "content": "This update includes minor bug fixes, UI adjustments and other performance improvements, including:"
       },
       {
-        "id": "c_2",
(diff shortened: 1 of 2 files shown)


MattLichtenstein:
> It's really unfortunate that, once again, we already started localizing these 12 hours ago :-(

@flodolo the string we are removing was already localized a while ago, shouldn't it have skipped needing translations in favor of using machine memory?

fwiw, we do still plan to use this string in the future

flodolo:
It's faster than translating a brand new strings, but it still shows up as a new string and needs to be reviewed by a human.
And it will remain in English until then, unless it's one of the few locales where we have enabled pretranslation (machine translation).

PR #8343, opened 2023-10-19
VPN-5657: Log file location

vinocher:
## Description

   Correct the log file location

## Reference

[VPN-5657](https://mozilla-hub.atlassian.net/browse/VPN-5657)

## Checklist
    
[VPN-5657]: https://mozilla-hub.atlassian.net/browse/VPN-5657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/windows/windowscommons.cpp
+++ b/src/platforms/windows/windowscommons.cpp
@@ -8,6 +8,7 @@
 #include <d3d11.h>
 #include <dxgi.h>
 #include <iphlpapi.h>
+#include <shlobj_core.h>
 
 #include <QDir>
 #include <QHostAddress>
@@ -22,6 +23,8 @@
 #define TUNNEL_SERVICE_NAME L"WireGuardTunnel$mozvpn"
 
 constexpr const char* VPN_NAME = "MozillaVPN";
+constexpr const char* WIREGUARD_DIR = "WireGuard";
+constexpr const char* DATA_DIR = "Data";
 
 namespace {
 Logger logger("WindowsCommons");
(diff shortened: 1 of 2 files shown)


src/platforms/windows/windowscommons.cpp
```diff
@@ -67,22 +70,43 @@ QString WindowsCommons::tunnelConfigFile() {
...
+        QDir wireGuardDataDir(wireGuardDir.filePath(DATA_DIR));
+
+        if (wireGuardDataDir.exists()) {
+          return wireGuardDataDir.filePath("log.bin");
+        }
+      }
+    }
   }
 
   return QString();
```
Gela, line 112: Can you add a `logger.debug() << ...` here to log that the log file path is empty?

src/platforms/windows/windowscommons.cpp
```diff
@@ -67,22 +70,43 @@ QString WindowsCommons::tunnelConfigFile() {
...
 }
 
+// static
 QString WindowsCommons::tunnelLogFile() {
-  QStringList paths =
-      QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation);
+  static QString tunnelLogFilePath = getTunnelLogFilePath();
```
Gela, line 75: for my understanding, are you declaring `tunnelLogFilePath` as static because we don't need to recalculate the path beyond the first initialization?

src/platforms/windows/windowscommons.cpp
```diff
@@ -67,22 +70,43 @@ QString WindowsCommons::tunnelConfigFile() {
...
-    QDir vpnDir(dir.filePath(VPN_NAME));
-    if (!vpnDir.exists()) {
-      continue;
-    }
+  if (SUCCEEDED(
+          SHGetKnownFolderPath(FOLDERID_ProgramFiles, 0, nullptr, &path))) {
+    auto guard = qScopeGuard([&] { CoTaskMemFree(path); });
+    return QString::fromWCharArray(path);
+  }
+  return QString();
```
Gela, line 88: From my understanding, if `SHGetKnownFolderPath` fails to locate the directory, it'll return an empty string. Would it make more sense to eliminate line 88 `return QString();`?

src/platforms/windows/windowscommons.cpp
Gela: Also similar question as above about an empty string being returned

src/platforms/windows/windowscommons.cpp
vinocher: Yes, this is an optimization to reduce file system activity because log file path doesn't change.

src/platforms/windows/windowscommons.cpp
vinocher: SHGetKnownFolderPath can fail  -- it can return an E_FAIL or other error code, causing the SUCCEEDED() macro to return false. We need to return an empty QString() on line 88 to handle that case.

PR #8365, opened 2023-10-23
VPN-5745 - Fix update message contents

brizental:
Tentative fix. Waiting for discussions on whether this is the change we want to go with.


brizental:
Apologies to the reviewers, moving this back to draft because turns out discussion were not over.

lesleyjanenorton (approved):
R+

PR #8371, opened 2023-10-23
VPN-5749: Enable annual upgrades on Stage desktop

lesleyjanenorton:
## Description

Per a request from QA, this PR enables the `annualUpgrade` feature on Stage desktop only. This means that the upgrade path in the Subscription Management view should now be visible by default for someone who: 
- Is on stage &&
- Is on a desktop device &&
- Has a monthly subscription which was purchased through SubPlat (not the App Store, not the Play Store)

## Reference

VPN-5749

## Checklist
    

mcleinman (approved):
👍🏻

PR #8397, opened 2023-10-26
VPN-5745 - Change the id for updated whats new and update messages

brizental:
Yeah, this is hacky. I thought it would be better than changing the string id so late though. Let me know.

Little bonus for addon JavaScript file logging. 

I also reverted the changes from https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8365 just to keep the repo in line with i18n repo.

### DIFF
--- a/src/addons/addonapi.h
+++ b/src/addons/addonapi.h
@@ -25,6 +25,7 @@ class AddonApi final : public QQmlPropertyMap {
 
   Q_INVOKABLE void connectSignal(QObject* obj, const QString& signalName,
                                  const QJSValue& callback);
+  Q_INVOKABLE void log(const QString& message);
 
   /**
    * @brief callback executed when a new AddonApi is created. Use it to add
--- a/addons/message_update_v2.18/manifest.json
+++ b/addons/message_update_v2.18/manifest.json
@@ -17,7 +17,7 @@
     "badge": "new_update",
     "blocks": [
       {
-        "id": "c_1",
+        "id": "c_1-1",
         "type": "text",
         "content": "This update includes minor bug fixes, UI adjustments and other performance improvements."
       },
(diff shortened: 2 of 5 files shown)


mcleinman (dismissed):
Can we trust that every language has a similar `[part we care about], [part we want to remove]` structure? I'm worried that if some languages uses something other than a comma or orders their sentence differently, we'll be left with a really bad message.

brizental:
> Can we trust that every language has a similar [part we care about], [part we want to remove] structure?

100%? No. I looked most translation that have been done up until now and they do follow that pattern.

mcleinman (commented):
Sounds great. If flod is good with this, I'm good with this. Thanks for the creative fix.

vinocher:
> Can we trust that every language has a similar `[part we care about], [part we want to remove]` structure? I'm worried that if some languages uses something other than a comma or orders their sentence differently, we'll be left with a really bad message.

Yes, it seems very likely that this will break in some languages.

flodolo (changes requested):
Sorry, it's a strong no for me. 

This is making a lot on assumptions on how locale structure their translations. 

This, for example, should break your logic right away

> Táto aktualizácia obsahuje drobné opravy chýb, úpravy používateľského rozhrania a ďalšie vylepšenia výkonu vrátane:

flodolo (approved):

src/addons/addonapi.h
```diff
@@ -25,6 +25,7 @@ class AddonApi final : public QQmlPropertyMap {
 
   Q_INVOKABLE void connectSignal(QObject* obj, const QString& signalName,
                                  const QJSValue& callback);
+  Q_INVOKABLE void log(const QString& message);
```
vinocher, line 28: Just curious. since there aren't any callers in this PR -- is future usage expected?

brizental: This is mostly useful for debugging purposes. I used it extensively on the previous approach, but removed all the log calls before submitting it for review. This has been a problem before, this addon scripts are hard to debug without some way to see the logs, see: https://mozilla-hub.atlassian.net/browse/VPN-4055.

PR #8400, opened 2023-10-26
[RFC] VPN-4414: Tech spec for suggested app exclusions

mcleinman:
## Description

This is the tech spec for suggested app exclusions. Some prior discussions occurred on [the Confluence doc](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/430866549/Suggesting+Apps+for+Split+Tunneling) that was deprecated for this PR.

[Rendered tech spec](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/vpn-4414-suggested-exclusion-tech-spec/docs/rfcs/0004-suggested-excluded-apps-tech-spec.md)

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4414

## Checklist
    

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+Epic: [VPN-4412: Suggest Apps to Exclude from VPN Protection](https://mozilla-hub.atlassian.net/browse/VPN-4412)
+
+Context / Problem Statement
+---------------------------
+
+We're looking to create tools that prompt the user in ways that help them stay more secure. Split tunneling (called 'App Exclusion' to users) is a feature that can reduce the need to turn off the VPN entirely by excluding traffic in specific apps from the VPN tunnel, but many users don't know about it. The current feature is in settings, and provides an alphabetical checklist of installed apps:
+
+<img src="./images/0004-01.png" height=500>
+
+This epic is for a new feature to proactively let users know when they have apps that could benefit from the split tunneling feature.
```
lesleyjanenorton, line 13: Could we add a slightly more detailed overview of how this feature works? I think this can mostly be inferred from the tasks but it would be helpful to have it summarized somewhere.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+    -   Additions to the list should be given the same amount of thought that adding new dependencies to our repo are given. That is, full research by our engineers to ensure nothing malicious is being snuck in. And erring on the side of not including any.
+        -   This statement (or something similar) should be added to the top of the file in a comment.
+    -   From Santiago: <https://github.com/citizenlab/test-lists/tree/master/lists>, <https://ooni.org/>
+    -   Santiago list for Android: <https://docs.google.com/spreadsheets/d/1j2REPQkcdPg5bNW5B3T-L8dnlNPPwAms1WlBiYvVwoc/edit#gid=0>
+-   (8-13 points) Deep linking into app for notification to work ([VPN-5034](https://mozilla-hub.atlassian.net/browse/VPN-5034)): Currently, our navigation code allows us to switch tabs, but does not permit moving to a deeper screen - so we can move a user to the settings screen, but can't move them into a menu or submenu within the screen (like App Exclusions). We need deep linking to be able to open the notification where we want.
```
lesleyjanenorton, line 44: We can refactor the App Exclusions view to be a `Screen` and call it easily enough with the Navigator as we do for `ScreenGetHelp` and others.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+We'd expect the `apps_excluded` metric to increase after this is implemented on platforms which have app exclusions. Both the number of people with more than 0 apps excluded and the average count of apps excluded should increase.
+
+If Nimbus is available by the time this feature is complete, we will use it to roll out the test.
+
+Security / Privacy
+------------------
+
+Since all the work is being done on device, the only security concerns are ensuring we don't log any private info.
+
+The excluded apps list will likely live in the Guardian repo. There are ways to weaponize the suggested apps list to weaken privacy of key apps, but for this attack vector to be used a Mozilla staffer would need to approve the PR with the bad-faith addition to the suggested app list.
```
lesleyjanenorton, line 127: Should we update this paragraph? On line 47 the suggestion is to keep the list in the client repo.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+
+Rollout considerations
+----------------------
+
+While the reworked app exclusion page will affect all platforms that support split tunneling, notifications will only be shown to users who have already allowed our app to send them. On mobile, we show the "approve notifications" screen immediately upon launching the app for the first time, which likely depresses approval rates. We may want to re-consider how we show this as part of onboarding.
```
lesleyjanenorton, line 132: More of a product question but should this be something users can specifically turn off? Maybe via the App Exclusions page or in App permissions -> Notifications?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+-   (3-5 points) Create a task for checking for getting suggested app exclusions, ensure this is well tested
+    -   The list of suggested app exclusions will live in the VPN client repo, and be served via addons.
+
+
+### Create daily task in app (8-12 points total)
+<img src="./images/0004-02.png" width=800 alt="New daily task">
+
+-   (3-5 points) Daily task to check for suggested apps that are installed on the device (after one week from first install), halt task if don't have notification permissions or if the platform doesn't have app exclusions feature. (Create dev option to immediately pull down the new list.)
+    -   This will be run by the app for desktop, and a [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) on Android. (iOS does not have split tunneling.)
+        -   After running the job, we will create a 24 hour countdown timer to run the job next. Additionally, we'll record the time for this job as a setting, and we'll check this timestamp on app/daemon launch - if more than 24 hours have passed we'll
```
lesleyjanenorton, line 55: Have we considered how we might avoid spamming the user in cases where they have more than 1 app we suggest they exclude? Maybe the list schema could include some kind of priority/ranking to help us make more impactful suggestions first?

lesleyjanenorton: Just realized my question above assumes we intend to send one notification per app suggestion, maybe that is not the case.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+
+Open Questions
+--------------
+
+-   What do we use to canonically track apps we may want to exclude? Is there a bundle ID? The name of app (fragile)? Something else? Do we use different IDs across platforms?
+-   How hard would these checks be to run in Android daemon or PeriodicWorkRequest? (If substantially harder, Santiago open to doing it in Android app - which reduces how often it runs, as the app must be open in order for it to run.) On desktop platforms, the app is always active when the VPN is active. This is not true on mobile, which is why this is considered separately.
+-   Any easy ways to categorize games? (See [VPN-4433](https://mozilla-hub.atlassian.net/browse/VPN-4433).)
+-   How will we decide which apps are initially on the suggested list?
+-   Does the suggested list just include specific apps, or can it include entire directories (like well-known Windows game directories) and/or allow wildcard matching?
+
```
lesleyjanenorton, line 29: - How can we validate apps before adding them to the list to be sure we don't recommend excluding an app with egregious tracking practices?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+----------------------
+
+While the reworked app exclusion page will affect all platforms that support split tunneling, notifications will only be shown to users who have already allowed our app to send them. On mobile, we show the "approve notifications" screen immediately upon launching the app for the first time, which likely depresses approval rates. We may want to re-consider how we show this as part of onboarding.
+
+Test Plan
+---------
+
+-   Fresh install of the app on a device that has a suggested app. Ensure that app is shown appropriately on the screen in VPN Settings. Ensure a notification suggesting its exclusion is popped at an appropriate time.
+-   Install an app on the suggested list on the device. Ensure the app is now shown appropriately on the screen in VPN settings. Ensure a notification suggesting its exclusion is popped at an appropriate time.
+-   Add an app on the device to the list on the repo, wait, and ensure it is offered as an app to exclude within 24 hours.
```
lesleyjanenorton, line 139: It might be helpful for QA to have a more expedient way to test this  - maybe an Inspector Command, or input in the dev menu, by which they can write new apps to the list at will?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Other options considered for suggested app files
+
+The goals for the file format:
+-   Easy to read by humans
+-   Easy for humans to add to, no matter their level of technical proficiency
+-   Easy for us to parse in the app
+
+Initially, there were two additional goals (no longer considerations):
+-   To have one file for all platforms, so that we'd remember to add a specific app to all platforms at once. With a conversation around potentially wanting different suggested app categories on mobile, this seems unimportant.
```
lesleyjanenorton, line 106: Curious why a single file is no longer a goal? Seems desirable but not essential.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,141 @@
...
+-   Allowing comments in the file. While we still could structure a text file to allow comments, this became less important when the plan became "one file per platform, one app per line".
```
lesleyjanenorton, line 107: Also curious why this is not a goal. Allowing comments would I think make parsing the list for human beings a lot easier.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: That only allows us to present the screen, right?

If so - We may want to be switching to the Settings tab beforehand (awaiting word from Santiago). That's possible, but not in the app today, right?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Yes, thanks! I missed a spot. Will be updated with the next push.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Thanks, yes - adding that with next update

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: From discussions w/ Santiago, we pop the same message for 1 or more than 1 new app, so that doesn't apply here. Next push will include some clearer text here.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Will be discussed on Monday's meeting

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Good call - I'll add to this doc a dev menu option to immediately run the daily job.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
lesleyjanenorton: > That only allows us to present the screen, right?

Correct. A truly nested deep link (in this case, where clicking back takes you to the settings main menu) is something that would have to be implemented.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: It seemed important when I thought we were going to exclude the same apps (or types of apps) on all platforms. But it comes at the expense of complexity. And after talking to Santiago, it seems like we may want to suggest excluding very different apps on desktop vs. mobile, so this makes less sense now.

Per a paragraph or two later, I was also a little stuck on how to make volunteer contributions easy, and what format best allowed that. I'm open to suggestions; I don't feel strongly on this one.

That said, I'm curious what comes out of Monday's meeting - we may end up adjusting this based on that discussion.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Fair enough. I'll move that back in.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Confirming w/ Santiago how he wants to handle this.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,142 @@
+Suggesting Apps for Split Tunneling
```
brizental, line 1: This document is missing the header components defined in the template. See: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/rfcs/0001-rfc-process.md#guide-level-explanation.

brizental: Also, the PR title is missing the `[RFC]` prefix as defined in the same document I linked before.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,142 @@
...
+
+Implementation (estimated 40-65 points total)
+---------------------------------------------------
+
+This work is broken into four sections, with the bulk of code coming in the middle two sections. Those two sections should each be built behind their own separate feature flag, though we'd expose both parts together via Nimbus.
+
+### Prerequisites (21-36 points total)
+
+-   (3-5 points) Answer the open questions above
+-   (5-8 points) Create a well-structured file (for our own research, as well as one users can add to via GH pull requests on the client repo) w/ apps to exclude (and update the repo's readme to explain this new type of contribution). Add this doc to addons. Add tests to ensure the doc's format integrity is maintained. This format should be designed to allow new platforms (like websites) to be added without breaking anything.
```
brizental, line 38: > Add this doc to addons.

Continuing the discussion from the Confluence.

How we are going to deploy this thorugh the addons framework needs to be better detailed and probably spun up into it's own task or group of tasks. Although in practice addons are just .rcc files served statically, they have a well defined structure in code, see: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/add-on.md#add-on-format.

Sure, .txt files can be added to .rcc files. Is that the best way to approach this though? Why .txt file and not something more easily digested from C++ code, akin to our featurelist or settinglist files? Even then, there are a few steps before that need to be figure out. Is this going to be a new type of addon? What type is that? What does the manifest.json for the new addon type look like? Is it going to be one addon per platform, one addon per suggested exclusion or one addon for everything? And so on.

I'd also question, if it is it absolutely crucial that this list have the dynamic properties of addons. We have been discussing for the past months the removal of a lot of the addons code. See VPN-5545 and most importantly VPN-5419. What are the advantages of serving this through addons? Is it worth the complexity trade-off over just having a file that is updated on each release?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Updated the title, thanks. I think I'm missing the reference in the doc to "header components" - sorry to be dense, but can you point out what you mean? (What I do see on the doc is: `I'd prefer we not be overly prescriptive about it, at least early on.`)

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: I think in light of your comments, and your reasonable request to want another set of tasks for this, we should consider this part dead, and fallback on the "just include the files in each release" plan. Will confirm with Santiago, and leave this open for 24 hours or so to ensure others can comment if they want to argue for serving this up dynamically.

mcleinman: @albionx ^

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
brizental: Hm, you are right. I seem to remember from the discussion on that first RFC that the first RFC itself would be used as a template, but if it's not in the doc it's not 🤷 I meant the list of stuff right at the top, but feel free to ignore.

mcleinman:
I believe I’ve addressed all open comments on the PR as well as the [confluence page](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/430866549/Suggesting+Apps+for+Split+Tunneling) (where this was originally hosted, and now has an outdated version of the doc with some lingering comment threads).
- I’ve added to the open questions section, based on Monday’s meeting. Several of the policy questions were ones that Santiago said he’d do some research on, I believe. I don’t believe these all need to be answered before we approve the RFC, but feel free to disagree with that.
- I’d like to propose closing this doc on Monday Nov 6th in the morning (west coast time). If anybody feels that this is not enough time to finish up discussions on this doc, please let me know - it’s flexible.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+The goals for the file format:
+-   Easy to read by humans
+-   Easy for humans to add to, no matter their level of technical proficiency
+-   Easy for us to parse in the app
+-   Allowing comments in the file. While we still could structure a text file to allow comments, this became less important when the plan became "one file per platform, one app per line".
+
+Initially, there was one additional goal (no longer consideration):
+-   To have one file for all platforms, so that we'd remember to add a specific app to all platforms at once. With a conversation around potentially wanting different suggested app categories on mobile, this seems unimportant.
+
+Other options considered included CSV, YAML, and JSON. While YAML could be useful (as it's easy for less-technical people to read), we don't have a good YAML parser in the app yet. We could add a library, of course. However, we'd need to write our own document format test, because the YAML file (or JSON or CSV) would need to be structured in a specific way.
```
brizental, line 107: Qt provides a JSON parser out of the box. Using JSON would allow us to use JSON Schema to define the format of the file and would make it very easy to test. That would also not be anything new for the codebase, we currently do exactly that for addon manifests and the schema check is run on the linter CI job already. See https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/.github/workflows/linters.yaml#L43-L45.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (8-13 points) Deep linking into app for notification to work ([VPN-5034](https://mozilla-hub.atlassian.net/browse/VPN-5034)): Currently, our navigation code allows us to switch tabs, but does not permit moving to a deeper screen - so we can move a user to the settings screen, but can't move them into a child menu or child's child menu within the screen (like App Exclusions). We need deep linking to be able to open the notification where we want.
+-   (2-5 points) Create initial list of recommended app exclusions
+
+### Create daily task in app (9-13 points total)
+<img src="./images/0004-02.png" width=800 alt="New daily task">
+
+-   (3-5 points) Daily task to check for suggested apps that are installed on the device (after one week from first install), halt task if don't have notification permissions or if the platform doesn't have app exclusions feature. (Create dev option to immediately pull down the new list.)
+    -   This will be run by the app for desktop, and a [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) on Android. (iOS does not have split tunnel.)
```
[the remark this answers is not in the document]
mcleinman: @strseb Can you speak to why you suggested a `PeriodicWorkRequest`? I'm having trouble finding your original comment about why it was well-suited.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Day 2
+
+These are important components that we are committed to delivering, but are outside the scope of this epic. They are items that will be done as soon as the epic is closed, but do not block the feature on shipping.
+
+-   Engineering metrics
+    -   Metrics will be created for a successful deep linking, as well as an erroneous deep linking (the app cannot figure out how to resolve the deep link).
+    -   Add performance metrics around load time of the Excluded Apps screen.
+    -   Create a metric for when the daily job was successful, and another one for when it fails.
+    -   A metric should be recorded when the notification is sent
```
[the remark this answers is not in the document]
mcleinman: This was Adrienne's recommendation to try to slim down implementation. I'm in agreement that this is small, but if we're trying to hold to a smallest possible MVP these do seem like the type of thing to get booted to post-launch. Realistically, this will likely go out to users at the same time (unless there is a release branch cut in between the main work and getting this phase two work done).

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (2-5 points) Create initial list of recommended app exclusions
+
+### Create daily task in app (9-13 points total)
+<img src="./images/0004-02.png" width=800 alt="New daily task">
+
+-   (3-5 points) Daily task to check for suggested apps that are installed on the device (after one week from first install), halt task if don't have notification permissions or if the platform doesn't have app exclusions feature. (Create dev option to immediately pull down the new list.)
+    -   This will be run by the app for desktop, and a [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) on Android. (iOS does not have split tunnel.)
+        -   After running the job, we will create a 24 hour countdown timer to run the job next. Additionally, we'll record the time for this job as a setting, and we'll check this timestamp on app/daemon launch - if more than 24 hours have passed we'll
```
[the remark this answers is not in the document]
mcleinman: Eeek, thanks. Adding the end of that sentence, and something about user clocks in the next push (coming shortly).

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+    -   Create a dev option to pull a new suggested apps list immediately.
```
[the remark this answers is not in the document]
mcleinman: Thanks. Fixed in the next push.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (3-5 points) Is there one or more new apps?
+    -   Check against setting - if we have ever sent a notification for this app before, don't send one (this should be done off the app's domain name or something, not marketing name - that can change)
+    -   Pop an OS-level notification (on Android, use a suggestion channel). The text is intentionally generic, and makes sense if there is one new app or multiple.
+    -   Add new app(s) to "have sent notification" list
+    -   (Create a dev option to clear the "have sent notification" list. Create another dev option to immediately run the daily check job.)
```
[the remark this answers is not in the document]
mcleinman: Good call. Done in next push.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (1 point) Add this notification type to the settings screen that allows per-type opt out (App permissions -> Notifications)
+-   (2 points) Ensure tapping notification leads to correct screen
+
+### Improve existing split tunnel screen (10-16 points total)
+[Figma mocks](https://www.figma.com/file/UZYzma7hlcfE5ke3z8jGbN/App-exclusions-suggestions?type=design&node-id=196-6366&mode=design&t=RL1hdfBQLMS1rKVa-0)
+
+<img src="./images/0004-03.png" width=500 alt="Updated flow for existing App Exclusion screen">
+
+-   (2 points) Run the task on the launch of screen
+    -   Update the "have seen these apps" settings list with all suggested apps that are on the system. (This means that if a new app is seen on the Settings screen before the daily job runs and we send a notification, we never send a notification. This was confirmed by Santiago.)
```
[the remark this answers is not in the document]
mcleinman: That is correct, adding some clarification. (And remember that while the notification is popped because of a new app or apps, that app (or apps) are never mentioned by name in the notification to the user.)

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (1-2 points) Add recommended section (only if 1 more more apps on the system are suggested) - pull any apps in the "have seen them" list
+-   (1 point) Sort recommended section
+-   (1 point) Sort "others" section
+-   (1-2 points) Update sort when something is added or removed
```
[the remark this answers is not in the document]
mcleinman: Correct. Adding clarifying language.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,139 @@
...
+-   (2 points) Ensure tapping notification leads to correct screen
+
+### Improve existing split tunnel screen (10-16 points total)
+[Figma mocks](https://www.figma.com/file/UZYzma7hlcfE5ke3z8jGbN/App-exclusions-suggestions?type=design&node-id=196-6366&mode=design&t=RL1hdfBQLMS1rKVa-0)
+
+<img src="./images/0004-03.png" width=500 alt="Updated flow for existing App Exclusion screen">
+
+-   (2 points) Run the task on the launch of screen
+    -   Update the "have seen these apps" settings list with all suggested apps that are on the system. (This means that if a new app is seen on the Settings screen before the daily job runs and we send a notification, we never send a notification. This was confirmed by Santiago.)
+-   (1-2 points) Add recommended section (only if 1 more more apps on the system are suggested) - pull any apps in the "have seen them" list
```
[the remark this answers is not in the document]
mcleinman: No, sorry - poor wording. I've updated this.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Thanks, I didn't know about that. I've updated this statement to be more accurate.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,142 @@
...
+The goals for the file format:
+-   Easy to read by humans
+-   Easy for humans to add to, no matter their level of technical proficiency
+-   Easy for us to parse in the app
+-   Allowing comments in the file. While we still could structure a text file to allow comments, this became less important when the plan became "one file per platform, one app per line".
+
+Initially, there was one additional goal (no longer consideration):
+-   To have one file for all platforms, so that we'd remember to add a specific app to all platforms at once. With a conversation around potentially wanting different suggested app categories on mobile, this seems unimportant.
+
+Other options considered included CSV, YAML, and JSON. While YAML could be useful (as it's easy for less-technical people to read), we don't have a good YAML parser in the app yet. We could add a library, of course. However, we'd need to write our own document format test, because the YAML file (or CSV) would need to be structured in a specific way. While we already have a JSON parser that we've custom wrote that could be used for this validation, a simple list is better than JSON for the first two goals of the file format.
```
brizental, line 110: We did not custom write it. We just wrote the schema, but [JSON schema](https://json-schema.org/) is not a tool we created. 

I believe it would probably be more reliable and easier to use JSON here. Because Qt already has tools to parse JSON and it is trivial to write a simple schema and validate it.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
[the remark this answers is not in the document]
strseb: At that point in time, I was under the impression that this notification can pop up at any time, therefore we cannot expect the daemon to be alive and need a scheudled OS task. However this expectation no longer be accurate :) 
If it's fine that it only happens if the user is connected (aka daemon alive) - we do not need to haggle with that.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: I don't see how JSON is more reliable and easier than a file with one entry per line, especially if we want to make it simple for less-technical people to contribute to this. Can you explain?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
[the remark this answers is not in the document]
mcleinman: Thanks. Updating this to use the daemon for Android.

mcleinman:
With the meeting Wednesday to discuss how items get on the list, I believe the only outstanding question for this doc is whether we're going to use a simple text file or JSON. I've asked a question above about how JSON helps meet our goals.

I'm planning on closing this doc tomorrow morning - hopefully that gives enough time for response for the last piece.

(And if there are more outstanding questions that should be agreed upon before locking this doc, please let me know.)

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
brizental: Technically I believe it would be easier because the tools for parsing and validating are already there. 

With regards to less technical people contributing, we could go with an extremely minimal JSON format here. Say we decide to keep almost exactly what is currently being proposed for `txt` files and have `android.json`, `windows.json` etc. The JSON format could look like:

```json
[
   "my.app.identifier",
   "my.other.app.identifier"
]
```

And so on, which just adds the quotes, commas and brackets more than what the `txt` . However, with the JSON schema we would be able to validate the format and create regular expression to validate the format of each identifier per platform without much overhead e.g. the JSON schema for `android.json` would look something like:

```json
{
  "type": "array",
  "items": {
    "type": "string",
    "pattern": "^[a-zA-Z][a-zA-Z0-9_]*(\\.[a-zA-Z][a-zA-Z0-9_]*)*$"
  }
}
```

mcleinman: Thanks for all this. I'll update the doc today or tomorrow.

mcleinman: Okay, I think I've come around. I'll update this doc today or tomorrow.

mcleinman:
The format for the doc of suggested apps has been updated. 

This document will close in 24-48 hours. If there is agreement on this doc format, it will close in 24 hours. If there is further discussion, I'd like to resolve it within 48 hours to finalize this doc.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Prerequisites (15-26 points total)
+
+-   (3-5 points) Answer the open questions above
+-   (5-8 points) Create a well-structured file (for our own research, as well as one users can add to via GitHub pull requests on the client repo) w/ apps to exclude (and update the README in the repo to explain this new type of contribution). These docs will live in our repo, and be bundled with the app. Add tests to ensure the doc's format integrity is maintained. This format should be designed to allow new platforms (like websites) to be added without breaking anything.
+    - The document format will be a single JSON file for all platforms. Within each platform there will be an array of IDs for the suggested apps. These will each be an array, and we'll only look at the first item in the array - if a second string exists in the array will be considered comments (as JSON doesn't have the concept of comments). (This is preferred to keys/values for binary/comments compactness/legibility.)
```
vinocher, line 39: Could this clearly call out that the suggested list is bundled with the app? Does 'docs' mean the list of the list + README? Currently this  is called out in the "Other Options Considered" section, but it isn't obvious to the user at this point in the doc.

vinocher: It sounds like this list will be updated only when the app is upgraded. If so, could that be mentioned?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+-   Allowing comments in the file. While we still could structure a text file to allow comments, this became less important when the plan became "one file per platform, one app per line".
+
+Initially, there was one additional goal (no longer consideration):
+-   To have one file for all platforms, so that we'd remember to add a specific app to all platforms at once. With a conversation around potentially wanting different suggested app categories on mobile, this seems unimportant.
+
+Other options considered included CSV, YAML, and plain text files. While YAML could be useful (as it's easy for less-technical people to read), we don't have a good YAML parser in the app yet. We could add a library, of course. However, we'd need to write our own document format test, because the YAML file (or CSV) would need to be structured in a specific way. 
+
+Ultimately, JSON version seemed simpler (and thus easier to build) - there aren't the same parsing concerns and it's easy for contributors.
+
+Metrics
```
vinocher, line 125: Sounds like a rollback of the list, because we accidentally added a bad app, will require an VPN app upgrade. Could that be mentioned here?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+    - The document format will be a single JSON file for all platforms. Within each platform there will be an array of IDs for the suggested apps. These will each be an array, and we'll only look at the first item in the array - if a second string exists in the array will be considered comments (as JSON doesn't have the concept of comments). (This is preferred to keys/values for binary/comments compactness/legibility.)
+    ```
+    {
+        "android":
+            [
+                ["my.app.identifier", "an optional comment"],
+                ["my.other.app.identifier"]
+            ],
+        "linux":
+            [
```
vinocher, line 48: Given that the suggested app list is bundled with the VPN app and the number of apps on a device is unlikely to change very much after the device's initial setup period, should we reduce the frequency of checks for better power management on desktop devices?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
vinocher: Also, does the design of a rollback need to be explicly called out?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+                ["whatever.identifier.linux.uses", "comments are still optional"]
+            ]
+    }
+```
+    -   Additions to the list should be given the same amount of thought that adding new dependencies to our repo are given. That is, full research by our engineers to ensure nothing malicious is being brought in. And erring on the side of not including any.
```
brizental, line 53: nit: Consider calling out that defining the actual steps / requirements for approval of new apps to the list is out of scope for this document, but will be done is a follow up workflow. I think that would be important, to be clear that we don't plan to handwave this -- "full research by our engineers to ensure nothing malicious is being brought in" is quite vague, which is fine for this document, but not fine for the feature itself.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+        -   This statement (or something similar) should be added to the top of the file in a comment.
+    -   From Santiago: <https://github.com/citizenlab/test-lists/tree/master/lists>, <https://ooni.org/>
+    -   Santiago list for Android: <https://docs.google.com/spreadsheets/d/1j2REPQkcdPg5bNW5B3T-L8dnlNPPwAms1WlBiYvVwoc/edit#gid=0>
```
brizental, line 56: nit: IIRC, these are being added as reminders, not suggestions. Consider calling that out explicitly.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+    - Immediately run the daily check job.
+
+### Improve existing split tunnel screen (10-16 points total)
+[Figma mocks](https://www.figma.com/file/UZYzma7hlcfE5ke3z8jGbN/App-exclusions-suggestions?type=design&node-id=196-6366&mode=design&t=RL1hdfBQLMS1rKVa-0)
+
+<img src="./images/0004-03.png" width=500 alt="Updated flow for existing App Exclusion screen">
+
+-   (2 points) Run the task on the launch of screen
+    -   Update the "have seen these apps" settings list with all suggested apps. (If a new suggested app is install and the Settings screen is opened by a user before the daily job runs to send a notification, we thus never send a notification. This behavior was confirmed by Santiago.)
+-   (1-2 points) Add recommended section (only if 1 more more apps on the system are suggested). Since we've just updated the settings list of recommended apps on the system (in prior bullet point), use this settings list - if any system app we discover also appears on that list, it goes in the recommended section.
```
brizental, line 85:
```suggestion
-   (1-2 points) Add recommended section (only if 1 or more apps on the system are suggested). Since we've just updated the settings list of recommended apps on the system (in prior bullet point), use this settings list - if any system app we discover also appears on that list, it goes in the recommended section.
```

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+-   (1 point) Sort recommended section - selected ones at top alphabetically, then the reset alphabetically
+-   (1 point) Sort "others" section - selected ones at top alphabetically, then the reset alphabetically
+-   (1-2 points) Update sort when something is checked or unchecked. The list dynamically updates when a user is on it.
+-   (1-2 points) Ensure search works, shows one section, is alphabetically sorted
+-   (2-5 points) Ensure accessibility is up to date
+-   (1 point) Add metrics for how many suggested (and non-suggested) apps are excluded\
```
brizental, line 91:
```suggestion
-   (1 point) Add metrics for how many suggested (and non-suggested) apps are excluded
```

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,153 @@
...
+
+Other options considered included dynamically serving the suggested lists from addons or our repo or Guardian, each of which added complexity for hosting/validating. With the rarity of the list being updated (and lack of time-sensitivity when it does), this draft bundles the lists with the app. Hosting the list in Guardian was difficult (which is a private repo, so the list would need to be hosted elsewhere to allow public suggestions).
+
+### Other options considered for suggested app files
+
+The goals for the file format:
+-   Easy to read by humans
+-   Easy for humans to add to, no matter their level of technical proficiency
+-   Easy for us to parse in the app
+-   Allowing comments in the file. While we still could structure a text file to allow comments, this became less important when the plan became "one file per platform, one app per line".
```
brizental, line 116: IIUC the plan has gone back to one file for all platforms.

brizental (approved):
LGTM! My concerns are all addressed. Thank you for the productive discussions, Matt! I don't feel confident being the only approver here though as split tunneling / app exclusions is not a part of the code I am super familiar with 😅 Just FYI.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: That's a typo from when we were considering multiple files, sorry - I've updated it to `doc`.

I'm holding off on updating the "where we're hosting" piece pending Monday's discussion.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Added a section; waiting to write it until after Monday's meeting.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Sounds like we're not bundling with the app at this point.

There are two reasons we'd want to alert a user: The user installed a new app (which is on the suggested list), or a the suggested list being updated with an app the user already has installed. I don't think this is going to be a very computationally expensive run (especially compared with how frequently we run health checks), and 24 hours feels like an okay cadence. I don't have any data for this, however. 

Do you think it will be somewhat expensive? If so, what do you suggest for a check cadence?

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: 👍🏻

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: thanks, done

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Thanks, good catch. Fixed.

mcleinman:
Update on timing: Several of us are meeting next Monday to talk through some of the last pieces. This document will stay open at least until then.

mcleinman:
I believe all outstanding questions have been addressed at this point.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,156 @@
...
+                ["my.other.app.identifier"]
+            ],
+        "linux":
+            [
+                ["whatever.identifier.linux.uses", "comments are still optional"]
+            ]
+    }
+    ```
+    -   Additions to the list should be given the same amount of thought that adding new dependencies to our repo are given. That is, full research by our engineers to ensure nothing malicious is being brought in. And erring on the side of not including any.
+        -   This statement (or something similar) should be added to the top of the file in a comment.
```
brizental, line 51: nit: This is not possible with JSON.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,156 @@
...
+    - Clear the "have sent notification" list.
+    - Immediately run the daily check job.
+
+### Improve existing split tunnel screen (10-16 points total)
+[Figma mocks](https://www.figma.com/file/UZYzma7hlcfE5ke3z8jGbN/App-exclusions-suggestions?type=design&node-id=196-6366&mode=design&t=RL1hdfBQLMS1rKVa-0)
+
+<img src="./images/0004-03.png" width=500 alt="Updated flow for existing App Exclusion screen">
+
+-   (2 points) Run the task on the launch of screen
+    -   Update the "have seen these apps" settings list with all suggested apps. (If a new suggested app is install and the Settings screen is opened by a user before the daily job runs to send a notification, we thus never send a notification. This behavior was confirmed by Santiago.)
```
brizental, line 81:
```suggestion
    -   Update the "have seen these apps" settings list with all suggested apps. (If a new suggested app is installed and the Settings screen is opened by a user before the daily job runs to send a notification, we thus never send a notification. This behavior was confirmed by Santiago.)
```

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,156 @@
...
+    -   Metrics will be created for a successful deep linking, as well as an erroneous deep linking (the app cannot figure out how to resolve the deep link).
+    -   Add performance metrics around load time of the Excluded Apps screen.
+    -   Create a metric for when the daily job was successful, and another one for when it fails.
+    -   A metric should be recorded when the notification is sent
+
+Rolling back the JSON file
+---------------------
+If we mistakenly add a suggested exclusion or an app changes in significant ways, we may need to quickly update the suggested apps file.
+
+[This section will be written after Monday's meeting, as it greatly depends on where the document is served from.]
```
brizental, line 104: I think this is out of date.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,156 @@
...
+### Other options considered for suggested app files
+
+The goals for the file format:
+-   Easy to read by humans
+-   Easy for humans to add to, no matter their level of technical proficiency
+-   Easy for us to parse in the app
+-   Allowing comments in the file.
+
+Initially, there was one additional goal (no longer consideration):
+-   To have one file for all platforms, so that we'd remember to add a specific app to all platforms at once. With a conversation around potentially wanting different suggested app categories on mobile, this seems unimportant. (However, if one file per platform seems more streamlined, it's still acceptable.)
```
brizental, line 122: IIUC the current design is for one file per platform?

mcleinman: Current design is one JSON file for all platforms, and I believe is consistent throughout this doc.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Good catch, updated.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Thanks, updated.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
brizental: Yeah, that is what I mean. 

> Initially, there was one additional goal (no longer consideration):
> -   To have one file for all platforms

This kind of implies we are not having a single file per platform?

docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
```diff
@@ -0,0 +1,154 @@
...
+Ultimately, JSON version seemed simpler (and thus easier to build) - there aren't the same parsing concerns and it's easy for contributors.
+
+Metrics
+-------
+
+We will create several metrics as part of this work; they are described in the appropriate spots in the "Implementation" section.
+
+We'd expect the `apps_excluded` metric to increase after this is implemented on platforms which have app exclusions. Both the number of people with more than 0 apps excluded and the average count of apps excluded should increase.
+
+If Nimbus is available by the time this feature is complete, we will use it to roll out the test.
```
lesleyjanenorton, line 133: Given the unknowns about the long-term shape of some aspects of this feature I think we should stipulate that it be rolled out as a Nimbus experiment. Nimbus availability also means we'd have the ability to disable it without a client update.

lesleyjanenorton (approved):
R+ with one comment. Thank you for your efforts and persistence in driving this forward.

docs/rfcs/0004-suggested-excluded-apps-tech-spec.md
mcleinman: Good catch, sorry I fully missed that. Updated.

docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
mcleinman: Good call, updated.

(We'll also be able to disable by turning off the feature.)

PR #8413, opened 2023-10-27
VPN-5592: Fix DNS settings layout (Qt 6.4)

MattLichtenstein:
## Description

Another Qt 6.4 layout issue 😪 

- Properly layout/render the DNS settings view after re-visiting the view while custom DNS is selected and has a value in the text field

## Reference

[VPN-5592: Linux: DNS Settings screen doesn't always render correctly](https://mozilla-hub.atlassian.net/browse/VPN-5592)


mcleinman (approved):
👍🏻

PR #8328, opened 2023-10-18
VPN-5680: Complete mobile onboarding without internet

MattLichtenstein:
## Description

- While attempting to request system permissions to configure the VPN in the final step of mobile onboarding, a failed subscription check was blocking the activation process due to no internet access. We are not attempting to enable (turn on) the VPN here (just request permissions) so we can ignore the results of this check

## Reference

[VPN-5680: [mobile] Unable to finalize the new onboarding flow after disconnecting from Wi-Fi](https://mozilla-hub.atlassian.net/browse/VPN-5680)

PR #8430, opened 2023-10-30
VPN-5532: Prevent persistent VPN config system prompt

MattLichtenstein:
## Description

- Prevents the VPN config system prompt from repeatedly showing when no selection (neither "Allow" or "Don't allow") has been made by the user.
  - This was due to the handshake timer that begins upon an activation request and retries to activate the VPN every 15 seconds, and is only stopped once a connection either succeeds or fails. In this case, the connection success/failure was delayed due to a required response by the user to either allow or not allow the vpn configuration to be made by the OS.
  - This caused onboarding to be automatically completed after the system prompt appears, ignoring it, and waiting 15 seconds 
- Repeatedly showing the VPN config system prompt when selecting "Don't allow" was resolved as part of #8328

## Reference

[VPN-5532](https://mozilla-hub.atlassian.net/browse/VPN-5532)

[VPN-5532]: https://mozilla-hub.atlassian.net/browse/VPN-5532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/android/androidvpnactivity.h
+++ b/src/platforms/android/androidvpnactivity.h
@@ -72,8 +72,12 @@ enum ServiceEvents {
   // The Daemon need's the app to ask for notification
   // permissions, to show the "you're connected" messages.
   EVENT_REQUEST_NOTIFICATION_PERMISSION = 8,
+  // Signals MozillaVPN that we have completed onboarding
   EVENT_ONBOARDING_COMPLETED = 9,
-  EVENT_ALLOW_VPN_CONFIG_OPTION_SELECTED = 10,
+  // Signals the ConnectionManager that it may now allow 
+  // activation retries now that the system vpn config modal
+  // has been responded to
+  EVENT_VPN_CONFIG_PERMISSION_RESPONSE = 10,
 };
 typedef enum ServiceEvents ServiceEvents;
 
@@ -97,7 +101,7 @@ class AndroidVPNActivity : public QObject {
   void eventStatisticUpdate(const QString& data);
   void eventActivationError(const QString& data);
   void eventOnboardingCompleted();
-  void eventAllowVpnConfigOptionSelected();
+  void eventVpnConfigPermissionResponse();
   void eventRequestGleanUploadEnabledState();
 
  private:
--- a/src/platforms/android/androidvpnactivity.h
+++ b/src/platforms/android/androidvpnactivity.h
@@ -74,7 +74,7 @@ enum ServiceEvents {
   EVENT_REQUEST_NOTIFICATION_PERMISSION = 8,
   // Signals MozillaVPN that we have completed onboarding
   EVENT_ONBOARDING_COMPLETED = 9,
-  // Signals the ConnectionManager that it may now allow 
+  // Signals the ConnectionManager that it may now allow
   // activation retries now that the system vpn config modal
   // has been responded to
   EVENT_VPN_CONFIG_PERMISSION_RESPONSE = 10,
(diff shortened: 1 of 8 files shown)


src/platforms/ios/ioscontroller.mm
```diff
@@ -160,6 +160,11 @@
       }
       onboardingCompletedCallback:^() {
         MozillaVPN::instance()->onboardingCompleted();
+      }
+      allowVpnConfigOptionSelectedCallback:^() {
```
MattLichtenstein, line 164: not opposed to better name ideas for this

src/platforms/ios/ioscontroller.swift
```diff
@@ -185,6 +185,10 @@ public class IOSControllerImpl : NSObject {
             tunnel.isEnabled = true
 
             return tunnel.saveToPreferences { saveError in
+                // User made a selection, at this point it is safe to run retries via
+                // ConnectionManager::startHandshakeTimer()
+                allowVpnConfigOptionSelectedCallback()
```
vinocher, line 190: Consider if something like onTunnelConfigurationCompleted() is a better name. Also, could the comment above be clarified a bit -- it is bit cryptic IMO.

src/platforms/ios/ioscontroller.mm
vinocher: Maybe something like onTunnelConfigurationCompleted(). 

Is it possible to do call connectionManager->startHandshakeTimer(); in the existing onboardingCompletedCallback()?

src/platforms/ios/ioscontroller.mm
```diff
@@ -160,6 +160,11 @@
...
+        ConnectionManager* connectionManager = MozillaVPN::instance()->connectionManager();
+        Q_ASSERT(connectionManager);
```
vinocher, line 166: Nit: this assert doesn't add much value IMO and the code base doesn't use asserts much when getting singletons.

src/platforms/android/androidvpnactivity.h
```diff
@@ -73,6 +73,7 @@ enum ServiceEvents {
   // permissions, to show the "you're connected" messages.
   EVENT_REQUEST_NOTIFICATION_PERMISSION = 8,
   EVENT_ONBOARDING_COMPLETED = 9,
+  EVENT_ALLOW_VPN_CONFIG_OPTION_SELECTED = 10,
```
vinocher, line 76: Consider if a name like EVENT_TUNNEL_CONFIGURATION_COMPLETED is better. Also, should a comment be added above this enum value?

vinocher:
An explanation of how the handshake timer causes the current behavior may be useful in the PR description.

src/platforms/android/androidvpnactivity.h
MattLichtenstein: hmm, do you think that naming is still appropriate even though the configuration could be rejected?

src/platforms/ios/ioscontroller.mm
vinocher: Consider something like tunnelConfigSavedCallback or tunnelConfigCompletedCallback, which matches the existing onboardingCompletedCallback

src/platforms/android/androidvpnactivity.h
vinocher: Sorry, please ignore that. Perhaps EVENT_VPN_CONFIG_PERMISSION_PROVIDED is better?

Issue #7772, opened 2023-08-15
Android Talkback screen reader reads content from previous screens


data-sync-user:
The Android Talkback screen reader experience on the Settings/Privacy features screen has two major problems:
1. Open the Privacy features screen and do the first Swipe Right. A long string is announced: “Mozilla VPN, VPN is off, No Signal, Check connection, Connection information, Exit server IP, Loading, Exit Server IPV6, Loading, It looks like your connection to this server location is a bit slow …” 
2. Subsequent Swipe Right operations read elements in the VPN home screen and the Settings screen, both of which are currently not visible.

Eventually, Swipe Right operations will read elements in the Privacy features screen.

For problem 1, it feels like the screen reader is reading previously hidden pages in the navigation stack.

* [~accountid:60acc5180c6f2d0068633c71] indicated that this doesn’t repro in iOS.
* Chris and I can repro on Android (Google Pixel).
* This doesn’t repro in Windows.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5392)
┆Reporter: Vinoo Cherian

data-sync-user:
➤ ayeddi commented:

Just in case, TalkBack guide:https://webaim.org/articles/talkback/ ( https://webaim.org/articles/talkback/|smart-link ), also you could activate the on-screen captions to track the announcement and to include the text output in screenshots by turning on the following feature in the Android OS’s Settings > Accessibility > TalkBack > Settings > Advanced Settings > Developer Settings > Display speech output

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this as fixed on Mozilla VPN 2.19.0 (2.202311021719), using Android 13 (Google Pixel 7a) and the Talkback screen reader no longer reads the content from the previous screens.

While being on the Settings / Privacy features screen and after swiping right, only the elements from the screen are narrated.

PR #8439, opened 2023-10-31
VPN-5392: Android screen reader reads hidden elements of previous pages in VPN navigation stack

vinocher:
## Description

TalkBack, the Android screen reader, reads hidden pages in the VPN app's navigation stack, resulting in a bad experience. 

The VPN Client app has a navigation stack, built with Qt's StackView, to navigate through its screens. When the user selects a new screen, the corresponding page is pushed on to the stack and the previous page is hidden by Qt. Similarly, when the stack selects "Back", the previous page is popped off the stack and Qt shows that page's content. 

So, the previous pages are in the UI tree, but hidden. Screen readers on most platforms don't read hidden elements. However, TalkBack reads them unless the element is explicitly marked or removed from the Accessibility tree. When TalkBack gestures are used to move to the previous/next item, it can navigate to and read those hidden elements, causing a confusing experience.

Fixed by removing hidden elements from the Accessibility tree by binding Accessible.ignored to the visible property.  As far as possible, the core custom component closest to the built-in Qt element was changed, to reduce the number of changes.

Also, removed some redundant QML property code that is not needed because the component already has those properties.

## Reference

[VPN-5392](https://mozilla-hub.atlassian.net/browse/VPN-5392), [Github 7772](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/7772)

## Checklist
    
[VPN-5392]: https://mozilla-hub.atlassian.net/browse/VPN-5392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/screens/home/controller/ConnectionTimer.qml
+++ b/src/ui/screens/home/controller/ConnectionTimer.qml
@@ -12,6 +12,9 @@ RowLayout {
     property real defaultWidth: MZTheme.theme.fontSize * 0.25
     property real narrowWidth: MZTheme.theme.fontSize * 0.05
     property var narrowCharacters: [" ", "\t", "\n", ":"]
+    property bool ignoreForAccessibility: false
+
+    Accessible.ignored: ignoreForAccessibility
 
     function formatSingle(value) {
         if (value === 0)
@@ -39,6 +42,7 @@ RowLayout {
             font.letterSpacing: 0
             font.pixelSize: MZTheme.theme.fontSize
             text: modelData
+            Accessible.ignored: ignoreForAccessibility || !visible
 
             Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
             Layout.preferredWidth: narrowCharacters.includes(modelData) ? narrowWidth : defaultWidth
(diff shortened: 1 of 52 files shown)


src/ui/screens/home/controller/ConnectionTimer.qml
```diff
@@ -12,6 +12,9 @@ RowLayout {
     property real defaultWidth: MZTheme.theme.fontSize * 0.25
     property real narrowWidth: MZTheme.theme.fontSize * 0.05
     property var narrowCharacters: [" ", "\t", "\n", ":"]
+    property bool ignoreForAccessibility: false
+
+    Accessible.ignored: ignoreForAccessibility
```
MattLichtenstein, line 17: why do we need to specifically ignore the `RowLayout` instead of just the `Text` component?

MattLichtenstein (approved):
Looks good!

src/ui/screens/home/controller/ConnectionTimer.qml
vinocher: I was trying to the keep the old behavior:
          ConnectionTimer {
            id: connectionTime
            Accessible.ignored: true
          }

But it isn't needed. Will remove this.

lesleyjanenorton (approved):
Thank you @vinocher !

PR #8463, opened 2023-11-03
[RFC] VPN-5813 Create tech design for connection manager

Gela:
## Description

This document proposes a new component, the ConnectionManager, to handle all connectivity-related checks during VPN activation and periodically once the VPN is activated. The motivation behind this is to standardize and improve the handling of VPN connectivity, which currently leaves users without clear explanations or troubleshooting guidance when issues arise. The ConnectionManager will run probes on all potential causes of connectivity issues and presents the user with a modal detailing the root cause and possible solutions. This new functionality will be encapsulated within the ConnectionManager to avoid complicating the existing controller logic.

The ConnectionManager will manage all probes, stored as distinct states within an enumeration class. It will sequentially cycle through each state, only proceeding to the next if the current probe is successful. If a probe fails, the ConnectionManager will emit a signal to the frontend, triggering a modal for the user explaining the issue and suggesting solutions. The cycle of probes restarts once the issue is resolved. If there are no ongoing probes, the ConnectionManager will enter an Idle state. The existing ConnectionHealth component, which sends periodic pings and determines VPN stability, will be used to recover from a probe failure within the ConnectionManager.

**Accepting feedback until Nov 10, 2023**

Please review and provide your feedback.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5813

## Checklist
    

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,75 @@
...
+
+The objective with the Connection Manager is to address these issues by running probes on all potential causes and presenting the user with an appropriate modal detailing the root cause of their problem. This approach aims to enhance the user's understanding of the issue and provide them with actionable steps to rectify it.
+
+To avoid further complicating the already complex logic of the controller, we have decided to encapsulate this new functionality within the Connection Manager. This object will interact with both the controller and connection health, serving as a dedicated entity for managing and improving VPN connectivity. This approach ensures a more streamlined and efficient handling of VPN connectivity issues, ultimately enhancing the user experience.
```
brizental, line 18:
```suggestion
To avoid further complicating the already complex logic of the controller, this document proposes to encapsulate this new functionality within the Connection Manager. This object will interact with both the controller and connection health, serving as a dedicated entity for managing and improving VPN connectivity. This approach ensures a more streamlined and efficient handling of VPN connectivity issues, ultimately enhancing the user experience.
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Proposed Solution
+
+We are introducing a new object, the Connection Manager, which will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionManager::State`. The Connection Manager will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful.
```
brizental, line 22:
```suggestion
We will introduce a new object, the Connection Manager, which will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionManager::State`. The Connection Manager will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful.
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,75 @@
...
+
+In the event of a probe failure, the Connection Manager will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted. Once the issue has been resolved, the cycle of probes commences anew. If there are no ongoing probes, the Connection Manager will enter an Idle state.
```
brizental, line 24: I would be interested in what are the proposed troubleshooting steps for each of the failed probes.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,75 @@
...
+  - Wait until Connection Health stability goes back to Stable or Unstable
+  - Restart the sequence of ConnectionManager checks
+
+![Diagram to explain how Controller, ConnectionManager and ConnectioHealth states interact](./CM_diagram.png)
+
+## Metrics and Measurements
+
+This change also means that prior to the activation of the VPN, there are multiple checks and probes that need to succeed; this will take some time (hopefully no more than a couple of seconds).
+
+To confidently determine that these checks do not delay the VPN activation by an unacceptable amount TBD, we need to measure the duration of the _time elapsed from toggling on the VPN to activation_. We are interested in the VPN activation duration because it delays the VPN being toggled on.
```
brizental, line 73: We might also be interested in probe success vs failure metrics.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,75 @@
...
+
+In the event of a probe failure, the Connection Manager will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted. Once the issue has been resolved, the cycle of probes commences anew. If there are no ongoing probes, the Connection Manager will enter an Idle state.
+
+The Connection Health, an existing component of our codebase, is tasked with sending periodic pings and determining the stability of the VPN based on the responses received (see `ConnectionHealth::healthCheckup()`). It classifies the VPN stability as `Stable`, `Unstable` (some missing responses), or No Signal (no responses received). When the VPN experiences issues with network connectivity or server availability, it naturally transitions into the No Signal state and, ideally, recovers once the issue is resolved. We can leverage this logic to recover from a probe failure within the Connection Manager.
```
brizental, line 26: > We can leverage this logic to recover from a probe failure within the Connection Manager.

Would you mind expanding a bit on how we would do that? Is ConnectionHealth listening for ConnectionManager probe failures? I see an arrow going from one to the other in the attached diagram, but the text only talk about "emmiting a signal to the front end". Is ConnectionHealth == front end in this case?

brizental: Wait, I just re-read "We will __not__ manipulate the Connection Health stability based on Connection Manager probes". I'm confused.

brizental: And then
>   - Wait until Connection Health stability goes back to Stable or Unstable
>  - Restart the sequence of ConnectionManager checks

So is ConnectionManager the one listening for ConnectionHealth?

brizental:
Please add [RFC] prefix to the PR title.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: They are specific to each probe, there are designa from the UX engs linked in the epic, but going into the specifics of what the frontend should look like are beyond the scope of this doc.
Internet probe Figma: https://www.figma.com/file/GZKigVZbnCisMJNU3Hc9xM/Connection-error-messaging?type=design&node-id=26-215&mode=design&t=kfYnUKuXQ9nsGDRR-0

docs/rfcs/0004-rfc-connectiomanager.md
Gela: Yes correct, if a ConnectionManager probe fails and ConnectionHealth stability is NoSignal, Connection Manager aborts all probes until the stability changes to stable or unstable.

If VPN is in NoSignal we don't advance to the next probe.



For example, if the VPN is already toggled on and the internet goes down:

1. VPN goes into NoSignal via Connection Health
2. Connection Manager internet probe fails
3. Connection Manager emits a signal to the front end about network being down
4. User resolves their internet issue somehow
5. Connection Health goes back to Stable
6. Now because connection health is no longer in NoSignal, the connection manager checks restart

mcleinman:
Rendered: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/gela/another_cm_techspec/docs/rfcs/0004-rfc-connectiomanager.md

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
+- Status: In progress
+- Date: 2023-11-02
+- Author: [@gela](https://github.com/gela)
+- RFC PR: TBD
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+
+Propose a new `ConnectionManager` component which handles all connectivity related checks once during VPN activation (`Controller::StateConnecting`) and then periodically once the VPN is already activated (`Controller::StateOn`); the controller will continue to exist as is, communicating with the frontend as the backend goes through various states such as `connecting` and `confirming` to reflect those to the user.
```
mcleinman, line 10: Given that you say "a new ConnectionManager component", it might be worth mentioning what is happening with the old class with that name.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+### Upon VPN Activation (`StateOff` to `StateConnecting`)
+Perform a single cycle going through all the Connection Manager probes.
+- If a probe fails:
+  - Set Controller to `StateOff`
+  - Stop subsequent probes from running
+  - emit a signal to the frontend communicating which probe failed
+- If all probes pass:
+  - Activate VPN (Controller `StateOn`)
+
+### VPN is already active (`StateOn`)
```
lesleyjanenorton, line 48: Based on our conversation yesterday and the note on L58, would it make sense to limit the scope of this proposal to the Connection Manager's role during activation (L39-46) and postpone stipulating behavior/handing of the periodic checks?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+
+Propose a new `ConnectionManager` component which handles all connectivity related checks once during VPN activation (`Controller::StateConnecting`) and then periodically once the VPN is already activated (`Controller::StateOn`); the controller will continue to exist as is, communicating with the frontend as the backend goes through various states such as `connecting` and `confirming` to reflect those to the user.
+
+## Motivation
+
+At present, the handling of VPN connectivity lacks standardization and consistency. There are instances when the VPN goes into No Signal due to a variety of connectivity-related issues, such as the network connection failing or the server location becoming unavailable. In these situations, users are often left without a clear explanation of what went wrong or guidance on how to troubleshoot and resolve the issue. They may encounter a background error message or lose signal without further information.
```
mcleinman, line 14: "lacks standardization and consistency" - You mean for users, right? Not in the code? That's what the rest of the paragraph implies, but it's probably worth being explicit.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+  - Wait until Connection Health stability goes back to Stable or Unstable
+  - Restart the sequence of ConnectionManager checks
+
+![Diagram to explain how Controller, Connection Manager and Connection Health states interact](./CM_diagram.png)
+
+## Metrics and Measurements
+
+This change also means that prior to the activation of the VPN, there are multiple checks and probes that need to succeed; this will take some time (hopefully no more than a couple of seconds).
+
+To confidently determine that these checks do not delay the VPN activation by an unacceptable amount TBD, we need to measure the duration of the _time elapsed from toggling on the VPN to activation_. We are interested in the VPN activation duration because it delays the VPN being toggled on. This also allows us to keep track of the duration of each operation which allows us to continually monitor the system health and identify any slow downs and changes.
```
mcleinman, line 73: Will this confirmation that it doesn't introduce an unacceptable delay be done in testing? By QA? Based on early metrics we get back from users?

Gela:
@mcleinman clicking on the circled button on any PR will also show the formatted version 

Screenshot 2023-11-03 at 2 07 08 PM

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+Perform periodic checks cycling through all the Connection Manager probes.
+- If a probe fails:
+  - Controller continues to stay in `StateOn`. We will __never__ toggle the VPN off without the user explicitly doing so themselves
+  - We will __not__ manipulate the Connection Health stability based on Connection Manager probes
+  - Stop subsequent probes from running
+  - emit a signal to the frontend communicating which probe failed
+- If all probes pass:
+  - Do nothing. VPN stays on and stable
+
+> Note: The second scenario where a check fails when the VPN is already active still needs some more thinking. There will likely be a follow up design doc on how we will approach that later on.
```
mcleinman, line 58: What do you mean by "needs some more thinking"? Can you add example of what downsides we might be pushing to a future doc?

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman, line 48: What happens when a user manually switches the VPN off while these checks are running? Are they cancelled? Can they finish with no impact? Do we still show a modal if something fails?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+
+In the event of a probe failure, the Connection Manager will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted. Once the issue has been resolved, the cycle of probes commences anew. If there are no ongoing probes, the Connection Manager will enter an Idle state.
+
+The Connection Health, an existing component of our codebase, is tasked with sending periodic pings and determining the stability of the VPN based on the responses received (see `ConnectionHealth::healthCheckup()`). It classifies the VPN stability as `Stable`, `Unstable` (some missing responses), or `NoSignal` (no responses received). When the VPN experiences issues with network connectivity or server availability, it naturally transitions into the No Signal state and, ideally, recovers once the issue is resolved. We can leverage this logic to recover from a probe failure within the Connection Manager.
```
mcleinman, line 26: I think this doc needs to discuss how ConnectionHealth can also change the UI for users. 

And what if we changed that current practice as part of this work - what if a bad periodic ConnectionHealth check just signals ConnectionManager to run its checks (if it's currently in idle, and not already checking)?

mcleinman (commented):
I think this is going to be a great improvement to our codebase. Curious about some clarifying questions.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+> Note: The second scenario where a check fails when the VPN is already active still needs some more thinking. There will likely be a follow up design doc on how we will approach that later on.
+
+### How to recover after a failed probe
+- If VPN is not in `StateOn` yet:
+  - Connection Manager will emit signal about the failed probe to notify the frontend
+- If VPN is in `StateOn`:  
+  - Wait until Connection Health stability goes back to Stable or Unstable
+  - Restart the sequence of ConnectionManager checks
+
+![Diagram to explain how Controller, Connection Manager and Connection Health states interact](./CM_diagram.png)
```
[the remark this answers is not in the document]
Gela: IT'S A RACE!! 🏁 (jk, I agree about moving other files in a separate folder)

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+
+The objective with the Connection Manager is to address these issues by running probes on all potential causes and presenting the user with an appropriate modal detailing the root cause of their problem. This approach aims to enhance the user's understanding of the issue and provide them with actionable steps to rectify it.
+
+To avoid further complicating the already complex logic of the controller, this document proposes encapsulating this new functionality within the Connection Manager. This object will interact with both the controller and connection health, serving as a dedicated entity for managing and improving VPN connectivity. This approach ensures a more streamlined and efficient handling of VPN connectivity issues, ultimately enhancing the user experience.
```
lesleyjanenorton, line 18: > serving as a dedicated entity for managing and improving VPN connectivity.

Would something like `dedicated entity for identifying connection problems and correctly communicating status changes to the UI` be more appropriate here? Not sure we are actually doing anything to improve VPN connectivity here.

docs/rfcs/0004-rfc-connectiomanager.md
lesleyjanenorton, line 10: Your call on this but I think we could add a line here saying something like `This proposal describes Connection Manager behavior during activation. Subsequent proposal(s) will address the behavior of the Connection Manager during the lifespan of a connection."

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Proposed Solution
+
+We will introduce a new object, the Connection Manager, which will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionManager::State`. The Connection Manager will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful.
```
lesleyjanenorton, line 22: > only proceeding to the next state if the current probe is successful.

> Upon encountering a failed probe, the progression through the remaining probes is immediately halted.

We'll still need to do some diagnostic work though right? Unless we work out a perfect sequence by which to run checks, it seems possible that we could fail at `check A` when the failure is actually a failing `check B`, and, without further detective work, emit the wrong error message.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,76 @@
...
+In the event of a probe failure, the Connection Manager will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted. Once the issue has been resolved, the cycle of probes commences anew. If there are no ongoing probes, the Connection Manager will enter an Idle state.
+
+The Connection Health, an existing component of our codebase, is tasked with sending periodic pings and determining the stability of the VPN based on the responses received (see `ConnectionHealth::healthCheckup()`). It classifies the VPN stability as `Stable`, `Unstable` (some missing responses), or `NoSignal` (no responses received). When the VPN experiences issues with network connectivity or server availability, it naturally transitions into the No Signal state and, ideally, recovers once the issue is resolved. We can leverage this logic to recover from a probe failure within the Connection Manager.
+
+## Implementation Plan
+
+```c++
+enum ConnectionManagerStates {
+  InternetProbe,
+  Firewall,
```
lesleyjanenorton, line 33: Are these the 'Phase 1' checks we expect to run during activation?

lesleyjanenorton: Also nit: `StateInternetProbe`, `StateFirewall`, etc.

docs/rfcs/0004-rfc-connectiomanager.md
lesleyjanenorton, line 26: 👍

docs/rfcs/0004-rfc-connectiomanager.md
lesleyjanenorton, line 73: 👍

docs/rfcs/0004-rfc-connectiomanager.md
brizental: I was not thinking about the front end specifically, but a more general explanation to how each failure can troubleshot (?) but the user. Regardless, I think the Figma link answers that. Would you mind referring to it in the document as a not? It helps to see the feature as it is meant.

docs/rfcs/0004-rfc-connectiomanager.md
brizental: The Figma dashboard really helped me understand things better. So, the connection health is like blunt tool that is just telling us whether there is signal or not. While the connection manager is a more precise tool that tells us exactly why things are weird and thus is able to provide more specific troubleshooting info to the front end. Is that correct?

brizental: > Yes correct, if a ConnectionManager probe fails and ConnectionHealth stability is NoSignal, Connection Manager aborts all probes until the stability changes to stable or unstable.

Ok, but in the 6 steps you described there is no point in which the connection manager actually listens for signals from the connection health. Is it a case where the effects are related but not actually tied to each other?

Gela: > While the connection manager is a more precise tool that tells us exactly why things are weird and thus is able to provide more specific troubleshooting info to the front end. Is that correct?

Yes that's right.

> in the 6 steps you described there is no point in which the connection manager actually listens for signals from the connection health. 

Step 5 and 6 are dependent on the connection health stability and where the connection manager is listening to the stability to determine when the connection manager should restart the checks

docs/rfcs/0004-rfc-connectiomanager.md
Gela: Likely both

docs/rfcs/0004-rfc-connectiomanager.md
Gela: Not any downsides necessary, we just need to work out the specific details that I don't want to get into in this document.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > I think this doc needs to discuss how ConnectionHealth can also change the UI for users.

Are you referring to the existing stability modes (stable, nosignal, unstable) in the UI, or something different? 

> And what if we changed that current practice as part of this work - what if a bad periodic ConnectionHealth check just signals ConnectionManager to run its checks (if it's currently in idle, and not already checking)?

I'm hesitant to change this as we didn't talk/discuss this in any of the connection manager discussions. It opens a whole other can of worms which we never agreed on

docs/rfcs/0004-rfc-connectiomanager.md
Gela: There is a note further down specifying that there will be a follow up for the second portion. Do you want me to get rid of any mentions altogether? I figured it would be helpful to give some context about the checks while the VPN is already on without going into too much details

docs/rfcs/0004-rfc-connectiomanager.md
Gela: What sort of diagnostics are you thinking? I was under the impression that we would identify the cause and then show the user some things they can try. For example if the network check fails, we'll tell them to check their internet. I'm not sure if there's much (or anything really) we as the VPN client can do to resolve any of the issues the connection manager identifies.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Are these the 'Phase 1' checks we expect to run during activation?

I haven't spent a ton of time thinking ahead about the phase 2 details yet but my initial response is that phase 1 and phase 2 have the same set of checks. The only major difference is that phase 1 runs them right at activation and phase 2 runs them periodically while the VPN is already active.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,80 @@
...
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+
+The existing `ConnectionManager` component will handle all connectivity related checks once during VPN activation (`Controller::StateConnecting`) and then periodically once the VPN is already activated (`Controller::StateOn`); the controller will be the entity which manages the controller states that the client goes through such as `connecting` and `confirming` as we activate and deactivate the VPN to reflect those to the user what is happening behind the scene.
+
+## Motivation
+
+At present, the handling of VPN connectivity lacks consistency for users. There are instances when the VPN goes into No Signal due to a variety of connectivity-related issues, such as the network connection failing or the server location becoming unavailable. In these situations, users are often left without a clear explanation of what went wrong or guidance on how to troubleshoot and resolve the issue. They may encounter a background error message or lose signal without further information.
```
vinocher, line 14: > They may encounter a background error message or lose signal without further information.

Consider mentioning the actual error to clarify the motivation for this feature. For example, if the internet connection is disabled, a "No Signal" error is displayed, without indicating that it is caused by lack of an internet connection.

Also, what does "background" mean?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,80 @@
...
+
+The objective with the Connection Manager is to address these issues by running probes on all potential causes and presenting the user with an appropriate modal detailing the root cause of their problem. This approach aims to enhance the user's understanding of the issue and provide them with actionable steps to rectify it.
```
vinocher, line 16: > "modal"

Does this mean a modal dialog box? Isn't this error displayed using an alert, which is not modal?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,80 @@
...
+## Implementation Plan
+
+```c++
+enum ConnectionManagerStates {
+  StateInternetProbe,
+  StateFirewall,
+  StateServerProbe,
+  StateCheckSubscription,
+  StateIdle,
+}
```
vinocher, line 41: > StateInternetProbe,
  StateFirewall,

Nit: Should the `State` prefix be removed because it is redundant? Also should verbs be used to name some of these states that are doing work? For example, ProbingInternet, CheckingFirewall, ProbingServer, CheckingSubscription, Idle. But it looks like Lesley's comment recommends adding the `State` prefix!

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,80 @@
...
+> Note: The second scenario where a check fails when the VPN is already active still needs some more thinking. There will be a follow up design doc on how we will approach that later on.
+
+### How to recover after a failed probe
+- If VPN is not in `StateOn` yet:
+  - Connection Manager will emit signal about the failed probe to notify the frontend
+- If VPN is in `StateOn`:  
+  - Wait until Connection Health stability goes back to Stable or Unstable
+  - Restart the sequence of ConnectionManager checks
+
+![Diagram to explain how Controller, Connection Manager and Connection Health states interact](./images/CM_diagram.png)
```
vinocher, line 71: > Diagram

Could the diagram show how ConnectionHealth communicates with ConnectionManager?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,80 @@
...
+
+In the event of a probe failure, the Connection Manager will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted. Once the issue has been resolved, the cycle of probes commences anew. If there are no ongoing probes, the Connection Manager will enter an Idle state.
+
+The Connection Health, an existing component of our codebase, is tasked with sending periodic pings and determining the stability of the VPN based on the responses received (see `ConnectionHealth::healthCheckup()`). It classifies the VPN stability as `Stable`, `Unstable` (some missing responses), or `NoSignal` (no responses received). When the VPN experiences issues with network connectivity or server availability, it naturally transitions into the No Signal state and, ideally, recovers once the issue is resolved. We can leverage this logic to recover from a probe failure within the Connection Manager.
+
```
vinocher, line 29: > We can leverage this logic to recover from a probe failure within the Connection Manager.

Could this interaction between ConnectionManager and ConnectionHealth be described in more detail? For example, if ConnectionHealth's probe indicates NoSignal, how does ConnectionManager use that information to identify the underlying cause?

My understanding is that when ConnectionHealth signals Unstable or NoSignal to ConnectionManager, the latter will run its probes to determine the underyling cause and then communicate that to the front end.

docs/rfcs/0004-rfc-connectiomanager.md
vinocher: Didn't we discuss that ConnectionManager would rationalize any failures reported by ConnectionHealth (by running the ConnectionManager probes to detect the underlying cause) and report that to the front end. So, ConnectionHealth would not directly report to the front end.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Also, what does "background" mean?

It's literally a "background error occurred" messaged with no additional context. I'm guessing it's referring to something going wrong with Guardian.

mcleinman:
> @mcleinman clicking on the circled button on any PR will also show the formatted version. Same with clicking on ... and "View File"
> 
> Screenshot 2023-11-03 at 2 07 08 PM

🤯 Thanks.

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: Can that be explicitly mentioned in the doc?

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: To me, this document is where we should be thinking this through, to make sure we don't back ourselves into a corner. This seems really closely tied to the decisions being made in this doc, so why shouldn't it be further explored here?

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: Can that cancellation of connection manager be spelled out in the doc?

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: If ConnectionManager rationalizes failures reported by ConnectionHealth, it seems extra important to have bad health checks kick off a ConnectionManager run (if it's currently idle).

mcleinman: Separately, I don't see the doc discussion limited to in-person discussions, which may be what you're saying (or I may be misunderstanding). I think seeing it written down after a group discussion can allow us to see potential interactions we hadn't previously.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: The errors would be shown to the user using a box that pops up in the client, [this Figma](https://www.figma.com/file/GZKigVZbnCisMJNU3Hc9xM/Connection-error-messaging?type=design&node-id=26-215&mode=design&t=kfYnUKuXQ9nsGDRR-0) shows what they would look like. User would need to explicitly dismiss the alert. I think that's a modal, right? If I'm using the wrong term I can change it to something more accurate.

Screenshot 2023-11-14 at 4 14 44 PM

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Didn't we discuss that ConnectionManager would rationalize any failures reported by ConnectionHealth (by running the ConnectionManager probes to detect the underlying cause) and report that to the front end. So, ConnectionHealth would not directly report to the front end.

Connection Health will continue to show "No Signal" or "Unstable" to the frontend as it does currently. I

docs/rfcs/0004-rfc-connectiomanager.md
brizental: My confusion here, IMO, stems for some poorly names classes. Specifically ConnectionManager. That name doesn't really convey "a more precise tool that tells us exactly why things are weird". I understand if this is out of context, worst case scenario take this as an FYI.

Gela: Renamed it to `ConnectionDiagnostics` but if the team feels more excited about `ConnectionDoctor` I could be persuaded easily... 👩🏽‍⚕️

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > If ConnectionManager rationalizes failures reported by ConnectionHealth, it seems extra important to have bad health checks kick off a ConnectionManager run (if it's currently idle).

@mcleinman I agree 100%. I was holding that off for part 2 but figured I might as well combine the 2 phases in one doc. I actually don't think we should be periodically running the checks while the VPN is active. It only makes sense to do so when the VPN goes into no signal. Updated the tech spec to reflect that

docs/rfcs/0004-rfc-connectiomanager.md
vinocher: Looks good, since it is a modal. I incorrectly assumed it was a modeless alert.

docs/rfcs/0004-rfc-connectiomanager.md
vinocher: Seems OK to be consistent with the other states.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+The objective with the Connection Diagnostics is to address these issues by running probes on all potential causes and presenting the user with an appropriate modal detailing the root cause of their problem. This approach aims to enhance the user's understanding of the issue and provide them with actionable steps to rectify it.
+
+To avoid further complicating the already complex logic of the controller, this document proposes encapsulating this new functionality within the Connection Diagnostics. This object will interact with both the controller and connection health, serving as a dedicated entity for identifying connection problems and correctly communicating status changes to the UI, ultimately enhancing the user experience.
+
+## Proposed Solution
+
+The Connection Diagnostics will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionDiagnostics::State`. The Connection Diagnostics will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful.
```
vinocher, line 22: Could this describe what will happen if ConnectionDiagnostics cannot determine the cause of an Unstable or NoSignal state that was detected by ConnectionHealth?

docs/rfcs/0004-rfc-connectiomanager.md
vinocher: > Connection Health will continue to show "No Signal" or "Unstable" to the frontend as it does currently.

If Connection Health detects "No Signal", it appears that there will be two alerts in the client app -- one from Connection Health showing the connection state, and potentially another from Connection Diagnostics indicating the cause of the anomalous connection state. Is that correct? See my other comment about the potential downsides of using a modal.

I think it makes sense to show the connection state alert as soon as possible, because it will take longer to run the Connection Diagnostics probes and determine the cause. If the cause can be determined, the initial alert could be subsequently modified to show both the connection state and cause. For example, the user would see "No Signal"  when Connection Health reports it. It will be replaced by something like "No Signal. Check your internet connection." when Connection Diagnostics completes its probes.

Or two alerts can be shown, one for the connection state and another for the cause.

Consider if Connection Diagnostics should be responsible for emitting all connection state related signals to the font end, for better encapsulation. (Connection Health can live inside Connection Diagnostics and notify it.)

mcleinman: After looking at the code, I believe that the health checks do not pop modals - they just end up calling signals that update the color of the icon in the system tray and on the main screen. I may be wrong here, and would love someone else to confirm.

mcleinman: I really like the idea of popping a modal as soon as possible, and updating it as we have more information.

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: Love `ConnectionDiagnostics`. This feels much better - the update I didn't know I was looking for, but feels perfect. Thank you.

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: Thinking more on this - 

If it does work in this way - ConnectionHealth updates colors and text on main screen, the connection diagnostic checks are what ends up popping the modals - that seems fair, but could be unintuitive to future engineers down the line. Can we leave some wayfinding comments in the code (both in ConnectionHealth and in ConnectionDiagnostics) explaining this setup?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+  - Stop subsequent probes from running
+  - emit a signal to the frontend communicating which probe failed
+  - VPN will remain in `StateOff`
+  - Connection Diagnostics will go into `StateIdle`
+- If all probes pass:
+  - Activate VPN (Controller `StateOn`)
+  - Connection Diagnostics will go into `StateIdle`
+
+### VPN is already active (`StateOn`) - Phase 2
+If the VPN is already active and Connection Health stability goes into no signal, initiate a diagnostics cycle to perform a set of checks to identify the underlying cause.
```
mcleinman, line 55: `SubscriptionMonitor` and `CaptivePortalDetection` currently listen for the stabilitychanged signal being emitted by health monitoring. This should stop, right, as it’s now going to be handled by connection diagnostic tool?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+
+- If a probe fails:
```
mcleinman, line 57: If we hit NoSignal state and run diagnostic checks, we'd expect that a probe would fail, right? Is there any corner case possiblity to hit NoSignal but end up with all successful checks? Even if we can't consider a way this could happen, what would the user see? Just a NoSignal red state with no modal? If that happens, we should at the very least log the heck out of it (and maybe send a Glean event), no?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+  - Set Controller to `StateOff`
+  - Stop subsequent probes from running
+  - emit a signal to the frontend communicating which probe failed
+  - VPN will remain in `StateOff`
+  - Connection Diagnostics will go into `StateIdle`
+- If all probes pass:
+  - Activate VPN (Controller `StateOn`)
+  - Connection Diagnostics will go into `StateIdle`
+
+### VPN is already active (`StateOn`) - Phase 2
```
mcleinman, line 54: If a connection health check ends up with `unstable`, we don't do anything, right? I'm pretty sure (and that makes sense), but just wanted to confirm.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+
+To avoid further complicating the already complex logic of the controller, this document proposes encapsulating this new functionality within the Connection Diagnostics. This object will interact with both the controller and connection health, serving as a dedicated entity for identifying connection problems and correctly communicating status changes to the UI, ultimately enhancing the user experience.
+
+## Proposed Solution
+
+The Connection Diagnostics will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionDiagnostics::State`. The Connection Diagnostics will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful. 
+
+In the event of a probe failure, the Connection Diagnostics will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted and the Connection Diagnostics will enter the Idle state.
```
mcleinman, line 24: How will this modal disappear? Will the user need to interact with it? 

What happens if we hit NoSignal and pop a modal - but then the connectivity issue resolves itself and we get a successful health check before the user interacts with the modal?

mcleinman (changes requested):
This is looking great. I really like the rename too. Just a handful of clarifying questions, but they're mostly small corner cases, trying to make sure we've considered everything. Thanks for this.

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
+- Status: In progress
+- Date: 2023-11-02
+- Author: [@gela](https://github.com/gela)
+- RFC PR: TBD
+- Implementation GitHub issue: TBD.
+
+
+## Summary
+
+The existing `ConnectionManager` class will be renamed to `ConnectionDiagnostics` and will be responsible for handling all connectivity related checks once during VPN activation (`Controller::StateConnecting`) and then periodically once the VPN is already activated (`Controller::StateOn`); the controller will be the entity which manages the controller states that the client goes through such as `connecting` and `confirming` as we activate and deactivate the VPN to reflect those to the user what is happening behind the scene.
```
brizental, line 10: Why do we need to run diagnostics while the VPN is just working as expected? I'd expect that to be necessary only once ConnectionHealth has reported instability or lack of signal.

docs/rfcs/0004-rfc-connectiomanager.md
brizental, line 24: What if there is instability in the network and all ConnectionDiagnostics probes succeed? Do we leave the user hanging or is there a catch all error modal?

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,74 @@
...
+  StateProbingInternet,
+  StateCheckingFirewall,
+  StateProbingServerLocation,
+  StateCheckingSubscription,
+  StateIdle,
+}
+```
+
+### Upon VPN Activation (`StateOff` to `StateConnecting`) - Phase 1
+Perform a single cycle going through all the Connection Diagnostics probes.
```
brizental, line 43: Why do we need to run these checks upon activation? Why not attempt to activate and only run the chceks if activation is unsuccesful?

brizental: Also, for my understanding who owns ConnectionDiagnostics? Controller?

docs/rfcs/0004-rfc-connectiomanager.md
brizental: Also what if ConnectionHealth recovers while ConnectionDiagnostics is still running probes? Does it abort? In that same vein, what if ConnectionHealth recovers while the modal is showing to users? Does it just disappear?

brizental: Just noticed this is the same question Matt just asked, :shrug:

docs/rfcs/0004-rfc-connectiomanager.md
brizental: Actually reading on it seems we only run diagnostics when we go into NoSignal. All good.

docs/rfcs/0004-rfc-connectiomanager.md
Gela: Thanks to Bea for expressing her dislike for the old `ConnectionManager` name which led to this one. ChatGPT had some other suggestions such as `ConnectivityDiagnosticsManager`, `NetProbe` and `NetDiagnostics` too but we settled on `ConnectionDiagnostics` 👌🏽

docs/rfcs/0004-rfc-connectiomanager.md
Gela: @vinocher @mcleinman Yes the way I had envisioned this would be that the VPN goes into no signal as it does currently, and the text updates in the controller to show the new stability value (e.g. `Unstable`). Then as we go through the various probes, as soon as the first one fails and we are able to determine the cause for network issues, we will show a popup/modal with the reason. The user has close this modal as they wish and they'll still be able to see the No Signal text in the controller. Does that sound good?

docs/rfcs/0004-rfc-connectiomanager.md
Gela: There is an arrow from Controller to Connection Diagnostics in the diagram that has the label ConnectionHealth::NoSignal, indicating that if connection health stability goes into no signal then we will kick off the diagnostics. Do you have a suggestion for how to show that better?

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > My understanding is that when ConnectionHealth signals Unstable or NoSignal to ConnectionManager, the latter will run its probes to determine the underyling cause and then communicate that to the front end.

Connection Diagnostics will not do anything if the Connection Health stability mode is `Unstable`. We only initiate the diagnostics check if stability is NoSignal

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Will this modal be automatically closed if the error is transitory and is resolved another change in the system or network? For example, what happens if this occurs while the client app is hidden and the VPN hits a transient connection error and then recovers?

Yes if the issue resolves itself and VPN goes back to stable after we pop up the modal then the modal will close automatically. I'll add this to the document. 

If VPN is in the background as this happens we will still show the same modal and handle it as if it was in the foreground, and if the error resolves itself we will close the modal automatically. @MattLichtenstein I think that's doable, right?

docs/rfcs/0004-rfc-connectiomanager.md
Gela: `CaptivePortalDetection` changes are still ongoing but as a result of a separate effort, moving forward we will not check for captive portal after the VPN is activated, this is also why I didn't include Captive portal in the list of probes.

Regarding Subscription Monitor, yes you are correct. It will be moved to Diagnostics

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Is there any corner case possiblity to hit NoSignal but end up with all successful checks? Even if we can't consider a way this could happen, what would the user see? Just a NoSignal red state with no modal?

Yes I'm adding a little more clarification on this to the tech spec but as you mentioned, if all probes pass and we are still in No Signal, then we won't show anything beyond "No Signal" in the controller

```
- If all probes pass but the stability is still `NoSignal`:
  - Something else has gone wrong and are not able to determine the cause. VPN continues to stay in No Signal but we will not show any further modals to the user
  - This will be recorded in the logs for debugging purposes
```

docs/rfcs/0004-rfc-connectiomanager.md
Gela: Yes that's right

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > How will this modal disappear? Will the user need to interact with it?

Either the user manually closes it OR VPN health goes out of No Signal, in that case the modal will close automatically. Added to the tech spec.

Gela: If all probes pass but VPN is still in No Signal (aka we failed to pinpoint the exact issue) then we won't show any modals to the user. At that point we'll just leave it in No Signal, not sure a catch-all modal would offer additional guidance

docs/rfcs/0004-rfc-connectiomanager.md
Gela: > Why do we need to run these checks upon activation? Why not attempt to activate and only run the cheks if activation is unsuccessful?

Hmmm I'd be open to that. @lesleyjanenorton what are your thoughts?

> Also, for my understanding who owns ConnectionDiagnostics? Controller?

Not sure I understand this question, can you elaborate?

docs/rfcs/0004-rfc-connectiomanager.md
brizental: I'm fine with that, maybe call it out in the doc.

Gela: @brizental Added to the doc about 20 seconds before your commend 😄

```
- If all probes pass but the stability is still `NoSignal`:
  - Something else has gone wrong and are not able to determine the cause. VPN continues to stay in No Signal but we will not show any further modals to the user
  - This will be recorded in the logs for debugging purposes
  ```

File: docs/rfcs/0004-rfc-connectiomanager.md (hunk 27 above)
Comment (brizental) [reply docs/rfcs/0004-rfc-connectiomanager.md:43]: Sure, I mean who is responsible for instantiating and managing the ConnectionDiagnostics object?

File: docs/rfcs/0004-rfc-connectiomanager.md (hunk 22 above)
Comment (mcleinman) [reply docs/rfcs/0004-rfc-connectiomanager.md:55]: Sorry, missed that it wasn't part of this. Sounds good.

Comment (mcleinman) [reply docs/rfcs/0004-rfc-connectiomanager.md:55]: What is the `CaptivePortalDetection` out of curiosity? Do you know if that effort will move it into this diagnostic class?

File: docs/rfcs/0004-rfc-connectiomanager.md (hunk 23 above)
Comment (Gela) [reply docs/rfcs/0004-rfc-connectiomanager.md:57]: @mcleinman I agree with recording the frequency of how often we end up in No Signal without a specific diagnostic. I don't think we should be getting into the specifics of what metrics to record here though, right? I assume that would be a Miro doc further down the line?

File: docs/rfcs/0004-rfc-connectiomanager.md (hunk 22 above)
Comment (Gela) [reply docs/rfcs/0004-rfc-connectiomanager.md:55]: We no longer check for captive portal upon network changes, only once during activation. Even though we no longer check for that after the VPN is already activated, I think it should still be in the Diagnostics so I'll add it to the states in this doc. 

Does this answer your question?

File: docs/rfcs/0004-rfc-connectiomanager.md (hunk 27 above)
Comment (Gela) [reply docs/rfcs/0004-rfc-connectiomanager.md:44]: > Sure, I mean who is responsible for instantiating and managing the ConnectionDiagnostics object?

The diagnostics tool runs when stability is `NoSignal` so I think the connection health class is responsible for instantiating the diagnostics and managing it.

File: docs/rfcs/0004-rfc-connectiomanager.md [hunk 28]
```diff
@@ -0,0 +1,83 @@
...
+
+## Proposed Solution
+
+The Connection Diagnostics will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionDiagnostics::State`. The Connection Diagnostics will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful. If all probes pass but VPN stability is still `NoSignal`, then we would not present any modals to the user, but the controller will continue to show "No Signal" as we do currently.
+
+In the event of a probe failure, the Connection Diagnostics will emit a signal to the frontend. This will trigger a modal to be displayed to the user, providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted and the Connection Diagnostics will enter the Idle state.
```
brizental, line 24: nit, the ConnectionDiagnostics actually doesn't know anything aobut the frontend, does it?
```suggestion
In the event of a probe failure, the Connection Diagnostics will emit a signal. The frontend will listen to that signal and display a modal to the user providing an explanation of the issue and suggesting potential solutions. Upon encountering a failed probe, the progression through the remaining probes is immediately halted and the Connection Diagnostics will enter the Idle state.
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,83 @@
...
+  StateIdle,
+}
+```
+
+### Upon VPN Activation (`StateOff` to `StateConnecting`) - Phase 1
+Perform a single cycle going through all the Connection Diagnostics probes.
+- If a probe fails:
+  - Set Controller to `StateOff`
+  - Stop subsequent probes from running
+  - emit a signal to the frontend communicating which probe failed
```
brizental, line 48:
```suggestion
  - emit a signal communicating which probe failed
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,83 @@
...
+  - Connection Diagnostics will go into `StateIdle`
+
+### VPN is already active (`StateOn`) - Phase 2
+If the VPN is already active and Connection Health stability goes into no signal, initiate a diagnostics cycle to perform a set of checks to identify the underlying cause.
+
+- If a probe fails:
+  - Controller continues to stay in `StateOn`. We will __never__ toggle the VPN off without the user explicitly doing so themselves
+  - We will __not__ manipulate the Connection Health stability based on Connection Diagnostics probes
+  - Stop subsequent Connection Diagnostics probes from running
+  - emit a signal to the frontend communicating which probe failed
```
brizental, line 62:
```suggestion
  - emit a signal communicating which probe failed
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,83 @@
...
+
+- If a probe fails but the issues resolves on its own and Connection Health stability goes back to `Stable` while the modal is up:
```
brizental, line 64:
```suggestion
- If a probe fails but the issue resolves on its own and Connection Health stability goes back to `Stable` while the modal is up:
```

docs/rfcs/0004-rfc-connectiomanager.md
```diff
@@ -0,0 +1,83 @@
...
+
+To avoid further complicating the already complex logic of the controller, this document proposes encapsulating this new functionality within the Connection Diagnostics. This object will interact with both the controller and connection health, serving as a dedicated entity for identifying connection problems and correctly communicating status changes to the UI, ultimately enhancing the user experience.
+
+## Proposed Solution
+
+The Connection Diagnostics will be responsible for managing all probes. These probes are stored as distinct states within an enumeration class, `ConnectionDiagnostics::State`. The Connection Diagnostics will sequentially cycle through each of these states in a systematic order, only proceeding to the next state if the current probe is successful. If all probes pass but VPN stability is still `NoSignal`, then we would not present any modals to the user, but the controller will continue to show "No Signal" as we do currently.
```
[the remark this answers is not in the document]
Gela: The duagnostics states under `Implementation Plan` are listed in the correct order of execution. I'll mention that specifically.

vinocher (approved):
Looks good!

docs/rfcs/0004-rfc-connectiomanager.md
lesleyjanenorton: Resolving this because it's outdated and I'm struggling to load 100% of the context for this comment back into memory.  I think what I was getting at is the possibility of sending the wrong error message to the user when, for random example, when running checks on activation, we fail at the subscription check but the reason for the failure is actually that there is no internet or a given server is unavailable.

lesleyjanenorton (approved):

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: I don't feel like it *needs* to be acknowledged here, but I want to make sure we don't forget it. If the eventual metrics doc is not engineer-led, I'm not sure this situation would be remembered. But I don't think you need to add something here.

docs/rfcs/0004-rfc-connectiomanager.md
mcleinman: I didn't know this change was coming, thank you! In relation to this doc - does that mean a captive portal check will be part of the monitoring, or not?

mcleinman: Nevermind, I just saw your update - sorry. This sounds great, thanks.

PR #8491, opened 2023-11-06
VPN-5421 -[Android] Update qt next to 6.6.0

strseb:
## Description
It seems to fix the actvity-recreate issue we need to fix 2 things: 
-> Update Qt to 6.6.0, so it fixes it's own issues stopping the recreation
-> Fixing a resource leak which would trow on the recreation of the activity. 

I'just noticed that the qt6 next task actually never used 6.5.0 like expected, so fixing that too and making it more explicit which toolchain tasks are using which versions of qt by infating the task parameters in a transform. 
That should make that file look less boilerplaty :)


lottie/CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_SOURCE_DIR}/lib/lottie/lottie.mjs.cmake
 )
 
+# Dont build the test's on android ... for now.
```
brizental, line 52: Please refer to the ticket where this will be done in this comment.

android/vpnClient/src/main/java/org/mozilla/firefox/vpn/qt/VPNActivity.java
```diff
@@ -201,4 +201,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
     }
     super.onActivityResult(requestCode, resultCode, data);
   }
+
+  @Override
+  protected void onDestroy() {
+    unbindService(mConnection);
+    super.onDestroy();
+  }
```
brizental, line 209: How come this is relate to the Qt update?

brizental (approved):

android/vpnClient/src/main/java/org/mozilla/firefox/vpn/qt/VPNActivity.java
strseb: We have been leaking this, however due to QT failing to recreate the activity in <6.6 we never noticed. Now we would see that.

PR #8496, opened 2023-11-07
Nit: Pin our Python dependencies 

strseb:
I have been wanting to do this since forever. 
Now that we have a proper conda env, the same version of python everywhere, let's pin our used versions.


brizental (approved):

bhearsum (approved):
I wouldn't call this a nit, but rather an important security and stability feature :)

PR #8498, opened 2023-11-07
No ticket - Nit: Fix wasm canvas container

brizental:
I recently changed the dimensos of the wasm canvas, because turns out that can cause testing issues.

Why? Because in order to click on something for tests it needs to be in screen, if it's below or above the fold the click is unsuccessful. Yes, we could have special cased WASM clicks and so on, but that sounds excessive, instead I changed the WASM canvas to always have the same dimensions as the Desktop client. (When we run these tests on mobile devices this may come back to bite us, but then we can come up with a more elegant solution e.g. always scroll to element before a click).

Anyways, that made the WASM client look weird. Just go to https://mozilla-mobile.github.io/mozilla-vpn-client/?branch=main and see for yourself. With this change it will look like this:

Screenshot 2023-11-07 at 17 39 02

A little bit better 👀

### DIFF
--- a/wasm/main.css
+++ b/wasm/main.css
@@ -182,11 +182,9 @@ canvas {
 .canvas-bg {
 	box-shadow: var(--boxShadow);
 	border-radius: 2rem;
-	overflow: hidden;
 	background-color: rgba(14, 13, 26, 0.05);
-	min-width: 400px;
-	height: 70vh;
-	max-height: 760px;
+	width: fit-content;
+	height: fit-content;
 }
 
 .wasm-loaded .canvas-bg {


wasm/main.css
```diff
@@ -162,11 +162,10 @@ canvas {
...
 
 .control-bg {
-	background: rgba(14, 13, 26, 0.05);
 	border-radius: 2rem;
-	max-width: 500px;
-	height: 300px;
-	overflow: hidden;
+	width: fit-content;
+	height: fit-content;
+	background-color: transparent !important;
```
brizental, line 168:
```suggestion
```

wasm/main.css
```diff
@@ -182,11 +181,9 @@ canvas {
 .canvas-bg {
 	box-shadow: var(--boxShadow);
 	border-radius: 2rem;
-	overflow: hidden;
+	overflow: scroll;
```
brizental, line 184:
```suggestion
```

strseb (approved):

PR #8504, opened 2023-11-08
Drop pidtracker class as it is not used anymore and causes compilation issues

Cimbali:
## Description

The pidtracker class is not used and is dead code, as linux split-tunnelling is using the apptracker instead (which is more robust and uses cgroups v2).

<details><summary>Since kernel 6.5.6 and specifically <a href="https://www.github.com/torvalds/linux/commit/743acf351bae1ff7ff4aaadd6a406d4d6091d90b">this cn_proc performance patch</a>, the C header <code>cn_proc.h</code> is not compatible with C++ anymore.</summary>

The code causing the issue is
```
static inline enum proc_cn_event valid_event(enum proc_cn_event ev_type)
{
    ev_type &= PROC_EVENT_ALL;
    return ev_type;
}
```
where `PROC_EVENT_ALL` is a `|`-aggregate of the enum values, but for C++ `PROC_EVENT_ALL` is an unsigned int instead of the underlying type of the enum (which is int). Confirmed with both gcc and llvm.
</details>

Therefore compilation of mozillavpn fails with the following error:
```
/usr/include/linux/cn_proc.h: In function ‘proc_cn_event valid_event(proc_cn_event)’:
/usr/include/linux/cn_proc.h:72:17: error: invalid conversion from ‘unsigned int’ to ‘proc_cn_event’ [-fpermissive]
   72 |         ev_type &= PROC_EVENT_ALL;
      |                 ^
      |                 |
      |                 unsigned int
```

This PR removes the dead code (class and associated files), which fixes the issue.

## Reference

none

## Checklist
    

strseb (approved):

oskirby (approved):
Thank you!

PR #8509, opened 2023-11-08
Nit: don't download wintun for each build

strseb:
## Description
We're currently downloading wintun on every init of the build folder from the interwebs. 
Now today that has been flaky, so let's have taskcluster fetch it and use that instead, so we don't hit the server everytime ,__,


oskirby (approved):
LGTM

PR #8510, opened 2023-11-08
VPN-4846: FPN trial subscription expiring message

MattLichtenstein:
## Description

- Message that will be sent to FPN users who were given a 6 month free trial
- Does not need translations
  - FPN was US-only so we're restricting this message to users who have there language set to english
- Timestamp is set to 20:00 GMT as we anticipate this message going out at that time or a little later

## Reference

[VPN-4846: Phase 3: Implement and release the in-app message for the free VPN users](https://mozilla-hub.atlassian.net/browse/VPN-4846)

### DIFF
--- a/addons/message_fpn_trial_subscription_expiring/isFpnTrial.js
+++ b/addons/message_fpn_trial_subscription_expiring/isFpnTrial.js
@@ -1,0 +1,8 @@
+(function(api, condition) {
+  //Enable message for users on the FPN 6 month trial plan only
+  const fpnPlanId = "price_1MzNRCJNcmPzuWtRMCwUWADu"
+  const subscriptionDataJson = JSON.parse(api.settings.subscriptionData)
+  const subscriptionPlanId = subscriptionDataJson["plan_id"]
+
+  subscriptionPlanId === fpnPlanId ? condition.enable() : condition.disable()
+});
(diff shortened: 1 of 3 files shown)


brizental (approved):

strseb (approved):

MattLichtenstein:
@flodolo merging without your review due to urgency, but no translations are necessary for this message addon

addons/message_fpn_trial_subscription_expiring/isFpnTrial.js
```diff
@@ -0,0 +1,8 @@
+(function(api, condition) {
+  //Enable message for users on the FPN 6 month trial plan only
+  const fpnPlanId = "price_1MzNRCJNcmPzuWtRMCwUWADu"
+  const subscriptionDataJson = JSON.parse(api.settings.subscriptionData)
+  const subscriptionPlanId = subscriptionDataJson["plan_id"]
+
+  subscriptionPlanId === fpnPlanId ? condition.enable() : condition.disable()
+});
```
MattLichtenstein, line 8: For future reference, a watcher should have been added to monitor `subscriptionData` changes so the condition could be re-evaluated when the subscription changes (eg signing out and back in, or signing into different account). Now, if a user switches from an account that shows this addon, to one that doesn't, they will still see it (and vice versa)

``` js
(function(api, condition) {
  function computeCondition() {

    //Enable message for users on the FPN 6 month trial plan only
    const fpnPlanId = "price_1J0Y1iKb9q6OnNsLXwdOFgDr"
    const subscriptionDataJson = JSON.parse(api.settings.subscriptionData)
    const subscriptionPlanId = subscriptionDataJson["plan_id"]

    subscriptionPlanId === fpnPlanId ? condition.enable() : condition.disable()
  }

  api.connectSignal(api.settings.subscriptionData, 'changed', () => computeCondition());

  computeCondition();
});
```

PR #8527, opened 2023-11-09
VPN-4604 - Add utm params to subscription url

brizental:
The params are added to the login URL and are passed on to the subscription purchase page.


lesleyjanenorton (approved):
Looks good to me. Did we square these values with Marketing?

brizental:
> Looks good to me. Did we square these values with Marketing?

This change came from a ticket filed by Santiago having Yeonjoo and Umit as collaborators. I believe they have done their due diligence.

PR #8541, opened 2023-11-13
Nit: we no longer need to pass the sentry secrets

strseb:
## Description
We fetch that info during runtime, so no need to have 2 ci build paths.

### DIFF
--- a/taskcluster/scripts/build/windows_clang_cl.ps1
+++ b/taskcluster/scripts/build/windows_clang_cl.ps1
@@ -31,7 +31,7 @@ $env:OPENSSL_USE_STATIC_LIBS = "TRUE"
 # Extract the sources
 $SOURCE_DSC = resolve-path "$FETCHES_PATH/mozillavpn_*.dsc"
 $SOURCE_VERSION = ((select-string $SOURCE_DSC -Pattern '^Version:') -split " ")[1]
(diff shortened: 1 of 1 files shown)


brizental (approved):

taskcluster/scripts/build/windows_clang_cl.ps1
```diff
@@ -31,7 +31,7 @@ $env:OPENSSL_USE_STATIC_LIBS = "TRUE"
 # Extract the sources
 $SOURCE_DSC = resolve-path "$FETCHES_PATH/mozillavpn_*.dsc"
 $SOURCE_VERSION = ((select-string $SOURCE_DSC -Pattern '^Version:') -split " ")[1]
-tar -xzvf (resolve-path "$FETCHES_PATH/mozillavpn_$SOURCE_VERSION.orig.tar.gz" -Relative)
+tar -xzvfh (resolve-path "$FETCHES_PATH/mozillavpn_$SOURCE_VERSION.orig.tar.gz" -Relative)
```
Gela, line 34: it looks like -h and -dereference do something similar. How do we know which one is the correct flag to use, is that dependent on the version and could it change?

strseb: Neither seems to work, so i'm removing it q_q

Gela: 🫠

Gela (approved):

PR #8546, opened 2023-11-13
CI: Windows - Remove unsed fetchs 

strseb:
## Description
Most of the build deps should live inside the conda-env, so neither go nor rustup should need to be fetched.


strseb:
> Did you mean to keep win-perl in fetch/kind.yml? 
Yes it's still used by one of the toolchain tasks :)

vinocher (approved):

Issue #8097, opened 2023-09-21
[a11y] When a radio button is selected, the new state is not narrated by the screen reader


data-sync-user:
**Important details:**

* The issue is reproducible for all radio buttons in the application, e.g. in App preferences -> DNS Settings.

**Affected versions:**

* 2.18.0 (2.202309200508);

**Tested Platforms:**

* Windows;

**Prerequisites:**

* Have an active account with a subscription made for Mozilla VPN;

* Have Mozilla VPN app installed on your device;

* Have a screen reader activated (Narrator, VoiceOver)

* Be in Settings ~~> App preferences~~ > Languages;

**Steps to reproduce:**

1. Using the keyboard, select another language;
1. Check how the screen reader narrates the state of the radio button.

**Expected result:**

* The screen reader narrates the “selected” state for the radio button.

**Actual result:**

The screen reader does not narrate the “selected” state of the radio button. 

If “Enter” key is pressed, the radio button is selected, but the screen reader does not narrate the action/state.

If “Space” key is pressed, the radio button is selected, but the screen reader narrates “space”.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5578)
┆Reporter: Raluca Monica Ilban

data-sync-user:
➤ Vinoo Cherian commented:

This is caused a Qt framework limitation. 

This change works around that limitation on Windows by triggering the screen reader to say ‘Selected’ when the Narrator shortcut CapsLock+Enter is pressed to select a radio button. This is triggered only if the focus hasn’t moved away from the radio button on selection, or when another reading that includes the selection state is not said by the screen reader.

FYI, CapsLock+Enter is the Narrator shortcut to activate a control.

data-sync-user:
➤ Raluca Monica Ilban commented:

Verified this as fixed on Mozilla VPN 2.19.0 (2.202311152028) on Windows 10.
When CapsLock+Enter is pressed, the radio button is selected and 'selected' state is narrated by the screen reader.

PR #8548, opened 2023-11-13
VPN-5578: Work around lack of built-in RadioButton selection a11y support on Windows

vinocher:
## Description

The selected state of a Radio Button is not read by the screen reader on Windows  when activated through the Accessibility PressAction command. This is a Qt limitation. Worked around by raising a "Selected" accessible notification. Accessible notifications currently are supported only in Windows by Qt. 

This workaround excludes the following scenarios, as explained in the code comment, to prevent a redundant notification:
1. When the radio button's selection causes focus to change, as seen in the DNS Settings screen, when custom DNS is selected.
2. When the radio button's selection causes the Accessible name to change, as seen in the Languages screen, when an unrelated new language is chosen.

Also, fixed the following:
1. Improved performance in the Languages screen by removing the unnecessary resetting of the model in  `Localizer::settingsChanged` when a new language is selected.  The model reset also broke the screen reading experience on Windows because the old radio item that raised the  focus notification was no longer present after the model reset, since it was replaced by a new radio button.
2. Handled hidden screens in the StackView when raising accessible notifications for elements with changed text due to a new language selection. This ensures that only elements with `Accessible.ignored == false` raise these notifications, preventing unnecessary notifications from hidden text elements (where `Accessible.ignored == true`), when the language is changed.
3. Used the highlight color from the themes file instead of hard coding, in the Languages screen

## Reference
[VPN-5578](https://mozilla-hub.atlassian.net/browse/VPN-5578), [GitHub 8097](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/8097)

## Checklist
    
[VPN-5578]: https://mozilla-hub.atlassian.net/browse/VPN-5578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -737,6 +737,9 @@ accessibility:
       - Used by screen readers to denote that a particular item is currently selected.
       - For example, given a segmented control with 2 items, where the first item is currently selected. We want the screen reader to read out "Selected, <item name>".
       - The %1 represents the name of the selected item.
+  selected:
+    value: "Selected"
+    comment: Used by screen reader to indicate that a user action selected an item.
 
 freeTrials:
   startYourFreeTrial:
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -738,7 +738,7 @@ accessibility:
       - For example, given a segmented control with 2 items, where the first item is currently selected. We want the screen reader to read out "Selected, <item name>".
       - The %1 represents the name of the selected item.
   selected:
-    value: "Selected"
+    value: Selected
     comment: Used by screen reader to indicate that a user action selected an item.
 
 freeTrials:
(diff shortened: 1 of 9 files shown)


src/translations/strings.yaml
```diff
@@ -737,6 +737,9 @@ accessibility:
       - Used by screen readers to denote that a particular item is currently selected.
       - For example, given a segmented control with 2 items, where the first item is currently selected. We want the screen reader to read out "Selected, <item name>".
       - The %1 represents the name of the selected item.
+  selected:
+    value: "Selected"
```
flodolo, line 741: You don't need quotes, there are no special characters here.

flodolo (changes requested):

src/translations/strings.yaml
vinocher: Thanks, fixed.

flodolo (approved):

MattLichtenstein (approved):
Qt accessibility support 😠 ✊ 

unfortunate that we have to do this, but nice workaround nonetheless!

PR #8551, opened 2023-11-13
VPN-5385 - register callback to make Android back work

mcleinman:
## Description

The intended flow of a back button press:
- `handleBackButton` is called
- Which calls `Navigator::instance()->eventHandled()`
- One of the first lines of that function is `if (!ExternalOpHandler::instance()->request(
          ExternalOpHandler::OpCloseEvent))`... 
- ... and in the middle of that function lies these two lines: `Q_ASSERT(m_ops.contains(op));` and `if (!m_ops.contains(op))`.
- Since we weren't registering this event, it was failing the assert (or on release, always returning `false` due to this if statement).

My guess is this bug was introduced in flattening the app (or making it a multi-app repo), but I didn't do the research to confirm this.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5385

## Checklist
    
### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -2281,6 +2281,8 @@ void MozillaVPN::registerExternalOperations() {
         new TaskControllerAction(TaskControllerAction::eDeactivate));
   });
 
+  eoh->registerExternalOperation(ExternalOpHandler::OpCloseEvent, []() {});
+
   eoh->registerExternalOperation(OpNotificationClicked, []() {});
 
   eoh->registerExternalOperation(


src/mozillavpn.cpp
```diff
@@ -2281,6 +2281,8 @@ void MozillaVPN::registerExternalOperations() {
         new TaskControllerAction(TaskControllerAction::eDeactivate));
   });
 
+  eoh->registerExternalOperation(ExternalOpHandler::OpCloseEvent, []() {});
```
vinocher, line 2284: Should this be registered on all platforms, since some platforms don't support this?

vinocher:
Should  Navigator::eventHandled be renamed? Perhaps it should be Navigator::handleOperation(ExternalOpHandler::Op operation) and take the operation as a parameter. Or if the operation is implicitly the Close operation, perhaps it should be renamed to Navigator::handleCloseOperation. Feel free to ignore this comment, because it relates to exisiting code.

lesleyjanenorton (commented):
Just a heads up that this will be affected by -possibly automagically resolved by (?)- #8534 in which `ExternalOpHandler` is removed alongside other tutorial related ephemera.

mcleinman:
> Just a heads up that this will be affected by -possibly automagically resolved by (?)- #8534 in which `ExternalOpHandler` is removed alongside other tutorial related ephemera.

Thanks for the heads up. I believe [this deletion](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8534/files#diff-bf5452c5d9bf001e7b2cf467b8ef9ea5d752617491b1aeaf711f7e5a674e25d6L311-L316) will automatigically make this work again.

That said, since we're coming up on the 2.19 deadline, is it harmful to get this in (in case the big removal doesn't get in there)?

src/mozillavpn.cpp
mcleinman: Within `bool Navigator::eventHandled()`, we check for eligible platforms before running the relevant code - it's wrapped in `#if defined(MZ_ANDROID) || defined(MZ_WASM)`. Do you think that suffices, or should we similarly wrap this here?

mcleinman: Also as a sidenote - this line will be removed when the PR Lesley mentioned gets merged in - so this is a temporary thing either way.

strseb (approved):

src/mozillavpn.cpp
vinocher: Sounds good.

vinocher (approved):

mcleinman:
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8534 has been merged in, and as expected that fixes this bug. Closing this PR.

PR #8555, opened 2023-11-14
VPN-5766 - Fix CountdownTimer generation timing

brizental:
When the generation of this object was changed to be lazy, it would be created inside the `VPNService.turnOn` method. That method is called in a thread that is not equipped to create timers causing a crash and thus the issue described on VPN-5766.


strseb (changes requested):
add tests

strseb (approved):

Issue #7988, opened 2023-09-12
Automatic restart of macosdaemon on crash


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-4228)
┆Reporter: Naomi Kirby

data-sync-user:
➤ Bianca Hidecuti commented:

Naomi Kirby, please take a look at my previous question when you get a chance. Thanks!

data-sync-user:
➤ Naomi Kirby commented:

The force_daemon_crash command added to the inspector for https://mozilla-hub.atlassian.net/browse/VPN-5931 ( https://mozilla-hub.atlassian.net/browse/VPN-5931|smart-link ) can be used to test this scenario. This command will cause the daemon to crash, after which the client will likely return to a disconnected state if it was connected, but should be remain usable and the user should be able to re-start the connection.

data-sync-user:
➤ Valentina Virlics commented:

Verified this while on MacOS M1 - 14, while using Mozilla VPN 2.20 (202312160138) build. The “force_daemon_crash“ command is available in Inspector, running it causes the VPN to disconnect and the “Background service restore“ message is displayed. Trying to turn the VPN ON is possible with success.

PR #8562, opened 2023-11-14
VPN-4228: MacOS daemon recovery on crash

oskirby:
## Description
In this PR we attempt to make the VPN client and daemon on MacOS more resilient to abnormal exit conditions in the VPN tunnel or daemon. To accomplish this goal, we made a few changes and tweaks:
- Check for and handle errors in the UNIX socket, and re-initialize of the socket closes on us.
- Remove the "try again" link in the backend failure banner, it doesn't do anything, and it never did anything.
- Permit the controller to re-initialize itself after a backend failure.
- Add optional timeouts on `LocalSocketController::write()` to catch cases where the daemon is hung.

This alone was sufficient to get the client to recover from a crash of the daemon. But there was still a problem: if the daemon crashes while the tunnel is active this will bring the tunnel down, but we don't get to run any cleanup code. This potentially results in leaving the user's DNS configuration in a broken state. To address this, the `DnsUtilsMacos` class was written to fork itself and handle this cleanup from a separate process.

While I was investigating this, I also had a look at converting the MacOS daemon into an on-demand startup mode, as supported by the `Sockets` dictionary in the Launchd [plist](https://www.manpagez.com/man/5/launchd.plist/). However, this would require moving the control socket path, which could bring with it potential compatibility issues and breakage. For this reason, I opted to leave the daemon's startup and keepalive behaviour as-is.

## Reference
Github issue #7988 ([VPN-4228](https://mozilla-hub.atlassian.net/browse/VPN-4228))
Github issue #8665 ([VPN-5931](https://mozilla-hub.atlassian.net/browse/VPN-5931))

## Checklist
    
[VPN-4228]: https://mozilla-hub.atlassian.net/browse/VPN-4228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/MZSystemAlert.qml
+++ b/nebula/ui/components/MZSystemAlert.qml
@@ -72,13 +72,7 @@ MZAlert {
                     target: alertBox
                     //% "Background service error"
                     alertText: qsTrId("vpn.alert.backendServiceError")
-                    //% "Restore"
-                    //: Restore a service in case of error.
-                    alertActionText: qsTrId("vpn.alert.restore")
                     visible: true
-                    onActionPressed: ()=>{
-                        VPN.backendServiceRestore();
-                    }
                 }
             },
             State {
--- a/nebula/ui/components/MZSystemAlert.qml
+++ b/nebula/ui/components/MZSystemAlert.qml
@@ -81,9 +81,6 @@ MZAlert {
                     target: alertBox
                     alertText: qsTrId("vpn.alert.backendServiceError")
                     visible: true
-                    onActionPressed: ()=>{
-                        VPN.backendServiceRestore();
-                    }
                 }
             },
             State {
@@ -93,9 +90,6 @@ MZAlert {
                     //% "Remote service error"
                     alertText: qsTrId("vpn.alert.remoteServiceError")
                     visible: true
-                    onActionPressed: ()=>{
-                        VPN.backendServiceRestore();
-                    }
                 }
             },
             State {
--- a/src/controllerimpl.h
+++ b/src/controllerimpl.h
@@ -51,6 +51,8 @@ class ControllerImpl : public QObject {
   // tunnel cannot be reactivated in system settings.
   virtual void deleteOSTunnelConfig(){};
 
+  virtual void testDaemonCrash() { /* not implemented */ }
+
   // This method is used to retrieve the VPN tunnel status (mainly the number
   // of bytes sent and received). It's called always when the VPN tunnel is
   // active.
--- a/src/controllerimpl.h
+++ b/src/controllerimpl.h
@@ -53,7 +53,7 @@ class ControllerImpl : public QObject {
 
   // This method attempts to force the daemon to crash, and is used for
   // testing the ability of the VPN to recover from a backend error.
-  virtual void testDaemonCrash() {}
+  virtual void forceDaemonCrash() {}
 
   // This method is used to retrieve the VPN tunnel status (mainly the number
   // of bytes sent and received). It's called always when the VPN tunnel is
--- a/src/controllerimpl.h
+++ b/src/controllerimpl.h
@@ -51,7 +51,7 @@ class ControllerImpl : public QObject {
   // tunnel cannot be reactivated in system settings.
   virtual void deleteOSTunnelConfig(){};
 
-  virtual void testDaemonCrash() { /* not implemented */ }
+  virtual void testDaemonCrash(){}
 
   // This method is used to retrieve the VPN tunnel status (mainly the number
   // of bytes sent and received). It's called always when the VPN tunnel is
--- a/src/controllerimpl.h
+++ b/src/controllerimpl.h
@@ -51,7 +51,9 @@ class ControllerImpl : public QObject {
   // tunnel cannot be reactivated in system settings.
   virtual void deleteOSTunnelConfig(){};
 
-  virtual void testDaemonCrash(){}
+  // This method attempts to force the daemon to crash, and is used for
+  // testing the ability of the VPN to recover from a backend error.
+  virtual void testDaemonCrash() {}
 
   // This method is used to retrieve the VPN tunnel status (mainly the number
   // of bytes sent and received). It's called always when the VPN tunnel is
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -4,23 +4,14 @@
 
 #include "localsocketcontroller.h"
 
-#include <QDir>
 #include <QFileInfo>
-#include <QHostAddress>
-#include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QJsonValue>
-#include <QStandardPaths>
 
 #include "errorhandler.h"
-#include "ipaddress.h"
 #include "leakdetector.h"
 #include "logger.h"
-#include "models/device.h"
-#include "models/keys.h"
-#include "models/server.h"
-#include "settingsholder.h"
 
 // When the daemon is unreachable, we will retry indefinitely using an
 // exponential backoff algorithm. The interval between retries starts at
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -16,14 +16,13 @@
 #ifdef MZ_MACOS
 #include <Security/Authorization.h>
 #include <Security/AuthorizationTags.h>
+#include <signal.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
 
 #include <QProcess>
 #include <QScopeGuard>
-
-#include <sys/socket.h>
-#include <sys/un.h>
-#include <signal.h>
-#include <unistd.h>
 #endif
 
 // When the daemon is unreachable, we will retry indefinitely using an
@@ -417,20 +416,23 @@ void LocalSocketController::testDaemonCrash() {
 
   // Create an authorization session.
   AuthorizationRef authRef;
-  AuthorizationFlags authFlags = kAuthorizationFlagDefaults | kAuthorizationFlagInteractionAllowed | kAuthorizationFlagPreAuthorize | kAuthorizationFlagExtendRights; 
-  OSStatus status = AuthorizationCreate(nullptr, kAuthorizationEmptyEnvironment, authFlags, &authRef);
+  AuthorizationFlags authFlags =
+      kAuthorizationFlagDefaults | kAuthorizationFlagInteractionAllowed |
+      kAuthorizationFlagPreAuthorize | kAuthorizationFlagExtendRights; 
+  OSStatus status = AuthorizationCreate(nullptr, kAuthorizationEmptyEnvironment,
+                                        authFlags, &authRef);
   if (status != errAuthorizationSuccess) {
     logger.error() << "Failed to acquire authorization:" << status;
     return;
   }
-  auto authGuard = qScopeGuard([&] {
-    AuthorizationFree(authRef, kAuthorizationFlagDefaults);
-  });
+  auto authGuard = qScopeGuard(
+      [&] { AuthorizationFree(authRef, kAuthorizationFlagDefaults); });
 
   // Acquire execution permissions.
-  AuthorizationItem authItems = { kAuthorizationRightExecute, 0, nullptr, 0 };
-  AuthorizationRights authRights = { 1, &authItems };
-  status = AuthorizationCopyRights(authRef, &authRights, nullptr, authFlags, nullptr);
+  AuthorizationItem authItems = {kAuthorizationRightExecute, 0, nullptr, 0};
+  AuthorizationRights authRights = {1, &authItems};
+  status = AuthorizationCopyRights(authRef, &authRights, nullptr, authFlags,
+                                   nullptr);
   if (status != errAuthorizationSuccess) {
     logger.error() << "Failed to copy authorization rights:" << status;
     return;
@@ -441,12 +443,13 @@ void LocalSocketController::testDaemonCrash() {
   QByteArray pidString = QString::number(pid).toUtf8();
   char killpath[] = "/bin/kill";
   char killsignal[] = "-SEGV";
-  char* const killargs[] = { killsignal, pidString.data(), nullptr };
+  char* const killargs[] = {killsignal, pidString.data(), nullptr};
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-  status = AuthorizationExecuteWithPrivileges(authRef, killpath, kAuthorizationFlagDefaults, killargs, nullptr);
-#pragma clang diagnostic pop
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wdeprecated-declarations"
+  status = AuthorizationExecuteWithPrivileges(
+      authRef, killpath, kAuthorizationFlagDefaults, killargs, nullptr);
+#  pragma clang diagnostic pop
   if (status != errAuthorizationSuccess) {
     logger.error() << "Failed to copy execute tool:" << status;
     return;
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -404,7 +404,7 @@ void LocalSocketController::clearAllTimeouts() {
   }
 }
 
-void LocalSocketController::testDaemonCrash() {
+void LocalSocketController::forceDaemonCrash() {
 #ifdef MZ_MACOS
   pid_t pid;
   socklen_t len = sizeof(pid);
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -13,6 +13,7 @@
 #include <QJsonValue>
 #include <QStandardPaths>
 
+#include "constants.h"
 #include "errorhandler.h"
 #include "ipaddress.h"
 #include "leakdetector.h"
@@ -22,11 +23,9 @@
 #include "models/server.h"
 #include "settingsholder.h"
 
-// How many times do we try to reconnect.
-constexpr int MAX_CONNECTION_RETRY = 10;
-
 // How long do we wait between one try and the next one.
-constexpr int CONNECTION_RETRY_TIMER_MSEC = 500;
+constexpr int CONNECTION_RETRY_INITIAL_MSEC = 500;
+constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000;
 
 namespace {
 Logger logger("LocalSocketController");
@@ -45,9 +44,15 @@ LocalSocketController::LocalSocketController() {
   connect(m_socket, &QLocalSocket::readyRead, this,
           &LocalSocketController::readData);
 
+  m_initializingInterval = CONNECTION_RETRY_INITIAL_MSEC;
   m_initializingTimer.setSingleShot(true);
   connect(&m_initializingTimer, &QTimer::timeout, this,
           &LocalSocketController::initializeInternal);
+
+  m_messageTimeout.setSingleShot(true);
+  connect(&m_messageTimeout, &QTimer::timeout, this, [&] {
+    this->errorOccurred(QLocalSocket::SocketTimeoutError);
+  });
 }
 
 LocalSocketController::~LocalSocketController() {
@@ -56,26 +61,20 @@ LocalSocketController::~LocalSocketController() {
 
 void LocalSocketController::errorOccurred(
     QLocalSocket::LocalSocketError error) {
-  logger.error() << "Error occurred:" << error;
+  logger.error() << "Error occurred:" << m_socket->errorString();
 
-  if (m_daemonState == eInitializing) {
-    if (m_initializingRetry++ < MAX_CONNECTION_RETRY) {
-      m_initializingTimer.start(CONNECTION_RETRY_TIMER_MSEC);
-      return;
-    }
-
-    emit initialized(false, false, QDateTime());
+  if (m_daemonState != eInitializing) {
+    REPORTERROR(ErrorHandler::ControllerError, "controller");
+    emit disconnected();
   }
 
-  REPORTERROR(ErrorHandler::ControllerError, "controller");
-  disconnectInternal();
+  m_initializingTimer.start(m_initializingInterval);
 }
 
 void LocalSocketController::disconnectInternal() {
   // We're still eReady as the Deamon is alive
   // and can make a new connection.
   m_daemonState = eReady;
-  m_initializingRetry = 0;
   m_initializingTimer.stop();
   emit disconnected();
 }
@@ -87,12 +86,17 @@ void LocalSocketController::initialize(const Device* device, const Keys* keys) {
   Q_UNUSED(keys);
 
   Q_ASSERT(m_daemonState == eUnknown);
-  m_initializingRetry = 0;
 
   initializeInternal();
 }
 
 void LocalSocketController::initializeInternal() {
+  // Perform an exponential backoff.
+  if (m_daemonState != eInitializing) {
+    m_initializingInterval = CONNECTION_RETRY_INITIAL_MSEC;
+  } else if (m_initializingInterval < CONNECTION_RETRY_MAXIMUM_MSEC) {
+    m_initializingInterval *= 2;
+  }
   m_daemonState = eInitializing;
 
 #ifdef MZ_WINDOWS
@@ -152,7 +156,7 @@ void LocalSocketController::checkStatus() {
 
     QJsonObject json;
     json.insert("type", "status");
-    write(json);
+    write(json, "status");
   }
 }
 
@@ -239,6 +243,10 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
   QString type = typeValue.toString();
 
   logger.debug() << "Parse command:" << type;
+  if (!m_messageExpected.isEmpty() && (m_messageExpected == type)) {
+    m_messageExpected.clear();
+    m_messageTimeout.stop();
+  }
 
   if (m_daemonState == eInitializing && type == "status") {
     m_daemonState = eReady;
@@ -343,9 +351,17 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
   logger.warning() << "Invalid command received:" << command;
 }
 
-void LocalSocketController::write(const QJsonObject& json) {
+void LocalSocketController::write(const QJsonObject& json,
+                                  const QString& expect) {
+  QByteArray payload = QJsonDocument(json).toJson(QJsonDocument::Compact);
+  payload.append('\n');
+
+  if (!expect.isEmpty()) {
+    m_messageExpected = expect;
+    m_messageTimeout.start(CONNECTION_RETRY_INITIAL_MSEC);
+  }
+
   Q_ASSERT(m_socket);
-  m_socket->write(QJsonDocument(json).toJson(QJsonDocument::Compact));
-  m_socket->write("\n");
+  m_socket->write(payload);
   m_socket->flush();
 }
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -384,6 +384,9 @@ void LocalSocketController::write(const QJsonObject& message,
 }
 
 void LocalSocketController::clearTimeout(const QString& responseType) {
+  // We assume that responses arrive in the same order, the local socket
+  // API has no mechanism to detect out-of-order responses, and any change
+  // here to add such a mechanism would risk a compatibility issue.
   for (QTimer* t : m_responseTimeouts) {
     QVariant timerResponseType = t->property("responseType");
     if (timerResponseType.type() != QVariant::String) {
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -359,7 +359,7 @@ void LocalSocketController::write(const QJsonObject& json,
   // throw an error if that response fails to arrive in a timely manner. This
   // is used to detect a crash or failure of the daemon.
   if (!expect.isEmpty()) {
-    QTimer *t = new QTimer(this);
+    QTimer* t = new QTimer(this);
     connect(t, &QTimer::timeout, this,
             [&] { this->errorOccurred(QLocalSocket::SocketTimeoutError); });
 
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -13,6 +13,15 @@
 #include "leakdetector.h"
 #include "logger.h"
 
+#ifdef MZ_MACOS
+#include <QProcess>
+
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <signal.h>
+#include <unistd.h>
+#endif
+
 // When the daemon is unreachable, we will retry indefinitely using an
 // exponential backoff algorithm. The interval between retries starts at
 // the initial value, and doubles after each failed connection attempt,
@@ -388,4 +397,28 @@ void LocalSocketController::clearAllTimeouts(void) {
     QTimer* t = m_expectedResponses.takeFirst();
     delete t;
   }
+}
+
+void LocalSocketController::testDaemonCrash() {
+#ifdef MZ_MACOS
+  pid_t pid;
+  socklen_t len = sizeof(pid);
+  int sd = m_socket->socketDescriptor();
+  if (getsockopt(sd, SOL_LOCAL, LOCAL_PEERPID, &pid, &len) < 0) {
+    return;
+  }
+  if ((pid <= 0) || (pid == getpid())) {
+    return;
+  }
+  logger.warning() << "Sending SIGSEGV to:" << pid;
+
+  QProcess proc;
+  QStringList args;
+  args.append("-e");
+  args.append(QString("do shell script \"kill -SEGV %1\" with administrator privileges").arg(pid));
+  proc.start("osascript", args);
+  if (proc.waitForStarted()) {
+    proc.waitForFinished();
+  }
+#endif
 }
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -20,8 +20,8 @@
 //
 // This means that if the daemon never starts, the steady-state behaviour will
 // have client to retry every 16 seconds indefinitely.
-constexpr int CONNECTION_RETRY_INITIAL_MSEC = 500; // 0.5 seconds
-constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000; // 16 seconds
+constexpr int CONNECTION_RETRY_INITIAL_MSEC = 500;    // 0.5 seconds
+constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000;  // 16 seconds
 
 namespace {
 Logger logger("LocalSocketController");
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -14,15 +14,15 @@
 #include "logger.h"
 
 #ifdef MZ_MACOS
-#include <Security/Authorization.h>
-#include <Security/AuthorizationTags.h>
-#include <signal.h>
-#include <sys/socket.h>
-#include <sys/un.h>
-#include <unistd.h>
+#  include <Security/Authorization.h>
+#  include <Security/AuthorizationTags.h>
+#  include <signal.h>
+#  include <sys/socket.h>
+#  include <sys/un.h>
+#  include <unistd.h>
 
-#include <QProcess>
-#include <QScopeGuard>
+#  include <QProcess>
+#  include <QScopeGuard>
 #endif
 
 // When the daemon is unreachable, we will retry indefinitely using an
@@ -420,7 +420,7 @@ void LocalSocketController::testDaemonCrash() {
   AuthorizationRef authRef;
   AuthorizationFlags authFlags =
       kAuthorizationFlagDefaults | kAuthorizationFlagInteractionAllowed |
-      kAuthorizationFlagPreAuthorize | kAuthorizationFlagExtendRights; 
+      kAuthorizationFlagPreAuthorize | kAuthorizationFlagExtendRights;
   OSStatus status = AuthorizationCreate(nullptr, kAuthorizationEmptyEnvironment,
                                         authFlags, &authRef);
   if (status != errAuthorizationSuccess) {
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -50,14 +50,11 @@ LocalSocketController::LocalSocketController() {
   m_initializingTimer.setSingleShot(true);
   connect(&m_initializingTimer, &QTimer::timeout, this,
           &LocalSocketController::initializeInternal);
-
-  m_responseTimeout.setSingleShot(true);
-  connect(&m_responseTimeout, &QTimer::timeout, this,
-          [&] { this->errorOccurred(QLocalSocket::SocketTimeoutError); });
 }
 
 LocalSocketController::~LocalSocketController() {
   MZ_COUNT_DTOR(LocalSocketController);
+  clearAllTimeouts();
 }
 
 void LocalSocketController::errorOccurred(
@@ -69,6 +66,7 @@ void LocalSocketController::errorOccurred(
     emit disconnected();
   }
 
+  clearAllTimeouts();
   m_initializingTimer.start(m_initializingInterval);
 }
 
@@ -247,10 +245,7 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
   QString type = typeValue.toString();
 
   logger.debug() << "Parse command:" << type;
-  if (!m_responseExpected.isEmpty() && (m_responseExpected == type)) {
-    m_responseExpected.clear();
-    m_responseTimeout.stop();
-  }
+  disarmTimeout(type);
 
   if (m_daemonState == eInitializing && type == "status") {
     m_daemonState = eReady;
@@ -364,11 +359,39 @@ void LocalSocketController::write(const QJsonObject& json,
   // throw an error if that response fails to arrive in a timely manner. This
   // is used to detect a crash or failure of the daemon.
   if (!expect.isEmpty()) {
-    m_responseExpected = expect;
-    m_responseTimeout.start(timeout);
+    QTimer *t = new QTimer(this);
+    connect(t, &QTimer::timeout, this,
+            [&] { this->errorOccurred(QLocalSocket::SocketTimeoutError); });
+
+    t->setProperty("responseType", QVariant(expect));
+    t->setSingleShot(true);
+    t->start(timeout);
+
+    m_expectedResponses.append(t);
   }
 
   Q_ASSERT(m_socket);
   m_socket->write(payload);
   m_socket->flush();
+}
+
+void LocalSocketController::disarmTimeout(const QString& responseType) {
+  for (QTimer* t : m_expectedResponses) {
+    QVariant qv = t->property("responseType");
+    if (qv.type() != QVariant::String) {
+      continue;
+    }
+    if (qv.toString() != responseType) {
+      continue;
+    }
+    m_expectedResponses.removeOne(t);
+    delete t;
+  }
+}
+
+void LocalSocketController::clearAllTimeouts(void) {
+  while (!m_expectedResponses.isEmpty()) {
+    QTimer* t = m_expectedResponses.takeFirst();
+    delete t;
+  }
 }
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -375,7 +375,7 @@ void LocalSocketController::write(const QJsonObject& message,
     t->setSingleShot(true);
     t->start(timeout);
 
-    m_expectedResponses.append(t);
+    m_responseTimeouts.append(t);
   }
 
   Q_ASSERT(m_socket);
@@ -384,7 +384,7 @@ void LocalSocketController::write(const QJsonObject& message,
 }
 
 void LocalSocketController::clearTimeout(const QString& responseType) {
-  for (QTimer* t : m_expectedResponses) {
+  for (QTimer* t : m_responseTimeouts) {
     QVariant qv = t->property("responseType");
     if (qv.type() != QVariant::String) {
       continue;
@@ -392,14 +392,14 @@ void LocalSocketController::clearTimeout(const QString& responseType) {
     if (qv.toString() != responseType) {
       continue;
     }
-    m_expectedResponses.removeOne(t);
+    m_responseTimeouts.removeOne(t);
     delete t;
   }
 }
 
 void LocalSocketController::clearAllTimeouts() {
-  while (!m_expectedResponses.isEmpty()) {
-    QTimer* t = m_expectedResponses.takeFirst();
+  while (!m_responseTimeouts.isEmpty()) {
+    QTimer* t = m_responseTimeouts.takeFirst();
     delete t;
   }
 }
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -31,7 +31,7 @@
 // with time between each retry clamped to the maximum value.
 //
 // This means that if the daemon never starts, the steady-state behaviour will
-// have client to retry every 16 seconds indefinitely.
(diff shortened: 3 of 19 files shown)


src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -150,41 +156,36 @@ bool DnsUtilsMacos::updateResolvers(const QString& ifname,
...
+  auto guard = qScopeGuard([&] { m_dnsSnapshotPid = -1; });
+  if (kill(m_dnsSnapshotPid, SIGINT) < 0) {
+    logger.debug() << "Failed to signal DNS restorer";
+    return false;
   }
 
-  m_prevServices.clear();
+  // Wait for the process to exit, and gather it's status.
+  int status = 0;
+  if (waitpid(m_dnsSnapshotPid, &status, 0) < 0) {
```
vinocher, line 178: To protect against a hung child process, should WNOHANG be used?

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -131,14 +123,28 @@ bool DnsUtilsMacos::updateResolvers(const QString& ifname,
   cfDictSetStringList(dnsConfig, kSCPropNetDNSServerAddresses, list);
   cfDictSetString(dnsConfig, kSCPropNetDNSDomainName, "lan");
 
-  // Backup each network service's DNS config, and replace it with ours.
+  // Take a snapshot of the DNS services, and start a process to restore it.
```
vinocher, line 126: Can this handle the multi-hop scenario where updateResolvers is called multiple times? It looks like multiple watchdogs will be forked in that case, is that correct?

src/localsocketcontroller.h
```diff
@@ -62,7 +62,10 @@ class LocalSocketController final : public ControllerImpl {
   std::function<void(const QString&)> m_logCallback = nullptr;
 
   QTimer m_initializingTimer;
-  uint32_t m_initializingRetry = 0;
+  uint32_t m_initializingInterval;
```
vinocher, line 65: Consider initializing this here.

src/localsocketcontroller.cpp
```diff
@@ -105,6 +108,7 @@ void LocalSocketController::initializeInternal() {
 #endif
 
   logger.debug() << "Connecting to:" << path;
+  m_socket->disconnectFromServer();
```
vinocher, line 111: Can this hang or wait? The docs indicate that this can wait. Do we know if it can handle a crashed daemon if the socket has data waiting to be read?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +351,17 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
   logger.warning() << "Invalid command received:" << command;
 }
 
-void LocalSocketController::write(const QJsonObject& json) {
+void LocalSocketController::write(const QJsonObject& json,
+                                  const QString& expect) {
```
vinocher, line 355: The `expect` parameter could use a comment. Also, should it and m_messageExpected be renamed to something like expectedResponse?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +351,17 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+  QByteArray payload = QJsonDocument(json).toJson(QJsonDocument::Compact);
+  payload.append('\n');
+
+  if (!expect.isEmpty()) {
+    m_messageExpected = expect;
+    m_messageTimeout.start(CONNECTION_RETRY_INITIAL_MSEC);
```
vinocher, line 361: Should this use m_initializingInterval?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +351,17 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
 }
 
-void LocalSocketController::write(const QJsonObject& json) {
+void LocalSocketController::write(const QJsonObject& json,
+                                  const QString& expect) {
+  QByteArray payload = QJsonDocument(json).toJson(QJsonDocument::Compact);
+  payload.append('\n');
+
+  if (!expect.isEmpty()) {
+    m_messageExpected = expect;
```
vinocher, line 360: This m_messageExpected design assumes that the daemon will respond before the next write. Is that always true?

Also, does LocalSocketController::getBackendLogs need to pass an expected message?

src/localsocketcontroller.cpp
```diff
@@ -87,12 +85,17 @@ void LocalSocketController::initialize(const Device* device, const Keys* keys) {
   Q_UNUSED(keys);
 
   Q_ASSERT(m_daemonState == eUnknown);
-  m_initializingRetry = 0;
 
   initializeInternal();
 }
 
 void LocalSocketController::initializeInternal() {
+  // Perform an exponential backoff.
```
vinocher, line 93: A comment that describes this design to detect a non-responding daemon would be useful here or in the header file.

nebula/ui/components/MZSystemAlert.qml
```diff
@@ -72,13 +72,7 @@ MZAlert {
                     target: alertBox
                     //% "Background service error"
                     alertText: qsTrId("vpn.alert.backendServiceError")
-                    //% "Restore"
-                    //: Restore a service in case of error.
-                    alertActionText: qsTrId("vpn.alert.restore")
```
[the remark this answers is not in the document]
Gela: @vinocher I believe removing the comment `//% "Restore"` removes the string moving forward. There's not much else to be done

nebula/ui/components/MZSystemAlert.qml
```diff
@@ -72,13 +72,7 @@ MZAlert {
...
                     visible: true
-                    onActionPressed: ()=>{
-                        VPN.backendServiceRestore();
```
Gela, line 80: Can we remove `backendServiceRestore()` from the code altogether since it's not doing anything?

```c++
void MozillaVPN::backendServiceRestore() {
  logger.debug() << "Background service restore request";
  // TODO
}
```

src/connectionmanager.cpp
```diff
@@ -187,8 +187,6 @@ void ConnectionManager::implInitialized(bool status, bool a_connected,
                  << "connected:" << a_connected
                  << "connectionDate:" << connectionDate.toString();
 
-  Q_ASSERT(m_state == ConnectionManager::StateInitializing);
```
Gela, line 190: Why was this behavior changed?

src/localsocketcontroller.cpp
```diff
@@ -22,11 +23,9 @@
 #include "models/server.h"
 #include "settingsholder.h"
 
-// How many times do we try to reconnect.
-constexpr int MAX_CONNECTION_RETRY = 10;
-
 // How long do we wait between one try and the next one.
-constexpr int CONNECTION_RETRY_TIMER_MSEC = 500;
+constexpr int CONNECTION_RETRY_INITIAL_MSEC = 500;
+constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000;
```
Gela, line 28: Can you expand on the comment on line 26 to explain why there are 2 values? Do we retry every 500ms, up until 16000ms?

brizental (commented):
Gave this an initial pass, I need to dig deeper still to understand the DNS changes and provide useful feedback in there. Submiting this partial either way.

For now, is it possible to write test for the new behaviour?

src/platforms/macos/daemon/dnsutilsmacos.h
```diff
@@ -45,7 +41,11 @@ class DnsUtilsMacos final : public DnsUtils {
   };
 
   SCDynamicStoreRef m_scStore = nullptr;
-  QMap<QString, DnsBackup> m_prevServices;
+  int m_dnsSnapshotPid = -1;
```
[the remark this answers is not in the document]
oskirby: I thought it already was, oops! I'll fix that.

nebula/ui/components/MZSystemAlert.qml
oskirby: This method is still callable through a few other code paths in the project, so I didn't feel it was quite ready for removal until we can clean the rest of them up. In particular, we can still get here from an `ErrorHandler::RemoteServiceError` or an `ErrorHandler::UnrecoverableError`

Although, the name no longer makes sense for these cases.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -219,5 +220,93 @@ void DnsUtilsMacos::backupService(const QString& uuid) {
...
+    QString uuid = service.section('/', 3, 3);
+    if (uuid.isEmpty()) {
+      continue;
+    }
+    snapshot[uuid] = backupService(uuid);
+  }
+  CFRelease(netServices);
+
+  // Fork the process. The parent/daemon continues on doing daemon stuff.
+  int pid = fork();
```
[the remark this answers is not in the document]
oskirby: Modern operating systems are very good about sharing memory pages, so the incremental memory usage should be fairly minimal to just do a simple `fork()` and wait. However, I do think it makes the job of this little DNS manager process much clearer to move it into its own executable, so I went and implemented it as a new command.

src/connectionmanager.cpp
oskirby: When the daemon crashes, the controller attempts to re-initialize it, which will trigger the asynchronous emission of another `ControllerImpl::initialized()` signal from any state, therefore this assert is no longer valid.

For example, if the daemon crashed while in `ConnectionManager::StateOn` we will wind up here when re-starting the daemon.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -219,5 +220,93 @@ void DnsUtilsMacos::backupService(const QString& uuid) {
...
+// Fork a child process to handle restoration of the DNS configuration.
+// This ensures that we always cleanup after ourselves, even if the
+// daemon happens to crash.
+bool DnsUtilsMacos::takeDnsSnapshot(void) {
+  // Get the list of current network services.
+  CFArrayRef netServices = SCDynamicStoreCopyKeyList(
+      m_scStore, CFSTR("Setup:/Network/Service/[0-9A-F-]+"));
+  if (netServices == nullptr) {
+    return false;
+  }
```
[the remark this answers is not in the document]
oskirby: Done!

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -219,5 +220,93 @@ void DnsUtilsMacos::backupService(const QString& uuid) {
...
   }
 
-  m_prevServices[uuid] = backup;
+  return backup;
+}
+
+// Fork a child process to handle restoration of the DNS configuration.
+// This ensures that we always cleanup after ourselves, even if the
+// daemon happens to crash.
+bool DnsUtilsMacos::takeDnsSnapshot(void) {
```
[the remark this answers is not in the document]
oskirby: Done!

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -219,5 +220,93 @@ void DnsUtilsMacos::backupService(const QString& uuid) {
...
+  // Get the list of current network services.
+  CFArrayRef netServices = SCDynamicStoreCopyKeyList(
+      m_scStore, CFSTR("Setup:/Network/Service/[0-9A-F-]+"));
```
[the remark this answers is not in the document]
oskirby: I couldn't find a schema for the format of this string, but it appears to be a UUID. I'll expand the regex to permit lowercase hexadecimal too.

src/localsocketcontroller.cpp
oskirby: When the daemon crashes, writing to the socket would expect to throw a `SIGPIPE` because there is no longer anything on the reading end of the pipe, so I think that it cannot hang in the event of a crash. If the daemon becomes unresponsive then this might result in a hang, so I guess `QLocalSocket::abort()` is probably the better function to call here.

src/localsocketcontroller.cpp
oskirby: I don't think so, but it was lazy of me to re-use `CONNECTION_RETRY_INITIAL_MSEC` here. I will add a separate constant for the response timeout.

src/localsocketcontroller.cpp
oskirby: This assumes that the daemon will respond before the next write of another synchronous command (eg: `status` or `logs`). Asynchronous commands such as `connect` don't expect a response because the `Controller` handles that separately, or the daemon may not even respond at all if the tunnel is unable to establish a handshake.

In the event that such a race condition occurs, the last `write()` call with an `expected` value takes precedence when checking for responses, in which case we will still be able to detect a crashed daemon.

oskirby: `LocalSocketController::getBackendLogs()` is the only other case where we expect a synchronous response from the daemon, but it's not clear how long the daemon may take to generate that response if the log file is significant in size, so I decided to ignore the problem for the time being.

I suppose I could add another optional parameter to specify a timeout, in which case we could allow a little more time for the `getBackendLogs` command specifically.

src/localsocketcontroller.cpp
```diff
@@ -13,6 +13,7 @@
 #include <QJsonValue>
 #include <QStandardPaths>
 
+#include "constants.h"
```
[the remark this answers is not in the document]
oskirby: It does, thank you for catching this.

src/localsocketcontroller.h
```diff
@@ -45,7 +45,7 @@ class LocalSocketController final : public ControllerImpl {
   void readData();
   void parseCommand(const QByteArray& command);
 
-  void write(const QJsonObject& json);
+  void write(const QJsonObject& json, const QString& expect = QString());
```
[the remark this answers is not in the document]
oskirby: The last message with an `expect` parameter is the one that takes precedence when driving this timeout mechanism, which should still work when trying to detect a crashed daemon, though it might be a little sloppier.

src/localsocketcontroller.cpp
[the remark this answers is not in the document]
oskirby: I believe that it is, but switching this out to `QLocalSocket::abort()` should make its effect even clearer and more robust.

nebula/ui/components/MZSystemAlert.qml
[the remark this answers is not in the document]
oskirby: `MozillaVPN::backendServiceRestore()` is still reachable through a few other code paths when handling other errors, but there is a reasonable case to remove it, because none of it has ever done anything useful.

src/localsocketcontroller.cpp
oskirby: Done!

src/localsocketcontroller.cpp
oskirby: Done!

oskirby:
> For now, is it possible to write test for the new behaviour?

Unfortunately, it is quite difficult to unit test the controller code as it interacts with the daemon and OS. A mock daemon would be a nice tool to have in order to simulate these kinds of failures, but that would be a large-ish project to write, but one that could enable us to write functional tests for these changes.

src/localsocketcontroller.cpp
oskirby: Done!

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: This should now be addressed by using a `QProcess` class, which waits for at most 30s when trying to terminate a child process.

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: In the multi-hop scenario, `updateResolvers()` is only called when setting up the exit hop, so it should only ever be called once when trying to activate the tunnel.

oskirby: The check for this is done in `Daemon::maybeUpdateResolvers()`

nebula/ui/components/MZSystemAlert.qml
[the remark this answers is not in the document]
oskirby: It has been removed.

oskirby: I have opted to remove the method.

src/localsocketcontroller.h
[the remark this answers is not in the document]
brizental: II wonder if this is an issue. Say the next to last reply is the one that hangs? ... I guess if the next to last reply hangs, but the last one doesn't that means the daemon is doing fine and just maybe had a hiccup. Hm. That deserves a comment in code though I think.

src/localsocketcontroller.cpp
oskirby: I have updated the response handling so that we can now handle multiple outstanding responses.

src/localsocketcontroller.h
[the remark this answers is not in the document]
oskirby: I have updated the response handling so that we can now handle multiple outstanding responses. It wasn't that hard and it's probably the better way to handle things.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -219,5 +220,93 @@ void DnsUtilsMacos::backupService(const QString& uuid) {
...
+  signal(SIGHUP, [](int s) { Q_UNUSED(s); });
+  if (sigwait(&set, &sig) != 0) {
+    // Something went horribly wrong.
+    exit(errno);
+  }
+  // If we receive a SIGHUP, then exit without restoring anything.
+  if (sig == SIGHUP) {
+    exit(0);
+  }
+
```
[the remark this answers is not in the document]
oskirby: No, I think it's important that we make every effort to fall-through to the cleanup code, and an `assert()` will terminate the process without cleaning up the DNS resolvers. We should just handle unexpected signals like a termination request.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -150,41 +156,36 @@ bool DnsUtilsMacos::updateResolvers(const QString& ifname,
...
+    return false;
   }
 
-  m_prevServices.clear();
+  // Wait for the process to exit, and gather it's status.
+  int status = 0;
+  if (waitpid(m_dnsSnapshotPid, &status, 0) < 0) {
+    logger.debug() << "Failed to receive child status";
+  } else if (!WIFEXITED(status)) {
+    logger.error() << "Failed to terminate DNS restorer";
```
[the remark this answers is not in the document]
oskirby: This is now handled by using `QProcess:waitForFinished()`

oskirby:
> > For now, is it possible to write test for the new behaviour?
> 
> Unfortunately, it is quite difficult to unit test the controller code as it interacts with the daemon and OS. A mock daemon would be a nice tool to have in order to simulate these kinds of failures, but that would be a large-ish project to write, but one that could enable us to write functional tests for these changes.

The more I think about this, the more I am starting to like it. The existence of a mock daemon would let us run functional tests against the production binary instead of needing a separate `dummyvpn` build by stubbing out all of the privileged functionality. I might just go and have a play with this concept in another PR.

src/localsocketcontroller.h
oskirby: Done!

src/localsocketcontroller.cpp
```diff
@@ -4,29 +4,21 @@
...
-constexpr int MAX_CONNECTION_RETRY = 10;
-
-// How long do we wait between one try and the next one.
-constexpr int CONNECTION_RETRY_TIMER_MSEC = 500;
+// When the daemon is unreachable, we will retry indefinitely using an
+// exponential backoff algorithm. The interval between retries starts at
+// the initial value, and doubles after each failed connection attempt,
+// with time between each retry clamped to the maximum value.
+constexpr int CONNECTION_RETRY_INITIAL_MSEC = 500;
+constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000;
```
brizental, line 21: nit: Please add comments with human readable times here, e.g.
```suggestion
constexpr int CONNECTION_RETRY_MAXIMUM_MSEC = 16000; // 16s
```

brizental: Also, I just realized this is 16s. Isn't that too little for an infinite retry loop?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +341,48 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+void LocalSocketController::disarmTimeout(const QString& responseType) {
+  for (QTimer* t : m_expectedResponses) {
+    QVariant qv = t->property("responseType");
+    if (qv.type() != QVariant::String) {
+      continue;
+    }
+    if (qv.toString() != responseType) {
+      continue;
+    }
+    m_expectedResponses.removeOne(t);
```
brizental, line 378: Consider extracting the queue mechanism into a testable unit.

src/mozillavpn.cpp
```diff
@@ -1165,11 +1165,6 @@ void MozillaVPN::controllerStateChanged() {
   NetworkManager::instance()->clearCache();
 }
 
-void MozillaVPN::backendServiceRestore() {
-  logger.debug() << "Background service restore request";
-  // TODO
-}
```
brizental, line 1171: Thank you for removing this. We need to notify QA and the team about this change somehow. Maybe we can create a separate 0 pointer for it just for tracking purposes?

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,266 @@
...
+                                                   qUtf8Printable(rstring),
+                                                   kCFStringEncodingUTF8);
+    CFArrayAppendValue(array, cfAddr);
+    CFRelease(cfAddr);
+  }
+  CFDictionarySetValue(dict, name, array);
+  CFRelease(array);
+}
+
+int MacOSDnsManager::run(QStringList& tokens) {
```
brizental, line 94: nit: It's a bit confusing not having this as the last function of the file.

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,266 @@
...
+    if (host.isNull()) {
+      stream << "Error: DNS Server \'" << addr << "\' is invalid." << Qt::endl;
+      return 1;
+    }
+    if (host.isBroadcast() || host.isMulticast()) {
+      stream << "Error: DNS Server \'" << addr << "\' is a multicast address."
+             << Qt::endl;
+      return 1;
+    }
+    // TODO: Should we prohibit the loopback address?
```
brizental, line 124: For my understanding, why would be the pros and cons of prohibiting that?

src/platforms/macos/daemon/macosdnsmanager.h
```diff
@@ -0,0 +1,33 @@
...
+
+ private:
+  QStringList enumerateNetServices();
+  static QString scErrorMessage();
+  void restoreSnapshot();
+  void clearSnapshot();
+  int waitForTermination();
+
+ private:
+  SCDynamicStoreRef m_scStore = nullptr;
```
brizental, line 29: Would it be possible to test the functionality of this class by mocking the value of this property? IIUC that is main outside dep?

brizental:
Question: Why does ths PR title say MacOS only? Doesn't this partially also apply to Windows given it also uses LocalSocketController?

src/localsocketcontroller.cpp
oskirby: This means that if the daemon never starts, the steady-state behaviour is for the client to retry every 16 seconds indefinitely.

brizental: Yeah I know, but still. Isn't that too small an interval? My experience with backoff intervals is all related to network requests though, so this is a genuine question lol in those cases it would be too small an interval.

oskirby:
> Question: Why does ths PR title say MacOS only? Doesn't this partially also apply to Windows given it also uses LocalSocketController?

I haven't thoroughly tested how the Windows daemon handles a crash situation while the tunnel is active, so I kind of suspect that there may be additional work to finish up that platform with regards to things this epic. But yes, because the `LocalSocketController` is shared by Windows too, Windows will gain the ability for the client to re-connect.

brizental:
> I haven't thoroughly tested how the Windows daemon handles a crash situation while the tunnel is active, so I kind of suspect that there may be additional work to finish up that platform with regards to things this epic. But yes, because the LocalSocketController is shared by Windows too, Windows will gain the ability for the client to re-connect.

👍 One last note and this can totally go in a follow up, but I'd suggest adding a "force daemon crash" option to the dev menu for QA to be able to test this.

oskirby:
> > I haven't thoroughly tested how the Windows daemon handles a crash situation while the tunnel is active, so I kind of suspect that there may be additional work to finish up that platform with regards to things this epic. But yes, because the LocalSocketController is shared by Windows too, Windows will gain the ability for the client to re-connect.
> 
> 👍
> 
> One last note and this can totally go in a follow up, but I'd suggest adding a "force daemon crash" option to the dev menu / inspector for QA to be able to test this.

That is not really possible. Root permissions are required in order to force a crash of the daemon, which the client does not possess. I think maybe the mock daemon is going to be required after all in order for any of this to be testable.

brizental:
> That is not really possible. Root permissions are required in order to force a crash of the daemon, which the client does not possess. I think maybe the mock daemon is going to be required after all in order for any of this to be testable.

Is there a mechanism to ask for a password on demand when running this specific dev task?

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: Allowing `localhost` allows users to run a DNS server or proxy on their local machine to intercept and manipulate DNS (eg: something like a Pi-Hole), effectively rolling their own solution for DNS-based ad or tracker blocking. However, we have seen some platforms (eg: iOS and Android) explicitly prohibit this (see: `DNSHelper::validateUserDNS()`).

However, given that this is a MacOS-specific tool, perhaps it is fine to permit the loopback address after all.

src/localsocketcontroller.cpp
oskirby: I don't think the interval is too small here. The local socket is a connection made entirely within the user's computer so the latency and resources needed to check if the daemon is active are quite small. Typically exponential backoffs are used when many clients are competing for a finite resource (either network bandwidth or time on a server) so that failures don't further compound problems of network congestion, but those concerns don't really apply here because the VPN client/daemon is a one-to-one relationship.

I kind of phrase it like, is it reasonable to expect the daemon to re-start itself and be ready to accept commands from the client within 16 seconds or so? And that feels like a yes.

src/localsocketcontroller.h
```diff
@@ -5,13 +5,17 @@
...
 #include <QLocalSocket>
 #include <QTimer>
 #include <functional>
 
 #include "controllerimpl.h"
 
+// For messages that are expected to generate a synchronous response, this
+// defines the default time that we will wait before assuming an error in
+// the daemon has occured.
+constexpr int CONNECTION_RESPONSE_TIMEOUT_MSEC = 500;
```
vinocher, line 17: Consider moving this into the class to avoid affecting the global name space.

src/localsocketcontroller.h
```diff
@@ -44,8 +50,19 @@ class LocalSocketController final : public ControllerImpl {
...
   void readData();
   void parseCommand(const QByteArray& command);
-
-  void write(const QJsonObject& json);
+  void disarmTimeout(const QString& responseType);
+  void clearAllTimeouts();
+
+  /**
+   * @brief Write a JSON message to the socket, sending it to the daemon.
+   *
```
vinocher, line 58: Are repsonseType or expectedResponseType better names? And should json be renamed to something like message or request or command?

src/localsocketcontroller.h
```diff
@@ -62,7 +79,11 @@ class LocalSocketController final : public ControllerImpl {
   std::function<void(const QString&)> m_logCallback = nullptr;
 
   QTimer m_initializingTimer;
-  uint32_t m_initializingRetry = 0;
+  uint32_t m_initializingInterval = 0;
+
+  // When a message to the daemon expects an immediate response, these
```
vinocher, line 84: Are names like m_responseTimeoutTimers or m_responseTimeouts better? The current name seems to indicate that it is a list of responses.

src/localsocketcontroller.cpp
```diff
@@ -343,9 +353,72 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+  // is used to detect a crash or failure of the daemon.
+  if (!expect.isEmpty()) {
+    QTimer* t = new QTimer(this);
+    connect(t, &QTimer::timeout, this,
+            [&] { this->errorOccurred(QLocalSocket::SocketTimeoutError); });
+
+    t->setProperty("responseType", QVariant(expect));
+    t->setSingleShot(true);
+    t->start(timeout);
+
```
vinocher, line 376: Consider if clearTimeout is a better name because it matches clearAllTimeouts.

src/localsocketcontroller.cpp
```diff
@@ -343,9 +353,72 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+  }
+
   Q_ASSERT(m_socket);
-  m_socket->write(QJsonDocument(json).toJson(QJsonDocument::Compact));
-  m_socket->write("\n");
+  m_socket->write(payload);
   m_socket->flush();
 }
+
+void LocalSocketController::disarmTimeout(const QString& responseType) {
```
vinocher, line 385: Multiple requests with the same expected response type are  assumed to receive responsese in the same order. That may be an acceptable assumption. Could a comment be added explaining that?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +353,72 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+  for (QTimer* t : m_expectedResponses) {
+    QVariant qv = t->property("responseType");
+    if (qv.type() != QVariant::String) {
+      continue;
+    }
```
vinocher, line 390: Nit: Remove void

src/localsocketcontroller.cpp
```diff
@@ -343,9 +353,72 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
   logger.warning() << "Invalid command received:" << command;
 }
 
-void LocalSocketController::write(const QJsonObject& json) {
+void LocalSocketController::write(const QJsonObject& json,
+                                  const QString& expect, int timeout) {
+  QByteArray payload = QJsonDocument(json).toJson(QJsonDocument::Compact);
```
vinocher, line 362: Should we ensure that code like DaemonAccessControl that validates commands should write empty responses for disallowed commands? Otherwise we will wrongly detect those cases as a disconnected daemon.

src/localsocketcontroller.cpp
```diff
@@ -45,37 +53,34 @@ LocalSocketController::LocalSocketController() {
...
           &LocalSocketController::readData);
 
+  m_initializingInterval = CONNECTION_RETRY_INITIAL_MSEC;
   m_initializingTimer.setSingleShot(true);
   connect(&m_initializingTimer, &QTimer::timeout, this,
           &LocalSocketController::initializeInternal);
 }
 
 LocalSocketController::~LocalSocketController() {
   MZ_COUNT_DTOR(LocalSocketController);
```
vinocher, line 63: Would clearAllTimeouts put us in an incorrect state if the daemon was delayed and then starts responding to previous responses?

src/platforms/macos/daemon/dnsutilsmacos.h
```diff
@@ -25,27 +22,11 @@ class DnsUtilsMacos final : public DnsUtils {
...
-
-    QString m_domain;
-    QStringList m_search;
-    QStringList m_servers;
-    QStringList m_sortlist;
-  };
-
-  SCDynamicStoreRef m_scStore = nullptr;
-  QMap<QString, DnsBackup> m_prevServices;
+  QProcess m_dnsManager;
```
vinocher, line 29: This could use a comment that it will be used to create a watch dog process for DNS cleanup.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -19,205 +17,52 @@ Logger logger("DnsUtilsMacos");
...
-                                                   kCFStringEncodingUTF8);
-    CFArrayAppendValue(array, cfAddr);
-    CFRelease(cfAddr);
+void DnsUtilsMacos::dnsManagerStdoutReady() {
+  for (;;) {
+    QByteArray line = m_dnsManager.readLine();
+    if (line.length() <= 0) {
+      break;
+    }
+    logger.debug() << QString::fromUtf8(line);
```
vinocher, line 41: Should this add a string like "DnsManagerStdout" so it is easier to find it in a log?

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -19,205 +17,52 @@ Logger logger("DnsUtilsMacos");
...
-    value = CFDictionaryGetValue(config, kSCPropNetDNSSortList);
-    if (value) {
-      backup.m_sortlist = cfParseStringList(value);
-    }
+  if (m_dnsManager.state() == QProcess::NotRunning) {
+    return true;
   }
 
-  m_prevServices[uuid] = backup;
+  m_dnsManager.terminate();
```
vinocher, line 66: updateResolvers and restoreResolvers could use some comments about how the dnsManager in a different process is being used for their operations.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -19,205 +17,52 @@ Logger logger("DnsUtilsMacos");
...
+  return m_dnsManager.waitForFinished();
```
vinocher, line 67: Should an explicit timeout be used here or is the default OK? Looks like the default is 30s, which may be too long.

src/platforms/macos/daemon/dnsutilsmacos.cpp
```diff
@@ -19,205 +17,52 @@ Logger logger("DnsUtilsMacos");
...
-    if (!dnsPath) {
-      continue;
-    }
-    SCDynamicStoreSetValue(m_scStore, dnsPath, dnsConfig);
-    CFRelease(dnsPath);
-  }
-  return true;
+  // Launch the DNS manager process.
+  m_dnsManager.start(qApp->applicationFilePath(), args);
+  return m_dnsManager.waitForStarted();
```
vinocher, line 58: Is waitForStarted's default timeout OK?

src/localsocketcontroller.cpp
oskirby: Without modifying the API, we have no way of distinguishing multiple responses of the same type, so this is the best we can do without risking a compatibility issue.

src/platforms/macos/daemon/macosdnsmanager.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef MACOSDNSMANAGER_H
+#define MACOSDNSMANAGER_H
+
+#include <systemconfiguration/scdynamicstore.h>
+
+#include <QMap>
+
+#include "command.h"
+
+class MacOSDnsManager final : public Command {
```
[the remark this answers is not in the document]
vinocher: Also, the naming is a bit confusing because it is a command and a DnsManager. Should those be separated (separation of concerns), by introducing a DnsManagerCommand which will handle the command line args and create the MacOSDnsManager?

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: The timeout here would only catch the case where the `macosdnsmanager` command hangs or something goes horribly wrong, so I think the default is fine here.

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,269 @@
...
+      return 1;
+    }
+  }
+
+  // Block this process and wait for a termination signal.
+  stream << "Waiting for termination signal." << Qt::endl;
+  int sig = waitForTermination();
+  stream << "Caught signal " << sig << ". Preparing to exit." << Qt::endl;
+
+  // If we exit on SIGHUP, leave the DNS configuration in-place.
```
vinocher, line 110: Could this comment or the a comment in DnsUtilsMacos::restoreResolvers indicate that   m_dnsManager.terminate() causes SIGHUP?

oskirby: Will do, but technically this sends a `SIGTERM`

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: I believe so, normally it should only take a few milliseconds to launch a new process, so if we encounter the 30s default timeout here something has gone horribly wrong.

src/localsocketcontroller.cpp
oskirby: There shouldn't be anything in the daemon that blocks for any significant amount of time (everything should be handled async), so it would be an error if the daemon were to block long enough to trigger one of these timeouts.

Also, `clearAllTimeouts()` is used here under two conditions:
1. To cleanup and garbage collect when running the destructor, and
2. When handling an `errorOccurred()` signal, and we are about to try and reconnect. We don't want timeouts from the previous connection to interfere with the re connection attempt.

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: There's nothing in the parent process that will send a `SIGHUP` I mostly just used it for testing. However, this signal in POSIX daemons is typically used to inform a daemon that its configuration has changed and may need to be updated and/or re-start itself.

I was kind of envisioning being able to use a `SIGHUP` to tell the DNS manager to change a DNS server without deactivating the VPN, but I never quite got that far.

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: Done!

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: Done!

src/localsocketcontroller.cpp
oskirby: Done!

src/localsocketcontroller.h
oskirby: I have updated the names accordingly.

src/localsocketcontroller.h
oskirby: Done!

src/localsocketcontroller.h
oskirby: Sure, I will rename it.

src/platforms/macos/daemon/dnsutilsmacos.h
oskirby: Sure, I will rename it.

src/platforms/macos/daemon/macosdnsmanager.h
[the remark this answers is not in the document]
oskirby: As for the naming, that was following the precedent set by the `MacOSDaemonServer` which also breaks from our typical `CommandFooBar` class naming convention.

src/platforms/macos/daemon/dnsutilsmacos.cpp
vinocher: A timeout here may need be handled by requesting process termination of the child process, so it can cleanup.

src/platforms/macos/daemon/macosdnsmanager.cpp
vinocher: Would be useful to the reader if that was added as a comment. Or consider removing that code path if necessary.

src/localsocketcontroller.cpp
vinocher: Sounds good. Perhaps add a comment with the assumption?

src/platforms/macos/daemon/macosdnsmanager.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <systemconfiguration/scdynamicstore.h>
+
+#include <QMap>
+
+#include "command.h"
+
+// This class adds a command to the Mozilla VPN that overwrites the system
+// DNS configuration, and restores it upon program exit. This provides a
+// watchdog mechansim to ensure that no matter how the daemon termanates
```
vinocher, line 16: Some typos here. Also, could this say this it runs in a different process?

src/localsocketcontroller.cpp
```diff
@@ -343,9 +357,104 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
+  // Execute 'kill' to terminate the daemon as though it crashed.
+  logger.warning() << "Sending SIGSEGV to:" << pid;
+  QByteArray pidString = QString::number(pid).toUtf8();
+  char killpath[] = "/bin/kill";
+  char killsignal[] = "-SEGV";
+  char* const killargs[] = {killsignal, pidString.data(), nullptr};
+
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wdeprecated-declarations"
+  status = AuthorizationExecuteWithPrivileges(
```
Gela, line 452: Looks like [AuthorizationExecuteWithPrivileges](https://developer.apple.com/documentation/security/1540038-authorizationexecutewithprivileg) is deprecated. Are there security concerns with using this?

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if (m_scStore != nullptr) {
+    restoreSnapshot();
+    CFRelease(m_scStore);
+  }
+  MZ_COUNT_DTOR(MacOSDnsManager);
+}
+
+int MacOSDnsManager::run(QStringList& tokens) {
+  Q_ASSERT(!tokens.isEmpty());
+  QString appName = tokens[0];
```
Gela, line 38:
```suggestion
  const QString appName = tokens[0];
```

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  if (tokens.length() <= 1) {
+    stream << "Error: At least one DNS server address is required." << Qt::endl;
+    stream << Qt::endl;
+    stream << "Usage: [options] ADDR [ADDR ...]" << Qt::endl;
+    return 1;
+  }
+
+  // Ensure that the DNS addresses we received are valid.
+  QStringList dnsAddrList = tokens.mid(1);
```
Gela, line 54:
```suggestion
  const QStringList dnsAddrList = tokens.mid(1);
```

src/platforms/macos/daemon/macosdnsmanager.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+    if (host.isBroadcast() || host.isMulticast()) {
+      stream << "Error: DNS Server \'" << addr << "\' is a multicast address."
+             << Qt::endl;
+      return 1;
+    }
+  }
+
+  // Prepare the updated DNS configuration.
+  CFMutableDictionaryRef dnsConfig = CFDictionaryCreateMutable(
```
Gela, line 69: const maybe? 
```suggestion
  const CFMutableDictionaryRef dnsConfig = CFDictionaryCreateMutable(
```

src/localsocketcontroller.cpp
oskirby: I think Apple's intention for deprecating this API is that they want to encourage developers to split privileged functionality out of their UI processes and move it into daemons instead, which is exactly how we have structured the VPN on MacOS. However, an IPC method to tell the daemon to crash itself seems even worse from a security perspective.

I think part of the concern with using this API is that it just trains users to accept a security login prompt without thinking, and doesn't do anything to describe (or restrict) what applications can do once they are authorized.

However, for the purposes of a testing command which can only be triggered from the inspector, I think this is fine.

oskirby: There is also the danger that at some point in the future Apple may remove this function entirely, in which case we may need to revisit how to test this feature. But lets cross that bridge when we get there.

src/platforms/macos/daemon/macosdnsmanager.h
oskirby: The forking and separation of processes is actually handled by the `QProcess` class in `DnsUtilsMacos`, so I think it would be a bit misleading to describe it here.

src/platforms/macos/daemon/dnsutilsmacos.cpp
oskirby: Done!

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: I have added a bit more to the comment, hopefully it helps.

src/controllerimpl.h
```diff
@@ -51,6 +51,10 @@ class ControllerImpl : public QObject {
   // tunnel cannot be reactivated in system settings.
   virtual void deleteOSTunnelConfig(){};
 
+  // This method attempts to force the daemon to crash, and is used for
+  // testing the ability of the VPN to recover from a backend error.
+  virtual void testDaemonCrash() {}
```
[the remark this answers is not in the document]
oskirby: Done

src/localsocketcontroller.cpp
```diff
@@ -4,29 +4,36 @@
...
+#  include <QScopeGuard>
+#endif
+
+// When the daemon is unreachable, we will retry indefinitely using an
+// exponential backoff algorithm. The interval between retries starts at
+// the initial value, and doubles after each failed connection attempt,
+// with time between each retry clamped to the maximum value.
+//
+// This means that if the daemon never starts, the steady-state behaviour will
+// have client to retry every 16 seconds indefinitely.
```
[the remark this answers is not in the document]
oskirby: I have refined the comment a bit, hopefully I got the grammar right this time.

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: Done

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: Done

src/platforms/macos/daemon/macosdnsmanager.cpp
oskirby: Done

src/localsocketcontroller.cpp
```diff
@@ -343,9 +357,104 @@ void LocalSocketController::parseCommand(const QByteArray& command) {
...
   Q_ASSERT(m_socket);
-  m_socket->write(QJsonDocument(json).toJson(QJsonDocument::Compact));
-  m_socket->write("\n");
+  m_socket->write(payload);
   m_socket->flush();
 }
+
+void LocalSocketController::clearTimeout(const QString& responseType) {
+  for (QTimer* t : m_responseTimeouts) {
+    QVariant qv = t->property("responseType");
```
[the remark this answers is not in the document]
oskirby: Done

src/localsocketcontroller.cpp
oskirby: That is an excellent point. The API, as designed, doesn't really have any mechanism to return an error in this scenario and that feels like a bigger oversight than I really intended to tackle in this PR. I think the behaviour has always been that when access is denied it manifests as a disconnected daemon, so nothing has really changed here.

There seems like a design question here of, how should the client behave when it is denied access?

oskirby: In particular, how to respond to the `activate()` method when access is denied seem to be more important. Without a response the client will behave as though the server is unreachable.

vinocher: Yes, it is an existing issue and I don't think anything needs to be done in this PR.

vinocher (approved):
Looks good!

Gela (approved):

PR #8570, opened 2023-11-16
VPN-3362: Add plan upgrade message for monthly users

MattLichtenstein:
## Description

- Add new in-app messages for non-IAP subscribers on a monthly plan that become enabled on day 14 and resurfaced on day 87 of the subscription period

- Add `setTimedCallback` to the addon api to support re-evaluating javascript conditions on a timer
  - This is useful because a user who has the app launched for long periods of time will now have this message enabled at the correct time, instead of waiting for a relaunch or change in `subscriptionData` to re-evaluate the conditions.

- Add `set_subscription_start_date` inspector command to dynamically change the `createdAt` timestamp (aka start date) of the subscription for easier testing
  - This only updates the timestamp locally, so it will be overwritten every time `taskGetSubscriptionDetails` re-polls guardian

- Increase the `periodicTaskTimer` interval to fire every 5 minutes instead of every 30 seconds on the staging environment
  - This timer triggers a number of tasks, including `taskGetSubscriptionDetails`, which mad testing cumbersome

- Add a new `Subscription` badge type for in-app messages

- Add `notify` to the messages schema which determines whether a system notification should be fired for this in-app message. Default value is `true`

- Fix a bug in `subscriptiondata.cpp` where we were caching device settings instead of subscription data

Other notes: 
- The timestamp for these messages is based off of the subscription start date, not when the message was received or enabled. The timestamp for these messages will always be 14 days and 87 days respectively after the subscription start date. 
  - This will be a little janky for users who are already past this threshold as they will now have a message from a date in the past (eg once this message is deployed, a user might have a message from 10 days ago, even though it wasn't there until now)

## Reference

[VPN-3362: Release a "Plan Upgrade" in-app message to monthly-plan non-IAP subscribers](https://mozilla-hub.atlassian.net/browse/VPN-3362)
[VPN-3044: Release support for In-App Messages to generate (or not) an OS notification](https://mozilla-hub.atlassian.net/browse/VPN-3044)


flodolo (commented):
Questions:
* Why do we need two addons that are identical in content? We're asking translators to translate these strings twice.
* Can we have better names than "14d" and "87d", given these will show up in Pontoon? Does "d" stand for "days"?

scripts/ci/jsonSchemas/message.json
```diff
@@ -32,6 +32,10 @@
       "type": "string",
       "description": "The type of this message"
     },
+    "notify": {
+      "type": "boolean",
+      "description": "Determines whether we trigger a system notification. Default: true"
+    },
```
brizental, line 38: This addresses https://mozilla-hub.atlassian.net/browse/VPN-3044. Please make sure to assign yourself :100:

src/addons/addonapi.cpp
```diff
@@ -92,6 +92,32 @@ void AddonApi::initialize() {
...
+          [callback]() { callback.call(); });
+
+  timer->start();
+}
+
+void AddonApi::clearTimers() {
+  // Clear all timers for a specific addon
+  QString addonId = m_addon->id();
+  if (m_timers.contains(addonId)) {
+    m_timers.remove(addonId);
```
brizental, line 117: Should this also stop the timer or will the removal do that already?

addons/message_plan_upgrade_14d/conditions.js
```diff
@@ -0,0 +1,40 @@
+(function(api, condition) {  
+  //Show this message after 14 days, but not after 87 days (there is a separate message for that)
+
+  //There are 2 identical (content-wise), yet separate messages because we want to give the illusion that we are 
+  //re-showing the same message 87 days in (by hiding the 14 day one and showing the 87 day one) without having to
+  //un-read or un-dismiss a message
```
brizental, line 6: Why would un-reading / un-dismissing be bad here?

src/addons/addonmessage.cpp
```diff
@@ -61,6 +61,9 @@ Addon* AddonMessage::create(QObject* parent, const QString& manifestFileName,
   message->m_date = messageObj["date"].toInteger();
   message->planDateRetranslation();
 
+  // if "notify" is not specified in the manifest, default to true
+  message->m_shouldNotify = messageObj["notify"].toBool(true);
```
brizental, line 65: Given this is a new API and the previous behavior used to be not showin gnotifications, should the default here be false?

src/constants.h
```diff
@@ -172,7 +172,7 @@ CONSTEXPR(uint32_t, releaseMonitorMsec, 21600000, 4000, 0)
 
 // in milliseconds, how often we should fetch the server list, the account and
 // so on.
-CONSTEXPR(uint32_t, schedulePeriodicTaskTimerMsec, 3600000, 30000, 0)
+CONSTEXPR(uint32_t, schedulePeriodicTaskTimerMsec, 3600000, 300000, 0)
```
brizental, line 175: Just because arithmetic is hard, please add a comment here with this time in hours or minutes.

brizental (changes requested):
Please add functional tests for the new addon enabling / disabling logic.

src/addons/addonmessage.cpp
MattLichtenstein: the previous behavior was to show notifications

scripts/ci/jsonSchemas/message.json
MattLichtenstein: two for one! thanks

addons/message_plan_upgrade_14d/conditions.js
MattLichtenstein: I remember investigating this route, but I don't remember why it wasn't working. This was before we had the discussion about timers. Let me reinvestigate

src/constants.h
vinocher: Is 5 minutes too long in debug for testing other scenarios? The PR description says that the previous 30 seconds made testing difficult. Could you please clarify why?

src/translations/strings.yaml
```diff
@@ -899,6 +899,9 @@ inAppMessaging:
   surveyBadge:
     value: Survey
     comment: A badge shown in a message signifying that this message contains a survey
+  subscriptionBadge:
+    value: Subscription
+    comment: A badge shown in a message signifying that this message contains information about the users subscription
```
vinocher, line 904: Nit: Typo "user's"

src/models/subscriptiondata.cpp
```diff
@@ -50,7 +50,7 @@ bool SubscriptionData::fromSettings() {
 
   logger.debug() << "Reading the subscription data from settings";
 
-  const QByteArray& json = settingsHolder->devices();
+  const QByteArray& json = settingsHolder->subscriptionData();
```
vinocher, line 53: Nice catch! Do you know what the symptoms of this bug was?

src/inspector/inspectorhandler.cpp
```diff
@@ -636,7 +638,38 @@ static QList<InspectorCommand> s_commands{
...
+          qint64 newCreatedAtTimestamp = arguments[1].toLongLong();
+
+          // get sub data json from settings
+          QByteArray subscriptionData =
+              SettingsHolder::instance()->subscriptionData();
+          QJsonDocument doc = QJsonDocument::fromJson(subscriptionData);
+          QJsonObject obj = doc.object();
+          QJsonObject subObj = obj["subscription"].toObject();
+          qlonglong createdAt =
+              subObj["created"].toVariant().toLongLong() * 1000;
```
vinocher, line 656: Why is this line needed, since createdAt is overwriten in line 659?

docs/messages.md
```diff
@@ -22,5 +22,6 @@ property object with the following properties:
 | title_comment | An optional comment to describe the meaning of the title | String | No |
 | subtitle_comment | An optional comment to describe the meaning of the subtitle | String | No |
 | date | The date the message was received (using seconds since epoch time) | Number | No |
-| badge | A label used to tag a message (options: `warning`, `critical`, `new_update`, `whats_new`, `survey` ) | String | No |
+| badge | A label used to tag a message (options: `warning`, `critical`, `new_update`, `whats_new`, `survey`, `subscription` ) | String | No |
+| notify | Should we notify the user about this message via system notifications | Boolean | No |
```
vinocher, line 26: SHould this indicate that default is to notify?

src/addons/addonapi.h
```diff
@@ -27,6 +28,9 @@ class AddonApi final : public QQmlPropertyMap {
                                  const QJSValue& callback);
   Q_INVOKABLE void log(const QString& message);
 
+  Q_INVOKABLE void setTimeout(int interval, const QJSValue& callback);
```
vinocher, line 31: Consider if a name like scheduleCallback or setTimedCallback are better names, and change clearTimers accordingly.

vinocher:
The PR description says:
> clearTimers should always be called before setTimeout to avoid one addon having multiple timers running (in it's current implementation)

While this may apply to these addons, isn't it conceivable that an addon may want multiple timers?

addons/message_plan_upgrade_87d copy/conditions.js
```diff
@@ -0,0 +1,34 @@
+(function(api, condition) {  
+  //Show this message after 87 days
+
+  //There are 2 identical (content-wise), yet separate messages because we want to give the illusion that we are 
+  //re-showing the same message 87 days in (by hiding the 14 day one and showing the 87 day one) without having to
+  //un-read or un-dismiss a message
+
+  api.connectSignal(api.subscriptionData, 'changed', () => computeCondition());
+
+  //Slight delay to as we wait for subscriptionData to initialize
```
vinocher, line 10: Typo "to as"

addons/message_plan_upgrade_87d copy/conditions.js
```diff
@@ -0,0 +1,34 @@
...
+  api.setTimeout(5000, () => computeCondition());
```
vinocher, line 11: Is this considered reliable to ensure that TaskGetSubscriptionDetails has run? Or should that be explicitly checked?

src/models/subscriptiondata.cpp
MattLichtenstein: Whenever we run `TaskGetSubscriptionDetails`, we compare the incoming json with what we have stored in settings before parsing it. Here we were always comparing new subscription data json with device settings json so they were always different and we would always have to re-parse (emitting `subscriptionDataChanged`). Don't know of any bugs this caused other than wasteful signal emission/binding reevaluations

src/inspector/inspectorhandler.cpp
MattLichtenstein: it's not 🙂

MattLichtenstein:
> While this may apply to these addons, isn't it conceivable that an addon may want multiple timers?

Sure. my only reasoning was to avoid overengineering, but if you think that is a likely scenario than we can support that

addons/message_plan_upgrade_87d copy/conditions.js
MattLichtenstein: I would say no, and I am looking for a better solution. We aren't really waiting on `TaskGetSubscriptionDetails` (if that task runs, and we have new subscription data, the `connectSignal` above will catch it), we're waiting on the subscriptionData from settings to be loaded into memory

MattLichtenstein: I think adding `emit changed();` [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/e1c86b0063dbe83dbbc5b809a03fcd3eee8d14be/src/models/subscriptiondata.cpp#L59) could resolve the issue

vinocher:
> Sure. my only reasoning was to avoid overengineering, but if you think that is a likely scenario than we can support that

I think keeping it as is by following the YAGNI principle seems correct. I was just curious.

src/addons/addonapi.cpp
```diff
@@ -92,6 +92,35 @@ void AddonApi::initialize() {
...
+  timer->setInterval(interval);
+  m_timers.insert(m_addon->id(), timer);
+
+  connect((m_timers.value(m_addon->id())), &QTimer::timeout, this,
+          [callback]() { callback.call(); });
+
+  timer->start();
+}
+
+void AddonApi::clearTimedCallbacks() {
```
vinocher, line 113: Shouldn't the new name changes be made in the header file too. Should clearTimedCallbacks be named clearTimedCallback because only one callback is beeing cleared? Should setTimedCallback call clearTimedCallback to clear any previous callback?

MattLichtenstein: you beat me to it... working on removing the `clearTimedCallback` function altogether and just using a single timer object instead of a map. setting one timer will overwrite the other one

src/models/subscriptiondata.cpp
```diff
@@ -50,12 +50,13 @@ bool SubscriptionData::fromSettings() {
...
   if (json.isEmpty() || !fromJsonInternal(json)) {
     return false;
   }
 
   m_rawJson = json;
+  emit changed();
```
vinocher, line 59: One concern is that it hasn't really changed, and does not follow the pattern in user.cpp, serverdata.cpp etc, where the signal is emitted only on an actual change.

src/addons/addonapi.cpp
```diff
@@ -90,6 +90,22 @@ void AddonApi::initialize() {
...
+}
+
+void AddonApi::setTimedCallback(int interval, const QJSValue& callback) {
+  if (!callback.isCallable()) {
+    logger.debug() << "No callback received";
+    return;
+  }
+
+  // Creates a timer for `interval` msec before executing `callback`
+  m_timer.setInterval(interval);
```
vinocher, line 104: Does the previous timer need to be stopped or does setInterval do that automatically?

addons/message_plan_upgrade_14d/manifest.json
flodolo, line 1: Thanks for unifying the 2 addons.

Can we rename the folder, e.g. `message_plan_upgrade_scheduled`, or something like that, given it's not just showing after 14 days (as the name here and folder imply)?

flodolo (approved):

src/addons/addonapi.cpp
```diff
@@ -90,6 +90,22 @@ void AddonApi::initialize() {
   if (s_constructorCallback) {
     s_constructorCallback(this);
   }
+
+  m_timer.setSingleShot(true);
+}
+
+void AddonApi::setTimedCallback(int interval, const QJSValue& callback) {
```
brizental, line 97:
```suggestion
void AddonApi::setTimeout(const QJSValue& callback, int interval) {
```

This design would mimic almost exactly the JavaScript setTimeout API, which I think is nice.

src/addons/addonapi.cpp
brizental, line 104: Why was this changed to an interval instead of a single shot timer?

And also why was the list dropped? This prevents us from creating more than one timer. If that is the proposed design we'd need to add a check at the top of this function to only allow setting a timer when no other timer is already set... But I'd rather just make it possible to have more than one timer.

src/addons/addonmessage.cpp
brizental: LOL, I retrieve my ~dumb~ comment.

addons/message_plan_upgrade_14d/conditions.js
```diff
@@ -0,0 +1,47 @@
...
+        api.setTimedCallback(fourteenDaysAfterSubscriptionStarted - now, () => computeCondition());
+        condition.disable()
+      }
+      //Between 14 and 87 days into the subscription, show message
+      else if (now >= fourteenDaysAfterSubscriptionStarted && now < eightySevenDaysAfterSubscriptionStarted) {
+        api.setTimedCallback(eightySevenDaysAfterSubscriptionStarted - now, () => computeCondition())
+        api.addon.date = fourteenDaysAfterSubscriptionStarted / 1000
+        condition.enable()
+      }
+      // After 87 days into the subscription, re-surface the message (undimiss, unread, updating timestamp)
```
brizental, line 39: Question: do we care if the user has not read or dismissed the message?

brizental (changes requested):
Still missing tests. Note that we can mock large jumps in current date using https://sinonjs.org/releases/latest/fake-timers/, for example.

src/models/subscriptiondata.cpp
MattLichtenstein: I suppose it depends on how you look at it

On one hand, all of the `Q_PROPERTY`'s that rely on this `changed()` signal being emitted ARE being changed. For example, if we were to load a UI that used this data before this subscriptiondata object was initialized, that UI would not update to reflect the correct values after subscriptiondata initialization

On the other hand, the subscription data on launch is being pulled from settings, so it would be the same data as the last session. In that respect, the data hasn't change

I don't see any particular drawbacks to emitting the signal here (are there any other places we catch this signal?), but I do also dislike the inconsistency with the other data model classed. What do you think about just creating a different signal. Maybe `Initialized()`?

src/addons/addonapi.cpp
MattLichtenstein: It's still a singleshot timer, `setInterval(msec)` just sets the timer interval, but doesn't necessarily repeat. But you're right, we don't need this as we can just pass the interval in when we start the timer.

For this particular message, only one timer is necessary, so in an effort to not overengineer and add unused complexity, I have simplified it to one overwriteable timer at a time

brizental: Duh, with regards to the single shot question...

Hum. Ok, let's go with the YAGNI principle. Still, this needs two things 1. unit tests 2. an error in case a timer is created while there is another timer still ongoing.

MattLichtenstein: if a timer is created while one is already running, we will overwrite the original timer

src/addons/addonapi.cpp
vinocher: IMO, I think this is a better name than the Javascript method. Switching the parameter order to match the setTimeout looks good.

src/models/subscriptiondata.cpp
vinocher: initialized sounds good.

src/addons/addonapi.cpp
MattLichtenstein: I also like the name `setTimedCallback` a little more. I'm not sure we need to line this function up with javascript usage.

src/addons/addonmessage.h
```diff
@@ -38,7 +38,15 @@ class AddonMessage final : public Addon {
       QString formattedDate READ formattedDate NOTIFY retranslationCompleted)
   Q_PROPERTY(Badge badge MEMBER m_badge WRITE setBadge NOTIFY badgeChanged)
 
-  enum Badge { None, Warning, Critical, NewUpdate, WhatsNew, Survey };
+  enum Badge {
+    None,
+    Warning,
+    Critical,
+    NewUpdate,
```
brizental, line 45: super nit... but NewUpdate is redundant.
```suggestion
    Update,
```

tests/functional/addons/09_message_upgrade_to_annual_plan/message_upgrade_to_annual_plan/conditions.js
```diff
@@ -0,0 +1,48 @@
+(function(api, condition) {
```
brizental, line 1: You don't need the duplicate test addon. You can just test the real thing.

tests/functional/testAddons.js
```diff
@@ -305,4 +305,105 @@ describe('Addons', function() {
...
+      //        "time",
+      //        true, 
+      //        '87 days after subscription created'],
+      [() => Date.now() - 1000 * 60 * 60 * 24 * 88,
+             "yesterday",
+             true, 
+             '88 days after subscription created'],
+    ];
+
+    const getNextTestCase = testCases[Symbol.iterator]();
```
brizental, line 339: Fancy 💅

tests/functional/testAddons.js
```diff
@@ -305,4 +305,105 @@ describe('Addons', function() {
...
+      }
+    }
+
+    // We call this once before all tests to set up the first test,
+    // we can't use beforeEach because that is executed after the guardian
+    // endpoints are overriden.
+    //
+    // We need to setup for the next test before it even starts for the
+    // overrides to apply.
+    setNextSubscriptionStarted(this.ctx);
```
brizental, line 367: Isn't it possible to do `before` or `beforeAll`?

tests/unit_tests/testaddonapi.cpp
```diff
@@ -186,4 +186,35 @@ void TestAddonApi::foobar() {
...
+      message, ":/addons_test/api_settimedcallback.js");
+  QVERIFY(!!a);
+
+  QTimer timer;
+
+  timer.setSingleShot(true);
+  timer.start(1000);
+
+  QObject::connect(&timer, &QTimer::timeout,
+                   [&]() { QVERIFY(a->conditionApplied()); });
```
brizental, line 217: Are you sure this assertion is executed before the test ends?

brizental (commented):
Test are looking _beautiful_. Thank you for taking the time to add them! Left some questions and comments, nothing major.

src/addons/addonmessage.h
MattLichtenstein: agree. hate it. but this is not new (linter change) and is already being used by all of the "what's new" addons

tests/functional/testAddons.js
MattLichtenstein: the comment above says 

> // We call this once before all tests to set up the first test,
    // we can't use beforeEach because that is executed after the guardian
    // endpoints are overriden.

brizental: I know but I wonderd if `before` and `beforeAll` would be called at an appropriate time.

tests/unit_tests/testaddonapi.cpp
MattLichtenstein: nope, will add a signal spy

tests/functional/testAddons.js
```diff
@@ -305,4 +305,103 @@ describe('Addons', function() {
...
+        await vpn.resetAddons('prod');
+
+        //Load messages
+        const loadedMessages = await vpn.messages();
+
+        //If the message is supposed to be enabled, lets check the timestamp
+        if (shouldBeAvailable) {
+            await vpn.waitForCondition(async () => (parseInt(await vpn.getMozillaProperty('Mozilla.Shared', 'MZAddonManager', 'count'), 10) > 0));
+            
+            //Check timestamp
```
brizental, line 381: nit: Maybe move this comment down to line 384.

brizental (approved):

vinocher (approved):

PR #8576, opened 2023-11-16
Draft: RFC- Firefox + VPN

strseb:
lel


docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+In this State the Extension can simply set the Proxy for those network requests to `null` to use the user's active server.
+If a custom exit location is set it must set it to the `socks5` proxy of that endpoint. No changes are needed on the Client as this is how MAC uses the current API. 
+
+
+#### State 2: Device VPN is OFF & Page VPN is On
+
+Simplified, the VPN has the following states:
+```cpp
+  enum State {
+    StateOff, // Noting protected
```
bakulf, line 59: Nothing

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+    (...) // unrelated intermediate states
+  };
+```
+An extension can send `{"t":"activate"}` to the VPN to turn on whole-device protection. 
+For this case, we propose to introduce a new state `StateOnPartial` that will be activated on a call of an extension to `{"t":"activate"}` instead. It is similar to `StateOn` with the following properties: 
+- The `allowedIp` for this activation is set to `10.124.0.0/20` so that only the Mullvad Exit Proxies are accessible on the VPN Network Device. Therefore, all other traffic will not use the VPN. 
+- A Mullvad DNS server will be set, so that the Mullvad proxies are resolvable. If a custom DNS is set, we need to set a Mullvad DNS as 2nd DNS server.
+- The Client UI does not show an activation. (or a new special UI state)
+> Q: Do we need a new UI state ? or like a "firefox is using the vpn?" 
+- If the Client activates, it will attempt to switch to `StateOn` and protect the whole device.
```
bakulf, line 70: Tell how. I guess we need a stop/start/switch.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+- An Extension may request `{"t":"deactivate"}` only if the client is in `StateOnPartial` and it was the activator. The Client will then go to `StateOff`.
```
bakulf, line 71: How to know it was the activator?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+
+The extension must then set the proxy to the `socks5` URL of the current `exitServer` OR, if a custom exit location is set, to the `socks5` proxy of that endpoint.
+
+#### State 3: Device VPN is On & Page VPN is Off
```
bakulf, line 75: Do we really want to cover this scenario? It seems that we are offering a way to workaround the VPN protection. 

In theory this proxy could be abused by malicious app.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+- opens a socks5 proxy locally on a random port
+- generates a username and password for this session
+- prints the port, username, password to STDOUT so that the parent process may gather that info. 
+
+The VPN client will then pass that info to the Extension inside it's `status` response as an extra field. The extension must use that proxy for pages wanting to bypass the VPN.
+
+For Platforms where split tunneling is supported, the proxy is permanently fixed in the split-tunnel rule list, making sure the vpn connection cannot be used. 
+
+For Platforms where split tunneling is not available, the proxy will use socket options to make sure the right network device is used.
+> Q: do we need more details here?
```
bakulf, line 89: Yes :)

Question: a malicious app can simulate the Firefox implementing the native messaging protocol. How can you prevent this?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+#### State 4: Device VPN is OFF & Page VPN is Off
+This State is trivial, no proxy needs to be set by the extension. 
+
+### Other API Changes 
+
+#### Renaming
+We will rename the current C++ Extension API classes to ... ExtensionAPI. They are currently under `server/` and called `ServerHandler` && `ServerConnection` which is an overloaded term for a VPN. 
+
+#### Split-Tunnel Lists
+Instead of reporting all excluded apps in `disabled_apps`, we will limit the response to executables containing `firefox`. The Extension may then limit it's funcitonality until resolved, as Firefox cannot access the VPN if it is excluded.
```
bakulf, line 104: Using name convention is another attacking vector. A malicious app can duplicate itself, rename the new binary as 'firefox-foo' and workaround the VPN protection.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+- prints the port, username, password to STDOUT so that the parent process may gather that info. 
+
+The VPN client will then pass that info to the Extension inside it's `status` response as an extra field. The extension must use that proxy for pages wanting to bypass the VPN.
+
+For Platforms where split tunneling is supported, the proxy is permanently fixed in the split-tunnel rule list, making sure the vpn connection cannot be used. 
+
+For Platforms where split tunneling is not available, the proxy will use socket options to make sure the right network device is used.
+> Q: do we need more details here? 
+
+If the daemon has support for per-app firewall rules (currently Windows), we will make sure only Firefox may access the port of the proxy to make sure other local apps cannot use the proxy to get the real IP of the user.
```
brizental, line 91: What if the daemon doesn't have it?

docs/rfcs/9999-vpn-firefox-integration.md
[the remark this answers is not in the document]
brizental: >  The Extension may then limit it's funcitonality until resolved, as Firefox cannot access the VPN if it is excluded.  

nit: Add this to the preconditions list. And also, it should have error state designs.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+
+![](https://github.com/mozilla-mobile/mozilla-vpn-client/assets/9611612/aa8331cf-f0cb-489b-a7e2-293bce422551) 
+
+VPNs currently allow you to select the country through which you want your traffic to emerge to the Internet and users use this to express a variety of geolocation preferences. These controls are not flexible enough - users can only select a single location through which all traffic should go through. 
+
+Therefore we want to propose a new Web-Extension that extends on top of the VPN+MAC API, giving users the options to:
+ - Enable The VPN only for Firefox 
+ - Disable The VPN only for Firefox 
+ - Exclude a Page from VPN protection
+ - Specify an Exit Location a Page.
```
Gela, line 23:
```suggestion
Therefore we want to propose a new Web-Extension that extends the VPN+MAC API, giving users the option to:
 - Enable the VPN only on Firefox 
 - Disable the VPN only on Firefox 
 - Exclude a Firefox tab from VPN protection
 - Specify an exit location on a tab

```

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+|-----------------|------------------|-------------------|
+| Page VPN is On  | 1                | 2                 |
+| Page VPN is Off | 3                | 4                 |
+
+In the following sections, we will explain what changes we need inside the client to support each of the cases and where the extension must proxy traffic to. 
+
+
+#### State 1: Device VPN is On & Page VPN is On
+In this State the Extension can simply set the Proxy for those network requests to `null` to use the user's active server.
+If a custom exit location is set it must set it to the `socks5` proxy of that endpoint. No changes are needed on the Client as this is how MAC uses the current API.
```
Gela, line 51:
```suggestion
In this state the extension can simply set the proxy for those network requests to `null` to use the user's active server.
If a custom exit location is set it must set it to the `socks5` proxy of that endpoint. No changes are needed on the client as this is how MAC uses the current API. 

```

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+  enum State {
+    StateOff, // Noting protected
+    StateOn, // Whole device protected
+    (...) // unrelated intermediate states
+  };
+```
+An extension can send `{"t":"activate"}` to the VPN to turn on whole-device protection. 
+For this case, we propose to introduce a new state `StateOnPartial` that will be activated on a call of an extension to `{"t":"activate"}` instead. It is similar to `StateOn` with the following properties: 
+- The `allowedIp` for this activation is set to `10.124.0.0/20` so that only the Mullvad Exit Proxies are accessible on the VPN Network Device. Therefore, all other traffic will not use the VPN. 
+- A Mullvad DNS server will be set, so that the Mullvad proxies are resolvable. If a custom DNS is set, we need to set a Mullvad DNS as 2nd DNS server.
```
vinocher, line 67: > we need to set a Mullvad DNS as 2nd DNS server.
Could this explain why a Mullvad DNS needs to be set as second DNS server?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+The extension must then set the proxy to the `socks5` URL of the current `exitServer` OR, if a custom exit location is set, to the `socks5` proxy of that endpoint.
+
+#### State 3: Device VPN is On & Page VPN is Off
+
+If the full device is protected and the VPN integration is enabled, we will utilize a Socks5 Proxy running on localhost that will make sure its outgoing traffic utilizes the non-vpn default route. 
+
+For that, we will create an external binary that:
+- opens a socks5 proxy locally on a random port
+- generates a username and password for this session
+- prints the port, username, password to STDOUT so that the parent process may gather that info.
```
vinocher, line 82: > prints the port, username, password to STDOUT

How will this be done on Windows? Won't an IPC mechanism be needed?

BTW, a diagram may be useful for the reader to understand some of these states.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+
+Therefore we want to propose a new Web-Extension that extends on top of the VPN+MAC API, giving users the options to:
+ - Enable The VPN only for Firefox 
+ - Disable The VPN only for Firefox 
+ - Exclude a Page from VPN protection
+ - Specify an Exit Location a Page.
+
+We already have integrated into Firefox with [Multi Account Containers](https://github.com/mozilla/multi-account-containers) which allows users to specify an exit location for a container. 
+The following parts expect knowledge on how the MAC Extension works, for a refresher scroll to appendix :) 
+
```
vinocher, line 27: The Appendix describes the role of the VPN Client. What happens on a mobile platform where the VPN is active but the client has been closed?

docs/rfcs/9999-vpn-firefox-integration.md
strseb: Right now using the api has arbitrary read on `Settingsholder::excludedApps`. So any person with local code execution can learn of any excluded apps. 
An attacker may trivially forge a binary that would bypass this rule, (still a user would need to manually add foo-firefox)  but the attacker would already know that its own binary foo-firefox is on the system? Also it can trivially know it's split tunneled by introspecting it's created sockets. 
Can you elaborate what the attack vector is?

docs/rfcs/9999-vpn-firefox-integration.md
strseb: >In theory this proxy could be abused by malicious app. 

I agree, however if we can only ship one of those states, the per tab split tunnel would be what Product would request 😅

docs/rfcs/9999-vpn-firefox-integration.md
strseb: I agree but also I'm a bit torn 😅 
On the one hand, the existence of the cli means any gce-exploit that can create a process can: `activate, deactivate, add itself to split-tunnel` 

On the other hand, i myself am not convinced that this means we should tack on more things. 

How would you feel if we had a "paring" flow? Something like: 
An extension would trigger a "permission prompt" on the client by sending an HMAC Public key. If the user accepts we added to a "well-known-keys" list. 
All other messages need's to be authed using that hmac [jwt maybe?](https://jwt.io/introduction). 


That would at least make it non trivial.

docs/rfcs/9999-vpn-firefox-integration.md
strseb: the MAC integration is not supported on mobile :) , added that to the appendix.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,164 @@
...
+----
+
+## Preconditions
+
+For this feature to be enabled, a Firefox user on the desktop needs to have the VPN installed, be logged in and have an active subscription.
```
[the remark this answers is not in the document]
strseb: If the preconditions are not met it will be hidden. - It is explicitly at this stage not another firefox surface for advertisement but a feature we want to offer to existing customers.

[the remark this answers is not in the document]
brizental: Got it. Please note that in the text.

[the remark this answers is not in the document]
brizental: FWIW, in the Figma designs the "Activate VPN" screens seem like the answer to my first question though.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,167 @@
...
+- Authors:
+    - [@strseb](https://github.com/strseb)
+    - [@lesleyjanenorton](https://github.com/lesleyjanenorton)
+    - [@oskirby](https://github.com/oskirby)
+- RFC PR: -
+- Implementation GitHub issue: -
+- [Design @Figma](https://www.figma.com/file/s13B7zs27cadZXUyvxobgW/Better-Together-Explorations?type=design&node-id=259-4830&mode=design&t=4BmtgXMzjnubd9gc-0)
+
+-----
+# Firefox + VPN Better together
```
brizental, line 12: I have a random question: if we are able to do split tunneling per URL domain, why don't we provide that feature in the app as well as in the extension?

brizental: Ah ok I think I understand that we actually can't do this, this is only possible due to the Proxy browser extension stuff.

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,167 @@
...
+
+Simplified, the VPN has the following states:
+```cpp
+  enum State {
+    StateOff, // Nothing protected
+    StateOn, // Whole device protected
+    (...) // unrelated intermediate states
+  };
+```
+An extension can send an `ExtensionMessage<activate>` to the VPN to turn on whole-device protection.
```
lesleyjanenorton, line 66: > to the VPN to turn on whole-device protection.

Should this be 'turn on partial protection'?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,167 @@
...
+- opens a socks5 proxy locally on a random port
+- generates a username and password for this session
+- prints the port, username, password to STDOUT so that the parent process may gather that info. 
+
+The VPN client will then pass that info to the Extension inside it's `status` response as an extra field. The extension must use that proxy for pages wanting to bypass the VPN.
+
+For Platforms where split tunneling is supported, the proxy is permanently fixed in the split-tunnel rule list, making sure the vpn connection cannot be used. 
+
+For Platforms where split tunneling is not available, the proxy will use socket options to make sure the right network device is used.
+> Q: do we need more details here?
```
lesleyjanenorton, line 91: > Q: do we need more details here? 

Eventually yes but since we're only concerned with Windows at the moment I don't think we need it for the purposes of this RFC.

docs/rfcs/9999-vpn-firefox-integration.md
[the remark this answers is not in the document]
strseb: yes

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,188 @@
...
+|-----------------|------------------|-------------------|
+| Page VPN is On  | 1                | 2                 |
+| Page VPN is Off | 3                | 4                 |
+
+In the following sections, we will explain what changes we need inside the client to support each of the cases and where the extension must proxy traffic to. 
+
+
+#### State 1: Device VPN is On & Page VPN is On
+In this state the extension can simply set the proxy for those network requests to `null` to use the user's active server.  
+If a custom exit location is set it must set it to the `socks5` proxy of that endpoint. No changes are needed on the client as this is how MAC uses the current API.
```
vinocher, line 57: > If a custom exit location is set it must set it to the `socks5` proxy of that endpoint.

Could some more information be added on how a custom exit location will be handled by setting it to the 'socks5' proxy, for the benefit of the reader?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,188 @@
...
+```mermaid 
+---
+title: State Diagramm v2
+---
+stateDiagram-v2
+    StateOff --> StateOn: activate
+    StateOff --> StateOnPartial: activatePartial
+    StateOnPartial --> StateOn: silentServerSwitch
+    StateOnPartial --> StateOff: deactivate
+    StateOn -->StateOff : deactivate
```
vinocher, line 90: > StateOn -->StateOff : deactivate

It appears that deactivating the Client in StateOn will turn off the VPN for a Firefox tab. Is that what we want? Or should we transition to StateOnPartial in that case?

docs/rfcs/9999-vpn-firefox-integration.md
```diff
@@ -0,0 +1,188 @@
...
+  };
+```
+An extension can currently send an `ExtensionMessage<activate>` to the VPN to turn on whole-device protection. 
+For this case, we propose to introduce a new state `StateOnPartial` that will be activated on a call of an extension to `ExtensionMessage<activate>` instead. It is similar to `StateOn` with the following properties: 
+- The `allowedIPAddressRanges` for this activation is set to `10.124.0.0/20` so that only the Mullvad Exit Proxies are accessible on the VPN Network Device. Therefore, all other traffic will not use the VPN. 
+- A Mullvad DNS server will be set, so that the Mullvad proxies are resolvable. If a custom DNS is set, we need to set a Mullvad DNS as 2nd DNS server.
+- The Client UI does not show an activation. (or a new special UI state)
+> Q: Do we need a new UI state ? or like a "firefox is using the vpn?" 
+- If the Client activates, it will attempt to switch to `StateOn` and protect the whole device.
+- An Extension may request `ExtensionMessage<deactivate>` only if the client is in `StateOnPartial`. The Client will then go to `StateOff`.
```
brizental, line 77: What happens if the client is in StateOn?

PR #8581, opened 2023-11-17
VPN-5889 Remove the ConnectionManager object and merge classes back in Controller

Gela:
## Description

The feed backs coming in for the Connection Diagnostics tech spec are starting to slow down and the general approach is agreed upon in this RFC (https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8463 ):

Moving forward, we are going to continue using the Controller class as we did prior to introducing ConnectionManager (i.e. the controller will communicate with the frontend to show connecting, disconnecting, on, off, etc).

Given that we’ve agreed on the big picture approach of keeping the controller as is, this PR merges the current controller and connection manager code back in the controller and remove any reference to Connection Manager since that name will no longer be used. This would essentially revert [this PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7774).

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5889


brizental (approved):
Make sure to get tests passing. Otherwise 👍

Issue #8423, opened 2023-10-30
[a11y] Connection Information elements are not narrated by the screen reader


data-sync-user:
**Important details:**

* The issue is not reproducible when navigating backwards, using CAPS+U key combination.
* Not reproducing on macOS;

**Affected versions:**

* 2.19.0 (2.202310280712);

**Tested Platforms:**

* Windows;

**Prerequisites:**

* Have an account with an active subscription for Mozilla VPN;

* Be signed in;

* Turn VPN on;

* Be in Connection Information screen

**Steps to reproduce:**

1. Using the keyboard, navigate in the Connection Information screen (TAB/SHIFT + TAB, CAPS +O).
1. Observe how the elements are narrated.

**Expected result:**

* All elements in the Connection Information screen are narrated.

**Actual result:**

* Connection Information title and exit server are not narrated when using TAB/SHIFT+TAB/CAPS+O key combinations.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5783)
┆Reporter: Raluca Monica Ilban

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this on Mozilla VPN 2.19.0 (2.202311181616), using Windows 10/11, and the following were observed:

  * when the entire app window is in focus, text elements from the CI screen are narrated;
  * when navigating backwards, using CAPS + U / CAPS + left arrow, text elements from the CI screen are narrated;
  * when navigation using CAPS + O / CAPS + right arrow, text elements from the CI screen are still NOT narrated;
    * text elements are narrated when navigating forward only if the user navigated backwards previously and the ‘Connection inforation’ text element is in focus;

Based on the above, it seems that the behavior did not change at all after the fix. Vinoo Cherian, could you please take a look at this? Thank you!

Attaching a video with the behavior as well. 

!bandicam 2023-11-20 11-09-32-825.mp4|width=1152,height=896!

data-sync-user:
➤ Vinoo Cherian commented:

Bianca Hidecuti, sorry, I misunderstood this issue. It looks like this one is about the order of navigation. This navigation order is currently by design because the close button is the last element in the Connection Information screen, so the user needs to navigate backwards to access the other elements. The design looks broken. I’m looking closer to see how this can be fixed.

data-sync-user:
➤ Vinoo Cherian commented:

[...]

PR #8596, opened 2023-11-20
VPN-5783: Change IPInfoPanel's a11y navigation order so that content is after the close button

vinocher:
## Description

The IPInfoPanel was before its close button in the QML, and focus was on the close button when the panel was opened. This meant that the user needed to navigate backwards from the Close button to read the panel text using a screen reader, which is not intuitive. Fixed by moving IPInfoPanel after its Close button. This allows the user to read the panel's text by navigating forward in a screen reader.

Also fixed the following:
Previously the IPInfoPanel had its z-order set to the same z-order as the buttons, which hid some visibility bugs. Specifically, the visible properties of connectionInfoToggleButton and ipInfoToggleButton were not correct. connectionInfoToggleButton  should be visible only if the IPInfoPanel  is not opened.

## Reference
[VPN-5783](https://mozilla-hub.atlassian.net/browse/VPN-5783), [GitHub 8423](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/8423)

## Checklist
    
[VPN-5783]: https://mozilla-hub.atlassian.net/browse/VPN-5783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/screens/home/controller/ControllerView.qml
+++ b/src/ui/screens/home/controller/ControllerView.qml
@@ -517,6 +517,66 @@ Item {
         }
     }
 
+    MZIconButton {
+        id: ipInfoToggleButton
+        objectName: "ipInfoToggleButton"
+
+        property var connectionInfoCloseText: MZI18n.GlobalClose
+
+        anchors {
+            right: parent.right
+            rightMargin: MZTheme.theme.windowMargin / 2
+            top: parent.top
+            topMargin: MZTheme.theme.windowMargin / 2
+        }
+        accessibleName: ipInfoPanel.isOpen
+            ? connectionInfoCloseText
+            : MZI18n.ConnectionInfoConnectionInformation
+        Accessible.ignored: !enabled || !visible
+        buttonColorScheme: MZTheme.theme.iconButtonDarkBackground
+        enabled: visible && VPNConnectionHealth.stability !== VPNConnectionHealth.NoSignal
+        opacity: visible ? 1 : 0
+        visible: (VPNController.state === VPNController.StateOn || VPNController.state === VPNController.StateSilentSwitching)
+            && !connectionInfoScreen.isOpen
+            && !connectionInfoScreen.isTransitioning
+        z: 1
+        onClicked: {
+            ipInfoPanel.isOpen = !ipInfoPanel.isOpen;
+
+            if (ipInfoPanel.isOpen) {
+                Glean.interaction.openConnectionInfoSelected.record({
+                    screen: "main",
+                });
+            } else {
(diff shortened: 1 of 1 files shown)


src/ui/screens/home/controller/ControllerView.qml
```diff
@@ -686,12 +648,11 @@ Item {
...
             ? connectionInfoCloseText
             : MZI18n.ConnectionInfoConnectionInformation
+        Accessible.ignored: !enabled || !visible
         buttonColorScheme: MZTheme.theme.iconButtonDarkBackground
         enabled: visible && VPNConnectionHealth.stability !== VPNConnectionHealth.NoSignal
         opacity: visible ? 1 : 0
-        visible: connectionInfoToggleButton.visible
-            && !connectionInfoScreen.isOpen
-            && !connectionInfoScreen.isTransitioning
+        visible: !connectionInfoScreen.isOpen && !connectionInfoScreen.isTransitioning
```
MattLichtenstein, line 655: Won't this make it IpInfo button visible when the VPN is off?

MattLichtenstein (commented):
Everything else looks good. 

Do you think it would make sense to move the VPNToggle after the IpInfo button while we're fixing navigation order here. Seems weird to me that we go from Connection info button > VPN toggle > IpInfo button when the VPN is on

src/ui/screens/home/controller/ControllerView.qml
vinocher: Yes. Thanks for catching.

MattLichtenstein:
> Yes, that seems like the correct order. However, that will mean that a keyboard user will need to do an extra tab to reach the Toggle button, the most important part of the UI. 

I'd think it makes more sense to have to have keyboard navigation flow in a natural and expected order than to save the user one extra button press.

> Also, not sure if we have any tests that depend on that order. Perhaps this should be done in another change.

Don't believe we have any tests that verify tab order, but I am totally fine with coming back to this in the future

src/ui/screens/home/controller/ControllerView.qml
vinocher: Fixed.

vinocher:
> I'd think it makes more sense to have to have keyboard navigation flow in a natural and expected order than to save the user one extra button press.
> Don't believe we have any tests that verify tab order, but I am totally fine with coming back to this in the future 

Yes, it makes sense to do this now. Fixed.

MattLichtenstein (approved):
👌

PR #8603, opened 2023-11-21
Move unit tests for linux to taskcluster?

strseb:
## Description
Currently to build tests, we pull our 3rdparty deps from the taskcluster index, pointing at main. 
This introduces a race between i.e updating qt in a pr and having the tests horribly fail due to that. 

Therefore they should live inside taskcluster, so they properly get the correct dependencies.

### DIFF
--- a/taskcluster/ci/test/unit.yml
+++ b/taskcluster/ci/test/unit.yml
@@ -21,14 +21,14 @@ ctest-linux:
             - qt-linux
     worker-type: b-linux
     worker:
-        docker-image: {in-tree: linux-qt6-build}
+        docker-image: {in-tree: linux-build-focal}
     run:
         cwd: '{checkout}'
         command: >-
             pip install -r requirements.txt &&
             cmake -S . -B build -GNinja -DCMAKE_PREFIX_PATH=${MOZ_FETCHES_DIR}/qt_dist/lib/cmake -DCMAKE_BUILD_TYPE=Debug &&
             cmake --build build/ --target build_tests &&
-            ctest  --test-dir .\build\vpn\ 
+            ctest  --test-dir ./build/vpn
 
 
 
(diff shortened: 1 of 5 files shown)


taskcluster/ci/test/unit.yml
```diff
@@ -0,0 +1,39 @@
...
+    worker-type: b-linux
+    worker:
+        docker-image: {in-tree: linux-qt6-build}
+    run:
+        cwd: '{checkout}'
+        command: >-
+            pip install -r requirements.txt &&
+            cmake -S . -B build -GNinja -DCMAKE_PREFIX_PATH=${MOZ_FETCHES_DIR}/qt_dist/lib/cmake -DCMAKE_BUILD_TYPE=Debug &&
+            cmake --build build/ --target build_tests &&
+            ctest  --test-dir .\build\vpn\
```
bhearsum, line 31: Nit: is there a reason these are backslashes instead of forward slashes?

taskcluster/ci/test/unit.yml
```diff
@@ -0,0 +1,39 @@
...
+    run:
+        using: run-task
+        use-caches: true
+
+ctest-linux:
+    description: "Run full C-Test"
+    fetches:
+        toolchain:
+            - qt-linux
+    worker-type: b-linux
```
bhearsum, line 22: With Firefox, we usually run tests on target OS' (usually this means Ubuntu for Linux). I see that these tests require doing a build first, so that's probably not viable here, but it may be worth considering for the future. (With Firefox, the build system is capable of packaging tests for consumption elsewhere - so the tests tasks end up downloading the build and the test packages.)

bhearsum (commented):
I started reviewing this while it was still a draft...I'll drop the comments I wrote up already but I'll hold off on any further reviewing until it's ready.

PR #8613, opened 2023-11-21
Chore: Fix Rust unit tests due to expired test certificate.

oskirby:
## Description
This addresses a TODO item in the Rust unit tests that were causing our unit test job to fail due to an expired certificate.

We had some tests in the Balrog module that tried to verify certificates that we fetched from the production servers, and those certificates recently expired. The Rust language doesn't really have a mechanism to mock things like the system time when running tests, so I opted to work around the issue by parsing a timestamp out of the certificates when compiled with `#[cfg(test)]`

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/signature/src/lib.rs
+++ b/signature/src/lib.rs
@@ -6,8 +6,8 @@ use crate::balrog::*;
 use crate::logger::*;
 use ring::signature;
 use std::ffi::CStr;
-use std::os::raw::c_uchar;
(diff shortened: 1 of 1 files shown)


brizental (approved):
I'll also go ahead and merge this to unblock other PRs. Hope you all don't mind 😰

PR #8662, opened 2023-11-27
Chore: Cherry pick translations report generation job to release branch

brizental:
We just need to taskcluster job to generate the translations report for this branch as well. The rest is a bonus and doesn't really matter.


strseb (approved):

strseb:
>Rust tests are failing because I didn't cherry pick that fix here... Should I? 
They are not required, so imho no.

PR #8692, opened 2023-11-28
Chore: Fix race in AuthenticationInApp functional test

oskirby:
## Description
The `testAuthenticationInApp.js` functional test has recently started to fail on Linux, and this appears to be due to a race between the functional test and the asynchronous handling of the 2FA network request. In particular, the tests click on the verification button and then immediately check the glean telemetry for the expected events, but those events are set only once the 2FA network requests are completed. This leaves a race condition where the tests fail to wait for the telemetry to be set.

To resolve the problem, I just added a few waits before checking telemetry on a successful 2FA verification.

## Checklist
    
### DIFF
--- a/tests/functional/testAuthenticationInApp.js
+++ b/tests/functional/testAuthenticationInApp.js
@@ -639,7 +639,7 @@ describe('User authentication', function() {
             await goToPasswordScreen(false);
         });
 
-        it("impression event is are recorded", async () => {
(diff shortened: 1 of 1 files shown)


mcleinman (approved):
👍🏻

lesleyjanenorton (approved):
✨🥇🍾  Thank you!

PR #8694, opened 2023-11-29
VPN-5948 - Abort controller activation when there is no current device

brizental:
The issue here is quite simple: we do not handle the case when there is no current device available when activating the controller. That was the cause of VPN-5948.

This can happen if a device is removed and immediatelly an activation is enqueued. 

A better option might be to just stop the task that is activating the VPN, this is a bit less drastic and maybe less risky?


brizental:
I agree this should be tested behaviour. I would like to write these tests, but as of right now that is not trivial. There is a ticket for that: https://mozilla-hub.atlassian.net/browse/VPN-4445

As discussed offline, I have added a label to this PR to signal that it needs tests, but will be merged without them.

PR #8698, opened 2023-11-29
VPN-5900: Do not send pings for server switch

mcleinman:
## Description

We want to prevent sending the pings if its due to a server switch. We can take advantage of the fact that all VPN activations come through Controller::activate, and at that point the current state informs us *why* we're switching. I have notes on the ticket that show my work for why I believe these are the only paths.

This isn't pretty, but is better than the alternatives that a couple of us could come up with.

Credit to @Gela for this approach.

## Reference

VPN-5900

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -659,7 +659,12 @@ void Controller::connected(const QString& pubkey,
   // We have succesfully completed all pending connections.
   logger.debug() << "Connected from state:" << m_state;
   setState(StateOn);
-  emit newConnectionSucceeded();
+
+  logger.debug() << "Connection happened due to server switch:"
+                 << isActivatingFromSwitch;
+  if (!isActivatingFromSwitch) {
+    emit newConnectionSucceeded();
+  }
 
   // In case the Controller provided a valid timestamp that
   // should be used.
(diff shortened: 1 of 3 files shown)


src/controller.cpp
```diff
@@ -659,7 +659,12 @@ void Controller::connected(const QString& pubkey,
   // We have succesfully completed all pending connections.
   logger.debug() << "Connected from state:" << m_state;
   setState(StateOn);
-  emit newConnectionSucceeded();
+
+  logger.debug() << "Connection happened due to server switch:"
+                 << isActivatingFromSwitch;
```
Gela, line 664: This debug statement threw me off for a second because the text initially made me assume that we've encountered a server switch but then realized that depends on the value of `isActivatingFromSwitch`. For the sake of clarity can this be changed to be more explicit in when we log the statement? Possibly:

```suggestion
  if (isActivatingFromSwitch)
  {
     logger.debug() << "Connection happened due to server switch. Do not collect telemetry regarding new connection.";
  }
```

Gela (approved):
This is amazing, thank you!! Is there a test for this already or can we add one?

src/controller.cpp
mcleinman: Added something similar, with logging for both ways. Good call on the confusion there.

mcleinman:
> This is amazing, thank you!! Is there a test for this already or can we add one?

We don't have any tests around silent switching or intentional server switching, as far as I can see. I don't believe any of our tests handle the actual VPN activation - we mock that out, and I'm not sure mocking these methods in `moccontroller` would be a helpful test, as we wouldn't be testing the actual code paths.

Gela (approved):

PR #8717, opened 2023-12-01
Nit: Use cmake --install for wasm

strseb:
That is such a nit. 

We're currently produce the "build" somewhere in the cmake bin dir. 
Let's make sure cmake --install can properly setup a working wasm installation anywhere.


wasm/index.html
```diff
@@ -20,7 +20,7 @@ <h1>Mozilla VPN - <span class="web-viewer">WebAssembly Viewer</span></h1>
       <div class="row">
         <div class="col-sm qml-content">
           <p class="loading-error-message">Trouble loading content</p>
-          <img src="img/loader.svg" class="loader" alt="loading icon" />
+          <img src="loader.svg" class="loader" alt="loading icon" />
```
strseb, line 23: my cmake glob* is flatting all folders and i am too lazy to fix that.

brizental (changes requested):
Please also update the developer documentation.

PR #8718, opened 2023-12-01
Nit: Unify setting target_compile_options

strseb:
## Description
We set `-werror` in multiple places with conditions. I wanted to add wasm to that. So let's have that condition in one place only.

### DIFF
--- a/scripts/cmake/utilities.cmake
+++ b/scripts/cmake/utilities.cmake
@@ -1,0 +1,21 @@
+
+# Set's Defaults like `-Wall -Werror` if we know it will not 
+# explode on that target + compiler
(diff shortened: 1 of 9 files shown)


brizental (approved):
💯

scripts/cmake/utilities.cmake
```diff
@@ -0,0 +1,21 @@
+
+# Set's Defaults like `-Wall -Werror` if we know it will not 
+# explode on that target + compiler
```
Gela, line 3:
```suggestion
# Sets Defaults like `-Wall -Werror` if we know it will not 
# explode on that target + compiler
```

Gela (approved):

PR #8722, opened 2023-12-01
VPN-5971 - Use Conda environment on MacOS auth tests

brizental:
The auth tests are currently broken for MacOS, this is probably going on for a long time.

The issue is the same we had and fixed for MacOS functional tests, so this is very stright forward.

We should monitor this failures more closely though.


brizental:
Great, got the test to run... But now it seems like they also are broken 👍

lesleyjanenorton (approved):

PR #8736, opened 2023-12-04
No ticket - Add sample Nimbus features file

brizental:
Adding this in now, because we will need the path for this file to finish setup of the sidecar container.

This feature is copy pasted from https://github.com/mozilla/blurts-server/blob/4331ae752d797eb1f88c4bf6e75fcd4a833a411d/config/nimbus.yaml

### DIFF
--- a/src/experimentation/nimbus.yaml
+++ b/src/experimentation/nimbus.yaml
@@ -1,0 +1,26 @@
+about:
+  description: Nimbus Feature Manifest for Monitor Web testing
+channels:
+  - staging
+  - production
+features:
+  example-feature:
+    description: An example feature
+    variables:
+      enabled:
+        description: If the feature is enabled
+        type: Boolean
+        default: false
+      something:
+        description: Another variable
+        type: Option<String>
+        default: null
+    defaults:
+      - channel: staging
+        value: { "enabled": true }
+      - channel: production
+        value: { "something": "wicked" }
+
+types:
+  objects: {}
+  enums: {}
--- a/src/experimentation/nimbus.yaml
+++ b/src/experimentation/nimbus.yaml
@@ -1,5 +1,5 @@
 about:
-  description: Nimbus Feature Manifest for Monitor Web testing
+  description: Nimbus Feature Manifest for Mozilla VPN testing
 channels:
   - vpn
 features:
--- a/src/experimentation/nimbus.yaml
+++ b/src/experimentation/nimbus.yaml
@@ -1,8 +1,7 @@
 about:
   description: Nimbus Feature Manifest for Monitor Web testing
 channels:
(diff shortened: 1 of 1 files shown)


src/experimentation/nimbus.yaml
```diff
@@ -0,0 +1,26 @@
+about:
+  description: Nimbus Feature Manifest for Monitor Web testing
+channels:
+  - staging
+  - production
```
yashikakhurana, line 5: we can just define one channel here, lets just say `vpn`

src/experimentation/nimbus.yaml
```diff
@@ -0,0 +1,26 @@
...
+        default: false
+      something:
+        description: Another variable
+        type: Option<String>
+        default: null
+    defaults:
+      - channel: staging
+        value: { "enabled": true }
+      - channel: production
+        value: { "something": "wicked" }
```
yashikakhurana, line 22: same here just keep one channel called `vpn`

yashikakhurana (commented):
The idea is that we will be using one channel only, and we can still target stage and prod when we launch the experiment, via a experimenter console, we do have stage https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus which targets stage only and prod set up https://experimenter.services.mozilla.com/nimbus, we will also update the monitor later with the same strategy

src/experimentation/nimbus.yaml
brizental, line 22:
```suggestion
```

src/experimentation/nimbus.yaml
```diff
@@ -0,0 +1,26 @@
...
+      enabled:
+        description: If the feature is enabled
+        type: Boolean
+        default: false
+      something:
+        description: Another variable
+        type: Option<String>
+        default: null
+    defaults:
+      - channel: staging
```
brizental, line 19:
```suggestion
      - channel: vpn
```

lesleyjanenorton (approved):

yashikakhurana (commented):
looks good to me, thank you @brizental 🎉 🎉

src/experimentation/nimbus.yaml
```diff
@@ -0,0 +1,23 @@
+about:
+  description: Nimbus Feature Manifest for Monitor Web testing
```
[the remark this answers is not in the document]
brizental:
```suggestion
  description: Nimbus Feature Manifest for Mozilla VPN testing
```

lesleyjanenorton: Oops, missed this the first time around. Let's update the description.

brizental:
Closing this because we are going to add this file to Guardian.

Issue #8743, opened 2023-12-04
Linux: Allow deep linking from notifications


data-sync-user:
Since Qt doesn’t provide this functionality, this is a platform-specific child ticket.

For each platform:

* When tapping a specific notification, the app must open to the App Exclusions setting screen. We’ll be later creating the new notification type as part of [https://mozilla-hub.atlassian.net/browse/VPN-5933](https://mozilla-hub.atlassian.net/browse/VPN-5933|smart-link), so in development we can pick any specific notification - “connected to server”, for example - to attach this to when testing the new functionality.
* This work probably consists of two pieces: First, tagging notifications with what category they are (which I don’t believe we currently do on all platforms). Second, looking for that tag when the notification opens the VPN app, and moving to the proper QML screen.
* For Linux, we can assume the app is currently running when the notification is clicked.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-5984)
┆Reporter: Matt Cleinman

data-sync-user:
➤ Bianca Hidecuti commented:

Naomi Kirby, is there anything that QA can verify here? Thanks!

data-sync-user:
➤ Naomi Kirby commented:

To test this, you should be able to cause the client to become visible and navigate to the Get Help view by opening the URL mozilla-vpn://nav/gethelp in a browser, and you can then return to the home screen by navigating to mozilla-vpn://nav/home

So far this is only implemented on Windows and Linux, but we have more tasks to implement this for Android too.

Also note that this functionality is likely to change as we complete more of this epic.

data-sync-user:
➤ Naomi Kirby commented:

Which version of Ubuntu are you using? I am beginning to suspect that this might be another quirk of Snap-based Firefox installations.

Could you see if the command xdg-open mozilla-vpn://nav/gethelp works on the terminal command line?

data-sync-user:
➤ Bianca Hidecuti commented:

Naomi Kirby, the issue is reproducing on Ubuntu 22.04 and Ubuntu 23.04.

Using the xdg-open mozilla-vpn://nav/gethelp command on the terminal opens the client and navigates to the Get help screen (on both Ubuntu 22.04 and 23.04) - I am attaching a video.

!ubuntu.mp4|width=1896,height=1060!

[...]

PR #8757, opened 2023-12-05
VPN-5984: Add support for deep linking on Linux

oskirby:
## Description
This PR attempts to add Linux support for deep linking via the `mozilla-vpn` URL scheme. This is done by registering our application as a MIME type handler for `x-scheme-handler/mozilla-vpn`. When a deep link is launched, the desktop environment will invoke the VPN client command while providing the URL as a positional argument.

Under the hood, we extend the UI pipe to pass the deep link to the running UI instance, and then connect it to the `Navigator` class with a new method: `Navigator::requestDeepLink()`. I have also included a quick and dirty implementation of the navigator code to handle these requests too by doing some path matching against the screen names.

For example, navigating to `mozilla-vpn://nav/gethelp` will bring the UI into the foreground and display `MozillaVPN::ScreenGetHelp`

## Reference
Github issue #8743 ([VPN-5984](https://mozilla-hub.atlassian.net/browse/VPN-5984))

## Checklist
    
[VPN-5984]: https://mozilla-hub.atlassian.net/browse/VPN-5984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/frontend/navigator.cpp
```diff
@@ -206,6 +206,23 @@ void Navigator::requestScreen(int requestedScreen,
...
 
+void Navigator::requestDeepLink(const QUrl& url) {
+  logger.debug() << "Recevied nav link:" << url.toString();
+
+  // Quick and dirty navigation handler.
+  // Lookup the screen by name using QMetaEnum
+  QString topPath = url.path().section('/', 0, 0, QString::SectionSkipEmpty);
+  QString name = QString("Screen%1").arg(topPath);
+  const QMetaEnum meta = QMetaEnum::fromType<MozillaVPN::CustomScreen>();
+  for (int index = 0; meta.keyCount(); index++) {
```
vinocher, line 217: Nice! Consider adding a comment that this is enumerating over all the screen names in the app, by using the CustomScreen enum type's value strings.

src/frontend/navigator.cpp
```diff
@@ -206,6 +206,23 @@ void Navigator::requestScreen(int requestedScreen,
...
+    const char* key = meta.key(index);
+    if ((key != nullptr) && name.compare(key, Qt::CaseInsensitive) == 0) {
+      requestScreen(meta.value(index), NoFlags);
```
vinocher, line 220: The second arg isn't needed because it is the default value.

src/eventlistener.cpp
```diff
@@ -96,24 +122,32 @@ bool EventListener::checkOtherInstances(const QString& windowName) {
   }
 #endif
 
+  // Try to wake the UI and bring it to the foreground.
   logger.debug() << "Try to communicate with the existing instance";
```
vinocher, line 126: Should the old log be retained?
  logger.debug() << "Terminating the current process";

src/eventlistener.cpp
```diff
@@ -49,21 +51,45 @@ EventListener::EventListener() {
...
-      }
-    });
-  });
+  logger.debug() << "EventListener input:" << input;
+  QString command = input.section(' ', 0, 0);
+  QString payload = input.section(' ', 1, -1);
+
+  if (command == "show") {
+    QmlEngineHolder* engine = QmlEngineHolder::instance();
+    engine->showWindow();
```
vinocher, line 72: Consider extracting a method for this block that handles the link command.

src/eventlistener.cpp
```diff
@@ -49,21 +51,45 @@ EventListener::EventListener() {
...
+  } else if (command == "link") {
+    QUrl url(payload);
+
+    // We only accept the mozilla-vpn scheme.
+    if (url.scheme() != Constants::DEEP_LINK_SCHEME) {
+      return;
```
vinocher, line 78: Typos

src/eventlistener.h
```diff
@@ -17,6 +17,11 @@ class EventListener final : public QObject {
   ~EventListener();
 
   static bool checkOtherInstances(const QString& windowTitle);
+  static bool sendCommand(const QString& data);
+  static bool sendDeepLink(const QUrl& url);
```
vinocher, line 21: Should these be moved to private?

src/commands/commandui.cpp
```diff
@@ -172,6 +168,15 @@ int CommandUI::run(QStringList& tokens) {
     // If there is another instance, the execution terminates here.
     if (!EventListener::checkOtherInstances(
             I18nStrings::instance()->t(I18nStrings::ProductName))) {
+      // If we are given URL parameters, send them to the UI socket and exit.
```
vinocher, line 171: Consider extracting a method for this block and sharing with the code that executes when another instance of the app is not found.

src/commands/commandui.cpp
```diff
@@ -172,6 +168,15 @@ int CommandUI::run(QStringList& tokens) {
...
+      for (const QString& value : tokens) {
```
vinocher, line 172: Is the "ui" parameter in `mozillavpn ui %u` being checked for and verified?

src/eventlistener.cpp
```diff
@@ -96,24 +122,32 @@ bool EventListener::checkOtherInstances(const QString& windowName) {
   }
 #endif
 
+  // Try to wake the UI and bring it to the foreground.
```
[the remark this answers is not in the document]
oskirby: I agree, I found this method's return values confusing. I will update it accordingly.

src/eventlistener.h
oskirby: Sure

src/frontend/navigator.cpp
oskirby: Done

src/eventlistener.cpp
oskirby: Should be fixed now

src/commands/commandui.cpp
oskirby: The handling of this command parameter takes place in CommandLineParser::parse()

src/eventlistener.cpp
```diff
@@ -49,21 +51,44 @@ EventListener::EventListener() {
     QLocalSocket* socket = m_server.nextPendingConnection();
     Q_ASSERT(socket);
 
-    connect(socket, &QLocalSocket::readyRead, socket, [socket]() {
-      QByteArray input = socket->readAll();
-      input = input.trimmed();
+    connect(socket, &QLocalSocket::readyRead, this,
+            &EventListener::socketReadyRead);
+    connect(socket, &QLocalSocket::disconnected, socket, &QObject::deleteLater);
```
vinocher, line 56: My understanding is that the socket is being deleted to free memory, although it would have been deleted when the parent QLocalServer was destroyed. Is that correct? Are we sure that the socket doesn't need to be used after it is disconnected (for example, to check connected state)? I suppose deleteLater will prevent most such cases.

Will the following be afffected by the socket being prematurely deleted on disconnect?

  `socket.disconnectFromServer();
  if (socket.state() != QLocalSocket::UnconnectedState) {
    socket.waitForDisconnected(1000);
  }``

src/commands/commandui.cpp
vinocher: Never mind, the differences in the code blocks may not make it worth it.

src/eventlistener.cpp
```diff
@@ -49,21 +51,44 @@ EventListener::EventListener() {
...
+  });
+}
 
-      logger.debug() << "EventListener input:" << input;
+void EventListener::socketReadyRead() {
+  QObject* obj = QObject::sender();
+  QLocalSocket* socket = qobject_cast<QLocalSocket*>(obj);
```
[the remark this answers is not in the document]
oskirby: It's a stylistic preference I guess, I am not a fan of lambda-capturing values, especially pointers as they can lead to use-after-free bugs and escape the ownership tracking of `QObjects`. Though, in this particular case it would have been fine.

src/eventlistener.cpp
oskirby: The leak seems to have been a bug in the code that existed before I got here, in particular the `QLocalServer` lives for the duration of the client UI instance, but the `QLocalSocket` connections are made for each message sent into the pipe. Failing to destroy them upon disconnection leaves them in a disconnected state that hang around effectively indefinitely and waste resources.

The `socket.disconnectFromServer()` function you mentioned is a separate socket in a separate process (the sender of the command). It will not be affected by the garbage collection of a disconnected socket in the receiver process.

src/eventlistener.cpp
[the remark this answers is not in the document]
oskirby: This could be made a bit tighter though by ensuring that the sender really is a `QLocalSocket`. I'll add a check for that.

src/eventlistener.cpp
oskirby: Done.

vinocher (approved):
Looks good!

PR #8764, opened 2023-12-06
Chore: Unify Wix installer files

oskirby:
## Description
We have three copies of the Wix installer files floating around in our repository, which makes it kind of a pain to make changes. Especially considering all three of these files are identical except for the relative path to the built artifacts. We should really just have one source of truth used by all places. This PR attempts to do just that.

The three copies were as follows:
- `MozillaVPN.wxs`: The Wix installer file used by QMake, not actually used by anthing that I can find.
- `MozillaVPN_cmake.wxs`: The Wix installer file used by CMake whenever you build the `msi` target.
- `MozillaVPN_prod.wxs`: The Wix installer file used by the `repackage-msi` task.

## Reference
This bothered me while writing #8762 so I decided to fix it.

## Checklist
    

strseb (approved):
Oh my god, thank you!

PR #8772, opened 2023-12-08
ci: Remove unused fetch-open-ssl task

ahal:
The `fetch-open-ssl` task [is failing](https://firefox-ci-tc.services.mozilla.com/tasks/J9O4fWO3T2-Nn6mT6Sq_Hw/runs/0/logs/public/logs/live.log) after rebuilding cached tasks for a CoT key rotation. To be clear, the failure has nothing to do with the rotation, the rotation is just surfacing a latent error that would have popped up sooner or later anyway.

In this case I can confirm that downloading the file at the requested url yields the same hash / size that the error is complaining about... But I'm confused why the new file is so much smaller than the old one. Something seems suspicious here.

### DIFF
--- a/taskcluster/ci/fetch/kind.yml
+++ b/taskcluster/ci/fetch/kind.yml
@@ -34,14 +34,6 @@ tasks:
             size: 158837368
         #artifact-prefix: vpn/win-perl
         fetch-alias: win-perl
-    open-ssl:
-        description: Open SSL Source
-        fetch:
-            type: static-url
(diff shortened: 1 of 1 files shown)


ahal:
@strseb what do you make of the difference in file size here? If this passes all CI should we land it anyway? Or should we dig into why there's such a big difference

ahal:
Actually... it looks like this fetch task might be unused, can we delete it instead?

gabrielBusta (approved):

strseb (approved):

PR #8786, opened 2023-12-11
Nit: Put Decision Tasks for branches on the TC index

strseb:
## Description

We have a bit of tooling that tries to pull specific taskcluster jobs by branch. i.e https://mozilla-mobile.github.io/mozilla-vpn-client/
However it's a bit more complex right now then it need's to. It pulls on the gh api the branches, the runs from each branch and from there tries to find the task group. From that it pulls the wasm task and renders it's result. 

We could simplify that and announce the latest pr decision task on the index, which would allow me to instantly get the taskgroup of a certain branch 😅

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -175,7 +175,8 @@ tasks:
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.latest.taskgraph.decision-${cron.job_name}
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.revision.${head_sha}.taskgraph.decision-${cron.job_name}
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.revision.${head_sha}.taskgraph.cron.${ownTaskId}
-
+                else:
+                  - index.mozillavpn.v2.${project}.branch.${short_head_branch}.latest.taskgraph.decision
           scopes:
             $if: 'tasks_for == "github-push"'
             then:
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -175,7 +175,8 @@ tasks:
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.latest.taskgraph.decision-${cron.job_name}
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.revision.${head_sha}.taskgraph.decision-${cron.job_name}
                       - index.mozillavpn.v2.${project}.branch.${head_branch}.revision.${head_sha}.taskgraph.cron.${ownTaskId}
-
+                else:
+                  - index.mozillavpn.v2.${project}.branch.${head_branch}.latest.taskgraph.decision
```
ahal, line 179: I think this might cause Decision tasks from pull-requests to overwrite Decision tasks from pushes. I think we need to namespace these out to make sure they stay separate from the push ones, or it could cause issues.

Maybe something like `index.mozillavpn.v2.${project}.pr.${short_head_branch}.latest.taskgraph.decision`?

Also whatever we choose, we're going to need to update `ci-configuration` to grant the scope necessary to add this to the index.

ahal (changes requested):

.taskcluster.yml
hneiva: @ahal According to the previous error, `${short_head_branch}` isn't available.

ahal: Ah, we need to define it then like this:
https://github.com/mozilla-mobile/firefox-android/blob/main/.taskcluster.yml#L142

If using `head_branch` *just works*, then maybe that's good enough, but this value comes from Github and it appears it can sometimes contain `refs/heads` (I'm not sure exactly under what circumstances though).

hneiva (changes requested):
@strseb also, can you please change the branch name to something more descriptive?

Issue #8016, opened 2023-09-14
[Linux] App exclusions are not working on Ubuntu 23.04 (Lunar) VM


data-sync-user:
**Important details:**

* Not reproducing on Linux Jammy;

**Mozilla VPN used versions:**

* 2.17.0 (2.202309131732);

**Platform tested:**

* desktop;

**Prerequisites:**

* Have a clean install of the Mozilla VPN;
* Have an account with subscription for Mozilla VPN;
* Have the VPN off;

**Steps to reproduce:**

1. Go to Settings -> App exclusions;
1. Search for the Firefox browser and check it;
1. Turn the VPN ON;
1. In Firefox access an IP check page ;
1. Observe the displayed IP address and location;

**Expected result:**

* User real IP address is displayed in the Firefox browser;

**Actual result:**

* The VPN IP address is displayed even if the browser is excluded from VPN protection;

**Notes:**

* Not having a physical machine to confirm for Lunar;
* Follow up from [https://mozilla-hub.atlassian.net/browse/VPN-1990](https://mozilla-hub.atlassian.net/browse/VPN-1990|smart-link)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5535)
┆Reporter: Valentina Virlics

data-sync-user:
➤ Lesley Norton commented:

Naomi Kirby thoughts here? Would this be expected to work running in a Lunar VM?

data-sync-user:
➤ Naomi Kirby commented:

Yes, it is expected that app exclusions should work within a VM.

data-sync-user:
➤ Naomi Kirby commented:

I think the root cause here is that our application matching logic has always been a little bit flaky, and it requires a bit of fragile state to be carried between the GTK launch event and the cgroup path detection, and this has finally broken on us in Ubuntu Lunar. It seems that something in the SNAP launch process is terminating the root PID by the time the cgroup gets created so we never actually associated a cgroup with its desktop file.

I think the path to fixing this is to abandon the use of the GTK launch event and use a heuristic matching between the cgroup path and the desktop file.

This is really the same underlying problem as https://mozilla-hub.atlassian.net/browse/VPN-4798 ( https://mozilla-hub.atlassian.net/browse/VPN-4798|smart-link ), so that will likely get fixed too.

data-sync-user:
➤ Naomi Kirby commented:

[...]

Issue #6917, opened 2023-05-16
Linux App exclusions lose track of running applications when daemon is restarted


data-sync-user:
On the Linux platform. App exclusions require a mapping between control groups and the running applications. This mapping is established by capturing launch events from the desktop environment and correlating them against the `.desktop` file ID. This approach carries with it an obvious flaw: When the daemon is restarted (eg: by updating the VPN software) it looses this mapping and can no longer apply app exclusions until the apps are restarted as well.

**Affected devices:**

* Ubuntu 20.04 and later;
* Fedora 36 and later;

**Prerequisites:**

* Have a subscription made for Mozilla VPN;
* Have Mozilla VPN installed;
* Have Mozilla VPN closed and OFF;

**Steps to reproduce:**

1. Start a Firefox browser session.
1. Update the Mozilla VPN client by re-installing the `.deb` file or restart the daemon with `sudo systemctl restart mozillavpn`
1. Open the VPN and sign-in if required;
1. Navigate to Settings > App Permissions;
1. Check the Firefox browser;
1. Turn the VPN ON;
1. In Firefox access the[https://whatismyipaddress.com/](https://whatismyipaddress.com/|smart-link)
1. Observe the displayed IP address and location;

**Expected result:**

* The real IP address is displayed;

**Actual result:**

* The VPN IP address is displayed even if the browser is checked in app permission;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-4798)
┆Reporter: Naomi Kirby

data-sync-user:
➤ Betty Fleming commented:

Not 2.15 blocker. Naomi Kirby do you want to weigh in on priority?

data-sync-user:
➤ Valentina Virlics commented:

Verified this while using Mozilla VPN v. 2.20.0 (2.20.0~build20231218) across Linux 20.04, 22.04, 23.04 and 23.10 versions.

App exclusion work as expected → checking the Firefox browser in the app exclusions list inside de VPN client triggers the real IP to be displayed in the Firefox browser.

!image-20231219-121249.png|width=1364,height=628!

PR #8801, opened 2023-12-12
VPN-4798: Linux cgroup application matching heuristics

oskirby:
## Description
App exclusions on Linux mostly make use of the GTK `Launched` signal over D-Bus to kick off application detection by connecting a `*.desktop` file and a control group, and this has worked okay-ish for the most part, but it leaves a lot of edge cases where applications can fall through the cracks if the PID that we get from the GTK launch event dies before the following control group is created. And it just so happens that this has become the case for Snap-packaged applications as of Ubuntu/23.04 Lunar.

To fix this, I think it's finally time that we get off our butts and do some proper heuristic matching between the control group name and typical launcher commands used for Linux desktop environments. To make this work, we first convert the `*.desktop` file into a desktop file ID and then try to parse a matching desktop file ID out of the cgroup path name.

So far this supports applications launched via the Gnome shell launcher, Snaps and Flatpaks. Perhaps with works with more desktop environments too? Only testing will tell!

Platforms tested thus far:
- Ubuntu 20.04 native apps
- Ubuntu 20.04 snaps
- Ubuntu 23.04 native apps
- Ubuntu 23.04 snaps
- Ubuntu 23.04 flatpaks
- Debian 12 native apps on Gnome shell
- Debian 12 native apps on KDE Plasma
- Fedora 38 native apps
- Fedora 38 flatpaks

## Reference
Github issue #6917 ([VPN-4798](https://mozilla-hub.atlassian.net/browse/VPN-4798))
Github issue #8016 ([VPN-5535](https://mozilla-hub.atlassian.net/browse/VPN-5535))
Freedesktop spec: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id

## Checklist
    
[VPN-4798]: https://mozilla-hub.atlassian.net/browse/VPN-4798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/daemon/apptracker.cpp
+++ b/src/platforms/linux/daemon/apptracker.cpp
@@ -40,9 +40,6 @@ AppTracker::~AppTracker() {
   MZ_COUNT_DTOR(AppTracker);
   logger.debug() << "AppTracker destroyed.";
 
-  for (AppData* data : m_runningApps) {
-    delete data;
-  }
   m_runningApps.clear();
 }
 
@@ -122,16 +119,15 @@ QString AppTracker::decodeUnicodeEscape(const QString& str) {
 }
 
 // Make an attempt to resolve the desktop ID from a cgroup scope.
-void AppTracker::appHeuristicMatch(AppData* data) {
-  QString scopeName = QFileInfo(data->cgroup).fileName();
+QString AppTracker::findDesktopId(const QString& cgroup) {
+  QString scopeName = QFileInfo(cgroup).fileName();
 
   // Reverse the desktop ID from a cgroup scope and known launcher tools.
   if (scopeName.startsWith("app-gnome-") ||
       scopeName.startsWith("app-flatpak-")) {
     QString raw = scopeName.section('-', 2, 2);
     if (!raw.isEmpty()) {
-      data->desktopId = QString("%1.desktop").arg(decodeUnicodeEscape(raw));
-      return;
+      return QString("%1.desktop").arg(decodeUnicodeEscape(raw));
     }
   }
 
@@ -140,8 +136,7 @@ void AppTracker::appHeuristicMatch(AppData* data) {
     qsizetype start = gnomeLaunchdPrefix.length();
     qsizetype end = scopeName.lastIndexOf('-');
     if (end > start) {
-      data->desktopId = scopeName.mid(start, end - start);
-      return;
+      return scopeName.mid(start, end - start);
     }
   }
 
@@ -157,8 +152,11 @@ void AppTracker::appHeuristicMatch(AppData* data) {
                            this);
   QString source = interface.property("SourcePath").toString();
   if (!source.isEmpty() && source.endsWith(".desktop")) {
-    data->desktopId = LinuxDependencies::desktopFileId(source);
+    return LinuxDependencies::desktopFileId(source);
   }
+
+  // Otherwise, we don't know the desktop ID for this control group.
+  return QString();
 }
 
 void AppTracker::cgroupsChanged(const QString& directory) {
@@ -179,12 +177,10 @@ void AppTracker::cgroupsChanged(const QString& directory) {
     if (oldScopes.removeAll(path) == 0) {
       // This is a new scope, let's add it.
       logger.debug() << "Control group created:" << path;
-      AppData* data = new AppData(path);
+      QString desktopId = findDesktopId(path);
+      m_runningApps[path] = desktopId;
 
-      m_runningApps[path] = data;
-      appHeuristicMatch(data);
-
-      emit appLaunched(data->cgroup, data->desktopId);
+      emit appLaunched(path, desktopId);
     }
   }
 
@@ -194,10 +190,9 @@ void AppTracker::cgroupsChanged(const QString& directory) {
     if (scopeInfo.absolutePath() == directory) {
       logger.debug() << "Control group removed:" << scope;
       Q_ASSERT(m_runningApps.contains(scope));
-      AppData* data = m_runningApps.take(scope);
+      QString desktopId = m_runningApps.take(scope);
 
-      emit appTerminated(data->cgroup, data->desktopId);
-      delete data;
+      emit appTerminated(scope, desktopId);
     }
   }
 }
--- a/src/platforms/linux/daemon/apptracker.cpp
+++ b/src/platforms/linux/daemon/apptracker.cpp
@@ -125,7 +125,7 @@ QString AppTracker::decodeUnicodeEscape(const QString& str) {
 //
 // However, at some point the separator between the app and UUID was
 // swapped from a dot to a dash. Which makes the parsing a bit of a pain.
-// 
+//
 // See: https://github.com/snapcore/snapd/blob/master/sandbox/cgroup/scanning.go
 QString AppTracker::snapDesktopId(const QString& scope) {
   static const QRegularExpression snapuuid(
--- a/src/platforms/linux/daemon/apptracker.cpp
+++ b/src/platforms/linux/daemon/apptracker.cpp
@@ -107,7 +107,7 @@ QString AppTracker::decodeUnicodeEscape(const QString& str) {
     bool okay;
     qsizetype start = match.capturedStart(0);
     QChar code = match.captured(0).mid(2).toUShort(&okay, 16);
-    if (okay) {
+    if (okay && (code != 0)) {
       result.replace(start, match.capturedLength(0), QString(code));
       offset = start + 1;
     } else {
--- a/src/platforms/linux/daemon/apptracker.cpp
+++ b/src/platforms/linux/daemon/apptracker.cpp
@@ -235,24 +235,3 @@ void AppTracker::cgroupsChanged(const QString& directory) {
     }
   }
 }
-
-QList<int> AppTracker::pids(const QString& cgroup) const {
-  QList<int> results;
-  QFile cgroupProcs(m_cgroupMount + cgroup + "/cgroup.procs");
-
-  if (cgroupProcs.open(QIODevice::ReadOnly | QIODevice::Text)) {
-    while (true) {
-      QString line = QString::fromLocal8Bit(cgroupProcs.readLine());
-      if (line.isEmpty()) {
-        break;
-      }
-      int pid = line.trimmed().toInt();
-      if (pid != 0) {
-        results.append(pid);
-      }
-    }
-    cgroupProcs.close();
-  }
-
-  return results;
-}
--- a/src/platforms/linux/daemon/apptracker.cpp
+++ b/src/platforms/linux/daemon/apptracker.cpp
@@ -121,6 +121,22 @@ void AppTracker::gtkLaunchEvent(const QByteArray& appid, const QString& display,
   }
 }
 
+// static
+// Expand unicode escape sequences.
+QString AppTracker::decodeUnicodeEscape(const QString& str) {
+  static const QRegularExpression re("\\x[0-9A-Za-z][0-9A-Za-z]");
+
+  QString result = str;
+  QRegularExpressionMatch match = re.match(result);
+  for (int i = match.lastCapturedIndex(); i > 0; i--) {
+    bool okay;
+    QChar code = match.captured(i).mid(2).toUShort(&okay, 16);
+    result.replace(match.capturedStart(i), match.capturedLength(i),
+                   QString(code));
+  }
+  return result;
+}
+
 void AppTracker::appHeuristicMatch(AppData* data) {
   // If this cgroup contains the last-launched PID, then we have a fairly
   // strong indication of which application this control group is running.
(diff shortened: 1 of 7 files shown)


src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -104,59 +91,116 @@ void AppTracker::userRemoved(uint userid) {
...
+// The naming convention for snaps follows one of the following formats:
+//   snap.<pkg>.<app>.service - assigned by systemd for services
+//   snap.<pkg>.<app>-<uuid>.scope - transient scope for apps
+//   snap.<pkg>.hook.<app>-<uuid>.scope - transient scope for hooks
+//
+// However, at some point the separator between the app and UUID was
+// swapped from a dot to a dash. Which makes the parsing a bit of a pain.
+//
+// See: https://github.com/snapcore/snapd/blob/master/sandbox/cgroup/scanning.go
+QString AppTracker::snapDesktopId(const QString& scope) {
```
vinocher, line 130: Should these string manipulation functions be moved to LinuxDependencies or some String utils file since they don't need any class state? Also, consider adding some tests for the string manipulation functions, now or later.

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -224,24 +221,28 @@ void DBusService::userRemoved(uint uid, const QDBusObjectPath& path) {
...
-void DBusService::appTerminated(const QString& cgroup, const QString& appId) {
-  logger.debug() << "terminate:" << cgroup;
+void DBusService::appTerminated(const QString& cgroup,
+                                const QString& desktopId) {
+  logger.debug() << "terminate:" << cgroup << "id:" << desktopId;
 
-  // HACK: Quick and dirty split tunnelling.
-  // TODO: Apply filtering to currently-running apps too.
-  if (m_excludedApps.contains(appId)) {
+  // Remove any firewall rules applied to this control group.
```
vinocher, line 244: Shouldn't this app also be removed from m_excludedApps?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -250,54 +251,70 @@ void DBusService::appTerminated(const QString& cgroup, const QString& appId) {
...
+    if (state == Excluded) {
+      // Excluded control groups are given special netfilter rules to direct
+      // their traffic outside of the VPN tunnel.
+      m_wgutils->excludeCgroup(cgroup);
+    }
+  }
+}
+
 /* Update the firewall for running applications matching the application ID. */
 bool DBusService::firewallApp(const QString& appName, const QString& state) {
```
vinocher, line 297: Is firewallApp() called anymore?

vinocher:
> cracks if the PID that we get from the GTK launch event dies before the following control group is created

Trying to understand this better:  Why does this always happen for snap-packaged apps? It sounds like this is not an edge case for snap apps, is that right?

src/platforms/linux/linuxdependencies.cpp
```diff
@@ -156,3 +156,31 @@ QString LinuxDependencies::kdeFrameworkVersion() {
 
   return QString();
 }
+
+// static
+QString LinuxDependencies::desktopFileId(const QString& path) {
```
vinocher, line 161: This method seems to return a non-empty string even if '/applications/' is not in the path. Should it try to detect error conditions and return an empty string in those cases, or is that not possible? Callers seem to pass on the result to other methods without additional checks.

src/platforms/linux/daemon/apptracker.h
```diff
@@ -36,23 +24,26 @@ class AppTracker final : public QObject {
...
 
-  QHash<QString, AppData*>::iterator begin() { return m_runningApps.begin(); }
-  QHash<QString, AppData*>::iterator end() { return m_runningApps.end(); }
-  AppData* find(const QString& cgroup) { return m_runningApps.value(cgroup); }
+  QStringList getRunningApps() const { return m_runningApps.keys(); }
+  QString getDesktopId(const QString& cgroup) const {
+    return m_runningApps.value(cgroup);
+  }
+  QStringList findByDesktopId(const QString& desktopId) const {
+    return m_runningApps.keys(desktopId);
```
vinocher, line 32: Is the cgroup path <--> desktopId mapping 1:1? Or can there be multiple cgroup paths that map to a desktopId?

src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -172,32 +216,29 @@ void AppTracker::cgroupsChanged(const QString& directory) {
...
 
-      emit appTerminated(data->cgroup, data->appId);
-      delete data;
+      emit appTerminated(scope, desktopId);
     }
   }
 }
 
-QList<int> AppData::pids() const {
+QList<int> AppTracker::pids(const QString& cgroup) const {
```
vinocher, line 239: Is this method called anymore?

src/platforms/linux/daemon/dbusservice.h
```diff
@@ -24,7 +25,11 @@ class DBusService final : public Daemon, protected QDBusContext {
   DBusService(QObject* parent);
   ~DBusService();
 
+  enum AppState { Active, Excluded, Blocked };
```
vinocher, line 28: Is this 'Blocked' state used?

src/platforms/linux/daemon/dbusservice.cpp
```diff
@@ -250,54 +251,70 @@ void DBusService::appTerminated(const QString& cgroup, const QString& appId) {
...
+  for (const QString& cgroup : m_appTracker->findByDesktopId(desktopId)) {
+    if (m_excludedCgroups.contains(cgroup)) {
+      m_wgutils->resetCgroup(cgroup);
+    }
+    m_excludedCgroups[cgroup] = state;
+    if (state == Excluded) {
+      // Excluded control groups are given special netfilter rules to direct
+      // their traffic outside of the VPN tunnel.
+      m_wgutils->excludeCgroup(cgroup);
+    }
```
vinocher, line 292: Should we check if the cgroup state is already in the expected state before doing this work?

src/platforms/linux/daemon/dbusservice.cpp
oskirby: It is a callable method via D-Bus, but at present the client doesn't make use of it.

src/platforms/linux/daemon/apptracker.h
oskirby: The desktopId -> cgroup mapping is one-to-many. Each cgroup matches at most one desktopId, but a desktopId can match multiple cgroups.

src/platforms/linux/daemon/apptracker.cpp
oskirby: No, I don't think so. I will remove it.

src/platforms/linux/daemon/dbusservice.cpp
oskirby: No, the gist of this, is that m_excludedApps holds the applications that the user wishes to exclude, so it kind of hangs around as a persistent setting, but m_excludedCgroups holds the matched applications that are currently running.

This signal occurs when the application closes, in which case the cgroup has been destroyed, but we still need to remember the setting in case the user re-opens it.

src/platforms/linux/linuxdependencies.cpp
oskirby: It's not entirely clear how to handle that edge case, but I wanted to try and err on the side of caution and make sure that we always return something suitable for cgroup matching. I guess taking the file name and using that might be a more suitable fallback for that case.

src/platforms/linux/daemon/dbusservice.h
oskirby: Not anymore. In the previous iteration of the Linux split tunneling feature, we had the ability to deny network access to applications entirely with this feature, and that's what this state did. But sometime last year we needed to migrate the split tunneling feature to cgroupsv2 and we lost that feature in the process but left it in the API.

This state is a vestigial reminder of that old code.

src/platforms/linux/daemon/dbusservice.cpp
oskirby: That felt like an unnecessary edge case to handle. If the cgroup is already in the expected state then this will simply remove the firewall rules for it, and then re-add them.

src/platforms/linux/daemon/dbusservice.cpp
vinocher: Should it be removed to reduce attack surface and for YAGNI?

oskirby:
> Trying to understand this better: Why does this always happen for snap-packaged apps? It sounds like this is not an edge case for snap apps, is that right?

I think this breakage occurs on snap because applications are launched as a two-stage process. The first step is that snaps create a `.desktop` file that the desktop environment can see, which contains a command like `/usr/bin/snap <target app>`. The GTK `Launched` event does successfully report the creation of this process, and we can split-tunnel it just fine, but that command is just a wrapper that creates a sandbox environment for the target app (including a new cgroup) and launches the target app within it. The detection of that new cgroup has always been brittle because it depends on matching the PID from the GTK `Launched` event and there is no real guarantee that PIDs will stay running for any length of time after that signal, which opens us up to race conditions.

So, the real fix happening here is the addition of the `AppTracker::snapDesktopId()` to try and figure out the desktopId of a snap without any dependency on the lifetime of the contained process IDs, and then converting the `AppTracker` to match apps by desktopId rather than the full file path. The rest of this PR is just trying to clean up and modernize some of my old code.

src/platforms/linux/daemon/dbusservice.cpp
oskirby: sure

src/platforms/linux/daemon/dbusservice.cpp
vinocher: Is there any risk of desktopId re-use that can impact m_excludedApps, because the ID is not removed when the app is terminated? For example, could an excluded app be uinstalled and another app installed with the same desktopId while the VPN is on? Or is this an unlikely edge case that is difficult to handle?

src/platforms/linux/daemon/dbusservice.h
oskirby: I have removed this state.

src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -104,59 +91,116 @@ void AppTracker::userRemoved(uint userid) {
...
-  // If this cgroup contains the last-launched PID, then we have a fairly
-  // strong indication of which application this control group is running.
-  for (int pid : data->pids()) {
-    if ((pid != 0) && (pid == m_lastLaunchPid)) {
-      logger.debug() << data->cgroup << "matches app:" << m_lastLaunchName;
-      data->appId = m_lastLaunchName;
-      data->rootpid = m_lastLaunchPid;
-      break;
+// Make an attempt to resolve the desktop ID from a cgroup scope.
+QString AppTracker::findDesktopId(const QString& cgroup) {
```
vinocher, line 158: Consider adding some examples of cgroup scope names, to make it easier for the reader to understand this method. Examples for various names that start with app-gnome-, app-flatpak-, gnome-launched- etc. would be helpful.

vinocher:
> I think this breakage occurs on snap because applications are launched as a two-stage process. The first step is that snaps create a `.desktop` file that the desktop environment can see, which contains a command like `/usr/bin/snap <target app>`. ...

Thanks for the explanation. It may be helpful to add some of this information as a class header comment in AppTracker.

src/platforms/linux/daemon/dbusservice.cpp
oskirby: I think this scenario would be commonly encountered if a user updates software on their device while the VPN is active, and I think the correct behavior in that case is to treat the newly-installed application identically to the previous version.

The recommendation in the freedesktop specification is for application names to follow a reverse-DNS naming convention, in which case they are expected to be a kind of globally unique identifier.

src/platforms/linux/daemon/apptracker.h
```diff
@@ -36,23 +24,25 @@ class AppTracker final : public QObject {
   void userCreated(uint userid, const QString& xdgRuntimePath);
   void userRemoved(uint userid);
 
-  QHash<QString, AppData*>::iterator begin() { return m_runningApps.begin(); }
-  QHash<QString, AppData*>::iterator end() { return m_runningApps.end(); }
-  AppData* find(const QString& cgroup) { return m_runningApps.value(cgroup); }
+  QStringList getRunningApps() const { return m_runningApps.keys(); }
+  QString getDesktopId(const QString& cgroup) const {
```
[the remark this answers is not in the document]
oskirby: They are no longer used after removing the D-Bus Firewall API methods. I will remove these helpers too.

src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -104,59 +91,116 @@ void AppTracker::userRemoved(uint userid) {
...
+    }
+
+    bool okay;
+    qsizetype start = match.capturedStart(0);
+    QChar code = match.captured(0).mid(2).toUShort(&okay, 16);
+    if (okay && (code != 0)) {
+      result.replace(start, match.capturedLength(0), QString(code));
+      offset = start + 1;
+    } else {
+      offset = match.capturedEnd(0);
```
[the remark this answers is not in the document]
oskirby: That's not a problem, the loop will terminate in such a case because the loop condition requires the offset to be contained within the string.

src/platforms/linux/daemon/apptracker.h
```diff
@@ -36,23 +24,25 @@ class AppTracker final : public QObject {
...
+    return m_runningApps.value(cgroup);
+  }
+  QStringList findByDesktopId(const QString& desktopId) const {
```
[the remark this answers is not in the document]
oskirby: Sure, I will rename things accordingly.

src/platforms/linux/daemon/apptracker.h
```diff
@@ -36,23 +24,25 @@ class AppTracker final : public QObject {
   void userCreated(uint userid, const QString& xdgRuntimePath);
   void userRemoved(uint userid);
 
-  QHash<QString, AppData*>::iterator begin() { return m_runningApps.begin(); }
-  QHash<QString, AppData*>::iterator end() { return m_runningApps.end(); }
-  AppData* find(const QString& cgroup) { return m_runningApps.value(cgroup); }
+  QStringList getRunningApps() const { return m_runningApps.keys(); }
```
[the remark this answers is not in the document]
oskirby: They are no longer used after removing the D-Bus Firewall API methods. I will remove these helpers too.

src/platforms/linux/daemon/apptracker.cpp
```diff
@@ -104,59 +91,116 @@ void AppTracker::userRemoved(uint userid) {
...
-    appIdName.chop(1);
+  QString result = str;
+  qsizetype offset = 0;
+  while (offset < result.length()) {
+    QRegularExpressionMatch match = re.match(result, offset);
+    if (!match.hasMatch()) {
+      break;
+    }
+
+    bool okay;
```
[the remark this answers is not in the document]
oskirby: Done

vinocher (approved):
Looks good! Comments are very helpful!

PR #8807, opened 2023-12-13
VPN-5984: Rename .desktop files into reverse-DNS format

oskirby:
## Description
We recently tried to add URL deeplinking on Linux, but it didn't quite work when launched from a snap-installed version of Firefox. It seems that snap requires the registered handler for MIME types have well-formed `*.desktop` file names, of which ours is not.

The official freedesktop specification recommends that desktop files are given names in reverse-DNS format, of which `mozillavpn.desktop` does not comply. This PR renames it to `org.mozilla.vpn.desktop`

## Reference
Github issue: #8743 ([VPN-5984](https://mozilla-hub.atlassian.net/browse/VPN-5984))
Snapcraft's URL handler check: [schemeHasHandler()](https://github.com/bboozzoo/snapd/blob/c3b4b3c749d5a9089b033cff6d9f6df38d454b9a/usersession/userd/launcher.go#L163-L175)
Freedesktop specification: [File naming](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming)

## Checklist
    
[VPN-5984]: https://mozilla-hub.atlassian.net/browse/VPN-5984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
Hmm, okay then, this PR actually changed nothing. Seems that both before and after this patch we get a system that can launch deep links from inside a snap sandbox (eg: `snap run --shell firefox -c "xdg-open mozilla-vpn://nav/gethelp"`) but I am unable to figure out why neither Firefox nor Chromium are able to detect the custom MIME type on Ubuntu 23.04.

PR #8808, opened 2023-12-13
VPN-4852: Implement MZBottomSheet and MZHelpSheet

MattLichtenstein:
## Description

- Create `MZBottomSheet`, a generic, reusable drawer/sheet component
- Create `MZHelpSheet`, based off of `MZBottomSheet`, but specifically tailored for our [educational content](https://mozilla-hub.atlassian.net/browse/VPN-4681)
- Notes about `MZHelpSheet`:
  - Resizes to fit it's content, up to a maximum size, then scrolls
  - Supports titles, text, primary buttons, and link buttons

Examples: 

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/735a2be2-860c-4bf4-b018-3b6e0f222d25

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/d60c5af8-f22e-4243-b2a6-8ccb66b3b57d

## Reference

[VPN-4852: Create new "MZSheet" Nebula component to support new guides](https://mozilla-hub.atlassian.net/browse/VPN-4852)

### DIFF
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -9,6 +9,23 @@ import QtQuick.Layouts 1.14
 import components 0.1
 import Mozilla.Shared 1.0
 
+//SUMMARY: MZBottomSheet opens a drawer from the bottom of the screen showing it's contentItem
+
+//USAGE:
+/*
+  MZBottomSheet {
+      contentItem: ColumnLayout {
+          Rectangle {
+              anchors.left: parent.left
+              anchors. right: parent.right
+              height: 20
+              color: "green"
+
+          }
+      }
+  }
+*/
+
 Drawer {
     id: drawer
 
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -8,7 +8,7 @@ import Mozilla.Shared 1.0
 Drawer {
     id: drawer
 
-    readonly property int maxHeight: parent.height - MZTheme.theme.sheetTopMargin //parent is the overlay
+    readonly property int maxHeight: window.height - MZTheme.theme.sheetTopMargin //parent is the overlay
 
     implicitWidth: window.width
     implicitHeight: maxHeight
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -8,7 +8,7 @@ import Mozilla.Shared 1.0
 Drawer {
     id: drawer
 
-    readonly property int maxHeight: window.height - MZTheme.theme.sheetTopMargin //parent is the overlay
+    readonly property int maxHeight: (Qt.platform.os === "ios" ? window.safeContentHeight : window.height) -  MZTheme.theme.sheetTopMargin
 
     implicitWidth: window.width
     implicitHeight: maxHeight
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 import QtQuick 2.15
 import QtQuick.Controls 2.14
 import QtQuick.Layouts 1.14
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -8,7 +8,7 @@ import Mozilla.Shared 1.0
 Drawer {
     id: drawer
 
-    readonly property int maxHeight: window.height - 32
+    readonly property int maxHeight: parent.height - MZTheme.theme.sheetTopMargin //parent is the overlay
 
     implicitWidth: window.width
     implicitHeight: maxHeight
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -1,0 +1,36 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import components 0.1
+import Mozilla.Shared 1.0
+
+Drawer {
+    id: drawer
+
+    readonly property int maxHeight: window.height - 32
+
+    implicitWidth: window.width
+    implicitHeight: maxHeight
+
+    topPadding: 0
+
+    dragMargin: 0
+    edge: Qt.BottomEdge
+    background: Rectangle {
+
+        radius: 8
+        color: MZTheme.theme.bgColor
+
+        Rectangle {
+            color: parent.color
+            anchors.bottom: parent.bottom
+            width: parent.width
+            height: parent.radius
+        }
+    }
+
+    Overlay.modal: Rectangle {
+        color: MZTheme.theme.overlayBackground
+    }
+}
--- a/nebula/ui/components/MZFlickable.qml
+++ b/nebula/ui/components/MZFlickable.qml
@@ -14,7 +14,7 @@ Flickable {
     property var flickContentHeight
     property bool contentExceedsHeight: height < contentHeight
     property bool hideScollBarOnStackTransition: false
-    property bool navbarVisible: navbar.visible
+    property bool addNavbarHeightOffset: navbar.visible
 
     //This property should be true if the flickable appears behind the main navbar
     interactive: contentHeight > height || contentY > 0
@@ -95,13 +95,13 @@ Flickable {
 
         const isItemBehindNavbar = item.mapToItem(window.contentItem, 0, 0).y + item.height > window.height - MZTheme.theme.navBarHeightWithMargins
         const isItemOffScreen = item.mapToItem(window.contentItem, 0, 0).y + item.height > window.height
-        const buffer = vpnFlickable.navbarVisible && isItemBehindNavbar ? MZTheme.theme.navBarHeightWithMargins : MZTheme.theme.contentBottomMargin
+        const buffer = vpnFlickable.addNavbarHeightOffset && isItemBehindNavbar ? MZTheme.theme.navBarHeightWithMargins : MZTheme.theme.contentBottomMargin
         const itemHeight = Math.max(item.height, MZTheme.theme.rowHeight) + buffer
         let destinationY
 
         //When focusing on an item either behind the navbar and/or that is partially or fully off screen in the downward direction
         //scroll to it so that it is visible
-        if((vpnFlickable.navbarVisible && isItemBehindNavbar) || (!vpnFlickable.navbarVisible && isItemOffScreen)) {
+        if((vpnFlickable.addNavbarHeightOffset && isItemBehindNavbar) || (!vpnFlickable.addNavbarHeightOffset && isItemOffScreen)) {
             const scrollDistance = item.mapToItem(window.contentItem, 0, 0).y + item.height - (window.height - buffer)
 
             destinationY = contentY + scrollDistance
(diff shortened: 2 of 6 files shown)


nebula/ui/components/qmldir
```diff
@@ -75,3 +75,5 @@ MZVerticalSpacer 0.1 MZVerticalSpacer.qml
 MZViewBase 0.1 MZViewBase.qml
 MZGuideList 0.1 MZGuideList.qml
 MZTipsAndTricksSection 0.1 MZTipsAndTricksSection.qml
+MZHelpSheet 0.1 helpSheet/MZHelpSheet
```
vinocher, line 78: Should the helpSheet dir be removed?

nebula/ui/components/MZBottomSheet.qml
```diff
@@ -0,0 +1,40 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.15
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import components 0.1
+import Mozilla.Shared 1.0
+
```
vinocher, line 11: Why is safeContentHeight only used on iOS and not other platforms?

nebula/ui/components/MZBottomSheet.qml
```diff
@@ -0,0 +1,40 @@
...
+    id: drawer
+
+    readonly property int maxHeight: (Qt.platform.os === "ios" ? window.safeContentHeight : window.height) -  MZTheme.theme.sheetTopMargin
+
+    implicitWidth: window.width
+    implicitHeight: maxHeight
+
+    topPadding: 0
+
+    dragMargin: 0
```
vinocher, line 22: Should this constant be moved into theme.js? Are we deliberately not using the standard  MZTheme.theme.cornerRadius value here?

nebula/ui/components/MZBottomSheet.qml
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.15
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
```
vinocher, line 8: Should we add a summary header comment to new components describing major features and usage?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+          {type: MZHelpSheet.BlockType.PrimaryButton, text: "Primary", margin: 16, action: () => { sheet.close(); MZNavigator.requestScreen(VPN.ScreenGetHelp) } },
+          {type: MZHelpSheet.BlockType.LinkButton, text: "Link", margin: 8, action: () => { MZUrlOpener.openUrl("https://mozilla.org") } }
+      ]
+    }
+*/
+
+MZBottomSheet {
+    id: bottomSheet
+
+    property alias iconSource: icon.source
```
vinocher, line 38: Nit: Extra line

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+    property alias title: title.text
+    property var model
+
+
+    enum BlockType {
+        Title,
+        Text,
+        PrimaryButton,
```
vinocher, line 46: Just wondering, is there a good QML practice to hint that this maxHeight property is defined in MZBottomSheet? As a reader, I find myself looking for definitions that are not in the file. For example, is it good practice to use bottomSheet.maxHeight here?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+        LinkButton
+    }
+
+    implicitHeight: Math.min(contentItem.implicitHeight, maxHeight)
+
+    contentItem: ColumnLayout {
+        spacing: 0
+
```
vinocher, line 54: Should the constants in this file be moved to theme.js?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+                Item {
+                    Layout.fillWidth: true
+                }
+
+                MZIconButton {
+                    id: iconButton
+
+                    Layout.rightMargin: MZTheme.theme.windowMargin / 2
+
```
vinocher, line 104: Should this line be removed?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+                    Layout.preferredHeight: MZTheme.theme.rowHeight
+                    Layout.preferredWidth: MZTheme.theme.rowHeight
+
+                    //                skipEnsureVisible: true // prevents scrolling of lists when this is focused
+
```
vinocher, line 109: This line should not be needed because MZButtonBase has it.

nebula/ui/components/MZFlickable.qml
```diff
@@ -14,6 +14,8 @@ Flickable {
     property var flickContentHeight
     property bool contentExceedsHeight: height < contentHeight
     property bool hideScollBarOnStackTransition: false
+    property bool navbarVisible: navbar.visible
```
vinocher, line 17: Is this property alias required although navbar is a global component?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+
+        MZFlickable {
+            id: flickable
+
+            readonly property int maxheight: bottomSheet.maxHeight - headerLayout.implicitHeight - headerLayout.Layout.topMargin
+
+            Layout.fillWidth: true
+            Layout.preferredHeight: Math.min(layout.implicitHeight + layout.anchors.topMargin + layout.anchors.bottomMargin, maxheight)
```
vinocher, line 142: When will navbar.visible be set to true after the Drawer is closed?

vinocher:
Were you able to solve the Accessibility issues in the Drawer?

nebula/ui/components/MZBottomSheet.qml
```diff
@@ -0,0 +1,40 @@
...
+import components 0.1
+import Mozilla.Shared 1.0
+
+Drawer {
+    id: drawer
+
+    readonly property int maxHeight: (Qt.platform.os === "ios" ? window.safeContentHeight : window.height) -  MZTheme.theme.sheetTopMargin
```
vinocher, line 15: Why is safeContentHeight only used on iOS and not other platforms?

nebula/ui/components/MZBottomSheet.qml
```diff
@@ -0,0 +1,40 @@
...
+    implicitWidth: window.width
+    implicitHeight: maxHeight
+
+    topPadding: 0
+
+    dragMargin: 0
+    edge: Qt.BottomEdge
+    background: Rectangle {
+
+        radius: 8
```
vinocher, line 26: Should this constant be moved into theme.js? Are we deliberately not using the standard MZTheme.theme.cornerRadius value here?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+*/
+
+MZBottomSheet {
+    id: bottomSheet
+
+    property alias iconSource: icon.source
+    property alias title: title.text
+    property var model
+
+
```
vinocher, line 42: Nit: Extra line

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+    implicitHeight: Math.min(contentItem.implicitHeight, maxHeight)
+
+    contentItem: ColumnLayout {
+        spacing: 0
+
+        ColumnLayout {
+            id: headerLayout
+
+            Layout.topMargin: 8
```
vinocher, line 58: Should the constants in this file be moved to theme.js?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+
+                MZIconButton {
+                    id: iconButton
+
+                    Layout.rightMargin: MZTheme.theme.windowMargin / 2
+
+                    Layout.preferredHeight: MZTheme.theme.rowHeight
+                    Layout.preferredWidth: MZTheme.theme.rowHeight
+
+                    //                skipEnsureVisible: true // prevents scrolling of lists when this is focused
```
vinocher, line 108: Should this line be removed?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+
+                    onClicked: bottomSheet.close()
+
+                    accessibleName: MZI18n.GlobalClose
+                    Accessible.ignored: !visible
```
vinocher, line 113: This line should not be needed because MZButtonBase has Accessible.ignored: !visible

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+            id: flickable
+
+            readonly property int maxheight: bottomSheet.maxHeight - headerLayout.implicitHeight - headerLayout.Layout.topMargin
+
+            Layout.fillWidth: true
+            Layout.preferredHeight: Math.min(layout.implicitHeight + layout.anchors.topMargin + layout.anchors.bottomMargin, maxheight)
+
+            flickContentHeight: layout.implicitHeight + layout.anchors.topMargin + layout.anchors.bottomMargin
+
+            navbarVisible: false
```
vinocher, line 146: When will navbar.visible be set to true after the Drawer is closed?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,240 @@
...
+
+    enum BlockType {
+        Title,
+        Text,
+        PrimaryButton,
+        LinkButton
+    }
+
+    implicitHeight: Math.min(contentItem.implicitHeight, maxHeight)
+
```
vinocher, line 51: Just wondering, is there a good QML practice to hint that this maxHeight property is defined in MZBottomSheet? As a reader, I find myself looking for definitions that are not in the file. For example, is it good practice to use bottomSheet.maxHeight here?

nebula/ui/components/MZFlickable.qml
MattLichtenstein: this property is used to tell the flickable whether or not to consider the navbar when calculating its `contentHeight` (eg do we need to be able to scroll more so that no content is stuck behind the navbar). Most of the time this is based on whether the navbar is actually visible or not, but in the case of these sheets that are overlayed on top of the screen, the navbar is still `visible == true`, but it is behind the sheet so does not need to be considered.

MattLichtenstein:
> Were you able to solve the Accessibility issues in the Drawer?

No 😞 it seems like an issue with the way macOS voiceover narrates content as it is not as much of an issue on iOS. The only solution I can think of would be to set `Accessible.ignored = true` for all components not within the drawer. 

Would be curious to know how this functions on Windows

nebula/ui/components/MZHelpSheet.qml
MattLichtenstein: In my opinion, not every integer literal needs naming. I think if there is no way to genericize an integer for reusability, theres not much of a point in adding it to theme.js

nebula/ui/components/MZBottomSheet.qml
MattLichtenstein: [see here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8808#discussion_r1428185741)

nebula/ui/components/MZHelpSheet.qml
MattLichtenstein: But I will go through this component and add constants where I see fit

nebula/ui/components/MZHelpSheet.qml
MattLichtenstein: Qt suggests always [referencing properties of parent components by using their id](https://doc.qt.io/qt-6/qml-codingconventions.html#unqualified-access), but does not mention referencing the the components own id. 

I get the confusion here because the property is defined in a different file, but it should be implied that if it is not defined in-line in the current file, then it would be part of it's super class

vinocher:
Is keyboard access to the Drawer's contents possible? (For example, can a user TAB through them?)

MattLichtenstein:
> Is keyboard access to the Drawer's contents possible? (For example, can a user TAB through them?)

Yes, keyboard navigation via tab doesn't exceed the bounds of the drawer

vinocher (approved):
Feature looks great!

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -0,0 +1,246 @@
...
+import QtQuick 2.15
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import components 0.1
+import Mozilla.Shared 1.0
+
+//SUMMARY: MZHelpSheet is a component derived from MZBottomSheet that opens a drawer from the bottom of the screen who's content
+//is always laid out in a consistnet way by passing in a model with the correct schema, and sizes to fit the content it contains
```
vinocher, line 13: > consistnet 

Typo

PR #8812, opened 2023-12-14
VPN-5994 Add inspector command to mock network connection being down

Gela:
## Description

Add inspector command `force_connection_down` to mock network connection being down.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5994

## Checklist
    
### DIFF
--- a/src/inspectornetworkwatcher.h
+++ b/src/inspectornetworkwatcher.h
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef INSPECTORNETWORKWATCHER_H
+#define INSPECTORNETWORKWATCHER_H
+
+#include "networkwatcherimpl.h"
+
+class InspectorNetworkWatcher final : public NetworkWatcherImpl {
+ public:
+  InspectorNetworkWatcher(QObject* parent);
+  ~InspectorNetworkWatcher();
+
+  void initialize() override;
+
+  NetworkWatcherImpl::TransportType getTransportType() override {
+    return TransportType_None;
+  };
+};
+
+#endif  // INSPECTORNETWORKWATCHER_H
--- a/src/networkwatcher.cpp
+++ b/src/networkwatcher.cpp
@@ -54,43 +54,25 @@ NetworkWatcher::~NetworkWatcher() { MZ_COUNT_DTOR(NetworkWatcher); }
 void NetworkWatcher::initialize() {
   logger.debug() << "Initialize";
 
+  NetworkWatcherImpl* impl = nullptr;
+
 #if defined(MZ_WINDOWS)
-  m_impl = new WindowsNetworkWatcher(this);
+  impl = new WindowsNetworkWatcher(this);
 #elif defined(MZ_LINUX)
-  m_impl = new LinuxNetworkWatcher(this);
+  impl = new LinuxNetworkWatcher(this);
 #elif defined(MZ_MACOS)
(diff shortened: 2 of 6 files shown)


src/networkwatcher.cpp
```diff
@@ -174,3 +175,31 @@ QString NetworkWatcher::getCurrentTransport() {
   return QString(metaEnum.valueToKey(type))
       .remove("TransportType_", Qt::CaseSensitive);
 }
+
+void NetworkWatcher::setImpl(NetworkWatcherImpl* networkWatcherImpl) {
+  logger.debug() << "setImpl";
+
+  m_impl = new InspectorNetworkWatcher(this);
```
strseb, line 182: You already created the InspectorNetworkWatcher and passed it into this function as an argument in the Command, so i'd suggest `m_impl = netWorkWatcherImpl`

src/networkwatcher.cpp
strseb, line 1: Nit: with your setImpl it would be cool if we can cleanup the constructor and remove the now duplicate code:) 
```
setImpl(new WindowsNetworkWatcher(this));
#elif defined(MZ_LINUX)
 setImpl( new LinuxNetworkWatcher(this));
#elif defined(MZ_MACOS)
...
#endif
```

strseb (changes requested):
Looks Great! Just 2 Comments :)

src/inspectornetworkwatcher.h
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef INSPECTORNETWORKWATCHER_H
+#define INSPECTORNETWORKWATCHER_H
+
+#include "networkwatcherimpl.h"
+
+class InspectorNetworkWatcher final : public NetworkWatcherImpl {
```
vinocher, line 10: Consider if names like InactiveNetworkWatcher or OfflineNetworkWatcher are better, instead of coupling it to use by Inspector.

src/networkwatcher.cpp
```diff
@@ -174,3 +175,31 @@ QString NetworkWatcher::getCurrentTransport() {
...
+             settingsHolder->captivePortalAlert();
+  m_reportUnsecuredNetwork = settingsHolder->unsecuredNetworkAlert();
+  if (m_active) {
+    m_impl->start();
+  }
+
+  connect(settingsHolder, &SettingsHolder::unsecuredNetworkAlertChanged, this,
+          &NetworkWatcher::settingsChanged);
+  connect(settingsHolder, &SettingsHolder::captivePortalAlertChanged, this,
+          &NetworkWatcher::settingsChanged);
```
vinocher, line 204: It should be possible to remove this if Basti's suggestion of using setImpl for the other watchers is followed, because NetworkWatcher::initialize will provide this functionality.

src/networkwatcher.cpp
strseb: Ohnoo sorry i completly missed that in my first review.  😭 
Now as this class creates and owns the NetworkWatcherImpl, it should also cleanup. 

So if m_impl is already containing an object it probably should delete that. Otherwise the old one may still send signals to the Watcher. 


so something like `if(m_impl) m_impl->deleteLater()` or putting m_impl in a smart pointer i.e `std::unique_ptr`

strseb (changes requested):

strseb (approved):

Gela:
Closing in the interest of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8834
Since GetTransportType() is deprecated as of Qt 6.3 so we should use reachability() instead.

PR #8819, opened 2023-12-15
Chore: Write test suite for SettingsHolder generated functions

brizental:
This is pre-work for my upcoming PR which actually completely refactors the Settings management.

The idea was to test the APIs as is, then change it all and check that the test are passing all the same.

Also, I removed an unused configuration for settings since I was at it.

### DIFF
--- a/tests/unit_tests/testsettingsholder.cpp
+++ b/tests/unit_tests/testsettingsholder.cpp
@@ -1,0 +1,120 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "testsettingsholder.h"
+
+#define SETTING(type, toType, getter, setter, remover, has, key, defaultValue, \
+                ...)                                                           \
+  void TestSettingsHolder::testGetSetCheckRemove_##getter() {                  \
+    SettingsHolder settingsHolder;                                             \
+                                                                               \
+    QSignalSpy spy(&settingsHolder, &SettingsHolder::getter##Changed);         \
+                                                                               \
+    QVariant v = QVariant(defaultValue);                                       \
+    switch (v.metaType().id()) {                                               \
+      case QMetaType::Bool: {                                                  \
+        QVERIFY(!settingsHolder.has());                                        \
+        QCOMPARE(QVariant(settingsHolder.getter()).toBool(), v.toBool());      \
+                                                                               \
+        bool expectedValue = !v.toBool();                                      \
+        settingsHolder.setter(QVariant(expectedValue).toType());               \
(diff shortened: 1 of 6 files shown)


tests/unit_tests/testsettingsholder.cpp
```diff
@@ -0,0 +1,120 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "testsettingsholder.h"
+
+#define SETTING(type, toType, getter, setter, remover, has, key, defaultValue, \
```
Gela, line 7: is the use of `\` necessary here? I looked at a few other files in the same directory and not seeing the same pattern (ex: testenv.cpp, testaddonindex.cpp, testchckedint.cpp)

tests/unit_tests/testsettingsholder.cpp
```diff
@@ -0,0 +1,120 @@
...
+                 expectedValue.join(","));                                     \
+        QVERIFY(settingsHolder.has());                                         \
+                                                                               \
+        settingsHolder.remover();                                              \
+        QCOMPARE(spy.count(), 2);                                              \
+        QVERIFY(!settingsHolder.has());                                        \
+        return;                                                                \
+      }                                                                        \
+      default: {                                                               \
+        Q_ASSERT(false);                                                       \
```
Gela, line 112: Can more information be logged here?

tests/unit_tests/testsettingsholder.cpp
```diff
@@ -0,0 +1,120 @@
...
+        QCOMPARE(spy.count(), 2);                                              \
+        QVERIFY(!settingsHolder.has());                                        \
+        return;                                                                \
+      }                                                                        \
+      case QMetaType::Int:                                                     \
+      case QMetaType::LongLong: {                                              \
+        QVERIFY(!settingsHolder.has());                                        \
+        QCOMPARE(QVariant(settingsHolder.getter()).toInt(), v.toInt());        \
+                                                                               \
+        auto expectedValue = 42;                                               \
```
Gela, line 70: Where does 42 come from?

tests/unit_tests/testsettingsholder.cpp
brizental: That is because this is macro.

tests/unit_tests/testsettingsholder.cpp
brizental: Given this is just a test case, do you think a comment would be enough?

tests/unit_tests/testsettingsholder.cpp
brizental: Just a random testing value.

strseb (approved):

Gela (approved):

PR #8820, opened 2023-12-15
[PRE-WORK] VPN-5854 - Refactor storage management

brizental:
~**This PR is on top of #8819, please ignore the first two commits on it.**~ That has been merged.

This PR is also pre-work for VPN-5854, in PR #8821.

Experiment feature have attached settings, but the way our SettingsHolder is implemented makes it hard to dynamically add settings. There is an underlying assumption that all settings will be defined at build time in the settingslist.h file.

In order to introduce the experimental settings features I had a few options that were all bit a hacky. Such as doing what we did for addon features and adding addon-specific (or in this case, experiment specific) functions to the SettingsHolder. Or just bypassing the SettingsHolder completely and just creating a different QSettings location for experiments settings.

None of these options made sense to me and I decided to refactor the whole settings managent in the app, to be more flexible towards dynamic creation of metrics. 

The new design has the following pieces.

- The **SettingsManager** singleton, which manages the underlying QSettings location and serves as a bucket for all Settings in the app.
- The **Setting** class, which is an abstraction over one single setting in the underlying QSettings store.
- The **SettingFactory** class, which exposes the only public constructors for Setting objects. It guarantees _all_ new Setting objects that are created are created in the heap and are immediatelly registered to the SettingsBase class. It also guarantees that no duplicates are ever added to the SettingsManager class.
- The **SettingGroup** class, which is an abstraction over the concept of a group of metrics. All Settings under a SettingGroup are under the same QSettings group as well i.e. the key for all of them share the same prefix followed by a slash. SettingGroups are useful for observing changes to a group of related metrics or for grouping metrics in a semantic way.

I have intentionally not renamed or moved the SettingsHolder class and files. The objective was to refactor everything without having to touch the whol codebase to change paths and calls to SettingsHolder functions. If we feel that is something we want to do, we can file a follow up ticket to do so.

> ~**Note** This ia draft because the QML test I added is failing. It is possible and I would love to get feedback on the approach while fixing that though.~ Fixed.


src/settings/settingsbase.h
```diff
@@ -0,0 +1,124 @@
...
+  static QString settingsFileName();
+
+  /**
+   * @brief Soft reset all registered settings.
+   *
+   * If the setting is configured to be removed on reset this will reset it,
+   * otherwise it will stay there.
+   *
+   */
+  static void reset();
```
vinocher, line 67: Is there any advantage in moving away from pattern used for other singletons, which use non-static methods and the caller obtains the singleton instance?

src/settings/settingsbase.h
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * All settings are created using the SettingsFactory, which creates a Setting
+ * object and registers it in the SettingsBase class.
+ *
+ * The only public constructor for Setting objects is the SettingsFactory
+ * constructor, therefore everytime a Setting object is created it is guaranteed
+ * to be registered in the settings base.
+ *
+ */
+class SettingsBase final : public QObject, public LogSerializer {
```
vinocher, line 34: This name seems to indicate that it is a base class. Consider renaming it to something like SettingsManager

src/mozillavpn.cpp
```diff
@@ -1805,7 +1808,16 @@ void MozillaVPN::registerNavigationBarButtons() {
...
-  connect(SettingsHolder::instance(), &SettingsHolder::addonSettingsChanged,
+  // A group of settings containing all the addon message settings.
+  SettingGroup messageSettingGroup =
+      SettingGroup(QString("%1/%2")
+                       .arg(Constants::ADDONS_SETTINGS_GROUP)
+                       .arg(ADDON_MESSAGE_SETTINGS_GROUP),
+                   true,  // remove when reset
+                   false  // sensitive setting
+      );
+
```
vinocher, line 1819: Consider if a design like the following is better, after introducing a singleton SettingsManager:

  `SettingGroup messageSettingGroup =
      SettingGroup(QString("%1/%2")
                       .arg(Constants::ADDONS_SETTINGS_GROUP)
                       .arg(ADDON_MESSAGE_SETTINGS_GROUP),
                   true,  // remove when reset
                   false  // sensitive setting
      );

  SettingsManager::instance()->register(messageSettingGroup);`

This improves the Single Responsibility Principle, by moving the registration action out of SettingGroup.

src/settings/settinggroup.h
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * It is allowed to create SettingGroups that are subsets or supersets of other
+ * SettingGroups. However, this can be tricky if the different groups are
+ * configured differently e.g. one has removeWhenReset set to true, while the
+ * other has removeWhenReset set to false. Whichever group register the
+ * underlying setting first will have their configurations enforced. There are
+ * Q_ASSERTs set up in debug mode to catch such situations.
+ *
+ */
+class SettingGroup : public QObject {
```
vinocher, line 40: Consider if moving the registration code into SettingsBase can provide a better design. (Better Single Responsibility). That may remove or reduce the requirement for the friend classes:

```
	  SettingGroup messageSettingGroup =
      SettingGroup(QString("%1/%2")
                       .arg(Constants::ADDONS_SETTINGS_GROUP)
                       .arg(ADDON_MESSAGE_SETTINGS_GROUP),
                   true,  // remove when reset
                   false  // sensitive setting
      );
      SettingsManager::instance()->register(messageSettingGroup);
```

src/settings/settingsbase.h
brizental: No strong opinions, but I prefer this format because it is more restrictive than exposing the `instance`. For example, although `serializeLogs` is a public function, since `instance` is not public, only the logHandler which explicitly is given access to the instance can access it.

src/settings/settinggroup.h
brizental: Would you mind clarifying where the registration you are referring to is right now? The SettingsGroup itself doesn't register anything, it's the SettingsFactory that registers settings in the SettingsBase.

src/settings/setting.h
```diff
@@ -0,0 +1,106 @@
...
+
+  QString key() const { return m_key; }
+
+  /**
+   * @brief Get the stored value for this setting. If not set, will return the
+   * default value.
+   *
+   * @return T
+   */
+  Q_INVOKABLE QVariant get() const;
```
brizental, line 28: TODO: These don't need to be Q_INVOKABLE AFAICT.

src/settings/settinggroup.cpp
```diff
@@ -0,0 +1,93 @@
...
+SettingGroup::~SettingGroup() { MZ_COUNT_DTOR(Setting); }
+
+void SettingGroup::addSetting(const QString& key) {
+  auto settingKey = getSettingKey(key);
+  auto setting = SettingFactory::createOrGetSetting(
+      settingKey, []() { return nullptr; }, m_removeWhenReset,
+      m_sensitiveSetting);
+
+  // Make this setting's getter available for QML.
+  this->setProperty(key.toUtf8().constData(), get(key));
```
brizental, line 40: This is the part that is not working.

src/settings/settinggroup.h
```diff
@@ -0,0 +1,108 @@
...
+  void set(const QString& key, QVariant value);
+
+  /**
+   * @brief Removes the stored values for all settings in this group.
+   *
+   */
+  void remove();
+
+ signals:
+  void changed();
```
brizental, line 91: TODO: Add doc.

src/settings/settinggroup.h
```diff
@@ -0,0 +1,108 @@
...
+
+ private:
+  void addSetting(const QString& key);
+  bool mayRecord(const QString& key);
+  QString getSettingKey(const QString& key) const {
+    return QString("%1/%2").arg(m_groupKey, key);
```
brizental, line 97: TODO: Is this working correctly?!

src/settings/settinggroup.h
vinocher: Sorry, I didn't look closely enough previously.

The friend classes in SettingsManager.h does break its encapsulation and makes it more difficult to understand the relationship between the classes. Consider removing those and introducing public methods which can be used by Setting and SetringGroup, instead of directly manipulating SettingManager's m_settings.

For example, introduce SettingsManager.getKeysInGroup(const QString& groupName), which can be used to get all those keys and add a slot to detect any changes in the group.

tests/unit_tests/testaddon.cpp
```diff
@@ -224,7 +226,14 @@ void TestAddon::conditions() {
...
   if (!settingKey.isEmpty()) {
-    SettingsHolder::instance()->setRawSetting(settingKey, settingValue);
+    auto s = SettingsManager::getSetting(settingKey);
+    if (!s) {
+      SettingFactory::createOrGetSetting(
+          settingKey, []() { return nullptr; }, true, false);
+      s = SettingsManager::getSetting(settingKey);
+    }
+
+    s->set(settingValue);
```
Gela, line 236:
```suggestion
    auto setting = SettingsManager::getSetting(settingKey);
    if (!setting) {
      SettingFactory::createOrGetSetting(
          settingKey, []() { return nullptr; }, true, false);
      setting = SettingsManager::getSetting(settingKey);
    }

    setting->set(settingValue);
```

tests/unit_tests/testaddon.cpp
```diff
@@ -690,12 +699,34 @@ void TestAddon::message_load_status_data() {
 }
 
 void TestAddon::message_load_status() {
+  Localizer l;
```
Gela, line 702: I don't think this is used anywhere

src/settings/settingsbase.h
vinocher: A concern is that the cognitive load associated with reading and maintaining this code has increased, because it is different from the established singleton patterns in the code base.

src/settings/settinggroup.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  setting->set(value);
+}
+
+void SettingGroup::remove() {
+  auto sb = SettingsManager::instance();
+  sb->m_settings.beginGroup(m_groupKey);
+  QStringList keys = sb->m_settings.allKeys();
+  sb->m_settings.remove("");
+  sb->m_settings.endGroup();
```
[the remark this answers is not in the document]
vinocher: See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8821 for more comments.

src/settings/settinggroup.h
brizental: It must be working because otherwise the test case `testLoadSettingsOnInit` would fail. This is werid though. I will change it to normal usage.

src/settings/settingsmanager.cpp
```diff
@@ -0,0 +1,128 @@
...
+  if (!s_instance) {
+    s_instance = new SettingsManager(qApp);
+  }
+
+  return s_instance;
+}
+
+#ifdef UNIT_TEST
+// static
+void SettingsManager::testCleanup() { delete s_instance; }
```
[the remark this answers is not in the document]
brizental: The destructor does that.

src/settings/settingsmanager.cpp
```diff
@@ -0,0 +1,128 @@
...
+        a_callback) {
+  std::function<void(const QString& name, const QString& logs)> callback =
+      std::move(a_callback);
+
+  QString buff;
+  QTextStream out(&buff);
+  foreach (Setting* setting, m_registeredSettings.values()) {
+    auto log = setting->log();
+    if (!log.isEmpty()) {
+      out << log << Qt::endl;
```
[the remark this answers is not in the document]
brizental: `out` is a variable created right before this loop. We are using `Qt:endl`....... 👀 ......... I don't know, that is what was there before. See: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/settingsholder.cpp#L180

tests/unit_tests/testaddon.cpp
brizental: This is a pattern in out tests. The constructor has side effects taht are required for the test to run correctly.

src/settings/settinggroup.h
brizental: Yeah ok... I learned about friend classes while working in this PR and maybe I got too excited. I'll do that.

src/settingsholder.cpp
```diff
@@ -32,19 +29,20 @@ SettingsHolder* SettingsHolder::instance() {
...
 
-SettingsHolder::SettingsHolder()
-    : m_settings(MozFormat, QSettings::UserScope,
-#ifndef UNIT_TEST
-                 "mozilla",
-#else
-                 "mozilla_testing",
-#endif
-                 Constants::SETTINGS_APP_NAME) {
+SettingsHolder::SettingsHolder() {
```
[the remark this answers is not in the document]
brizental: https://mozilla-hub.atlassian.net/browse/VPN-6046

src/settings/setting.h
```diff
@@ -0,0 +1,109 @@
...
+ signals:
+  /**
+   * @brief This signal is emmited whenever the underlying storage value related
+   * to the setting is changed.
+   *
+   * Check the documentation of the updater functions above to understand when
+   * the underlying value is changed or not.
+   *
+   */
+  void changed() const;
```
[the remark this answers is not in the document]
brizental: Same comment as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8821#discussion_r1433086587

src/settings/qsettingsconnector.h
```diff
@@ -0,0 +1,64 @@
...
+#define QSETTINGSCONNECTOR_H
+
+#include <QObject>
+#include <QSettings>
+
+/**
+ * @brief Exposes APIs to interact with an instance of QSettings.
+ *
+ */
+class QSettingsConnector : public QObject {
```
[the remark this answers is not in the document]
brizental: Yeah, good call. I'll rename it.

src/settings/settinggroup.cpp
```diff
@@ -0,0 +1,82 @@
...
+  QStringList keys = m_settingsConnector->getAllKeys(m_groupKey);
+
+  m_settingsConnector->remove("", m_groupKey);
+
+  foreach (const QString& key, keys) {
+    auto fullKey = getSettingKey(key);
+    auto setting = SettingsManager::instance()->getSetting(fullKey);
+    setting->changed();
+  }
+}
```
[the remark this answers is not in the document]
brizental: The group signal will be emmited whenever a setting in the group emmits it's signal already. So we don't need to explicitly do that here.

tests/unit_tests/settings/testsettingsmanager.cpp
```diff
@@ -0,0 +1,160 @@
...
+  Q_UNUSED(setting)
+
+  // Try to get it, not it's not nullptr.
+  Q_ASSERT(SettingsManager::instance()->getSetting("doesexist"));
+}
+
+void TestSettingsManager::testReset() {
+  // Create a setting that should be reset
+  auto doReset = SettingsManager::createOrGetSetting(
+      "doreset", []() { return QVariant(); }, true, false);
```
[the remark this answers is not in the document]
brizental: Yeah, I've done that is most of the code and only skipped it in test code. I'll add the comments.

src/settings/setting.cpp
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+void Setting::reset() const {
+  if (!m_removeWhenReset) {
+    return;
+  }
+
+  remove();
+}
```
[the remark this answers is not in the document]
brizental: The `remove` function will take care of emmiting the signal or not.

src/settings/setting.h
```diff
@@ -0,0 +1,112 @@
...
+
+ private:
+ private:
+  QString m_key;
+
+  std::function<QVariant()> m_defaultValue;
+
+  bool m_sensitiveSetting;
+  bool m_removeWhenReset;
+
```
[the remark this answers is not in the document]
brizental: I've seen you ask for that in other place as well. Curious: why is that better? Am I not forced to initialize these in the constructor if they don't have any value anyways?

src/settings/settingsmanager.h
```diff
@@ -0,0 +1,170 @@
...
+   * @param removeWhenReset Whether or not this settings in this group should
+   * actually be removed when `reset` is called.
+   * @param sensitiveSetting Whether or not settings in this group are
+   * sensitive settings i.e. settings that must not be logged in plain text.
+   * @param acceptedKeys When this list has at least one member, only the listed
+   * keys will be allowed for storing under this group. Other keys will be
+   * ignored.
+   */
+  static SettingGroup createSettingGroup(
+      const QString& groupKey, bool removeWhenReset = true,
```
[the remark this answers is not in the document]
brizental: I _thought_ the latter would be possible, but then we get other issues such as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8820#discussion_r1433124666. Changing this to return a pointer.

src/settings/settingsmanager.cpp
```diff
@@ -0,0 +1,159 @@
...
+  foreach (Setting* setting, m_registeredSettings.values()) {
+    Q_ASSERT(setting);
+    setting->changed();
+  }
+
+  m_registeredSettings.clear();
+  // Free the memory for everything that was in the map.
+  qDeleteAll(m_registeredSettings);
+}
+
```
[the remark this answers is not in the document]
brizental: YES! Good catch!

src/settings/settingsmanager.cpp
```diff
@@ -0,0 +1,159 @@
...
+  Q_ASSERT(s_instance == this);
+  s_instance = nullptr;
+}
+
+QString SettingsManager::settingsFileName() { return m_settings.fileName(); }
+
+void SettingsManager::registerSetting(Setting* setting) {
+  Q_ASSERT(setting);
+
+  if (m_registeredSettings.contains(setting->key())) {
```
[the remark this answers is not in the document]
brizental: What are you referring to here? I am assuming you are commenting on outdated code... I accidentally commited debug logs yesterday.

src/settings/settingsmanager.h
```diff
@@ -0,0 +1,170 @@
...
+   * meantime by another application.
+   *
+   * This function is called automatically from QSettings's
+   * destructor and by the event loop at regular intervals,
+   * so you normally don't need to call it yourself.
+   *
+   * Description copied from: https://doc.qt.io/qt-6/qsettings.html#sync
+   *
+   */
+  void sync() { instance()->m_settings.sync(); };
```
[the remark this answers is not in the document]
brizental: Yeah, I went a little overboard on this one because it's a weirdo for me. I can just be more succint.

src/settings/qsettingsconnector.cpp
```diff
@@ -0,0 +1,41 @@
...
+void QSettingsConnector::setValue(const QString& key, QVariant value) const {
+  m_settings->setValue(key, value);
+}
+
+void QSettingsConnector::remove(const QString& key,
+                                const QString& group) const {
+  m_settings->beginGroup(group);
+  m_settings->remove(key);
+  m_settings->endGroup();
+}
```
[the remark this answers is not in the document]
brizental: The SettingsConnector doesn't know about the SettingsManager.

src/settings/settingsmanager.h
```diff
@@ -0,0 +1,170 @@
...
+#endif
+
+ private:
+  SettingsManager(QObject* parent);
+
+  void registerSetting(Setting* setting);
+
+ private:
+  QMap<QString, Setting*> m_registeredSettings;
+  QSettings m_settings;
```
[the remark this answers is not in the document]
brizental: It doesn't necessarily load them. When you create a SettingGroup yes it loads settings on init. But that is not true for a Setting object. So, if a setting is registered with the SettingsManager, it knows about it, but it may not have any value in the QSettings underlying storage.

tests/unit_tests/settings/testsettingsmanager.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  // Now reset!
+  SettingsManager::instance()->reset();
+
+  // Fun fact, reset will _not_ unregister. Check that is true.
+  Q_ASSERT(SettingsManager::instance()->getSetting("doreset"));
+  Q_ASSERT(SettingsManager::instance()->getSetting("donotreset"));
+
+  // But it will clear the storage, if the setting is setup to do that.
+  QVERIFY(doReset->get().isNull());
```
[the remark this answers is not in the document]
brizental: Nope, the caller is not expected to do so. I am thinking now that hardReset should not unregister the settings either. Neither method is guaranteed to be called in a situation where unregistering is the right move, which could lead to segfaults due to the unexpected behaviour. Remove also shouldn't unregister and so on.

Think about the situation where the user does a hardReset and then keeps using the app. (That is not possible currently, there is only hardResetAndQuit, but there is also nothing stopping us from implementing it.) In this situation suddenly calling any method on the SettingsHolder will result in a segfault.

Another option is to use shared pointers, but that would create the issue that a pointer to a setting exists that the SettingsManager is not aware of -- which invites another host of issues.

I'll go ahead and remove unregistering logic from everywhere but the destructor. Let me know if you disagree.

Gela (approved):

src/settings/setting.h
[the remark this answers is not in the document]
vinocher: See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers

src/settings/settingsmanager.cpp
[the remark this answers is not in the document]
vinocher: It looks good now -- I must have seen some old code.

src/settings/settingsmanager.h
```diff
@@ -0,0 +1,181 @@
...
+ * lifetime of the QApplication.
+ *
+ * There are two types of settings currently in this codebase:
+ *
+ * * Static settings are settings defined at build time in a declarative manner
+ * e.g. the SETTING macro or the EXPERIMENTAL_FEATURE macro.
+ * * Dynamic settings are settings which are created on demand at runtime.
+ *
+ * The only public constructor for Setting objects is
+ * SettingsManager::instance()->createOrGetSetting, therefore everytime a
```
vinocher, line 28: The Settings constructor is public, so it could be created using new Settings(...). However, not a big deal.

vinocher (approved):
Looks great!

src/settings/settingsmanager.h
```diff
@@ -0,0 +1,181 @@
...
+  // The actual underlying storage.
+  QSettings m_settings;
+
+  // APIs to access the QSettings underlying storage.
+  SettingsConnector m_settingsConnector;
+
+#ifdef UNIT_TEST
+  friend class TestSettingsManager;
+  friend class TestSettings;
+#endif
```
vinocher, line 178: Are these needed anymore?

src/settings/settingsmanager.h
brizental: Yes. I think though, since the only public constructor requires SettingsConnector it might as well be private, because it's just too hard to use outside of the SettingsManager :D

src/settings/setting.h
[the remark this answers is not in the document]
brizental: Ah! Ok got it. I assumed the compiler would forme to initialize an un-initialized member!

src/settings/settingsmanager.h
brizental: Yes, but not all of them.

PR #8822, opened 2023-12-15
Use SSL server for Adjust proxy

brizental:
Requests to the adjust proxy fail because it's not using an encrypted connection.

This fixes it.

### DIFF
--- a/src/adjust/adjustproxy.cpp
+++ b/src/adjust/adjustproxy.cpp
@@ -5,7 +5,7 @@
 #include "adjustproxy.h"
 
 #include <QHostAddress>
-#include <QTcpSocket>
+#include <QSslSocket>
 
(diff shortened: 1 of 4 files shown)


src/adjust/adjustproxy.cpp
```diff
@@ -15,7 +15,7 @@ namespace {
 Logger logger("AdjustProxy");
 }  // namespace
 
-AdjustProxy::AdjustProxy(QObject* parent) : QTcpServer(parent) {
+AdjustProxy::AdjustProxy(QObject* parent) : QSslServer(parent) {
```
bakulf, line 18: What certs do we use here?

brizental:
We are not going ahead with this approach.

PR #8824, opened 2023-12-15
Adjust fixes for Android client: add a network conf, change how token is passed to gradle

strseb:
A write up of the issues on hand is in the ticket: https://mozilla-hub.atlassian.net/browse/VPN-5822?focusedCommentId=809828

## Description

Instead of a bool, let's enforce https for all known domains the client can talk to.

### DIFF
--- a/android/resources/all/xml/network_security_config.xml
+++ b/android/resources/all/xml/network_security_config.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true">
+        <trust-anchors>
+            <certificates src="system" />
+        </trust-anchors>
+    </base-config>
+    <domain-config cleartextTrafficPermitted="false">
+        <domain includeSubdomains="true">mozilla.org</domain>
+        <domain includeSubdomains="true">firefox.com</domain>
+        <domain includeSubdomains="true">mozgcp.net</domain>
+        <domain includeSubdomains="true">sentry.io</domain>
+        <domain includeSubdomains="true">apple.com</domain>
+        <domain includeSubdomains="true">google.com</domain>
+        <domain includeSubdomains="true">github.io</domain> 
+        <trust-anchors>
+            <certificates src="system" />
+        </trust-anchors>
+    </domain-config>
+</network-security-config>
--- a/android/resources/all/xml/network_security_config.xml
+++ b/android/resources/all/xml/network_security_config.xml
@@ -12,7 +12,9 @@
         <domain includeSubdomains="true">sentry.io</domain>
         <domain includeSubdomains="true">apple.com</domain>
         <domain includeSubdomains="true">google.com</domain>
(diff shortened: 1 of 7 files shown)


android/resources/all/xml/network_security_config.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true">
+        <trust-anchors>
+            <certificates src="system" />
+        </trust-anchors>
+    </base-config>
+    <domain-config cleartextTrafficPermitted="false">
+        <domain includeSubdomains="true">mozilla.org</domain>
```
lesleyjanenorton, line 9: ++ mozaws.net and allizom.org I think?

android/resources/all/xml/network_security_config.xml
```diff
@@ -0,0 +1,20 @@
...
+        <domain includeSubdomains="true">firefox.com</domain>
+        <domain includeSubdomains="true">mozgcp.net</domain>
+        <domain includeSubdomains="true">sentry.io</domain>
+        <domain includeSubdomains="true">apple.com</domain>
+        <domain includeSubdomains="true">google.com</domain>
+        <domain includeSubdomains="true">github.io</domain>
```
brizental, line 15: As discussion on Slack, I feel this is very fragile given it is totally reliant on us _remembering_ to put new domains here as they are used. A few ideas: Could we use wildcards here? Or could we use IP addresses and then provide the loopback address as an exception instead of making the base case `true`? 

I was also going to ask if this is necessary at all, because AFAIK we enforce HTTPS on production for all requests on the C++ side. However turns out I could not find that code to link to, so now I am second guessing it 😅

strseb: >Or could we use IP addresses and then provide the loopback address as an exception instead of making the base case true? 

Tested that, didnt work. 

>Could we use wildcards here?

No, sorry.

> I feel this is very fragile given it is totally reliant on us remembering to put new domains here as they are used.

It's a layer of security, so the goal of this file is to reduce risk, not to be an impenetrable new sec policy. That would be  `usesCleartextTraffic="false"`

>I was also going to ask if this is necessary at all, because AFAIK we enforce HTTPS on production for all requests on the C++ side. However turns out I could not find that code to link to, so now I am second guessing it 😅

Not all requests originate from c++ tho. Glean on android i.e. uses a kotlin worker to post messages.

scripts/android/cmake.sh
```diff
@@ -190,7 +183,7 @@ cd .tmp/src/android-build/
 if [[ "$RELEASE" ]]; then
   print Y "Generating Release APK..."
   ./gradlew compileReleaseSources
-  ./gradlew assemble || die
```
strseb, line 193: Thanks @mcleinman thats waaay better!

brizental (approved):

lesleyjanenorton (approved):
👍

Issue #7912, opened 2023-09-05
blank home screen in UI after update from 2.15.0 to 2.16.1


bbbart:
**Describe the bug**
After upgrading from 2.15.0 to 2.16.1 the GUI shows no option on the home pane. The messages and settings seem to work fine, but there is no way to select a server or enable the VPN.

**VPN version:**
2.16.1

**Affected Platforms:**
Arch Linux

**Prerequisites:**
None that I know of. Happens after a fresh reboot too.

**Steps to reproduce:**
Just open the GUI

**Expected result:**
To be able to connect to the VPN after selecting a server.

**Actual result:**
No way to enable the VPN or to select a server.

**Additional notes:**
 ![image](https://github.com/mozilla-mobile/mozilla-vpn-client/assets/3244673/7ca97f68-9a42-44e0-b991-f0ccb5bf6cab)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-5472)

bbbart:
(enabling the VPN from the command line works just fine)

arnobl:
This still happens with 2.17.0 on Fedora 38. Qt version 6.5.2

arnobl:
Well, I removed all the "Mozilla VPN" configuration folders and 'vpn.moz' and it now works as expected.

arnobl:
Well, well, my fix fixes nothing in fact. This still happens quite randomly. 
If I am logged in, this happens, nevermind whether the app starts automatically. If I close the app and start it again, the app is displayed correctly.
If a clean data, the login page is correctly visible.

bbbart:
It seems to be happening randomly here as well, but more often than not it's blank. I'm still at 2.16.1.

arnobl:
Here is my log when I manually start for the first time the VPN on a fresh startup:

```
[05.10.2023 08:28:50.124] (LogHandler) Debug: Log file: /home/ablouin/.local/share/mozillavpn.log
[05.10.2023 08:28:50.219] (Localizer) Debug: Language excluded: bg completeness: 0
[05.10.2023 08:28:50.222] (Localizer) Debug: Language excluded: da completeness: 0
[05.10.2023 08:28:50.222] (Localizer) Debug: Language excluded: dsb completeness: 0
[05.10.2023 08:28:50.222] (Localizer) Debug: Language excluded: fa completeness: 0
[...]

PR #8843, opened 2023-12-19
VPN-5472: Merge Mozilla.Shared.qmlcomponents into Mozilla.VPN module

oskirby:
## Description
We have received a few reports of users encountering a blank UI screen on some Linux distributions. So far this seems to manifest on Ubuntu 20.04 intermittently. When this occurs, we find a QML error in the log of the form: 
```
Warning: qrc:/ui/screens/ScreenHome.qml:9:1: module "Mozilla.VPN" is not installed (ScreenHome.qml:9)
```

However, this `Mozilla.VPN` QML module is created programmatically by the program. I suspect that this error occurs because something in the QML cache isn't able to figure out that this module exists, creating this module with `qt6_add_qml_module()` might do the trick.

## Reference
Github issue: #7912 ([VPN-5472](https://mozilla-hub.atlassian.net/browse/VPN-5472))

## Checklist
    
[VPN-5472]: https://mozilla-hub.atlassian.net/browse/VPN-5472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
Bah, humbug. Okay, so this did actually fix the QML cache loading jankiness, but it completely broke the QML_NAMED_ELEMENT() classes from the C++ world. I guess this means I'm on the right track at least.

src/productshandler.cpp
```diff
@@ -44,6 +44,11 @@ ProductsHandler::ProductsHandler(QObject* parent) : QAbstractListModel(parent) {
   Q_ASSERT(!s_instance);
   s_instance = this;
 
+  // If web purchases are supported, then we don't need to do anything.
```
mcleinman, line 47: Why move this in?

nebula/ui/components/filterproxymodel.cpp
```diff
@@ -4,12 +4,9 @@
 
 #include "filterproxymodel.h"
 
-#include "logger.h"
-#include "qmlengineholder.h"
-
-namespace {
-Logger logger("FilterProxyModel");
-}
+#include <QDebug>
```
mcleinman, line 7: Why the changes in this file?

mcleinman (commented):
Looks good. A couple questions so I can better understand this, as I don't know this part of the code super-well.

I don't have a Linux box or VM currently, so I'm not able to test it.

nebula/ui/components/filterproxymodel.cpp
oskirby: In order to move this class into the Nebula class, it was important to make sure that the C++ code doesn't depend on code in the `mozillavpn` target, otherwise this would result in a dependency loop. This mostly just required that we drop the use of the VPN's logging infrastructure and just do it in pure Qt.

src/addons/addonmessage.h
```diff
@@ -22,8 +22,6 @@ constexpr const char* ADDON_MESSAGE_SETTINGS_STATUS_KEY = "state";
 class AddonMessage final : public Addon {
   Q_OBJECT
   Q_DISABLE_COPY_MOVE(AddonMessage)
-  QML_NAMED_ELEMENT(MZAddonMessage)
```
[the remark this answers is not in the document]
oskirby: Generally I was finding that the QML types registered implicitly by the `mozillavpn` target were failing to make it into the QML namespace. To work around it, this PR is attempting to move all the QML type registration into the `mozillavpn-ui` target and we do it in `src/ui/types/MZAddonMessage.h` with the `QML_FOREIGN()` macro.

src/productshandler.cpp
oskirby: I couldn't find a way to conditionally register a singleton type purely using `qt_add_qml_module()` in a way that would respect the feature flags so I figured the cleanest solution was to register all the QML types in the `Mozilla.VPN` module (and to do so consistently regardless of feature flagging).

In order to replicate the intended behaviour of the C++ code that this replaces, it was necessary to ensure that the `ProductsHandler` class doesn't do anything when the feature flag is disabled even if we happen to instantiate it.

mcleinman (approved):
Thanks for those explanations. Code looks great to me, but I'm not able to test it - I don't have a Linux box set up. 

Naomi, your call on whether you want to merge or wait for additional 👍🏻 on this one.

src/commands/commandui.cpp
```diff
@@ -309,61 +309,6 @@ int CommandUI::run(QStringList& tokens) {
...
-#endif
-
-    qmlRegisterSingletonInstance("Mozilla.VPN", 1, 0, "VPNPurchase",
-                                 PurchaseHandler::instance());
-
-    if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
-      qmlRegisterSingletonInstance("Mozilla.VPN", 1, 0, "VPNProducts",
-                                   ProductsHandler::instance());
-    }
-
```
lesleyjanenorton, line 366: This wasn't introduced by your changes but does it still make sense to register MZAccessibleNotification and MZI18n here instead of qmlengineholder? Seems odd to be handling this in two places.

nebula/ui/components/forms/MZSearchBar.qml
```diff
@@ -6,7 +6,7 @@ import QtQuick 2.0
 import QtQuick.Layouts 1.14
 
 import Mozilla.Shared 1.0
-import Mozilla.Shared.qmlcomponents 1.0
+import Mozilla.VPN 1.0
```
lesleyjanenorton, line 9: Do we still need the Mozilla.VPN import here

nebula/ui/components/MZGuideList.qml
```diff
@@ -6,7 +6,7 @@ import QtQuick 2.15
 import QtQuick.Layouts 1.15
 
 import Mozilla.Shared 1.0
-import Mozilla.Shared.qmlcomponents 1.0
+import Mozilla.VPN 1.0
```
lesleyjanenorton, line 9: Same question here. Do we still need the .VPN import?

lesleyjanenorton (commented):
And here I was really hoping this was just an issue of adding an import statement or two. Oof. Thanks for taking this on @oskirby! 

A few things:
- Why move the composer dir into src/ui? Is it possible to move it to nebula/ui instead? NBD if not but so far only QML types in Nebula are prefixed with "MZ". (Another matter for our tech debt pile: we no longer need to separate the QML according to what can be 'shared' or not) 
- I'm still completely baffled as to why Mozilla.Shared works and Mozilla.VPN randomly fails (although to your point on Slack, maybe Shared fails too and we just don't know it yet).

lesleyjanenorton:
Also want to note: My questions don't need to be answered if you're ready to merge. I'm just baffled and curious/wondering.

lesleyjanenorton (approved):

tests/qml/CMakeLists.txt
```diff
@@ -161,8 +147,6 @@ target_sources(qml_tests PRIVATE
     ${MZ_SOURCE_DIR}/errorhandler.h
     ${MZ_SOURCE_DIR}/feature/feature.cpp
     ${MZ_SOURCE_DIR}/feature/feature.h
-    ${MZ_SOURCE_DIR}/filterproxymodel.cpp
```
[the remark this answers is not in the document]
oskirby: We do, it's just included as a part of the nebula library now instead of needing to be manually included in the tests.

src/commands/commandui.cpp
oskirby: I agree, it is a bit weird registering things in multiple places like this. However, my ideal solution, if I were up to another refactoring challenge, would be to combine the `Mozilla.VPN` and `Mozilla.Shared` modules together. They are vestigial remains of our attempt to modularize the codebase.

nebula/ui/components/qmldir
```diff
@@ -15,8 +15,8 @@ MZCheckBox 0.1 MZCheckBox.qml
 MZCheckBoxRow 0.1 MZCheckBoxRow.qml
 MZChevron 0.1 MZChevron.qml
 MZClickableRow 0.1 MZClickableRow.qml
-MZCollapsibleCard 0.1 MZCollapsibleCard.qml
 MZComposerView 0.1 MZComposerView.qml
+MZCollapsibleCard 0.1 MZCollapsibleCard.qml
```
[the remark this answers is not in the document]
oskirby: Oops! I was messing around with moving the `MZComposerView` around a bunch, and I guess I forgot to put it back where I found it.

oskirby:
>     * Why move the composer dir into src/ui? Is it possible to move it to nebula/ui instead? NBD if not but so far only QML types in Nebula are prefixed with "MZ". (Another matter for our tech debt pile: we no longer need to separate the QML according to what can be 'shared' or not)
I wanted to; it would have cleaned things up nicely and left all the composer stuff in the `nebula` library. Unfortunately, there's a dependency between the composer classes and the `AddonApi` which would result in a dependency loop between `nebula` and `mozillavpn`.

PR #8850, opened 2023-12-20
Use the `release-promotion` action from `mozilla-taskgraph`

ahal:
## Description

We upstreamed the `release-promotion` action to a shared location in `mozilla-taskgraph` so we can stop copy/pasting this everywhere it's needed. The action in `mozilla-taskgraph` is identical to the one that was here.

## Checklist
    

ahal:
To avoid regressions over the holidays, I'll hold off landing this until next year.

strseb (approved):

PR #8834, opened 2023-12-18
VPN-4987 [Part 1] - Check for internet connectivity upon activation + VPN-5994 - Inspector command

Gela:
## Description

This PR does 2 things:
1. Check for internet connection upon activation. This is the backend code only, [frontend](https://mozilla-hub.atlassian.net/browse/VPN-3711) will be completed in a later sprint
2. Add inspector command force_connection_down to mock network connection being down.

Please note that functional tests cannot be added until the front end work is done.

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-5994
    https://mozilla-hub.atlassian.net/browse/VPN-4987

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #8853, opened 2023-12-20
VPN-6042 Check QNetworkInformation::instance() exists first

Gela:
## Description

Check QNetworkInformation::instance() exists first before attempting to access its reachability. If it doesn't exist, return reachability::unknown.

## Reference

   https://mozilla-hub.atlassian.net/browse/VPN-6042

## Checklist
    
### DIFF
--- a/src/networkwatcher.cpp
+++ b/src/networkwatcher.cpp
@@ -170,5 +170,8 @@ void NetworkWatcher::notificationClicked(NotificationHandler::Message message) {
 }
 
 QNetworkInformation::Reachability NetworkWatcher::getReachability() {
-  return QNetworkInformation::instance()->reachability();
+  if (QNetworkInformation::instance()) {
+    return QNetworkInformation::instance()->reachability();
+  }
+  return QNetworkInformation::Reachability::Unknown;
 }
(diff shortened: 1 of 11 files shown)


vinocher:
Why was a different NetworkWatcher added for each platform?  (Instead of using https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8834?)

src/networkwatcher.cpp
```diff
@@ -170,5 +170,8 @@ void NetworkWatcher::notificationClicked(NotificationHandler::Message message) {
 }
 
 QNetworkInformation::Reachability NetworkWatcher::getReachability() {
-  return QNetworkInformation::instance()->reachability();
+  if (QNetworkInformation::instance()) {
+    return QNetworkInformation::instance()->reachability();
+  }
+  return QNetworkInformation::Reachability::Unknown;
```
vinocher, line 176: Why was a different NetworkWatcher added for each platform? (Instead of using https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8834?)

Gela: It follows the same pattern as `getCurrentTransport()`, anywhere that function used to exist is now replaced by `getReachability()` by [this](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8835/files) PR.

vinocher: As discussed offline, the NetworkWatcherImpl changes are not required because QNetworkInformation is already cross-platform.

vinocher (approved):

PR #8859, opened 2023-12-21
VPN-5854 - Implement infrastructure for experimental features

brizental:
> Note: I reccomend reviewing this commit by commit.
> I did some moving of files around in a housekeeping effort and looking at the whole diff here will be confusing.

As described in: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/rfcs/0006-cirrus-experiments.md#3-apply-features

Unfortunately, the exact design as propsoed in the RFC could not be implemented -- my bad.

~The QML API for experiment settings will be:~ EDIT: QML API will be done in a follow up, https://mozilla-hub.atlassian.net/browse/VPN-6057

```
MZSettings.onboardingCtaStyleExperiment.get('color')
MZSettings.onboardingCtaStyleExperiment.get('fontFamily')
```

And the C++ API for experiment settings will be:

```
settingsHolder.onboardingCtaStyleExperiment().get('color')
settingsHolder.onboardingCtaStyleExperiment().get('fontFamily')
```


src/cmake/sources.cmake
```diff
@@ -24,7 +24,7 @@ set_property(TARGET mozillavpn-sources PROPERTY INTERFACE_INCLUDE_DIRECTORIES
 target_sources(mozillavpn-sources INTERFACE
     ${CMAKE_CURRENT_SOURCE_DIR}/accessiblenotification.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/feature/featurelistcallback.h
-    ${CMAKE_CURRENT_SOURCE_DIR}/feature/featurelist.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/feature/experimentalfeaturelist.h
```
brizental, line 27: Should not have removed the other one.

src/settingsholder.h
```diff
@@ -31,6 +33,12 @@ class SettingsHolder final : public QObject {
 #include "settingslist.h"
 #undef SETTING
 
+#define EXPERIMENTAL_FEATURE(experimentId, ...) \
+  Q_PROPERTY(SettingGroup experimentId READ experimentId)
```
brizental, line 37: I should test this, pretty sure it's not working as the type is wrong. It's actually `SettingsGroup*`.

vinocher:
A lot of the code in this PR seems to be the same as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8820. Does this PR need to be rebased? I'm a bit confused about which remaining changes need to be reviewed.

brizental:
> A lot of the code in this PR seems to be the same as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8820. Does this PR need to be rebased? I'm a bit confused about which remaining changes need to be reviewed. 

Hey Vinoo, yeah sorry. This needs to be rebased!

tests/unit_tests/testsettingsholder.cpp
```diff
@@ -117,4 +117,22 @@
 #include "settingslist.h"
 #undef SETTING
 
+#define EXPERIMENTAL_FEATURE(experimentId, experimentDescription, \
```
vinocher, line 120: Why is EXPERIMENTAL_FEATURE being re-defined for these tests (instead of using feature.h/cpp)?

src/settingsholder.h
vinocher: Yes, looks like it.

vinocher (approved):

src/settingsholder.h
brizental: Ok, fun times. This is indeed not working and I've spent the whole day debugging it to no success.

I am 99% sure there is a Qt MOC bug here. Some stuff I learned: 

Q_PROPERTY, Q_INVOKABLE, Q_SIGNAL, Q_OBJECT etc. These are _not_ preprocessor macros. See: https://forum.qt.io/topic/79689/the-qt-moc-and-preprocessor

From Qt5 on, it is allowed to use these special MOC macros inside preprocessor macros. Before that, they would get ignored. See: https://bugreports.qt.io/browse/QTBUG-35

Now, the weird thing here is that it looks like we are experiencing is a re-appearance of QTBUG-35. All MOC macros I add inside this specific EXPERIMENTAL_FEATURE definition get ignored. I inspected the moc_ file generated and indeed the generated code expected is not there. If I manually add the experiment settings code, it works fine:

```
// This works!
Q_PROPERTY(SettingGroup* myExperimentalFeature READ myExperimentalFeature
                 WRITE experimentSetter NOTIFY myExperimentalFeatureChanged)
Q_SIGNAL void myExperimentalFeatureChanged();
```

The behaviour is not exactly like QTBUG-35 though. Given we have the SETTING macro definition right below this one and that one has MOC macros inside it that work just fine. I spent today trying to understand the issue in order to report a QTBUG. But I was not successful. So now this has become a 2024 Bea problem 👋

brizental: I guess what I should have done is punted this to a follow up ticket so as not to block this one 🤔 Leaving this as a note to start the new year doing that.

tests/unit_tests/testsettingsholder.cpp
brizental: To test the experimental feature APIs inside the SettingsHolder.

src/settingsholder.h
brizental: https://mozilla-hub.atlassian.net/browse/VPN-6057

PR #8864, opened 2023-12-22
VPN-4687: Add DNS settings help sheet

MattLichtenstein:
## Description

- Implement help sheet for DNS settings
- Slight tweaks to help sheet UI

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/7ee24b0c-bedc-4ab8-84c6-4b324d399c4b

## Reference

[VPN-4687: Implement the 'Custom DNS Settings' in-product help sheet](https://mozilla-hub.atlassian.net/browse/VPN-4687)


src/translations/strings.yaml
```diff
@@ -928,6 +928,20 @@ devices:
     value: My devices
     comment: Title for the menu bar when viewing the devices or device limit pages
 
+helpSheets: 
+  dnsTitle:
+    value: Custom DNS settings
+    comment: Title label for the custom dns help sheet
+  dnsHeader: 
+    value: What is custom DNS?
```
flodolo, line 936:
```suggestion
    value: What is a custom DNS?
```

Sounds off without articles?

flodolo (approved):
Not blocking on English, but I think the title should be updated.

src/translations/strings.yaml
```diff
@@ -928,6 +928,20 @@ devices:
...
+    comment: Header label for the custom dns help sheet
+  dnsBody1:
+    value: Whenever you connect to a website, a DNS (domain name system) first turns the domain name (e.g. www.mozilla.org) into an IP address that allows your internet traffic to reach that destination.
```
vinocher, line 939: Just checking if this content has been approved by UX? The help text doesn't seem to explain what 'resolve websistes' means in the screen UI.

src/ui/screens/settings/ViewDNSSettings.qml
```diff
@@ -21,6 +21,25 @@ MZViewBase {
...
     property bool dnsSelectionChanged: false
     readonly property string telemetryScreenId : "dns_settings"
+    property Component rightMenuButton: Component {
+        Loader {
+            active: MZFeatureList.get("helpSheets").isSupported
+            sourceComponent: MZIconButton {
+                onClicked: helpSheetLoader.active = true
+
+                accessibleName: MZI18n.GlobalHelp
+                Accessible.ignored: !visible
```
vinocher, line 31: MZButtonBase has `Accessible.ignored: !visible`, so this should not be needed.

nebula/ui/components/MZMenu.qml
```diff
@@ -103,10 +103,11 @@ Item {
 
         sourceComponent: menuBar.rightButtonComponent
         onItemChanged: {
-            if (item instanceof Text) {
+            //item.item is for right button components that might be inside a loder itself
```
vinocher, line 106: Typo

nebula/ui/components/MZMenu.qml
```diff
@@ -103,10 +103,11 @@ Item {
...
+            if (item instanceof Text || item.item instanceof Text) {
```
vinocher, line 107: For the item.item condition, should the first item be checked if it is an instance of Loader? Also, are any null checks required?

src/mozillavpn.cpp
```diff
@@ -1381,6 +1381,11 @@ void MozillaVPN::registerUrlOpenerLabels() {
   uo->registerUrlLabel("upgradeToAnnualUrl", []() -> QString {
     return Constants::upgradeToAnnualUrl();
   });
+
+  uo->registerUrlLabel("sumoDns", []() -> QString {
```
vinocher, line 1385: Is sumoDns used?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -82,7 +81,8 @@ MZBottomSheet {
                         id: icon
                         anchors.centerIn: parent
 
-                        sourceSize.width: MZTheme.theme.iconSize
+                        source: "qrc:/nebula/resources/tip-filled.svg"
```
vinocher, line 84: Was the iconSource property alias removed because MZHelpSheet is expected to always use tip-filled.svg?

nebula/ui/components/MZHelpSheet.qml
```diff
@@ -96,24 +96,31 @@ MZBottomSheet {
...
-                    id: iconButton
-
                     Layout.rightMargin: MZTheme.theme.windowMargin / 2
 
                     Layout.preferredHeight: MZTheme.theme.rowHeight
                     Layout.preferredWidth: MZTheme.theme.rowHeight
 
+                    //Hacky workaround because for some reason, when opening a sheet via
+                    //the right menu button in MZMenu, this close button is in a
+                    //"Hovered" state, even though it wasn't hovered
```
vinocher, line 119: Could this be because the mouseup event that ends the hovered state is sent to the helo sheet, so the the right menu button's screen doesn't receive it, and assumes that it is still hovered?

src/mozillavpn.cpp
```diff
@@ -1381,6 +1381,11 @@ void MozillaVPN::registerUrlOpenerLabels() {
...
+    return "https://support.mozilla.org/en-US/kb/"
+           "how-do-i-change-my-dns-settings";
```
vinocher, line 1387: Should these strings be moved into Constants, like the other UrlLabels?

src/translations/strings.yaml
MattLichtenstein: @traceyvantyghem

traceyvantyghem: Yes, agree with Flod: "What is a custom DNS?" (Sorry about the delayed reply)

src/translations/strings.yaml
MattLichtenstein: @traceyvantyghem

nebula/ui/components/MZMenu.qml
MattLichtenstein: Hmm, null checks for what? `item`? If `item` is null, `item instanceof Text` will evaluate to `false`

nebula/ui/components/MZHelpSheet.qml
MattLichtenstein: It is the "X" button on the help sheet that is in the hovered state when the right menu button on the source screen is clicked.

Am I misunderstanding your comment?

src/translations/strings.yaml
lesleyjanenorton: +1 @vinocher . We don't seem to be answering the question from dnsHeader 'What is a _custom_ DNS?').

lesleyjanenorton (changes requested):

nebula/ui/components/MZHelpSheet.qml
vinocher: Yes, I was wondering it the mouse down and up are going to two different targets, causing this issue. I don't think anything needs to be addressed in this comment.

nebula/ui/components/MZMenu.qml
vinocher: Please ignore the null check in the comment.

src/translations/strings.yaml
traceyvantyghem: @MattLichtenstein I agree that updating the title makes sense (per Flod's suggestion). 

I didn't work on this content (and don't know much about the subject matter), so I checked in with Chris, but he didn't have any additional updates. Chris is returning to content lead on VPN, so I guess we can leave the body content as-is.

vinocher (approved):
Looks good!

lesleyjanenorton (approved):

Issue #8396, opened 2023-10-26
Long app name causes UI issues in App exclusions list


data-sync-user:
**Important details:**

* not sure if this is intended or not, but after renaming an app, the app is excluded from the App exclusions list;

**Affected versions:**

* 2.18.0 (2.202310241849)

**Tested Platforms:**

* Windows 10/11;

**Prerequisites:**

* Have Mozilla VPN app installed and open;
* Have a Mozilla account with an active subscription;
* Be signed in;  

**Steps to reproduce:**

1. On your device, rename any app that is part of Mozilla VPN's app exclusion list, to a longer name (about 80 chars);
1. Go to Mozilla VPN and navigate to Settings - App exclusions and add the renamed app from step 1;
1. Observe the screen;

**Expected result:**

* There are no UI issues for the app with the long name;

**Actual result:**

* There are UI issues -> the long name of the app overlaps with the app below;

**Notes:**

* See the attached ss;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-5771)
┆Reporter: Bianca Hidecuti

PR #8866, opened 2023-12-23
VPN-5771 - Fix UI issues with long app names in app exclusions (#8396)

janriokrause:
# Description

When adding an app with a long app name, the UI causes problems with overlapping text. Removing the fixed height solves this. I could not use Windows for testing, so I used the `dummyapplistprovider.cpp`.

# Reference

* https://mozilla-hub.atlassian.net/browse/VPN-5771
* #8396

# Screenshot

screenshot-vpn-5771

### DIFF
--- a/src/ui/screens/settings/appPermissions/AppPermissionsList.qml
+++ b/src/ui/screens/settings/appPermissions/AppPermissionsList.qml
@@ -40,7 +40,7 @@ ColumnLayout {
         id: col2
         objectName: "appExclusionsList"
 
-        spacing: MZTheme.theme.windowMargin / 2
+        spacing: MZTheme.theme.vSpacing
 
         Layout.preferredWidth: parent.width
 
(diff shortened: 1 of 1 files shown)


strseb:
Lol thanks Jan ♥️ 
@MattLichtenstein  quick r?

src/ui/screens/settings/appPermissions/AppPermissionsList.qml
```diff
@@ -40,7 +40,7 @@ ColumnLayout {
         id: col2
         objectName: "appExclusionsList"
 
-        spacing: MZTheme.theme.windowMargin / 2
+        spacing: MZTheme.theme.vSpacing
```
MattLichtenstein, line 43: since we are getting rid of the cell's height, I like the extra spacing between rows here, but this also adds extra spacing between the "clear all" button and the first item in the list. 

can we put the list (repeater) inside it's own layout so it can have separate spacing

MattLichtenstein (changes requested):

strseb:
closing as stale, feel free to reopen

PR #8875, opened 2023-12-28
VPN-5986 Add openssl-3 on Qt 6.4+

strseb:
## Description

So from Qt 6.4 we need to also bundle openssl 3.x.x. As before, i'm going to use kdab :) 
We're keeping the 1.x, as our rust builds seem to rely on those q_q 

Screenshot (02 01

### DIFF
--- a/src/cmake/android_openssl.cmake
+++ b/src/cmake/android_openssl.cmake
@@ -7,7 +7,6 @@
 
 include(ExternalProject)
 
-
 ExternalProject_Add(ndk_openssl
   URL https://maven.google.com/com/android/ndk/thirdparty/openssl/1.1.1q-beta-1/openssl-1.1.1q-beta-1.aar
   URL_HASH SHA256=a5b05c4b362d35c022238ef9b2e4e2196248adea3bac9dd683845ee75a3a8d66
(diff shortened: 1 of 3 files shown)


src/cmake/android_openssl.cmake
```diff
@@ -54,6 +56,26 @@ add_custom_command(
...
+find_package(Qt6 COMPONENTS Core)
+if( ${Qt6_VERSION} VERSION_GREATER_EQUAL 6.4.0)
+include(FetchContent)
+# Google does not provide openssl v 3.0
+# so let's use the builds from kdab
+  FetchContent_Declare(
+    android_openssl
+    DOWNLOAD_EXTRACT_TIMESTAMP true
+    URL      https://github.com/KDAB/android_openssl/archive/refs/heads/master.zip
+  #      URL_HASH MD5=c97d6ad774fab16be63b0ab40f78d945 #optional
```
brizental, line 72: Why is this commented out?

src/cmake/android_openssl.cmake
```diff
@@ -54,6 +56,26 @@ add_custom_command(
         COMMAND ${CMAKE_COMMAND} -E copy ${_OPENSSL_CRYPTO_MODULE}/libs/android.${ANDROID_ABI}/libcrypto.so ${_OPENSSL_LIBS}/libcrypto.so)
 
 
-get_property(crypto_module GLOBAL PROPERTY OPENSSL_CRYPTO_MODULE)
-get_property(ssl_module GLOBAL PROPERTY OPENSSL_SSL_MODULE)
-get_property(openssl_libs GLOBAL PROPERTY OPENSSL_LIBS)
+# In case of newer versions of QT, for QtSLL we need to 
+# have 3.x.x bundled alongside.
+# Some rust crates seem to need 1.x? so ugh. both i guess.
```
brizental, line 61: Did you check if this is a config thing or something we can update on the Rust side at all?

strseb: theorecially the openssl bindings should support that i think

PR #8873, opened 2023-12-27
VPN-4691: Add location selection help sheet

MattLichtenstein:
## Description

**NOTE: Depends on [#8864: VPN-4687: Add DNS settings help sheet](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8864)**

- Implement help sheet for location selection
- Remove collapsible help cards for recommended servers and multihop

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/1206ee69-9c9a-4a8e-a7a0-5ee416562bf5

## Reference

[VPN-4691: Implement the 'Location Selection' in-product help sheet](https://mozilla-hub.atlassian.net/browse/VPN-4691)

PR #8878, opened 2023-12-28
VPN-4684: Add privacy features help sheet

MattLichtenstein:
## Description

**NOTE: Depends on [#8864: VPN-4687: Add DNS settings help sheet](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8864)**

- Implement help sheet for privacy features

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/15353801/dfc3e63a-1d19-4397-ba8a-8022be18d534

## Reference

[VPN-4684: Implement the 'Privacy features' in-product help sheet](https://mozilla-hub.atlassian.net/browse/VPN-4684)


src/translations/strings.yaml
```diff
@@ -928,6 +928,19 @@ devices:
     value: My devices
     comment: Title for the menu bar when viewing the devices or device limit pages
 
+  privacyTitle:
+    value: Privacy features
+    comment: Title label for the privacy features help sheet
+  privacyHeader:
+    value: What do privacy features do for me?
```
flodolo, line 935: "What do…do" doesn't sound great?

"What can…do" looks like a minimal change here.

src/translations/strings.yaml
```diff
@@ -928,6 +928,19 @@ devices:
...
+    comment: Header label for the privacy features help sheet
+  privacyBody1:
+    value: Privacy features give you greater control over the types of content you see, helping you stay safe online.
+    comment: Body label for the privacy features help sheet
+  privacyBody2:
+    value: If you activate these features, they’ll overwrite any custom DNS you’re using. These protections are not a substitute for taking other security precautions—for example, while blocking malware you should still avoid downloading attachments in strange emails.
```
flodolo, line 941:
```suggestion
    value: If you activate these features, they’ll overwrite any custom DNS you’re using. These protections are not a substitute for taking other security precautions — for example, while blocking malware you should still avoid downloading attachments in strange emails.
```

flodolo (changes requested):

src/translations/strings.yaml
MattLichtenstein: @traceyvantyghem

src/ui/screens/settings/privacy/ViewPrivacy.qml
```diff
@@ -16,6 +16,25 @@ MZViewBase {
...
     objectName: "privacySettingsView"
 
+    property Component rightMenuButton: Component {
+        Loader {
+            active: MZFeatureList.get("helpSheets").isSupported
+            sourceComponent: MZIconButton {
+                onClicked: helpSheetLoader.active = true
+
+                accessibleName: MZI18n.GlobalHelp
+                Accessible.ignored: !visible
```
vinocher, line 26: MZButtonBase has `Accessible.ignored: !visible` this already. Same comment for related PRs.

src/translations/strings.yaml
```diff
@@ -928,6 +928,19 @@ devices:
...
+    comment: Title label for the privacy features help sheet
+  privacyHeader:
+    value: What do privacy features do for me?
+    comment: Header label for the privacy features help sheet
+  privacyBody1:
+    value: Privacy features give you greater control over the types of content you see, helping you stay safe online.
+    comment: Body label for the privacy features help sheet
+  privacyBody2:
+    value: If you activate these features, they’ll overwrite any custom DNS you’re using. These protections are not a substitute for taking other security precautions — for example, while blocking malware you should still avoid downloading attachments in strange emails.
+    comment: Body label for the privacy features help sheet
```
vinocher, line 942: Nit: Do we need to break up the body into two strings? Is it more flexible for localization if a single string (and a single Text field in the UI) is used? 

Feel free to ignore this comment.

flodolo: In this case, one string provides more flexibility for translation.

MattLichtenstein: For layout/spacing reasons, we have to use two text components here

src/translations/strings.yaml
traceyvantyghem: Yes, I think this is a good suggestion, especially because "do" implies they do it by default, but "can" is more aligned with the fact that users can turn things off or on here.

Anyway, I agree, I think we should update to "What can privacy features do for me?"

src/mozillavpn.cpp
```diff
@@ -1381,6 +1381,11 @@ void MozillaVPN::registerUrlOpenerLabels() {
   uo->registerUrlLabel("upgradeToAnnualUrl", []() -> QString {
     return Constants::upgradeToAnnualUrl();
   });
+
+  uo->registerUrlLabel("sumoPrivacy", []() -> QString {
+    return "https://support.mozilla.org/en-US/kb/"
```
lesleyjanenorton, line 1386: Same comment here, let's remove "en-US" so that moz.org can determine and reroute to the correct locale.

src/ui/screens/settings/privacy/ViewPrivacy.qml
```diff
@@ -106,5 +125,26 @@ MZViewBase {
...
+            title: MZI18n.HelpSheetsPrivacyTitle
+
+            model: [
+                {type: MZHelpSheet.BlockType.Title, text: MZI18n.HelpSheetsPrivacyHeader},
+                {type: MZHelpSheet.BlockType.Text, text: MZI18n.HelpSheetsPrivacyBody1, margin: 8},
+                {type: MZHelpSheet.BlockType.Text, text:MZI18n.HelpSheetsPrivacyBody2, margin: 16},
+                {type: MZHelpSheet.BlockType.LinkButton, text: MZI18n.GlobalLearnMore, margin: 16, action: () => { MZUrlOpener.openUrlLabel("sumoPrivacy") } },
+            ]
+
+            onClosed: helpSheetLoader.active = false
```
lesleyjanenorton, line 146: Do we anticipate these help sheets existing outside of Loaders? Any reason to not create a separate component for the Loader/HelpSheet duo? Seems like we could save ourselves a bit of duping.

lesleyjanenorton (changes requested):
Thanks Matt! Same URL comment here: We just need to remove the locale from the URL so that moz.org can route to the correct page based on the user's browser language setting.

src/ui/screens/settings/privacy/ViewPrivacy.qml
MattLichtenstein: agree. [see here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8873#discussion_r1457908929)

lesleyjanenorton (approved):

vinocher (approved):

flodolo (approved):

PR #8887, opened 2024-01-02
VPN-5594 Use a diffrent Settings File for testing and running the client!

strseb:
## Description

The Functional tests are using the same Settings file as the normal vpn client. 
This set's settings like "only use localhost requests" which means i always need to nuke that file after running tests. 
Let's have this file per-test suite.

### DIFF
--- a/src/settings/settingsmanager.cpp
+++ b/src/settings/settingsmanager.cpp
@@ -27,13 +27,12 @@ constexpr const char* SETTINGS_ORGANIZATION_NAME = "mozilla_testing";
 #endif
 
 #if defined UNIT_TEST
-  constexpr const char* SETTINGS_APP_NAME = "vpn_unit";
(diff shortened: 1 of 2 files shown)


brizental (approved):
!!! GO BASTI !!! 💯 🏆

brizental:
There is a ticket for this https://mozilla-hub.atlassian.net/browse/VPN-5594, btw.

src/settings/settingsmanager.cpp
```diff
@@ -27,6 +26,15 @@ constexpr const char* SETTINGS_ORGANIZATION_NAME = "mozilla";
 constexpr const char* SETTINGS_ORGANIZATION_NAME = "mozilla_testing";
 #endif
```
brizental, line 27: I thin this doesn't need to be different anymore now that the app name is different.

mcleinman (approved):
Thank you!!!

PR #8899, opened 2024-01-03
VPN-4307 Use registerAndReturnError to replace deprecated API for managing LoginItems

Gela:
## Description

Update the API we use to manage loginItems and startAtBoot. What we were using(`SMLoginItemSetEnabled()`) is deprecated on MacOS 13 and beyond and is replaced by `registerAndReturnError` and `unregisterAndReturnError`. To continue backwards compatibility, if the OS is older than MacOS 13 we still fall back to the old way of managing login items.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-4126
https://mozilla-hub.atlassian.net/browse/VPN-4307

## Checklist
    
### DIFF
--- a/src/platforms/macos/macosutils.mm
+++ b/src/platforms/macos/macosutils.mm
@@ -38,33 +38,38 @@ QString MacOSUtils::computerName() {
 void MacOSUtils::enableLoginItem(bool startAtBoot) {
   logger.debug() << "Enabling login-item";
 
-  NSString* appId = MacOSUtils::appId();
+  NSString * appId = MacOSUtils::appId();
   Q_ASSERT(appId);
 
-  NSString* loginItemAppId =
+  NSString * loginItemAppId =
     QString("%1.login-item").arg(QString::fromNSString(appId)).toNSString();
-  
+
   // For macOS 13 and beyond, register() and unregister() methods
   // are used for managing login items since SMLoginItemSetEnabled() is deprecated.
   // For versions prior to macOS 13, SMLoginItemSetEnabled() is used.
-    if (@available(macOS 13, *)) {
-      // Use register() or unregister() based on the startAtBoot flag
-      NSError *error = nil;
-      if (startAtBoot) {
-        if (![[SMAppService mainAppService] registerAndReturnError:&error]) {
-          logger.error() << "Failed to register Mozilla VPN LoginItem: " << error.localizedDescription;
-        } else {
-          logger.debug() << "Mozilla VPN LoginItem registered successfully.";
-        }
+  if (@available(macOS 13, *)) {
+    // Use register() or unregister() based on the startAtBoot flag
+    NSError * error = nil;
+
+    if (startAtBoot) {
+      if (![
+          [SMAppService mainAppService] registerAndReturnError: & error
+        ]) {
+        logger.error() << "Failed to register Mozilla VPN LoginItem: " << error.localizedDescription;
       } else {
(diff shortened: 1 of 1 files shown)


src/platforms/macos/macosutils.mm
```diff
@@ -42,11 +44,35 @@
...
+          logger.debug() << "Mozilla VPN LoginItem registered successfully.";
+        }
+      } else {
+        if (![[SMAppService mainAppService] unregisterAndReturnError:&error]) {
+          logger.error() << "Failed to unregister Mozilla VPN LoginItem: " << error.localizedDescription;
+        } else {
+          logger.debug() << "LoginItem unregistered successfully.";
+        }
+      }
+    } else
```
mcleinman, line 69: Did you consider using something like?
```
#if defined
...
#else
...
#endif
```
I think that's the pattern we mostly use elsewhere in the code.

mcleinman (commented):
Thanks!

src/platforms/macos/macosutils.mm
```diff
@@ -42,11 +44,35 @@
...
+  #endif
+  {
+    CFStringRef cfs = (__bridge CFStringRef)loginItemAppId;
+    Boolean ok = SMLoginItemSetEnabled(cfs, startAtBoot ? YES : NO);
```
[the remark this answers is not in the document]
Gela: In my personal testing it didn't seem to fail but according to VPN-4126 it sometimes fails to start at boot on MacOS 13+.

src/platforms/macos/macosutils.mm
```diff
@@ -42,11 +44,35 @@
...
-  CFStringRef cfs = (__bridge CFStringRef)loginItemAppId;
-
-  Boolean ok = SMLoginItemSetEnabled(cfs, startAtBoot ? YES : NO);
-  logger.debug() << "Result: " << ok;
+    QString("%1.login-item").arg(QString::fromNSString(appId)).toNSString();
+  
+  // For macOS 13 and beyond, register() and unregister() methods
+  // are used for managing login items since SMLoginItemSetEnabled() is deprecated.
+  // For versions prior to macOS 13, SMLoginItemSetEnabled() is used.
+  #if defined(__MAC_13_0) && (__MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_13_0)
```
vinocher, line 52: The #ifdef doesn't seem to make sense here, because this occurs at build time. We need to dynamically determine the version of the macOS on the user's machine, which is performed using `if (@available(macOS 13, *)) `

src/platforms/macos/macosutils.mm
```diff
@@ -42,11 +42,32 @@
...
-  CFStringRef cfs = (__bridge CFStringRef)loginItemAppId;
-
-  Boolean ok = SMLoginItemSetEnabled(cfs, startAtBoot ? YES : NO);
-  logger.debug() << "Result: " << ok;
+    QString("%1.login-item").arg(QString::fromNSString(appId)).toNSString();
+  
+  // For macOS 13 and beyond, register() and unregister() methods
+  // are used for managing login items since SMLoginItemSetEnabled() is deprecated.
+  // For versions prior to macOS 13, SMLoginItemSetEnabled() is used.
+    if (@available(macOS 13, *)) {
```
vinocher, line 50: Nit: The alignment of the if and else statments and some of the blocks below need to be changed.

vinocher (approved):

PR #8911, opened 2024-01-05
VPN-5637: Enable annual plan upgrade in production

MattLichtenstein:
## Description

- Turn on the annual upgrade feature flag for the production environment

## Reference

[VPN-5637: Turn the VPN plan upgrades feature ON](https://mozilla-hub.atlassian.net/browse/VPN-5637)

### DIFF
--- a/src/feature/featurelistcallback.h
+++ b/src/feature/featurelistcallback.h
@@ -41,9 +41,6 @@ bool FeatureCallback_iosOrAndroid() {
 // -------------------------
 
 bool FeatureCallback_annualUpgrade() {
-//  if (Constants::inProduction()) {
-//    return false;
-//  }
   if (FeatureCallback_iosOrAndroid()) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


src/feature/featurelistcallback.h
```diff
@@ -41,9 +41,9 @@ bool FeatureCallback_iosOrAndroid() {
 // -------------------------
 
 bool FeatureCallback_annualUpgrade() {
-  if (Constants::inProduction()) {
-    return false;
-  }
+//  if (Constants::inProduction()) {
+//    return false;
+//  }
```
MattLichtenstein, line 46: Commented instead of removed to more easily revert if needed

lesleyjanenorton: Let's remove it. It's not much additional effort to revert this patch or rewrite the block later if we need to.

lesleyjanenorton (approved):
R+ with 1 ask :) Thanks Matt!

PR #8937, opened 2024-01-10
VPN-5871: Passwordless auth: string update

jswinarton:
## Reference

Fixes changes requested on [VPN-5871](https://mozilla-hub.atlassian.net/browse/VPN-5871).

## Checklist
    
- [ ] I have commented my code PARTICULARLY in hard to understand areas (N/A)
- [ ] I have added thorough tests where needed (N/A)

[VPN-5871]: https://mozilla-hub.atlassian.net/browse/VPN-5871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -705,7 +705,7 @@ inAppAuth:
   ssoInstructionLine1:
     value: "To finish setting up your Mozilla account, create a password in your account settings:"
     comment: User has tried to log in using an SSO-linked account that does not have a password. The user needs to log in to their Mozilla Account to set a password before proceeding.
-  ssoInstructionLine2-2:
+  ssoInstructionLine2a:
     value: 1. Go to <a href="open-account">Create password</a> in your Mozilla account settings.
     comment: User has tried to log in using an SSO-linked account that does not have a password. The user needs to log in to their Mozilla Account to set a password before proceeding. Do not remove or translate the markup '<a href="open-account">'.
   ssoInstructionLine3:
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -706,7 +706,7 @@ inAppAuth:
     value: "To finish setting up your Mozilla account, create a password in your account settings:"
     comment: User has tried to log in using an SSO-linked account that does not have a password. The user needs to log in to their Mozilla Account to set a password before proceeding.
   ssoInstructionLine2:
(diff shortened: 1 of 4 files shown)


src/translations/strings.yaml
```diff
@@ -706,7 +706,7 @@ inAppAuth:
     value: "To finish setting up your Mozilla account, create a password in your account settings:"
     comment: User has tried to log in using an SSO-linked account that does not have a password. The user needs to log in to their Mozilla Account to set a password before proceeding.
   ssoInstructionLine2:
```
flodolo, line 708: You need to update the string ID here.
https://mozilla-l10n.github.io/documentation/localization/making_string_changes.html

flodolo (changes requested):

src/translations/strings.yaml
flodolo: Yes, that's the only way to make sure we have the string retranslated. Maybe adding `a` (`ssoInstructionLine2a`) works a bit better than adding `-1`?

jswinarton: Thanks! After discussing with the team I ended up changing it to `ssoInstructionLine2-2` because this is the convention we are using elsewhere.

MattLichtenstein (approved):

src/translations/strings.yaml
```diff
@@ -705,8 +705,8 @@ inAppAuth:
   ssoInstructionLine1:
     value: "To finish setting up your Mozilla account, create a password in your account settings:"
     comment: User has tried to log in using an SSO-linked account that does not have a password. The user needs to log in to their Mozilla Account to set a password before proceeding.
-  ssoInstructionLine2:
-    value: 1. Go to <a href="open-account">Change password</a> in your Mozilla account settings.
+  ssoInstructionLine2-2:
+    value: 1. Go to <a href="open-account">Create password</a> in your Mozilla account settings.
```
MattLichtenstein, line 709: forgot to update the id in the UI as well

MattLichtenstein (changes requested):

src/translations/strings.yaml
jswinarton: Actually, turns out hyphens are not allowed by QML, so I have taken your suggestion.

MattLichtenstein (approved):

flodolo (approved):

PR #4958, opened 2022-11-17
Build VPN client as a Flatpak

oskirby:
## Description
To make the VPN client viable for installation on a few more platforms that we don't have native packages for, we can now build the client as a Flatpak. This is not ready for users just yet and there may be plenty of undiscovered breakage, but as a proof of concept it does function.

Because Flatpaks run in a sandbox, we had to do away with the `linuxdaemon` and replace it with a new controller using the `org.freedesktop.NetworkManager` portal to setup the Wireguard tunnel. This gives us less control over the network stack meaning that we may not be able to implement split tunnelling or native multihop support, but it does mean that we can run without root permissions.

I feel kind of gross by depending on `libnm` and having to convert between `glib` and `Qt` programming models, there's a small part of me that wants to throw it out and talk to NetworkManager directly using `QtDbus`

Some stuff that doesn't work, and probably should:
- [x] System tray icon is not shown.
- [x] Settings are not saved across restarts.
- [x] `m_excludedAddresses` is not honoured by the controller.
- [ ] Closing the flatpak does not always remove the connection.
- [x] Launching the application twice does not wake up backgrounded clients.
- [x] `LinuxPingSender` can send and receive pings.
- [ ] Generate signal on Wireguard handshake completion (would be very nice for multihop support).

## Reference
Github issue #1014 ([VPN-48](https://mozilla-hub.atlassian.net/browse/VPN-48))

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

Issue #1014, opened 2021-05-12
Flatpak package for Linux


rugk:
It would be nice if you could publish this as a [flatpak](https://flatpak.org/) on [flathub](https://flathub.org/) e.g.

Flatpaks are a new software distribution mechanism for Linux distros, can thus installed on any distro and are easy to update. 

Here is [how to get started](http://docs.flatpak.org/en/latest/getting-started.html).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-48)

lolrepeatlol:
Would it be possible to have a Flatpak be a VPN since it's sandboxed?

rugk:
I'm not sure and very likely depends on how it's done and how it integrates into the system. Online I could only find https://linuxpunx.com/2021/04/05/flatpak-and-whatip/, which seems to indicate it can at least have access to the network interfaces.

andersk:
FYI, I’m packaging `mozillavpn` for Nix at NixOS/nixpkgs#147975. (Nix is unrelated to Flatpak, but satisfies a similar sort of use case: it installs on any Linux distro without relying on or interfering with system packages.)

HiruNya:
I tried my hand at this but unfortunately I got stuck at the requirement for PolKit which [doesn't seem to be supported](https://lists.freedesktop.org/archives/flatpak/2018-August/001249.html) (that link is from a while ago so I'm not sure if anything has changed since then). I don't know enough about how a VPN and PolKit work so I gave up :sweat_smile: . Depending on what Mozilla VPN needs root privileges for, you might be able to replace it with [Portals](https://docs.flatpak.org/en/latest/sandbox-permissions.html#portals).

ahjolinna:
any hope? or is this dead/abandoned?

herzenschein:
I've been testing the current flatpak based on 6.6 and it works well.

In case others aren't aware of this, you can build the flatpak by navigating to the linux/flatpak folder and either:

```bash
# Building in one go
flatpak-builder build --install-deps-from=flathub --force-clean --ccache --install org.mozilla.vpn.yml
# Building the flatpak manually
flatpak-builder flatpak-build-dir --force-clean --install-deps-from flathub --repo=mozillavpn-master org.mozilla.vpn.yml
flatpak-builder --run flatpak-build-dir org.mozilla.vpn.yml mozillavpn # This just tests whether the flatpak runs
[...]

PR #8939, opened 2024-01-10
Flatpak support, round two

oskirby:
## Description
Every distro, yea-ah
Build your software, yea-ah
Every distro, yea-ah
Build your software right
Flatpak's back, alright!

With a bit of free time, I decided to make another attempt to get a working flatpak build into shape. This PR attempts to add a Github CI job to build a flatpak and to get a starting point of functionality working.

Some things which don't work, and are left for future work:
- Start on boot
- App exclusions
- Multi-account containers
- Building on taskcluster
- Pushing to flathub
- Handshake timing for true multihop.
## Reference
Github issue: #1014 ([VPN-48](https://mozilla-hub.atlassian.net/browse/VPN-48))
Previous attempt: #4958

## Checklist
    
[VPN-48]: https://mozilla-hub.atlassian.net/browse/VPN-48?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lesleyjanenorton:
So put your 'wares up in the air?

.github/workflows/linux.yaml
```diff
@@ -41,6 +42,48 @@ jobs:
           name: Sources
           path: .tmp
 
+  flatpak:
```
strseb, line 45: Love it, however could we consider moving this to tc from the get-go?
Releng already has support to release to flathub, so we could use that from the getgo 
https://github.com/mozilla-releng/scriptworker-scripts/tree/master/pushflatpakscript

oskirby: I've given it a solid try, but unfortunately we can't get flatpaks building in TC without being able to stand up a privileged container, which will require some work on releng's part. This is due to the need to create sandboxed environments in which the build occurs and is not posssible without extra permissions. (See: https://github.com/flatpak/flatpak/issues/3027)

That being said, it should be pretty easy to push to flathub even without building. IIRC that just amounts to updating a git repository with the YAML manifests.

lesleyjanenorton (approved):

PR #8956, opened 2024-01-15
VPN-5855 - Update TaskGetFeatureList to use new Guardian endpoint and scheduling

brizental:
The `/featurelist` Guardian enpoint is being update to also return experimental features from Nimbus, see https://github.com/mozilla-services/guardian-website/pull/1688. 

This pull request updates TaskGetFeatureList to use the new endpoint. The new enpoint 

1. is a POST endpoint, 
2. expects either an Authorization header from which to derive the experimenter id or the experimenter id to be in the body of the request

This pull request also updates the scheduling of the calls to this endpoint. Previously this endpoint would only be called once the user was logged in, on a timer. Now it should be called

1. As soon as possible when the application is started
2. Regardless of being logged in or not
3. Periodically
4. Whenever the experimenter id changes

#### About the experimenter id

The experimenter id is used by Nimbus to calculate the experimental features that should be enable and with which parameters. 

As described in https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/rfcs/0006-cirrus-experiments.md, we will use the FxA id as the experimenter id. However, if the user is logged out we will generate a random UUID and use that instead.

Therefore, whenever the user logged in state changes TaskGetFeatureList needs to be re-run with a different experimenter id. Note that the way this is implemented, the logged out experimenter id is only calculated once i.e. it doesn't get recalculated everytime the user logs out. As I write this I realize we should probably rotate it as well when telemetry is disabled -- will send a commit with that.


tests/unit_tests/feature/testtaskgetfeaturelistworker.cpp
```diff
@@ -0,0 +1,162 @@
...
+      qDebug() << updatedListOfTasks.count();
+      // If it is not a nullptr, the cast was succesfull.
+      QVERIFY(task);
+    }
+
+    // Update the initial list of tasks before destruction of the worker.
+    initialListOfTasks = TaskScheduler::tasks();
+  }
+
+  qDebug() << "Heer!";
```
brizental, line 143: TODO: Remove debug log.

src/feature/taskgetfeaturelistworker.h
```diff
@@ -0,0 +1,56 @@
...
+ * to determine which experimental features should be enabled for this user.
+ *
+ * On a logged out session of the VPN app, that will be a randomly generated
+ * UUID stored under the unauthedExperimenterId setting.
+ *
+ * On a logged in session of the VPN app, that will be the FxA id which is not
+ * passed in the request in plain text, but is gleaned from the JWT user token
+ * sent to the server through the Authorization header.
+ */
+class TaskGetFeatureListWorker final : public QObject {
```
vinocher, line 33: The naming convention in this codebase assumes that classes which start with Task* are tasks that are derived from Task. Consider if something like FeatureListTaskScheduler or GetFeatureListTaskScheduler are better names.

src/networkrequest.h
```diff
@@ -46,7 +50,9 @@ class NetworkRequest final : public QObject {
 
   QNetworkRequest& requestInternal() { return m_request; }
 
-  void auth(const QByteArray& authorizationHeader);
+  static const QByteArray authorizationHeader();
+
+  void auth(const QByteArray& authorizationHeader = "");
```
vinocher, line 55: This is old code, but should this have been named setAuthHeader() or setAuthorizationHeader()? Otherwise it is confusing at the call site.

tests/unit_tests/feature/testtaskgetfeaturelist.cpp
```diff
@@ -0,0 +1,148 @@
...
+  settingsHolder.setToken("aToken");
+
+  TaskGetFeatureList task;
+  task.run();
+
+  QVERIFY(featureListRequest.hasRawHeader("Authorization"));
+  QCOMPARE(featureListRequestBody, "{}");
+
+  // Reset request handlers.
+  NetworkRequest::resetRequestHandler();
```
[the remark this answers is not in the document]
brizental: Hm I guess not really, we could just call this on teardown of this test suite. However, given we are stubbing the request handler inside these tests, I'd rather reset it before leaving the test case itself so as not to leak state from this test into other tests.

tests/unit_tests/feature/testtaskgetfeaturelistworker.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "feature/taskgetfeaturelistworker.h"
+#include "settingsholder.h"
+#include "task.h"
+#include "taskscheduler.h"
+
+void TestTaskGetFeatureListWorker::testTaskIsScheduledOnStart() {
+  SettingsHolder settingsHolder;
+
+  // Stop the TaskScheduler so we can inpect the tasks scheduled.
+  TaskScheduler::stop();
```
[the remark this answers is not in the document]
brizental: That is a great call. I completely disconsidered messing with the event loop itself and jumped straight into creating this method. I'll change this.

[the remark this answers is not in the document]
brizental: Hm, what you are saying makes sense. However if I remove the `stop` the tests stop passing. The task is executed before we even get to the assertions. I'll dig a little bit on this to understand.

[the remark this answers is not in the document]
brizental: Ok so, I believe this task is being run synchronously. See: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/taskscheduler.cpp#L59-L62

The task is being added to the queue and immediatelly getting executed. 

The `stop` here is required, so I am leaving it. Let me know if you think there is a better way to do this.

src/feature/getfeaturelistscheduler.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+GetFeatureListScheduler::GetFeatureListScheduler(QObject* parent)
+    : QObject(parent) {
+  MZ_COUNT_CTOR(GetFeatureListScheduler);
+}
+
+GetFeatureListScheduler::~GetFeatureListScheduler() {
+  MZ_COUNT_DTOR(GetFeatureListScheduler);
+
+  qDebug() << "Destructing!";
```
vinocher, line 20: Shouldn't logger be used here?

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+void TestGetFeatureListScheduler::testTaskIsScheduledOnStart() {
+  SettingsHolder settingsHolder;
+
+  // Stop the TaskScheduler so we can inpect the tasks scheduled.
+  TaskScheduler::stop();
+
+  auto initialListOfTasks = TaskScheduler::tasks();
+
+  GetFeatureListScheduler worker;
```
vinocher, line 23: Nit: Consider renaming worker to scheduler

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
```diff
@@ -0,0 +1,160 @@
...
+  // give the worker time to schedule the first periodic task.
+  QTest::qWait(250);
+
+  auto updatedListOfTasks = TaskScheduler::tasks();
+
+  // We are expecting two tasks, the task scheduled on start + 1.
+  QCOMPARE(initialListOfTasks.count() + 2, updatedListOfTasks.count());
+
+  // Check that the tasks are TaskGetFeatureList, just in case.
+  // Check that the tasks are TaskGetFeatureList, just in case.
```
vinocher, line 62: Nit: Remove duplicate line, here and elsewhere in this file

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
```diff
@@ -0,0 +1,160 @@
...
+    initialListOfTasks = TaskScheduler::tasks();
+  }
+
+  // Sleep more 250, if the timer hasn't stopped this will schedule more
+  // tasks. But it shouldn't.
+  QTest::qWait(250);
+
+  // Also change the token, this should also not schedule any tasks unless
+  // the signal connection was done incorrectly.
+  settingsHolder.setToken("aToken");
```
vinocher, line 149: Should this token be removed in test cleanup?

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
```diff
@@ -0,0 +1,160 @@
...
+  TaskScheduler::stop();
+
+  auto initialListOfTasks = TaskScheduler::tasks();
+
+  GetFeatureListScheduler worker;
+  worker.start(1 * 1000 * 60 * 60);  // 1hr
+
+  settingsHolder.setToken("aToken");
+  settingsHolder.setToken("anotherToken");
+  settingsHolder.removeToken();
```
vinocher, line 88: Should tests restore the original token, so it doesn't affect the machine that the test is running on?

tests/unit_tests/feature/testtaskgetfeaturelistworker.cpp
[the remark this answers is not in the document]
vinocher: Yes, the scheduler will schedule synchronously if there isn't a pending task, so this seems to be needed. Thanks, and please ignore my comment.

vinocher (approved):
Looks good!

src/feature/getfeaturelistscheduler.cpp
brizental: Dang, good catch. This is a debug log.

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
brizental: The settings holder is destructed at the end of this test case and that takes care of it already.

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
brizental: Same resply as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8956#discussion_r1466512725

brizental:
Merging this is blocked until https://github.com/mozilla-services/guardian-website/pull/1688 is merged and released to staging.

tests/unit_tests/feature/testgetfeaturelistscheduler.cpp
vinocher: But doesn't setToken write the token to the backing settings file? How does the SettingHolder destruction help with that?

brizental: Correct, however the SettingsHolder destructor will also clean that for tests: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/settingsholder.cpp#L64-L68. This is a bit hacky, but I did it so I did not have to touch most tests that already expect this to be the default behaviour.

strseb (approved):

PR #8971, opened 2024-01-17
VPN-6020: null check telemetryScreenId in MZSignOut

MattLichtenstein:
## Description

- Only record `signOutSelected` events from the settings menu, instead of from all sign out buttons app-wide
- Fixes broken sign out button (button did not do anything) on other screens (subscription not found, subscription is linked with another account, update view) caused by referencing an undefined variable

## Reference

[VPN-6020: [iOS] “Sign out” button from “Problem confirming subscription” error screens is not working](https://mozilla-hub.atlassian.net/browse/VPN-6020)

### DIFF
--- a/nebula/ui/components/MZSignOut.qml
+++ b/nebula/ui/components/MZSignOut.qml
@@ -14,7 +14,7 @@ MZFooterLink {
 
     onClicked: () => {
         preLogoutCallback();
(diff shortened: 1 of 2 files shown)


Gela:
Are we okay with changing the behavior to only collect sign out telemetry from Settings only? What about the rest?

lesleyjanenorton (approved):

PR #8972, opened 2024-01-17
VPN-5932: Change where we save a new device model

mcleinman:
## Description

This bug existed for the following situation:
- Have an account with five devices.
- Sign in on a new device
- Remove a device from the account as part of the sign in flow
- End up on main screen
- Force kill app
- Reopen it
- Note that you need to sign in again - not what we expected

The list of valid devices comes from the Guardian, and is saved in local settings (as the device list JSON). Upon launch, we check if the current device is in those devices, and if not, we ask the user to sign in again.

When we sign into the app when an account already has max devices, we retrieve the device list from Guardian twice - once upon initial launch, and a second time after we remove a device and add our current one. *However, we were only saving it to settings after that first retrieval.* Thus, on subsequent launch, the current device wasn't in the saved list, signing the user out.

I updated where we save the device JSON to local settings - It's now in the function where we process the new device list from Guardian. This fixes the bug, and hopefully prevents future iterations of this bug.

Note that we save the user data in the same spot, but I am not updating it as part of this bug. Arguably, we should save it to disk in the function where we process the server response, as we're changing devices here. However, there are no known issues with it currently and I'm trying to keep this patch as small as possible.

## Reference

VPN-5932

## Checklist
    
### DIFF
--- a/src/models/devicemodel.cpp
+++ b/src/models/devicemodel.cpp
@@ -51,6 +51,7 @@ bool DeviceModel::fromJson(const Keys* keys, const QByteArray& s) {
   }
 
   m_rawJson = s;
+  writeSettings();
   return true;
 }
 
--- a/src/models/devicemodel.cpp
+++ b/src/models/devicemodel.cpp
@@ -51,7 +51,9 @@ bool DeviceModel::fromJson(const Keys* keys, const QByteArray& s) {
   }
 
   m_rawJson = s;
+#if !defined(UNIT_TEST)
   writeSettings();
+#endif
   return true;
 }
 
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -535,7 +535,6 @@ void MozillaVPN::completeAuthentication(const QByteArray& json,
   }
 
   m_private->m_user.writeSettings();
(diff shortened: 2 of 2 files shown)


src/models/devicemodel.cpp
```diff
@@ -51,6 +51,9 @@ bool DeviceModel::fromJson(const Keys* keys, const QByteArray& s) {
   }
 
   m_rawJson = s;
+#if !defined(UNIT_TEST)
+  writeSettings();
```
vinocher, line 55: As this is now a side effect of reading the device model from the Json, the previous code that used writeSettings() in the caller appears to be a better design.

AFAIK, DeviceModel::fromJson is called from MozillaVPN::completeAuthentication (with the list of devices without the current device) and MozillaVPN::accountChecked (with the list of devices including the current device), and both of them called writeSettings. So, I don't understand the cause of the bug. Could you please clarify if I'm missing something?

mcleinman:
Closing this in favor of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9011. Thanks to @vinocher's comments, I believe I found a better solution.

PR #8976, opened 2024-01-18
"VPN-6037 - fix de-serialisation of canned notifications" 

strseb:
## Description

This is a really embarrasing bug lol. The data class for the notification has the requestScreen as `String?` however in the json we assume `String!`, therefore it throws if we do not pass that. 

Sooooo given this is clearly something a unit test would have catched, let's add junit test support on tc.

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
@@ -189,7 +189,7 @@ data class CannedNotification(
                         messages.getString("disconnectedBody"),
                     ),
                     messages.getString("productName"),
-                    messages.getString("requestedScreen"),
+                    messages.optString("requestedScreen"),
                 )
             } catch (e: Exception) {
                 Log.e("NotificationUtil", "Failed to Parse Notification Object $value")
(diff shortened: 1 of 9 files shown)


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
```diff
@@ -189,7 +189,7 @@ data class CannedNotification(
                         messages.getString("disconnectedBody"),
                     ),
                     messages.getString("productName"),
-                    messages.getString("requestedScreen"),
```
strseb, line 192: this is the fix lol

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/NotificationUtil.kt
```diff
@@ -189,7 +189,7 @@ data class CannedNotification(
...
+                    messages.optString("requestedScreen"),
```
strseb, line 192: this is the fix lol

brizental (approved):
r+, the task is not passing -- but I am sure you'll figure that out in a bit.

Two more thoughts, that can be follow up work: we need to add documentation on how to run these tests locally and/or make them runnable with ctest.

PR #8978, opened 2024-01-18
Do not deactivate if the VPN is on and server becomes unavailable

Gela:
## Description

Do not deactivate if the VPN is in StateOn or SilentSwitch and the server location becomes unavailable. Because this the user can miss that the VPN was deactivated which will leave them unprotected. Instead stay in kill switch mode until the issue is resolved. 

Follow up issue to write a functional test for this: https://mozilla-hub.atlassian.net/browse/VPN-6117

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-6033

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -290,8 +290,7 @@ void Controller::serverUnavailable() {
 
   m_nextStep = ServerUnavailable;
 
-  if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting ||
+  if (m_state == StateSwitching || m_state == StateConnecting ||
       m_state == StateConfirming || m_state == StateCheckSubscription) {
     deactivate();
     return;


src/controller.cpp
```diff
@@ -290,8 +290,7 @@ void Controller::serverUnavailable() {
 
   m_nextStep = ServerUnavailable;
 
-  if (m_state == StateOn || m_state == StateSwitching ||
-      m_state == StateSilentSwitching || m_state == StateConnecting ||
+  if (m_state == StateSwitching || m_state == StateConnecting ||
```
lesleyjanenorton, line 293: Thank you for taking this on @Gela! Two things: 
- Do we know how we enter into this server unavailable routine while in `NoSignal` and should we add logic so that we _don't_ go into `serverUnavailable()` while in `NoSignal`?  
- Is there value in deactivating here at all? Perhaps we can stipulate that deactivation must be user initiated and remove the block entirely.

src/controller.cpp
```diff
@@ -290,8 +290,7 @@ void Controller::serverUnavailable() {
 
   m_nextStep = ServerUnavailable;
 
-  if (m_state == StateOn || m_state == StateSwitching ||
```
vinocher, line 293: What is the new user experience after this change? I think the user will need to go to the app, toggle the VPN and try again. Is that correct?

Gela: Yes that's correct

src/controller.cpp
Gela: > Do we know how we enter into this server unavailable routine while in NoSignal and should we add logic so that we don't go into serverUnavailable() while in NoSignal?

I'm not very clear on how/why we go into `NoSignal` and if that's intended. I'm not sure how to reproduce that scenario to debug currently.

> Is there value in deactivating here at all? Perhaps we can stipulate that deactivation must be user initiated and remove the block entirely.

If we're in StateConnecting and the server location becomes unavailable, I think it makes sense to deactivate as opposed to remaining in the Connecting... state. Similarly in StateSwitching the VPN has to deactivate to switch servers anyway. I think it's fair here to deactivate since the VPN wouldn't have been active anyway. What are your thoughts?

src/controller.cpp
Gela: They get a message at the top of the client that let's them know that the server location has become unavailable. This happens before they even deactivate. 

I added some log statements to capture what's happening

oskirby:
I think this is actually the better approach, given that the silent server switching can happen autonomously in the background and it is better to make the problem visible in the form of a lost internet connection rather than leaving the user suddenly unprotected (eg: if they have the VPN in the background).

src/controller.cpp
oskirby: To test the `NoSignal` state in the past I have done things to break the connection to the wireguard server, for example:

1. Disconnect Wi-Fi/Ethernet and then re-connect after it tries server switching. 
2. Locate the server's IP and then install a `/32` route to drop all packets to it....

  - Linux: `ip route add blackhole <server-ip>`
  - MacOS: `route -v add -net <server-ip>/32 127.0.0.1 -blackhole`

oskirby (approved):

src/controller.cpp
```diff
@@ -290,12 +290,15 @@ void Controller::serverUnavailable() {
...
-      m_state == StateSilentSwitching || m_state == StateConnecting ||
+  if (m_state == StateSwitching || m_state == StateConnecting ||
       m_state == StateConfirming || m_state == StateCheckSubscription) {
+    logger.info() << "Server location is unavailable and we are not in "
+                     "StateOn. Deactivate!";
     deactivate();
     return;
   }
+  logger.info() << "Server location is unavailable. Do not deactivate the VPN "
```
vinocher, line 300: Consider adding a comment explaining why we aren't deactivating.

vinocher (approved):
FYI, an additional comment was added.

Bobinsz:
My confidence in the VPN has risen, after reading through the steps coders and code reviwers have been taking... with Logical reasoning and what seems to be a thorough look at what was causing the Kill-Switch to deactivate and allow the net connection to revert to unprotected traffic, without user interaction.

I've continued to have the occational complete discconnection, with the pop-up server unavaiable whilst running 2.22.1 (2.202405141843) )

If i'm honest, as a user I'm actually more frustrated by the connection timer changes that have occured over the past few updates... The "Secure & Private" timer no longer resets upon a reconnection attempt... and seems to just show the Total time the VPN has been active for.

I've had unstable traffic whilst using the VPN, without the Kill-Switch failure that I initially raised... but at a simple glance, there isn't an easy way to see if the VPN is constantly trying to establish a stable connection with this new "Secure & Private" timer behaviour.

PR #8984, opened 2024-01-19
Fix JUnit testing for android

strseb:
## Description
This pr attempts to enable running the unit tests written in #8976. 
I originally tried to get a build artifact from the android-build task and then run the test suite against that, 
however given how glean on android works, i was unable to create a portable artifact 😭  - so for now i'm just building the project as normal.


brizental (approved):
I'd suggest filing a ticket for investigating not having to rebuild the app for tests and instead using the built stuff from other tasks. WDYT?

PR #9013, opened 2024-01-24
VPN-6129 - Create daemon-specific metrics

mcleinman:
## Description

The connection health metrics all have a ping lifetime. However, in [this PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7124/) they were added to the daemon ping - and previously, two of them were already on both `main` and `session` pings. Thus, they're being misreported - for example, we increment one metric (`connection_health_stable_count`) for both Android daemon and app session health checks - and clear it whenever either ping is sent.

This PR adds new variables for the daemon, and removes the existing metrics from `main` and `daemon` pings.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-6129

## Checklist
    
### DIFF
--- a/src/telemetry/metrics.yaml
+++ b/src/telemetry/metrics.yaml
@@ -1762,7 +1762,7 @@ session:
       - vpn-telemetry@mozilla.com
     expires: never
 
-  daemon_connection_health_no_signal:
+  daemon_conn_health_no_signal:
     type: event
     lifetime: ping
     send_in_pings:
@@ -1788,7 +1788,7 @@ session:
       - vpn-telemetry@mozilla.com
     expires: never
 
-  daemon_connection_health_unstable:
+  daemon_conn_health_unstable:
     type: event
     lifetime: ping
     send_in_pings:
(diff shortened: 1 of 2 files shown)


mcleinman:
This needs a data steward review. However, I want to make sure this approach is agreed upon before requesting that review. Thus, keeping this on draft for now.

@brizental , despite being on draft, please review - I want to make sure this approach makes sense.

brizental:
> and clear it whenever either ping is sent.

This is incorrect. Even though the metric is lifetime ping, there are different instances of each metric per ping so that they are cleared based on the lifetime of each ping separately. I am not sure what is being misreported about these metrics, but I am skeptical that this would fix it.

src/telemetry/metrics.yaml
```diff
@@ -1776,3 +1761,73 @@ session:
       - mcleinman@mozilla.com
       - vpn-telemetry@mozilla.com
     expires: never
+
+  daemon_conn_health_no_signal:
```
brizental, line 1765: Creating these new metrics would have the same effect as sending the same metric is multiple pings, it just makes the metric design a bit repetitive.

brizental (changes requested):

mcleinman:
Thanks. Closing this, pending more investigation.

PR #9017, opened 2024-01-24
Run all tests in testSubscription

MattLichtenstein:
## Description

- Looks like I forgot to remove some debug/test code in eef4311 which prevents all other tests in the `testSubscriptions` suite from running 😬 

## Reference

N/A


brizental (approved):
AAAAAAAAAAAAAAA We need a lint for this.

PR #9022, opened 2024-01-25
Experiment: Set codecov upload token

oskirby:
## Description
In doing a bit of investigation on code coverage tooling, I took another look at what's going on with codecov, and it seems that we are missing upload data quite often due to rate limiting of the github API. It seems that there is a known workaround for this simply by setting a codecov upload token.

So, let's find out what happens when we set it.

## Reference
[Codecov issue](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954)

## Checklist
    
### DIFF
--- a/.github/workflows/linux_tests.yaml
+++ b/.github/workflows/linux_tests.yaml
@@ -164,6 +164,7 @@ jobs:
         if: steps.generateGrcov.outcome == 'success'
         with:
           directory: .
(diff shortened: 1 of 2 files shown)


oskirby:
Well, the 404 errors went away, but the coverage report still fails to produce a meaningful diff.

PR #9028, opened 2024-01-27
Update RPM build to Fedora 39

devurandom:
Fedora 37 is EOL since 2023-12-05 (cf. https://docs.fedoraproject.org/en-US/releases/eol/).

Current builds (tried e.g. `mozillavpn-2.20.0~build20240125-1.x86_64`
downloaded from https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/7661282763)
depends upon `libQt6Network.so.6(Qt_6.5_PRIVATE_API)(64bit)` but since
Fedora 39 is on Qt 6.6 no current package provides that symbol.

Also update from YUM to DNF:
* https://fedoraproject.org/wiki/Yum_to_DNF_Cheatsheet
* https://www.redhat.com/sysadmin/create-rpm-package

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -46,7 +46,7 @@ jobs:
     needs: source-bundle
     runs-on: ubuntu-latest
     container:
-      image: fedora:37
+      image: fedora:39
 
     steps:
       - name: Download Source Package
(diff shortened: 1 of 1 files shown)


devurandom:
Watching the builds I noticed that some other job (outside GitHub Actions) already builds an RPM for Fedora 38, i.e. against Qt 6.6, which also works fine with Fedora 39: https://firefox-ci-tc.services.mozilla.com/tasks/Jyfj9U09TxeC0CpEzH0Vqg
Maybe it makes sense to completely remove https://github.com/mozilla-mobile/mozilla-vpn-client/actions/workflows/linux.yaml and point people to that other service?

oskirby (approved):

PR #9069, opened 2024-02-02
Adding logging to investigate VPN-5766

mcleinman:
## Description

From prior and current investigation, iOS is crashing somewhere in this function. We don't have solid repro steps, though it's happening often enough to know we'll get a few samples a week from QA. These log lines are to start narrowing in on where the issue is. After we get crash logs from QA, we'll remove these log lines (and possibly add more logging in to figure out exactly where the problem lies). Then we'll fix it.

## Reference

VPN-5766 (sorta)

## Checklist
    
### DIFF
--- a/ConnectionHealth.swift
+++ b/ConnectionHealth.swift
@@ -1,0 +1,8 @@
+//
+//  ConnectionHealth.swift
+//  compat
+//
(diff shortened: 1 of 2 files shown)


brizental (approved):
Consider logging the errno on line 113 as well.

PR #9070, opened 2024-02-02
VPN-5466 - Bring back LAN Bypass setting on iOS

brizental:
This is here mostly for illustration purposes for now, therefore it's a draft and doesn't really require a review.

Notes:

1. CheckBox enabled setting is not doing anything. @MattLichtenstein anything that immediatelly pops out? TBH, I did not deeply investigate it given the state of this PR, but if already know what it is, good.
2. Do we want to add telemetry to the new button?
3. Can't write tests for this, because it's iOS and we don't run tests there yet.
4. A lot of copy paste from https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5096

Screenshot: 
Screenshot 2024-02-02 at 13 05 19

### DIFF
--- a/src/platforms/ios/ioscontroller.swift
+++ b/src/platforms/ios/ioscontroller.swift
@@ -116,7 +116,7 @@ public class IOSControllerImpl : NSObject {
         }
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
+    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  excludeLocalNetworks: Bool, allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
         IOSControllerImpl.logger.debug(message: "Connecting")
 
         TunnelManager.withTunnel { tunnel in
@@ -142,7 +142,7 @@ public class IOSControllerImpl : NSObject {
             var peerConfigurations: [PeerConfiguration] = []
             peerConfigurations.append(peerConfiguration)
 
-            var interface = InterfaceConfiguration(privateKey: privateKey!)
+            var interface = InterfaceConfiguration(privateKey: self.privateKey!)
 
             if let ipv4Address = IPAddressRange(from: deviceIpv4Address!),
                let ipv6Address = IPAddressRange(from: deviceIpv6Address!) {
@@ -152,7 +152,7 @@ public class IOSControllerImpl : NSObject {
 
             let config = TunnelConfiguration(name: VPN_NAME, interface: interface, peers: peerConfigurations)
 
-            return self.configureTunnel(config: config, reason: reason, serverName: serverIpv4AddrIn + ":\(serverPort )", gleanDebugTag: gleanDebugTag, isSuperDooperFeatureActive: isSuperDooperFeatureActive, installationId: installationId, disconnectOnErrorCallback: disconnectOnErrorCallback, onboardingCompletedCallback: onboardingCompletedCallback, vpnConfigPermissionResponseCallback: vpnConfigPermissionResponseCallback)
+            return self.configureTunnel(config: config, reason: reason, serverName: serverIpv4AddrIn + ":\(serverPort )", excludeLocalNetworks: excludeLocalNetworks, gleanDebugTag: gleanDebugTag, isSuperDooperFeatureActive: isSuperDooperFeatureActive, installationId: installationId, disconnectOnErrorCallback: disconnectOnErrorCallback, onboardingCompletedCallback: onboardingCompletedCallback, vpnConfigPermissionResponseCallback: vpnConfigPermissionResponseCallback)
         }
     }
 
--- a/src/platforms/ios/ioscontroller.swift
+++ b/src/platforms/ios/ioscontroller.swift
@@ -156,7 +156,7 @@ public class IOSControllerImpl : NSObject {
         }
     }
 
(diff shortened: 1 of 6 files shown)


MattLichtenstein:
> CheckBox enabled setting is not doing anything. @MattLichtenstein anything that immediatelly pops out? TBH, I did not deeply investigate it given the state of this PR, but if already know what it is, good.

The issue is you have the `enabled` property bound to an enum instead of an expression ([here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9070/files#diff-acab1d9449771c3371d8f4e7f6fc0309b9a9931e1077913d647a85dae31b7f01R87))

actual: `enabled: VPNController.StateOff`

expected: `enabled: VPNController.state === VPNController.StateOff`

You can also get rid of the `enabled` binding that you added to the MZCheckBox [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9070/files#diff-513b45286f8b5b56dc463b7f7bc4d166e3ab5c0cf9bd92e2006f27a0c71c3c55R33) since disabling the MZCheckBoxRow disables all of it's children

brizental:
> The issue is you have the enabled property bound to an enum instead of an expression ([here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9070/files#diff-acab1d9449771c3371d8f4e7f6fc0309b9a9931e1077913d647a85dae31b7f01R87))

:woman_facepalming:

brizental:
There are two commits I sent and are not showing up here yet @lesleyjanenorton , maybe wait a bit and if there are not two commits after https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9070/commits/5b3fa0ec5ce29ad0e3109475a9b01bce21964403 then this is not up to date and I'll see what is up tomorrow 🤦‍♀️

src/settingslist.h
```diff
@@ -709,6 +709,18 @@ SETTING_STRINGLIST(subscriptionTransactions,        // getter
                    false,                           // remove when reset
                    true                             // sensitive (do not log)
 )
+
+SETTING_BOOL(localNetworkAccess,        // getter
+             setLocalNetworkAccess,     // setter
+             removeLocalNetworkAccess,  // remover
+             hasLocalNetworkAccess,     // has
+             "localNetworkAccess",      // key
+             true,                      // default value
```
brizental, line 718: TODO: Change this to `false`.

src/translations/strings.yaml
```diff
@@ -360,6 +360,10 @@ settings:
     value: Clear all
     comment: Button label to clear all apps from the exclusions list
 
+localNetworkAccess:
+  label: "Local network access"
```
flodolo, line 364: Why the quotes?

src/translations/strings.yaml
```diff
@@ -360,6 +360,10 @@ settings:
...
+  subLabel: "Access printers, streaming sticks and all other devices on your local network"
```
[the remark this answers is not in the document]
flodolo: No, same IDs is fine.

src/translations/strings.yaml
brizental: Copy/Paste :spaghetti: Will remove.

src/ui/screens/settings/ViewPreferences.qml
```diff
@@ -67,6 +68,27 @@ MZViewBase {
...
+                anchors {
+                    right: parent.right
+                    left: parent.left
+                    rightMargin: VPNTheme.theme.windowMargin
+                }
+
+                labelText: MZI18n.LocalNetworkAccessLabel
+                subLabelText: MZI18n.LocalNetworkAccessSubLabel
+                isChecked: MZSettings.localNetworkAccess
+                enabled: VPNController.state === VPNController.StateOff
```
mcleinman, line 84: Should part of the text explain that this is only enabled when the VPN is off? If not, users will show up here with the VPN active and have no indication as to why the setting is disabled.

mcleinman (changes requested):
This looks great overall, just a couple questions.

src/ui/screens/settings/ViewPreferences.qml
brizental: That is a good point. There are other features that depend on the VPN being turned off. I'll check how they handle it.

brizental: Interesting, we actually don't have this behaviour anywhere in the app. I am pretty sure we used to. I'll ask today during stand up.

src/ui/screens/settings/ViewPreferences.qml
```diff
@@ -17,22 +18,25 @@ MZViewBase {
     property string _languageTitle: ""
     readonly property string telemetryScreenId : "app_preferences"
 
+    property bool vpnIsOff: VPNController.state === VPNController.StateOff
+    property bool isIOS: Qt.platform.os === "ios"
```
mcleinman, line 22: Nit: Move these two to `MZUiUtils.js`?

src/platforms/ios/ioscontroller.swift
```diff
@@ -116,7 +116,7 @@ public class IOSControllerImpl : NSObject {
         }
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
+    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  excludeLocalNetworks: Bool, allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
```
mcleinman, line 119:
```suggestion
    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, excludeLocalNetworks: Bool, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
```

I believe this fixes the crash

mcleinman (changes requested):
This is crashing every time I try to activate the VPN. I believe this is because the function's arguments are not in the expected order.

brizental:
> This is crashing every time I try to activate the VPN. I believe this is because the function's arguments are not in the expected order.

That is what I get for trusting month old code 🤦‍♀️ Thank you so much for testing this.

mcleinman (approved):
Looks good, thanks.

ValentinaPC:
@brizental Does this have a Jira ticket? Thanks!

PR #9090, opened 2024-02-06
Introduce mz_add_new_module CMake utility

brizental:
This is extracted from my YOLO sprint PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961

The modularization itself is a lot of code, but maybe we can sneakly introduce the utility to create modules and if we happen to create a new module (please raise your hand now @strseb 👀 ) we can create it in this format?

Usage examples can be found all over that PR:

- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-2660e919af9eddfb4d2b5933f6a0166edde95992e5813193b6f7eec51978a4da
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-76a02f3395033bf965cc8168ee9e36408f914cd1b6a4f57960998abc979e087d
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-b4e86c3c2ea7c1aa8ff606965604a64b8d68866dea660318e35e2543f37372b7
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-383906385b56894602bbb61b1d5968470b4eb5324959949af41c74449342f027
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-55ffab2043edfdc5b4edc06e31a6016365d051aa221a5d08004a793114589cc9
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-4b4dca15b7841d4d947d55cecf910e76f6c3dba6e2e01548a13ef2686f74839e
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-e27c8e325de5fa1be8ed0bd037301e77ba8c3bb756768837972f1fb79f1a2570
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-daa941ca76de1dba7afdcf49665d0cc636faacb747eb37c114add8282e3c5e8b
- https://github.com/mozilla-mobile/mozilla-vpn-client/pull/8961/files#diff-aca153d79dfeafc1bdc6c942f55b024cd97b8b2ee211fcdc833d44195fe64c88

---

One strong principle I used when desinging the module was that they were self contained. Meaning the module defines and contains all of it's dependencies, source files, test files, include directories, compile definitions and so on.

Targets that depend of the module just link to it and don't need to do anything else.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,9 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.20)
 
-option(BUILD_TESTS "Whether or not to build test targets" ON)
+# The source file GENERATED property is visible from all directory scopes when set.
+cmake_policy(SET CMP0118 NEW)
 
 message("Configuring for ${CMAKE_GENERATOR}")
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
@@ -25,7 +26,7 @@ project("Mozilla VPN" VERSION ${APP_VERSION} LANGUAGES C CXX
 include(CTest)
 add_custom_target(build_tests)
 set_target_properties(build_tests PROPERTIES
-        EXCLUDE_FROM_ALL ${BUILD_TESTS}
+        EXCLUDE_FROM_ALL TRUE
         FOLDER "Tests"
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,10 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-cmake_minimum_required(VERSION 3.20)
-
-# The source file GENERATED property is visible from all directory scopes when set.
-cmake_policy(SET CMP0118 NEW)
+cmake_minimum_required(VERSION 3.18)
 
 message("Configuring for ${CMAKE_GENERATOR}")
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@
 
 cmake_minimum_required(VERSION 3.18)
 
+option(BUILD_TESTS "Whether or not to build test targets" ON)
+
 message("Configuring for ${CMAKE_GENERATOR}")
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 if(NOT (IS_MULTI_CONFIG OR DEFINED CMAKE_BUILD_TYPE))
@@ -23,7 +25,7 @@ project("Mozilla VPN" VERSION ${APP_VERSION} LANGUAGES C CXX
 include(CTest)
 add_custom_target(build_tests)
 set_target_properties(build_tests PROPERTIES
-        EXCLUDE_FROM_ALL TRUE
+        EXCLUDE_FROM_ALL ${BUILD_TESTS}
         FOLDER "Tests"
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ cmake_minimum_required(VERSION 3.20)
 # The source file GENERATED property is visible from all directory scopes when set.
 cmake_policy(SET CMP0118 NEW)
 
+option(BUILD_TESTS "Whether or not to build test targets" ON)
+
 message("Configuring for ${CMAKE_GENERATOR}")
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 if(NOT (IS_MULTI_CONFIG OR DEFINED CMAKE_BUILD_TYPE))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,9 @@
 
 cmake_minimum_required(VERSION 3.20)
 
+# The source file GENERATED property is visible from all directory scopes when set.
+cmake_policy(SET CMP0118 NEW)
+
 message("Configuring for ${CMAKE_GENERATOR}")
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 if(NOT (IS_MULTI_CONFIG OR DEFINED CMAKE_BUILD_TYPE))
--- a/scripts/android/cmake.sh
+++ b/scripts/android/cmake.sh
@@ -138,6 +138,7 @@ if [[ "$RELEASE" ]]; then
     -DANDROID_SDK_ROOT=$ANDROID_SDK_ROOT \
     -DCMAKE_BUILD_TYPE=Release \
     -DADJUST_TOKEN=$ADJUST_SDK_TOKEN \
+    -DBUILD_TESTS=OFF \
     -GNinja \
     -S . -B .tmp/
 else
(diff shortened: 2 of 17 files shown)


scripts/cmake/utilities.cmake
```diff
@@ -19,3 +22,438 @@ function(mz_target_handle_warnings MZ_TARGET)
...
+# Example:
+# MZ_ADD_NEW_MODULE(
+#     TARGET_NAME MyModule
+#     INCLUDE_DIRECTORIES ${CMAKE_CURRENT_SOURCE_DIR}/include
+#     GENERATED_SOURCES ${CMAKE_CURRENT_BINARY_DIR}/generated_sources.cpp
+#     SOURCES src/file1.cpp src/file2.cpp
+#     TEST_SOURCES tests/test_file.cpp
+#     MZ_DEPENDENCIES mz_module
+# )
+function(mz_add_new_module)
```
brizental, line 95: This is the main function, I suggest reviewing from here and then looking at the other as necessary.

scripts/cmake/utilities.cmake
```diff
@@ -19,3 +22,438 @@ function(mz_target_handle_warnings MZ_TARGET)
...
+# - LINUX_DEPENDENCIES: (Optional) List of Linux-specific dependencies for the module.
+# - WINDOWS_DEPENDENCIES: (Optional) List of Windows-specific dependencies for the module.
+# - WASM_DEPENDENCIES: (Optional) List of WebAssembly-specific dependencies for the module.
+# - DUMMY_DEPENDENCIES: (Optional) List of dummy dependencies for the module.
+#
+# Example:
+# MZ_ADD_NEW_MODULE(
+#     TARGET_NAME MyModule
+#     INCLUDE_DIRECTORIES ${CMAKE_CURRENT_SOURCE_DIR}/include
+#     GENERATED_SOURCES ${CMAKE_CURRENT_BINARY_DIR}/generated_sources.cpp
```
brizental, line 90: Oh, this is wrong. And I need to send a new commit...

brizental: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9090/commits/ac4f63f29b656f75f45247cc41dad977e4cc8093

scripts/cmake/utilities.cmake
```diff
@@ -19,3 +22,437 @@ function(mz_target_handle_warnings MZ_TARGET)
...
+    elseif(${MZ_PLATFORM_NAME} STREQUAL "dummy")
+        list(APPEND LOCAL_ALL_SOURCES ${MZ_GENERATE_SOURCES_LIST_DUMMY_SOURCES})
+    else()
+        message(FATAL_ERROR "MZ_PLATFORM_NAME must be set before creating modules.")
+    endif()
+
+    # 2. Separate out Swift and ObjC sources
+
+    set(SWIFT_SOURCES ${LOCAL_ALL_SOURCES})
+    list(FILTER SWIFT_SOURCES INCLUDE REGEX "(.*)\\s\\w\\i\\f\\t$")
```
oskirby, line 396: I am not sure about this regex. If the goal is to separate `.swift` and `.mm` files, shouldn't this be:
`list(FILTER SWIFT_SOURCES INCLUDE REGEX "\\.swift$")`

The dot is a special character, so `(.*)` matches anything (including stuff without a literal dot). As is, I think this would match a string like `example.foobarswift`

scripts/cmake/utilities.cmake
```diff
@@ -19,3 +22,437 @@ function(mz_target_handle_warnings MZ_TARGET)
...
+
+    list(APPEND LOCAL_LINK_LIBRARIES ${MZ_GENERATE_LINK_LIBRARIES_EXTRA_DEPENDENCIES})
+
+    # 6. Test dependency handling
+
+    set(LOCAL_TEST_LINK_LIBRARIES ${LOCAL_LINK_LIBRARIES})
+
+    # Replace dependencies for test dependencies that start with `replace-`
+    set(REPLACER_DEPENDENCIES ${MZ_GENERATE_LINK_LIBRARIES_TEST_DEPENDENCIES})
+    list(FILTER REPLACER_DEPENDENCIES INCLUDE REGEX "^\\r\\e\\p\\l\\a\\c\\e\\-")
```
oskirby, line 338: I don't think alphanumeric chars require escaping in CMake Regex, so this could be simplified to `list(FILTER REPLACER_DEPENDENCIES INCLUDE REGEX "^replace-")`

scripts/cmake/utilities.cmake
brizental: Yeah, honestly no idea why I am escaping each character here. Good call on the `.` as well :facepalm:

CMakeLists.txt
```diff
@@ -2,7 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.20)
```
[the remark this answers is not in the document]
brizental: The change is required so that we can enable CMP0118. That policy makes things much much easier when we want to set sources as GENERATED. Right now, that is not required, I mean... none of this is required, this function is not used anywhere :D

If we don't set the policy we would need to break a little bit the encapsulation of modules, because the GENERATED property would need to be set in multiple places. That said, let's cross that bridge when we get there. This was YOLO sprint work and it is highly possible there is some other way to handle GENERATED files thta allows us not to increase the CMake version.

scripts/cmake/utilities.cmake
```diff
@@ -19,3 +22,437 @@ function(mz_target_handle_warnings MZ_TARGET)
...
+        "" # single-value args
+        "TARGET_NAME;TEST_COMMAND;PARENT_TARGET;SOURCES;DEPENDENCIES" # multi-value args
+        ${ARGN})
+
+    # Test targets are executable targets.
+    qt_add_executable(${MZ_ADD_TEST_TARGET_NAME}
+        ${MZ_ADD_TEST_SOURCES}
+    )
+    set_target_properties(${MZ_ADD_TEST_TARGET_NAME} PROPERTIES
+        EXCLUDE_FROM_ALL TRUE
```
strseb, line 238: I'm not sure we should keep doing that :) 
This makes ctest wieerd, given that we will be registering the test in ctest, even if it does not exist and there is nothing that will build it. 

In "old" cmake land we have the build_tests target as a proxy to trigger all build things. 
We should keep that, so that in CI we can easy do `cmake build_tests && ctest /build` but also we should have  `-DBUILD_TESTS=OFF` like qt and others that change if tests are added to `ALL` - as that allows for a nice devex i.e test on save.

CMakeLists.txt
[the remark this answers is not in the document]
oskirby: I'll take a stab at seeing what it takes to update CMake on the PPA. It might not be as hard as I think it is.

scripts/cmake/utilities.cmake
oskirby: I agree, the use of `EXCLUDE_FROM_ALL` was done simply to try and keep the CI jobs from burning too much CPU time if they weren't going to run tests. But we should probably invert this so that test are built by default and the CI builds can just specify the targets they need (eg: `cmake --build <dir> --target mozillavpn`)

oskirby (approved):
I think this is looking good.

scripts/android/cmake.sh
```diff
@@ -156,7 +156,7 @@ fi
 
 print Y "Compiling apk_install_target in .tmp/"
 # This compiles the client and generates a mozillavpn.so
-cmake --build .tmp -j$JOBS
+cmake --build .tmp --target mozillavpn -j$JOBS
```
strseb, line 159: That's not the target you want - that just builds libmozillavpn.so, on android qt add's other default targets that will also generate the gradle project, hence why the builds are not working  
keep the default, just flag the tests off :)

strseb (approved):

strseb:
Before i forget, this needs probably a docs page :)

CMakeLists.txt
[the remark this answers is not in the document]
oskirby: The [ppa:okirby/qt6-backports](https://launchpad.net/~okirby/+archive/ubuntu/qt6-backports) should now have CMake 3.22 available, backported from Ubuntu 22.04/Jammy, so we can add this policy back to help with dependencies. Though, you will also need to rebase ontop of #9127 to make sure the taskcluster worker is updated.

PR #9096, opened 2024-02-07
VPN-5187: Expose "Factory reset" functionality

MattLichtenstein:
## Description

- Exposes the "Reset and quit" functionality that was previously only available in the developer menu, to users via the "Get help" settings
- Resetting the VPN currently resets the following settings: 
  - All persistent settings stored in vpn.moz (list can be found in [settingslist.h](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/settingslist.h))
  - iOS only: VPN tunnel system config
  - Authentication details. The user is signed out

## Reference

[VPN-5187: Implement new 'factory reset' UX in the GUI](https://mozilla-hub.atlassian.net/browse/VPN-5187)


src/ui/screens/settings/ViewReset.qml
```diff
@@ -0,0 +1,189 @@
...
+            text: MZI18n.ResetSettingsBody1
+            color: MZTheme.theme.fontColor
+            horizontalAlignment: Text.AlignLeft
+        }
+
+        MZInterLabel {
+            Layout.topMargin: 16
+            Layout.fillWidth: true
+
+            text: "<ul style='margin-left: -20px;'><li>%1</li></ul>".arg(MZI18n.HelpSheetsPrivacyTitle)
```
mcleinman, line 56: Why are we styling text inline?

mcleinman (commented):
Just one thing, for me to understand better. This is looking great.

src/ui/screens/settings/ViewReset.qml
MattLichtenstein: It's markdown to get a bulleted list

mcleinman: Do we do similar things other places in the app? I don't see many. Is there any cleaner way to do a bulleted list? (Genuine question, I'm not familiar w/ Qt best practices.)

mcleinman:
This factory reset option appears on the "Get help" menu on the initial log in screens. Is this intended? Presumably you're already in a reset state at that point.

src/ui/screens/settings/ViewReset.qml
MattLichtenstein: I believe it's a standard practice. Alternatively we could build something ourselves like a rounded rectangle + text side by side, but I'm not sure that is any more beneficial. We use the same technique in a few places like guides and in-app messages, but, afaik we don't have bulleted list anywhere else

MattLichtenstein:
> This factory reset option appears on the "Get help" menu on the initial log in screens. Is this intended? Presumably you're already in a reset state at that point.

Good question! Spoke with Santiago, and he is fine with the option being accessible from any "Get help" access point. Makes sense because a previously logged in user would still have all of their settings intact (recent connections, privacy features, etc) and could want to dispose of them before logging back in

mcleinman (approved):
Thanks!

src/ui/screens/settings/ViewReset.qml
lesleyjanenorton: Curious: Did you happen to try this with a screen reader? I'm wondering how Qt exposes the text value and if the markup could cause any grief there.

PR #9104, opened 2024-02-07
VPN-6104: Turn on help sheets feature flag

MattLichtenstein:
## Description

- Enables helps sheets

## Reference

[VPN-6104: Turn the in-product help sheets feature on](https://mozilla-hub.atlassian.net/browse/VPN-6104)

### DIFF
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -544,15 +544,14 @@ static QList<InspectorCommand> s_commands{
                        return obj;
                      }},
 
-    InspectorCommand{"is_feature_enabled",
-                     "Check if a feature is enabled", 1,
+    InspectorCommand{"is_feature_enabled", "Check if a feature is enabled", 1,
                      [](InspectorHandler*, const QList<QByteArray>& arguments) {
                        QJsonObject obj;
                        QString featureName = arguments[1];
                        auto featureModel = FeatureModel::instance();
                        QObject* featureObj = featureModel->get(featureName);
 
-                       if(!featureObj) {
+                       if (!featureObj) {
                          obj["error"] = "Feature does not exist";
                          return obj;
                        }
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -544,6 +544,26 @@ static QList<InspectorCommand> s_commands{
                        return obj;
                      }},
 
+    InspectorCommand{"is_feature_enabled",
+                     "Check if a feature is enabled", 1,
+                     [](InspectorHandler*, const QList<QByteArray>& arguments) {
+                       QJsonObject obj;
+                       QString featureName = arguments[1];
+                       auto featureModel = FeatureModel::instance();
+                       QObject* featureObj = featureModel->get(featureName);
+
+                       if(!featureObj) {
+                         obj["error"] = "Feature does not exist";
+                         return obj;
+                       }
+
+                       Feature* feature = static_cast<Feature*>(featureObj);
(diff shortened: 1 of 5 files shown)


brizental (approved):

tests/functional/testMultihop.js
```diff
@@ -63,12 +68,6 @@ describe('Server list', function() {
             queries.screenHome.serverListView.VPN_COLLAPSIBLE_CARD,
             'expanded') === 'false');
 
-    await vpn.waitForQuery(
-        queries.screenHome.serverListView.ENTRY_BUTTON.visible());
-
-    await vpn.waitForQuery(
-        queries.screenHome.serverListView.EXIT_BUTTON.visible());
-
```
MattLichtenstein, line 71: these checks didn't seem to serve any purpose, so I removed them and renamed this test

tests/functional/testSettings.js
```diff
@@ -63,6 +67,11 @@ describe('Settings', function() {
   });
 
   it('Checking the tips and tricks settings', async () => {
+    //tips and tricks feature is hidden when help sheets are enabled, so skip this test when help sheets are enabled
+    if (await vpn.isFeatureEnabled('helpSheets')) {
```
jswinarton, line 71: I think it would be better to use `this.skip()` rather than returning directly, so the test doesn't report as passing. https://mochajs.org/#inclusive-tests (ctrl+f for "skip at runtime")

src/inspector/inspectorhandler.cpp
```diff
@@ -544,6 +544,25 @@ static QList<InspectorCommand> s_commands{
                        return obj;
                      }},
 
+    InspectorCommand{"is_feature_enabled", "Check if a feature is enabled", 1,
```
jswinarton, line 547: What is the difference between enabling a feature and flipping a feature on? A comment might be helpful here (or maybe there is some documentation that I missed).

jswinarton (approved):
r+wc

src/inspector/inspectorhandler.cpp
```diff
@@ -544,6 +544,25 @@ static QList<InspectorCommand> s_commands{
...
+                       QString featureName = arguments[1];
+                       auto featureModel = FeatureModel::instance();
+                       QObject* featureObj = featureModel->get(featureName);
+
+                       if (!featureObj) {
+                         obj["error"] = "Feature does not exist";
+                         return obj;
+                       }
+
+                       Feature* feature = static_cast<Feature*>(featureObj);
```
vinocher, line 559: Couldn't this be replaced with:
`const Feature* feature = Feature::getOrNull(featureName);
                       if (!feature) {`

tests/functional/testSettings.js
```diff
@@ -1045,6 +1054,11 @@ describe('Settings', function() {
             // This test cannot run in wasm
             return;
         }
+
+        if (await vpn.isFeatureEnabled('helpSheets')) {
```
vinocher, line 1058: Perhpas add a comment about why this is being skipped?

vinocher (approved):

src/inspector/inspectorhandler.cpp
MattLichtenstein: Learned this the hard way. We use the term "flipping" and "flipped" to describe features that were toggled on/off manually (either through the UI, or the inspector), otherwise, the list of flipped on/off features is empty. 

The command I have added is simply to see if the feature is on or off,.

Will add a comment as well!

tests/functional/testSettings.js
MattLichtenstein: did not know this existed! thanks!

PR #9113, opened 2024-02-08
VPN-6152 - Allow WebExt to talk to VPN

strseb:
## Description

To let the extension allow talking to the client, we need to make sure the right id is there c:
Looking at:
https://github.com/mozilla-extensions/mozilla-vpn-extension/blob/main/src/manifest.json#L8

we chose mozilla.com instead of .org


lesleyjanenorton (approved):
👍

PR #9116, opened 2024-02-09
Fix: Taskcluster/MacOS ensure a fresh install of miniconda on each run

oskirby:
## Description
It seems that the miniconda installation directory gets cached between task runs on our MacOS workers, and we run the installer with the `-u` option (for upgrade), which means that we never really install Conda and are just kind of relying on the cached version to be correct.

Unfortunately, when there are multiple versions of Conda being used by our taskcluster jobs, such as when upgrading packages, this can result in broken packages. To ensure that Conda is correctly installed, we should flush these caches when the job starts and do a fresh install.

## Checklist
    

mcleinman (approved):
Thank you!

PR #9126, opened 2024-02-09
VPN-5521: Enable new onboarding

MattLichtenstein:
## Description

- Enables the New Onboarding feature by default
- Amend functional tests
  - Ensure functional tests pass with both the old and new onboarding
  - Skip old/new onboarding when not specifically being tested

## Reference

[VPN-5521: Turn on the New Onboarding feature flag](https://mozilla-hub.atlassian.net/browse/VPN-5521)

### DIFF
--- a/tests/functional/helper.js
+++ b/tests/functional/helper.js
@@ -415,8 +415,13 @@ module.exports = {
     }
   },
 
-  async authenticateInApp(
-      clickOnPostAuthenticate = false, acceptTelemetry = false) {
+  async authenticateInApp(skipOnboarding = true) {
+    if (skipOnboarding) {
+      await this.setSetting('onboardingCompleted', 'true')
+      await this.setSetting('postAuthenticationShown', 'true')
+      await this.setSetting('telemetryPolicyShown', 'true')
+    }
+    
     // This method must be called when the client is on the "Get Started" view.
     await this.waitForInitialView();
 
@@ -445,13 +450,12 @@ module.exports = {
     await this.waitForMozillaProperty(
         'Mozilla.VPN', 'VPN', 'userState', 'UserAuthenticated');
 
-    if (clickOnPostAuthenticate) {
+    if (!skipOnboarding && !await this.isFeatureEnabled('newOnboarding')) {
       await this.waitForQuery(queries.screenPostAuthentication.BUTTON.visible());
       await this.clickOnQuery(
           queries.screenPostAuthentication.BUTTON.visible());
       await this.wait();
-    }
-    if (acceptTelemetry) {
+
       await this.waitForQuery(queries.screenTelemetry.BUTTON.visible());
       await this.clickOnQuery(queries.screenTelemetry.BUTTON.visible());
       await this.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
--- a/tests/functional/testCaptivePortal.js
+++ b/tests/functional/testCaptivePortal.js
@@ -69,8 +69,13 @@ describe('Captive portal', function() {
     assert.equal(vpn.lastNotification().title, null);
   });
 
-  it('Captive portal in the Telemetry policy view', async () => {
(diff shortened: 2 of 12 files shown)


lesleyjanenorton (approved):
👍

Might be nice to have helper fn for the 3-line post-auth settings block since it doesn't appear to change and we're using it in a few files.

MattLichtenstein:
> 👍
> 
> Might be nice to have helper fn for the 3-line post-auth settings block since it doesn't appear to change and we're using it in a few files.

@lesleyjanenorton Would you mind elaborating? I assume you are not referring to the function to complete the post auth screen [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9126/files#diff-ecac75875a7bfd126997889026cef83ea780e46d4c1f0e985db256eabe718c14R442-R447)

tests/functional/helper.js
```diff
@@ -345,7 +345,13 @@ module.exports = {
 
   // TODO - The expected staging urls are hardcoded, we may want to
   // move these hardcoded urls out if testing in alternate environments.
-  async authenticateInBrowser(clickOnPostAuthenticate, acceptTelemetry, wasm) {
+  async authenticateInBrowser(wasm, skipOnboarding = true) {
+    if (skipOnboarding) {
+      await this.setSetting('onboardingCompleted', 'true')
+      await this.setSetting('postAuthenticationShown', 'true')
+      await this.setSetting('telemetryPolicyShown', 'true')
+    }
```
brizental, line 353: :100:

tests/functional/testAuthenticationInApp.js
```diff
@@ -301,6 +301,11 @@ describe('User authentication', function() {
         await vpn.flipFeatureOn('inAppAccountCreate');
       }
 
+      //skip onboarding. normally done in helper::authenticateInApp(), but this test logs in manually
+      await vpn.setSetting('onboardingCompleted', 'true')
+      await vpn.setSetting('postAuthenticationShown', 'true')
+      await vpn.setSetting('telemetryPolicyShown', 'true')
```
brizental, line 307: nit: Create a "skipOnboarding" helper method?

tests/functional/testCaptivePortal.js
```diff
@@ -58,8 +58,13 @@ describe('Captive portal', function() {
     assert.equal(vpn.lastNotification().title, null);
   });
 
-  it('Captive portal in the Post authentication view', async () => {
-    await vpn.authenticateInApp();
+  it('Captive portal in the Post authentication view', async function () {
+    //Post auth view does not exist in new onboarding
+    if (await vpn.isFeatureEnabled("newOnboarding")) {
+      this.skip();
```
brizental, line 64: I don't get it. Shouldn't we remove these tests?

MattLichtenstein: I felt a bit reluctant to remove anything related to old onboarding (post auth and telemetry screens) until the new onboarding is released and is confirmed working well enough that we don't have to revert it.

I plan on removing everything related to the old onboarding, including tests, when we delete the feature flag

MattLichtenstein: The real question to me was should we be running tests on the old onboarding, even when the new onboarding is disabled. I concluded that it just doesn't make sense to run functional tests on code that isn't active (without going into dev settings), it would just be too confusing with all of the permutations, so thats the reason I am skipping most post auth/telemetry screen tests until they can be removed

brizental: If the code is still in there we should keep on testing it even though it is not enabled. In this case I suggest flipping the feature flag on for all tests. Think about it this way: we don't it to break in case we need to re-enable it. When the whole thing is removed we remove the tests as well.

mcleinman (approved):
👍🏻

mcleinman:
@brizental do you have any changes/thoughts, or are we ready to merge this?

brizental (approved):

PR #9129, opened 2024-02-10
VPN-6041 iOS: Swift logging doesn't save to VPN log stream

Gela:
…troller

## Description

We are dropping all swift client side logs messages that are not in the network extension. This is problematic because we use the iOS logs to determine issues for the clients, and not seeing vital information such as VPN connection state for example could make it harder/impossible to effectively draw conclusions from the iOS logs.

Because in iOSLogger we were filtering the messages by only logging what's coming from `.appex` previously, we were missing this data. Simply removing the if statement `if (Bundle.main.bundlePath.hasSuffix(".appex")` doesn't resolve this because now we end up with some duplicate logs coming from the main app. 

The simplest fix I could think of which doesn't involves rewriting the logging code is to append the class name to the log messages coming from the swift controller, and expanding the if statement check to allow messages coming from `IOSSwiftController`. I would argue that this is a good practice in general because otherwise with the log statement we can't even tell where they're coming from. For example just seeing "Connecting" in logs gives no context as to where it's coming from. 

## Reference

    https://mozilla-hub.atlassian.net/browse/VPN-6041

## Checklist
    
### DIFF
--- a/src/platforms/ios/ioscontroller.swift
+++ b/src/platforms/ios/ioscontroller.swift
@@ -60,19 +60,19 @@ public class IOSControllerImpl : NSObject {
 
         switch session.status {
         case .connected:
-            IOSControllerImpl.logger.debug(message: "STATE CHANGED: connected")
+            IOSControllerImpl.logger.debug(message: "(IOSSwiftController) Debug: STATE CHANGED: connected")
         case .connecting:
-            IOSControllerImpl.logger.debug(message: "STATE CHANGED: connecting")
+            IOSControllerImpl.logger.debug(message: "(IOSSwiftController) Debug: STATE CHANGED: connecting")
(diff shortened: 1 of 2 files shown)


mcleinman (changes requested):
I prefer a more general solution for this - we're going to be adding more Swift files (hopefully), so I'd like something that could handle more than just one file.

I just investigated for a bit. I'm not quite sure how easy it would be to send all app swift logs up to the C++ logging layer. I'd be comfortable with adding a third logging section of front end swift logs. Not ideal, but better than the current state. We could use `getBackendLogs` to send both the backend and swift front end logs to the C++ layer when we need to share the logs. We'd have to use a different file for those logs, of course. (And delete them when `clearAppexLogs` is called.)

PR #9140, opened 2024-02-12
Nit: No more npm updates from dependabot

strseb:
## Description
None of our node code is shipped to users. 
All of that is internal dev tools, the testing suite or our docs page. 

Therefore imho dependabot is more noise then useful. 

WDYT?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -68,10 +68,10 @@ updates:
     labels:
       - "dependencies"
   # Functional Tests etc
-  - package-ecosystem: "npm"
-    directory: "/"
-    schedule:
-      interval: "weekly"
+  #- package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


strseb:
> Maybe we keep it enable for security updates only on tools and docs stuff?

For some reason security dependencies are it's own dependabot feature and not configurable with this file. 
We can disable the normal npm check and still have dependabot file whenever there is an update that addresses a secbug :) 
We currently do not have that active, but i would do so.

brizental:
> We can disable the normal npm check and still have dependabot file a pr whenever there is an update that addresses a secbug :)
> We currently do not have that active, but i would do so.

SGTM.

brizental (approved):
Approving under the assumption we will enable the sec updates only.

PR #9057, opened 2024-01-31
VPN-5860: Improve health metrics

mcleinman:
## Description

The first round of health metrics from Super Dooper dashboards needed improvement:
- We were spamming Glean with events when connections were down, which was too noisy for useful engineering and data insights.
- It was still hard to answer questions like "how long are connections stable for?"

This PR attempts to address this. The current solution was designed in collaboration w/ Yeonjoo (data) and Santiago (product). We're adding 9 new metrics, and removing a bunch as well. We think that some subset of these 9 will be most useful, but aren't quite sure. Thus, these metrics all will live for about 6 months, and we'll then evaluate the data and decide which ones to keep in perpetuity.

These metrics are reworked both in the app and in the Android daemon.

The metrics (3 of each, for stable, unstable, and no connection):
- An event for when the connectivity *changes* to this state.
- A counter for when a health check returns this state, even if it was already in the state.
- A timing distribution for how long it stays in a given state. (This only exists on Android daemon and desktop, as the mobile apps are frequently relaunched during a VPN session.)

The counters can be compared in relative terms, but should not be considered as markers of time. There is a race condition in the counter, and it cannot be relied upon as having one count per second. In many cases, we conduct a health check due to the timer timeout. Then milliseconds later, a ping returns and we do a health check due to that ping returning. However, if the ping returned before the timeout, the timer would have been cancelled we would have only recorded one. Fixing this is outside the scope of this PR, and the resulting data is still valuable.

I also renamed a couple Android booleans, to make it clear when the silent server switch (or "reconnect") is coming from the app, and when it's coming from the daemon.

I've tested this on macOS, Android, and iOS.

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-5860?focusedCommentId=825842

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #9146, opened 2024-02-13
VPN-4979: (part 1) iOS network extension health checks

mcleinman:
## Description

**Important: This PR adds a new git submodule, [SwiftyPing](github.com/samiyr/SwiftyPing/). It didn't seem to make sense to roll our own ping sender, and this code seems clean to me (and uses the MIT license).**

I'm breaking down VPN-4979 into 3 PRs, to be stacked on one another:
1. This one, which adds health monitoring
2. One that adds silent server switching in. It's going to refactor a little bit of how it's done for Android so the code can be shared, and if we're trying to keep PRs small it should be a separate one.
3. One that adds the telemetry in. It should be done after https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9057 is merged, as it will be based on those new metrics.

The app and the Android daemon do health checks differently:
- Desktop: Sending pings out basically constantly. Health check every second based on the last 32 pings.
- Android daemon: Sending out a check every 30 seconds. Sees what it gets back from those pings, and makes a health decision based on which pings are reached.

Since we want to be respectful of device battery life, we need to carefully balance "not sending out too many checks" with "don't leave the device in a messed up state". This PR sends out 20 pings every 15 seconds, and calculates health status based on that.

We don't currently have iOS testing infrastructure, so not adding tests for this.

## Reference

VPN-4979

## Checklist
    

mcleinman:
This needs review, as we're now continuing this work.

.gitmodules
```diff
@@ -27,3 +27,6 @@
 [submodule "3rdparty/sentry"]
 	path = 3rdparty/sentry
 	url = https://github.com/getsentry/sentry-native/
+[submodule "3rdparty/swiftyping"]
```
strseb, line 30: As we ship this with the client please update Licence.md :)

src/platforms/ios/iostunnel.swift
```diff
@@ -100,6 +102,11 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
                     GleanMetrics.Pings.shared.daemonsession.submit(reason: .daemonStart)
                 }
 
+                // The internal DNS endpoint results is in the 10.x.x.x restricted IP range. For reasons I couldn't
+                // determine, pinging restricted IP addresses failed. Thus, using other ones.
+                let endpointIP = String(describing: tunnelConfiguration.peers.first?.endpoint?.host ?? "8.8.8.8")
```
strseb, line 107: Why start connectionhealth if `tunnelConfiguration.peers.first?.endpoint?.host` is empty? 🤔 

I'm tending to either assert it's there, cause otherwise we should hopefully (🤞 ) have an adapterError, or we should not start the timer. Wdyt?

src/platforms/ios/PingTest.swift
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import NetworkExtension
+
+class PingTest {
```
strseb, line 8: 🤩 NICE, even if we do not have automation, could you add a small doc on how to run those type of things on the machine? (if it is runnable ofc :D )

strseb (approved):
r+wc, thanks !

strseb:
Also small comment re build failures, might need to rebase as this should fix the conda wierdness https://github.com/mozilla-mobile/mozilla-vpn-client/commit/f4c660aa784b995ddd7fd3f6ffc2133408a7da71

.gitmodules
mcleinman: Good call, added.

src/platforms/ios/iostunnel.swift
mcleinman: Changed this to a guard with an assertion, and a note about why we don't change behavior on prod. Let me know if you think we should take a different approach.

src/platforms/ios/PingTest.swift
mcleinman: This was poorly named. I renamed it to `PingAnalyzer`. It isn't a test, it's the helper function that sends test pings out.

PR #9192, opened 2024-02-29
VPN-6041: Add Swift logs from app

mcleinman:
## Description

We've had a problem for a while where only the Swift logs from the network extension are included in the log file output by the app. 

Ideally, the swift logs would be inline with the rest of the app logs (from C++ land). I looked into adding a objc bridging header and writing directly to the main logger. Unfortunately, the main logger class uses a lot of Qt imports, and the build fails due to these deeper dependencies. After I did that research, I went with this approach - a somewhat hacky fix. (Given that untangling this would be non-trivial and the current status of the app, I went with a simpler solution.)

I've added a third section of logs to the log output - Swift logs from the main app will exist, but be in their own section. This is done by hackily appending the Swift log section to the backend logs, so it looks no different to the main app. Where possible, I maintained the same code patterns that existed for the backend (network extension) logs.

I had to update our calls from C++ to iOS logging - you'll notice updates to functions with names like `errorToConsoleWithMessage`. Without this, the C++ app logs were logged in two places - the main app logs (as expected) and these new Swift logs. This is because we were sending all C++ logs to Swift-land - but seemingly only to log them to the console. The easiest way in the Swift logging class to separate out "C++ logs to print to console only" and "Swift logs to print to console and to save to disk" was to give those C++ logs their own functions, which is done as part of this PR.

[Sample logs from a fresh run of the app](https://github.com/mozilla-mobile/mozilla-vpn-client/files/14454079/mozillavpn.3.log)

## Reference

VPN-6041

## Checklist
    

src/platforms/ios/ioslogger.swift
```diff
@@ -35,46 +37,76 @@ public class IOSLoggerImpl : NSObject {
         )
     }
 
-    @objc func debug(message: String) {
+    @objc func debugToConsole(message: String) {
+        os_log("%{public}@", log: self.log, type: .debug, message)
```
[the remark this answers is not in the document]
mcleinman: It is for me - I'm seeing both C++ and Swift logs in the Xcode console.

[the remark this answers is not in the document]
brizental: I mean the Console app. Either way, I guess that would be a separate issue.

src/platforms/ios/ioslogger.swift
```diff
@@ -102,6 +134,31 @@ public class IOSLoggerImpl : NSObject {
             logger.error(message: "Unable to access log file at \(appexLogFileURL): \(error)")
         }
     }
+
+    private static func withAppSwiftLogFile(_ f: (_ handle: FileHandle) throws -> Void) {
```
brizental, line 138: nit: The only difference between this function and `withAppexLogFile` is the state of the empty file. I think we can merge these two functions into one that gets the data from an empty log file as a string.

src/platforms/ios/ioslogger.swift
```diff
@@ -35,46 +37,76 @@ public class IOSLoggerImpl : NSObject {
...
+    @objc func debugToConsole(message: String) {
+        os_log("%{public}@", log: self.log, type: .debug, message)
+    }
+
+    @objc func infoToConsole(message: String) {
+        os_log("%{public}@", log: self.log, type: .info, message)
+    }
+
+    @objc func errorToConsole(message: String) {
+        os_log("%{public}@", log: self.log, type: .error, message)
```
brizental, line 49: nit: To avoid code repetition, maybe we can extract the `os_log` call into a separate function or keep using the `log` fucntion and rip out the part that write to the log file into a separate function? Maybe we can just keep using `debug`, `info` and `error` but pass `writeToLogFile: bool` as an argument.

mcleinman: I removed some of the repetition by creating a more general function to go from obj-c to swift. I was having trouble finding a way to use the iOS enum in obj-c code, so I made my own after spending time researching/attempting. (Let me know if you have a code snippet that works for it.)

mcleinman:
@lesleyjanenorton per our discussion, I made a few small edits. I also looked at all the Swift logs - I couldn't find any that include superfluous names/data, and none that include IP addresses or anything like that.

brizental (approved):

PR #9216, opened 2024-03-06
VPN-5766: Fix iOS crash

mcleinman:
## Description

Big credit to @oskirby for coming up with the fix after the source of the problem was identified.

iOS is closing sockets when the device is closed (and probably while the app is backgrounded), causing this crash when the device is turned back on. This makes sense as to why QA is running into this the most - as they are more likely to launch the app and immediately background it.

To reproduce:
1. Ensure VPN is off
2. Cold launch of app
3. In about a second or two, press power button to turn screen off
4. In 5-10 seconds, unlock phone (this part seemed inconsistent - the crash doesn’t happen if it’s only a couple seconds)
5. Crash happens

This PR does 3 things:
- Improves logging (removes some logging we added to diagnose the problem, adds some better logging for any future issues)
- Adds the `MSG_NOSIGNAL` flag to the socket. (Big thanks to Naomi for this.)
- If server latency checks are running when the app state changes to non-foreground, it stops the checks and sets it up to refresh the server list on the next launch. (FWIW, we already had similar code for health checks - the other place that uses sockets.) 

(This third piece - My first attempt at this functionality was to essentially pause the running check. In `ServerLatency::maybeSendPings`, I checked if we were on iOS and not foregrounded, and if so did an early return while setting a timer to check again in a few seconds. For reasons I don't quite understand, when we came back to foreground the latency checks were *very very slow*, and so I went with this current approach. It's not the best - as it restarts the entire check - but given that in most cases a server check isn't interrupted by backgrounding the app, it seems like an okay tradeoff.)

## Reference

VPN-5766

## Checklist
    
### DIFF
--- a/src/platforms/macos/macospingsender.cpp
+++ b/src/platforms/macos/macospingsender.cpp
@@ -69,7 +69,6 @@ MacOSPingSender::~MacOSPingSender() {
 }
 
 void MacOSPingSender::sendPing(const QHostAddress& dest, quint16 sequence) {
-  logger.debug() << "Creating address";
   quint32 ipv4dest = dest.toIPv4Address();
   struct sockaddr_in addr;
   bzero(&addr, sizeof(addr));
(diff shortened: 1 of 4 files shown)


src/platforms/macos/macospingsender.cpp
```diff
@@ -69,24 +69,21 @@ MacOSPingSender::~MacOSPingSender() {
 }
 
 void MacOSPingSender::sendPing(const QHostAddress& dest, quint16 sequence) {
-  logger.debug() << "Creating address";
```
oskirby, line 72: Thank you for tidying this up.

oskirby (approved):
Looks good to me.

PR #9231, opened 2024-03-13
VPN-5101: cleanup logs after they are exported for iOS and desktop

mcleinman:
## Description

While we had the code to cleanup logs, we weren't using it. While this ticket was about iOS, I also added relevant code for all non-Android platforms.

Reason for not implementing on Android:
- On desktop platforms, we save the log file, so it can always be found.
- On iOS, we only clear logs if we get a positive signal that the user did something with the logs. (If they cancel out of the share sheet screen, the logs are not cleared.)
- On Android, we don't have a way to discover whether the user did something with the share sheet - and thus they could have canceled out without sharing the logs anywhere. In those cases, we don't want to clear the logs. And since we can't tell which type of case we're in, I'm being conservative and keeping the logs for Android.

## Reference

VPN-5101

## Checklist
    
### DIFF
--- a/src/loghandler.cpp
+++ b/src/loghandler.cpp
@@ -449,15 +449,10 @@ bool LogHandler::viewLogs() {
   return ok;
 #endif
 
-  if (writeAndShowLogs(QStandardPaths::DesktopLocation)) {
-    return true;
-  }
-
-  if (writeAndShowLogs(QStandardPaths::HomeLocation)) {
-    return true;
-  }
-
-  if (writeAndShowLogs(QStandardPaths::TempLocation)) {
+  if (writeAndShowLogs(QStandardPaths::DesktopLocation) ||
+      writeAndShowLogs(QStandardPaths::HomeLocation) ||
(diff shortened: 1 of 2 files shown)


brizental:
> Reason for not implementing on Android:

Are logs never cleared on Andriod then?

mcleinman:
> > Reason for not implementing on Android:
> 
> Are logs never cleared on Andriod then?

Correct. I believe currently logs aren't cleared on any platform. After this PR, they'll be cleared on all platforms except Android. Without an easy way to tell if the Android logs are shared (after we pass it to the share sheet), I think it's worth having improvements for the other platforms and having different behavior on Android. 

Alternatively, we can just change behavior on iOS - this ticket was about iOS, and in my research I realized we weren't clearing logs on any platform currently.

brizental (approved):
Got it. Let's file a separate ticket for Android in this case. In an ideal world we would add tests for this behaviour... but I'll leave that decision to you 💯

PR #9239, opened 2024-03-18
VPN-4769 Add Clang-Tidy 

strseb:
## Description
oEgssVo7shAD4AJyB2A4IhEVNN0DNBgE4zB6af~tplv-nhvfeczskr-1 250 0

I don't think i need to pitch why we should use clang-tidy. 
given it's a tidbit non intuitive to use and i would need to write a script anyway, why not just hide all that in a cmake target to build a report. 

So if you configure the project cmake will tell you if you are ready for clang-tidy, , otherwise it will tell you what to do. 
If all ready, we add a new target  `cmake --build <build_dir> --target clang_tidy_report`, builds a folder containing a .yaml file for each target. each file contains the notes, errors and suggestions. 
The task "fails" if checks that marked as errors throw (see .clang-tidy), allowing us to inform and enforce in CI. 

If suggestions are generated they can then be aplied with `clang-apply-replacements`

Also i added a docs page :) 

## Reference
https://mozilla-hub.atlassian.net/browse/VPN-4769

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,12 @@
+Checks: >  
+  clang-diagnostic-*
+  clang-analyzer-*,
+#  modernize-*,
+#  cppcoreguidelines-*,
+#  -modernize-use-trailing-return-type,
+WarningsAsErrors: >
+  clang-analyzer-security.*
+HeaderFilterRegex: ''
+AnalyzeTemporaryDtors: false
+# Use our Clang Format
+FormatStyle: file
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,9 +1,9 @@
 Checks: >  
   clang-diagnostic-*
   clang-analyzer-*,
-#  modernize-*,
+  -modernize-use-trailing-return-type,
+  modernize-*,
 #  cppcoreguidelines-*,
-#  -modernize-use-trailing-return-type,
 WarningsAsErrors: >
   clang-analyzer-security.*
 HeaderFilterRegex: ''
--- a/scripts/cmake/clang_tidy.cmake
+++ b/scripts/cmake/clang_tidy.cmake
@@ -95,10 +95,3 @@ function(mz_add_clang_tidy aTarget)
     add_dependencies(clang_tidy_all ${aTarget}_clang_tidy)
     add_dependencies(clang_tidy_fix_all ${aTarget}_clang_tidy_fix)
 endfunction() 
-
-add_custom_target(clang_tidy_diff
-   COMMAND git diff main | python3 ${PROJECT_SOURCE_DIR}/scripts/clang-tidy-diff.py -clang-tidy-binary ${CLANG_TIDY_EXECUTABLE} -p ${CMAKE_BINARY_DIR} 
-   DEPENDS ${aTarget}
-   COMMENT "Running clang-tidy against a diff ${branch} -> main"
-   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} 
-)
--- a/scripts/cmake/clang_tidy.cmake
+++ b/scripts/cmake/clang_tidy.cmake
@@ -1,0 +1,104 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+
+# This file adds a function 
+# mz_add_clang_tidy( <SOME_TARGET> )
+#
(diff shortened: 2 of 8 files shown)


scripts/run-clang-tidy.py
strseb, line 1: Given i call clang-tidy <HUGE BLOB OF SOURCES>, this python script will fork out multiple clang-tidy workers, otherwise it would be singlethreaded and slow :c

.clang-tidy
strseb, line 1: Obviously we need to talk about those rules :) But this is kind of the minimum thing we pass, so i'd imagine step by step we discuss what we want to enable end refactor until we are green there :)

scripts/cmake/clang_tidy.cmake
```diff
@@ -0,0 +1,97 @@
...
+# <Some_TARGET> -> A target containing c++ source files.
+# 
+# It will generate 2 targets
+#
+# <some_target>_clang_tidy -> Runs Clang Tidy for all files in that target
+# <some_target>_clang_tidy_fix -> Runs Clang Tidy for all files in that target and attempts to fix them 
+#
+# Also it will add 2 targets:
+# "clang_tidy_all" "clang_tidy_fix_all", to run all 
+#  respective targets at once for convinence :)
```
lesleyjanenorton, line 19: *convenience :)

scripts/cmake/clang_tidy.cmake
```diff
@@ -0,0 +1,97 @@
...
+
+
+
+## First Check that we can run clang tidy: 
+# We need: 
+# -> Clang tidy to be found 
+# -> the request for compile_commands to be generated 
+# -> a generator that can actually generate that file
+# 
+# If they are not fullfilled, generate a dummy function and exit.
```
lesleyjanenorton, line 29: *fulfilled

scripts/run-clang-tidy.py
```diff
@@ -0,0 +1,557 @@
...
+                task_queue.put(name)
+
+        # Wait for all threads to be done.
+        task_queue.join()
+        if len(failed_files):
+            return_code = 1
+
+    except KeyboardInterrupt:
+        # This is a sad hack. Unfortunately subprocess goes
+        # bonkers with ctrl-c and we start forking merrily.
```
lesleyjanenorton, line 527: >  we start forking merrily

😂

scripts/cmake/clang_tidy.cmake
```diff
@@ -0,0 +1,97 @@
...
+    get_target_property(aTarget_SOURCE_FILES ${aTarget} SOURCES)
+    get_target_property(aTarget_INTERFACE_FILES ${aTarget} INTERFACE_SOURCES)
+    list(APPEND aTarget_SOURCE_FILES ${aTarget_INTERFACE_FILES})
+
+    # Filter for .cpp files and check source directory location 
+    # Otherwise we might have files from MOC there. 
+    list(FILTER aTarget_SOURCE_FILES INCLUDE REGEX ".*\\.cpp$")
+    list(FILTER aTarget_SOURCE_FILES EXCLUDE REGEX "${CMAKE_BINARY_DIR}/.*")
+
+    add_custom_target(${aTarget}_clang_tidy
```
brizental, line 81: Why do these need to be full blown targets? Wouldn´t it be possible to attach the clang tidy command to each target?

Even so, is the idea that we run clang-tidy everytime we build?

brizental (commented):
Not really sure sure about adding this to CMake, given itś not a build step. I think a script by itself should suffice.

lesleyjanenorton (approved):
✨ welcome back, basti!

scripts/cmake/clang_tidy.cmake
strseb: Fair question! 
I did consider both the "oh let's just use clang tidy as compiler wrapper" and a find "*.cpp" | clang-tidy script. 

-> compiler wrapper imho hidden complexity given i am sure the number will be >1 . i.e i want to use a build cache, or a 2nd linter like clazy, this might get complicated 😅 

-> Building a script looked simple but quickly went out of scope for me: 
- Just looking all at files, how can i make sure they are built? i.e windowscommons on a linux worker is not. Are the adjust sources built, aka was the right config key passed? So i'd need to parse the compile database to get an accurate file list. Then i'd need to write a filter to not lint automoc, uniffi and other generated things, then i'd need a lookup mechanism so clang-tidy is the same as clang used for config, i.e clang-14 vs a clang-17-tidy on path, or android-ndk-clang.  
Which cmake all knows, which is the reason i dropped the script idea 👉 👈

PR #9244, opened 2024-03-19
Run Clang-Tidy on taskcluster

strseb:
This PR adds taskcluster jobs to run clang-tidy. Given we have diffrent c++ source files for each plattform i did create Tasks covering Android,Windows,Macos which should cover most users. 

iOS and Linux should be easy to add, but given this pr is already 27 commit's this will be a Followup. 

I also touched the clang-tidy file, adding a few checks, however only `clang-analyzer-security.*` will produce errors, therefore can block merging, for now :)

### DIFF
--- a/src/cmake/shared-sources.cmake
+++ b/src/cmake/shared-sources.cmake
@@ -247,3 +247,8 @@ include(${CMAKE_SOURCE_DIR}/src/platforms/${MZ_PLATFORM_NAME}/sources.cmake)
 include(${CMAKE_SOURCE_DIR}/src/cmake/sentry.cmake)
 
 mz_add_clang_tidy(shared-sources)
+if(TARGET shared-sources_clang_tidy_report)
+    add_dependencies(shared-sources_clang_tidy_report
+        qtglean
+    )
+endif()
--- a/src/cmake/shared-sources.cmake
+++ b/src/cmake/shared-sources.cmake
@@ -250,5 +250,6 @@ mz_add_clang_tidy(shared-sources)
 if(TARGET shared-sources_clang_tidy_report)
     add_dependencies(shared-sources_clang_tidy_report
         qtglean
+        translations
     )
 endif()
--- a/src/cmake/shared-sources.cmake
+++ b/src/cmake/shared-sources.cmake
@@ -246,6 +246,8 @@ endif()
 include(${CMAKE_SOURCE_DIR}/src/platforms/${MZ_PLATFORM_NAME}/sources.cmake)
 include(${CMAKE_SOURCE_DIR}/src/cmake/sentry.cmake)
 
+add_dependencies(shared-sources translations)
+
 mz_add_clang_tidy(shared-sources)
 if(TARGET shared-sources_clang_tidy_report)
     add_dependencies(shared-sources_clang_tidy_report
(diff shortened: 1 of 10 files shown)


strseb:
> Does clang-tidy need to build the files or is it just a static analyzer?

Yes,ish? It is invoking clang to get the [AST](https://clang.llvm.org/docs/IntroductionToTheClangAST.html) so it's only skipping generating bitcode from that AST, so Compilation would need to work to get a proper ast, so we can't just drop it any file.

brizental:
> Yes,ish? It is invoking [clang ](https://devblogs.microsoft.com/cppblog/exploring-clang-tooling-part-1-extending-clang-tidy/)to get the [AST](https://clang.llvm.org/docs/IntroductionToTheClangAST.html) so it's only skipping generating bitcode from that AST, so Compilation would need to work to get a proper ast, so we can't just drop it any file.

I know I am being annoying about this but... It seems excessive to run a static analyzer multiple times in different platforms just to analyze all files when it can be done by any of them only once. For local development I am still not a big fan, but I get that attempting to figure out what is built for each platform is a pain and something locally we might need more.

Idea: what about running cland-tidy only filed touched since commit X? Or only on staged files if no commit is provided?

strseb:
> analyze all files when it can be done by any of them only once. 

for a few files that might be true but consider i.e controller.cpp, it's used in all builds aka not `/plattforms` but 
```c++
#if defined(MZ_FLATPAK)
#  include "platforms/linux/networkmanagercontroller.h"
#elif defined(MZ_LINUX)
#  include "platforms/linux/linuxcontroller.h"
#elif defined(MZ_MACOS) || defined(MZ_WINDOWS)
#  include "localsocketcontroller.h"
#elif defined(MZ_IOS)
#  include "platforms/ios/ioscontroller.h"
#elif defined(MZ_ANDROID)
#  include "platforms/android/androidcontroller.h"
#else
#  include "platforms/dummy/dummycontroller.h"
#endif
```
On each platform we a have completely different shapes of the input object, depending on the compiler arguments. Even on "cross-plattform" sources.  
On excessiveness, i really do not think that ~10 minutes per target is excessive when it covers the whole codebase, that's roughly 1,5 functional test suites :) 



> Idea: what about running cland-tidy only filed touched since commit X? Or only on staged files if no commit is provided?
Sure i like that, for local development, on ci this would hide existing errors.

brizental:
Ok, I think I misunderstood your first answer i.e. clang-tidy isnot just a static analyzer 🤦‍♀️

src/cmake/shared-sources.cmake
```diff
@@ -246,4 +246,12 @@ endif()
 include(${CMAKE_SOURCE_DIR}/src/platforms/${MZ_PLATFORM_NAME}/sources.cmake)
 include(${CMAKE_SOURCE_DIR}/src/cmake/sentry.cmake)
 
+add_dependencies(shared-sources translations)
+
 mz_add_clang_tidy(shared-sources)
+if(TARGET shared-sources_clang_tidy_report)
```
brizental, line 252: Why is this condition required here?

brizental (approved):
r+wc

src/cmake/shared-sources.cmake
strseb: We do not generate that target i.e if there is no clang-tidy binary found, so the target may or may not exist.

brizental: Sure but the dependencies are being added into this target, so if we don't build this the `add_depencies` also doesn't do anything.

brizental: FWIW, if you don't even create it the `TARGET` keyword is prob unnecessary. It's a nit though.

PR #9248, opened 2024-03-19
VPN-6215 : Fix navbar visibility after viewing reset options

lesleyjanenorton:
## Description

This PR removes `property bool wasNavbarVisible` from ViewReset.qml to fix VPN-6215 because:
- `wasNavbarVisible` overwrites the [default navbar visibility handling](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/162e1b947fcd83252d73dbbe4340f72010d160ca/src/ui/main.qml#L179-L190) which causes the navbar to erroneously stay visible after logout. 
- The `wasNavBarVisible` property is not needed given navbar visibilit is handled in the above code chunk, even when on the Get help menu and Get help sub pages.

## Reference

VPN-6215

## Checklist
    

mcleinman (approved):
👍🏻 Nice find!

PR #9251, opened 2024-03-20
VPN-6236: Re-introduce Balrog staging as a feature

oskirby:
## Description
We need to do some testing against Balrog staging over the next few months. We used to use Balrog stage for QA testing, but this resulted in a lot of manual work for SRE, and our QA channel was moved to Balrog prod in #4741 in order to better automate our release and QA process. We are basically adding this back, but putting it behind a feature flag.

However, the content signing root CA for Balrog staging is up for renewal, so it is also important to implement a little more flexibility when handling the root hash pinning when switching to Balrog stage. In order to do this, we have refactored the API to the `signature` crate in order to migrate the root hash verification job into the Qt/C++ world.

## Reference
Github PR removing balrog stage: #4741
Bugzilla staging CA rotation: [1882192](https://bugzilla.mozilla.org/show_bug.cgi?id=1882192)
JIRA ticket: [VPN-6236](https://mozilla-hub.atlassian.net/browse/VPN-6236)

## Checklist
    
[VPN-6236]: https://mozilla-hub.atlassian.net/browse/VPN-6236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/signature/src/balrog.rs
+++ b/signature/src/balrog.rs
@@ -16,7 +16,6 @@
  */
 
 use asn1_rs::ToDer;
-use hex;
 use ring::digest;
 use x509_parser::prelude::*;
 
@@ -115,7 +114,7 @@ impl<'a> Balrog<'_> {
     }
 
     /* Ensure that the certificate chain is valid. */
-    pub fn verify_chain(&self, current_time: i64, root_hash: &str) -> Result<(), BalrogError> {
+    pub fn verify_chain(&self, current_time: i64) -> Result<(), BalrogError> {
         if self.chain.is_empty() {
             return Err(BalrogError::CertificateNotFound);
         }
@@ -143,14 +142,6 @@ impl<'a> Balrog<'_> {
         /* Verify the root certificate. */
         let root_cert = self.chain.last().unwrap();
         Self::verify_cert_chain_pair(root_cert, root_cert, current_time)?;
-
-        let hash_stripped = root_hash.replace(":", "");
-        let Ok(hash_decoded) = hex::decode(hash_stripped) else {
-            return Err(BalrogError::RootHashParseFailed);
-        };
-        if self.root_hash.is_empty() || (self.root_hash != hash_decoded) {
-            return Err(BalrogError::RootHashMismatch);
-        }
 
         /* Success! */
         Ok(())
@@ -346,11 +337,10 @@ impl<'a> Balrog<'_> {
         input: &[u8],
         signature: &str,
         current_time: i64,
-        root_hash: &str,
         leaf_subject: &str,
     ) -> Result<(), BalrogError> {
         /* Verify things. */
-        self.verify_chain(current_time, root_hash)?;
+        self.verify_chain(current_time)?;
         self.verify_leaf_hostname(leaf_subject)?;
         self.verify_content_signature(input, signature)?;
     
@@ -364,7 +354,7 @@ mod test {
     use super::*;
 
     // Copied from https://github.com/mozilla/application-services/blob/main/components/support/rc_crypto/src/contentsignature.rs
-    const ROOT_HASH: &str = "3C:01:44:6A:BE:90:36:CE:A9:A0:9A:CA:A3:A5:20:AC:62:8F:20:A7:AE:32:CE:86:1C:B2:EF:B7:0F:A0:C7:45";
+    const ROOT_HASH_HEX: &str = "3C01446ABE9036CEA9A09ACAA3A520AC628F20A7AE32CE861CB2EFB70FA0C745";
     const VALID_CERT_CHAIN: &[u8] = include_bytes!("../assets/valid_cert_chain.pem");
     const VALID_INPUT: &[u8] = b"{\"data\":[],\"last_modified\":\"1603992731957\"}";
     const VALID_SIGNATURE: &str = "fJJcOpwdnkjEWFeHXfdOJN6GaGLuDTPGzQOxA2jn6ldIleIk6KqMhZcy2GZv2uYiGwl6DERWwpaoUfQFLyCAOcVjck1qlaaEFZGY1BQba9p99xEc9FNQ3YPPfvSSZqsw";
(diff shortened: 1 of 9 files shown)


signature/src/balrog.rs
```diff
@@ -361,14 +316,30 @@ impl<'a> Balrog<'_> {
             &pubkey, &algorithm, &sig_asn1, &message,
         )?)
     }
+
+    /* Wrapper function to verify everything at once. */
```
bhearsum, line 320: I will note that you are now verifying a _little_ less than before here, because `balrog.cpp` is validating the hash of the root cert -- but I also think that it's OK to separate out verification of the signature, and verification of what made the signature.

(Mostly just calling this out in case you wanted to update comments.)

bhearsum (approved):
> However, the root CA for Balrog staging is up for renewal, so it is also important to implement a little more flexibility when handling the root hash pinning when switching to Balrog stage. In order to do this, we have refactored the API to the signature crate in order to migrate the root hash verification job into the Qt/C++ world.

Just to clarify, there's nothing changing when it comes to Balrog SSL certificates. What's changing is the certificate used to generate the `content-signature` header in Balrog's response. (That certificate lives in Autograph, which Balrog talks to to sign the responses.)

This looks fine to me, but I will note that I am far from a crypto expert. @moz-hwine - is there anyone else you think ought to review this?

src/update/balrog.cpp
```diff
@@ -461,3 +485,34 @@ void Balrog::propagateError(NetworkRequest* request,
...
   REPORTNETWORKERROR(error, m_errorPropagationPolicy, "balrog");
 }
+
+// static
+QString Balrog::balrogUrl() {
+  const QString product = "FirefoxVPN";
+  QString hostname = Constants::BALROG_PROD_HOSTNAME;
+  QString channel = "release";
+
+  if (Feature::get(Feature::Feature_stagingUpdateServer)->isSupported()) {
```
bhearsum, line 495: Does this allow for runtime switching to stage? If not, is that something you want to do? (It would probably make future testing easier not to require special builds...)

oskirby: Yes, this feature can be switched at runtime through the developer menu.

src/update/balrog.cpp
```diff
@@ -181,10 +184,31 @@ bool Balrog::checkSignature(Task* task, const QByteArray& x5uData,
...
+  if (!compute_root_certificate_hash(x5uData.constData(), x5uData.length(),
+                                     rootHash.data(), rootHash.length())) {
+    logger.error() << "Failed to calculate root certificate hash";
+    return false;
+  }
+  bool hashOkay = false;
+  for (const QString& fingerprint : rootCertHashes()) {
+    QByteArray fpDecode = QByteArray::fromHex(fingerprint.toUtf8());
+    Q_ASSERT(fpDecode.length() == hashSize);
+    if (rootHash == fpDecode) {
```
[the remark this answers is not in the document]
oskirby: Yes, that is correct. That is handled in the C++ code in `Balrog::fetchSignature()` which parses the `Content-Signature` header and then spawns a network request to download the certificate chain to use for this verification step.

signature/src/balrog.rs
oskirby: Good callout, thank you.

src/constants.h
```diff
@@ -231,14 +231,17 @@ PRODBETAEXPR(QString, captivePortalUrl, "http://%1/success.txt",
...
-    "update.json");
-constexpr const char* AUTOGRAPH_ROOT_CERT_FINGERPRINT =
+constexpr const char* BALROG_PROD_HOSTNAME = "aus5.mozilla.org";
+constexpr const char* AUTOGRAPH_PROD_FINGERPRINTS =
     "97e8ba9cf12fb3de53cc42a4e6577ed64df493c247b414fea036818d3823560e";
 
+constexpr const char* BALROG_STAGE_HOSTNAME =
+    "stage.balrog.nonprod.cloudops.mozgcp.net";
+constexpr const char* AUTOGRAPH_STAGE_FINGERPRINTS =
+    "c0f05d59b1fde25780854c32fae8faba8481c233b4c1d390cca5f2cea81930ee\n"  // cas-new-2024-03-12
```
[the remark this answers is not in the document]
oskirby: I think it does make the declaration cleaner, but there isn't a nice method to convert a `const char*[]` into a QStringList like we get with `QString(fingerprints).split('/n')`

src/feature/featurelistcallback.h
```diff
@@ -174,4 +174,12 @@ bool FeatureCallback_webPurchase() {
 #endif
 }
 
+bool FeatureCallback_hasBalrog() {
+#if defined(MVPN_BALROG)
```
lesleyjanenorton, line 178: Should this also be restricted to `FeatureCallback_inStaging()` for now?

src/feature/featurelist.h
```diff
@@ -202,6 +202,13 @@ FEATURE(startOnBoot,            // Feature ID
         QStringList(),          // feature dependencies
         FeatureCallback_startOnBoot)
 
+FEATURE(stagingUpdateServer,        // Feature ID
+        "Staging Update Server",    // Feature name
+        FeatureCallback_hasBalrog,  // Can be flipped on
+        FeatureCallback_false,      // Can be flipped off
```
lesleyjanenorton, line 208: Does this mean that once the feature is enabled it can't be disabled?

lesleyjanenorton (approved):

src/feature/featurelistcallback.h
oskirby: I think it's fine to leave this enabled for prod as well, that gives us the ability to test updates and key switching in prod too.

src/feature/featurelist.h
oskirby: I don't really understand the logic behind our feature flagging methods, but it can be turned off just fine. However, when I replace this with `FeatureCallback_true` it resulted in a default-enabled feature which we definitely don't want here.

hwine:
@bhearsum wrote:
> This looks fine to me, but I will note that I am far from a crypto expert. @moz-hwine - is there anyone else you think ought to review this?

Neither am I (a crypto or rust expert), so:
For crypto review, please follow the [2nd bullet under "special cases" here](https://mozilla-hub.atlassian.net/wiki/spaces/SLP/pages/27394064/Licensing+Runbook+Software+Licensing+And+Code+Contribution+Between+Mozilla+And+Other+Orgs+Projects)

PR #9252, opened 2024-03-20
Fix TC android builds

strseb:
## Description

in #9239 i added support for clang tidy, trying to not add dependencies if the clang_tidy target was not created. 
I did not consider that those can depend on targets that are not created for all plattforms i.e sentry is not built on 32bit android.
This caused all 32bit android builds to go bust on main, while the PR was green. 
 
given this, let's add a util that adds dependencies between 2 targets if they both exist, so it's a bit more clean :)


mcleinman (approved):
👍🏻  Thanks!

PR #9259, opened 2024-03-21
Nit: Compiler Warnings as Errors when using MSVC

strseb:
## Description
For clang& gcc we set `-Wall -Werror -Wno-conversion`
for windows we skip this. I just noticed a warning and thought, let's fix them all :)


src/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -670,7 +670,7 @@ bool WindowsFirewall::blockTrafficTo(const IPAddress& addr, uint8_t weight,
   filter.weight.uint8 = weight;
   filter.subLayerKey = ST_FW_WINFW_BASELINE_SUBLAYER_KEY;
 
-  FWPM_FILTER_CONDITION0 cond[1] = {0};
+  FWPM_FILTER_CONDITION0 cond[1] = {};
```
strseb, line 673: i could also write `{{0}}`, but it's the same.

oskirby (approved):
These changes look good to me.

PR #9276, opened 2024-03-25
VPN-6120 - Remove Tips and Tricks related code

brizental:
This started with me just removing the Tips and Tricks modal code... 

Then @mcleinman rightfully pointed out to me that that whole screen is gone.

Then I removed the screen code.

If we don't need that screen, then we also don't need guides. So then I also removed guides code.

While removing guides code, I noticed there were still some tutorials code lying around... So I removed that as well.


mcleinman (approved):
👍🏻 Thanks for this. Should `ScreenTipsAndTricks` and `ViewTipsAndTricks` also go?

brizental:
> 👍🏻 Thanks for this. Should ScreenTipsAndTricks and ViewTipsAndTricks also go?

Ha, I didn't realize that was goen as well. Will remove.

mcleinman:
I think we should keep guides in addons for older clients. What do you think?

brizental:
> I think we should keep guides in addons for older clients. What do you think?

Makes me sad, but you are right. I added it back, even the tutorials 😞

src/ui/screens/settings/ViewPreferences.qml
```diff
@@ -31,6 +31,14 @@ MZViewBase {
 
         spacing: MZTheme.theme.windowMargin
 
+        Item {
```
mcleinman, line 34: Should this be in here?

nebula/ui/components/MZToggleRow.qml
```diff
@@ -59,13 +59,18 @@ RowLayout {
         }
     }
 
+    Text {
```
mcleinman, line 62: Are the changes in this file related to the PR?

nebula/ui/components/MZSettingsToggle.qml
```diff
@@ -15,11 +15,14 @@ CheckBox {
     property var uiState: MZTheme.theme.uiState
     property alias forceFocus: vpnSettingsToggle.focus
     property string accessibleName
+    property bool accessibleEnabled
```
mcleinman, line 18: Is this related to the PR?

mcleinman (commented):
Thank you for this! I have a few questions about some parts.

flodolo:
> I think we should keep guides in addons for older clients. What do you think?

Is this really necessary? I'm concerned about keeping around content (which is translated) for the time being.

brizental:
> Is this really necessary? I'm concerned about keeping around content (which is translated) for the time being.

The main concern here is that users in older versions of the application would attempt to go into the Tips and tricks screen and would just get a whiet screen. A middle ground here would be to leave one single guide there, such as the "How to VPN", just so users still see something in that screen. I will bring this option to the team and see how they feel about it.

flodolo:
> [EDIT 2] As discussed over slack, we are leaving guides for now. We will go ahead and remove tutorials though.

For the future: edit to comments don't notify folks, I didn't realize there were updates here ;-)

flodolo (approved):

docs/Components/guides.md
flodolo, line 1: > As discussed over slack, we are leaving guides for now. We will go ahead and remove tutorials though.

Should this stay then?

mcleinman: This looks right to me, as we're doing two things in this PR:
App - Everything related to guides/tutorials is being removed, we don't want them in the app.
Addons: Guides are being kept, to support old clients.

mcleinman (approved):
Looks good, thanks.

PR #9285, opened 2024-03-26
VPN-6224: Prevent default checkboxes from appearing on hover 

lesleyjanenorton:
## Description
This is a hack to prevent default checkboxes from appearing over our stylized toggles on Windows. We do  the same thing [here ](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/d9aca561a638cf2985ac9d60ed56502844cd6b85/nebula/ui/components/MZCheckBox.qml#L25)in MZCheckbox. 

Here is a crummy video taken with my iPhone since screen casting on Windows seems (at least at current energy level) non-trivial. 

https://github.com/mozilla-mobile/mozilla-vpn-client/assets/22355127/3023bd79-d10f-4149-81fd-4ef6cff13443

## Reference

VPN-6224

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZSettingsToggle.qml
+++ b/nebula/ui/components/MZSettingsToggle.qml
@@ -24,6 +24,11 @@ CheckBox {
 
     implicitHeight: MZTheme.theme.toggleHeight
     implicitWidth: MZTheme.theme.toggleWidth
+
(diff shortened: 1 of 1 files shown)


mcleinman (approved):
👍🏻

PR #9288, opened 2024-03-28
VPN-4979 part 2: Add silent server switching to iOS network extension

mcleinman:
## Description

This is the second of 3 PRs for VPN-4979:
1. Set up connection health monitoring ([done](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9146))
2. Add silent server switching (this one)
3. Add telemetry (to be done after this one is merged)

Some notes:
- Let me know if you have any questions about the iOS/Swift stuff - I'm happy to explain.
- There is only one backup server. A second silent server switch will fail. (We need to set the backup servers when the app is running. We can add additional backup servers if desired, it wouldn't be too hard.)
- When you change locations, it loads a fresh backup server (in the proper region).
- This works with multihop.
- In connectionhealth, the timer start moved. The old spot would start a timer even if the setup failed, which is not what we want.
- Includes small refactor of how we select the backup server, done to share code between iOS and Android.
- Includes small refactor of how we update the iOS config to make it more generalizable. This also removes some code passed to the callback (see `self.adapter.getRuntimeConfiguration`). The only callsite for this message (`ioscontroller.swift`, line 215 in this PR) does not use the returned data, and thus it is unnecessary to gather and return it.
- Includes small refactor of how we send server data from Obj C++ to Swift to make it clearer and easier to send a backup server's info.

Testing:
- I've added a dev menu debug option to force an iOS daemon (AKA network extension) silent server switch.
- The connection information (the `i` in the top right when connected to VPN) will be out of date after forcing the silent server switch. It is refreshed when the app relaunches. Since the network extension's silent server switch should only fire when the app is quit (otherwise the app's silent server switch should kick in), the only time this should be obvious is when testing by activating a silent server switch from that debug menu.
- Two ways to test: Use a "what is my IP" website, or inspect the app's connection info menu (but you must hard quit the app and relaunch it after activating the silent server switch).

## Reference

VPN-4979

## Checklist
    

mcleinman:
ConnectionHealth seems to be broken. It's throwing a `PingError.hostnotfound` when setting it up. I'm going to investigate that and fix it in another PR. It's still possible to test this work with the dev menu option.

src/controller.cpp
```diff
@@ -1059,3 +1059,15 @@ bool Controller::deactivate() {
   m_impl->deactivate(stateToReason(m_state));
   return true;
 }
+
+void Controller::forceDaemonSilentServerSwitch() {
+#ifdef MZ_IOS
+  if (m_impl) {
```
brizental, line 1065: Consider logging errors if m_impl is not present.

src/controller.cpp
```diff
@@ -1059,3 +1059,15 @@ bool Controller::deactivate() {
...
+    logger.debug() << "Sending server switch message to iOS Network Extension";
+    m_impl->forceDaemonSilentServerSwitch();
+  }
+#else
+  logger.debug() << "Server switch debug feature only available for iOS. Not "
```
brizental, line 1070: Why make this API even available in platforms other than iOS?

src/models/serverdata.h
```diff
@@ -64,6 +64,7 @@ class ServerData final : public QObject {
   const QString& exitCityName() const { return m_exitCityName; }
   QString localizedExitCountryName() const;
   QString localizedExitCityName() const;
+  const Server backupServer(const QString& currentPublicKey);
```
brizental, line 67: nit:
```suggestion
  const Server backupServer(const QString& currentPublicKey) const;
```

src/platforms/android/androidcontroller.cpp
```diff
@@ -179,15 +179,8 @@ void AndroidController::activate(const InterfaceConfig& config,
...
-  const QList<Server> serverList =
-      vpn->controller()->currentServer().exitServers();
-  Server* fallbackServer = nullptr;
-  foreach (auto item, serverList) {
-    if (item.publicKey() != config.m_serverPublicKey) {
-      fallbackServer = &item;
-      break;
-    }
-  }
+  Server* fallbackServer =
```
brizental, line 182: nit: Since this function doesn't return a pointer anymore, this can be just 'Server', no?

src/platforms/ios/ioscontroller.swift
```diff
@@ -116,52 +135,31 @@ public class IOSControllerImpl : NSObject {
...
-
-            var peerConfigurations: [PeerConfiguration] = []
-            peerConfigurations.append(peerConfiguration)
-
-            var interface = InterfaceConfiguration(privateKey: self.privateKey!)
-
-            if let ipv4Address = IPAddressRange(from: deviceIpv4Address!),
-               let ipv6Address = IPAddressRange(from: deviceIpv6Address!) {
-                interface.addresses = [ipv4Address, ipv6Address]
+            let config = createConfig(for: serverData, allowedIPAddressRanges: allowedIPAddressRanges)
```
brizental, line 145: Potentially out-of-scope: why do we provide the allowedIPAddressRanges to iOS if it doesn't use it?

src/platforms/ios/ioscontroller.swift
```diff
@@ -116,52 +135,31 @@ public class IOSControllerImpl : NSObject {
         }
     }
 
-    @objc func connect(dnsServer: String, serverIpv6Gateway: String, serverPublicKey: String, serverIpv4AddrIn: String, serverPort: Int,  excludeLocalNetworks: Bool, allowedIPAddressRanges: Array<VPNIPAddressRange>, reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
+    @objc func connect(serverData: VPNServerData, fallbackServer: VPNServerData?, excludeLocalNetworks: Bool, allowedIPAddressRanges: [VPNIPAddressRange], reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
```
brizental, line 138: Nice call putting this all in a struct.

src/platforms/ios/ioscontroller.swift
```diff
@@ -116,52 +135,31 @@ public class IOSControllerImpl : NSObject {
...
-            interface.dns = [DNSServer(address: dnsServerIP!), DNSServer(address: ipv6GatewayIP!)]
 
-            let config = TunnelConfiguration(name: VPN_NAME, interface: interface, peers: peerConfigurations)
-
-            return self.configureTunnel(config: config, reason: reason, serverName: serverIpv4AddrIn + ":\(serverPort )", excludeLocalNetworks: excludeLocalNetworks, gleanDebugTag: gleanDebugTag, isSuperDooperFeatureActive: isSuperDooperFeatureActive, installationId: installationId, disconnectOnErrorCallback: disconnectOnErrorCallback, onboardingCompletedCallback: onboardingCompletedCallback, vpnConfigPermissionResponseCallback: vpnConfigPermissionResponseCallback)
+            return self.configureTunnel(config: config, reason: reason, serverName: serverData.ipv4AddrIn + ":\(serverData.port )", fallbackConfig: fallbackConfig, excludeLocalNetworks: excludeLocalNetworks, gleanDebugTag: gleanDebugTag, isSuperDooperFeatureActive: isSuperDooperFeatureActive, installationId: installationId, disconnectOnErrorCallback: disconnectOnErrorCallback, onboardingCompletedCallback: onboardingCompletedCallback, vpnConfigPermissionResponseCallback: vpnConfigPermissionResponseCallback)
```
brizental, line 153: nit: Consider adding a function to the VPNServerData struct that returns serverName and skip the extra argument?

src/platforms/ios/ioscontroller.swift
```diff
@@ -178,6 +176,8 @@ public class IOSControllerImpl : NSObject {
             configHack["isSuperDooperFeatureActive"] = isSuperDooperFeatureActive
             configHack["gleanDebugTag"] = gleanDebugTag
             configHack["installationId"] = installationId
+
+            configHack["fallbackConfig"] = fallbackConfig?.asWgQuickConfig() ?? ""
```
brizental, line 180: Random: at this point we should consider renaming this 'configHack' to something more semantic. We are using it for so many things already.

src/platforms/ios/iostunnel.swift
```diff
@@ -203,35 +205,43 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        }
+    }
+
+    func silentServerSwitch() {
+        if let fallbackServerConfig = ((protocolConfiguration as? NETunnelProviderProtocol)?.providerConfiguration?["fallbackConfig"] as? String), !fallbackServerConfig.isEmpty {
+            self.logger.info(message: "Sending silent server switch in Network Extension")
+            updateServerConfig(with: fallbackServerConfig)
+            (protocolConfiguration as? NETunnelProviderProtocol)?.providerConfiguration?["fallbackConfig"] = ""
+        } else {
+            logger.info(message: "Silent server switch called, but no fallbackConfig available")
```
brizental, line 244: nit: logger.error.

src/platforms/ios/ioscontroller.swift
```diff
@@ -300,4 +300,56 @@ public class IOSControllerImpl : NSObject {
...
+        guard let session = TunnelManager.session else {
+            IOSControllerImpl.logger.info(message: "No tunnel session available to send silentServerSwitch message");
+            return
+        }
+
+        let message = TunnelMessage.silentServerSwitch;
+        IOSControllerImpl.logger.info(message: "Sending new message \(message)");
+        do {
+            try session.sendProviderMessage(message.encode()) { _ in
+                // IMPORANT: Must keep this log line (or other code) in this closure.
```
brizental, line 314: IMPORTANT*

src/platforms/android/androidcontroller.cpp
```diff
@@ -179,15 +179,8 @@ void AndroidController::activate(const InterfaceConfig& config,
...
+      vpn->controller()->currentServer().backupServer(config.m_serverPublicKey);
   QJsonObject jFallbackServer;
   if (fallbackServer) {
```
brizental, line 185: This check will always be true now that server is nota. pointer anymore.

brizental (changes requested):

src/controller.cpp
mcleinman: I couldn't find a way, but I may have missed something. We need to trigger this from the `ViewDeveloperMenu` QML screen. We have a connection to the controller class, but not any individual platform implementations. (Adding a connection just to iOS's controller seemed heavy handed.) And then within the controller class, we only can call functions on the abstract class. (We can't cast the `m_impl` to `ioscontroller`, for example.)

I'm very open to suggestions for better ways to handle this, because I agree this is pretty messy, especially for a debug option needed just for testing.

src/platforms/ios/ioscontroller.swift
mcleinman: 🤦🏻 Thanks, updated

src/controller.cpp
mcleinman: Added, thanks

src/platforms/ios/iostunnel.swift
mcleinman: It's not necessarily an error - 
1. In geos that only have one server, we expect it to be empty. 
2. If we've already used the fallback server, we'd also expect to be empty.

In light of that, do you think it's okay to keep it as an info?

src/platforms/ios/ioscontroller.swift
mcleinman: Good call. I've updated it to `configManipulation`. If you have an even better name, please let me know - I don't think `configManipulation` is the best one possible, but it's the best I'm coming up with right now.

src/platforms/ios/ioscontroller.swift
mcleinman: We do, in `ioscontroller.swift` around line 334.

Or do you mean "our architecture doesn't use it"? Great question. It is part of the Wireguard config, which is what is happening there. I'd want to dig in to make sure there isn't something surprising happening before we rip it out - feel free to add a spike ticket if you want - I'd be happy to dig in (outside this ticket).

src/platforms/android/androidcontroller.cpp
mcleinman: Good catch, updated. Thanks.

src/platforms/ios/ioscontroller.swift
brizental: Yeah I mean the architecture. We have a bunch of logic around this list, but it is a constant list (0.0.0.0/0, ::) so all of that is kind of unnecessary and even confusing if you are not aware this is unused.

src/platforms/ios/iostunnel.swift
brizental: Yeap, that sounds reasonable.

src/controller.cpp
brizental: We can wrap the header declaration in ifdef MZ_IOS. But no hard sell, this is just a nit.

src/platforms/ios/ioscontroller.swift
brizental: Hm, SGTM... Brainstorming: 'configExtension'? 'extraConfig'? 'customConfig'? 'mozConfig'?

src/platforms/ios/ioscontroller.swift
brizental: Ah, duh 🤦‍♀️ That's right.

src/platforms/ios/ioscontroller.swift
mcleinman: I'm not aware, but I also have dug into the wireguard internals to see what (if anything) they do for it.

src/platforms/ios/ioscontroller.swift
mcleinman: I like `customConfig; I've updated this.

src/controller.cpp
mcleinman: Ah sorry, I thought you were going for something otherwise - in my opinion, it's an ugly amount and type of code for a one-platform debug option. I might keep it as is for now - in case we want to add this into Android (for example) in the future.

src/platforms/ios/ioscontroller.swift
brizental: Wireguard uses that to figure out what should go through the tunnel or not. But we dont rely on it for iOS, we use 'excludeLocalNetworks' instead -- https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/controller.cpp#L513-L524.

PR #9310, opened 2024-04-02
VPN-2656 Support Windows CLI 

strseb:
## Description
So far when building in MZ_DEBUG we allocated a console instead of using the perfectly fine ones in my IDE >:( 
This pr changes so that we attach to a parent console if it exists, so the client behaves the same way like we do on MacOS. (just as broken lmao) 

This also enables use to use the cli. 
image

## Reference

https://mozilla-hub.atlassian.net/browse/VPN-2656

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4,11 +4,31 @@
 
 #include "commandlineparser.h"
 #include "leakdetector.h"
+#include "stdio.h"
+#ifdef MZ_WINDOWS
+
+#  include <windows.h>
+
+#  include <iostream>
+
+#endif
 
 Q_DECL_EXPORT int main(int argc, char* argv[]) {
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -4,13 +4,33 @@
 
 #include "commandlineparser.h"
 #include "leakdetector.h"
+#include "stdio.h"
+#ifdef MZ_WINDOWS
+
+#  include <windows.h>
+
+#  include <iostream>
+
```
brizental, line 13: nit: Unnecessary white space.

brizental (approved):
Love it, really wan an annoying behaviour.

lesleyjanenorton (approved):
👍

PR #9326, opened 2024-04-04
VPN-6290 - Stop using the glean sdk generator script

brizental:
~We should do the same for Android. Let's do it in a separate PR because this is currently breaking CI IIUC.~

Doing Android here since fixing things for iOS broke them for Android.

### DIFF
--- a/android/buildSrc/src/main/kotlin/Config.kt
+++ b/android/buildSrc/src/main/kotlin/Config.kt
@@ -3,9 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object Config {
-    const val compileSdkVersion = 33
+    const val compileSdkVersion = 34
     const val buildToolsVersion = "30.0.3"
     const val minSdkVersion = 26
-    const val targetSdkVersion = 33
+    const val targetSdkVersion = 34
     const val ndkVersion = "23.1.7779620"
 }
--- a/android/daemon/build.gradle
+++ b/android/daemon/build.gradle
@@ -22,28 +22,17 @@ buildscript {
 
     dependencies {
         classpath SharedDependencies.com_android_tools_build_gradle
-        classpath "org.mozilla.telemetry:glean-gradle-plugin:52.7.0"
     }
 }
 
 plugins {
     id 'org.jetbrains.kotlin.plugin.serialization'
-    id "com.jetbrains.python.envs" version "0.0.31"
 }
 
 apply plugin: "com.android.library"
 apply plugin: "org.jetbrains.kotlin.android"
 
-apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
 
-ext.gleanNamespace = "mozilla.telemetry.glean"
-ext.gleanYamlFiles = [
-    "$rootDir/../../../src/telemetry/metrics.yaml",
-    "$rootDir/../../../src/telemetry/interaction_metrics.yaml",
-    "$rootDir/../../../src/telemetry/impression_metrics.yaml",
-    "$rootDir/../../../src/telemetry/outcome_metrics.yaml",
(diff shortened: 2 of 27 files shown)


brizental:
@strseb I am assuming now we don't need to run gradle twice in cmake.sh? From local testing it _seems_ to work -- the apk was created fine with just `cmake build --build -j${nproc}`... Let me know though.

android/daemon/build.gradle
```diff
@@ -52,6 +41,8 @@ android {
         minSdkVersion Config.minSdkVersion
         targetSdkVersion Config.targetSdkVersion
 
+        buildConfigField "int", "VERSION_CODE", System.getenv("VERSIONCODE") ?: "99999"
+        buildConfigField "String", "VERSION_NAME", '"' + (System.getenv("SHORTVERSION") ?: "") + '"'
         buildConfigField  "String", "VERSIONCODE" ,  '"' +System.getenv("VERSIONCODE") + '"'
```
strseb, line 46: Why twice?

android/vpnClient/build.gradle
```diff
@@ -67,6 +61,14 @@ android {
             includeAndroidResources = true
         }
     }
+
+    sourceSets {
+        main {
+            kotlin.srcDirs = [
+                "${System.getProperty('user.dir')}/../../clienttelemetry"
```
strseb, line 68: Not a fan of this, can you either infer that from `$rootDir` or get that copied into the project?

android/buildSrc/src/main/kotlin/Config.kt
```diff
@@ -3,9 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object Config {
-    const val compileSdkVersion = 33
+    const val compileSdkVersion = 34
     const val buildToolsVersion = "30.0.3"
     const val minSdkVersion = 26
-    const val targetSdkVersion = 33
+    const val targetSdkVersion = 34
```
strseb, line 9: There is one more thing to to 
https://developer.android.com/about/versions/14/behavior-changes-14#fgs-types. 

In the daemon android manifest you need to set a type: https://developer.android.com/about/versions/14/changes/fgs-types-required

android/daemon/build.gradle
brizental: Oh, one is "VERSIONCODE" and the other is "VERSION_CODE" 👀 

It's good you commented here. It reminds me I had a little issue with these variables. 

You and the internet had told me I should use `versionName` and `versionCode` here and they would set these variables. That didn't work though, that is why I used `buildConfigField` here instead.

Also, I set these in a bunch of places, but I could not figure out how to just set it once. I don't like the way it is right now :(

android/vpnClient/build.gradle
brizental: I can infer from `$rootDir`. But that would require me to assume the name of the build folder, which doesn't seem right. I also didn't want to generate these files outside of the build folder... I can copy them into the projec thought. Hadn't thought of that.

brizental: > I can infer from $rootDir. But that would require me to assume the name of the build folder, which doesn't seem right.

Neverming me being wrong here... 🤦‍♀️

android/settings.gradle
```diff
@@ -9,11 +9,22 @@ buildCache {
...
     }
 }
+
+def repositoryRoot = "$rootDir/../../.."
+def gleanVendoredPath = "$repositoryRoot/3rdparty/glean"
+
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("gleanLibs") {
+            from(files("$gleanVendoredPath/gradle/libs.versions.toml"))
```
strseb, line 19: Neeeat

strseb (approved):

PR #9343, opened 2024-04-08
VPN-6299: Implement start at boot for Flatpaks

oskirby:
## Description
Sandboxed applications don't have access to `/etc` which prevents us from being able to register a traditional autostart handler. Instead, the recommended way to launch an application on startup is to make use of the XDG [Background](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html) portal.

## Reference
JIRA Task: [VPN-6299](https://mozilla-hub.atlassian.net/browse/VPN-6299)

## Checklist
    
[VPN-6299]: https://mozilla-hub.atlassian.net/browse/VPN-6299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/xdgstartatbootwatcher.cpp
+++ b/src/platforms/linux/xdgstartatbootwatcher.cpp
@@ -4,10 +4,6 @@
 
 #include "xdgstartatbootwatcher.h"
 
-#include "leakdetector.h"
-#include "logger.h"
-#include "settingsholder.h"
-
 #include <QDBusConnection>
 #include <QDBusMessage>
 #include <QDBusObjectPath>
@@ -15,13 +11,18 @@
 #include <QDBusPendingReply>
 #include <QRandomGenerator>
 
+#include "leakdetector.h"
+#include "logger.h"
+#include "settingsholder.h"
+
 namespace {
 Logger logger("XdgStartAtBootWatcher");
 }
 
 constexpr const char* XDG_PORTAL_SERVICE = "org.freedesktop.portal.Desktop";
 constexpr const char* XDG_PORTAL_PATH = "/org/freedesktop/portal/desktop";
-constexpr const char* XDG_PORTAL_BACKGROUND = "org.freedesktop.portal.Background";
+constexpr const char* XDG_PORTAL_BACKGROUND =
+    "org.freedesktop.portal.Background";
 constexpr const char* XDG_PORTAL_REQUEST = "org.freedesktop.portal.Request";
 
 XdgStartAtBootWatcher::XdgStartAtBootWatcher() : QObject() {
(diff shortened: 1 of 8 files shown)


mcleinman:
This is an area I'm pretty unfamiliar with (both in specifics of our code and generalities of Linux). I'm going to remove myself and let @strseb handle this one. If you end up needing me to review, feel free to re-assign to me.

src/platforms/linux/xdgstartatbootwatcher.cpp
```diff
@@ -0,0 +1,156 @@
...
+  logger.debug() << "StartAtBoot responded:" << response;
+  for (auto i = results.cbegin(); i != results.cend(); i++) {
+    logger.debug() << "StartAtBoot" << QString("%1:").arg(i.key())
+                   << i.value().toString();
+  }
+}
+
+// Try to find the window identifier for an XDG desktop portal request.
+// https://flatpak.github.io/xdg-desktop-portal/docs/window-identifiers.html
+QString XdgStartAtBootWatcher::parentWindow() {
```
strseb, line 82: Consider making this `std::excpeted<const QString,SomeErrorEnum>` so the caller need's to check if this succeed.

src/platforms/linux/xdgstartatbootwatcher.cpp
```diff
@@ -0,0 +1,156 @@
...
+  QVariantMap options;
+  options["autostart"] = QVariant(startAtBoot);
+  options["background"] = QVariant(true);
+  options["commandline"] = QVariant(cmdline);
+  options["handle_token"] = QVariant(m_token);
+
+  QDBusMessage call = QDBusMessage::createMethodCall(
+      XDG_PORTAL_SERVICE, XDG_PORTAL_PATH, XDG_PORTAL_BACKGROUND,
+      "RequestBackground");
+  call << parentWindow();
```
strseb, line 124: Seems this can be "" if the windowing system is unsupported. Does it still make sense to do the call, or should this be checked and aborted if not?

src/platforms/linux/xdgstartatbootwatcher.cpp
```diff
@@ -0,0 +1,156 @@
...
+  QDBusConnection bus = QDBusConnection::sessionBus();
+  QDBusPendingReply<QDBusObjectPath> reply = bus.asyncCall(call);
+  QDBusPendingCallWatcher* watcher = new QDBusPendingCallWatcher(reply, this);
+  connect(watcher, &QDBusPendingCallWatcher::finished, this,
+          &XdgStartAtBootWatcher::callCompleted);
+  connect(watcher, &QDBusPendingCallWatcher::finished, watcher,
+          &QObject::deleteLater);
+}
+
+void XdgStartAtBootWatcher::callCompleted(QDBusPendingCallWatcher* call) {
```
strseb, line 137: Consider checking for nullptr

src/platforms/linux/xdgstartatbootwatcher.cpp
```diff
@@ -0,0 +1,156 @@
...
+  QDBusPendingReply<QDBusObjectPath> reply = *call;
+  if (reply.isError()) {
+    logger.error() << "Failed to set startOnBoot:" << reply.error().message();
+    return;
+  }
+
+  // We need to rebind our signals if the reply path changed.
+  QString path = reply.argumentAt<0>().path();
+  logger.debug() << "Expecting XDG response at:" << path;
+  if (!path.isEmpty() && path != m_replyPath) {
```
strseb, line 147: Given there is no else, consider doing an early return :)

strseb (approved):
LGTM, r+wc

src/platforms/linux/xdgstartatbootwatcher.cpp
oskirby: The XDG specs go out of their way to say that the window handle should be an empty string if the windowing system is unsupported. The purpose of the window handle in this case is to tell the windowing system where to display a UAC-style popup message if the action requires user approval, and if we provide an empty string, the window system just throws it in the middle of the screen.

Note that I have yet to find any desktop environment that actually displays such a prompt for this permission, so this code is all just theoretical anyways.

src/platforms/linux/xdgstartatbootwatcher.cpp
oskirby: I think it's more in-line with the XDG specs to return an empty string here, as that is the expected value of the window handle in the event that the windowing system doesn't support it.

src/platforms/linux/xdgstartatbootwatcher.cpp
oskirby: Will do!

src/platforms/linux/xdgstartatbootwatcher.cpp
strseb: I see thanks for explaining :)

PR #9351, opened 2024-04-10
Modularize Leakdetector

strseb:
## Description

Okay, so VPN-6267 is, like, totally ruining my life. Gotta modularize connectionhealth (and add tests, obviously).
But, there's this annoying dependency on leakdetector. This PR handles that, so the fun can begin.
Added tests, because I'm not a monster. Now I can actually start fixing connectionhealth without breaking everything.

XOXO, basti


src/utils/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+add_library(mz_utils INTERFACE)
+
+add_subdirectory(leakdetector)
+
+target_link_libraries(mz_utils INTERFACE mz_utils_leakdetector)
+
```
brizental, line 10: nit: Unnecessary whitespace.

src/utils/leakdetector/leakdetector.cpp
```diff
@@ -36,16 +39,18 @@ LeakDetector::~LeakDetector() {
...
+      (*m_out) << "  - ptr: " << l.key() << " size:" << l.value() << Qt::endl;
     }
   }
 
   if (!hasLeaks) {
-    out << "No leaks detected." << Qt::endl;
+    (*m_out) << "No leaks detected." << Qt::endl;
   }
+
+  (*m_out).flush();
```
brizental, line 53: Why add this change?

strseb: Given that we now can add any QIODevice as an output target to leakdetector `<<` does not necessarily would send it all out, that depends on the device impl. Imagine it's a tcp stream sending 512 bytes packages at a time and we only feed "<< 'a'", depending on the stream state it might buffer a bit until it can write a full package.  
If we call flush we make sure that everything is written out sync at this point if it has not been yet. 
I.e if we would ever pipe this to a file we could end up with just a few parts written and the rest lost in the buffer. :)

PR #9355, opened 2024-04-11
Nit: Fix Broken inspector and update workflow

strseb:
1. Remove DoxyGen -> We're not using that. 
2. Fix inspector and cleanup -> It was still using an old version of ubuntu and node:16 - let's use the conda env instead. Also needed to upgrade parcel. 
3. We do not need to run those tasks unless they have been changed (or cron hits), sooo let's save some cpu-cycles


mcleinman (approved):
👍🏻

PR #9359, opened 2024-04-12
VPN-6251: fix backend logs

mcleinman:
## Description

After a bunch of debugging, I realized that the problem went away when we didn't flush the logs.

And I remembered that I recently changed the behavior of log flushing, which would be why this is a recent bug: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9231 (It's me, hi, I'm the problem it's me)

Digging into *why* flushing logs caused the problem, it was because of a race condition between asking for backend logs and flushing the logs. We reset the `m_logCallback` after we get the daemon's response with backend logs (as we'd expect), but we also clear `m_logCallback` when asking to flush the logs. Since both requests were being made before the daemon returned with the backend logs, `m_logCallback` was null by the time we wanted to use it with the first request.

This chunk of code is in 3 spots in the class, and I don't believe it's necessary here. I can't think of any race issues that could arise from removing this chunk of code, but I'd ask reviewers to think through this and confirm. (I'm also happy to get on a call and talk through the logic together if you want.)

## Reference

VPN-6251

## Checklist
    
### DIFF
--- a/src/localsocketcontroller.cpp
+++ b/src/localsocketcontroller.cpp
@@ -194,11 +194,6 @@ void LocalSocketController::getBackendLogs(
 void LocalSocketController::cleanupBackendLogs() {
   logger.debug() << "Cleanup logs";
 
-  if (m_logCallback) {
-    m_logCallback("");
-    m_logCallback = nullptr;
-  }
-
   if (m_daemonState != eReady) {
     return;
   }


mcleinman:
> > We reset the m_logCallback after we get the daemon's response with backend logs (as we'd expect), but we also clear m_logCallback when asking to flush the logs. Since both requests were being made before the daemon returned with the backend logs, m_logCallback was null by the time we wanted to use it with the first request.
> 
> A few thoughts:
> 
>     1. It sounds like we are using a single container for the callback when we actually need some sort of queue abstractoin to hold these callbacks.
> 
>     2. It would be nice to add unit tests to guarantee this is working as expected and to explore a bit of the other edge cases empirically. Consider extracting the local socket log stuff into a separate isolated abstraction if unable to test localsocketcontroller itself.

The short version - I’m going to work on adding testing to prevent future regressions here (your suggestion #2), but I think strengthening the app-daemon communication would be a heavy lift whose payoff isn’t worthwhile in my opinion. It would require a whole bunch of code, potentially introducing more bugs (and by design, allowing for the possibility of a whole new class of bugs, even worse than this minor bug fixed here).

This PR changes m_logCallback (the variable that introduced the bug here). It is now only used in the 2 way communication between app and daemon for the logs command. I can’t figure out why it was included in the cleanLogs command, and in absence of anyone else providing a reason why it would have been, I’m chalking it up to “accidentally repeated code when creating a new command”. Thus, unless we for some reason use this variable again, the race condition here disappears.

More notes from my thinking this through:
- We can’t use tasks, as we need to run this *immediatedly* as we’re blocking on getting these logs out there. We could not block, but that wold have to be just on desktop (and even then, it’s still not a great pattern for usability), and it seems more straightforward to keep logic here similar between desktop and mobile.
- The weird race condition here was in the app layer, not the daemon layer. So we’d have to block the app layer from running additional commands until it gets a confirmation from the daemon layer that the command successfully completed.
- However, the daemon does not send a message back to the app layer for all commands. (It does for some, including the log ones that are pertinent here.) If we create some sort of list/queue on the app layer and block it on getting confirmation from the daemon layer, we’d need to re-architect the daemon layer to always return, for all commands and even in case of error in the daemon layer or the absence of the daemon layer or such - otherwise we’d be introducing the possibility of a new class of bugs where further app-issued commands are blocked due to not hearing back from the daemon. We could change commands to be a class/protocol, and then add tests that force all of them to complete. However, this would add a lot of complexity to the app-daemon communication. I’m not sure it’s worth adding all of this overhead - and potential for new bugs - because one bug popped up once. I’m personally of the opinion that we shouldn’t add heavily engineered systems to try to prevent a smaller bug that has occurred once - especially when that new system has a serious potential for introducing a new class of bugs.

mcleinman:
I've been thinking some more and examining this logging code, and I don't think writing tests here is worthwhile. I believe tests added on this ticket should have caught the bug that this ticket is about, and that is a non-trivial lift that could introduce new bugs. I propose merging this as is, and if we ever encounter a similar bug then we should take the time to refactor this logging code to be testable. 

With one non-high priority bug coming out of the current architecture, I think there is a higher risk that we introduce a bug in a refactor to make these components testable* than there is in fixing the bug and hoping we don't reintroduce a similar bug later on. Yes, I'm suggesting that making this testable is riskier than not. 

There is a lot of spaghetti code involved in writing our logs (you can see a rough trace of what happens on my notes in the associated ticket). We could refactor that to be able to test this area. Testing an individual unit would not have caught this bug, it was caused by the series of actions (including the back-and-forth between the app and the daemon) that happen after clicking for logs. Thus, this isn’t appropriate to be a unit test, and seems more appropriate for functional tests. 

The functional test would roughly be:
1. Turn on VPN
2. Go get logs
3. Confirm that logs don’t include the string “no logs” anywhere, which is what happens when the backend logs return no data.

We don’t currently have any functional test helper functions to retrieve the saved logs. In theory, this shouldn't be that difficult. However, the bug in question came from the `LogHandler::viewLogs`, which is what is called when a user requests logs. Unfortunately, it's another function or two down where the filenames are determined (which we'd need to open the log and examine the contents). We'd have to refactor this to allow the final disposition (show logs / return filename /etc.) to be controlled from `viewLogs` (which we'd call in prod and in functional tests). And even then, the filename comes as a callback, so we'd have to further refactor this area to allow the functional tests - which don't wait for callbacks, I believe - to use this refactored function. 

An alternative would be to not call `viewLogs` in the tests, and instead have another path to retrieving the logs. This would *not* have caught the bug this ticket was all about (and would still require some refactor for the callbacks) - and I think adding testing for this ticket should be specific to tests that would have caught this bug.

I would like to test this, and I'm possibly neglecting to consider a way that a test could be written without a large refactor that would have caught this bug. I'm open to feedback/suggestions.

* The tests should catch bugs, including in a refactor. But we may fail to consider a specific test case (especially with the spaghetti code that this area currently is), and that could end up allowing a bug through.

brizental (approved):
> I've been thinking some more and examining this logging code, and I don't think writing tests here is worthwhile.

I disagree. It's fine not to do the refactor and testing of this code in this PR. In that case let's just file a ticket for it for documentation purposes.

PR #9366, opened 2024-04-15
VPN-6267 Respect Errors of DNSPingsender

strseb:
## Description

In the logs we see errors, however they are currently unhandled. 
So let's assume start as failable. Also we can now abort whenever something is wierd.

### DIFF
--- a/src/connectionhealth.cpp
+++ b/src/connectionhealth.cpp
@@ -42,6 +42,10 @@ ConnectionHealth::ConnectionHealth() : m_dnsPingSender(QHostAddress()) {
 
   connect(&m_dnsPingSender, &DnsPingSender::recvPing, this,
           &ConnectionHealth::dnsPingReceived, Qt::QueuedConnection);
+  connect(&m_dnsPingSender, &DnsPingSender::criticalPingError, this, [this]() {
+    m_dnsPingInitialized = false;
+    m_dnsPingTimer.stop();
+  });
 
   connect(&m_dnsPingTimer, &QTimer::timeout, this, [this]() {
     m_dnsPingSequence++;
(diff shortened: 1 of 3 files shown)


src/dnspingsender.cpp
```diff
@@ -59,13 +59,13 @@ DnsPingSender::DnsPingSender(const QHostAddress& source, QObject* parent)
 
 DnsPingSender::~DnsPingSender() { MZ_COUNT_DTOR(DnsPingSender); }
 
-void DnsPingSender::start() {
+bool DnsPingSender::start() {
   auto state = m_socket.state();
-  if (state != QAbstractSocket::UnconnectedState) {
+  if (true) {
```
brizental, line 64: Hum, is this a mistake? Doesn't this invalidate the whole rest of this function?

src/connectionhealth.cpp
```diff
@@ -103,14 +107,13 @@ void ConnectionHealth::startIdle() {
...
-  m_dnsPingSender.sendPing(QHostAddress(PING_WELL_KNOWN_ANYCAST_DNS),
-                           m_dnsPingSequence);
-
+  if (m_dnsPingSender.start()) {
+    m_dnsPingTimer.start(PING_INTERVAL_IDLE_SEC * 1000);
+    // Send an initial ping right away.
+    m_dnsPingTimestamp = QDateTime::currentMSecsSinceEpoch();
+    m_dnsPingSender.sendPing(QHostAddress(PING_WELL_KNOWN_ANYCAST_DNS),
+                             m_dnsPingSequence);
+  }
```
brizental, line 116: Can we add unit tests for this?

brizental: Let's at least add a ticket for this... Seems a bit pointless, I know, but still.

brizental (approved):

PR #9368, opened 2024-04-15
[Refactor] Use std::chrono_literals

strseb:
## Description
We express time in the project in `uint  varname_MILLISEC` which makes some code hard to read and reason about. 
Also we're moving the "type" into the varname instead of expressing this via the typesystem. 
Like how often did we call periodic tasks again? 
```
CONSTEXPR(uint32_t, schedulePeriodicTaskTimerMsec, 3600000, 300000, 0) 
````

Given [qTimer supports std::chrono](https://doc.qt.io/qt-6/qtimer.html#start-2) we can just use said literals to express time in the typesystem instead of in the variable name :) 
i.e 

```
CONSTEXPR(auto, schedulePeriodicTask, 60min, 5min, 0ms)
```

Note this does not remove all cases, just the most wierd ones :)


brizental (approved):
Fancy!

PR #9388, opened 2024-04-17
VPN-6208 Enable Beetmover apt

strseb:
## Description

From our side we have nothing that would block us to publish to the Mozilla APT instead of the PPA. 
So this pr (fixes my failed merge) and also enables beetmover-apt.


ahal (approved):
Sorry for the really late review.. this got lost in my review queue.

This lgtm, would like Gabriel to take a quick look too though.

gabrielBusta (approved):

gabrielBusta:
Looks good to me.
Let me know if the tasks run into any issues (I suspect there might be some scopes/script-worker config issues)

PR #9394, opened 2024-04-18
Nit: Assert the branch param is really a branch

strseb:
## Description

Make sure that if you put a query that is not an existing branch , that we pick one that exitsts instead :)

### DIFF
--- a/tools/wasm_chrome/branch_selector.mjs
+++ b/tools/wasm_chrome/branch_selector.mjs
@@ -106,21 +106,26 @@ class BranchSelector extends HTMLElement {
     return data;
   }
 
-  render() {
+  async render() {
     const urlParams = new URLSearchParams(window.location.search);
-    const selectedBranch = urlParams.get("branch") || DEFAULT_SELECTED_BRANCH;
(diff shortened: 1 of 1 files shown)


tools/wasm_chrome/branch_selector.mjs
```diff
@@ -106,21 +106,26 @@ class BranchSelector extends HTMLElement {
...
   }
 
-  render() {
+  async render() {
     const urlParams = new URLSearchParams(window.location.search);
-    const selectedBranch = urlParams.get("branch") || DEFAULT_SELECTED_BRANCH;
+    let selectedBranch = urlParams.get("branch") || DEFAULT_SELECTED_BRANCH;
+    const branches = await this.getData();
+    // If the branch in the param does not exit, roll back to main :)
+    if(!branches.some(branch => branch.name == selectedBranch)){
```
brizental, line 114: ignorable nit: This could just be a `find` instead. The only difference is that it stops as soon as it finds an occurance.

brizental (approved):

PR #9416, opened 2024-04-22
No ticket: Fix QML console warnings

lesleyjanenorton:
## Description

While investigating what has ruffled our Android build's proverbial feathers in v2.21, I noticed the console complaining loudly over every appearance of an MZSettingsToggle {}. 
This PR fixes `Warning: qrc:/nebula/components/MZSettingsToggle.qml:172:17: QML MZUIStates: The current style does not support customization of this control (property: "indicator" item: MZUIStates_QMLTYPE_9(0x60000330ebc0, parent=0x0, geometry=0,0 0x0, z=-1)). Please customize a non-native style (such as Basic, Fusion, Material, etc).` and `Warning: qrc:/nebula/components/MZSettingsToggle.qml:24:5: Unable to assign [undefined] to int (MZSettingsToggle.qml:24)`. 

🪐 **UPDATE:**  After more clicking around I realized these warnings are triggered other input components as well. Second commit addresses these as well. 

Related Qt bug: https://bugreports.qt.io/browse/QTBUG-109438

 

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZCheckBox.qml
+++ b/nebula/ui/components/MZCheckBox.qml
@@ -6,6 +6,9 @@ import QtQuick 2.5
 import QtQuick.Controls 2.14
 import QtQuick.Layouts 1.14
 
+// https://bugreports.qt.io/browse/QTBUG-109438
+import QtQuick.Controls.Basic
+
 import Mozilla.Shared 1.0
 import "qrc:/nebula/utils/MZUiUtils.js" as MZUiUtils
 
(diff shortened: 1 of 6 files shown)


mcleinman (changes requested):
Thanks for this. Given that it's fixed it 6.5/6.6 according to that Qt bug, can you add something to the comments in the code about how we can remove the import once we're on a later version of Qt?

lesleyjanenorton:
@mcleinman  That is a good call out but also a really strange one because I'm using 6.6 for local mac builds and still seeing this. Will do some more digging in their bug list to see if there is any mention of a regression in 6.6, etc.

strseb:
The QTBug seems to be Fixed in Qt 6.6.0+ which we are moving to. So closing this :) - Please re-open if needed.

PR #9429, opened 2024-04-23
Reconnect VPN when rebooting or process is killed

mcleinman:
## Description

To reactivate the network extension process, we set up a rule to always use the VPN. We must remove this rule when the user disconnects the VPN via the app.

This means that auto-connect on reboot is active for iOS, with no user option to turn it off. (If the VPN is off when rebooting it will remain off when the phone reboots.)

The user can access both the VPN toggle and "Connect On Demand" in iOS system settings. If a user toggles off the VPN from system settings without toggling "Connect On Demand", the VPN will immediately reconnect. I don't believe there is any way to move the auto connect logic into the system extension (turning it off somewhere in `stopTunnel`), so I think there is no way around this. Once the PR is merged, I'll let our support colleagues know (in the rare case people write in about this).

This PR is similar to https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4897/, though this one is far simpler because there is no user option to disable this functionality. (To ensure it reconnects when the process is ended, we need to always force this to be true - hence the lack of user option.)

## Reference

VPN-6065, VPN-3113, VPN-4230

## Checklist
    
### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -403,7 +403,7 @@ int CommandUI::run(QStringList& tokens) {
       Navigator::instance()->requestDeepLink(url);
     }
     // Whenever the Client is re-opened with a new url
-    // pass that to the navigaot
+    // pass that to the navigator
     QObject::connect(
         AndroidVPNActivity::instance(), &AndroidVPNActivity::onOpenedWithUrl,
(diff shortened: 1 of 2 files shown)


oskirby (approved):
I am not much of a swift expert, but this looks pretty straightforward and I don't see any glaring issues.

PR #9441, opened 2024-04-25
VPN-6313: Disable in-app auth on desktop in Staging

lesleyjanenorton:
## Reference

VPN-6313

## Checklist
    

src/feature/featurelistcallback.h
```diff
@@ -69,6 +69,13 @@ bool FeatureCallback_captivePortal() {
 #endif
 }
 
+bool FeatureCallback_inAppAuthentication() {
+  if (Constants::inProduction() || FeatureCallback_iosOrAndroid()) {
```
strseb, line 73: Try ifdef MZ_DUMMY -> return true. 

Currently the functional tests all assume we do in app auth. we would need to roll back those patches 😅

mcleinman (approved):
👍🏻

strseb (approved):

PR #9443, opened 2024-04-25
VPN-6340 Remove Multithreading Footgun

strseb:
## Description

`runOnAndroidThreadSync` was a thing in Qt5, so obviously when upgrading we just replaced it with `AndroidCommons::runOnAndroidThreadSync`. 

This obviously can lead to deadlocks, i.e if we're waiting to run a task on the main thread but i.e an Accessbility Service is trying to read the screen info from the Qt Thread. 
None of the callsites actually require the code to be run in sync, so it's better for the callsites to use the QFuture returned and decide what to do :)

### DIFF
--- a/src/platforms/android/androidvpnactivity.cpp
+++ b/src/platforms/android/androidvpnactivity.cpp
@@ -28,7 +28,7 @@ Logger logger("AndroidVPNActivity");
 }  // namespace
 
 AndroidVPNActivity::AndroidVPNActivity() {
-  AndroidCommons::runOnAndroidThreadSync([]() {
+  QNativeInterface::QAndroidApplication::runOnAndroidMainThread([]() {
     // Hook in the native implementation for startActivityForResult into the JNI
     JNINativeMethod methods[]{
         {"handleBackButton", "()Z", reinterpret_cast<bool*>(handleBackButton)},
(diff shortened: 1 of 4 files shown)


src/platforms/android/androidvpnactivity.cpp
```diff
@@ -188,10 +188,11 @@ void AndroidVPNActivity::onAppStateChange() {
   // those screens.
   auto state = MozillaVPN::instance()->state();
   bool isSensitive = state == App::StateAuthenticating;
-  AndroidCommons::runOnAndroidThreadSync([isSensitive]() {
-    QJniObject::callStaticMethod<void>(CLASSNAME, "setScreenSensitivity",
-                                       "(Z)V", isSensitive);
-  });
+  QNativeInterface::QAndroidApplication::runOnAndroidMainThread(
```
strseb, line 191: This i.e is a guaranteed deadlock when pressing "register" with talkback on :)

src/platforms/android/androidvpnactivity.cpp
```diff
@@ -188,10 +188,11 @@ void AndroidVPNActivity::onAppStateChange() {
   // those screens.
   auto state = MozillaVPN::instance()->state();
   bool isSensitive = state == App::StateAuthenticating;
-  AndroidCommons::runOnAndroidThreadSync([isSensitive]() {
-    QJniObject::callStaticMethod<void>(CLASSNAME, "setScreenSensitivity",
-                                       "(Z)V", isSensitive);
-  });
```
strseb, line 194: This i.e is a guaranteed deadlock when pressing "register" with talkback on :)

mcleinman (approved):
Makes sense. Great catch!

brizental (approved):

PR #5225, opened 2022-12-15
Remove python lxml dependency

oskirby:
## Description
I have noticed in my travels that we are encountering some build friction related to the `lxml` python module. On MacOS this sometimes causes pip to take an extremely long time to resolve dependencies and I have encountered the occasional build error on Windows for similar reasons.

This module is extremely close in functionality to Python's built-in `xml.etree` module, so let's just use that instead.

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #9448, opened 2024-04-26
VPN-5175 - Get languages, currencies and server name translations from the l10n repository strings

brizental:
Ok, this is a large one. 

I considered separating each commit into a separate PR, but 🤷‍♀️ Let me know if you'd prefer I do that.

Regardless, I recomend reviewing this per commit. Each commit handles a different part of the task.

We had the `languages.json` and `servers.json` files that included translations for: language name, currency symbols and server names. This PR moves away from that to actually use translations from Pontoon for these.

On previous PRs (https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9411, https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/435) I have added the `extras.xliff` file with language and server name strings to take place of the JSON files.

#### 1. Currency symbols (221aa63f5beb4977950b53723d2f22c2981e371b)

After discussions with @flodolo and Santiago, we decided not to have translations for these anymore. It seems pointless, because all translations are the same. 

On the first commit of this pull request, I have update localizer.cpp to either grap currency symbol translations from Qt translations or from a hardcoded list of currencies that we support.

#### 2. Using extras.xliff (3afc2ed39d2aff8c7cdb6d6994e70da6f7670763)

On this commit I updated the build to include strings from  extas.xliff into i18nstrings_p.cpp and friends.

#### 3. Laguage names (721effb30173b24fab9cf53462338026a83086ab)

On this commit I remove code that generates and consumes languages.json and add code to get language names from I18nStrings. I also had to include some new code generation in order to have a list of language names in all supported locales, the alternative to that would be to load a QTranslator just to grab a string than unload it -- not good. That means there is also some build changes to generate the new string and well as to update the I18nStrings class to be able to get strings by id.

#### 4. Server names (d6830e1f6126a47ca91820057670416511b825a3)

Finally, this commit adds logic to get city and country names. All of the infrastrcuture that makes this possible was done already in the previous commits so this is mechanical work.

I also removed the code that would get the server names form servers.json.


scripts/utils/generate_strings.py
```diff
@@ -147,6 +171,20 @@ class I18nStrings final : public QQmlPropertyMap {
             """    __Last,
   };
 
+  static String getString(const QString& s) {
+  """
+        )
+
+        for i, key in enumerate(strings):
+            output.write(f"    {'if' if i == 0 else '} else if '} (s == \"{key}\") {{\n")
```
flodolo, line 179:
```suggestion
            output.write(f"    {'if' if i == 0 else '  } else if '} (s == \"{key}\") {{\n")
```

scripts/utils/generate_strings.py
```diff
@@ -147,6 +171,20 @@ class I18nStrings final : public QQmlPropertyMap {
...
+            output.write(f"        return {key};\n")
+
+        output.write(
+            """    } else {
```
flodolo, line 183:
```suggestion
            """      } else {
```

scripts/utils/generate_strings.py
```diff
@@ -147,6 +171,20 @@ class I18nStrings final : public QQmlPropertyMap {
...
+          return Empty;
+        }
```
flodolo, line 185:
```suggestion
      }
```

Alignment seems to be slightly off when looking at the .h file generated

scripts/utils/generate_strings.py
```diff
@@ -238,18 +276,25 @@ def serialize(string):
...
 
-    if not args.source:
+    if not args.sources:
         exit("No source argument.")
 
     # If no output directory was provided, use the current directory.
     if args.output is None:
         args.output = os.getcwd()
 
-    # Parse the inputs for their sweet juicy strings.
```
flodolo, line 248: 🙄

flodolo:
Looks like GitHub ate one of my comment. Don't know much about C++, but is there a switch/case structure to use instead of so many "if else"?

scripts/utils/generate_strings.py
brizental: Heh, I was hoping I could just quietly remove this.

scripts/utils/generate_language_names_map.py
```diff
@@ -0,0 +1,60 @@
...
+            root = tree.getroot()
+
+            for trans_unit in root.findall('.//{urn:oasis:names:tc:xliff:document:1.2}trans-unit'):
+                unit_id = trans_unit.get('id')
+                if unit_id.startswith('languages.'):
+                    if lang_code == 'en':
+                        value = trans_unit.find('.//{urn:oasis:names:tc:xliff:document:1.2}source')
+                    else:
+                        value = trans_unit.find('.//{urn:oasis:names:tc:xliff:document:1.2}target')
+                    if value and value.text is not None:
```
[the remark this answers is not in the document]
brizental: Yeap I found this as well and fixed it already. Turns out Python is not Javascript.

scripts/utils/generate_strings.py
```diff
@@ -36,7 +36,31 @@ def construct_mapping(self, node, deep=False):
...
+
+    strings = {}
+
+    ns = {'x':'urn:oasis:names:tc:xliff:document:1.2'}
+    tree = etree.parse(xliff_file)
+    root = tree.getroot()
+
+    for node in root.xpath('//x:trans-unit', namespaces=ns):
+        id = node.get('id')
+        cpp_id = pascalize(id.replace('.', '_'))
```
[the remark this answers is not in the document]
brizental: Oh :eyes: I am pretty sure this is not fine. Good call! Will fix.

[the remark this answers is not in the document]
brizental: Well, hum. I wonder actually. Because this is building just fine... VSCode color codes the special character variant name differently, but maybe VSCode is just confused. I will check C++ specs before I fix this.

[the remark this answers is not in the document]
brizental: The spec seems to leave it up to the compiler... I will fix it :/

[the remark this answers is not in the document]
brizental: This is being extracted from the XLIFF id attribute. I am assuming in there it is not an issue, but let me know.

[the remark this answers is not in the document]
brizental: Yeah, exactly. My question is if it an issue that the XLIFF file has ids with unicode characters i.e. should I fix that in the l10n repo as well.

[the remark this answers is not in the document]
flodolo: No, that's not an issue for our side.

src/translations/extras/extras.xliff
```diff
@@ -162,7 +162,7 @@
       <trans-unit id="servers.br" xml:space="preserve">
         <source>Brazil</source>
       </trans-unit>
-      <trans-unit id="servers.SaoPaulo" xml:space="preserve">
+      <trans-unit id="servers.SãoPaulo" xml:space="preserve">
```
flodolo, line 165: Is this strictly necessary? This file is translated, this string will show up as new.

brizental: Argh! Great catch... I just wanted to change this in the test extras.xliff files 🙄 
```suggestion
      <trans-unit id="servers.SaoPaulo" xml:space="preserve">
```

oskirby:
Regarding the Linux build failures. The package sources for RPM-based distributions maintain their dependency information in `linux/mozillavpn.spec` instead of `linux/debian/control`. That fill will need to be updated as well if we need to add `lxml` as a dependency.

However! Unless we need to use `lxml` specifically, for XML parsing we have generally gotten away with `
import xml.etree.ElementTree as etree` as a suitable replacement. It's a got the same basic API and nearly the same functionality but it's a python built-in and generally does the job just as good as `lxml` without needing to bring in extra dependencies.

oskirby:
Also: The last time we made use of `lxml` in our build tooling we encountered some friction when resolving python deps (see: #5225) we might want to keep an eye out on that to make sure we don't cause a regression.

brizental:
> Looks like GitHub ate one of my comment. Don't know much about C++, but is there a switch/case structure to use instead of so many "if else"?

Plot twist: the Windows C++ compiler spits out a fatal error when there are more than 128 nesting blocks 🅰️ 🅰️ 🅰️ 🅰️ Going to need to rethink this.

brizental:
> Also: The last time we made use of lxml in our build tooling we encountered some friction when resolving python deps (see: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/5225) we might want to keep an eye out on that to make sure we don't cause a regression.

No worries, I just removed that dependency.

scripts/utils/generate_strings.py
```diff
@@ -36,7 +37,31 @@ def construct_mapping(self, node, deep=False):
...
+    if not os.path.isfile(xliff_file):
+        exit(f"Unable to find {xliff_file}")
+
+    strings = {}
+
+    tree = etree.parse(xliff_file)
+    root = tree.getroot()
+
+    for node in root.findall('.//{urn:oasis:names:tc:xliff:document:1.2}trans-unit'):
+        # Remove any unexpected characters e.g. São Paulo -> SoPaulo
```
flodolo, line 50:
```suggestion
        # Keep only ASCII alphabetic characters e.g. São Paulo -> SoPaulo
```

How does it work if we need to update a string with a new ID, e.g. for "cs" Czech Republic -> Czechia?

brizental: The code is not aware of updates. As far as it is concerned, that would be a new string. IIUC that is also how Pontoon will treat a new string id right?

flodolo: Let's say you update the string and use a new ID in `extras.xliff`, e.g. `servers.cz` ⇾ `servers.cz2`, but then you'd need some mapping to still store the content of `servers.cz2` as `ServersCz`.

I suppose with this regex it would work out of the box, since numbers are removed, but it feels like an unexpected side effect than a planned behavior.

brizental: Ok, it has taken me a while, but I think I see what you mean. 

If we need to change any of the root translations we can't just change the values is the `<source>` tags of the `extras.xliff` file, we actually need to create a new string id! 

This is a great call, but something we can handle when working on https://mozilla-hub.atlassian.net/browse/VPN-6350. The `extras.xliff` file is supposed to be automatically updated as the Mullvad server list is changed, we can add logic to create the mapping and so on in there. WDYT?

flodolo: That works for me. I just wanted to raise it now, so we don't paint ourselves into a corner.

flodolo (approved):
Python scripts look good to me 👍🏼

scripts/utils/generate_strings.py
brizental: Perfect, I'll add a note about it in the ticket description so I don't forget either.

lesleyjanenorton (approved):
👍

scripts/utils/generate_language_names_map.py
```diff
@@ -0,0 +1,61 @@
+#! /usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import argparse
+import xml.etree.ElementTree as etree
+
+def extract_strings_with_id(directory):
```
brizental, line 10: TODO: We don't need the map to include ALL language name translations, just the native translation.

brizental:
Test failure was unrelated and being worked on in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9499. Therefore I force merged.

Issue #9174, opened 2024-02-25
Killswitch Failure - Issue #8826 Remains prevalent after update to v2.20


Bobinsz:
Please see Issue #8826 

After updating to 2.20.0 (2.202401231837), this issue remains a problem. When the popup occurs stating "Network failure" choose a new location, the VPN releases all control of the network connection and all traffic defaults back to unprotected traffic.

data-sync-user:
➤ Santiago Andrigo commented:

Adding 3 points as the default for any bug

Bobinsz:
I'm glad to see that this issue has been escalated and has now been assigned to someone from Mozilla, who appears to be a specialist "Code Reviewer".

This may or may-not be helpful, but I felt I didn't include as much detail in my follow-up bug report as I should have, relying mainly on the referral to #8826 

Please be advised that going from: v2.19.0 (2.202311270624) to v2.20.0 (2.202401231837), there has been a noticeable reduction in the frequency of the failure popup "Network Failure - Choose a New Location" 

But the underlying issue is still there, when that popup occurs, the VPN releases all control of the network connection and all traffic defaults back to unprotected traffic, as if no VPN was being used & without any user interaction.

I hope this additional information is useful

Many Thanks

Bobs

Bobinsz:
My confidence in the VPN has risen, after reading through the steps coders and code reviwers have been taking... with Logical reasoning and what seems to be a thorough look at what was causing the Kill-Switch to deactivate and allow the net connection to revert to unprotected traffic, without user interaction.

I've continued to have the occational complete discconnection, with the pop-up server unavaiable whilst running 2.22.1 (2.202405141843) )

If i'm honest, as a user I'm actually more frustrated by the connection timer changes that have occured over the past few updates... The "Secure & Private" timer no longer resets upon a reconnection attempt... and seems to just show the Total time the VPN has been active for.

I've had unstable traffic whilst using the VPN, without the Kill-Switch failure that I initially raised... but at a simple glance, there isn't an easy way to see if the VPN is constantly trying to establish a stable connection with this new "Secure & Private" timer behaviour.

Bobinsz:
P.S When I suffer from an unstable connection... my actual net connection is 100% fine... It's the VPN connections that are unstable for some reason. Deactivating the VPN results in perfect net traffic, but its not how I want my net traffic to be, open to the world... Hence my signing upto the Mozilla VPN, virtually at launch, with continuity in sub until today!

Issue #8826, opened 2023-12-17
KillSwitch Failure - Network Released without user interaction


Bobinsz:
**Describe the bug**
When a Network connection fails, the KillSwitch works correctly stopping all network traffic, displaying the error: "No Signal - Check Connection"

However if the VPN fails to re-establish a connection to the server choice of the user within a set frame of time the VPN provides the following popup warning: "Server Unavailable - Choose a new Location" & without any user interaction to un-secure the connection, the VPN app completely releases control of the network connections... the KillSwitch no longer functions and all traffic defaults back to un-protected channels, as if the user was not using the VPN in the first place!

**VPN version:**
v2.19.0 (2.202311270624)

**Affected Platforms:**
Edition	Windows 10 Pro
Version	22H2
Installed on	‎25/‎03/‎2023
OS build	19045.3693
Experience	Windows Feature Experience Pack 1000.19053.1000.0

**Prerequisites:**
  <Optional: Any special steps taken not related to the app to trigger the bug>

**Steps to reproduce:**
Network Failure, showing error: ""No Signal - Check Connection"
With continued failures to reconnect resulting in the Error popup: "Server Unavailable - Choose a new Location"

**Expected result:**
The KillSwitch works correctly for the initial Network failure. However the VPN completely releases control, without user interaction should the subsequent failure occur ("Server Unavailable - Choose a new Location". allowing unprotected network traffic because of the failure. 

The KillSwitch & VPN should Not release the Network to unprotected traffic, under any circumstance, even with the connection failure popup... without user interaction. There should be no defaulting back to unprotected traffic without user interaction!

**Actual result:**
Once the "Server Unavailable - Choose a new Location" popup occurs - after reconnection attempts have failed - the VPN and KillSwitch stop functioning completely and release all Network traffic back to Unprotected Network traffic as if the VPN wasn't being used in the first place. Without any user interaction.

**Additional notes:**
  <Optional: Any Screenshots, Videos, Logs>

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-6033)

data-sync-user:
➤ Gela Malek Pour commented:

Lesley Norton mentioned that the deactivation is likely happening due to this function:

void Controller::serverUnavailable() {
  logger.error() << "server unavailable";

  m_nextStep = ServerUnavailable;

  if (m_state == StateOn || m_state == StateSwitching ||
[...]

PR #9460, opened 2024-04-29
VPN-6228: Don't deactivate during silent server switches

lesleyjanenorton:
## Description

In #8826 and #9174 a subscriber reports occasionally getting erroneously disconnected from the VPN (through no action of their own) and seeing the 'Server Unavailable' modal upon reopening the client. 

We hypothesize this is caused by the subscriber's connection becoming `Unstable`, whereupon we go into a silent server switch which fails sending us into the server unavailable routine and eventual disconnection. 

#8978 attempted to resolve this by adding logic to the server unavailable routine to check if the state is `StateSilentSwitching` or `StateOn` and bypassing the call the deactivate if so. This treatment was insufficient because, by the time we get to `ServerUnavailable`, the user is either in `StateConfirming` or `StateConnecting`. 

In this PR we remove the call to deactivate from the server unavailable routine entirely. This is in keeping with behavior outlined in this [PRD](https://mozilla-hub.atlassian.net/wiki/spaces/PXI/pages/130809909/Improve+our+handling+of+network+connectivity+issues) reiterating that we should never tear down the tunnel for any reason other than the subscriber has specifically clicked a button, or taken some action, to knowingly do so. 

Changed behavior:
- If we fail to connect, or reconnect after a silent server switch, we show the `ServerUnavailable` modal but do not deactivate the VPN. We stay in `NoSignal` and continue attempting to reestablish the connection.
- If the connection eventually succeeds and we move out of `NoSignal`, the `ServerUnavailable` modal will be hidden. 

There is a drawback: We could show the server unavailable modal when the failure is actually caused by something else (a lack of internet for example). But this is actually already the case. This PR is meant as a stop-gap ahead of future work to untangle the Controller and improve connection failure communication to the subscriber ([PRD](https://mozilla-hub.atlassian.net/wiki/spaces/PXI/pages/130809909/Improve+our+handling+of+network+connectivity+issues)).

VPN-6228

## Checklist
    
### DIFF
--- a/src/connectionhealth.cpp
+++ b/src/connectionhealth.cpp
@@ -93,7 +93,6 @@ void ConnectionHealth::stop() {
   m_dnsPingTimer.stop();
 
   stopTimingDistributionMetric(m_stability);
-  setStability(Stable);
 }
 
 void ConnectionHealth::startActive(const QString& serverIpv4Gateway,
@@ -343,8 +342,6 @@ void ConnectionHealth::stopTimingDistributionMetric(
       mozilla::glean::connection_health::stable_time.stopAndAccumulate(
           m_metricsTimerId);
   }
-  m_metricsTimerId = -1;  // used as a signal to prevent turning it off twice
-                          // when ConnectionHealth moves between idle and stop.
 #endif
 }
 
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -297,21 +297,18 @@ void Controller::serverUnavailable() {
   // to deactivate the VPN in the case of server location becoming unavailable
   // because the user may not notice that they are no longer protected which can
   // cause the traffic to leak outside the tunnel without their knowledge.
-  if (m_state == StateOn || m_state == StateOnPartial) {
+  if (m_state == StateOn || m_state == StateSilentSwitching || m_state == StateOnPartial) {
+    logger.info() << "Server location is unavailable in StateOn or "
+                     "StateSilentSwitching. Do not deactivate.";
     return;
   }
+
+  logger.info() << "Server location is unavailable and we are not in StateOn "
+                   "or StateSilentSwitching. Deactivate.";
 
   m_nextStep = ServerUnavailable;
-
-  if (m_state == StateSwitching || m_state == StateConnecting ||
-      m_state == StateConfirming) {
-    logger.info() << "Server location is unavailable and we are not in "
-                     "StateOn. Deactivate!";
-    deactivate();
-    return;
-  }
-  logger.info() << "Server location is unavailable. Do not deactivate the VPN "
-                   "automatically.";
+  deactivate();
+  return;
 }
 
 void Controller::updateRequired() {
@@ -617,8 +614,11 @@ void Controller::activateNext() {
   if (m_initiator == ExtensionUser) {
     return;
   }
-  // Move to the confirming state if we are awaiting any connection handshakes.
-  setState(StateConfirming);
+
+  if (m_state != StateSilentSwitching) {
+    // Move to the StateConfirming if we are awaiting any connection handshakes
+    setState(StateConfirming);
+  }
 }
 
 void Controller::setState(State state) {
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -290,24 +290,9 @@ void Controller::handshakeTimeout() {
 }
 
 void Controller::serverUnavailable() {
-  logger.error() << "server unavailable";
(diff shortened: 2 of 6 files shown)


src/connectionhealth.cpp
```diff
@@ -93,7 +93,6 @@ void ConnectionHealth::stop() {
   m_dnsPingTimer.stop();
 
   stopTimingDistributionMetric(m_stability);
-  setStability(Stable);
```
mcleinman, line 96: Why do we need to change this file as part of this?

src/controller.cpp
```diff
@@ -290,28 +290,10 @@ void Controller::handshakeTimeout() {
...
-  // If VPN is already active and server location becomes unavailable we do not
-  // deactivate the VPN or proceed with next steps. We specifically do not want
-  // to deactivate the VPN in the case of server location becoming unavailable
-  // because the user may not notice that they are no longer protected which can
-  // cause the traffic to leak outside the tunnel without their knowledge.
-  if (m_state == StateOn || m_state == StateOnPartial) {
-    return;
-  }
-
-  m_nextStep = ServerUnavailable;
```
mcleinman, line 304: Can we can remove `ServerUnavailable` in `controller.h` (around line 193)?

src/controller.cpp
```diff
@@ -617,8 +599,11 @@ void Controller::activateNext() {
   if (m_initiator == ExtensionUser) {
     return;
   }
-  // Move to the confirming state if we are awaiting any connection handshakes.
-  setState(StateConfirming);
+
+  if (m_state != StateSilentSwitching) {
```
mcleinman, line 603: This makes sense generally, but can you help me understand what would have happened if we didn't wrap this in the if statement?

mcleinman (commented):
This looks good. A few questions on some of the code to make sure I'm understanding what happens here.

Additionally, will this change result in *always* showing the server unavailable modal if there is no internet? Or just in certain conditions?

src/connectionhealth.cpp
lesleyjanenorton: Without this we won't detect the move from NoSignal back to Stable and know to hide the ServerUnavailable modal.

lesleyjanenorton:
> Additionally, will this change result in always showing the server unavailable modal if there is no internet? Or just in certain conditions?

This is actually already the case on desktop. If a connection goes south for any reason and we can't reconnect, we'll currently disconnect (after exhausting available servers) and show this modal.

src/controller.cpp
lesleyjanenorton: Sure, if we change the state here during a silent server switch we will see the 'switching' UI on the front-end and potentially cause confusion for the subscriber.

src/controller.cpp
lesleyjanenorton: ... I will confirm and report back. I think so.

lesleyjanenorton: removed

mcleinman:
> > Additionally, will this change result in always showing the server unavailable modal if there is no internet? Or just in certain conditions?
> 
> This is actually already the case on desktop. If a connection goes south for any reason and we can't reconnect, we'll currently disconnect (after exhausting available servers) and show this modal.

And this won't happen quicker with this PR, right? It still tries to go through server switches first? ("dead" servers are grayed out in the location list, but that seems okay.)

lesleyjanenorton:
> And this won't happen quicker with this PR, right? It still tries to go through server switches first? ("dead" servers are grayed out in the location list, but that seems okay.)

right! The flow up until server unavailable is essentially unchanged. The time to get to that modal depends on how many other servers there are in that location and this doesn't change that. Dead servers are greyed out in the location list, but that is also currently the case.

mcleinman (approved):
Thanks for talking through some of this with me!

Bobinsz:
My confidence in the VPN has risen, after reading through the steps coders and code reviwers have been taking... with Logical reasoning and what seems to be a thorough look at what was causing the Kill-Switch to deactivate and allow the net connection to revert to unprotected traffic, without user interaction.

I've continued to have the occational complete discconnection, with the pop-up server unavaiable whilst running 2.22.1 (2.202405141843) )

If i'm honest, as a user I'm actually more frustrated by the connection timer changes that have occured over the past few updates... The "Secure & Private" timer no longer resets upon a reconnection attempt... and seems to just show the Total time the VPN has been active for.

I've had unstable traffic whilst using the VPN, without the Kill-Switch failure that I initially raised... but at a simple glance, there isn't an easy way to see if the VPN is constantly trying to establish a stable connection with this new "Secure & Private" timer behaviour.

PR #9462, opened 2024-04-29
VPN-6356: Fix assert in QTest::mouseEvent()

oskirby:
## Description
I have been seeing a bunch of auth functional tests failing in CI due to an assert in `QTest::mouseEvent()` due to a null `QWindow` pointer. After some digging, I think this is really a consequence of the bug noted by [VPN-3697](https://mozilla-hub.atlassian.net/browse/VPN-3697), where we encounter a race condition between the rendering thread and the GUI thread.

To resolve this, we should poke the rendering thread and wait for the frame draw to finish before submitting the click to the window. This should ensure that the QML scene graph is synchronized with the QML content before we process the click event.

Some subtle changes that revealed some bugs:
- Sending the click to `item->window()` can actually click things that are offscreen, and sending the click to the root window revealed a handful of test bugs.
- There were a lot of test bugs that were being hidden by the `QTest::qWait()` workaround.

## Reference
JIRA issue: [VPN-3697](https://mozilla-hub.atlassian.net/browse/VPN-3697)
JIRA issue: [VPN-6356](https://mozilla-hub.atlassian.net/browse/VPN-6356)

## Checklist
    
[VPN-3697]: https://mozilla-hub.atlassian.net/browse/VPN-3697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-3697]: https://mozilla-hub.atlassian.net/browse/VPN-3697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPN-6356]: https://mozilla-hub.atlassian.net/browse/VPN-6356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -358,17 +358,18 @@ static QList<InspectorCommand> s_commands{
                      [](InspectorHandler*, const QList<QByteArray>& arguments) {
                        QJsonObject obj;
 
+#ifndef MZ_WASM
                        // Ensure window rendering is finished.
                        QQuickWindow* window = qobject_cast<QQuickWindow*>(
-                          QmlEngineHolder::instance()->window());
+                           QmlEngineHolder::instance()->window());
                        QSignalSpy spy(window, &QQuickWindow::afterFrameEnd);
                        window->update();
                        if (!spy.isValid()) {
                          logger.debug() << "QSignalSpy is invalid";
+                       } else if (!spy.wait(150)) {
+                         logger.debug() << "Never got an afterFrameEnd signal";
                        }
-                       if (!spy.wait(150)) {
-                        logger.debug() << "Never got an afterFrameEnd signal";
-                       }
+#endif
 
                        QObject* qmlobj =
                            InspectorUtils::queryObject(arguments[1]);
--- a/src/inspector/inspectorhandler.cpp
+++ b/src/inspector/inspectorhandler.cpp
@@ -391,71 +391,70 @@ static QList<InspectorCommand> s_commands{
                        return obj;
                      }},
 
-    InspectorCommand{"scrollview", "Scroll a view until an item is centered", 2,
-                     [](InspectorHandler*, const QList<QByteArray>& arguments) {
-                       QJsonObject result;
+    InspectorCommand{
+        "scrollview", "Scroll a view until an item is centered", 2,
+        [](InspectorHandler*, const QList<QByteArray>& arguments) {
+          QJsonObject result;
 
-                       QObject* viewobj =
-                           InspectorUtils::queryObject(arguments[1]);
-                       if (!viewobj) {
-                         logger.error() << "Did not find view object";
-                         result["error"] = "Object not found";
-                         return result;
-                       }
-                       QQuickItem* view = qobject_cast<QQuickItem*>(viewobj);
-                       if (!view) {
-                         logger.error() << "View is not a QQuickItem object";
-                         result["error"] = "Object not found";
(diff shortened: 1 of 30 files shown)


src/inspector/inspectorhandler.cpp
```diff
@@ -357,6 +358,15 @@ static QList<InspectorCommand> s_commands{
                      [](InspectorHandler*, const QList<QByteArray>& arguments) {
                        QJsonObject obj;
 
+                       // Ensure window rendering is finished.
+                       QQuickWindow* window = qobject_cast<QQuickWindow*>(
+                          QmlEngineHolder::instance()->window());
+                       QSignalSpy spy(window, SIGNAL(afterRendering()));
+                       window->update();
+                       if (!spy.wait(150)) {
+                        logger.debug() << "Never got an afterRendering signal";
```
brizental: nit: logger.warn

src/inspector/inspectorhandler.cpp
```diff
@@ -357,6 +358,19 @@ static QList<InspectorCommand> s_commands{
                      [](InspectorHandler*, const QList<QByteArray>& arguments) {
                        QJsonObject obj;
 
+                       QQuickWindow* window = qobject_cast<QQuickWindow*>(
+                           QmlEngineHolder::instance()->window());
+#ifndef MZ_WASM
```
brizental, line 363: nit: Please add a comment here on why WASM is being skipped.

src/ui/screens/home/servers/ServerCountry.qml
```diff
@@ -22,7 +22,7 @@ MZClickableRow {
     property var currentCityIndex
     property alias serverCountryName: countryName.text
     property var cityList: cityListVisible ? cityLoader.item : cityLoader
-    property var busy: cityListVisible ? scrollAnimation.running : false
+    property var busy: cityListVisible && (scrollAnimation.running || listOpenTransition.running || listClosedTransition.running)
```
brizental, line 25: nit: Does something like this work on QML?

```suggestion
    property var busy: cityListVisible && transitions.find(t => t.running)
```

brizental (approved):

src/ui/screens/home/servers/ServerCountry.qml
oskirby: This construction does not seem to work in QML. It generates an error:
`qrc:/ui/screens/home/servers/ServerCountry.qml:25: TypeError: Property 'find' of object [object Object] is not a function (ServerCountry.qml:25)`

src/inspector/inspectorhandler.cpp
oskirby: The short answer is that this results in a crash on WASM, and this workaround only seems to be needed on systems that use multithreaded rendering engines (of which WASM is not). I will add some more comments and make the intent a little clearer.

strseb (approved):

src/ui/screens/home/servers/ServerCountry.qml
brizental: Looks like translations is not an array although it looks like one. 🤷‍♀️

PR #9465, opened 2024-04-30
Nit: Only try to unbound the Daemon if it exists

strseb:
## Description
In case we kill the activity before the serviceconnection is valid, for example when writing speedy integration tests, 
this will cause a crash :)


brizental (approved):

mcleinman (approved):
👍🏻

PR #9469, opened 2024-04-30
VPN-5266: Update timer

mcleinman:
## Description

This streamlines the timer to have the same behavior across all platforms: Server switches (silent from app, silent from daemon [only available on mobile], or user-generated) do *not* reset the timer.

Some notes:
- Added a new debug option to initiate a silent server switch from the app.
- The Controller is now fully in charger of the timer, and Android and iOS give the start time only when the app boots (if the daemon is currently running.)
- We use `isSwitchingServer` (renamed from `shouldSkipSessionTelemetry`, but it has the same meaning) to decide whether to reset the timer or not.
- Similar logic was done iOS only as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/6527, and so that work is backed out here.
- This PR fixes a bug in the prior implementation - If a user switched servers during a given session, the ending telemetry (for the app, not daemon) was never recorded.
- There is one known bug in this implementation, and it will be fixed under VPN-2367. (It exists in main already, so not fixing as part of this.) The bug: If an iOS VPN session changes servers, and then the app is restarted, the app will report the timer from the "change server" time, not the "initial connection" time.

I've tested this extensively to ensure telemetry and timer works as expected, no matter whether the VPN was started or stopped from the app, tile, system settings, onboarding, etc. The only platforms I'm not able to test are Windows and Linux. Given that macOS and Windows have identical code, I'm not worried about Windows, as I tested on macOS. @oskirby could you confirm that the timer does not reset on Linux (and we don't send telemetry) when switching servers, as well as silent switches (which now have that new debug option).

## Reference

VPN-5266

## Checklist
    

strseb (changes requested):
This should get a functional test, given we can turn on/off/switch :)

tests/functional/testServers.js
```diff
@@ -152,6 +152,12 @@ describe('Server', function() {
             'VPN is on';
       });
 
+      await vpn.waitForCondition(async () => {
+        return await vpn.getQueryProperty(
+                   queries.screenHome.CONNECTION_TIMER.visible(), 'text') ===
+            '00:00:00';
```
strseb, line 158: Given this is running on 2 Different processes via networking this can be flaky (given we don't know when this will be handled by the client). 
How about we test that it only increments ? i.e 
``` Turn on
 (keep the other checks, i.e name, location  so it's unlikely to be 0)
 let oldTimer = checkTimer()
 Switch servers 
 assert( oldTimer <  checkTimer() ) # We now know it did not reset. 
 
```

strseb (approved):
Thanks! :)

oskirby:
Tested and confirmed on Linux - a silent server switch no longer resets the timer.

oskirby (approved):
LGTM!

tests/functional/testServers.js
mcleinman: It's definitely not guaranteed to be instantaneous, but as long as it hits this within the first full second this should be true. If this test ends up being flaky on the server, I'm definitely happy to update it.

PR #9473, opened 2024-05-02
No ticket - Set Glean log level to INFO

brizental:


mcleinman (changes requested):
Is it possible to fork this logic based on whether it's a debug build or not? Would be nice to set it to debug on VPN debug build.

brizental:
> Is it possible to fork this logic based on whether it's a debug build or not? Would be nice to set it to debug on VPN debug build.

Sure, but I don't think this is neessary. This is only setting the log level for Glean's logs. We shouldn't have a need for Glean specific logs most of the time... Note that we also don't have debug logs enabled for other libraries used by the application, not even for Qt.

mcleinman (approved):
That's a good point about external libraries, I'm in agreement. Thanks.

PR #9475, opened 2024-05-02
VPN-6344 Add missing permission

strseb:
## Description

> Declare new permission to use foreground service types
If apps that target Android 14 use a foreground service, they must declare a specific permission, based on the foreground service type, that Android 14 introduces. 

https://developer.android.com/about/versions/14/changes/fgs-types-required#permission-for-fgs-type

We did add a type here 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/android/daemon/src/main/AndroidManifest.xml#L9 
We did not add the permission, so the connection fails


brizental (approved):
Oops!

PR #9476, opened 2024-05-02
Nit, try to fix source-test-android

strseb:
## Description

I really should fix that upstream 😅 
The conda-rust-package set's `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER` to a nonexistent path in the env which makes cargo fail to find the linker. Therefore compilation of the crate fails -> source_test cannot continue.


brizental (approved):
> I really should fix that upstream 😅

👍

PR #9484, opened 2024-05-03
VPN-5969: Update settings screen

mcleinman:
## Description

Updates Settings screens per written details and mocks on ticket. 

I've kept the string ids the same as we'd rather keep the existing translations than have no translation for a bit - and let them slowly get adjusted as needed.

(I touched some test files, and the formatting got patched up - there aren't significant changes on most of those lines that just lose some whitespace.)

## Reference

VPN-5969

## Checklist
    

addons/message_upgrade_to_annual_plan/manifest.json
```diff
@@ -49,7 +49,7 @@
         "id": "c_4",
         "type": "button",
         "style": "link",
-        "content": "Get help",
+        "content": "Help",
```
flodolo, line 52: This needs a new ID.

flodolo (approved):
Every string change requires a new ID (addons, strings.yaml)

flodolo (changes requested):
(wrong button 🙄)

addons/message_upgrade_to_annual_plan/manifest.json
mcleinman: Fixed - changed everything to having new IDs

lesleyjanenorton (approved):
Looks great!

src/ui/screens/devices/ViewDeviceLimit.qml
```diff
@@ -19,15 +19,15 @@ MZViewBase {
...
             onClicked: {
                 Glean.sample.getHelpClickedInitialize.record();
                 MZNavigator.requestScreen(VPN.ScreenGetHelp);
             }
         }
     }
 
-    //% "My devices"
+    //% "Devices"
     _menuTitle: qsTrId("vpn.devices.myDevices")
```
flodolo, line 31: Realized only today that this and another string were changed without a new ID, so localization will reflect the English text only from the next version of VPN 😞

PR #9492, opened 2024-05-06
Nit: Fixup Licence MD and add browserstack notice

strseb:
## Description
Browserstack has an open-source project we could apply: 
https://www.browserstack.com/open-source

That should give us enough runtime (for freeeeee)  to test some real things on android - so i just tried to sign us up and got this  automated response: 

> Received project URL: https://github.com/mozilla-mobile/mozilla-vpn-client. One or more of the following validation checks failed: 1) Project should be at least 3 months old. 2) Project must contain valid open source licence. 3) Project's README.md file must contain the text 'This project is tested with BrowserStack.' without quotes. You would have received an email regarding the same, please check it for further details. 

1st: 
Fix up the license detection, we currently mash them all together in a readme style, github expects licence.md simply to be a copy of the lience. See i.e https://github.com/common-voice/common-voice on the right side it's listed as MPL-2.0 whereas we are just "See licence" :D - So i've moved that into a licence readme and added an mpl-2.0 licence.md 

2nd: 
Drop a "this is tested with browserstack" line into the readme.


lesleyjanenorton (approved):
👍

strseb:
Pausing this for now, seems browserstack devices already have a vpn enabled when booked for their services, sooo i cannot enable the vpn on those devices >:(

PR #9493, opened 2024-05-06
Nit: Fix Android-Armv7 builds

strseb:
## Description
This is the error: https://firefox-ci-tc.services.mozilla.com/tasks/Wi-gr5dLRval_TEDLPp8OQ/runs/0/logs/public/logs/live.log
```
Use release config./scripts/android/cmake.sh: line 134: /builds/worker/fetches/Qt/6.6.3/android_armv7/bin/qt-cmake: Permission denied
```
Looking at the packed conda env, it seems the exectutable bit was not set on this file, so let's manually do that.


brizental (approved):
How did this ever work though

PR #9400, opened 2024-04-19
VPN-5872: QML UI module cleanup

oskirby:
## Description
I am attempting to see if I can get the QML cache generation working in the `Mozilla.VPN` module, which should speed up first-boot load times significantly. To do this, we need to stop using `.qrc` files for loading QML content, and register them as QML sources in the `qt_add_qml_module()` CMake function instead.

The reason I stopped short of doing this last time we had to refactor the QML registration, amounted to two issues, which should now be dealt with:
- The generated resources for QML content was required to have a path matching the module URL. But we were mixing `qrc:/ui` and `import Mozilla.VPN` meaning that one or the other had to change.
- The QML type registration was getting lost during linking, and I think this has been resolved with the use of `Q_IMPORT_QML_MODULE(Mozilla_VPNPlugin)`

And finally, in order to remove the `ui.qrc` file altogether, we needed to dive into the translation string scraper job, and fix up [VPN-5515](https://mozilla-hub.atlassian.net/browse/VPN-5515) while we were at it.

## Reference
JIRA issue [VPN-5872](https://mozilla-hub.atlassian.net/browse/VPN-5872)
JIRA issue [VPN-5515](https://mozilla-hub.atlassian.net/browse/VPN-5515)
Qt blog: [QML Modules in Qt 6.2](https://www.qt.io/blog/qml-modules-in-qt-6.2)

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-5872]: https://mozilla-hub.atlassian.net/browse/VPN-5872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPN-5515]: https://mozilla-hub.atlassian.net/browse/VPN-5515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9498, opened 2024-05-07
Fix QML paths for Qt6.5+

strseb:
## Description
In https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9400/
we switched to qml_modules and the paths were updated. On my Mac with Qt 6.6.3 the app no longer finds main.qml.
In Qt6.5+ the way qml_modules path works have been changed https://doc.qt.io/qt-6/qt-cmake-policy-qtp0001.html. 

```
The OLD behavior of this policy is that, the RESOURCE_PREFIX argument for qt_add_qml_module() defaults to ":/".

The NEW behavior of this policy is that the RESOURCE_PREFIX argument for qt_add_qml_module() defaults to ":/qt/qml/". The new behavior ensures that modules are put into the [QML Import Path](https://doc.qt.io/qt-6/qtqml-syntax-imports.html#qml-import-path) and can be found without further setup.
```

Looking at 
```
const QUrl url(QStringLiteral("qrc:/Mozilla/VPN/main.qml"));
  engine->load(url);
```

We actually assume the "OLD" behavior, so let's enable that :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,7 +139,7 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten" AND
 endif()
 
 if(QT_KNOWN_POLICY_QTP0001)
-    qt_policy(SET QTP0001 NEW)
+    qt_policy(SET QTP0001 OLD)
 endif()
 
 message("Using Qt version ${Qt6_VERSION}")


oskirby:
Ah! Such shame! I can't believe I forgot to test this on Qt 6.6.

oskirby (approved):

PR #9501, opened 2024-05-07
Nit: [Sentry] - Pass version using semver

strseb:
## Description
One annoying thing right now is that sentry does not parse our versions. So if we mark something fixed on 2.23 (which is on stage) but we push a 2.22.1 hotfix still with that bug, the bug will be re-opened until a newer release is created. 

That is because sentry currently goes by "when was this release first seen" vs "what does the number say". 

For that we need to pass that string formatted using semver. 
With that pr i created a test crash: see semver is now "yes" for 2.23.
https://mozilla.sentry.io/releases/mozilla-vpn%402.23.0/?project=4504197915607040


lesleyjanenorton (approved):
✨

PR #9505, opened 2024-05-07
VPN-6350 - Add script to update server names and rule to do that periodically

brizental:
We are getting the latest server list from: "https://api.mullvad.net/public/relays/wireguard/v2", I have asked and this is returning the correct list of servers we get from Guardian the only difference is that this is a public API so it just makes everthing easier.

There were two concerns listed in the ticket:

1. Updating a string without changing the string id -- we don't want to do that. I kept it in mind, note the final script never updates the text in a source tag, it only adds text to a new string that it has created. 
2. Removing strings if they re not in the list returned by Mullvad -- we also don't want to do that. The script doesn't remove, it just adds. So that is tken care of.

---

Also note that yes there are strings currently missing. I am specifically not adding them here because I would like to leave it for the script to do once it is merged just to check everything works as expected.


.github/workflows/i18n_server_list_update.yaml
```diff
@@ -0,0 +1,37 @@
+name: i18n Updater
+# This Workflow will periodically go in:
+# 1. Update extras.xliff with any new servers names Mullvad may have added
+# 2. File a pr with the changes to extras.xliff
+
+on:
+  schedule:
+    - cron: "0 7 * * *" # Every Day 7:00 UTC
```
brizental, line 8: Maybe too much? LMK.

.github/workflows/i18n_server_list_update.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          submodules: "true"
+      - name: Update server name strings
+        run: |
+          scripts/ci/update_server_names.py
+      - uses: peter-evans/create-pull-request@v5
```
brizental, line 27: @strseb Just checking, but if there are no changes it just won't create the PR right?

flodolo: Correct, but also the latest version of this action is v6

.github/workflows/i18n_server_list_update.yaml
```diff
@@ -0,0 +1,37 @@
...
+  pull-requests: write
+  contents: write
+
+jobs:
+  main:
+    name: Update server name strings
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
```
flodolo, line 21:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/i18n_server_list_update.yaml
flodolo, line 27:
```suggestion
      - uses: peter-evans/create-pull-request@v6
```

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+    country_names = []
+    city_names = []
+
+    try:
+        response = requests.get(MULLVAD_SERVER_LIST_API)
+
+        if response.status_code == 200:
+            data = response.json()
+
+            locations = data.get('locations', None)
```
flodolo, line 25:
```suggestion
            locations = data.get('locations', {})
```

Set the fallback to an empty dictionary and drop the `if locations`.

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+            if locations:
+                for country_code, location in locations.items():
+                    country_code = country_code.split("-")[0]
+
+                    country_name = location.get("country", None)
+
+                    # We don't want duplicates
+                    # so let's check if this country code is in the list already
+                    found_country = [c for c in country_names if c.get("code") == country_code]
```
flodolo, line 34: Why make `country_names` a list of dictionary instead of just a dictionary?

```python
            locations = data.get('locations', {})
            for location_code, location in locations.items():
                country_code = location_code.split("-")[0]
                country_name = location.get("country", None)
                if country_name and country_code not in country_names:
                    country_names[country_code] = country_name

                city = location.get("city", None)
                if city and city not in city_names:
                    city_names.append(city)
```

And later

```python
for country_code, country_name in countries.items():
        string_map[country_code] = country_name
```

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+    output = ""
+    for chunk in string.split(" "):
+        output += chunk[0].upper()
+        output += chunk[1:]
+    return output
+
+
+def get_city_translation_id(str):
+    # Remove the state suffix e.g. São Paulo, SP -> São Paulo
+    str = str.split(",")[0]
```
flodolo, line 68:
```suggestion
    str = str.split(",")[0].strip()
```

Probably good to avoid traling spaces.

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+    except ValueError as e:
+        print("Error parsing server list:", e)
+
+    return {
+        "countries": country_names,
+        "cities": city_names,
+    }
+
+
+def pascalize(string):
```
flodolo, line 58: You could make this a one-liner

```python
return "".join([chunk.title() for chunk in string.split(" ")])
```

In any case, I would use [.title()](https://www.w3schools.com/python/ref_string_title.asp) instead of working on characters (I assume we don't have city names starting with numbers or symbols).

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+    # No missing strings, return early
+    if not missing:
+        print("Server list is up to date!")
+        sys.exit(0)
+
+    missing_string_map = { key: value for key, value in string_map.items() if key in missing}
+
+    # Iterate over the entries and add them to the XLIFF tree
+    servers_node = root.find(".//{urn:oasis:names:tc:xliff:document:1.2}*[@original='../src/apps/vpn/ui/screens/home/ViewServers.qml']")
+    if not servers_node:
```
flodolo, line 122:
```suggestion
    if servers_node is None:
```

flodolo (changes requested):
We already have a Malmö server

```
      <trans-unit id="servers.Malmö" xml:space="preserve">
        <source>Malmö</source>
      </trans-unit>
```

That means something is not working as expected? Which ID is the code ingesting this XLIFF expecting?

.github/workflows/i18n_server_list_update.yaml
flodolo: Definitely too much, once a week is probably enough…

.github/workflows/i18n_server_list_update.yaml
brizental: This is just in line with the versions used by our other Github actions. I will send a commit where I update this and the create-pull-request actions everywhere just to keep the pattern.

scripts/ci/update_server_names.py
brizental: > Why make country_names a list of dictionary instead of just a dictionary?

No good reason other than I did not think too much about it. Will change, making it a just a dict makes more sense.

scripts/ci/update_server_names.py
```diff
@@ -0,0 +1,156 @@
...
+        new_unit.set("{urn:oasis:names:tc:xliff:document:1.2}xml:space", "preserve")
+
+        source_node = etree.SubElement(new_unit, "{urn:oasis:names:tc:xliff:document:1.2}source")
+        source_node.text = source
+
+    # Save the new tree to extras.xliff file
+    with open(xliff_path, "w", encoding="utf-8") as fp:
+        etree.register_namespace('',"urn:oasis:names:tc:xliff:document:1.2")
+        # Fix indentation
+        reindent(root)
```
oskirby, line 138: What happens if we don't reindent the file?

brizental: The file formatting becomes completely different from what is the l10n repo.

flodolo (approved):

brizental:
Force merging as the CI failures are completely unrelated -- something funky happening with the Taskcluster cache.

.github/workflows/i18n_server_list_update.yaml
```diff
@@ -0,0 +1,37 @@
+name: i18n Updater
```
flodolo, line 1: This has the same name of an existing workflow, it should be changed 
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/

flodolo: It also doesn't work because there is no Python setup step. Opening a PR.

PR #9514, opened 2024-05-08
VPN-6301: Flatpak builds on taskcluster

oskirby:
## Description
It might be possible to get flatpaks building on Taskcluster by making use of a `generic-worker` instead of going through Docker. Let's explore this idea.

## Reference
JIRA issue: [VPN-6301](https://mozilla-hub.atlassian.net/browse/VPN-6301)

## Checklist
    
[VPN-6301]: https://mozilla-hub.atlassian.net/browse/VPN-6301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
Neat! It works!
It's not exactly the friendliest with regards to downloading dependencies in the build tasks, but it does produce builds.

oskirby:
FWIW, another approach to take here would be to have the `flatpak` and `flatpak-builder` packages installed in the base operating system, in which case this task wouldn't need the step of creating a podman worker as we can just rely on the flatpak isolation framework but that wouldn't have been as fun :)

ahal:
> FWIW, another approach to take here would be to have the flatpak and flatpak-builder packages installed in the base operating system, in which case this task wouldn't need the step of creating a podman worker as we can just rely on the flatpak isolation framework but that wouldn't have been as fun :)

This might be a good approach to move to, though we probably don't want to have a special VM image just for this.. We often use toolchain tasks for this kind of thing. Using toolchains to install `flatpak` and `flatpak-builder` at task runtime is probably the cleanest way to do it.

But for now this approach seems fine.

taskcluster/kinds/build/kind.yml
```diff
@@ -15,6 +15,7 @@ transforms:
 kind-dependencies:
     - fetch
     - toolchain
+    - docker-image
```
ahal, line 18: Interesting, I'm surprised we weren't depending on `docker-image` already!

taskcluster/scripts/podman-worker.py
```diff
@@ -0,0 +1,84 @@
...
+# Load the podman image.
+image_filename = os.path.join(moz_fetches_dir, 'image.tar.zst')
+print(f"Loading podman image from: {image_filename}")
+sys.stdout.flush()
+image_data = load_podman_image(image_filename)
+image_config = image_data["Config"]
+
+# Prepare the task arguments
+worker_env = get_task_payload()["env"]
+worker_args = ['podman', 'run', '--rm', '--privileged']
```
ahal, line 49: Just noting that using `--privileged` should be fine because we aren't running this as root, and also `generic-worker` creates a new user for each task. So unlike Docker which forwards commands to a daemon that does run as root, using this flag with podman should be safe from a task isolation standpoint.

ahal (approved):
Thanks, while this is a lot just to get a task going, this is pushing some new ground and gives folks a place to look if they want to something similar. This could even be a good starting point for a more official place in Taskgraph's `run` transforms or similar.

strseb (approved):
⭐

PR #9518, opened 2024-05-09
Fix workflow to update server names

flodolo:
Workflow is currently broken (missing deps), and has the same name of another workflow.

### DIFF
--- a/.github/workflows/i18n_server_list_update.yaml
+++ b/.github/workflows/i18n_server_list_update.yaml
@@ -1,4 +1,4 @@
-name: i18n Updater
+name: i18n Update Servers
 # This Workflow will periodically go in:
 # 1. Update extras.xliff with any new servers names Mullvad may have added
 # 2. File a pr with the changes to extras.xliff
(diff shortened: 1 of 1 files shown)


flodolo:
Confirmed that it's working (manually triggered from branch, created #9519)

.github/workflows/i18n_server_list_update.yaml
```diff
@@ -21,6 +21,11 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: "true"
+      - uses: actions/setup-python@v5
```
flodolo, line 24: Using v5 directly, as there's another PR trying to update all the workflows #9520

brizental (approved):

brizental:
I guess we should drop 96bbdec138ba7f483be8807e5fd87ca4423a8165 before merge.

flodolo:
> .



> I guess we should drop [96bbdec](https://github.com/mozilla-mobile/mozilla-vpn-client/commit/96bbdec138ba7f483be8807e5fd87ca4423a8165) before merge.

Yes, I can manually trigger the workflow after this is merged. I originally didn't realize the PR was opened against the branch.

PR #9520, opened 2024-05-09
Reduce warnings in GitHub workflows

flodolo:
There are currently over 100 warnings in checks, mostly due to obsolete GitHub actions.

### DIFF
--- a/.github/workflows/linux_tests.yaml
+++ b/.github/workflows/linux_tests.yaml
@@ -29,7 +29,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y $(./scripts/linux/getdeps.py -a linux/debian/control)
 
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.9"
           cache: "pip"
@@ -37,7 +37,7 @@ jobs:
 
       - name: Cache grcov
         id: cache-grcov
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: grcov-build/
           key: ${{ runner.os }}-grcov-v0.8.13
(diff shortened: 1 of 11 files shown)


.github/workflows/linux_tests.yaml
```diff
@@ -167,7 +167,7 @@ jobs:
               -o ${{ runner.temp }}/artifacts/functional_lcov.info
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
```
flodolo, line 170: Apparently Codecov requires a token?
https://github.com/codecov/codecov-action/issues/1274

flodolo:
> No files were found with the provided path: /Users/runner/work/_temp/artifacts. No artifacts will be uploaded.

This error is not due to updates, it's been there all the time (macOS tests).

.github/workflows/macos_tests.yaml
```diff
@@ -130,7 +132,7 @@ jobs:
           MVPN_BIN: ./build/dummyvpn
 
       - name: Uploading artifacts
```
flodolo, line 134: I couldn't help noticing that this looks very similar to the Linux workflow, which doesn't fail.

That workflow has an additional step though, using the same artifacts folder, so that might be the reason?

```
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v4
        if: steps.generateGrcov.outcome == 'success'
        with:
          directory: .
          flags: functional_tests
          name: codecov-poc
          files: ${{ runner.temp }}/artifacts/functional_lcov.info
          verbose: true
```

.github/workflows/linux_tests.yaml
flodolo: Someone needs to set up a secret (I have access to the key in Codedov, but not this repo).
https://docs.codecov.com/docs/adding-the-codecov-token#github-actions

oskirby: Honestly, we should just strip out the codecov checks, they haven't worked since Raphael left the team.

flodolo: Let me know if I should remove these steps, or merge this as is ;-)

oskirby: Merge it as-is. I can follow up with another PR to strip out the codecov stuff.

PR #9523, opened 2024-05-09
Remove code coverage from github actions.

oskirby:
## Description
Codecov has never really worked all that great, and hasn't had someone to maintain it since @mozrokafor left the team. So in the interest of cutting down on needless work and maintenence burden, we should just remove it from our CI jobs.

## Reference
None

## Checklist
    
### DIFF
--- a/.github/workflows/macos_tests.yaml
+++ b/.github/workflows/macos_tests.yaml
@@ -51,7 +51,7 @@ jobs:
           mkdir -p build/cmake
           cmake -S $(pwd) -B build/cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
             -DCMAKE_PREFIX_PATH=${{ github.workspace }}/qt_dist/lib/cmake
-          cmake --build build/cmake -j$(nproc) --target dummyvpn
+          cmake --build build/cmake --target dummyvpn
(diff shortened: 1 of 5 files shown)


oskirby:
Well that was wickedly fast! I am guessing that the profiling compiler options necessary for code coverage was massively slowing down the builds.

.github/workflows/macos_tests.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
           mkdir -p build/cmake
           cmake -S $(pwd) -B build/cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
             -DCMAKE_PREFIX_PATH=${{ github.workspace }}/qt_dist/lib/cmake
-          cmake --build build/cmake -j$(nproc) --target dummyvpn
```
oskirby, line 54: This is a sneaky one that keeps catching us out every now and then... `nproc` doesn't exist on MacOS, so this kind of expression expands to `cmake --build build/cmake -j --target dummyvpn` which gets interpreted as either zero or infinite parallelism - neither of which produces speedy builds.

flodolo (approved):
That's a lot more removal that I was expecting 👍🏼

PR #9524, opened 2024-05-09
Fix GH pages workflow (update deploy action)

flodolo:
There is a step in this workflow that only runs for pushes on main, so I missed it.

Not completely sure that updating the actio version will fix it, but it seemed to work on my fork.


oskirby (approved):
Looks good to me!

PR #9525, opened 2024-05-09
Experiment: Try using b-linux-large workers for Taskcluster build tasks

oskirby:
## Description
We have a larger GCP worker available to us on Taskcluster. While we aren't frequently blocked by taskcluster jobs, it might be nice to get an idea of how much faster they could be. Especially if we have dreams of doing more testing in taskcluster.

## Reference
None.

## Checklist
    
### DIFF
--- a/taskcluster/kinds/build/linux.yml
+++ b/taskcluster/kinds/build/linux.yml
@@ -22,6 +22,7 @@ task-defaults:
               path: /builds/worker/artifacts
     run:
         using: run-task
+        checkout: false
         use-caches: true
         command: /builds/worker/builder.sh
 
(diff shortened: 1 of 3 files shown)


oskirby:
Well, that is pretty cool. This resulted in faster build times by a factor of three! The question is, do we need it? IMO our development cycles right now are typically blocked by the functional tests as the weakest link so a faster Taskcluster build, while nice, doesn't really speed us up by as much as we would like.

strseb:
> Well, that is pretty cool. This resulted in faster build times by a factor of three! The question is, do we need it? IMO our development cycles right now are typically blocked by the functional tests as the weakest link so a faster Taskcluster build, while nice, doesn't really speed us up by as much as we would like.

wooow. Well then we might need to check the economics, the android runners already require us to book worker hours. 
If the this worker is not more expensive by the factor of 3 this might make sense to use anyway.

PR #9530, opened 2024-05-10
VPN-5156 - Do not fallback to English on language name map

brizental:
For some reason I thought it would be a good idea to fallback to the translation for "English" in whatever language if the translation for the actual language name was not present. So, say the translation for `es_CL` is missing for `es_CL` we would instead get the translation for `en` ("Inglés") instead of leaving it blank and letting Qt do the translation... 🤦‍♀️


mcleinman:
Thanks for this! "letting Qt do the translation" - can you share a bit more about how that works in this case?

brizental:
Literally we let Qt do it. There is a funcion to grab the native language name from the QLocale.

mcleinman (approved):
👍🏻

lesleyjanenorton (approved):

PR #9531, opened 2024-05-10
VPN-6375 - Fix QML modules import

brizental:
Ok, something funky is happening with the Qt6.6 imports.

Here is the thing, this patch fixes it. But I kinda hate the way it fixes it, makes no sense.

Opening as a draft, as food for thought. It seems sometime Qt6.6 is unable to resolve QML files in the same module???? cc @lesleyjanenorton for QML knowledge.

### DIFF
--- a/src/ui/CMakeLists.txt
+++ b/src/ui/CMakeLists.txt
@@ -5,6 +5,7 @@
 qt_add_qml_module(mozillavpn-ui
     VERSION 1.0
     URI Mozilla.VPN
+    NO_PLUGIN_OPTIONAL
     STATIC
     OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Mozilla/VPN
     QML_FILES
--- a/nebula/ui/components/MZScreenBase.qml
+++ b/nebula/ui/components/MZScreenBase.qml
@@ -36,8 +36,10 @@ Item {
             _menuIconButtonMirror:  stackview.depth !== 1 && MZLocalizer.isRightToLeft
             _iconButtonAccessibleName: stackview.depth === 1 ? MZI18n.GlobalClose : MZI18n.GlobalGoBack
             _menuOnBackClicked: () => maybeRequestPreviousScreen()
-            titleComponent: stackview.currentItem.titleComponent ? stackview.currentItem.titleComponent : null
-            rightButtonComponent: stackview.currentItem.rightMenuButton ? stackview.currentItem.rightMenuButton : null
+            titleComponent: stackview.currentItem && stackview.currentItem.titleComponent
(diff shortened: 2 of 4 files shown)


lesleyjanenorton (commented):
> But I kinda hate the way it fixes it, makes no sense.

I may actually hate this less than the qmldir files which were (I think) facilitating this before. I think my only concern here is that there are related bugs waiting to be found. For instance we'll definitely need to add similar import statements to ViewServers.qml, ViewMultiHop.qml and maybe some of the in-app auth views as well.

brizental:
> I may actually hate this less than the qmldir files which were (I think) facilitating this before.

Interesting. If this is the case, then we might have another easier and better fix. CMake is building the qmldir file for us now, it's not like they are gone. Maybe we just need to point the build to the right place where the file is now 🤔

brizental:
Ok, [fcb07a6](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9531/commits/fcb07a62f4b5886b858763c532a1ed2bafb6db00) fixes the issue for sure without what you initialliy mentioned @lesleyjanenorton . Sending it in just in case, but I am going to keep digging here.

This is quite annoying and although everything looks correct it's like Qt ignores the qmldir file that it itself writes.

src/ui/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@
 qt_add_qml_module(mozillavpn-ui
     VERSION 1.0
     URI Mozilla.VPN
+    NO_PLUGIN_OPTIONAL
```
brizental, line 8: Ok, so here is the thing: this fixes the issue and I don't know why.

With the removal of qmldir files manually written by developers
and the usage of qt_add_qml_module, it's not like there will not
be any more qmldir files, it's that they will be generated for us.

When I looked at the generated files I noted they have this in the header

```
linktarget mozillavpn-uiplugin
plugin mozillavpn-uiplugin
classname Mozilla_VPNPlugin
typeinfo mozillavpn-ui.qmltypes
prefer :/Mozilla/VPN/
```

These are not defined in the lottie and nebula qmldir files -- which work.

I started playing around with options that would change this such as
NO_PLUGIN (which did nothing) and NO_PLUGIN_OPTINAL which fixed the import
issues we were having :shrug.

brizental: Oh yeah, just for the record. After I aded this property
this is what the header looks like:

```
linktarget mozillavpn-uiplugin
plugin mozillavpn-uiplugin
classname Mozilla_VPNPlugin
typeinfo mozillavpn-ui.qmltypes
prefer :/Mozilla/VPN/
```

lesleyjanenorton (approved):
👍

PR #9556, opened 2024-05-13
Fix android build script

mcleinman:
## Description

I believe we missed a character in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9510/. Build timestamp was being set to the string `s`, and thus it was failing downstream.

## Reference

VPN-6390

## Checklist
    

oskirby (approved):
Good catch! Thank you!

PR #9561, opened 2024-05-14
try to fix qml imports

strseb:
## Description

i do think setting the policy to OLD broke things, my bad 🙏 . 
```
qt_policy(SET QTP0001 NEW)
```

-> So let's keep it at new and backport the changes of this policy to 6.2.4
->Set the resource prefix `/qt/qml`, it's the default value after 6.5
-> Add `/qt/qml` to the path by default.

This seems to fix everything on Windows (6.2.4/6.6.0) tested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,7 +139,7 @@ if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Emscripten" AND
 endif()
 
 if(QT_KNOWN_POLICY_QTP0001)
-    qt_policy(SET QTP0001 OLD)
+    qt_policy(SET QTP0001 NEW)
 endif()
 
 message("Using Qt version ${Qt6_VERSION}")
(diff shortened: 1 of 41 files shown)


brizental:
We should undo the NO_PLUGIN_OPTIONAL thing I did before if this is the fix.

lesleyjanenorton (approved):
worksonmymachine-1 


Do we still need the NO_PLUGIN_OPTIONAL change from #9531 ?

mcleinman:
This works for me on Android for 6.6

oskirby:
Works for my MacOS machine (tested both Qt 6.2.4 and Qt 6.6.3).

oskirby (approved):
As a thought for some future cleanup - we don't really need to use the path so prominently and in many cases we can load QML objects by type rather than URL.

For example, a `StackView::push()` can accept an `Item`, `Component` or `url` so in theory something like `stackView.push(ViewAboutUs{..})` should work too.

But that's too much work for now. Let's merge this and get the breakage fixed.

PR #9569, opened 2024-05-15
No ticket - Exit with failure when leaks are detected

brizental:


oskirby:
Just to be completely clear - the leak detector only runs when the application is compiled with Debug, right? Because we have some persistent leaks and I really *don't* want this to cause users to experience abnormal exit conditions when the use our release builds.

brizental:
> Just to be completely clear - the leak detector only runs when the application is compiled with Debug, right? 
 Yes.

strseb:
stale

PR #9571, opened 2024-05-15
DRAFT 😵‍💫 Use Qt 6.6.3 for macos functional tests and windows/mac unit tests 

lesleyjanenorton:
## Description

Here we update the following jobs to use Qt 6.6:
- macos functional tests
- windows unit tests
- macos unit tests 

To follow later: 
- Linux unit and functional tests (Once we've moved / are closer to moving to static Linux builds)
- Wasm tests

## Reference

VPN-6368

## Checklist
    
### DIFF
--- a/tests/unit_tests/testlocalizer.cpp
+++ b/tests/unit_tests/testlocalizer.cpp
@@ -277,7 +277,7 @@ void TestLocalizer::formattedDate_data() {
 
   QTest::addRow("en - future")
       << "en" << QDateTime(QDate(2000, 1, 1), QTime(10, 0), QTimeZone(0))
-      << QDateTime(QDate(2000, 1, 1), QTime(11, 0), QTimeZone(0)) << "10:00 AM"
+      << QDateTime(QDate(2000, 1, 1), QTime(11, 0), QTimeZone(0)) << "10:00 AM"
       << (qint64)(14 * 3600);
   QTest::addRow("es_ES - future")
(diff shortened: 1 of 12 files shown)


tests/unit_tests/testlocalizer.cpp
```diff
@@ -295,7 +295,7 @@ void TestLocalizer::formattedDate_data() {
 
   QTest::addRow("en - one hour ago")
       << "en" << QDateTime(QDate(2000, 1, 1), QTime(10, 0), QTimeZone(0))
-      << QDateTime(QDate(2000, 1, 1), QTime(9, 0), QTimeZone(0)) << "9:00 AM"
```
lesleyjanenorton, line 298: It strangely seems that 6.6 wants a _narrow_ no-break space here but changing this to a _narrow_ no-break white space (U+202F) causes this test to fail on Linux where we are still on 6.2.4. If anyone has immediate thoughts or guesses about why this is different in 6.6, I'd love to hear them.

jcristau (approved):
taskcluster bits lgtm

oskirby:
As a part of this work, I think we also need to bump the default value of `CMAKE_OSX_DEPLOYMENT_TARGET` in the top level CMakeLists.txt file in order to keep up with the [Qt Macos](https://doc.qt.io/qt-6.5/macos.html) supported platforms.

lesleyjanenorton:
Closing in favor of #9620 .

PR #9577, opened 2024-05-16
VPN-6330: More Linux static packaging fixes

oskirby:
## Description
To fix the missing fonts for Qt static builds, we need to rebuild Qt with fontconfig support (which also requires `-system-freetype`) and we patch up a missing `libxcb-cursor0` dependency too while we are at it.

However! Because this just so happened to touch the Qt build scripts, we also needed to address some breakage in MacOS due to a busted homebrew install. In this case, I opted to just ditch homebrew and use conda because it's already a part of our build tooling anyways. 

## Reference
JIRA issue: [VPN-6330](https://mozilla-hub.atlassian.net/browse/VPN-6330)

## Checklist
    
[VPN-6330]: https://mozilla-hub.atlassian.net/browse/VPN-6330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/linux/debian/control
+++ b/linux/debian/control
@@ -44,6 +44,7 @@ Architecture: any
 Depends: libcap2-bin (>=1:2.32),
          wireguard (>=1.0.20200319),
          wireguard-tools (>=1.0.20200319),
+         libxcb-cursor0,
          libqt6quick6 (>=6.2.0~),
          libqt6widgets6 (>=6.2.0~),
          libqt6gui6 (>=6.2.0~),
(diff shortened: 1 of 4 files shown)


oskirby:
Annoyingly, this was a negligibly small change to the Linux side of things - but unfortunately because I touched the Qt build scripts, I now have to deal with all the tech debt hiding in the MacOS build workers to get it going again.

lesleyjanenorton (approved):
🪄✨

PR #9583, opened 2024-05-17
VPN-6402: fix timer

mcleinman:
## Description

In VPN-5266, I put this line in the wrong spot. We want to set this variable to true every time we do a silent switch. But I put it in a block such that it only got set for `eServerCoolDownNeeded`. When changing DNS, this wasn't updated, and thus the timer was stuck on 0:00:00.

## Reference

VPN-6402

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -625,13 +625,12 @@ bool Controller::silentSwitchServers(
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


oskirby (approved):
Looks good!

PR #9596, opened 2024-05-25
VPN-6300: Refactor CryptoSettings and implement for Flatpaks

oskirby:
## Description
`CryptoSettings` were never really a class and just a namespace of static methods, which made it challenging to implement for flatpaks because we need to interact with an XDG portal to fetch secrets and we need some interaction with extra state and signals.

Further complicating matters for Flatpaks, the Secrets portal states that we may be given a `token` string back by the platform and we need to store it somewhere for the next time we fetch secrets. This presents a chicken-and-egg problem in that we can't store it in the settings if they're encrypted. To address this, we add a new `EncryptionChachaPolyV2` file format, which adds a metadata field to the file header where plaintext values can be stored. These values are then passed back to `CryptoSettings::getKey()` during key lookup.

This is a PR that has been kicking around in my local branch for a while and I figured it was time to get it moving again. In this work we:
1. Refactor `CryptoSettings` into a virtual class which can be instantiated by a platform implementation.
2. Refactor some common XDG portal helpers into an `XdgPortal` class.
3. Refactor `CryptoSettings::getKey()` to return a `QByteArray`. Not really necessary but it just makes it more Qt-ish.
4. Implement an `XdgCryptoSettings` that connects to to the XDG [org.freedesktop.portal.Secret](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Secret) portal.
5. Provide a mock `CryptoSettings` implementation for testing.
6. Extend the `CryptoSettings` to provide additional metadata for key lookup.
7. Add a new `EncryptionChachaPolyV2` file format to embed plaintext metadata.
8. Write unit tests for `CryptoSettings`

This turned into a pretty gnarly PR, but it seemed like the best way to do it. Some other solutions considered to the token storage problem:
 - Ignore it altogether: From my testing on Gnome and KDE it seems that neither of these two platforms make use of the `token` and if we just omit it then the secrets portal works fine. We don't really have a viable solution for key rotation via `CryptoSettings::resetKey()` though but I guess we could do some shenanigans with inserting a salt into the unused bytes of the nonce or something.
 - Add a second (plaintext) settings file just to store the salt and token. This also works, and it was even what I was using earlier in this PR. But then it becomes another artifact to manage (and delete) and what happens if it gets out of sync with the settings file?

## Reference
JIRA issue: [VPN-6300](https://mozilla-hub.atlassian.net/browse/VPN-6300)

## Checklist
    
[VPN-6300]: https://mozilla-hub.atlassian.net/browse/VPN-6300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcleinman (approved):
Approval for iOS/macOS - that part looks good. Thanks.

The crypto stuff generally looks great, but would either want someone closer to crypto to review it, or for me to take a lot longer to go over it with a fine toothed comb.

src/cryptosettings.cpp
```diff
@@ -12,16 +12,66 @@
...
 
 namespace {
-
 Logger logger("CryptoSettings");
+CryptoSettings* s_instance = nullptr;
+}  // namespace
 
-uint64_t lastNonce = 0;
+// static
+QSettings::Format CryptoSettings::format() {
```
strseb, line 39: I think this function does 2 things: 
-> be a factory for cryptosettings 
-> now returning a QSettingsFormat. 

i get why format is creating it (after reading the cpp ) but just by reading the header it's not clear why it would do that. Consider splitting the factory and format creator
```
if (!s_instance) {
 CryptoSettings::create()
}
```

src/cryptosettings.cpp
```diff
@@ -12,16 +12,66 @@
...
+  static QSettings::Format format =
+      QSettings::registerFormat("moz", readFile, writeFile);
+
+  // Create a platform crypto settings implementation, if supported.
+  if (!s_instance) {
+#if defined(UNIT_TEST)
+    s_instance = new DummyCryptoSettings();
```
strseb, line 46: That is not needed , as the constructor does that, no?

src/cryptosettings.h
```diff
@@ -5,35 +5,48 @@
...
     NoEncryption,
     EncryptionChachaPolyV1,
+    EncryptionChachaPolyV2,
   };
 
+  explicit CryptoSettings();
+  virtual ~CryptoSettings();
+
+  static QSettings::Format format();
+  static QByteArray generateRandomBytes(qsizetype length);
```
strseb, line 25: no need to be public, can be protected

src/cryptosettings.h
```diff
@@ -5,35 +5,48 @@
...
-class CryptoSettings final {
+class CryptoSettings {
  public:
   enum Version {
     NoEncryption,
     EncryptionChachaPolyV1,
+    EncryptionChachaPolyV2,
   };
 
+  explicit CryptoSettings();
```
strseb, line 21: this could be protected instead of public, given we expect a factory `format()` to create it,no ?

src/platforms/android/androidcryptosettings.cpp
```diff
@@ -52,58 +53,39 @@ inline void jni_clear() {
...
-  if (getKey(key)) {
-    logger.debug() << "Encryption supported!";
-    return CryptoSettings::EncryptionChachaPolyV1;
+  if (getKey(QByteArray()).isEmpty()) {
+    logger.debug() << "No encryption";
+    return CryptoSettings::NoEncryption;
   }
-  logger.debug() << "No encryption";
-  return CryptoSettings::NoEncryption;
+  logger.debug() << "Encryption supported!";
```
strseb, line 89: Small nit: I don't think the log lines are intesting except logger.debug() << "No encryption"; which probably now should be an error state, as to get here we need `(jni_hasKey()) && getKey(QByteArray()).isEmpty()`, which hopefully is unlikely as that means we have a big bug somehow.

src/platforms/linux/xdgportal.h
```diff
@@ -0,0 +1,44 @@
...
+
+constexpr const char* XDG_PORTAL_SERVICE = "org.freedesktop.portal.Desktop";
+constexpr const char* XDG_PORTAL_PATH = "/org/freedesktop/portal/desktop";
+
+constexpr const char* XDG_PORTAL_BACKGROUND =
+    "org.freedesktop.portal.Background";
+constexpr const char* XDG_PORTAL_REQUEST = "org.freedesktop.portal.Request";
+constexpr const char* XDG_PORTAL_SECRET = "org.freedesktop.portal.Secret";
+
+class XdgPortal : public QObject {
```
strseb, line 18: Nit: Could you add a few lines of Doc text about this class? 
as i'm not familiar with portals that would be nice 😅

tests/unit/testcryptosettings.cpp
strseb, line 1: Thank you so much for adding tests! :)

strseb (approved):
r+wc 

I still need to debug why the unit test is segfaulting on my macos machine only :/ 
(Client works fine tho)

src/cryptosettings.h
oskirby: Actually, I don't think it's even possible to instantiate the base class at all given that it has unimplemented virtual functions. I'll make it protected.

For the unit tests we actually just create the derived class directly, so there is one niche use case where we don't use `CryptoSettings::format()`

src/platforms/android/androidcryptosettings.cpp
oskirby: I will tidy this up as per your recommendations.

src/platforms/linux/xdgportal.h
oskirby: Will do

src/cryptosettings.cpp
oskirby: I will split this into a separate `CryptoSettings::create()` method.

src/cryptosettings.cpp
```diff
@@ -12,16 +12,66 @@
...
-}  // namespace
+  return format;
+}
+
+CryptoSettings::CryptoSettings() {
+  logger.info() << "Creating CryptoSettings";
+  if (s_instance) {
+    delete s_instance;
+  }
+  s_instance = this;
```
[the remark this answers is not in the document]
oskirby: Will do

PR #9614, opened 2024-05-28
Clang-Tidy: Enforce modernize-use-auto

strseb:
## Description
We currently have clang-tidy but no enforced rules :) 

use auto is simple, instead of writing the Type twice: 
```
NetworkRequest* request = new NetworkRequest(static_cast<Task*>(parent()));
```
just auto deduct 
```
auto* request = new NetworkRequest(static_cast<Task*>(parent()));
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,6 +14,7 @@ Checks: >
   misc*,
   -misc-no-recursion
 WarningsAsErrors: >
+  modernize-use-auto,
   clang-analyzer-security.*
 AnalyzeTemporaryDtors: false
 # Use our .clang-format file when applying fixes.
--- a/src/addons/addon.cpp
+++ b/src/addons/addon.cpp
@@ -684,7 +684,7 @@ void Addon::maybeLoadLanguageFallback(const QString& code) {
 }
 
 bool Addon::createTranslator(const QLocale& locale) {
-  QTranslator* translator = new QTranslator(this);
+  auto translator = new QTranslator(this);
   m_translators.append(translator);
   QCoreApplication::installTranslator(translator);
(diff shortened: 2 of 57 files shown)


oskirby:
I have mixed feelings on the use of auto - I like it when eliminating redundant type expressions in the same line, for example:

```
// this sucks
NetworkRequest* foo = new NetworkRequest(something);
// but this is good
auto foo = new NetworkRequest(something);
```

On the other hand, when the type isn't expressed by left-hand-side of the assignment expression, then I am surprisingly less okay with using auto as it now requires context to know what the type evaluates too:

```
// this is fine actually
NetworkRequest* example = createRequestFor("zanzibar");
// this makes things harder to understand because I have to go look up the declaration of createReqestFor();
auto confusing = createRequestFor("mykonos");
```

strseb:
>// this is fine actually
NetworkRequest* example = createRequestFor("zanzibar");
// this makes things harder to understand because I have to go look up the declaration of createReqestFor();
auto confusing = createRequestFor("mykonos");

Ah I just have seen this
> auto* request = fxaHawkPostRequest(m_task, "/v1/totp/create",

 in my PR, that is actually not enforced by clang tidy and this rule.
That was just my lazy search replace haha. i'll remove that. Sorry. I agree this is not really clear.

@lesleyjanenorton, @mcleinman - Please drop your thoughts too :)

mcleinman:
I agree with Naomi on that readability, but I may also be an old curmudgeon. But since it sounds like it's not needed, I'm all for this. (In general, I think auto-forcing some style stuff - that we all agree on - is good.) Thanks!

PR #9615, opened 2024-05-28
Add Qt 6.6.3 for Windows and macOS to TC (for #9571)

lesleyjanenorton:
This PR adds qt-6.6.3 toolchain tasks for Windows and Mac. It also re-enables some of the qt features I _think_ we'll need for the unit tests to pass when we update them to use 6.6.


taskcluster/kinds/toolchain/qt.yml
```diff
@@ -26,6 +26,23 @@ qt-macos-6.6:
         symbol: TL(qt-mac-6.6)
     worker-type: b-macos
 
+qt-macos-6.6.3:
```
oskirby, line 29: I would prefer to leave the toolchain tasks with just the major/minor version numbers in their name, and just bump the fetch task's URL and SHA whenever we need to update the patch version to pull in fixes. Qt promises ABI compatibility as long as the major/minor is unchanged, so there is no need to create additional jobs for this kind of update.

TLDR: Please leave the toolchain tasks as `qt-<platform>-6.6` and just update the source tarball fetch.

lesleyjanenorton: Sounds good, makes sense. The 6.6 Windows toolchain task, alias, and symbol include the patch number (`-6.6.0`) do you know if this was done for some specific reason or are we good to to change these `-6.6`?

oskirby: I don't know of any reason why the patch number was added - I think we are good to change it to `-6.6`

lesleyjanenorton:
As CI approaches 1 hour...

strseb (approved):

oskirby (approved):

PR #9616, opened 2024-05-28
Experiment: Run autoupdate check on flatpak-pydeps.yml

oskirby:
## Description
Right now flatpaks are built with their own custom python packages in `linux/flatpak-pydeps.yml` and this has fallen out of sync with our source of truth (the top level `requirements.txt`) it would be good if this could be automatically kept up to date.

Note - this doesn't automatically re-trigger the workflows. This is due to [limitations set by GitHub](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).
>  When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
I don't think this can be done in a way that doesn't screw up dependabot PRs when it comes to merging and rebasing. Going to close this for the time being.

PR #9618, opened 2024-05-29
Nit: Combine App::state and App::userState to fix test races

oskirby:
## Description
Spotted in the wild: a flaky [testSubscription.js](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/9281903149/job/25539154427?pr=9617).

This was hauntingly familliar to a bug I fixed in one of my experimental hackathons in which `vpn.authenticateInApp()` returns success as soon as we get to the `UserAuthenticated` state but the model initialization is still waiting on a bunch of tasks to finish. If this happens during the initial test setup, and we quit before those jobs finish -  we are left with a settings file that is incomplete (the token is set but no keys are) and all the subsequent test jobs fail to load.

To fix this properly, I think the best approach is to combine the `App::state` and `App::userState` states together to eliminate potential race conditions between them, and to add a new `userAuthenticated` property that returns `true` once the auth is completed.

## Reference
Cherry-picking a bunch of stuff from PR #9584
Flaky test run: [testSubscription.js](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/9281903149/job/25539154427?pr=9617)

## Checklist
    
### DIFF
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -40,7 +40,7 @@ void App::setState(int state) {
   logger.debug() << "Set state:" << state;
 
   m_state = state;
-  emit userAuthenticationChanged();
+  emit userAuthenticationMaybeChanged();
   emit stateChanged();
 }
 
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -36,8 +36,6 @@ App::~App() { MZ_COUNT_DTOR(App); }
 
 int App::state() const { return m_state; }
 
-App::UserState App::userState() const { return m_userState; }
-
 void App::setState(int state) {
   logger.debug() << "Set state:" << state;
 
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -30,9 +30,6 @@ App::App(QObject* parent) : QObject(parent) {
                                     WasmNetworkRequest::postResource,
                                     WasmNetworkRequest::postResourceIODevice);
 #endif
-
-  connect(SettingsHolder::instance(), &SettingsHolder::tokenChanged, this,
-          [this]() { emit userAuthenticationChanged(); });
 }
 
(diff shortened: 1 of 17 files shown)


oskirby:
This is confusing having both a `State` and `UserState` that both kind of do the same thing - perhaps a bigger refactoring is in order here. Is this a legacy of the attempted split-app suite thing?

src/app.cpp
```diff
@@ -36,23 +36,14 @@ App::~App() { MZ_COUNT_DTOR(App); }
 
 int App::state() const { return m_state; }
 
-App::UserState App::userState() const { return m_userState; }
-
 void App::setState(int state) {
   logger.debug() << "Set state:" << state;
 
   m_state = state;
+  emit userAuthenticationChanged();
```
mcleinman, line 43: Why is this emitted here? Isn't it possible the user state didn't change?

mcleinman (changes requested):
I really, really like this change. Thank you!

One question on a change. Thanks!

src/app.cpp
oskirby: There is no longer a `userState` to change, but the `userAuthenticated` property is derived from the state now, so when the state changes we need to emit a notification that `userAuthenticated` may have changed value as well.

mcleinman: Agreed fully, though I missed that we check if user is authenticated in the one place we listen to that signal.

I think we should rename this to `userAuthenticationMaybeChanged` or something like that - to make clear that's its not guaranteed that the state changed when that is emitted.

oskirby: Done!

mcleinman (approved):
Thanks, looks great!

PR #9638, opened 2024-06-04
ci: split 'beetmover-apt' into promote and ship tasks

ahal:
## Description

We want to have two tasks, one for promote and one for ship. The bucket / scopes that gets used should depend only on which of these tasks is running, and the tasks should then get filtered out at the target phase.

We also remove the 'filter_shipping' transforms because these tasks will get filtered at the target phase. The generation (aka full) phase doesn't need to be doing any pre-filtering of its own.

## Checklist
    
### DIFF
--- a/taskcluster/kinds/beetmover-apt/kind.yml
+++ b/taskcluster/kinds/beetmover-apt/kind.yml
@@ -12,19 +12,24 @@ transforms:
 kind-dependencies:
     - beetmover-promote
 
-tasks:
-    push-to-gcr:
-        from-deps:
-            with-attributes:
-                build-type:
-                    - linux64/release-deb
-        run-on-tasks-for: [action]
-        description: description
-        worker-type: beetmover-apt
-        worker:
-            product: vpn
-            max-run-time: 1800
-            gcs-sources: [] # Set from Transform beetmover_apt
-        
+task-defaults:
+    description: Import .deb package into Google Artifact Registry
+    run-on-tasks-for: [action]
+    from-deps:
+        with-attributes:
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
This is fine, but one thing I noticed while sanity checking is that we end up with two attributes for shipping phase:
```
beetmover-apt-promote-linux64-deb:
  attributes:
    always_target: false
    kind: beetmover-apt
    primary-kind-dependency: beetmover-promote
    run_on_projects:
    - all
    run_on_tasks_for:
    - action
    shipping-phase: promote-client
    shipping_phase: null
```

Elsewhere (eg: [this Firefox release promotion action](https://firefoxci.taskcluster-artifacts.net/JHWgIpH3TZGUGvg4kVItYA/0/public/task-graph.json)) we only have the underscore version in attributes. We should probably adjust this repo to only use the underscore version in attributes at some point, but it ought not to block this work.

ahal:
> This is fine, but one thing I noticed while sanity checking is that we end up with two attributes for shipping phase

Yeah I noticed that too, the underscore version gets added by the `task.py` transforms. I think it's always been the case, and unfortunately I made `mozilla-taskgraph` use the dash version (as I implemented most of `mozilla-taskgraph` to suppor this repo).

I agree we should switch to underscore.. it'll require a major bump to `mozilla-taskgraph` though (sigh)

hneiva:
I have a feeling the "Add landing label" job is failing because it's coming from a fork. The error suggests the `permissions.pull-requests: write` isn't set, but it is.

PR #9639, opened 2024-06-05
VPN-5992 Allow extensions to request Access to the Proxy Network

strseb:
This pull request introduces support for a new StateOnPartial state in the controller, enhancing the VPN's state management. It also adds an ActivationPrincipal enum to track whether a connection was initiated by a user or an extension, along with corresponding permission checks for deactivation. Additionally, this update includes several new methods and refines existing ones to handle the new state and activation principal, ensuring improved VPN control and security.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -274,14 +274,17 @@ void Controller::handshakeTimeout() {
   ++m_connectionRetry;
   emit connectionRetryChanged();
   logger.info() << "Connection attempt " << m_connectionRetry;
+  InterfaceConfig& exit_hop = m_activationQueue.last();
+  Q_ASSERT(exit_hop.m_hopType != InterfaceConfig::HopType::MultiHopEntry);
+
   if (m_connectionRetry == 1) {
     logger.info() << "Connection Attempt: Using Port 53 Option this time.";
     // On the first retry, opportunisticly try again using the port 53
     // option enabled, if that feature is disabled.
-    activateInternal(ForceDNSPort, RandomizeServerSelection);
+    activateInternal(ForceDNSPort,exit_hop.m_allowedIPAddressRanges, RandomizeServerSelection);
     return;
   } else if (m_connectionRetry < CONNECTION_MAX_RETRY) {
-    activateInternal(DoNotForceDNSPort, RandomizeServerSelection);
+    activateInternal(DoNotForceDNSPort,exit_hop.m_allowedIPAddressRanges,  RandomizeServerSelection);
     return;
   }
 
@@ -349,7 +352,8 @@ void Controller::logout() {
 }
 
 void Controller::activateInternal(DNSPortPolicy dnsPort,
-                                  ServerSelectionPolicy serverSelectionPolicy) {
+                                  std::optional<QList<IPAddress>> allowedIPList = {}, 
+                                  ServerSelectionPolicy serverSelectionPolicy = RandomizeServerSelection) {
   logger.debug() << "Activation internal";
   Q_ASSERT(m_impl);
 
@@ -389,7 +393,7 @@ void Controller::activateInternal(DNSPortPolicy dnsPort,
   exitConfig.m_serverIpv4AddrIn = exitServer.ipv4AddrIn();
   exitConfig.m_serverIpv6AddrIn = exitServer.ipv6AddrIn();
   exitConfig.m_serverPort = exitServer.choosePort();
-  exitConfig.m_allowedIPAddressRanges = getAllowedIPAddressRanges(exitServer);
+  exitConfig.m_allowedIPAddressRanges = allowedIPList.value_or(getAllowedIPAddressRanges(exitServer));
   exitConfig.m_dnsServer = DNSHelper::getDNS(exitServer.ipv4Gateway());
 #if defined(MZ_ANDROID) || defined(MZ_IOS)
   exitConfig.m_installationId = settingsHolder->installationId();
@@ -579,6 +583,16 @@ QList<IPAddress> Controller::getAllowedIPAddressRanges(
   return list;
 }
 
+// static
+ QList<IPAddress> Controller::getExtensionProxyAddressRanges(){
+  return {
+    IPAddress{
+      QHostAddress{MULLVAD_PROXY_RANGE}, 
+      MULLVAD_PROXY_RANGE_LENGTH}
+  };
+}
+
+
 void Controller::activateNext() {
   MozillaVPN* vpn = MozillaVPN::instance();
   const Device* device = vpn->deviceModel()->currentDevice(vpn->keys());
@@ -748,7 +762,7 @@ void Controller::disconnected() {
 
   if (nextStep == None &&
       (m_state == StateSwitching || m_state == StateSilentSwitching)) {
-    activate(m_nextServerData, m_nextServerSelectionPolicy);
+    activate(m_nextServerData, m_iniator, m_nextServerSelectionPolicy);
     return;
   }
 
@@ -757,7 +771,7 @@ void Controller::disconnected() {
   if (m_state != StateConfirming) {
     emit recordConnectionEndTelemetry();
   }
-
+  m_iniator = Null;
   setState(StateOff);
 }
 
@@ -945,6 +959,7 @@ QString Controller::currentServerString() const {
 // for now they will reside here
 
 bool Controller::activate(const ServerData& serverData,
+                          ActivationInitiator initiator, 
                           ServerSelectionPolicy serverSelectionPolicy) {
   logger.debug() << "Activation" << m_state;
   if (m_state != Controller::StateOff &&
@@ -959,6 +974,13 @@ bool Controller::activate(const ServerData& serverData,
   logger.debug() << "Set isSwitchingServer to" << isSwitchingServer;
 
   m_serverData = serverData;
+
+  // If the Activation is Triggered by a WebExtension, 
+  // Only Protext the Proxy Address Range.
+  std::optional<QList<IPAddress>> allowedIPList = initiator == ExtensionUser ?
+      std::optional(getExtensionProxyAddressRanges()):
+      std::nullopt;
+  m_iniator = initiator;
 
 #ifdef MZ_DUMMY
   emit currentServerChanged();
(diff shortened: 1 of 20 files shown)


mcleinman:
Since @lesleyjanenorton has more context, I'm going to let her review this one. If you want a review from me as well, let me know - I'll take the time to read the docs for this project so I can do this PR justice.

src/controller.cpp
```diff
@@ -342,16 +345,19 @@ void Controller::logout() {
...
 
-void Controller::activateInternal(DNSPortPolicy dnsPort,
-                                  ServerSelectionPolicy serverSelectionPolicy) {
+void Controller::activateInternal(
+    DNSPortPolicy dnsPort,
+    ServerSelectionPolicy serverSelectionPolicy = RandomizeServerSelection,
+    ActivationPrincipal initiator = ClientUser) {
   logger.debug() << "Activation internal";
   Q_ASSERT(m_impl);
+  m_iniator = initiator;
```
lesleyjanenorton, line 360: Should this be m_initiator?

src/controller.cpp
```diff
@@ -705,7 +722,11 @@ void Controller::connected(const QString& pubkey) {
 
   // We have succesfully completed all pending connections.
   logger.debug() << "Connected from state:" << m_state;
-  setState(StateOn);
+  if (m_iniator == ExtensionUser) {
```
lesleyjanenorton, line 725: Same question

src/controller.cpp
```diff
@@ -757,7 +784,7 @@ void Controller::disconnected() {
   if (m_state != StateConfirming) {
     emit recordConnectionEndTelemetry();
   }
-
+  m_iniator = Null;
```
lesleyjanenorton, line 787: 🤓

src/controller.cpp
```diff
@@ -1022,33 +1051,46 @@ bool Controller::activate(const ServerData& serverData,
...
               clearRetryCounter();
-              activateInternal(DoNotForceDNSPort, serverSelectionPolicy);
+              activateInternal(DoNotForceDNSPort, serverSelectionPolicy,
+                               initiator);
             });
 
     connect(request, &QObject::destroyed, task, &QObject::deleteLater);
     return true;
   }
+  if (m_state == StateOnPartial) {
```
lesleyjanenorton, line 1071: Thinking out loud: Do we need to move to `StateConnecting` here since we don't care about showing any connecting UI in the client? I'm wondering if we actually need something like a `StatePartialConnecting` ? I'm thinking it could be painful to convince the front-end to not respond to `StateConnecting` here.

src/inspector/inspectorhandler.cpp
```diff
@@ -764,7 +764,7 @@ void InspectorHandler::initialize() {
 #ifdef MZ_WASM
   WasmInspector::instance();
 #else
-  if (!Constants::inProduction()) {
+  if (true) {
```
lesleyjanenorton, line 767: yolo

src/statusicon.cpp
```diff
@@ -73,8 +73,6 @@ void StatusIcon::activateAnimation() {
 }
 
 void StatusIcon::animateIcon() {
-  logger.debug() << "Animate icon";
```
lesleyjanenorton, line 76: thank you

src/systemtraynotificationhandler.cpp
```diff
@@ -223,8 +223,6 @@ void SystemTrayNotificationHandler::updateContextMenu() {
 }
 
 void SystemTrayNotificationHandler::updateIcon() {
-  logger.debug() << "Update icon";
```
lesleyjanenorton, line 226: Why didn't we do this before?  🎉

lesleyjanenorton (commented):
Will keep going through this tomorrow but it is looking good! Just a few comments/Qs so far.

src/controller.cpp
strseb: We need to save the initiator passed to activateInternal, it's not getting set anywhere else 😓 (except for resetting it on deactivation)

strseb: Or did you point out the obvious typo? Hahaha. yes.

src/inspector/inspectorhandler.cpp
strseb: removed that :)

lesleyjanenorton: not defeated, just delayed :)

src/ui/screens/home/controller/VPNToggle.qml
```diff
@@ -19,7 +19,8 @@ MZButtonBase {
 
     function handleClick() {
         toolTip.close();
-        if (VPNController.state !== VPNController.StateOff) {
+        if (VPNController.state !== VPNController.StateOff &&
+            VPNController.state !== VPNController.StateOnPartial) {
```
lesleyjanenorton, line 23: One scenario I think we'll need to resolve but not in this PR...
- We connect via the extension, we're in StateOnPartial...
- We connect via the client, now we're in StateOn...
- We deactivate from the client... we're in StateOff... but then that means Firefox is now also unprotected(?) 

This is how it is supposed to work from the PRD but I'm guessing we'll get foxfooder feedback saying this is unexpected behavior so we might think about how to persist the extension vpn state through a deactivation in the client.

src/webextensionadapter.cpp
```diff
@@ -122,6 +137,10 @@ QJsonObject WebExtensionAdapter::serializeStatus() {
 
   {
     Controller::State state = vpn->controller()->state();
+    if (state == Controller::StateOnPartial) {
+      state = Controller::StateOn;  // Old extensions like MAC dont know and
+                                    // dont need to know about this state.
```
lesleyjanenorton, line 142: We can always update MAC to handle this if we for some reason need to

tests/functional/helper.js
```diff
@@ -352,18 +352,33 @@ module.exports = {
...
         'Mozilla.Shared', 'MZUrlOpener', 'lastUrl');
   },
-
-  async waitForCondition(condition, waitTimeInMilliSecs = 500) {
+  /**
+   * Calls condition() every waitTimeInMilliSecs, promise resolves
+   * if condition returns true.
+   *
+   * @param {()=> Promise<Boolean>} condition -  A Callable Function, returning
+   *     true if the Condition is fulfulledbcondition
```
lesleyjanenorton, line 360:
```suggestion
   *     true if the Condition is fulfilledbycondition
```

tests/functional/helper.js
```diff
@@ -352,18 +352,33 @@ module.exports = {
...
+        if (description != '') {
+          // If there is a description mark the assert
+          assert(true, 'Condition passed: ' + description);
+        }
         clearTimeout(timeout)
         return;
       }
       // Asserting here produces a more useful backtrace for diagnosing tests.
-      assert(active, "Condition timed out");
+      assert(active, 'Condition timed out: ' + description);
```
lesleyjanenorton, line 381: helpful!

tests/functional/testWebExtensionApi.js
```diff
@@ -0,0 +1,89 @@
...
+    sentToClient(new ExtensionMessage('deactivate'), sock);
+    await vpn.waitForCondition(async () => {
+      sentToClient(new ExtensionMessage('status'), sock);
+      const msg = await readResponseOfType('status', messagePipe);
+      return msg.status.vpn === 'StateOff';
+    });
+    sock.destroy();
+  });
+
+  it('A Webextension can NOT deactivate the VPN if it was activated in Client',
```
lesleyjanenorton, line 63: Related to other comment: Eventually we _will_ allow deactivating the vpn for Firefox in this situation right? This may be a terrible idea but I'm almost wondering if we'll want another state machine for the extension to make keeping track of situations like _client is on but Firefox VPN is off_ a bit easier. Otherwise if we do allow the user to remove Firefox traffic from the tunnel when the vpn was activated from the client, but tell the extension that the VPN is 'on', how will the extension know that Firefox is excluded/"off"?

src/controller.cpp
```diff
@@ -1019,36 +1046,53 @@ bool Controller::activate(const ServerData& serverData,
...
   clearRetryCounter();
-  activateInternal(DoNotForceDNSPort, serverSelectionPolicy);
+  activateInternal(DoNotForceDNSPort, serverSelectionPolicy, initiator);
   return true;
 }
 
-bool Controller::deactivate() {
+bool Controller::deactivate(ActivationPrincipal user) {
   logger.debug() << "Deactivation" << m_state;
+  if (m_initiator > user) {
```
lesleyjanenorton, line 1085: Eventually we want to embellish this to allow deactivating the VPN _for Firefox_ from the extension in this situation though right?

src/controller.cpp
```diff
@@ -388,8 +397,7 @@ void Controller::activateInternal(DNSPortPolicy dnsPort,
   exitConfig.m_serverPublicKey = exitServer.publicKey();
   exitConfig.m_serverIpv4AddrIn = exitServer.ipv4AddrIn();
   exitConfig.m_serverIpv6AddrIn = exitServer.ipv6AddrIn();
-  exitConfig.m_serverPort = exitServer.choosePort();
```
lesleyjanenorton, line 391: Just verifying we don't need this? I see we're setting it again further down but just checking.

lesleyjanenorton (approved):
R+ Left a few open questions but what we want to do with them is probably outside the scope of this PR.

src/ui/screens/home/controller/VPNToggle.qml
strseb: Given that the proxy api works on request basis, and we stream the state of the vpn we can simply request to re-enable the vpn to partial with the next request. Given we set a proxy in that case, the request would have a small delay. 
Or we check before disconnecting if there is an active web-extension connection. however then we rely on the port continuesly beeing open.

src/webextensionadapter.cpp
strseb: True but also mac should not care about our internal states 😅

src/controller.cpp
strseb: probably not. :) 
A: We cannot know that whatever is talking using the web-extension api really is firefox. 
B: We don't need to disable the vpn for firefox to not use the vpn connection see https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9676

tests/functional/testWebExtensionApi.js
strseb: The Webextension state machine is going to be diffrent from the client. 
If full device protection is enabled, the extension can use the localproxy to "exlude firefox"
-> if it's in partialOn we can just not use a proxy. 

Thinking about this more, i think we might need to tell the extension actually if it is partially or full-device on. 
But in no situation we need the extension to downgrade from full device protection :)

src/webextensionadapter.cpp
strseb: Let's do that, we probably should start thinking about versioning the api too.

src/controller.cpp
strseb: Ahhh thank you! that is still needed

PR #9645, opened 2024-06-06
Remove iOS arch fixups

mcleinman:
## Description

Without this file, I believe the xcode project uses it's default - which is building for both simulator and device. With that, I'm not sure this file is needed.

Optimistically putting up this PR based on a brief chat we had about this - but I may be missing something here. @oskirby , please reject this PR if I am.

## Reference

N/A

## Checklist
    

oskirby:
I think this is still needed - the reason is that Qt uses the build architecture to select between device and simulator and until they fix up [QTBUG-101276](https://bugreports.qt.io/browse/QTBUG-101276) we need to force the iOS simulator builds to target the x86_64 architecture.

That being said, I am willing to be proven wrong if the removal of this file still results in linkable and runnable iOS simulator builds on a M1/M2 mac.

mcleinman:
Compiling with Xcode, I'm able to run builds on simulator on Apple Silicon with this removed.

oskirby (approved):
In that case, I approve!

PR #9657, opened 2024-06-11
VPN-6423: Fix TestQmlPath on Qt 6.6.0 and later

oskirby:
## Description
The `TestQmlPath` test has been failing for me on Qt 6.6.0 and later, and after a bit digging I traced the crash down into the depths of Qt, which was failing at `qquickpathview.cpp:68:75` with the expression `QGuiApplicationPrivate::platformIntegration()` returning a null pointer. This occurs because the app unit tests are invoked as `QCoreApplication` instead of a `QGuiApplication` so the platform integration plugins are never initialized, leading to the segmentation fault in the QML engine.

There are two reasonable ways to fix this issue:
1. We move this test case into the QML Tests, which does initialize the Gui and QML engine, or
2. We just convert the app unit tests into a GUI application.

I have gone with option 2 here.

## Reference
JIRA issue: [VPN-6423](https://mozilla-hub.atlassian.net/browse/VPN-6423)

## Checklist
    
[VPN-6423]: https://mozilla-hub.atlassian.net/browse/VPN-6423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit_tests/CMakeLists.txt
+++ b/tests/unit_tests/CMakeLists.txt
@@ -169,6 +169,10 @@ if(NOT BUILD_ADJUST_SDK_TOKEN)
     )
 endif()
 
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    list(APPEND UNIT_TEST_ARGS -platform offscreen)
+endif()
+
 ## Add the tests to be run, one for each test class.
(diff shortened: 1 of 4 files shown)


lesleyjanenorton (approved):
🎉

PR #9661, opened 2024-06-13
Fix iOS toolchain generation

strseb:
## Description

if `scripts/macos/conda_setup_qt.sh` fails due to a network error we still continue and cache that. 
This butched the iOS conda toolchain. So let's fix that, and let it fail so tc can resheudle..


lesleyjanenorton (approved):
nice! ✨

PR #9674, opened 2024-06-19
VPN-6363: Better LAN exclusions

oskirby:
## Description
The way we set up LAN exclusions has begun to diverge between platforms, and I think we have some bugs in Linux. Specifically I have found that the IP routing policy of `ip rule add table main suppress_prefixlength 0` has basically meant that we defer to the `main` routing table for all destinations, except the default route. This kindof works, except what we really intended was to send all traffic to the wireguard interface first and then punt traffic back to the main table if it wasn't really supposed to go to the wireguard interface.

To fix this, I think it's about time that we move the routing exclusion logic out of the client and make it a part of the Daemon setup - it has always kinda been this way for iOS anyways so we might as well make it the standard procedure.

Fun fact, it turns out that Android actually has an API for this that would be much better suited to setting up address-range exclusions: [VpnService.Builder.excludeRoute()](https://developer.android.com/reference/android/net/VpnService.Builder#excludeRoute(android.net.IpPrefix))

## Reference
JIRA Issue [VPN-6363](https://mozilla-hub.atlassian.net/browse/VPN-6363)
JIRA Issue [VPN-6482](https://mozilla-hub.atlassian.net/browse/VPN-6482)
JIRA Issue [VPN-5865](https://mozilla-hub.atlassian.net/browse/VPN-5865)
Previous PR #9257

## Checklist
    
[VPN-6363]: https://mozilla-hub.atlassian.net/browse/VPN-6363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-5865]: https://mozilla-hub.atlassian.net/browse/VPN-5865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/cmake/linux.cmake
```diff
@@ -53,11 +53,6 @@ if(NOT BUILD_FLATPAK)
     target_sources(mozillavpn PRIVATE
         ${CMAKE_SOURCE_DIR}/3rdparty/wireguard-tools/contrib/embeddable-wg-library/wireguard.c
         ${CMAKE_SOURCE_DIR}/3rdparty/wireguard-tools/contrib/embeddable-wg-library/wireguard.h
-        ${CMAKE_SOURCE_DIR}/src/daemon/daemon.cpp
-        ${CMAKE_SOURCE_DIR}/src/daemon/daemon.h
-        ${CMAKE_SOURCE_DIR}/src/daemon/dnsutils.h
-        ${CMAKE_SOURCE_DIR}/src/daemon/iputils.h
-        ${CMAKE_SOURCE_DIR}/src/daemon/wireguardutils.h
```
mcleinman, line 60: This is because it's also in the shared sources cmake file, right?

src/daemon/daemon.cpp
```diff
@@ -115,10 +116,28 @@ bool Daemon::activate(const InterfaceConfig& config) {
...
+    // Bring the interface up.
+    if (supportIPUtils()) {
+      if (!iputils()->addInterfaceIPs(config)) {
+        return false;
+      }
+      if (!iputils()->setMTUAndUp(config)) {
+        return false;
+      }
+    }
+
```
mcleinman, line 134: Just so I better understand, why move this up?

src/daemon/wireguardutils.h
```diff
@@ -41,23 +41,9 @@ class WireguardUtils : public QObject {
...
-    Q_UNUSED(prefix);
-    return true;
-  }
-  virtual bool deleteExclusionRoute(const IPAddress& prefix) {
-    Q_UNUSED(prefix);
-    return true;
-  }
+  virtual bool updateRoutePrefix(const IPAddress& prefix) = 0;
+  virtual bool deleteRoutePrefix(const IPAddress& prefix) = 0;
+  virtual bool excludeLocalNetworks(const QList<IPAddress>& addresses) = 0;
```
mcleinman, line 46: Why did we need these updates? (Just trying to better understand this PR.)

src/platforms/macos/daemon/wireguardutilsmacos.cpp
```diff
@@ -265,10 +266,10 @@ bool WireguardUtilsMacos::deleteRoutePrefix(const IPAddress& prefix) {
   if (!m_rtmonitor) {
     return false;
   }
+
   if (prefix.prefixLength() > 0) {
-    return m_rtmonitor->insertRoute(prefix);
+    return m_rtmonitor->deleteRoute(prefix);
```
mcleinman, line 271: Was this a separate bug?

mcleinman (commented):
I didn't review the Linux and Windows portions, as I have no context on how we do routing on those platforms. If nobody else with expertise in that area can review this, I can dig in and learn (and review those portions of the PR).

Everything else looks good - just a few questions to make sure I understand it.  Thanks - this seems great!

src/cmake/linux.cmake
oskirby: That is correct, this is just some cleanup that I missed in #9584 and I stumbled across it while working on this PR.

src/daemon/daemon.cpp
oskirby: The root cause of the issue here is that on Linux and MacOS we need to wait for the interface to go up before we can add routes to it. In the last iteration of work on the Linux firewall (PR #9257) we added some code to check for the interface going up and then load the routing table, which worked fine when the only thing we were loading was a default route, but now that we are are installing multiple routing table entries it gets a lot more complicated to follow that same pattern.

The easier (and cleaner solution) is to just bring the interface up immediately after creation, and before trying to configure a peer and its routes.

src/platforms/macos/daemon/wireguardutilsmacos.cpp
oskirby: Yes, I think this was a bug. The impact of the bug was small because the routes are mostly automatically deleted upon interface destruction anyways - but it could result in zombie routes in the table if the interface doesn't go down (eg: during a silent server switch).

src/daemon/wireguardutils.h
oskirby: The `addExclusionRoute()` and `deleteExclusionRoute()` methods weren't being used anywhere so it seemed best to remove them in favor of the new `excludeLocalNetworks()` method.

mcleinman: Fully missed those other ones subclassed this one. 🤦🏻

mcleinman (approved):
**This approval should be considered partial.** I didn't review the Linux and Windows portions, as I have no context on how we do routing on those platforms. (If nobody else with expertise in that area can review this, I can dig in and learn and review those portions of the PR.)

src/platforms/windows/daemon/wireguardutilswindows.h
```diff
@@ -52,7 +50,7 @@ class WireguardUtilsWindows final : public WireguardUtils {
   std::unique_ptr<WireGuardAPI> m_wireguard_api;
   ulong m_deviceIpv4_Handle = 0;
 
-  WindowsRouteMonitor m_routeMonitor;
+  WindowsRouteMonitor* m_routeMonitor = nullptr;
```
strseb, line 53: Nitpick, maybe a QPointer, so that 
```
   delete m_routeMonitor;
    m_routeMonitor = nullptr;
``` 

could be condensed to 
`m_routeMonitor->deleteLater()`

src/platforms/windows/daemon/wireguardutilswindows.cpp
```diff
@@ -566,10 +584,23 @@ bool WireguardUtilsWindows::deleteRoutePrefix(const IPAddress& prefix) {
...
+  bool result = true;
+  for (const IPAddress& prefix : addresses) {
+    if (!m_routeMonitor->addExclusionRoute(prefix)) {
+      result = false;
+    }
+  }
+
+  // Permit LAN traffic through the firewall.
+  if (!WindowsFirewall::instance()->enableLanBypass(addresses)) {
+    result = false;
```
strseb, line 602: Should we be removing the exclusion route here, given if we end up here that route exist but might be blocked by the firewall?

strseb (approved):

PR #9686, opened 2024-06-24
VPN-6473: Add addon condition for OS version

mcleinman:
## Description

New addon conditions that looks at version of OS on a per-OS level.

Not sure what here triggered @flodolo 's review - let me know if I somehow messed with translations.

## Reference

VPN-6473

## Checklist
    
### DIFF
--- a/addons/message_survey_staging_30m/manifest.json
+++ b/addons/message_survey_staging_30m/manifest.json
@@ -6,6 +6,8 @@
   "type": "message",
   "conditions": {
     "env": "staging",
+    "min_os_version": {"ios": "14.0", "macos": "12.4"},
+    "max_os_version": {"ios": "24", "macos": "19.4"},
     "translation_threshold": 0,
     "trigger_time": 1800
   },
--- a/docs/Components/Addons/index.md
+++ b/docs/Components/Addons/index.md
@@ -87,6 +87,8 @@ Add-ons can enable and disable themselves using the `conditions` key in the mani
 | locales | An array of locales to be checked | Array of string | No | Yes |
 | min_client_version | The min client version | String | No | No |
 | max_client_version | The max client version | String | No | No |
+| min_os_version | The max OS version | String | No | No |
+| max_os_version | The max OS version | String | No | No |
 | platforms | An array of platforms to be checked | Array of string | No | No |
 | settings | An array of Condition Setting object. See below | Array of Condition Setting object | No | No |
 | trigger_time | A number identifying the number of seconds from the first execution of the client | Integer |  No | Yes |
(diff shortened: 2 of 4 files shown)


oskirby (commented):
I think the change is sound, but I am worried about how this is handled for backwards compatibility reasons. Looking at how conditions are evaluated (see `Addon::evaluateConditions()` it looks like clients that don't support a condition in their CPP code will report the addon as disabled.

IIUC, this means that setting a `min_os_version` or `max_os_version` condition would mean that a client on version < 2.23 would never see the addon. And the specific usecase we have for this change (hiding the update notification for iOS 14 users) would just wind up hiding the message for all users who aren't on the latest VPN release.

Is there something we could do in the Javascript side of things in `addons/message_update_v2.23/enable.js` to achieve the same effect without requiring an Addon API change?

oskirby:
> Is there something we could do in the Javascript side of things in `addons/message_update_v2.23/enable.js` to achieve the same effect without requiring an Addon API change?

I think that `api.env.osVersion` should contain the details we need in Javascript.

mcleinman:
> IIUC, this means that setting a `min_os_version` or `max_os_version` condition would mean that a client on version < 2.23 would never see the addon. And the specific usecase we have for this change (hiding the update notification for iOS 14 users) would just wind up hiding the message for all users who aren't on the latest VPN release.

Good catch, you're correct. I thought the opposite was true, and didn't test. 

> Is there something we could do in the Javascript side of things in `addons/message_update_v2.23/enable.js` to achieve the same effect without requiring an Addon API change?

I was hoping to do something more generalizable/easy for future use, but I think you're right - this is going to need to be custom JS whenever we need to do something like this. Will work that route, then close this ticket

addons/message_survey_staging_30m/manifest.json
```diff
@@ -6,6 +6,8 @@
   "type": "message",
   "conditions": {
     "env": "staging",
+    "min_os_version": {"ios": "14.0", "macos": "12.4"},
```
lesleyjanenorton, line 9: Just curious: was it not possible to hide/show the message via enable.js with `api.env.osVersion`? These changes look right to me, but would mean that if we ever wanted to show alternate UI in a single message depending on the osVersion we'd need to ship separate messages.

mcleinman: We can - and per conversation with Naomi above, I'm going to go that route - but my initial goal was to find a solution that was a condition, as it's easier to re-use (in my opinion) and less prone to bugs (if the condition code is solid). But given that old clients dismiss addons with conditions they don't expect, that doesn't seem like a workable approach.

mcleinman:
Question - should we keep the code for this new condition so that we can use it down the line (once a sufficient number of users are on a version that includes it)? Or just pull it out, figuring we can always do it via javascript?

mcleinman:
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9689 is up for review. Keeping this one up until we decide whether  want to add a condition or not. (I'm happy either way.)

PR #9689, opened 2024-06-25
VPN-6473: check for OS in addons

mcleinman:
## Description

Checking OS via a JS condition.

I don't believe this adds any translations, but perhaps those strings did something weird? I'm not sure why @flodolo got auto-added to this PR.

This PR overrides https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9686. However, I'm keeping that one open for now - I asked whether we want to modify that to add the condition in, so that we can use it in the future once enough users are on a version that include that condition.

Also, I might be doing over-defensive code with this check: `if (!osVersion || (typeof osVersion !== "string") || osVersion.length === 0)`. I don't write JS often; please let me know if there is a better way to do this.

## Reference

VPN-6473

## Checklist
    
### DIFF
--- a/addons/message_update_v2.23/manifest.json
+++ b/addons/message_update_v2.23/manifest.json
@@ -4,7 +4,8 @@
   "name": "Update to Mozilla VPN 2.23",
   "type": "message",
   "conditions": {
-    "max_client_version": "2.22.9"
+    "max_client_version": "2.22.9",
+    "javascript": "osCheck.js"
   },
   "javascript": {
     "enable": "enable.js"
--- a/addons/message_update_v2.23/osCheck.js
+++ b/addons/message_update_v2.23/osCheck.js
@@ -1,0 +1,23 @@
+// Disable on iOS 13 and earlier
+(function(api, condition) {
+  // First check for iOS...
+  const isIOS = (api.env.platform === "ios");
+  if (!isIOS) {
+    condition.enable();
+    return;
+  }
+
+  // ...then check for iOS 14 or later
+  const minVersion = 14;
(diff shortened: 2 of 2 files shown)


addons/message_update_v2.23/manifest.json
```diff
@@ -4,7 +4,8 @@
   "name": "Update to Mozilla VPN 2.23",
   "type": "message",
   "conditions": {
-    "max_client_version": "2.22.9"
+    "max_client_version": "2.22.9",
+    "javascript": "osCheck.js"
   },
   "javascript": {
     "enable": "enable.js"
```
lesleyjanenorton, line 11: Nice! Any reason to not just do this in enable.js where the other show/hide decision making happens?

lesleyjanenorton (approved):
Nice! Looks right, smells right. Left one question but R+

addons/message_update_v2.23/manifest.json
mcleinman: `enable` is run when the addon is enabled. We use it to modify the data in the addon based on conditions. 

There is an addon condition type (conditions are used to decide whether to even show the addon) of `javascript`, which can enable arbitrary javascript as part of a test as to whether to show the addon. It seems more appropriate to use a `condition` for this, and I'm not sure it's even possible to use an `enable` js snippet to halt usage of the addon. If you know of a way, I'm happy to explore.

mcleinman: Discussed in person, we're all comfortable w/ this. Going to merge.

PR #9724, opened 2024-07-10
chore: update Taskgraph and dependencies

ahal:
## Checklist
    
### DIFF
--- a/taskcluster/kinds/beetmover-ship/kind.yml
+++ b/taskcluster/kinds/beetmover-ship/kind.yml
@@ -19,7 +19,7 @@ task-defaults:
     from-deps:
         kinds: [beetmover-promote]
         group-by: all
-        set-name: false
+        set-name: null
         unique-kinds: false
     worker-type: beetmover
     worker:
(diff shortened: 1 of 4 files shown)


taskcluster/kinds/beetmover-ship/kind.yml
```diff
@@ -19,7 +19,7 @@ task-defaults:
     from-deps:
         kinds: [beetmover-promote]
         group-by: all
-        set-name: false
+        set-name: null
```
hneiva, line 22: I'm curious why do we need null here instead of false?

hneiva (approved):
LGTM

taskcluster/kinds/beetmover-ship/kind.yml
ahal: I think it was an oversight / bug in the refactor around this.

ahal:
Decision task is failing due to some difference in cached task routes for pull requests. I'll need to update the grants then it should start passing. Making this draft for now.

PR #9733, opened 2024-07-19
VPN-6495: fix iOS server switching bug

mcleinman:
## Description

A few lines were added in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9639, and they broke server switching on iOS. (Notably, it works on macOS - this seems to be iOS-specific.) While I'm deeply curious why an additional time setting the state to Connecting would cause this behavior (and would love to dig deeper to understand *why* this caused the behavior), the most time-judicious fix seems to be wrapping this new code in a platform flag to not run it on iOS.

## Reference

VPN-6495

## Checklist
    

lesleyjanenorton (approved):
👍

PR #9741, opened 2024-07-22
VPN-6252: Add Daemon Documentation

oskirby:
## Description
We have a need to better document how the daemon works, why we need it, and how it is expected to recover in the event of a crash. This PR should add a new document at `docs/Components/daemon.md` with the desired details.

## Reference
JIRA Issue: [VPN-6252](https://mozilla-hub.atlassian.net/browse/VPN-6252)

## Checklist
    
[VPN-6252]: https://mozilla-hub.atlassian.net/browse/VPN-6252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/Components/daemon.md
+++ b/docs/Components/daemon.md
@@ -30,7 +30,7 @@ and describe a common set of commands:
    peer has been successfully established. The public key of the Wireguard peer is provided as an argument to this
    signal.
 
- - `disconnected`: An asychronous signal sent from the daemon to inform the GUI client that the VPN has been terminated
+ - `disconnected`: An asynchronos signal sent from the daemon to inform the GUI client that the VPN has been terminated
    and the connection to all Wireguard peers has been stopped.
 
 ## Platform Implementations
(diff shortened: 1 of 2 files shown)


docs/Components/daemon.md
```diff
@@ -0,0 +1,149 @@
...
+[Wireguard NT](https://github.com/WireGuard/wireguard-nt) kernel driver is installed by the windows service. This driver
+is instantiated when bringing up the VPN tunnel to handle the connection.
+
+Additionally, a [split-tunneling](https://github.com/mullvad/win-split-tunnel) driver may also be provided, which can be used
+by the VPN to allow some applications to access the network directly and bypass the VPN. The daemon will check for this driver
+at startup, and register it for split tunneling if possible.
+
+### Android daemon???
+
+TODO: How does Android work, exactly?
```
strseb, line 111: The Daemon is an [Android Service](https://developer.android.com/develop/background-work/services) communicating using [Bindings ](https://developer.android.com/develop/background-work/services/bound-services) with the Client (or widgets). 

The protocol is similar to the desktop daemons as we exchange json blobs but the format on the wire is `{requestType:int,data:json}` (...we could actually think about using the desktop protocol at some point)

To activate the daemon will request a raw handle of a TUN device using [Android's VPN API](https://developer.android.com/reference/android/net/VpnService.Builder). It load's wireguard-go into it's process and calls into wg-go with the handle and the config, which will then spawn a goroutine handling traffic.

strseb (approved):

docs/Components/daemon.md
oskirby: Thank you!

PR #9752, opened 2024-07-27
Add some help for fedora installation (from Fedora40 plasma)

rloutrel:
## Description

    Add information for Fedora users: since no official client package is provided, a lot will have to go through those steps.
    
    I am still facing issues at the compiling step, but at least dependencies are installed. I will update once compile/install is working.

## Reference

    My experience trying to install it

## Checklist
    
### DIFF
--- a/docs/Building/linux.md
+++ b/docs/Building/linux.md
@@ -40,6 +40,11 @@ Configure
 
     cmake -S . -B build -DCMAKE_PREFIX_PATH=$(Qt install path)/gcc_64/lib/cmake/
 
+> hint about $(Qt install path): If you forgot where you installed it, you can use the following
+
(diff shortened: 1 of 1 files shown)


rloutrel:
About the remaining issues:
```shell
$ cmake --build build -j$(nproc)
gmake: Makefile: No such file or directory
gmake: *** No rule to make target 'Makefile'.  Stop.
```

Any clues? I am a bit lost. I understand, that it is something about the makefile and the target is not detected properly, but if some expert could point out where to search, it would really help.

strseb:
Maybe the configuration step did fail (`cmake -S . -B build -D_SOME_ARGS`) ?
We don't have a "makefile" target ourselves so i think this might be connected to the cmake-generator.

PR #9757, opened 2024-07-31
VPN-6523: Fix Android LAN exclusions

oskirby:
## Description
In PR #9674 we tried to improve the LAN exclusion implementation for Windows and Linux. However, this introduced a bug in Android where the LAN exclusion calculation would cause routes to the internal Mulvad network to get lost. This causes the VPN's connectivity checks to fail, and results in the client entering the `No Signal` state shortly after connecting.

The cause of the bug is that `IPAddress::excludeAddresses()` doesn't handle routes nested inside an excluded route. For example, `excludeAddresses({"0.0.0.0/0", "10.64.0.1/32"},{"10.0.0.0/8"})` will ignore the route to `10.64.0.1` because it is excluded by `10.0.0.0/8` despite having a longer prefix length.  

## Reference
JIRA issue: [VPN-6523](https://mozilla-hub.atlassian.net/browse/VPN-6523)
Previous PR: #9674

## Checklist
    
[VPN-6523]: https://mozilla-hub.atlassian.net/browse/VPN-6523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/android/androidcontroller.cpp
+++ b/src/platforms/android/androidcontroller.cpp
@@ -162,14 +162,24 @@ void AndroidController::activate(const InterfaceConfig& config,
   jServer["publicKey"] = config.m_serverPublicKey;
   jServer["port"] = (double)config.m_serverPort;
 
-  // TODO: This could be done better with VpnService.Builder.excludeRoute()
-  // we just need a way to get this config down that far.
-  const auto localRoutes = Controller::getExcludedIPAddressRanges().flatten();
-  const auto fullAllowedIPs =
-      IPAddress::excludeAddresses(config.m_allowedIPAddressRanges, localRoutes);
+  // This could be done better with VpnService.Builder.excludeRoute()
+  // but that requires API level 33 (Android 13).
+  const auto lanRoutes = Controller::getExcludedIPAddressRanges();
   QJsonArray jAllowedIPs;
-  foreach (auto item, fullAllowedIPs) {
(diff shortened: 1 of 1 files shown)


src/platforms/android/androidcontroller.cpp
```diff
@@ -162,14 +162,24 @@ void AndroidController::activate(const InterfaceConfig& config,
   jServer["publicKey"] = config.m_serverPublicKey;
   jServer["port"] = (double)config.m_serverPort;
 
-  // TODO: This could be done better with VpnService.Builder.excludeRoute()
-  // we just need a way to get this config down that far.
-  const auto localRoutes = Controller::getExcludedIPAddressRanges().flatten();
-  const auto fullAllowedIPs =
-      IPAddress::excludeAddresses(config.m_allowedIPAddressRanges, localRoutes);
+  // This could be done better with VpnService.Builder.excludeRoute()
```
oskirby, line 165: I actually went down the rabbit hole of implementing this with `VpnService.Builder.excludeRoute()` and it worked great, and only afterwards did I realize that it requires a higher API level than we ship with. So this implementation here is just crudely working around the issue, but in a way that doesn't require a change to our minimum Android version.

strseb (approved):

mcleinman (approved):
Questions answered via real time convo. Thanks!

PR #9809, opened 2024-08-26
VPN-6448 try to scheudle a reboot on w10 (2.24 only)

strseb:
## Description
Given QA's ticket mentions they uninstall the old version, using wix_upgrade detected is not sufficient. So let's just "ask" w10 users to reboot after installation :)

### DIFF
--- a/windows/installer/MozillaVPN.wxs
+++ b/windows/installer/MozillaVPN.wxs
@@ -168,9 +168,16 @@
     <CustomAction Id="LaunchVPNAfterUpdate" Impersonate="yes" FileKey="MozillaVPNExecutable" ExeCommand="ui --updated" Return="asyncNoWait" />
     <InstallExecuteSequence>
    <!-- This will schedule a reboot only if the OS is Windows 10 -->
-      <ScheduleReboot After="InstallFinalize">VersionNT = 603</ScheduleReboot>
-      <Custom Action="LaunchVPNFirstExecution" After="InstallFinalize">NOT WIX_UPGRADE_DETECTED AND NOT(VersionNT = 603)</Custom>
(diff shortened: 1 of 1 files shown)


windows/installer/MozillaVPN.wxs
```diff
@@ -167,8 +179,17 @@
     <CustomAction Id="LaunchVPNFirstExecution" Impersonate="yes" FileKey="MozillaVPNExecutable" ExeCommand="" Return="asyncNoWait" />
     <CustomAction Id="LaunchVPNAfterUpdate" Impersonate="yes" FileKey="MozillaVPNExecutable" ExeCommand="ui --updated" Return="asyncNoWait" />
     <InstallExecuteSequence>
-       <Custom Action="LaunchVPNFirstExecution" After="InstallFinalize">NOT WIX_UPGRADE_DETECTED</Custom>
-       <Custom Action="LaunchVPNAfterUpdate" After="InstallFinalize">WIX_UPGRADE_DETECTED</Custom>
+      <!-- This will schedule a reboot only if the OS is Windows 10 -->
+      <ScheduleReboot After="InstallFinalize">
```
oskirby, line 183: As I understand this, it will schedule a reboot after both fresh installs and upgrades even though we only need it on upgrades. It's not really the worst outcome, just inconvenient.

oskirby (approved):
I guess we don't technically need to solve this now, but as-written this installer file will always reboot for Windows 10 users. However, in the future when users are upgrading say from 2.24 -> 2.25 that reboot is no longer required, but the installer will still force it.

In order to minimize inconvenience to our users, it would be nice if this reboot could be made conditional upon the previously installed VPN version.

strseb:
> I guess we don't technically need to solve this now, but as-written this installer file will always reboot for Windows 10 users. However, in the future when users are upgrading say from 2.24 -> 2.25 that reboot is no longer required, but the installer will still force it. 

Ah dang i filed it against main, i actually just wanted this to land on the 2.24 branch.

PR #9812, opened 2024-08-27
Update Cargo/Rust to 1.75.0

oskirby:
## Description
I have noticed a bunch dependebot updates are starting to fail because we are building on too old of a Rust version (1.69) which used to be the most recent thing available on Ubuntu focal. Well, that's no longer the case and we now have 1.75.0 available as a security update.

Let's bump the versions and see what happens!

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
Uhoh. Turns out the Apple `rust-std` packages don't install on Windows anymore.

PR #9815, opened 2024-08-28
VPN-6569: Compute balrog installer hashes as QByteArrays

oskirby:
## Description
Balrog computes its file hashes by converting the hash to a hex string and then comparing it to the hash string received from Balrog. While this works, it is sloppy and it means that hash comparisons can fail depending on case sensitivity. The better approach is to compare hashes as a `QByteArray` so that the string case doesn't matter.

## Reference
JIRA issue: [VPN-6569](https://mozilla-hub.atlassian.net/browse/VPN-6569)

## Checklist
    
[VPN-6569]: https://mozilla-hub.atlassian.net/browse/VPN-6569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/update/balrog.cpp
+++ b/src/update/balrog.cpp
@@ -292,7 +292,8 @@ bool Balrog::processData(Task* task, const QByteArray& data) {
     return false;
   }
 
-  QString hashValue = obj.value("hashValue").toString();
+  QString hashString = obj.value("hashValue").toString();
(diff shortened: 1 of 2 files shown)


mcleinman (approved):
Thank you!

PR #9817, opened 2024-08-28
VPN-2876: Stop loading screens persistently

lesleyjanenorton:
## Description

Here we:
- Stop loading the Home, Settings, and Messages screens "Persistently" in favor of loading them "Temporarily"
- Fix UI flickering in ScreenHome when the Screen loads and which becomes much more noticeable when the view is loaded every time the home button is clicked.  
  - Annoyingly: I can't say with 100% confidence _why_ these changes, mostly to how we name states and assign property changes, fixes the flicker but I suspect that assigning state changes conditionally with `when: someConditionIsTrue` was causing the FOUC.  
  - Also annoyingly: The main toggle still fades in somewhat noticeably. This is fixable but less trivial to fix and, given the a11y wins we gain by doing this, is I think acceptable. 

<table>
<th>Before flicker fixes <br>(Note the toggle wiggliness)</th><th>After flicker fixes</th>
<tr>
<td>

https://github.com/user-attachments/assets/7ec22884-41ca-48b2-9b9b-d6a4785e4d9c

</td>
<td>

https://github.com/user-attachments/assets/c88e2991-dd39-4aca-a1fc-a0293ae7d7e1

</td>
</tr>
</table>

## Reference

VPN-2876

## Checklist
    
### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -1507,7 +1507,7 @@ void MozillaVPN::registerNavigatorScreens() {
       });
 
   Navigator::registerScreen(
-      MozillaVPN::ScreenHome, Navigator::LoadPolicy::LoadPersistently,
+      MozillaVPN::ScreenHome, Navigator::LoadPolicy::LoadTemporarily,
       "qrc:/qt/qml/Mozilla/VPN/screens/ScreenHome.qml",
       QVector<int>{App::StateMain}, [](int*) -> int8_t { return 99; },
       []() -> bool { return false; });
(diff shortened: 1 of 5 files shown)


mcleinman (approved):
Thanks! Looks great.

I believe that loadPersistently is now no longer used anywhere in the app. We could pull it out, but I'm fine leaving it (at least for now) in case we want to go back to using the old method at some point.

`testNavigator.cpp` in the unit tests use loadPersistently - I'm wondering if we should update this, as we don't load persistently anywhere else in the app at this point.

PR #9820, opened 2024-08-28
WIP: Repackage .deb builds as mozillavpn-beta

oskirby:
## Description
We have had much recurring pain trying to upload release candidates to the staging APT repository, and it would be nice if we could find a better way to package and upload release candidates. This PR attempts to accomplish just that. We attempt this by doing the following:

1. We add repackage jobs to unpack a `.deb` file, edit its `control` file and repack it. There are two such jobs:
   - `repackage-deb-beta`: Sets `Replaces` and `Provides` to `mozillavpn` and renames the package to `mozillvpn-beta`. This ensures that it will be considered an alternate version of `mozillavpn` as far as the Debian packaging world is concerned. 
   - `repackage-deb-release`: Changes the package version to the contents of `version.txt`, thereby stripping out the build suffix (eg: `2.26.0~build20241129` becomes `2.26.0`) and we also add the version to the filename.
2. We add a transform that renames a taskcluster `release-artifact` to append a version number. This should allow the release jobs to pick up the repackaged `.deb` files rather than the one output by the build tasks.
3. We add a transform to the beetmover tasks so that promotions depend on `repackage-deb-beta` and shipping depends on `repackage-deb-release`.
4. We modify the `beetmover-apt` task so that it pushes to the release bucket for both `promote` and `ship` tasks. This was done by copying the `in_production` check from `beetmover-promote` and this is where I get a bit fuzzy on what the correct thing to do is. 

Thus we wind up with a chain of tasks that should go something like this:
```
build-linux64/release-deb --+--> repackage-deb-beta -> beetmover-promote -> beetmover-apt
                            |
                            +--> repackage-deb-release -> beetmover-ship -> beetmover-apt
```

## Reference
JIRA Issue: [RELENG-1062](https://mozilla-hub.atlassian.net/browse/RELENG-1062)

## Checklist
    
[RELENG-1062]: https://mozilla-hub.atlassian.net/browse/RELENG-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/taskcluster/mozillavpn_taskgraph/transforms/add_version.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/add_version.py
@@ -10,27 +10,23 @@ transforms = TransformSequence()
 
 # Append the VPN client version to the filenames of the release-artifacts.
 @transforms.add
-def append_version(config, tasks):
+def add_version(config, tasks):
     # Get the version number from version.txt
     cwd = os.path.dirname(os.path.realpath(__file__))
     with open(os.path.join(cwd, '..', '..', '..', 'version.txt'), 'r') as fp:
         version = fp.readline().strip()
     
-    # If this is a beetmover promotion job, change the version to "beta" instead.
+    # If this is a beetmover promotion job, change the version to "beta" instead
     if config.kind == "beetmover-promote":
         version = "beta"
 
     for task in tasks:
-        print(f"DEBUG: append_version on {task['name']}")
         if task["name"] not in ("deb-release", "linux64-deb"):
             yield task
             continue
 
         # Get the release artifacts and append -<version> to their filenames.
-        print(f"DEBUG: task keys: {task.keys()}")
         for artifact in task["attributes"]["release-artifacts"]:  
-            print(f"DEBUG: release-artifact {artifact['name']}")
-            print(f"DEBUG: release-artifact {artifact['path']}")
             if "name" in artifact:
                 root, ext = os.path.splitext(artifact["name"])
                 artifact["name"] = root + "-" + version + ext
--- a/taskcluster/mozillavpn_taskgraph/transforms/add_version.py
+++ b/taskcluster/mozillavpn_taskgraph/transforms/add_version.py
@@ -15,10 +15,14 @@ def append_version(config, tasks):
     cwd = os.path.dirname(os.path.realpath(__file__))
     with open(os.path.join(cwd, '..', '..', '..', 'version.txt'), 'r') as fp:
         version = fp.readline().strip()
+    
+    # If this is a beetmover promotion job, change the version to "beta" instead.
+    if config.kind == "beetmover-promote":
+        version = "beta"
 
     for task in tasks:
         print(f"DEBUG: append_version on {task['name']}")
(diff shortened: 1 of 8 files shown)


taskcluster/mozillavpn_taskgraph/transforms/add_version.py
```diff
@@ -0,0 +1,37 @@
...
+@transforms.add
+def add_version(config, tasks):
+    # Get the version number from version.txt
+    cwd = os.path.dirname(os.path.realpath(__file__))
+    with open(os.path.join(cwd, '..', '..', '..', 'version.txt'), 'r') as fp:
+        version = fp.readline().strip()
+    
+    # If this is a beetmover promotion job, change the version to "beta" instead
+    if config.kind == "beetmover-promote":
+        version = "beta"
```
bhearsum, line 21: Does `apt upgrade` still work when the version doesn't change? (Do we care if it doesn't?)

taskcluster/mozillavpn_taskgraph/transforms/beetmover_apt.py
```diff
@@ -43,6 +43,11 @@ def get_gcs_sources(dependent_task):
 
 @transforms.add
 def beetmover_apt(config, tasks):
+    is_production = (
+        config.params["level"] == "3" and config.params["tasks_for"] == "action"
+    )
+    bucket = "release" if is_production else "dep"
```
bhearsum, line 49: Just to be 100% clear: the idea is that both the `promote` and `apt` tasks will push to the exact same debian repo (but with different package names)? If so, this block looks like it ought to be fine. (We'll be able to verify the `dep` path in the staging repo, but the `release` one obviously can only be fully tested in a real release.)

bhearsum (commented):
@jcristau should probably have a look at this at some point, as he's our resident debian expert.

I'm going to help get you set up with the staging repo to test this in any case.

taskcluster/mozillavpn_taskgraph/transforms/beetmover_apt.py
oskirby: So, my understanding is that `beetmover-promote` and `beetmover-ship` send packages to the production archive.mozilla.org site whenever they are triggered by shipit and we want to follow the same logic for the `beetmover-apt` task.

The promotion and shipping phases on archive.mozilla.org are distinguished by different directory paths (eg: promotion puts artifacts into the `candidates` directory). The objective of this PR is to distinguish promotion and shipping in the APT repository by changing the package name.

taskcluster/mozillavpn_taskgraph/transforms/add_version.py
bhearsum: ah, right, sure, that's fine

taskcluster/mozillavpn_taskgraph/transforms/beetmover_apt.py
bhearsum: Are we talking about uploading to archive.m.o, the debian repository, or both? (Both are handled by beetmover, but they are distinct things can be done or not done independently AFAIK.)

oskirby:
We are going to abandon this work - we make do without.

PR #9821, opened 2024-08-28
VPN-6488: update string for iOS exclude LAN feature

mcleinman:
## Description

Updated strings per ticket and conversation with Santiago.

## Reference

VPN-6488

## Checklist
    
### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -356,7 +356,7 @@ settings:
 localNetworkAccess:
   labelTitle: Exclude local network traffic
   subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
-  vpnMustBeOffTip: VPN must be off to edit “Exclude local network traffic”
+  disabledSubLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers. The VPN must be off to edit this setting.
 
 dnsOverwriteDialog:
   titleDNS: Override DNS settings?
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -354,9 +354,9 @@ settings:
     comment: Button label to clear all apps from the exclusions list
 
 localNetworkAccess:
-  label: Local network access
+  labelTitle: Exclude local network traffic
   subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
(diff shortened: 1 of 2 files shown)


src/translations/strings.yaml
```diff
@@ -354,9 +354,9 @@ settings:
     comment: Button label to clear all apps from the exclusions list
 
 localNetworkAccess:
-  label: Local network access
+  labelTitle: Exclude local network traffic
   subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
-  vpnMustBeOff: VPN must be off to edit “Local network access”
+  vpnMustBeOffTip: VPN must be off to edit “Exclude local network traffic”
```
flodolo, line 359: Wouldn't we better off with something more generic, e.g. `VPN must be off to edit this setting`?

flodolo (approved):

strseb (approved):

src/translations/strings.yaml
mcleinman: Good suggestion. I'll defer to Santiago for this one, as he crafted the language here. He's off this week, but I'll ping him next week - and will hold this PR until he's back.

mcleinman: Santiago wanted to remove this warning completely, and instead put the text in the label. 

@flodolo  Please let me know if the preferred way to have this translated is two separate strings that we sometimes combine, or the way that it currently is written.

That is, should the 2 strings be:
```
subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
disabledSubLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers. The VPN must be off to edit this setting.
```
(and we show either `subLabel` or `disabledSubLabel`)
or
```
subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
disabledRequirement: The VPN must be off to edit this setting.
```
(and we either show `subLabel` or `subLabel + disabledRequirement`)

Thanks!

flodolo: Combining would only work if you have the 2 strings clearly separated. In this case, you would need to know how to glue together 2 pieces based on the language.

flodolo (approved):

src/translations/strings.yaml
mcleinman: That was my gut, so seems like we're good with how it's written. Thank you!

PR #9823, opened 2024-08-29
EXT: Add the current client version

strseb:
## Description
The Extension should know the Client Version, as we will have a "minimum" client version needed to run the extension.


mcleinman (approved):
👍🏻

PR #9841, opened 2024-09-06
Add task to cache cargo crates and use it for MacOS/Windows builds

oskirby:
## Description
The rust crates used by our project are a fairly sizeable chunk of data to download on each job, and for the most part they are unlikely to change. This makes them well suited to be cached as a toolchain task. Furthermore, the existence of such a toolchain means that some of our build jobs could build directly out of git rather than needing to wait on `build-source/vpn` to download and vendor the crates for us.

Some notes:
- Android (and presumably iOS) can't build with vendored crates right now as the `embedded-uniffi-bindgen` runs into crate version errors.
- Linux builds are still relying on `build-source/vpn` because they also need some golang vendoring, and we hope to mirror the offline build environment for PPAs.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
Hmm, turns out the Android tasks don't build when the crates are vendored. It seems attempting to run the `embedded-uniffi-bindgen` uses Glean's `Cargo.lock` instead of the VPN's version and it falls apart due to version mismatches.

mcleinman (approved):
👍🏻 Thank you!

PR #9843, opened 2024-09-06
Nit: Add Commands to help troubleshoot errors

strseb:
## Description
We have a few states where the error can only properly be solved by the VPN client. i.e Not Subbed, not authed. 
So let's give the extension the ability to summon the vpn window to the foreground to address those issues :) 

## Demo 
https://github.com/user-attachments/assets/0d1a12fc-b3b9-4ce5-ae86-cde5769f861a


lesleyjanenorton (approved):
👍

PR #9865, opened 2024-09-12
Avoid Start loops for localproxy

strseb:
Ohno who wrote all that? 
4207562-spider-point

## Description
The biggest culprit for an endless loop is this
```c++
mCrashSignal = QObject::connect(
      mCurrentProcess, &QProcess::finished, [this](){ ..., start()}
```

So whenever the proxy would exit, we would try to reboot it, which sounds sensible. *IN THEORY* 

Except that i did forget to remove that listener when we actually want to stop it. i.e we switch servers or disconnect, or extra fun in `QObject::connect(mCurrentProcess, &QProcess::errorOccurred,`. 

This means on any error we call  
```
QProcess::errorOccurred -> calls stop() -> triggers QProcess::finished -> calls start() -> which given we have a m_process calls kill() -> WHICH TRIGGERS QProcess::errorOccurred  -> which calls stop() -> ....
```

- so stop now actually removes signals from the about to be killed process. No loops. 
- The unexpected kill check `&QProcess::finished` schedules the start by putting an event on the loop, which means qt will (as its then marked with delete later) cleanup the old process first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ linux/netfilter/vendor/
 .tmp
 .env
 .venv
+.cache/clangd
 __pycache__
 lottie.mjs
 qml_lint_result.txt
--- a/src/proxycontroller.cpp
+++ b/src/proxycontroller.cpp
@@ -5,18 +5,20 @@
 
 #include "proxycontroller.h"
 
+#include <QObject.h>
+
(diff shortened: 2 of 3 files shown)


lesleyjanenorton (approved):
🪄✨

PR #9866, opened 2024-09-12
VPN-6475: Parse and log Guardian errors

oskirby:
## Description
Sometimes it can be difficult to diagnose the cause of an error caused by an interaction with Guardian, or any other Mozilla web service. This is made extra tricky when the error handling in the `NetworkRequest` class treats the response body as sensitive and scrubs it from the logs.

To try and make diagnosis a little better, let's make an attempt to parse the error responses and log them if they happen to be parseable as such. To my knowledge, there are no API errors we get back from Guardian which contain sensitive information.

The format of these errors ought to match the [ApiErrorJson](https://github.com/mozilla-services/guardian-website/blob/2eb7ad4188db6bd5d4f48a15dd2768f89c143bc6/server/lib/errors.ts#L77-L82) type in the Guardian website:
```ts
interface ApiErrorJson {
  code: number;
  errno: number;
  error: string;
}
```

## Reference
JIRA Issue: [VPN-6475](https://mozilla-hub.atlassian.net/browse/VPN-6475)

## Checklist
    
### DIFF
--- a/src/models/apierror.h
+++ b/src/models/apierror.h
@@ -1,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef APIERROR_H
+#define APIERROR_H
+
+#include <QString>
+#include <QObject>
+
+class QByteArray;
+
+class ApiError final {
+  Q_GADGET
+
+ public:
+  ApiError() {};
+  ApiError(const ApiError& other);
+  ApiError& operator=(const ApiError& other);
+
+  // Error codes transcribed from guardian-website/server/lib/errors.ts
+  enum ErrorNumbers : int {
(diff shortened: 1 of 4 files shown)


src/models/apierror.h
```diff
@@ -0,0 +1,78 @@
...
+    InvalidTokenError = 147,
+    InvalidSkuError = 148,
+    MissingTokenError = 149,
+    MissingSkuError = 150,
+    FxaTokenValidationError = 151,
+    FxaPlansError = 152,
+    FxaSupportTicketError = 153,
+    FxaSupportTicketNotReturned = 154,
+    SupportTicketIssueTextSizeError = 155,
+    SupportTicketMissingLogsError = 157,
```
mcleinman, line 55: Missing 156 (`SupportTicketMissingIssueTextError`) and 160 (`AdjustMissingOrInvalidMethod`)

mcleinman (changes requested):
I think we need two more codes. Otherwise, looks great!

mcleinman (approved):
Thanks!

PR #9867, opened 2024-09-13
Add extra runtime dependencies for static builds

oskirby:
## Description
One finding that came out of the experimental [functional testing](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9863) pull request, is that we are actually missing a bunch of runtime package dependencies for the Linux static build. In a traditional dynamically linked build, these are included as dependencies of the Qt libraries, but after they are stripped out in the static build, there is nothing that requires them to be installed during package installation.

It's not particularly noticeable on your typical Linux installation because these are brought in by the desktop environment, but in a headless environment (such as a Docker container) they are typically missing.

## Reference
Found by PR #9863

## Checklist
    

mcleinman (approved):
👍🏻

PR #9873, opened 2024-09-13
VPN-6512: Update Apple in app purchase to use StoreKit2

mcleinman:
## Description

StoreKit (formerly called StoreKit2)  has been the modern way to build In App Purchase for a few years now. And Apple has announced Original StoreKit is deprecated after iOS 18.0. We needed to update the library we use, which is done in this PR. 

StoreKit2 is only available on iOS 15 and greater. Since it was trivial to maintain support for iOS 14 and earlier by using the original StoreKit APIs, there is forking logic all over this.

For a swift-only app, StoreKit2 is fairly straightforward. However, that doesn't work for our app. This is a mix of Objective C++ and Swift:
- Callbacks into our Qt code shared across platforms cannot be written in Swift, so those need to be in C++ or Objective C++.
- StoreKit2 is a Swift-only API, so code that interacts with it must be written in Swift. (StoreKit1 was available both in Swift and Objective-C, and given our app's architecture we use the Obj-C APIs.)
  
## Testing

I've tested:
- The happy path (subscribe in app)
- Cancelling the transaction when the IAP sheet comes up, rather than completing the purchase
- An interrupted transaction (interrupted by new App Store rules that the user must agree to)
- Restoring an Apple IAP subscription purchased outside the app
- The old, StoreKit1 happy path
- probably a few more things I'm not remembering

## Reference

VPN-6512

## Checklist
    
### DIFF
--- a/src/constants.cpp
+++ b/src/constants.cpp
@@ -94,6 +94,7 @@ QString Constants::apiUrl(ApiEndpoint endpoint) {
 #endif
 #ifdef MZ_IOS
       {ApiEndpoint::PurchasesIOS, "/api/v1/vpn/purchases/ios"},
+      {ApiEndpoint::PurchasesIOSv2, "/api/v2/vpn/purchases/ios"},
 #endif
 #ifdef MZ_WASM
       {ApiEndpoint::PurchasesWasm, "/api/v1/vpn/purchases/wasm"},
--- a/src/platforms/ios/iosiaphandler.h
+++ b/src/platforms/ios/iosiaphandler.h
@@ -5,6 +5,7 @@
 #ifndef IOSIAPHANDLER_H
 #define IOSIAPHANDLER_H
 
+#include "Mozilla-Swift.h"
 #include "purchaseiaphandler.h"
 
 class IOSIAPHandler final : public PurchaseIAPHandler {
@@ -18,7 +19,8 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
 
  public slots:
   void productRegistered(void* product);
-  void processCompletedTransactions(const QStringList& ids);
+  void processCompletedTransactions(const QStringList& ids,
+                                    const QString transactionIdentifier);
   void noSubscriptionFoundError();
 
  protected:
@@ -28,6 +30,12 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
  private:
   void* m_delegate = nullptr;
   int discountToDays(void* discount);
+  // This is a void (and cast to InAppPurchaseHandler as needed) as the
+  // InAppPurchaseHandler needs to store Product, which is only available in iOS
+  // 15+. So the entire class must be marked as only available in iOS 15. But we
+  // can't mark this variable as only available in iOS 15+ here. Hence, we have
+  // this variable as a `void`, and cast it to InAppPurchaseHandler as needed.
+  void* swiftIAPHandler;
 };
 
 #endif  // IOSIAPHANDLER_H
--- a/src/platforms/ios/iosiaphandler.h
+++ b/src/platforms/ios/iosiaphandler.h
@@ -19,6 +19,7 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
 
  public slots:
   void productRegistered(void* product);
+  void processCompletedTransactions(const QStringList& ids);
   void processCompletedTransactions(const QStringList& ids,
                                     const QString transactionIdentifier);
   void noSubscriptionFoundError();
--- a/src/platforms/ios/iosiaphandler.h
+++ b/src/platforms/ios/iosiaphandler.h
@@ -36,7 +36,7 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
   // 15+. So the entire class must be marked as only available in iOS 15. But we
   // can't mark this variable as only available in iOS 15+ here. Hence, we have
   // this variable as a `void`, and cast it to InAppPurchaseHandler as needed.
-  void* swiftIAPHandler;
+  void* swiftIAPHandler = nullptr;
 };
 
 #endif  // IOSIAPHANDLER_H
--- a/src/platforms/ios/iosiaphandler.mm
+++ b/src/platforms/ios/iosiaphandler.mm
@@ -203,7 +203,25 @@ IOSIAPHandler::IOSIAPHandler(QObject* parent) : PurchaseIAPHandler(parent) {
   MZ_COUNT_CTOR(IOSIAPHandler);
 
   if (@available(iOS 15, *)) {
-    swiftIAPHandler = [[InAppPurchaseHandler alloc] init];
+    swiftIAPHandler = [[InAppPurchaseHandler alloc]
+        initWithErrorCallback:^(void) {
+          logger.debug() << "Subscription error with StoreKit2.";
+          QMetaObject::invokeMethod(this, "stopSubscription", Qt::QueuedConnection);
+          QMetaObject::invokeMethod(this, "subscriptionCanceled", Qt::QueuedConnection);
+        }
+        successCallback:^(NSString* productIdentifier, NSString* transactionIdentifier) {
+          if (App::instance()->userAuthenticated()) {
+            logger.debug() << "Subscription completed with StoreKit2. Starting validation.";
+            QMetaObject::invokeMethod(
+                this, "processCompletedTransactions", Qt::QueuedConnection,
+                Q_ARG(QStringList, {QString::fromNSString(productIdentifier)}),
+                Q_ARG(QString, QString::fromNSString(transactionIdentifier)));
+          } else {
+            logger.debug() << "Subscription completed with StoreKit2. User not signed in.";
+            QMetaObject::invokeMethod(this, "stopSubscription", Qt::QueuedConnection);
+            QMetaObject::invokeMethod(this, "subscriptionCanceled", Qt::QueuedConnection);
+          }
+        }];
   } else {
     m_delegate = [[IOSIAPHandlerDelegate alloc] initWithObject:this];
     [[SKPaymentQueue defaultQueue]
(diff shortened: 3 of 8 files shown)


src/constants.cpp
```diff
@@ -94,6 +94,7 @@ QString Constants::apiUrl(ApiEndpoint endpoint) {
 #endif
 #ifdef MZ_IOS
       {ApiEndpoint::PurchasesIOS, "/api/v1/vpn/purchases/ios"},
+      {ApiEndpoint::PurchasesIOSv2, "/api/v2/vpn/purchases/ios"},
```
oskirby, line 97: Is there a corresponding PR to the guardian website that I can compare against?

src/platforms/ios/iosiaphandler.h
```diff
@@ -28,6 +31,12 @@ class IOSIAPHandler final : public PurchaseIAPHandler {
  private:
   void* m_delegate = nullptr;
   int discountToDays(void* discount);
+  // This is a void (and cast to InAppPurchaseHandler as needed) as the
+  // InAppPurchaseHandler needs to store Product, which is only available in iOS
+  // 15+. So the entire class must be marked as only available in iOS 15. But we
+  // can't mark this variable as only available in iOS 15+ here. Hence, we have
+  // this variable as a `void`, and cast it to InAppPurchaseHandler as needed.
+  void* swiftIAPHandler;
```
oskirby, line 39: Can we simply use a forward class declaration of in this case? IIUC, it's fine if the type doesn't get defined so long as it's used as a pointer. eg: 

```
class InAppPurchaseHandler;

class IOSIAPHandler {
    // ... some stuff ...
    InAppPurchaseHandler* swiftIAPHandler = nullptr;
};
```

oskirby: Oh, and we should also provide an initial value here of `nullptr`

src/platforms/ios/iosiaphandler.mm
```diff
@@ -214,6 +237,9 @@ - (void)paymentQueueRestoreCompletedTransactionsFinished:(nonnull SKPaymentQueue
 
   [delegate dealloc];
   m_delegate = nullptr;
+  if (@available(iOS 15, *)) {
+    swiftIAPHandler = nullptr;
```
oskirby, line 241: Do we need to invoke a destructor or `free()` kind of method here, or does Objective-C do that automatically?

src/platforms/ios/iosiaphandler.mm
```diff
@@ -222,26 +248,81 @@ - (void)paymentQueueRestoreCompletedTransactionsFinished:(nonnull SKPaymentQueue
     productIdentifiers = [productIdentifiers setByAddingObject:product.m_name.toNSString()];
   }
 
-  logger.debug() << "We are about to register" << [productIdentifiers count] << "products";
-
-  SKProductsRequest* productsRequest =
-      [[SKProductsRequest alloc] initWithProductIdentifiers:productIdentifiers];
+  logger.debug() << "We are about to register" << [productIdentifiers count] << "products...";
```
oskirby, line 251: nit: I don't like splitting messages across multiple log lines. I would prefer that we just have a single message in each of the control flow paths.

src/tasks/purchase/taskpurchase.cpp
```diff
@@ -22,9 +22,11 @@ Logger logger("TaskPurchase");
 
 #ifdef MZ_IOS
 // static
-TaskPurchase* TaskPurchase::createForIOS(const QString& receipt) {
+TaskPurchase* TaskPurchase::createForIOS(const QString& receipt,
+                                         bool isOlderOS) {
```
oskirby, line 26: The meaning of `older` is going to change over time, I would recommend that we provide a version number here and the implementation of this method can decide which API to use.

src/tasks/purchase/taskpurchase.h
```diff
@@ -52,7 +52,12 @@ class TaskPurchase final : public Task {
  private:
   Op m_op;
 #ifdef MZ_IOS
-  QString m_iOSReceipt;
+  // m_iOSData is the originalTransactionId for iOS 15+, and the receipt for
+  // earlier versions of iOS
+  QString m_iOSData;
+  // `@available(iOS 15, *)` is not available in C++, so this value must be
+  // passed in and saved
+  bool isOlderOS;
```
oskirby, line 60: nit: our coding convention prefixes member variables with `m_`, and I would recommend following that trend (or abandoning it entirely if we feel it's silly). I don't have a strong preference, but I just like consistency :)

src/platforms/ios/iosiaphandler.swift
```diff
@@ -0,0 +1,214 @@
...
+
+          let numberOfMonthsInSubscription = monthsInSubscription(for: subscription.subscriptionPeriod.unit, with: subscription.subscriptionPeriod.value)
+          let monthlyPrice: Double = (product.price as NSDecimalNumber).doubleValue / Double(numberOfMonthsInSubscription)
+
+          let numberFormatter = NumberFormatter()
+          numberFormatter.numberStyle = .currency
+          numberFormatter.locale = product.priceFormatStyle.locale
+          guard let monthlyPriceString = numberFormatter.string(from: NSNumber(floatLiteral: monthlyPrice)) else {
+            InAppPurchaseHandler.logger.error(message: "Monthly price could not be converted to a string with numberFormatter: \(monthlyPrice)")
+            assertionFailure()
```
oskirby, line 68: What happens if we hit this assert here? Generally speaking, an assert should indicate a failure of the programming logic. Could we get here if there was a misconfiguration in the appstore or Guardian or something? While that would be a serious error on our end, it probably doesn't necessitate app termination or sending us coredumps.

oskirby (approved):
I think this looks good. I left a couple of questions that might just be due to my inexperience with Objective-C and Swift.

src/constants.cpp
mcleinman: It was added here: https://github.com/mozilla-services/guardian-website/pull/1541/

src/platforms/ios/iosiaphandler.swift
mcleinman: Hitting that assert: On a debug build we would crash, on a release build we'd log something but the user would have no impact and the app would continue. There is no way this should ever fail, so an `assertionFailure` feels appropriate here to me.

src/tasks/purchase/taskpurchase.h
mcleinman: Good call, thank you. Will be updated on the next push. (I also like consistency!)

src/platforms/ios/iosiaphandler.mm
mcleinman: Fair, I'm on board with that. Will be updated in the next push.

src/tasks/purchase/taskpurchase.cpp
mcleinman: Doing this as `if (@available(iOS 15, *))` feels more in line with iOS coding norms, rather than pulling the OS version and hoping that format never changes.

I would *never* want something like `isOlderOS` to be a variable used in an API or anything, but in this case it's contained fully within `TaskPurchase` (and it's sole call site, `iosiaphandler.mm`). It changes which API we call within `TaskPurchase`. It also won't be there for long; I'd expect we'll move off iOS 14 within a year or so. If there is eventually a v3 of the API, it will be trivial to change the now-modern-API to be the one that is used when `isOlderOS` is true.*

We could have `TaskPurchase` decide which API to use based on parsing a version number that is passed to it. That seems far more brittle than a boolean, though - and I'm inclined to use a boolean unless there is a good reason not to.

How would you feel about a naming update like `doesUseIOS14API`? Is that too specific?

* isOlderOS is only `true` when the user is on iOS 14. iOS 15 and above uses the newer style. Two notes about this -
1. I considered having a discussion about whether we should stop supporting iOS 14 at this point because of this - and change out the code to only support StoreKit2 (rather than have this forking logic). I figured it wasn't too hard to have this forking logic, and so kept it for now.
2. iOS 14 is the oldest version of iOS we support. It came out 4 years ago. I'd imagine that within the next year or so, we'll stop supporting it. (At that point, we can remove a bunch of this code, including the `isOlderOS` variable.)
3. Thus, If we ever have a v3 of this API, we almost certainly will have stopped supporting iOS 14 by then, and we'll again 2 branches for the forking logic.

mcleinman: Sorry for the essay. I'm open to talking it out or doing it the way you're considering. It seems to me that it will be brittle, and given the short-term use case of this forking logic across 2 classes, a boolean feels most appropriate and least bug-prone. Thoughgts?

src/platforms/ios/iosiaphandler.mm
mcleinman: Once there is no reference to the object, it will automatically be destroyed.

src/platforms/ios/iosiaphandler.swift
oskirby: Thank you. I am okay with leaving this `assertionFailure()` in then.

src/platforms/ios/iosiaphandler.h
mcleinman: Good call on nullptr, added.

We can't use `InAppPurchaseHandler` here because the `InAppPurchaseHandler` class is marked as `@available(iOS 15, *)` at the top. That means in iOS 14, the app doesn't know about the class - even the name. (It seemed to make sense to mark the entire class as needing 15+ rather than marking nearly every function in the class, but I'm open to other thoughts. It felt most in line with the intent of availability marking.) However, we can't mark things as available in C++, so we're in a bit of a weird spot. (I tried to explain this in the note above this line as it's nonstandard, though if it's not clear please let me know.)

src/platforms/ios/iosiaphandler.swift
```diff
@@ -0,0 +1,214 @@
...
+
+  private func handlePurchaseResult(_ purchaseResult: Product.PurchaseResult) async {
+    switch purchaseResult {
+    case .success(let verificationResult):
+      switch verificationResult {
+      case .verified(let transaction):
+        InAppPurchaseHandler.logger.info(message: "StoreKit subscription returned success and verified")
+        let originalTransactionIdentifier: String = "\(transaction.originalID)"
+        successCallback(NSString(string: transaction.productID), NSString(string: originalTransactionIdentifier))
+        await transaction.finish()
```
oskirby, line 111: This may just be more Swiftignorance, but do we need a break here? This is the only `case` that doesn't have one.

src/tasks/purchase/taskpurchase.cpp
```diff
@@ -56,9 +58,18 @@ TaskPurchase::TaskPurchase(Op op) : Task("TaskPurchase"), m_op(op) {
 TaskPurchase::~TaskPurchase() { MZ_COUNT_DTOR(TaskPurchase); }
 
 void TaskPurchase::run() {
+#ifdef MZ_IOS
+  // APIv2 returns 200, APIv1 returns 201
```
oskirby, line 62: This is a relatively new addition to the VPN codebase, but you can also do:
```
request = new NetworkRequest(this);
request->addExpectedStatus(200);
request->addExpectedStatus(201);
```
In which case, the `NetworkRequest` will accept either status code as a successful result. We had a need for that in the Balrog handler, because we could get multiple status codes back from the update API.

oskirby (approved):
Looks good!

src/tasks/purchase/taskpurchase.cpp
mcleinman: In each scenario, we have only one expected status - I'd rather fail if we get a 200 for APIv1, as that is unexpected. What do you think about that? Is that too strict?

src/platforms/ios/iosiaphandler.swift
mcleinman: `break`s aren't needed in Swift, and this was a byproduct of my doing too much C++ lately and not thinking about it clearly. I've removed them all here.

PR #9818, opened 2024-08-28
VPN-6460: Change default for desktop start on boot

mcleinman:
## Description

Change the desktop default for start on boot. This pattern is used elsewhere in the file.

## Reference

VPN-6460

## Checklist
    
- [x] My code follows the style guidelines for this project
- [x] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [x] I have performed a self review of my own code
- [x] I have commented my code PARTICULARLY in hard to understand areas
- [x] I have added thorough tests where needed

PR #9878, opened 2024-09-16
VPN-6593: Improve start on boot

mcleinman:
## Description

https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9818/ introduced a couple issues (in order of importance):
1) When upgrading to 2.24, users who have never changed their "start on boot" preference got updated from false to true. (Today I learned that the default setting is not written in local storage - Updates to a settings are written to local storage, and otherwise we use the default. So changing the default affects existing users.)
2) The VPN was turned on at the end of onboarding (this is only expected for iOS), with a session timer that is not 0:00 (as it got turned on during the user's onboarding).

This PR fixes this by creating a temporarily boolean used during onboarding, and sets the "start on boot" setting to that temporary boolean as the user passes the onboarding screen.

## Reference

VPN-6593 and VPN-6588

## Checklist
    
### DIFF
--- a/src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
+++ b/src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
@@ -15,6 +15,8 @@ ColumnLayout {
     objectName: "onboardingStartSlide"
 
     property string telemetryScreenId: "connect_on_startup"
+    // `startAtBootCheckbox` was added to change the default `startAtBoot` setting to true. We  couldn't simply change the `startAtBoot` default in `settingslist.h`
+    // without affecting older VPN clients. See https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9878
     property bool startAtBootCheckbox: true
 
     signal nextClicked()
--- a/src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
+++ b/src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
@@ -15,6 +15,7 @@ ColumnLayout {
     objectName: "onboardingStartSlide"
 
     property string telemetryScreenId: "connect_on_startup"
+    property bool startAtBootCheckbox: true
 
     signal nextClicked()
     signal backClicked()
@@ -61,11 +62,11 @@ ColumnLayout {
         MZSettingsToggle {
             objectName: "startAtBootToggle"
 
-            checked: MZSettings.startAtBoot
+            checked: startAtBootCheckbox
             onClicked: {
-                MZSettings.startAtBoot = !MZSettings.startAtBoot
+                startAtBootCheckbox = !startAtBootCheckbox
 
-                if (MZSettings.startAtBoot) {
+                if (startAtBootCheckbox) {
                     Glean.interaction.connectOnStartupEnabled.record({
                         screen: root.telemetryScreenId,
                     });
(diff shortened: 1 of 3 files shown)


src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
```diff
@@ -61,11 +62,11 @@ ColumnLayout {
...
             objectName: "startAtBootToggle"
 
-            checked: MZSettings.startAtBoot
+            checked: startAtBootCheckbox
             onClicked: {
-                MZSettings.startAtBoot = !MZSettings.startAtBoot
+                startAtBootCheckbox = !startAtBootCheckbox
 
-                if (MZSettings.startAtBoot) {
+                if (startAtBootCheckbox) {
```
lesleyjanenorton, line 69: Unrelated to this PR but does it seem to you like we should we be collecting this telemetry when the view is destroyed or when the "Next" button is clicked, instead of here when the checkbox/toggle is clicked? IIUC, if the user disables this checkbox, we'll record a `connectOnStartupDisabled` event, but they'll have to re-enable the toggle before we'll record `connectOnStartupEnabled`.

src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
```diff
@@ -61,11 +62,11 @@ ColumnLayout {
         MZSettingsToggle {
             objectName: "startAtBootToggle"
 
-            checked: MZSettings.startAtBoot
+            checked: startAtBootCheckbox
```
lesleyjanenorton, line 65: If carting around the `startAtBootCheckbox` var makes you nervous you could also give the toggle an ID and then do: `Component.onCompleted : { toggleID.checked = true }` and then `if (checked)` on L65 and `MZSettings.startAtBoot = toggleID.checked` on L113.

src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
```diff
@@ -15,6 +15,7 @@ ColumnLayout {
     objectName: "onboardingStartSlide"
 
     property string telemetryScreenId: "connect_on_startup"
+    property bool startAtBootCheckbox: true
```
lesleyjanenorton, line 18: A comment might be helpful here. I can see us forgetting why this exists down the road.

lesleyjanenorton (approved):
R+ 👍  Left a few thoughts / comments but no action required!

src/ui/screens/onboarding/OnboardingStartSlideDesktop.qml
mcleinman: I see reasons for leaving it here, or putting down with the "next" work. I don't feel strongly on this one.
- Move it to "next": This would let us only record it once for this screen.
- Leave it here: It records the user changing their mind (from the default or their previous thought). Also, moving it  to "next" would mean we'd always record at least one change on this screen, even if the user didn't change anything (unless we add a new `didChangeSwitch` boolean, but that feels like possibly overengineering this). Additionally, it's how it currently functions, so it's easier to compare data across client versions if we leave it here.

I'm inclined to leave it here, but could easily be talked into moving it to the "next" logic.

strseb:
@lesleyjanenorton @mcleinman - Qa just pointed out to me that the regressing PR https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9818 - Is not on the release branch of 2.24 😅

PR #9879, opened 2024-09-16
VPN-6580 change 'App Exclusions' to 'Excluded Apps'

mcleinman:
## Description

A quick string change.

- I also changed dev-only strings (test descriptions and some markdown docs) for consistency for future maintainers of the app.
- I couldn't find a spot where we still use `infoText2`, so I removed that string.
- The help info sheet for the feature already used the "excluded apps" nomenclature, so there was nothing to update there.

## Reference

VPN-6580

## Checklist
    
### DIFF
--- a/docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
+++ b/docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
@@ -6,7 +6,7 @@ Epic: [VPN-4412: Suggest Apps to Exclude from VPN Protection](https://mozilla-hu
 Context / Problem Statement
 ---------------------------
 
-We're looking to create tools that prompt the user in ways that help them stay more secure. Split tunneling (called 'App Exclusion' to users) is a feature that can reduce the need to turn off the VPN entirely by excluding traffic in specific apps from the VPN tunnel, but many users don't know about it. The current feature is in settings, and provides an alphabetical checklist of installed apps:
+We're looking to create tools that prompt the user in ways that help them stay more secure. Split tunneling (called 'Excluded Apps' to users) is a feature that can reduce the need to turn off the VPN entirely by excluding traffic in specific apps from the VPN tunnel, but many users don't know about it. The current feature is in settings, and provides an alphabetical checklist of installed apps:
 
 <img src="./images/0005-01.png" height="500" />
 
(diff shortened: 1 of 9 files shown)


docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
```diff
@@ -154,7 +154,7 @@ The excluded apps list will likely live in the client repo. There are ways to we
 Considerations for roll out
 ----------------------
 
-While the reworked app exclusion page will affect all platforms that support split tunneling, notifications will only be shown to users who have already allowed our app to send them. On mobile, we show the "approve notifications" screen immediately upon launching the app for the first time, which likely depresses approval rates. We may want to re-consider how we show this as part of onboarding.
+While the reworked exclulded apps page will affect all platforms that support split tunneling, notifications will only be shown to users who have already allowed our app to send them. On mobile, we show the "approve notifications" screen immediately upon launching the app for the first time, which likely depresses approval rates. We may want to re-consider how we show this as part of onboarding.
```
strseb, line 157: typo :)

docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
strseb, line 1: Hmmmm... i thought the point of these that they are historic docs (once accepted) not docs we need to keep up to date for ever 😅 Appreciate it tho :)

strseb (approved):

docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
mcleinman: Thanks, fixed

docs/rfcs/0005-suggested-excluded-apps-tech-spec.md
mcleinman: Yeah, wasn't sure that was strictly necessary (and I wouldn't ding someone else's review for not including it), but it came up in a global search for the term so I figured it would help it be clearer for the future. (Though had I changed logic relating to that doc, I wouldn't have updated the doc. 🤷🏻 )

src/translations/strings.yaml
```diff
@@ -146,15 +146,10 @@ splittunnel:
...
-  infoText2:
-    value:
-      - For any questions about setting app exclusions, we’ve provided some helpful tips on our website.
-    comment:
-      - This displays as informational text below split tunneling settings. The link directs to a sumo article which will help certain users who are having network issues using this feature.
   infoLinkText: Learn more
   infoCardDescription:
-    value: VPN must be off to edit App exclusions
-    comment: Displays inside the information card at the top of the app exclusions screen when the VPN is on
+    value: VPN must be off to edit Excluded apps
```
flodolo, line 151: To change an existing string you need a new ID.

src/translations/strings.yaml
```diff
@@ -323,7 +318,7 @@ systray:
 
 settings:
   privacySettings: Privacy features
-  appExclusionSettings: App exclusions
+  appExclusionSettings: Excluded apps
```
flodolo, line 321: Same here, needs a new ID.

flodolo (changes requested):
Is it just me or we keep going back and forth on the terminology for this feature? Was the impact on SuMO accounted for here?

We still have strings in Pontoon referencing `Excluded apps` from the previous round.

src/translations/strings.yaml
mcleinman: I wrote it that way initially. But do you think the translations for "App exclusions" and "Excluded apps" will be substantially different? If not, perhaps we don't need to bother the translators with this one. I'll defer to you.

mcleinman:
> Is it just me or we keep going back and forth on the terminology for this feature? Was the impact on SuMO accounted for here?
> 
> We still have strings in Pontoon referencing `Excluded apps` from the previous round.

I'll let Santiago speak to the terminology updates on this. I know some places (the widget?) already use `Excluded apps`, so this is standardizing the naming.

src/translations/strings.yaml
flodolo: In general, if we think a fix it's worth for English (for consistency), we also want it in other locales. The translations will be definitely different between the two.

mcleinman: Sounds good, I've updated the PR to use new string IDs.

src/translations/strings.yaml
mcleinman: Per above, these now have new IDs.

flodolo (approved):

albionx:
@flodolo Yes, the SUMO changes are accounted for here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1916518
It will go live once this version is out.

PR #9839, opened 2024-09-05
VPN-6365: Add mozilla-vpn-keyring package

oskirby:
## Description
This adds a new Debian package, `mozilla-vpn-keyring` that should attempt install the GPG signing keys for the Mozilla APT repository, and install the repository as a source for the `mozillavpn` package. This is mostly just collecting the [Install Firefox on Linux](https://support.mozilla.org/en-US/kb/install-firefox-linux#w_install-firefox-deb-package-for-debian-based-distributions-recommended) procedure into a `.deb` file and automating it.

This is a bit of a hot mess to test right now, because:
- The Mozilla APT repository doesn't actually contain a `mozillavpn` package.
- I have no idea if the PPA automation workflow changes work, because it doesn't run in a PR.

My testing strategy is as follows:
- Checkout this branch.
- Build the package: `(cd linux/keyring && dpkg-buildpackage -b -us -uc)`
- Install the package: `sudo apt install ./linux/mozilla-vpn-keyring_0.9_all.deb`
- Update the debian package lists: `sudo apt-get update`
- You should see some messages that package lists have been download from `https://packages.mozilla.org/`
- You should not see any messages about GPG keys or signing problems.

The transition strategy for users on the PPA would look something like this:
 1. We ship a `mozillavpn` to the Mozilla APT repository (due to happen in the 2.24 release).
 2. We add a `postinst` script to `mozilla-vpn-keyring` that manually removes the PPA.
 3. We ship `mozilla-vpn-keyring` to the PPA sometime before 2.25.
 4. We add `mozilla-vpn-keyring` as a dependency for the PPA version of the `mozillavpn` package in version 2.25.
   a. Upon installation of the PPA `mozillavpn` package the user will also get `mozilla-vpn-keyring` installed as a dependency.
   b. The `mozilla-vpn-keyring` package removes the PPA and installs the APT repository.
   c. The next time the user does a package update `mozillavpn` is installed from the APT repository.

Some testing might be necessary to understand how, and if, the `mozilla-vpn-keyring` package is handled once `mozillavpn` no longer depends on it (after the PPA -> APT transition). It would be problematic if it got autoremoved as this would prevent users from getting future updates from the APT repository.  

To verify that this is the expected key, and to make sure I am not trying to do something evil and sneaky here. You can verify that this is the expected GPG key by doing:

```
[user@example]$ cat linux/keyring/mozilla-apt-keyring.gpg | gpg --dry-run --show-keys
pub   rsa2048 2021-05-04 [SC]
[...]

PR #9882, opened 2024-09-17
Fix GNUPGHOME variable in PPA automation workflow

oskirby:
## Description
In the change to the PPA automation workflow, we did a bit of refactoring to submit a second package, the `mozilla-vpn-keying`, to the testing PPAs. However, we missed that the github `runner` context isn't defined at the job level which means we can't set the `GNUPGHOME` variable globally like we can all the others.

## Reference
Introduced by #9839

## Checklist
    

mcleinman (approved):
👍🏻

PR #9884, opened 2024-09-17
Nit: Suppress Balrog network errors

oskirby:
## Description
The update check task has been fairly noisy in the logs which is due to Balrog returning a 404 status code whenever there are no updates available and the `NetworkRequest` task handles this as an error when it is really the expected the response in this case.

To suppress some of the noise in the logs, let's extend the `NetworkRequest` class so that it can accept more than one expected status code, and let's not treat it as an error when we happen to receive one of the expected status codes.

After this change, the log messages for a typical run of Balrog, when no updates are available, now looks like this:
```
[17.09.2024 10:44:32.700] (Updater) Debug: Updater created
[17.09.2024 10:44:32.701] (Balrog) Debug: Balrog created
[17.09.2024 10:44:32.701] (Balrog) Debug: URL: https://aus5.mozilla.org/json/2/FirefoxVPN/2.25.0/Darwin_x86/release-cdntest/14.6/update.json
[17.09.2024 10:44:32.701] (NetworkRequest) Debug: Network request created by TaskRelease
[17.09.2024 10:44:32.770] (NetworkRequest) Debug: Network header received
[17.09.2024 10:44:32.770] (NetworkRequest) Debug: Network reply received - status: 404 - expected: [200,404]
[17.09.2024 10:44:32.770] (Balrog) Debug: Request completed
[17.09.2024 10:44:32.770] (Balrog) Info: No updates available.
[17.09.2024 10:44:32.770] (Balrog) Debug: Balrog released
[17.09.2024 10:44:32.770] (Updater) Debug: Updater released
```

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/networkrequest.cpp
+++ b/src/networkrequest.cpp
@@ -68,8 +68,7 @@ void NetworkRequest::resetRequestHandler() {
 }
 #endif
 
-NetworkRequest::NetworkRequest(Task* parent, int status)
-    : QObject(parent) {
+NetworkRequest::NetworkRequest(Task* parent, int status) : QObject(parent) {
   MZ_COUNT_CTOR(NetworkRequest);
   logger.debug() << "Network request created by" << parent->name();
 
@@ -296,8 +295,8 @@ void NetworkRequest::processData(QNetworkReply::NetworkError error,
     return;
   }
 
-  // Otherwise, if we get this far and an expected response code was provided, but
-  // but not found. Then handle this as an error.
+  // Otherwise, if we get this far and an expected response code was provided,
+  // but not found. Handle this as an error.
   if (!m_expectedStatusCodes.isEmpty()) {
     // If the status was found, it should be handled earlier in this function.
     Q_ASSERT(!m_expectedStatusCodes.contains(status));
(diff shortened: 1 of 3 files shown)


src/networkrequest.cpp
```diff
@@ -248,6 +264,14 @@ void NetworkRequest::processData(QNetworkReply::NetworkError error,
...
 #endif
 
+  // Check for expected HTTP reponse statuses. This can distinguish successful
+  // responses (eg: 200 vs. 201), but it allows for expected error conditions
+  // (eg: 404 when it's okay for the resource not to exist).
+  if ((status != 0) && m_expectedStatusCodes.contains(status)) {
+    emit requestCompleted(data);
+    return;
+  }
+
```
mcleinman, line 274: Why keep line 283 once we've added this block? I'm not against it, just curious.

mcleinman (commented):
Looks great, just one question. Thanks!

src/networkrequest.cpp
oskirby: The behaviour before this PR was that an `m_expectedStatusCode` of zero was a special value meaning "accept any HTTP status code" and we retain that behaviour when the `m_expectedStatusCodes` is empty. In such a case, this `if` statement will evaluate to `false` and we fall-through all the way down to the `emit requestCompleted()` down on line 283.

mcleinman (approved):
👍🏻

PR #9889, opened 2024-09-19
SplitTunnel 01: Use a factory for creation

strseb:
Stacked PRs:
 * #9892
 * #9891
 * #9890
 * __->__#9889

--- --- ---

### SplitTunnel 01: Use a factory for creation

Currently you need to create the object and then call
the correct functions in order to have an initialized object.

Similar to wgnt lets make sure that no "invalid" object can even exist.
so that we can assert you will not call i.e. start() on a closed handle.
Given that it's not cool to have constructors throw exceptions lets add
a create() factory that will spit out nullptr or a working instance.

### DIFF
--- a/src/platforms/windows/daemon/windowssplittunnel.cpp
+++ b/src/platforms/windows/daemon/windowssplittunnel.cpp
@@ -4,9 +4,15 @@
 
 #include "windowssplittunnel.h"
 
+#include <qassert.h>
+
+#include <memory>
+
 #include "../windowscommons.h"
 #include "../windowsservicemanager.h"
 #include "logger.h"
+#include "platforms/windows/daemon/windowsfirewall.h"
+#include "platforms/windows/daemon/windowssplittunnel.h"
 #include "platforms/windows/windowsutils.h"
 #include "windowsfirewall.h"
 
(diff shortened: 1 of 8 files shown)


src/platforms/windows/daemon/windowssplittunnel.cpp
```diff
@@ -144,96 +150,112 @@ ProcessInfo getProcessInfo(HANDLE process, const PROCESSENTRY32W& processMeta) {
...
   ;
-
-  if (m_driver == INVALID_HANDLE_VALUE) {
+  if (driverFile == INVALID_HANDLE_VALUE) {
     WindowsUtils::windowsLog("Failed to open Driver: ");
+    return nullptr;
+  }
+  // 04: Make sure the Windows Firewall has created the sublayer
+  // otherwise the driver will fail to initialize
+  if (!fw || !fw->init()) {
```
oskirby, line 203: Minor nit: I get that this hasn't really changed, but it is kindof odd for the firewall init to be happening here. It would be a little more appropriate for `WindowsDaemon` to do the init, and then pass in a reference to the WFP sublayer as an argument to `WindowsSplitTunnel::create()`

oskirby: How exactly does the split tunnel driver figure out the WFP sublayer to use?

oskirby (approved):

src/platforms/windows/daemon/windowssplittunnel.cpp
strseb: The Sublayer is hardcoded in the driver...

PR #9903, opened 2024-09-23
VPN-5407: Start DnsPingSender

mcleinman:
## Description

In PingHelper.cpp, we have this block:
```
m_pingSender = PingSenderFactory::create(m_source, this);

// Some platforms require root access to send and receive ICMP pings. If
// we happen to be on one of these unlucky devices, create a DnsPingSender
// instead.
 if (!m_pingSender->isValid()) {
    m_pingSender->deleteLater();
    m_pingSender = new DnsPingSender(m_source, this);
} 
```

If you comment out this entire snippet except `m_pingSender = new DnsPingSender(m_source, this);` (that is, pretend we need root access), this bug reliably repros a few seconds after turning the VPN on, as well as when doing a cold launch of the app when the VPN is active.

`DnnsPingSender` has a `start` method. This should be called before `sendPing` it seems. (That is how it works in `ConnectionHealth`, the only other part of the code that uses `DnsPingSender`.) However in `PingHelper`, we initialize `DnsPingSender` but we never call `start`. But we do call `sendPing`.

To provide some more evidence that this likely the root issue: It makes sense why this bug is only on certain platforms - it's the rare platforms that enter the if block in the snippet above.

For the fix, we're removing the entire `start` method and incorporating it into the initializer. We're also fixing up a related piece on the mac ping sender.

## Reference

VPN-5407

## Checklist
    
### DIFF
--- a/src/pinghelper.cpp
+++ b/src/pinghelper.cpp
@@ -12,6 +12,7 @@
 #include "logger.h"
 #include "pingsender.h"
 #include "pingsenderfactory.h"
+#include "platforms/dummy/dummypingsender.h"
 
 // Maximum window size for ping statistics.
 constexpr int PING_STATS_WINDOW = 32;
@@ -47,10 +48,14 @@ void PingHelper::start(const QString& serverIpv4Gateway,
   // we happen to be on one of these unlucky devices, create a DnsPingSender
   // instead.
   if (!m_pingSender->isValid()) {
+    logger.warning() << "PingSenderFactory is not valid, trying DnsPingSender.";
     m_pingSender->deleteLater();
     m_pingSender = new DnsPingSender(m_source, this);
-    if (!((DnsPingSender*)m_pingSender)->start()) {
-      logger.error() << "Unable to start DNSPingSender.";
+    if (!m_pingSender->isValid()) {
+      logger.error()
+          << "DnsPingSender is also not valid, using DummyPingSender.";
+      m_pingSender->deleteLater();
+      m_pingSender = new DummyPingSender(m_source, this);
     }
   }
 
--- a/src/pinghelper.cpp
+++ b/src/pinghelper.cpp
@@ -49,6 +49,9 @@ void PingHelper::start(const QString& serverIpv4Gateway,
   if (!m_pingSender->isValid()) {
     m_pingSender->deleteLater();
     m_pingSender = new DnsPingSender(m_source, this);
+    if (!((DnsPingSender*)m_pingSender)->start()) {
+      logger.error() << "Unable to start DNSPingSender.";
+    }
   }
 
   connect(m_pingSender, &PingSender::recvPing, this, &PingHelper::pingReceived,
--- a/src/connectionhealth.cpp
+++ b/src/connectionhealth.cpp
@@ -128,7 +128,7 @@ void ConnectionHealth::startIdle() {
   m_dnsPingInitialized = false;
   m_dnsPingLatency = PING_TIME_UNSTABLE.count();
 
-  if (m_dnsPingSender.start()) {
+  if (m_dnsPingSender.isValid()) {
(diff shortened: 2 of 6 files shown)


src/pinghelper.cpp
```diff
@@ -49,6 +49,9 @@ void PingHelper::start(const QString& serverIpv4Gateway,
   if (!m_pingSender->isValid()) {
     m_pingSender->deleteLater();
     m_pingSender = new DnsPingSender(m_source, this);
+    if (!((DnsPingSender*)m_pingSender)->start()) {
```
strseb, line 52: My 2 Thoughts: 
Looking at the other pingsenders, they seem to do be ready once the constructor was finished. 
Given dnspingsender inherits pingsender - It should behave the same imho. 
It also should override `PingSender::isValid()`  and the if should use that as the input.

strseb (approved):
Great analysis! 
r+wc - i think the patch is solving the issue, but we could make that class better.

src/pinghelper.cpp
oskirby: I agree, I think if `DnsPingSender` is the only one to require a `start()` method, then it would be much better to refactor that one ping sender to behave consistently with the other ping senders. Specifically, they should all follow the contract that:
- Sockets are created (and bound) in the Constructor if they need to be long-lived.
- Fallibility is of that creation/binding is communicated via `isValid()`

oskirby (changes requested):
I think Basti already covered everything I would want to say here, but I am gonna request that we do the refactoring to `DnsPingSender` to clean this up instead of adding special cases to `PingHelper`

src/pinghelper.cpp
mcleinman: I agree that I don't like this structure (initialize, then `start`, then ready to `sendPing`). I was trying to do the lightest fix possible, while not changing much of the code. I'm open to doing a bigger fix here. I agree that we should move all of the `start` code to the initializer, and be better about using `isValid`. Will put up a new draft shortly.

src/pinghelper.cpp
```diff
@@ -49,6 +49,9 @@ void PingHelper::start(const QString& serverIpv4Gateway,
...
+      logger.error() << "Unable to start DNSPingSender.";
```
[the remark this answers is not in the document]
mcleinman: I've just made it a dummy ping sender for now, and log it. I agree it's better to bubble up further, but this seems like an improvement from what is currently there, and I'm not sure the amount of time it would take to bubble up further is worth it right now, given all the other priorities we have. Open to having my mind changed.

src/pinghelper.cpp
mcleinman: Did this here. Additionally, I realized the mac ping sender wasn't using `isValid` properly, so I updated that. (The default implementation is to always return `true`.) Perhaps this will fix some other unexpected behavior.

src/pinghelper.cpp
```diff
@@ -47,8 +48,13 @@ void PingHelper::start(const QString& serverIpv4Gateway,
   // we happen to be on one of these unlucky devices, create a DnsPingSender
   // instead.
   if (!m_pingSender->isValid()) {
+    logger.warning() << "PingSenderFactory is not valid, trying DnsPingSender.";
     m_pingSender->deleteLater();
     m_pingSender = new DnsPingSender(m_source, this);
+    if (!m_pingSender->isValid()) {
+      logger.error() << "DnsPingSender is also not valid, using DummyPingSender.";
+      m_pingSender = new DummyPingSender(m_source, this);
```
strseb, line 56: NIT: delete the now invalid dnspingsender :)

mcleinman: Good catch! Done.

oskirby (approved):
Looks good, thanks!

PR #9906, opened 2024-09-24
FXVPN-218: Socks Proxy bypass without split tunneling

oskirby:
## Description
A couple months ago, I made an attempt to get the firefox extension working on MacOS by binding the proxy connections to an interface with `IP_BOUND_IF` to try and get them to bypass the VPN connection. Unfortunately, this work never got anywhere so the MacOS implementation was abandoned.

However, when some of our devs started to experience some difficulty in getting their split tunneling drivers working on Windows, I figured it might be worth revisiting this work by doing the same thing on Windows. This should allow us to implement firefox extension in a way that doesn't require the split tunneling driver at all.

Along the way, we also implemented it for Linux, mostly because it was the easiest way to test out all the connection handling changes (it's easy there because we just need a way to set `SO_MARK` and we don't have to mess around with the routing table and firewall).

## Reference
JIRA Issue: [FXVPN-218](https://mozilla-hub.atlassian.net/browse/FXVPN-218)

## Checklist
    
[FXVPN-218]: https://mozilla-hub.atlassian.net/browse/FXVPN-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extension/socks5proxy/bin/main.cpp
+++ b/extension/socks5proxy/bin/main.cpp
@@ -4,7 +4,6 @@
 
 #include <QCommandLineParser>
 #include <QCoreApplication>
-#include <QDateTime>
 #include <QLocalServer>
 #include <QRandomGenerator>
 #include <QString>
@@ -12,6 +11,7 @@
 #include <QTimer>
 
 #include "socks5.h"
+#include "verboselogger.h"
 
 #ifdef __linux__
 #  include "linuxbypass.h"
@@ -19,55 +19,6 @@
 #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
 #  include "windowsbypass.h"
 #endif
-
-struct Event {
-  QString m_newConnection;
-  qint64 m_when;
-};
-
-class BoxcarAverage final {
-  public:
-   BoxcarAverage(int buckets = 8) : m_buckets(buckets) { advance(); }
-
-   void addSample(qint64 sample) {
-    m_data[0] += sample;
-   }
-
-   void advance() {
-    if (m_data.length() >= m_buckets) {
-      m_data.resize(m_buckets - 1);
-    }
-    m_data.push_front(0);
-   }
-
-   qint64 average() const {
-    qint64 sum = 0;
-    for (auto x : m_data) {
-      sum += x;
-    }
-    return m_data.isEmpty() ? 0 : sum / m_data.length();
-   }
-
-  private:
-   const int m_buckets;
-   QVector<qint64> m_data;
-};
-
-static QString bytesToString(qint64 bytes) {
-  if (bytes < 1024) {
-    return QString("%1b").arg(bytes);
-  }
-
-  if (bytes < 1024 * 1024) {
-    return QString("%1Kb").arg(bytes / 1024);
-  }
-
-  if (bytes < 1024 * 1024 * 1024) {
-    return QString("%1Mb").arg(bytes / (1024 * 1024));
-  }
-
-  return QString("%1Gb").arg(bytes / (1024 * 1024 * 1024));
-}
 
 struct CliOptions {
   uint16_t port = 0;
@@ -132,89 +83,6 @@ static CliOptions parseArgs(const QCoreApplication& app) {
   return out;
 };
 
-static void startVerboseCLI(const Socks5* socks5) {
-  static QList<Event> s_events;
-  static QString s_lastStatus;
-  static QTimer timer;
-
-  // Run a boxcar average of the data trasnfered.
-  constexpr const int max_counters = 8;
-  static BoxcarAverage rx_average(max_counters);
-  static BoxcarAverage tx_average(max_counters);
-
-  auto cleanup = [&]() {
-    // Update the boxcar average.
-    rx_average.advance();
-    tx_average.advance();
-  
-    qint64 now = QDateTime::currentMSecsSinceEpoch();
-    QMutableListIterator<Event> i(s_events);
-    while (i.hasNext()) {
-      if ((now - i.next().m_when) > 1000) {
-        i.remove();
-      }
-    }
-  };
-
-  auto printStatus = [socks5]() {
-    QString output;
-    {
-      QTextStream out(&output);
-      out << "Connections: " << socks5->connections();
-
-      QStringList addresses;
-      for (const Event& event : s_events) {
-        if (!event.m_newConnection.isEmpty() &&
-            !addresses.contains(event.m_newConnection)) {
-          addresses.append(event.m_newConnection);
-        }
-      }
-
-      out << " [" << addresses.join(", ") << "]";
-      out << " Up: " << bytesToString(tx_average.average()) << "/s";
-      out << " Down: " << bytesToString(rx_average.average()) << "/s";
-    }
-
-    output.truncate(80);
-    while (output.length() < 80) output.append(' ');
-    QTextStream out(stdout);
-    out << output << '\r';
-    s_lastStatus = output;
-  };
-  QObject::connect(socks5, &Socks5::connectionsChanged,
-                   [printStatus]() { printStatus(); });
-  QObject::connect(
-      socks5, &Socks5::incomingConnection,
-      [printStatus](Socks5Connection* conn) {
-        s_events.append(Event{conn->clientName(),
-                              QDateTime::currentMSecsSinceEpoch()});
-        printStatus();
-      });
-
-  QObject::connect(
-      socks5, &Socks5::dataSentReceived,
-      [](qint64 sent, qint64 received) {
-        tx_average.addSample(sent);
-        rx_average.addSample(received);
-      });
-
-  QObject::connect(&timer, &QTimer::timeout, [printStatus, cleanup]() {
-    cleanup();
-    printStatus();
-  });
-  timer.start(1000);
-
-  // Install a message handler that plays nice with the verbose output.
-  // Clears the current line - prints the log message - reprints the status.
-  qInstallMessageHandler(
-    [](QtMsgType type, const QMessageLogContext& ctx, const QString& msg) {
-      QTextStream out(stdout);
-      out << QString(80, ' ') << '\r';
-      out << msg << "\r\n";
-      out << s_lastStatus << '\r';
-    });
-}
-
 int main(int argc, char** argv) {
   QCoreApplication app(argc, argv);
   QCoreApplication::setApplicationName("socksproxy");
@@ -255,7 +123,7 @@ int main(int argc, char** argv) {
   }
 
   if (config.verbose) {
-    startVerboseCLI(socks5);
+    new VerboseLogger(socks5);
   }
 
   QObject::connect(socks5, &Socks5::incomingConnection,
--- a/extension/socks5proxy/bin/main.cpp
+++ b/extension/socks5proxy/bin/main.cpp
@@ -126,11 +126,6 @@ int main(int argc, char** argv) {
     new VerboseLogger(socks5);
   }
 
-  QObject::connect(socks5, &Socks5::incomingConnection,
-                   [](Socks5Connection* conn) {
-                     qDebug() << "Connection from peer" << conn->clientName();
-                   });
-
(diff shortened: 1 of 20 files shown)


extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,351 @@
...
+      continue;
+    }
+
+    // Ignore unrouteable addresses.
+    if (addr.isLinkLocal() || addr.isMulticast() || addr.isLoopback()) {
+      continue;
+    }
+
+    // Store the address.
+    // TODO: Compare amongst multiple addresses on an interface.
```
oskirby, line 237: This TODO actually leads to breakage of IPv6 on my device because it winds up selecting a non-routeable IPv6 source address. So, this isn't really fit for purpose until it gets addressed.

oskirby: I added a bit more filtering around address selection to prevent using local addresses and it seems to have gotten IPv6 working again. I'm going to leave this TODO in here for a bit longer, because I think we could still do better in terms of source address selection, but it seems to be perfectly workable as-is now.

extension/socks5proxy/bin/main.cpp
```diff
@@ -171,16 +186,47 @@ int main(int argc, char** argv) {
...
 
   QObject::connect(socks5, &Socks5::incomingConnection,
-                   [](const QString& peerAddress) {
-                     qDebug() << "Connection from  on port" << peerAddress;
+                   [](Socks5Connection* conn) {
+                     qDebug() << "Connection from peer" << conn->clientName();
                    });
 
+#ifdef __linux__
+  LinuxBypass* bypass = new LinuxBypass(socks5);
```
bakulf, line 226: because you don't use `bypass`, probably it's enough to do: `new LInuxByPass(socks5)` without setting the object.

extension/socks5proxy/bin/main.cpp
```diff
@@ -171,16 +186,47 @@ int main(int argc, char** argv) {
...
+#elif defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+  WindowsBypass* bypass = new WindowsBypass(socks5);
```
bakulf, line 228: same here

extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// Called by the kernel on network interface changes.
+// Runs in some unknown thread, so invoke a Qt signal to do the real work.
+static void addrChangeCallback(PVOID context, PMIB_UNICASTIPADDRESS_ROW row,
+                               MIB_NOTIFICATION_TYPE type) {
+  WindowsBypass* bypass = static_cast<WindowsBypass*>(context);
+  Q_UNUSED(type);
+
+  // Invoke the route changed signal to do the real work in Qt.
```
bakulf, line 41: this comment is wrong, right?

extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,376 @@
...
+  NotifyIpInterfaceChange(AF_UNSPEC, netChangeCallback, this, true,
+                          &m_netChangeHandle);
+  NotifyUnicastIpAddressChange(AF_UNSPEC, addrChangeCallback, this, true,
+                               &m_addrChangeHandle);
+  NotifyRouteChange2(AF_UNSPEC, routeChangeCallback, this, true,
+                     &m_routeChangeHandle);
+}
+
+WindowsBypass::~WindowsBypass() {
+  CancelMibChangeNotify2(m_netChangeHandle);
```
bakulf, line 76: This DTOR is never called. If yes, I wonder what guarantees that this object is correctly destroyed while the callbacks could be executed on any possible thread.

extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,376 @@
...
+  return result;
+}
+
+quint64 WindowsBypass::getVpnLuid() const {
+  // Get the LUID of the wireguard interface, if it's up.
+  NET_LUID luid;
+  GUID vpnInterfaceGuid = WIREGUARD_NT_GUID;
+  if (ConvertInterfaceGuidToLuid(&vpnInterfaceGuid, &luid) != NO_ERROR) {
+    return 0;
+  } else {
```
bakulf, line 164: no `else` after a `return`.

extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,376 @@
...
+      quint32 rowAddr = row->Address.Ipv4.sin_addr.s_addr;
+      addr.setAddress(qFromBigEndian<quint32>(rowAddr));
+    } else if (row->Address.si_family == AF_INET6) {
+      addr.setAddress(row->Address.Ipv6.sin6_addr.s6_addr);
+    } else {
+      continue;
+    }
+
+    // Ignore unrouteable addresses.
+    // TODO: Or would !isGlobal() || isUniqueLocalUnicast() be more consise?
```
bakulf, line 234: I vote for `isGlobal + isUniqueLocalUnicast`.

extension/socks5proxy/bin/windowsbypass.cpp
```diff
@@ -0,0 +1,376 @@
...
+  int bestLength = -1;
+  ULONG bestMetric = ULONG_MAX;
+  const MIB_IPFORWARD_ROW2* bestMatch = nullptr;
+  const QVector<MIB_IPFORWARD_ROW2>* table;
+  int family;
+
+  if (dest.protocol() == QAbstractSocket::IPv4Protocol) {
+    family = AF_INET;
+    table = &m_routeTableIpv4;
+  } else {
```
bakulf, line 276: should you check if actually dest.protocol() is IPv6Protocol?

extension/socks5proxy/bin/windowsbypass.h
```diff
@@ -0,0 +1,54 @@
...
+class QAbstractSocket;
+class QHostAddress;
+
+struct _MIB_IPFORWARD_ROW2;
+
+class WindowsBypass : public QObject {
+  Q_OBJECT
+
+ public:
+  WindowsBypass(Socks5* proxy);
```
bakulf, line 23: explicit

extension/socks5proxy/bin/windowsbypass.h
```diff
@@ -0,0 +1,54 @@
...
+#include <QObject>
+#include <QVector>
+
+class Socks5;
+class QAbstractSocket;
+class QHostAddress;
+
+struct _MIB_IPFORWARD_ROW2;
+
+class WindowsBypass : public QObject {
```
bakulf, line 19: final

extension/socks5proxy/src/socks5connection.cpp
```diff
@@ -72,12 +75,31 @@ ServerResponsePacket createServerResponsePacket(uint8_t rep,
...
+}
+
+Socks5Connection::Socks5Connection(QLocalSocket* socket) {
   connect(m_inSocket, &QIODevice::readyRead, this,
           &Socks5Connection::readyRead);
 
+  m_socksPort = 0;
+  // TODO: Some magic may be required here to resolve the entity of which app
+  // tried to connect. Some breadcrumbs:
+  //   - Linux: SO_PEERCRED can get us the cllient PID, from which we can get
```
bakulf, line 95: client

extension/socks5proxy/src/socks5connection.cpp
```diff
@@ -254,7 +277,79 @@ qint64 Socks5Connection::proxy(QIODevice* a, QIODevice* b) {
...
 
-void Socks5Connection::configureOutSocket() {
+void Socks5Connection::dnsResolutionFinished(quint16 port) {
+  QDnsLookup* lookup = qobject_cast<QDnsLookup*>(QObject::sender());
+
+  // Garbage collect the lookup when we're finished.
+  m_dnsLookupAttempts--;
+  auto guard = qScopeGuard([lookup]() {
+    if (lookup->isFinished()) {
+      lookup->deleteLater();
```
bakulf, line 287: you delete the sender? Tell me why this is fine.

extension/socks5proxy/src/socks5connection.h
```diff
@@ -5,18 +5,20 @@
...
+#include <QLocalSocket>
 #include <QObject>
 #include <QTcpSocket>
 
 class Socks5Connection final : public QObject {
   Q_OBJECT
 
  public:
-  Socks5Connection(QTcpSocket* socket, uint16_t port);
+  Socks5Connection(QTcpSocket* socket);
```
bakulf, line 18: explicit to both

extension/socks5proxy/src/socks5connection.cpp
```diff
@@ -72,12 +75,31 @@ ServerResponsePacket createServerResponsePacket(uint8_t rep,
...
+  // TODO: Some magic may be required here to resolve the entity of which app
+  // tried to connect. Some breadcrumbs:
+  //   - Linux: SO_PEERCRED can get us the cllient PID, from which we can get
+  //            the cgroup name, systemd scope and parse out the application ID.
+  //   - Windows: GetNamedPipeClientProcessId() and GetProcessImageFileNameA()
+  //              can get us the path to the calling executable.
+  //   - MacOS: SecTaskCopySigningIdentifier() can be used to grab information
+  //            about processes and their code signatures. Somewhere in there
+  //            I would expect to find the application ID too.
+  m_clientName = "TODO";
```
bakulf, line 102: are you actually going to implement this TODO?

bakulf (approved):

extension/socks5proxy/bin/windowsbypass.cpp
oskirby: As a note, there are two dimensions to consider here for doing proper IPv6 source address selection.

First, we could join a network which has multiple global prefixes and to do selection correctly, we are supposed to choose the source address which best matches the destination address. This would require storing multiple addresses and doing that selection later at socket creation.

Second, Windows will typically create multiple IPv6 addresses for each prefix, but with multiple address suffixes. We only need to store one of these, and we probably want the address with a randomized suffix, as this helps preserve the users privacy by preventing tracking of the user by their MAC address (which can be leaked in the suffix).

extension/socks5proxy/src/socks5connection.cpp
oskirby: We delete the sender if, and only if, the lookup is finished after exiting this function. This allows us to restart the lookup and continue DNS resolution simply by calling `lookup->lookup()` but if no such continuation is attempted then the `QDnsLookup` object gets garbage collected.

I suppose it is a little bit convoluted on how this works and prevents leaks or use-after-frees.

oskirby: For now, the only such case is when we receive a `CNAME` record and we need to restart the DNS lookup using the record pointed to by the `CNAME`

extension/socks5proxy/src/socks5connection.cpp
oskirby: I would like to down the road. I think it will be necessary at some point to enforce local security policies regarding which applications are permitted to bypass the VPN.

extension/socks5proxy/bin/windowsbypass.cpp
oskirby: The `WindowsBypass` class sets the `Socks5` as its parent object, meaning that their lifecycles are tied together. The DTOR will be called when the parent goes out of scope.

Also note, that `QObjects` without an explicit parent set are created as children of `qApp` and get destructed when the application exits.

If for some reason none of these DTORs run before the process quits, then it falls to the kernel cleanup misbehaving processes and safely de-register the callbacks.

lesleyjanenorton (approved):
R+ - Working on my machine

PR #9907, opened 2024-09-25
Add Webext Feature flag

strseb:
## Description

The Extension itself cannot know on which plattform it is on, and if that is supported. 
So let's have the client send a signal whether it supports "webextension" 

Also while at it, i consolidated a few `FeatureListCallback_X` with a consteval func, so in the end the compiler will evaluate to a simple `()=>true|false` lambda same as before but i think it looks neat to not change to a diffrent file to check when a feature is enabled 😅

### DIFF
--- a/src/feature/featurelist.h
+++ b/src/feature/featurelist.h
@@ -41,11 +41,8 @@ FEATURE(annualUpgrade,         // Feature ID
         FeatureCallback_true,  // Can be flipped on
         FeatureCallback_true,  // Can be flipped off
         QStringList(),         // feature dependencies
-        byPlatform({
-            .windows = true,
-            .macos = true,
-            .gnu_linux = true,
-        }))
+        byPlatform(
+            {.windows = true, .macos = true, .gnu_linux = true, .wasm = true}))
 
 FEATURE(appReview,              // Feature ID
         "App Review",           // Feature name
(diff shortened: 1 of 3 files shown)


strseb:
Okay gcc/clang 12 hates it.

src/feature/featurelist.h
```diff
@@ -38,14 +41,21 @@ FEATURE(annualUpgrade,         // Feature ID
         FeatureCallback_true,  // Can be flipped on
         FeatureCallback_true,  // Can be flipped off
         QStringList(),         // feature dependencies
-        FeatureCallback_annualUpgrade)
+        enableForPlatform({
+            .windows = true,
+            .macos = true,
+            .linux = true,
```
mcleinman, line 47: To match prior, it should have WASM here.

src/feature/featurelist.h
```diff
@@ -214,18 +225,37 @@ FEATURE(unsecuredNetworkNotification,      // Feature ID
...
+
+FEATURE(webExtension,      // Feature ID
+        "webExtension",    // Feature name
+        enableForPlatform({// Can be flipped on
+                           .windows = true,
+                           .macos = true,
+                           .linux = true}),
+        FeatureCallback_true,  // Can be flipped off
+        QStringList(),         // feature dependencies
+        enableForPlatform({    // default value
```
mcleinman, line 260: This means "default true for windows, default false elsewhere", right? The naming `enableForPlatform` starts to feel a little weird - perhaps `platformBoolean` or something like that?

mcleinman (approved):
Thanks for this. A couple things that you could change, but nothing that blocks this, given current timelines.

src/feature/featurelist.h
strseb: thanks! fixed.

PR #9908, opened 2024-09-25
Let's reboot the driver 

strseb:
## Description
Have you tried turning it on an off again, but make it for drivers!


src/platforms/windows/daemon/windowssplittunnel.cpp
```diff
@@ -673,3 +691,25 @@ bool WindowsSplitTunnel::detectConflict() {
   CloseServiceHandle(servicehandle);
   return err == ERROR_SERVICE_DOES_NOT_EXIST;
 }
+
+bool WindowsSplitTunnel::isRunning() { return getState() == STATE_RUNNING; }
+
+QString WindowsSplitTunnel::stateString() {
```
strseb, line 697: Adding this for better logs, in santiagos logs you can see state is `18446744073709551615` which is -1 as uint64 😅

lesleyjanenorton (approved):

PR #9911, opened 2024-09-26
Fix script for getting unreleased branches when there are none

flodolo:
## Description

`i18n_update.yml` has been [failing for over a week](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/workflows/i18n_update.yaml) (the PR goes through to `main`, it's just the next step failing). The reason is that there are no unreleased branches.

I think this should fix, but hard to tell considering there are no unreleased branches at the moment.

```
InvalidOperation: /home/runner/work/mozilla-vpn-client/mozilla-vpn-client/.github/scripts/get_unreleased_branches.ps1:34
Line |
  34 |  if ( $UNRELEASED_BRANCHES.GetType() -Eq [string] ){
     |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | You cannot call a method on a null-valued expression.
Error: Process completed with exit code 1.
```
    
### DIFF
--- a/.github/scripts/get_unreleased_branches.ps1
+++ b/.github/scripts/get_unreleased_branches.ps1
@@ -1,10 +1,10 @@
 #
 # This script fetches all remote branches
-# And filteres out branches that have a github release connected to that. 
-# Draft releases will count too. 
+# And filteres out branches that have a GitHub release connected to that.
+# Draft releases will count too.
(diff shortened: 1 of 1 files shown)


strseb (approved):
thanks!

PR #9914, opened 2024-09-26
Update rust to v1.75

oskirby:
## Description
This is a second attempt at upgrading our Cargo/Rust tooling. The last attempt got bogged down in trying to test the Windows changes locally and I got lost in a sea of refactoring. This time let's just try to do the minimal changes necessary to get Windows going and so long as it builds in CI, let's assume it works locally too for devs that use Conda.

For whatever reason, my machine has never had a working Conda environment for Windows (I just use MSVC and it works great for me).

## Reference
Previous attempt: #9812 

## Checklist
    
### DIFF
--- a/.github/workflows/auth_tests.yml
+++ b/.github/workflows/auth_tests.yml
@@ -61,6 +61,7 @@ jobs:
 
       - name: Install build dependencies
         run: |
+          ./scripts/macos/conda_install_extras.sh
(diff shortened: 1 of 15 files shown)


oskirby:
Huh, this should now only be bumping the versions of rust in the Conda environment, and Windows is still failing to build. I am beginning to wonder if this has always been busted and it only works now because of some caching thing? Or maybe the older versions of rust erroneously included some windows libs?

Issue #9895, opened 2024-09-21
flatpak-vpn-crates.json is out of date - Perhaps a lint is needed?


aaronschif:
**Describe the bug**

flatpak-vpn-crates.json is out of date. A simple `./flatpak-update-crates.sh ` seems to fix it. Creating an issue because this seems like it should be some sort of lint. A lint might run to check that Cargo.lock and flatpak-vpn-crates.json are consistent and prevent a merge otherwise.

**VPN version:**
  <Please fill this in>

**Affected Platforms:**
  <Please fill this in>

**Prerequisites:**
  <Optional: Any special steps taken not related to the app to trigger the bug>

**Steps to reproduce:**
   <Please fill this in>

**Expected result:**
  <Please fill this in>

**Actual result:**
  <Please fill this in>

**Additional notes:**
  <Optional: Any Screenshots, Videos, Logs>

PR #9915, opened 2024-09-27
Attempt to auto-update flatpak crates

oskirby:
## Description
After my last messing around with the Flatpak manifests, it turns out that it wasn't a good idea to check in a static version of `flatpak-vpn-crates.json` since our `Cargo.lock` is updated automatically by Dependabot and the two files will quickly fall out of sync. Let's make another attempt to write some automation to ensure it's updated.

The last time I tried this, it got stuck because Github prevents workflows from retriggering themselves (this runs the risk of creating an infinite loop). So we would wind up with a PR that would automatically update the flatpak manifests, but then none of the required checks wound run, leaving the PR in an un-mergeable state. However, there are some [recommended workarounds](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs) that we can use.

To try and mitigate the risk of an infinite loop, we set up this workflow so that it only runs on dependabot PRs, and only when there are no changes to the `linux/flatpak` directory.

## Reference
Github issue #9895 
My last attempt: #9616

## Checklist
    
### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -118,8 +118,7 @@ jobs:
   dependabot-cargo:
     name: "Update flatpak crates"
     runs-on: ubuntu-latest
-    #if: ${{ github.event_name == 'pull_request'}} && startsWith(gitub.head_ref, 'dependabot/cargo/')
-    if: ${{ github.event_name == 'pull_request'}}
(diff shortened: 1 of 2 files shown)


mcleinman (approved):
Looks good.

PR #9318, opened 2024-04-03
VPN-6285 - Add metric to record data transmitted per connection session

brizental:

PR #9917, opened 2024-09-27
VPN-6285: Metrics for bandwidth usage

mcleinman:
## Description

This is a continuation of https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9318, which Bea had mostly completed before she left. **Data review has been completed on that PR.**

Notes:
1. Per conversation w/ Santiago, we are not going to record these metrics on iOS. The custom distribution metric does not exist on iOS stock Glean (which is used in the network extension), and thus we can’t use this metric as designed for other platforms. We'd either have to use a different metric type on iOS (or on all platforms), and Santiago and I believe we can extrapolate the data on iOS effectively enough to not make this work necessary.
   a. Unfortunately, this doesn't mean that iOS takes no work. If we don't remove the metric, we get a compile error as the non-existant metric type is mentioned in `VPNMetrics.swift`. Thus, some ugly `sed` work was added (along with a comment) to the iOS Glean cmake files to remove these metrics completely for iOS.
1. I found some potential memory retain bugs in iOS, and left a fix for them as part of this PR.
1. Bea wrote the MemoryDistribution scaffolding, before it was decided to use a CustomDistribution for this work instead. Her intention was to leave this work in the PR, to prevent having to re-do this work later. (There is a bit of work required for each new metric type we use in the client, due to our custom Glean work.) I'm split on whether to leave this untested (and currently unused) code in (I did leave a note), or pull it out. For now, I've left it in - but I'm very open to removing it.
1. I removed a check from `Controller::getStatus`, as without it we couldn't get the data transfer data as the VPN was being turned off (without a decent sized refactor). This felt very safe, as `getStatus` called in just 2 places: First, `ConnectionHealth`, where it’s called right after confirming that controller is StateOn. And second, in `Telemetry`, when we’re recording `dataTransferTelemetry`. Thus, I feel comfortable removing the check for controller state within `Controller::getStatus`: We need it removed for telemetry’s usage of it, and it doesn’t matter for `ConnectionHealth` because we already do a check within `ConnectionHealth` before calling `getStatus`.
1. Similar to `Controller::getStatus`, I updated some Android-specific code to remove a check. That check, similarly, was blocking us from getting data as the tunnel turns off. And I can't find a code path where that check is necessary.
1. Important for testing: The short session timer debug option is broken on Android. I filed a ticket: https://mozilla-hub.atlassian.net/browse/VPN-6629
1. I tested this on Android (manually setting a short debug timer option in the code) and macOS, and made sure the app continues to work on iOS.

## Reference

VPN-6285

## Checklist
    

src/telemetry/metrics.yaml
```diff
@@ -676,6 +676,66 @@ session:
...
+    send_in_pings:
+      - vpnsession
+      - daemonsession
+    description: |
+      The amount of data sent through the VPN tunnel.
+
+      This metric is accumulated periodically on a 3hr interval
+      while the VPN is turned on.
+
+      This metric is recorded in **bytes**. It is not collected on iOS.
```
travis79, line 690: There is a `unit:` property that you could use for this information.  Something like `unit: bytes` would show up at the top level of the metric and be easier to use in analysis and in the tooling like Glean Dictionary.

oskirby:
I'm amazed at how much work has gone into the custom distribution metrics, it seems surprising that there wasn't a metric available that could have done this for us.

mcleinman:
> I'm amazed at how much work has gone into the custom distribution metrics, it seems surprising that there wasn't a metric available that could have done this for us.

Every metric we use has gone through this process, since we're essentially [rolling our own rust bindings for Glean](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/docs/Components/qtglean.md). Every time we use a new metric type, we have to do this. (Though in most cases, the "add a metric" PR was separate from the "use that new metric" PR.)

So far, every metric we've added has been rock solid. But if there are ever breakages or regressions or anything with this system, I think it would be worth a discussion about whether we should be maintaining our own Glean (essentially) with a team this size. That said, this system has been working surprisingly well thus far.

src/telemetry/metrics.yaml
mcleinman: Thank you! Updated this for both metrics.

oskirby (approved):
As best as I can tell from the code, I think this looks fine and I don't see anything wrong with merging it.

I don't have a good picture in my mind as to what shape the resulting data will look like and whether it will meet our measurement needs though, so I am taking it on faith that those questions have been resolved.

Issue #9624, opened 2024-06-01
VPN application clears all setup data on restart of the computer


hereandnowlive:
**Describe the bug**
When I restart my laptop it will wipe all progress of the previous mozilla vpn setup and act like its a fresh install every time. So basically I have to go through the initial setup every time the computer is restarted.

**VPN version:**
2.22

**Affected Platforms:**
Linux Mint

**Prerequisites:**
None

**Steps to reproduce:**
Restart Laptop (Asus Zenbook 14) and open Mozilla VPN

**Expected result:**
program remembers settings

**Actual result:**
Program treats it as a new install with no saved settings

**Additional notes:**
None

data-sync-user:
➤ Santiago Andrigo commented:

Expected to be addressed with Flatpaks.

XavierCLL:
I'm experiencing this error with the latest version 2.28.0 on Arch Linux (no flatpak). This is the first time I've encountered it, so something happened between versions 2.27 and 2.28. This is a very annoying bug.

Issue #6607, opened 2023-04-06
Requires re-auth on every boot and device is not recognized


aktaboot:
**Describe the bug**
On every reboot, the application asks to re-authenticate and does not recognize the current device, I'm guessing It recreates a device keypair, and since device list is full I have to remove the previous device-keypair for that device.

**VPN version:**
2.14.1

**Affected Platforms:**
Linux

**Prerequisites:**
 Reboot

**Steps to reproduce:**
Reboot

**Expected result:**
App will not ask to re-authenticate and will recognize the device

**Actual result:**
App asks to reauth, and does not recognize the device

**Additional notes:**
Logs:
[mozillavpn.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/files/11173285/mozillavpn.txt)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-4562)

data-sync-user:
➤ Juan Zapata commented:

[^mozillavpn-2023-4-13.txt]

data-sync-user:
➤ Betty Fleming commented:

not a blocker for 2.15

data-sync-user:
➤ Lesley Norton commented:

Should we close this as a dupe of https://mozilla-hub.atlassian.net/browse/VPN-4332 ( https://mozilla-hub.atlassian.net/browse/VPN-4332|smart-link ) and change the priority of https://mozilla-hub.atlassian.net/browse/VPN-4332 ( https://mozilla-hub.atlassian.net/browse/VPN-4332|smart-link ) to high?

data-sync-user:
➤ Matt Cleinman commented:

According to Naomi Kirby , this was fixed in 2.15.0. These logs are from 2.14.1. An upgrade should fix this one.

data-sync-user:
➤ Matt Cleinman commented:

Turns out, these were uplifted to 2.14

data-sync-user:
➤ Matt Cleinman commented:

When it boots up, it doesn’t thing it has the encrypted file. But by the time we hit CryptoSettings::readFile (or perhaps from the start), it tries to read the encrypted route. But then discovers the device doesn't have encryption, so fails.

data-sync-user:
➤ Matt Cleinman commented:

I believe that fully deleting the app (and all settings) and then reinstalling it will fix this condition.

data-sync-user:
➤ Matt Cleinman commented:

[...]

Issue #6150, opened 2023-02-24
libcrypto fails to retrieve the key


data-sync-user:
Bianca can reproduce this issue with a VM using the latest ubuntu release.

In the log, after a restart, I see this:

```
[24.02.2023 13:47:17.772] (CryptoSettings) Info: Encrypted settings with libsecrets is supported
[24.02.2023 13:47:17.834] (CryptoSettings) Debug: Key not found. Let's create it.
```

Because of this, the client resets.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-4249)
┆Reporter: Andrea Marchesini

data-sync-user:
➤ Andrea Marchesini commented:

VPN-3176

PR #9922, opened 2024-09-27
VPN-6422: Migrate to XDG Secrets Portal

oskirby:
## Description
We have a steady stream of users reporting that their Linux clients are frequently loosing their settings across reboot. This looks and smells like a failure in the `LinuxCryptoSettings` class to retrieve the settings encryption keys from the `libsecret` API. We have seen this before, and it typically results from the user's keyring being locked at the time that they launch the VPN client.

Fortunately, there is a better way to generate an application encryption key using the XDG [Secrets Portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Secret.html), and we already have it implemented for Flatpaks. This PR attempts to provide a migration strategy for users who have settings encrypted with `libsecret` to switch over to the XDG secrets instead.

We accomplish this by doing the following:
1. Extend the `CryptoSettings::getKey()` method to pass in the encrypted settings file version. This lets us distinguish if a file was encrypted with `libsecret` or the XDG portal.
2. Replace `LinuxCryptoSettings` with a wrapper that checks for a key in `libsecret` to use as a fallback if a V1 file is found, otherwise it just wraps `XdgCryptoSettings` instead.
3. Add some code to delete the keys in `libsecret` once the user attempts to write a file with an XDG secret.

Along the way, we found a bug (or quirk?). When the application is launched via the old `/etc/xdg/autostart` handlers, this results in the application being unassociated with a systemd scope. This means that any attempt to use an XDG portal results in the portals being unable to figure out the application ID. This can result in the application getting a different set of keys than when launched via the `.desktop` file. Different keys results in a failure to read the settings files, which results in an apparent loss of settings. Luckily for us, this can be addressed by using the XDG [Background Portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html) for launching at boot... which we also have implemented as a part of the Flatpak work.

There is a potential bug in here though, or maybe a just a developer annoyance. Because the XDG secrets portal is sensitive to the application ID (which in turn is derived from the systemd scopes it is launched in). We wind up with different encryption keys if we launch it via the application menu (which winds up with the proper scopes and app IDs) vs. when we start it manually on the command line (which gets no scope).

And there seems to be something odd in the RPM build world that breaks when trying to include the Qt6 private GUI libraries (for the XDG Portal helpers). This doesn't seem to cause problems on Fedora 39 or later, so we have also included an update to the Fedora build images here as well (they were long overdue for an update anyways as the old versions have been EOL for a number of months).

## Reference
JIRA Issue: [VPN-6422](https://mozilla-hub.atlassian.net/browse/VPN-6422)
User bug reports:
- #9624
- #6607
- #6150 

## Checklist
    
[VPN-6422]: https://mozilla-hub.atlassian.net/browse/VPN-6422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cryptosettings.cpp
+++ b/src/cryptosettings.cpp
@@ -229,7 +229,7 @@ bool CryptoSettings::writeFile(QIODevice& device,
                                const QSettings::SettingsMap& map) {
   logger.debug() << "Writing the settings file";
 
-  // Without a SryptoSettings we can only write plaintext settings.
+  // Without a CryptoSettings we can only write plaintext settings.
   if (s_instance == nullptr) {
     return writeJsonFile(device, map);
   }
@@ -332,7 +332,7 @@ bool CryptoSettings::writeEncryptedChachaPolyFile(
     header.append(1, metadata.length() & 0xff);
     header.append(1, (metadata.length() >> 8) & 0xff);
     header.append(metadata);
-  } else if (fileVersion != EncryptionChachaPolyV1) {
+  } else {
     logger.error() << "Unsupported encrypted file version:" << header[0];
     return false;
   }
--- a/src/cryptosettings.cpp
+++ b/src/cryptosettings.cpp
@@ -186,7 +186,7 @@ bool CryptoSettings::readEncryptedChachaPolyFile(Version fileVersion,
     return false;
   }
 
-  QByteArray key = getKey(metadata);
+  QByteArray key = getKey(fileVersion, metadata);
   if (key.isEmpty()) {
     logger.error() << "Something went wrong reading the key";
     return false;
@@ -228,13 +228,13 @@ bool CryptoSettings::readEncryptedChachaPolyFile(Version fileVersion,
 bool CryptoSettings::writeFile(QIODevice& device,
                                const QSettings::SettingsMap& map) {
   logger.debug() << "Writing the settings file";
-  CryptoSettings::Version version = NoEncryption;
 
-  if (s_instance) {
-    version = s_instance->getSupportedVersion();
+  // Without a SryptoSettings we can only write plaintext settings.
+  if (s_instance == nullptr) {
+    return writeJsonFile(device, map);
   }
 
-  switch (version) {
+  switch (s_instance->getPreferredVersion()) {
     case NoEncryption:
       return writeJsonFile(device, map);
     case EncryptionChachaPolyV1:
@@ -299,7 +299,8 @@ bool CryptoSettings::writeEncryptedChachaPolyFile(
   QByteArray nonce(NONCE_SIZE, 0x00);
   memcpy(nonce.data(), &m_lastNonce, sizeof(m_lastNonce));
 
-  QByteArray key = getKey(QByteArray());
+  Version fileVersion = getPreferredVersion();
+  QByteArray key = getKey(fileVersion, QByteArray());
   if (key.isEmpty()) {
     logger.error() << "Something went wrong reading the key";
     return false;
@@ -310,24 +311,29 @@ bool CryptoSettings::writeEncryptedChachaPolyFile(
   }
 
   QByteArray header;
-  QByteArray metadata = getMetaData();
-  if (metadata.isEmpty()) {
+  if (fileVersion == CryptoSettings::EncryptionChachaPolyV1) {
     // Encrypted V1 Header: Just the version.
-    header.append(1, CryptoSettings::EncryptionChachaPolyV1);
-  } else if (metadata.length() <= UINT16_MAX) {
+    header.append(1, fileVersion);
+  } else if (fileVersion == CryptoSettings::EncryptionChachaPolyV2) {
+    QByteArray metadata = getMetaData();
+    if (metadata.length() > UINT16_MAX) {
+      logger.error() << "Failed to write encrypted file header: too long";
+      return false;
+    }
+
     // Encrypted V2 Header:
     // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     // | Version | Reserved  | Metadata Length | Metadata  |
     // |  8-bit  |   8-bit   |     16-bit      | variable  |
     // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     header.reserve(ENCRYPTED_V2_HEADER_SIZE + metadata.length());
-    header.append(1, CryptoSettings::EncryptionChachaPolyV2);
+    header.append(1, fileVersion);
     header.append(1, 0);
     header.append(1, metadata.length() & 0xff);
     header.append(1, (metadata.length() >> 8) & 0xff);
     header.append(metadata);
-  } else {
-    logger.error() << "Failed to write encrypted file header: too long";
+  } else if (fileVersion != EncryptionChachaPolyV1) {
+    logger.error() << "Unsupported encrypted file version:" << header[0];
     return false;
   }
 
--- a/src/platforms/linux/linuxcryptosettings.cpp
+++ b/src/platforms/linux/linuxcryptosettings.cpp
@@ -94,7 +94,8 @@ void LinuxCryptoSettings::resetKey() {
   XdgCryptoSettings::resetKey();
 }
 
-QByteArray LinuxCryptoSettings::getKey(Version version, const QByteArray& metadata) {
+QByteArray LinuxCryptoSettings::getKey(Version version,
+                                       const QByteArray& metadata) {
   if (version == EncryptionChachaPolyV1) {
     // A legacy key is being requested - get it from the libsecrets API.
     return m_legacyKey;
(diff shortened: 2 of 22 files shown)


ahal:
r+ for the taskcluster changes, though removing myself so this gets a proper review.

src/platforms/linux/linuxcryptosettings.h
```diff
@@ -6,20 +6,22 @@
 #define LINUXCRYPTOSETTINGS_H
 
 #include "cryptosettings.h"
```
mcleinman, line 8: Can this line be removed now?

src/platforms/linux/linuxcryptosettings.cpp
```diff
@@ -46,108 +55,54 @@ LinuxCryptoSettings::LinuxCryptoSettings() {
...
+QByteArray LinuxCryptoSettings::getKey(Version version,
+                                       const QByteArray& metadata) {
+  if (version == EncryptionChachaPolyV1) {
+    // A legacy key is being requested - get it from the libsecrets API.
+    return m_legacyKey;
+  } else {
+    // Otherwise, use the XDG secrets portal for all new files and keys.
+    // Clear the legacy key once an attempt has been made to write a new file.
+    clearLegacyKey();
+    return XdgCryptoSettings::getKey(version, metadata);
```
mcleinman, line 106: Should we confirm this returns a key? And if it returns an empty bytearray try to fall back to legacy? Not sure what scenarios, if any, this would help in.

src/platforms/linux/linuxcryptosettings.cpp
```diff
@@ -46,108 +55,54 @@ LinuxCryptoSettings::LinuxCryptoSettings() {
...
-                     "(incompatible reply)";
+  /* Try to lookup the encryption key. This can fail for a variety of reasons.
+   * It's also not reliable, which is why we want to move users onto the XDG
+   * Secrets portal for managing their encrypted settings.
+   */
+  GError* err = nullptr;
+  auto schema = cryptosettings_get_schema();
+  gchar* password = secret_password_lookup_sync(schema, nullptr, &err, nullptr);
+  if (err != nullptr) {
+    g_error_free(err);
```
mcleinman, line 66: Do we want to log the error in this block, as we do currently?

src/cryptosettings.cpp
```diff
@@ -228,13 +228,13 @@ bool CryptoSettings::readEncryptedChachaPolyFile(Version fileVersion,
 bool CryptoSettings::writeFile(QIODevice& device,
                                const QSettings::SettingsMap& map) {
   logger.debug() << "Writing the settings file";
-  CryptoSettings::Version version = NoEncryption;
 
-  if (s_instance) {
-    version = s_instance->getSupportedVersion();
+  // Without a SryptoSettings we can only write plaintext settings.
```
mcleinman, line 232: When do we not have SyrptoSettings? Is it better to fail to plaintext, or fail to not saving the key?

src/cryptosettings.cpp
```diff
@@ -310,24 +311,29 @@ bool CryptoSettings::writeEncryptedChachaPolyFile(
...
     header.reserve(ENCRYPTED_V2_HEADER_SIZE + metadata.length());
-    header.append(1, CryptoSettings::EncryptionChachaPolyV2);
+    header.append(1, fileVersion);
     header.append(1, 0);
     header.append(1, metadata.length() & 0xff);
     header.append(1, (metadata.length() >> 8) & 0xff);
     header.append(metadata);
-  } else {
-    logger.error() << "Failed to write encrypted file header: too long";
+  } else if (fileVersion != EncryptionChachaPolyV1) {
```
mcleinman, line 335: We have an early return for ChachaPolyV1. Could this have any bad consequences, especially for non-Linux systems?

mcleinman (commented):
A bunch of questions, mostly to make sure I understand this area of the code and the tradeoffs made here.

I'd like this to be a +2 (2 VPN engineers sign off), given the security implications and my limited prior work in this area of the codebase. I'd be comfortable being one of the signoffs after I make sure I'm understanding this - happy to setup time to talk live about my questions, or you're welcome to answer them here - whatever is easier.

taskcluster/kinds/build/linux.yml
```diff
@@ -55,24 +55,24 @@ linux/noble:
         name: linux-noble
         type: build
 
-linux/fedora-fc37:
-    description: "Linux Build (Fedora/37)"
+linux/fedora-fc39:
```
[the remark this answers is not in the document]
oskirby: This PR enables the `XdgPortal` class for all Linux builds (no longer just the Flatpaks), which has a dependency on `Qt6::GuiPrivate` in order to resolve the window handle. It turned out that this dependency doesn't build cleanly on Fedora 37/38.

These versions of Fedora have also been EOL for a couple of months now, so I figured the cleanest solution was just to do an upgrade...  but I could be convinced to ignore the problem and do that upgrade in another PR.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -57,13 +57,13 @@ tasks:
         definition: linux-dpkg-build
         args:
             DOCKER_BASE_IMAGE: ubuntu:noble
-    linux-build-fedora-fc37:
-        symbol: I(linux-fedora-fc37)
+    linux-build-fedora-fc39:
```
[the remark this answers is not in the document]
oskirby: See comment above for Fedora 38. TDLR - this introduces a new dependency that fails to build on Fedora 37.

src/platforms/linux/xdgportal.cpp
```diff
@@ -14,6 +14,7 @@
 #  include <qpa/qplatformintegration.h>
 #else
 #  include <QGuiApplication>
+#  include <QWindow>
```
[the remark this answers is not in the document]
oskirby: Now that this file is built on all Linux versions, we found out that it contains a bug on Qt < 6.5.0. In particular, `QGuiApplication` only includes a foreward declaration of `QWindow` but does not define that class. Which means that the call to get the X11 window handle (`holder->window()->winId()`) will fail due to there being on definition of `QWindow::winId()`

src/cryptosettings.cpp
oskirby: Oops, that was a bit of a typo :)

When `s_instance` is null, this means that there is no platform provider for fetching keys in which case the only format we can possibly support is the plaintext settings (which is what `writeJsonFile()` writes). Nothing has really changed in this method in terms of logic, I just reorganized it to eliminate the local variable.

src/cryptosettings.cpp
oskirby: Uh, good catch, this if/else flow is bit of a mess.

The intention was that we can `writeEncryptedChachaPolyFile` to write an encrypted settings file then call `getPrefferedKey()` to determine which format the platform wants to use to select V1 or V2. If the platform gives us anything other than V1 or V2 then we shouldn't have gotten here or something is wrong with the `switch` statement in `CryptoSettings::writeFile()`.

src/platforms/linux/linuxcryptosettings.cpp
oskirby: Sure, I can put the error message back.

taskcluster/kinds/build/linux.yml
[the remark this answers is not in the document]
mcleinman: Thanks. That makes sense. Do we need to ensure any help docs or download pages are updated w/ this info?

taskcluster/kinds/docker-image/kind.yml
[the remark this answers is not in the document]
mcleinman: 👍🏻

src/platforms/linux/xdgportal.cpp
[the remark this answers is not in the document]
mcleinman: good find

src/cryptosettings.cpp
mcleinman: Thanks. Makes sense now.

mcleinman (approved):
I've added @lesleyjanenorton to this review - Like I said earlier, given that this is security-related and something I'm not that familiar with, I'd love a second r+ on this one. 

r+ from me, thanks for answering a bunch of questions as I get more familiar with this part of the code. I'm not marking "approve" yet so we can't merge it until we get a second reviewer who looks at it.

lesleyjanenorton (approved):
👍 On behalf of our futures selves, let's make sure to tell the folks in support about the launch quirk and maybe add something to the docs.

oskirby:
After a bit of research, I think we could workaround the startup quirk by creating a Systemd scope on startup if we find that the application is launched outside of one. Basically we would need to invoke the [startTransientUnit()](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.systemd1.html) method over D-Bus and request a scope name which matches the XDG desktop portal [parser implementation](https://github.com/flatpak/xdg-desktop-portal/blob/07284c093478afbb48305f3fe533ad5078a0085c/src/xdp-app-info-host.c#L80)

PR #9931, opened 2024-10-02
VPN-3112: Auto start breadcrumb on Android

mcleinman:
## Description

Adding a row with a link to the help article for Android start on boot. (Users need to set this up from Android Settings themselves.)

While the Figma (attached to the ticket) for this has link icon center justified (along with the toggle), I made this top justified to match what the toggle actually looks like.

While there is a similar looking component on the help screen, it is not a row - and didn't work here. I also tried adjusting another component to make it work as a link row, but ultimately Lesley helped me realize that we needed to create a new component for this situation.

Screenshot 27

## Reference

VPN-3112

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZBottomSheet.qml
+++ b/nebula/ui/components/MZBottomSheet.qml
@@ -17,7 +17,7 @@ import Mozilla.Shared 1.0
       contentItem: ColumnLayout {
           Rectangle {
               anchors.left: parent.left
-              anchors. right: parent.right
+              anchors.right: parent.right
               height: 20
               color: "green"
 
--- a/nebula/ui/components/MZLinkRow.qml
+++ b/nebula/ui/components/MZLinkRow.qml
@@ -9,14 +9,15 @@ import Mozilla.Shared 1.0
         
         
 RowLayout {
-    id: toggleRow
+    id: linkRow
     Layout.fillWidth: true
     Layout.rightMargin: MZTheme.theme.windowMargin
     Layout.leftMargin: MZTheme.theme.windowMargin
-    property string destinationUrl: ""
     property alias title: label.text
     property alias subLabelText: subLabel.text
     property var accessibleName: ""
+
+    signal clicked()
 
     spacing: 16
     Accessible.name: accessibleName
@@ -69,6 +70,6 @@ RowLayout {
             source: "qrc:/nebula/resources/externalLink.svg"
             anchors.centerIn: parent
         }
-        onClicked: MZUrlOpener.openUrl(destinationUrl)
+        onClicked: linkRow.clicked()
     }
 }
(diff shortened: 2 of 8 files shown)


nebula/ui/components/MZBottomSheet.qml
```diff
@@ -17,7 +17,7 @@ import Mozilla.Shared 1.0
       contentItem: ColumnLayout {
           Rectangle {
               anchors.left: parent.left
-              anchors. right: parent.right
+              anchors.right: parent.right
```
mcleinman, line 20: This is just documentation, so doesn't change anything. (If you expand this snippet you'll see it's a comment.) I noticed this typo, and fixed it.

nebula/ui/components/MZLinkRow.qml
```diff
@@ -0,0 +1,74 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.5
+import QtQuick.Layouts 1.14
+
+import Mozilla.Shared 1.0        
+        
+        
+RowLayout {
+    id: toggleRow
```
lesleyjanenorton, line 12: nit: linkRow

src/ui/screens/settings/ViewPreferences.qml
```diff
@@ -31,6 +31,16 @@ MZViewBase {
...
         spacing: MZTheme.theme.windowMargin
 
+        MZLinkRow {
+            objectName: "androidStartAtBootLink"
+
+            accessibleName: _startAtBootTitle
+            title: _startAtBootTitle
+            subLabelText: MZI18n.SettingsStartAtBootSubtitle
+            visible: Qt.platform.os === "android"
+            destinationUrl: "https://support.mozilla.org/kb/how-enable-always-vpn-android"
```
lesleyjanenorton, line 41: Suggestion: For the sake of consistency with `MZExternalLinkListItem{}` and `MZToggleRow` we could
- move this sumo URL to constants.h and then register a new url label for it in mozillavpn.cpp
- change the onClicked handler on L73 of `MZLinkRow` to be `linkRow.clicked()` and add `signal clicked()` somewhere around L20 of `MZLinkRow` (check out MZToggleRow as an example)
- drop the `destinationUrl` property from `MZLinkRow` and then change this line to `onClicked: MZUrlOpener.openUrlLabel("theNewLabel")`

lesleyjanenorton (approved):
R+ with 1 non-blocking comment :)

nebula/ui/components/MZLinkRow.qml
mcleinman: Thanks, fixed.

src/ui/screens/settings/ViewPreferences.qml
mcleinman: Thanks, updated

PR #9946, opened 2024-10-09
VPN-6640: Fix crash on reset and quit

mcleinman:
## Description

Reset and quit was crashing the app, caused by #9873.

Something that made this take a lot longer: There is a pre-existing issue that when on a debug build, reset and quit doesn't work (at least when the VPN is active). That predates #9873, and is not fixed here. But you must build a release build to have this work.

I tested for both the iOS 15+ code path and the iOS 14 code path:
- reset before logged in
- reset when logged in, VPN on
- reset when logged in, VPN off

## Reference

VPN-6640

## Checklist
    

oskirby (approved):
Looks good!

PR #9949, opened 2024-10-09
VPN-5138: fix pascal case

mcleinman:
# Draft - one more piece to address

## Description

    Describe your changes

## Reference

VPN-5138

## Checklist
    
### DIFF
--- a/src/localizer.cpp
+++ b/src/localizer.cpp
@@ -56,7 +56,7 @@ QString toPascalCase(const QString& s) {
   for (int i = 0; i < words.size(); i++) {
     QString word = words.at(i);
     if (!word.isEmpty()) {
-      words[i] = word.at(0).toUpper() + word.mid(1);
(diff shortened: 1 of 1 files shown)


src/localizer.cpp
```diff
@@ -50,19 +50,27 @@ QString toUpper(const QLocale& locale, QString input) {
...
+      } else {
+        words[i] = word.at(0).toUpper() + word.mid(1);
+      }
     }
   }
 
   return words.join("");
 }
 
+QString toPascalCase(const QString& s) { return changeCasing(s, true); }
```
lesleyjanenorton, line 70: Might be helpful to add a comment for our future selves and any poor souls who come after us about when we expect to use toPascalCase vs toLanguageId.

lesleyjanenorton (approved):
R+

src/localizer.cpp
mcleinman: Good call, thanks. Added comments.

PR #9950, opened 2024-10-10
FXVPN-213: Fix timer wonkiness in extension

lesleyjanenorton:
## Description
Occasionally, when starting the VPN from the extension, the client sends an invalid initial `connectedSince` value to the extension resulting in the extension timer erroneously reporting that the extension has been connected for > 480,000 hours. This PR checks that `m_connectedTimeInUTC` is valid before converting it to milliseconds and sending it to the extension and creates a new `timStampForExtension` if it isn't. 

## Reference

FXVPN-213
## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -344,7 +344,12 @@ qint64 Controller::connectionTimestamp() const {
     case Controller::State::StateSilentSwitching:
       [[fallthrough]];
     case Controller::State::StateSwitching:
-      return m_connectedTimeInUTC.toMSecsSinceEpoch();
+      if (m_connectedTimeInUTC.isValid()) {
+        return m_connectedTimeInUTC.toMSecsSinceEpoch();
+      }
+      const QDateTime& initialTimeStampForExtension =
+          QDateTime::currentDateTimeUtc();
+      return initialTimeStampForExtension.toMSecsSinceEpoch();
   }
   Q_UNREACHABLE();
 }


src/controller.cpp
```diff
@@ -344,7 +344,12 @@ qint64 Controller::connectionTimestamp() const {
     case Controller::State::StateSilentSwitching:
       [[fallthrough]];
     case Controller::State::StateSwitching:
-      return m_connectedTimeInUTC.toMSecsSinceEpoch();
+      if (m_connectedTimeInUTC.isValid()) {
+        return m_connectedTimeInUTC.toMSecsSinceEpoch();
+      }
+      const QDateTime& initialTimeStampForExtension =
```
mcleinman, line 350: Could we use `clearConnectedTime` or `resetConnectedTime`? Or even just set `m_connectedTimeInUTC` here if it's invalid?

lesleyjanenorton: Possibly but I'm hesitant to do anything to modify the client timer given this value is only for passing to the extension.

mcleinman: I really think we should at least set `m_connecteedTimeInUTC` here - if the timer doesn't currently exist (but should), I don't see a harm in starting the timer. You understand this part of the codebase better than I do though, and so feel free to merge as-is if this seems potentially harmful. I trust you.

mcleinman (approved):
Thanks!

PR #9951, opened 2024-10-11
VPN-6629: Fix debug daemon timer on Android

mcleinman:
## Description

There is a debug setting for short session timers - sending pings every 3 minutes when a VPN session is active, rather than the typical every 3 hours.

History of this:
- It was initially written as a `lazy var`, which caused a crash in some situations. (VPN-6629)
- It was then updated to not be lazy, but that caused the value to always default to `false` - as the `mConfig` wasn't saved when the computed var was trying to use it. (#8555)

In this PR, I change the timer to use `onTick` for debug timer pings. I had initially tried to just move the timer setup work to right before the timer is started, as then we'd have `mConfig` already set. However, that requires being on a UI thread (it's a CountDownTimer thing). I'm confident there is a way to do this, but I'm not an expert on Android - and I spent a while testing stuff out to get this UI thread from a background daemon, getting nothing working. This alternative way of handling it should suffice, and tests okay for both the debug option being active and not active.

## Reference

VPN-6629

## Checklist
    

lesleyjanenorton (approved):
👍

PR #9952, opened 2024-10-11
VPN-6626: Install socks proxy as a service

oskirby:
## Description
Time for some more wild experiments in the VPN extension's proxy. This time we try to make it autonomous as a windows service. For now, this just starts the service using a fixed port number (8123) but eventually we will probably want to make this dynamic (or use named pipes).

In this PR:
- Finally get around to testing named pipe support on Windows.
- Add firewall code to automatically install the split tunneling rules.
- Add a `WinSvcThread` class to handle the Windows service glue.
- Generate a sensible logfile so that we can inspect what the service is doing.
- Add a new service in the Wix installer to start the proxy automatically.
- Install the proxy tool as a systemd service.

This also adds a firewall policy to restrict the connecting clients to just web browsers. We do this by enumerating the browsers found in the registry key `HKEY_LOCAL_MACHINE\SOFTWARE\Clients\StartMenuInternet` and building firewall rules to block everything which doesn't originate from one of the browser binaries.

Regarding named pipes, I think ultimately that would be the best way to run this service, since it allows us to perform verification of the client application. And in theory Firefox [supports](https://github.com/mozilla/gecko-dev/blob/a753e3b7b09b368a9ba694bee5db1deb03734a45/netwerk/socket/nsSOCKSSocketProvider.cpp#L24-L38) connecting to them as a SOCKS proxy. However, I cannot get it to work locally. I suspect that Firefox has some kind of sandboxing permissions around named pipe access which is preventing the extension from connecting to them. This is probably worthy of some future research into how we want to setup the security around the VPN extension.

## Reference
JIRA issue: [VPN-6626](https://mozilla-hub.atlassian.net/browse/VPN-6626)

## Checklist
    
[VPN-6626]: https://mozilla-hub.atlassian.net/browse/VPN-6626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extension/socks5proxy/bin/winfwpolicy.cpp
+++ b/extension/socks5proxy/bin/winfwpolicy.cpp
@@ -97,7 +97,8 @@ WinFwPolicy::~WinFwPolicy() {
   }
 }
 
-void WinFwPolicy::fwpmSublayerChanged(uint changeType, const QUuid& subLayerKey) {
+void WinFwPolicy::fwpmSublayerChanged(uint changeType,
+                                      const QUuid& subLayerKey) {
   // Ignore everything except sublayer creation.
   if (changeType != FWPM_CHANGE_ADD) {
     return;
--- a/extension/socks5proxy/bin/winfwpolicy.cpp
+++ b/extension/socks5proxy/bin/winfwpolicy.cpp
@@ -8,9 +8,9 @@
 #include <fwpmu.h>
 #include <iptypes.h>
 
-#include <QString>
 #include <QScopeGuard>
 #include <QSettings>
+#include <QString>
 #include <QUuid>
 
 #include "winutils.h"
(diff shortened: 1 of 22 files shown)


extension/socks5proxy/bin/winfwpolicy.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+// Fixed GUID of the VPN Killswitch firewall sublayer
+constexpr const QUuid KILLSWITCH_FW_GUID(0xc78056ff, 0x2bc1, 0x4211, 0xaa, 0xdd,
+                                         0x7f, 0x35, 0x8d, 0xef, 0x20, 0x2d);
+
+// Fixed GUID of our own firewall sublayer
+constexpr const QUuid LOCALPROXY_FW_GUID(0x0555706c, 0x4468, 0x4ec6, 0xb4, 47,
+                                         0x20, 0xe7, 0x4a, 0x10, 0x06, 0xe7);
+
+WinFwPolicy::WinFwPolicy(QObject* parent) : QObject(parent) {
```
strseb, line 28: Totally nitpicking here. Maybe a factory function, given the constructor can fail.

strseb (approved):
Great stuff, r+wc

PR #9972, opened 2024-10-18
VPN-6407: 15 minutes between daemon server switches

mcleinman:
## Description

Based on the data coming back from this ticket, we decided to do server switches from the mobile daemons at most once every 15 minutes. On iOS this is simply changing a variable (we had previously done one at most every 5 minutes), and on Android we had to build some new logic in.

## Reference

VPN-6407

## Checklist
    
### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
@@ -36,8 +36,8 @@ class ConnectionHealth(service: VPNService) {
     private var lastHealthStatus = ConnectionStability.Stable
     private var connectionHealthTimerId: GleanTimerId? = null
 
-    private val SERVER_SWITCH_COOLDOWN_TIME = 15 // 15 minutes
-    private var nextPossibleServerSwitch = LocalDateTime.now().plusMinutes(SERVER_SWITCH_COOLDOWN_TIME)
+    private val SERVER_SWITCH_COOLDOWN_MINUTES: Long = 15
+    private var nextPossibleServerSwitch = LocalDateTime.now().plusMinutes(SERVER_SWITCH_COOLDOWN_MINUTES)
 
     var mActive = false
     var mVPNNetwork: Network? = null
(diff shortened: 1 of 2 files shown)


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -242,11 +246,18 @@ class ConnectionHealth(service: VPNService) {
             if (fallbackServerIsReachable) {
                 recordMetrics(ConnectionStability.Unstable)
 
+                if (LocalDateTime.now() < nextPossibleServerSwitch) {
+                    Log.i(TAG, "Want to switch servers, but it has not been enough time since last server switch")
```
oskirby, line 250: How often does this connection check run? Just want to be sure we aren't going to spam the logs if there's a long time to wait before the next possible server switch?

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -242,11 +246,18 @@ class ConnectionHealth(service: VPNService) {
...
                 recordMetrics(ConnectionStability.Unstable)
 
+                if (LocalDateTime.now() < nextPossibleServerSwitch) {
+                    Log.i(TAG, "Want to switch servers, but it has not been enough time since last server switch")
+                    taskDone()
+                    return@Runnable
+                }
+
                 Log.i(TAG, "Switch to fallback VPN server")
                 // We the server is online but the connection broke up, let's rest it
```
oskirby, line 256: This comment seems badly formatted it would be nice to touch it up while we're here.

oskirby (approved):
I think this looks good enough to merge, but I left a few minor comments.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
mcleinman: Every 30 seconds. In a worst case scenario, it would appear in the logs 29 times before a switch. Since we log most other potential routes of the health check (logging *something*) every 30 seconds, this doesn't seem overkill in my opinion.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
mcleinman: Thanks, updated

PR #9980, opened 2024-10-22
VPN-6408 and VPN-6678: Turn daemon server switches into round robin of servers, and fix DNS bug

mcleinman:
## Description

There are a few things in this PR, apologies. If you'd like them broken out into separate PRs, please let me know and I'll happily restructure.
1) (no ticket) Adding a developer debug option for daemon silent switching on Android. This exists on iOS already, and this was extended for Android as well. (This was necessary to test other work in this PR.)
2) (VPN-6408) Currently daemon silent server switches (which exist only on mobile, as on desktop the client is always running) will attempt a single switch to a new server. Once they've used their backup server, they never again try to switch until the VPN is again activated from the client. This updates it to use up to 4 servers, and rotate between them.
3) (VPN-6678) In looking at the code and testing my work, I discovered a bug - Special DNS (custom or privacy DNS) isn't maintained when moving to the backup server - it falls back to using the main Mullvad DNS. (This currently exists in prod, though it doesn't look like there was a ticket for it before I filed one today.) I've fixed this bug by looking at whether the main DNS is used on the main server, and setting DNS for the fallback servers appropriately.

## Reference

VPN-6408 and VPN-6678

## Checklist
    
### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
@@ -291,7 +291,7 @@ class VPNService : android.net.VpnService() {
             throw Error("no json config provided")
         }
         Log.sensitive(tag, json.toString())
-        val jServer: JSONObject = getServerConfig(json, useFallbackServer)
+        val jServer: JSONObject = getNextServerConfig(json, useFallbackServer)
         val wireguard_conf = buildWireguardConfig(jServer, json)
         val wgConfig: String = wireguard_conf.toWgUserspaceString()
         if (wgConfig.isEmpty()) {
@@ -589,10 +589,11 @@ class VPNService : android.net.VpnService() {
         return confBuilder.build()
     }
 
-    private fun getServerConfig(obj: JSONObject, useFallbackServer: Boolean): JSONObject {
+    private fun getNextServerConfig(obj: JSONObject, useFallbackServer: Boolean): JSONObject {
         if (!useFallbackServer) {
             currentServerConfig = 0
         } else {
+            // Potential future work: Use server weights when selecting next server here and in iOS.
             val numServers = obj.getJSONArray("servers").length()
             currentServerConfig = (currentServerConfig + 1) % numServers
         }
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
@@ -291,7 +291,7 @@ class VPNService : android.net.VpnService() {
             throw Error("no json config provided")
         }
         Log.sensitive(tag, json.toString())
-        val jServer: JSONObject =  getServerConfig(json, useFallbackServer)
+        val jServer: JSONObject = getServerConfig(json, useFallbackServer)
         val wireguard_conf = buildWireguardConfig(jServer, json)
         val wgConfig: String = wireguard_conf.toWgUserspaceString()
         if (wgConfig.isEmpty()) {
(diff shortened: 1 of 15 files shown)


mcleinman:
With https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9972 merged into `main`, this is now ready for review.

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
```diff
@@ -588,6 +589,16 @@ class VPNService : android.net.VpnService() {
         return confBuilder.build()
     }
 
+    private fun getServerConfig(obj: JSONObject, useFallbackServer: Boolean): JSONObject {
```
oskirby, line 592: Just a minor nitpick, do we really need the `useFallbackServer` parameter anymore? That seems like it should be covered by the `currentServerConfig` variable.

Also, this function is not idempotent, it actually does the rotation between servers, so a better name might be `getNextServerConfig()` to imply that it's iterating.

src/models/serverdata.cpp
```diff
@@ -295,12 +295,18 @@ void ServerData::setExitServerPublicKey(const QString& publicKey) {
   m_exitServerPublicKey = publicKey;
 }
 
-const Server ServerData::backupServer(const QString& currentPublicKey) const {
```
oskirby, line 298: Ideally, the backup servers should be selected according to the weighting as well....  but that is hard to do so I accept that this should be fine for fallback and recovery purposes. (See the mess of an algorithm in `Server::weightChooser()` for how it's done for the primary server) but a comment would be nice for our future selves.

src/platforms/ios/ioscontroller.swift
```diff
@@ -161,27 +161,30 @@ public class IOSControllerImpl: NSObject {
         }
     }
 
-    @objc func connect(serverData: VPNServerData, fallbackServer: VPNServerData?, excludeLocalNetworks: Bool, allowedIPAddressRanges: [VPNIPAddressRange], reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
+    @objc func connect(serverData: [VPNServerData], excludeLocalNetworks: Bool, allowedIPAddressRanges: [VPNIPAddressRange], reason: Int, gleanDebugTag: String, isSuperDooperFeatureActive: Bool, installationId: String, disconnectOnErrorCallback: @escaping () -> Void, onboardingCompletedCallback: @escaping () -> Void, vpnConfigPermissionResponseCallback: @escaping (Bool) -> Void) {
```
oskirby, line 164: nit: Does all of this really have to be on one line? It's kind of hard to read.

oskirby (approved):
Looks good - the only comments I've got are nitpicks and random philosophical journeys.

src/models/serverdata.cpp
mcleinman: Philosophizing - I *almost* went that way, but this idea of additional backups isn't relevant to desktop. (We pass just the main server to the platform-specific controller, and then the mobile controller requests additional backup servers. And for desktop controllers, they get the main server and that's all they ever need - any server switches will come from the client.)

android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/VPNService.kt
mcleinman: Updated to `getNextServerConfig`

`useFallbackServer` now just controls whether we use the first server in the config (when false) or the next server in line (when true). I think using the client-selected main server on initial connection seems helpful - and we thus would need to keep `useFallBackServer` in order to reset `currentServerConfig`. (If the daemon is maintained between VPN activations - I'm not sure if it is - we'd otherwise never reset `currentServerConfig` back to 0, and potentially take a non-first server on initial connection to a new geo, for instance.)

PR #9984, opened 2024-10-24
Re:#9966 Fix Crash in MZFlickable

strseb:
Soooo i did set up a 6.8 dev build and was able to repro the crash seen in #9966 - this is a fun one. 
In 6.8.0 we get the "onHeightChange" before the window has finished to initialize, so that when we call ` mapToItem(window.contentItem)` we defer a nullptr and crash.

### DIFF
--- a/extension/socks5proxy/bin/CMakeLists.txt
+++ b/extension/socks5proxy/bin/CMakeLists.txt
@@ -18,6 +18,7 @@ if(WIN32)
     target_sources(socksproxy PRIVATE
         windowsbypass.cpp
         windowsbypass.h)
+    target_link_libraries(socksproxy PRIVATE Iphlpapi.lib)
 
     install(FILES
         ${CMAKE_CURRENT_BINARY_DIR}/socksproxy.exe
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -207,23 +207,6 @@ int CommandUI::run(QStringList& tokens) {
     EventListener eventListener;
 #endif
 
-#ifdef false
-    // This enables the qt-creator qml debugger on debug builds.:
-    // Go to QtCreator: Debug->Start Debugging-> Attach to QML port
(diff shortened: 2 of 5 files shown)


extension/socks5proxy/bin/CMakeLists.txt
strseb, line 1: Qt 6.8.0 no longer links to this it seems so we need to manually declare that :)

src/commands/commandui.cpp
strseb, line 1: This errored and i think noone is using it, so let's just remove it.

src/platforms/windows/daemon/windowsfirewall.cpp
strseb, line 1: qAsConst is depricated in 6.8.0

extension/socks5proxy/bin/CMakeLists.txt
oskirby: I think we tend to handle this using pragmas: `#pragma comment(lib, "Iphlpapi")`

oskirby (approved):
Looks good.

Just one minor nitpick about how to link libraries on Windows, but it's not worth holding the patch up if you feel strongly about it.

XavierCLL:
Hi @strseb 

I am not sure if the purpose of this PR is fix all the problems with Qt6.8, I applied this patch (also this one 9961) to the stable version, and the UI is usable, but one thing is not working, when I click in the list of servers "Select location", the UI crash with the following traceback:

[trace.txt](https://github.com/user-attachments/files/17532325/trace.txt)

strseb:
Hey @XavierCLL - i was not expecting one small patch to magically fix all issues that might arise with 6.8 :) 
Given that we're still shipping the client on most plattforms with qt6.6 i really appreciate i really appreciate the testing, posting traces and bug's filed :) 
Let's pull that stacktrace into a new bug and we'll tackle it from there.

PR #9995, opened 2024-10-29
Nit: Fix MSVC warning

strseb:
## Description
Clang auto infers that all code-paths return a value .... msvc is unable to do so and therefore throws an error :)


oskirby (approved):
Looks good to me

PR #9998, opened 2024-10-29
Bug 1927890 - Hardcode beetmover-apt-promote task to level 1 worker

ahal:
For now there isn't a good place for us to import the candidate package during the promote phase for QA to be able to test. So we use the "staging" apt repo. However, this lives in the non production GCP project, so we have to use the level 1 workers.

In the future we'll want to either:
1. Create a second apt repo in the production GCP project that we can use for QA prior to shipping.
2. Import the package into the main `mozillavpn` production apt repo but with a different name (e.g include `rc###` or similar). Then add a repackage task that renames it to the proper name during the ship phase.

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/taskcluster/config.yml
+++ b/taskcluster/config.yml
@@ -105,6 +105,14 @@ workers:
             implementation: beetmover-import-from-gcs-to-artifact-registry
             os: scriptworker
             worker-type: 'mozillavpn-{level}-beetmover'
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1927898 for the better fix.

PR #10001, opened 2024-10-30
Nit: More ipv6+splittunnel fixes

strseb:
- **WIP WIP - look into killer network**
- **Properly filter out link local stuff**
- **cleanup**
- **cleanup**

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/scripts/windows/tools-make-mozillavpn-dir-writable.ps1
+++ b/scripts/windows/tools-make-mozillavpn-dir-writable.ps1
@@ -1,5 +1,0 @@
-$acl = Get-Acl "C:\Program Files\Mozilla\Mozilla VPN"
-$permission = "Everyone","FullControl","Allow"
-$rule = New-Object System.Security.AccessControl.FileSystemAccessRule $permission
-$acl.SetAccessRule($rule)
-$acl | Set-Acl "C:\Program Files\Mozilla\Mozilla VPN"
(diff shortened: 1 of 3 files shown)


oskirby:
This might be going way off on a tangent, but when I was implementing the split tunnel bypass logic in the proxy extension, I ran into a lot of similar difficulties with IPv6 and source address determination as well. I found that things performed pretty badly when my Windows machine had multiple network interfaces or when switching between interfaces.

Some learnings that I found from that work:

1. An interface can have a lot of IPv6 addresses, lots of which aren't routeable. You probably only want to use global addresses (eg: `addr.isGlobal() && !addr.isUniqueLocalUnicast()`). Note that ULA addresses (eg: `fc00::abcd`) are considered global by IPv6 standards, but technically don't route to the internet.
2. Addresses can change over time, you probably want to make sure we aren't using a stale address by filtering addresses by the DAD state (eg: `row->DadState == IpDadStatePreferred`). This one caught me out a bunch of times when switching between Wifi and ethernet. Because the routes would change and an otherwise good-looking address stuck around but was marked as stale.
3. Changing interfaces is also a giant pain in the ass (eg when switching from WiFi to Ethernet) and to make sense of that we also need to hook into the routing table and watch for changes to the default route and then re-run the address selection algorithm.

If you want to see how it is handled over in the socks proxy world, check out `WindowsBypass::refreshAddresses()` of `extension/socks5proxy/bin/windowsbypass.cpp`

PR #10013, opened 2024-11-04
Nit: Cleanup "BackendFailures" in controller.cpp

strseb:
Time for some cleanup.

## Description
This was the old hearbeat check
```c++
void MozillaVPN::heartbeatCompleted(bool success) {
  logger.debug() << "Server-side check done:" << success;

  if (!success) {
    m_private->m_controller.backendFailure();
    return;
  }

  if (m_state != StateBackendFailure) {
    return;
  }

  if (!modelsInitialized() || m_userState != UserAuthenticated) {
    setState(StateInitialize);
    return;
  }

  maybeStateMain();
}
```
So there we actually needed the controller to deativate and come back later. 
Nowadays we only care during auth for heartbeat, we so vpn can't be activated anymore. So let's remove the controller code and rename all the screens to heartbeat failure, as otherwise a confused engineer might think this could have something to do with the daemon :)

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -166,13 +166,6 @@ MozillaVPN::MozillaVPN() : App(nullptr), m_private(new MozillaVPNPrivate()) {
 
   connect(&m_private->m_controller, &Controller::readyToUpdate, this,
           [this]() { setState(StateUpdateRequired); });
-
-  connect(&m_private->m_controller, &Controller::readyToBackendFailure, this,
-          [this]() {
-            TaskScheduler::deleteTasks();
-            setState(StateBackendFailure);
-          });
-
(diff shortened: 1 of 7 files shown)


src/mozillavpn.cpp
```diff
@@ -1880,14 +1875,6 @@ void MozillaVPN::registerInspectorCommands() {
         AddonManager::instance()->fetch();
         return QJsonObject();
       });
-
-  InspectorHandler::registerCommand(
-      "force_backend_failure", "Force a backend failure", 0,
```
strseb, line 1885: There is actually a force_heartbeat_command aswell, so both did the same things 🫠

mcleinman (approved):
👍🏻

PR #10014, opened 2024-11-04
VPN-6213 part 2 - Pull colors into one file

mcleinman:
## Description

We had colors in two different files. We didn't have a good reason for this, so all colors have now been moved to one file. In the process, we removed some duplicate color hashes, fixed the identically named colors, and fixed some color-related log warnings. It may be easier to review this PR commit by commit.

- I tweaked colors in just one instance I believe: `#1CC5A0` `green70`, which is #1CC4A0. This difference is imperceptible, and allows us to reuse a color.
- We previously had both theme.red (red60) and colors.red (red50). All uses in QML code had been theme.red, so I made the only remaining colors.red = red60. The colors.js file had referenced the old color.red in a few places, and I updated those spots to directly use red50.
- We had color.input and theme.input; I changed one to inputState.

This PR builds on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9983, which has been merged to main.

Next few PRs for this ticket, to be built on top of one another:
1. Everything below the comment in the color file should not have hashes. Also fix alpha bug.
2. Remove unused color names - not base colors, but theming colors
3. Variable colors to general words, and break out variable colors into their own file to allow easy swapping for different themes

## Reference

VPN-6213

## Checklist
    

oskirby (approved):
My goodness, that's a lot of changes! Kind of overwhelming to review, but as far as I can tell it looks good.

PR #10017, opened 2024-11-06
Nits for mordor

strseb:
Some things found while we played with the config files: 
- **Dont try to give a pointer to an unallocated vector**
- **Dont try to create an address from an empty string**
- **Use correct byteorder for wireguard**


oskirby (approved):
Looks good!

PR #10042, opened 2024-11-14
basti/kn intervention

strseb:
This Pr adds the most basic code needed to make sure we handle intels KN service. 

For this i want to add a concept of "Interventions" as ``` 3rdparty Software that interferes with us and therefore we intervene```. Hopefully, this will be an empty folder. 

For now let's add:
-> a detection for kn, 
-> an API for the Extension to query detected interventions
-> Telemetry for kn, to make a rough case of how much we *really* should be focusing this on: should this just be a sumo article or do we really need to intervene.

### DIFF
--- a/src/interventions/killernetwork.h
+++ b/src/interventions/killernetwork.h
@@ -11,6 +11,14 @@ class WindowsServiceManager;
 
 namespace Intervention {
 
+/**
+ * @brief Intervention for Intel Killer Network.
+ *
+ * Intels Killer network services silently breaks
+ * the extensionproxy and the split-tunnel driver.
+ *
+ * This class currenlty only detects if a system will be affected.
+ */
 class KillerNetwork : public QObject {
   Q_OBJECT
  public:
@@ -26,17 +34,6 @@ class KillerNetwork : public QObject {
    * @return const QString&
    */
   static const QString id;
-
-  // Create a new Intervention.
-  KillerNetwork(WindowsDaemon* parent);
-
- private:
-  // EventCallbacks for windowsDaemon
-  void onVpnActivation();
-  void onVpnDeactivation();
-  std::unique_ptr<WindowsServiceManager> m_svm;
-
-  bool m_interfered = false;
 };
 
 };  // namespace Intervention
--- a/src/interventions/killernetwork.h
+++ b/src/interventions/killernetwork.h
@@ -14,10 +14,10 @@ namespace Intervention {
 /**
  * @brief Intervention for Intel Killer Network.
  *
- * Intels Killer network services silently breaks
- * the extensionproxy and the split-tunnel driver.
+ * Intel's Killer Network services silently breaks
+ * the extension proxy and the split-tunnel driver.
  *
- * This class currenlty only detects if a system will be affected.
+ * This class currently only detects if a system will be affected.
  */
 class KillerNetwork : public QObject {
   Q_OBJECT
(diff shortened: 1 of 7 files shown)


src/interventions/killernetwork.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Intervention {
+
+/**
+ * @brief Intervention for Intel Killer Network.
+ *
+ * Intels Killer network services silently breaks
+ * the extensionproxy and the split-tunnel driver.
+ *
+ * This class currenlty only detects if a system will be affected.
```
mcleinman, line 20:
```suggestion
 * Intel's Killer Network services silently breaks
 * the extension proxy and the split-tunnel driver.
 *
 * This class currently only detects if a system will be affected.
```

src/webextensionadapter.cpp
```diff
@@ -4,6 +4,9 @@
 
 #include "webextensionadapter.h"
 
+#include <qjsonarray.h>
+#include <qjsonobject.h>
```
mcleinman, line 8: Aren't these already included on lines 11 and 13?

mcleinman (commented):
A couple small nits, but looks good. Holding off on approval until we have an r+ from a data steward.

src/webextensionadapter.cpp
strseb: truee, i have a new vscode addon that probably added that >:(

strseb:
DATA REVIEW REQUEST
1. What questions will you answer with this data?

We're trying to measure how many users are having Intels "Killer Network" software installed. 
It is a traffic prioritization engine pre-installed on some OEM's, and it breaks the VPN Split tunnel and the VPN Extension. 

2. Why does Mozilla need to answer these questions? Are there benefits for users?

We want for a short period measure how many people are affected by this. 
This will allow us to (them remove the probe) and formulate a solution. 
A low impact count would just require a sumo article, a high impact count would make it valuable to notifiy the user about this breakage. 


3. What alternative methods did you consider to answer these questions?
   Why were they not sufficient?

We did try to measure impact by observing web forums complaining about breakage, but the sample size is too small and skewed towards technical users. 

4. Can current instrumentation answer these questions?

No.

5. List all proposed measurements and indicate the category of data collection for each
   measurement, using the Firefox data collection categories found on the Mozilla wiki.

Measurement Name | Measurement Description | Data Collection Category | Tracking Bug
---------------- | ----------------------- | ------------------------ | ------------
`sample.killer_network_detected` | The Conflicting Service 'Killer network' is detected  | technical | https://mozilla-hub.atlassian.net/browse/VPN-6606

6. Please provide a link to the documentation for this data collection which
   describes the ultimate data set in a public, complete, and accurate way.

This collection is Glean so is documented [in the Glean Dictionary](https://dictionary.telemetry.mozilla.org).

7. How long will this data be collected?

This collection has expiry '2025-06-30'.

8. What populations will you measure?

All channels, countries, and locales. No filters.

9. If this data collection is default on, what is the opt-out mechanism for users?

These collections are Glean. The opt-out can be found in the product's preferences.

10. Please provide a general description of how you will analyze this data.

We will use looker to check the results.

11. Where do you intend to share the results of your analysis?

They will be shared internally with the team. 

12. Is there a third-party tool (i.e. not Glean or Telemetry) that you
    are proposing to use for this data collection?

No.

strseb:
Taggging @travis79 for a data-review :)

strseb:
@mcleinman  this is ready for review, i've pulled the telemetry from this pr and moving this to another

src/webextensionadapter.cpp
```diff
@@ -33,6 +32,14 @@
 #  include "platforms/windows/windowsutils.h"
 #endif
 
+#if defined(MZ_LINUX) && !defined(MZ_FLATPAK)
```
mcleinman, line 35: Nice, thanks!

mcleinman (approved):
Looks good, thanks!

PR #10043, opened 2024-11-14
Fix iOS build failures on Taskcluster

oskirby:
## Description
The iOS builds have been failing on Taskcluster for a while now, and it seems like the problem is related to the `SDKROOT` environment being set. I am guessing this is due to a change on the MacOS worker machines. To fix it, we basically just need to unset the environment variable and let CMake do its thing.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

mcleinman (approved):
👍🏻 Thanks!

PR #10045, opened 2024-11-14
FXVPN-32 add message (and fix bug involving lists and shared strings)

mcleinman:
## Description

This adds the in-app message and notification for the web extension.

Notes:
- Before this goes live, we'll need to do a few final touch ups (being tracked as FXVPN-280):
  - Confirm the pref will flip (otherwise we need to change a condition)
  - Confirm the version this will go live with (otherwise we need to change a condition)
  - Update the URL for AMO
  - Update the datestamp (if needed)
- I included some fixes for addons shared strings work, originally done as https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9763. There were some bugs when using unordered or ordered lists, which have been fixed in this PR.
- I removed sample strings from the `strings.yaml` file, as we now have enough strings that we don't need those samples.
- I also fixed up some documentation

## Testing
This code will not be seen until 2.25.1. Thus, you need to change the `conditions` block. Given that QA will not be able to test this after merging to `main`, we need at least one other VPN engineer to test this code. To test:

To test:
1. Change the conditions block to the following (substituting the platform for another one, depending on what you're using):
```
  "conditions": {
    "min_client_version": "2.23.1",
    "enabled_features": ["webExtension"],
    "trigger_time": 1000,
    "platforms": ["macos"]
  },
```
2. After this is merged, we’ll get updated translation files from the translation repo. For now, manually add these lines to the end of `./3rdparty/i18n/en/addons/strings.xliff`:
```
<trans-unit id="vpn.tryFirefoxExtension.title">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Try the Firefox extension</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.subtitle">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Personalize your VPN protections to meet your web browsing needs.</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.bullet1">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Turn off VPN for specific websites</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.bullet2">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Set different locations for different websites, to see the web as you prefer</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.bullet3">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Keep protection for Firefox on, even when the Mozilla VPN app is off</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.finalLine">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Not using Firefox to browse? Give it a try. </source>
      </trans-unit>      
      <trans-unit id="vpn.tryFirefoxExtension.getExtension">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Get the extension</source>
      </trans-unit>
      <trans-unit id="vpn.tryFirefoxExtension.downloadFirefox">
        <note annotates="source" from="developer">Standard text in a composer block</note>
        <source>Download Firefox</source>
      </trans-unit>
```
3. Follow the instructions in the `How to implement and test add-ons` section of `./docs/Components/Addons/index.md`.
4. Compile and run the app.
5. In the debug menu, turn on the `webExtension` feature. Hard quit and relaunch the app.
6. You should see the message and notification when you relaunch the app.

## Reference

FXVPN-32

## Checklist
    

flodolo (approved):

addons/message_try_firefox_extension/manifest.json
```diff
@@ -0,0 +1,60 @@
+{
+  "api_version": "0.1",
+  "id": "message_try_firefox_extension",
+  "name": "Try the Firefox extension",
+  "type": "message",
+  "conditions": {
+    "min_client_version": "2.25.1",
+    "enabled_features": ["webExtension"],
```
lesleyjanenorton, line 8: ++`localProxy`

addons/message_try_firefox_extension/manifest.json
```diff
@@ -0,0 +1,60 @@
...
+    "trigger_time": 1209600,
+    "platforms": ["windows"]
+  },
+  "message": {
+    "date": 1733157651,
+    "usesSharedStrings": true,
+    "shortVersion": "N/A",
```
lesleyjanenorton, line 15: should this be "" ?

src/mozillavpn.cpp
```diff
@@ -1282,6 +1282,15 @@ void MozillaVPN::registerUrlOpenerLabels() {
             : settingsHolder->captivePortalIpv4Addresses().first());
   });
 
+  uo->registerUrlLabel("downloadExtension", []() -> QString {
+    // TODO: This link MUST be updated with a link to our extension
+    return "https://addons.mozilla.org/en-US/firefox/";
+  });
+
+  uo->registerUrlLabel("downloadFirefoxWindows", []() -> QString {
+    return "https://www.mozilla.org/firefox/windows/";
```
lesleyjanenorton, line 1291: If we do `https://www.mozilla.org/firefox/new/` it will take them directly to the desktop install page and bedrock will automagically determine the platform. We probably want to attach some UTM parameters to this link too.

addons/message_try_firefox_extension/manifest.json
oskirby: Actually, we just removed the `localProxy` feature in PR #10058 since it's no longer something that the client can toggle (it's just a service that happens to exist on Windows and Linux). The "feature" still gets reported through the web extension, but it's value just checks if the corresponding serivce is running.

oskirby: Hmm, maybe that was a mistake and I should put it back as an un-toggle-able feature.

mcleinman:
This is reviewable - I've put it on draft as a reminder to not merge it before https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10065.

src/mozillavpn.cpp
mcleinman: Thanks, updated the URL (and name). I left a message for Santiago on the associated ticket to see if he wants UTM params on either of these links.

addons/message_try_firefox_extension/manifest.json
mcleinman: @lesleyjanenorton , were you suggesting adding `localProxy` in addition to `webExtension`?

(I'm looking for the name of a feature that will be the thing we "flip on" to make this go live for users - to be clear this is not a user-toggleable thing.)

addons/message_try_firefox_extension/manifest.json
mcleinman: Updated this. It doesn't matter, as it's never shown - it's used for update/what's new messages (to drop a version number in the "Update to ___" slot). 

Potentially this was a bad design choice of mine to always require it (messages using shared strings don't pass validation without it). I'm not inclined to update it right now, given how rarely we have messages like this - though let me know if you think it's needed now.

mcleinman:
Taking off draft as #10065 is merged

addons/message_try_firefox_extension/manifest.json
oskirby: To summarize a zoom discussion: I'm not in favor of adding a `localProxy` feature flag back. The proxy, being a windows/systemd service required administrative privileges to start and stop and the client UI does not hold such privileges... to make this controllable by a feature flag it would have to be managed through the daemon and that is a lot of work for not a lot of gain.

mcleinman: Updated per our meeting

mcleinman:
I removed the feature trigger (as it no longer is accurate) and I updated the version to something far in the future (so this message won't accidentally get shown) - I've updated [the follow-up ticket](https://mozilla-hub.atlassian.net/jira/software/c/projects/FXVPN/issues/FXVPN-280?jql=project%20%3D%20%22FXVPN%22%20ORDER%20BY%20created%20DESC) as appropriate.

addons/message_try_firefox_extension/manifest.json
```diff
@@ -0,0 +1,59 @@
+{
+  "api_version": "0.1",
+  "id": "message_try_firefox_extension",
+  "name": "Try the Firefox extension",
+  "type": "message",
+  "conditions": {
+    "min_client_version": "3.25.1",
```
lesleyjanenorton, line 7: So that QA can start testing this we could do `"env": "staging",` instead.

lesleyjanenorton (approved):

addons/message_try_firefox_extension/manifest.json
mcleinman: Thanks, I thought I was forgetting one. I've updated this, and will update that follow up ticket in a moment.

addons/message_try_firefox_extension/getExtension.js
```diff
@@ -0,0 +1,3 @@
+((api) => {
+  api.urlOpener.openUrlLabel('downloadExtension');
```
oskirby, line 2: Same comment as for `downloadFirefox.js` this could done directly as `api.urlOpener.openUrl('https://foo.bar/baz/...')`

oskirby (approved):
Probably good enough to merge. Just a few nits.

addons/message_try_firefox_extension/downloadFirefox.js
```diff
@@ -0,0 +1,3 @@
+((api) => {
+  api.urlOpener.openUrlLabel('downloadFirefox');
```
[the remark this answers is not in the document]
oskirby: I don't like that to change the URL we have to change the client, which kind of makes the addon system unnecessarily brittle.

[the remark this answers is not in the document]
mcleinman: I was just following convention here, and I don't love it either.

That said, nearly all addon changes take client work (for non-English languages), as we need to release new client versions to get updated translations - a brand new addon without a client update wouldn't be translated.

Since it's a nit, I'm going to leave this for now. If we want to change the convention, let's do it for all addons at once.

PR #10052, opened 2024-11-19
Fix flaky AuthenticationInAppAccountCreationNav tests

oskirby:
## Description
At last, my cup runneth over with frustration when it comes to the `AuthenticationInAppAccountCreationNav` tests, and it's time to figure out why they are flaky. It seems the problem has to do with the fact that we have multiple QML loaders at play in the authentication screens, and for the most part we want to be waiting on the Authentication screen's loader to finish rather than the global loader.

We should also be checking the auth state as well to ensure we get the correct state transitions.

I should probably look at the Addons tests too at some point, but those are scary.

## Reference
So, so many test retries. Arggggggghhhhh

## Checklist
    

mcleinman (approved):
Thank you!!!!

PR #10053, opened 2024-11-19
Use caching to speed up functional tests

oskirby:
## Description
The functional tests are slow, and it sucks. Here are some tweaks to make it go faster:
 - Install `ccache` which we'll use to cache compiled objects.
 - Setup `actions/cache@v4` to save the results of the ccache folder on every `push`
 - Setup `actions/cache@v4` to pull the ccache folder from the PR's base commit.
 - Refactor the test addon build into a reusable workflow and cache the addons since they rarely change.

Some future work to make it go even faster:
 - The rust libraries don't benefit from `ccache` and they are slow to build. However, they don't change very often so maybe we should just cache the generated libs?
 - Miniconda is sloooooow. Unfortunately, caching it is complicated because of how it interracts with pip packages.
 - Miniconda is kinda too big to cache. Github workers tend to have 1-2GB of cache space available and the Miniconda package cache is something like 1.1GB in size.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/.github/workflows/build_addons.yaml
+++ b/.github/workflows/build_addons.yaml
@@ -1,0 +1,44 @@
+# A reusable workflow to build addons
+on:
+  workflow_call:
+    inputs:
+      test-addons-name:
+        required: true
+        type: string
+
+jobs:
+  build-test-addons:
+    name: Build Test Addons
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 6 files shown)


mcleinman (approved):
Thanks!

strseb (approved):

PR #10054, opened 2024-11-19
VPN-6213 (part 6): Restructure files, add linter checks

mcleinman:
## Description

This is the final PR for restructuring our colors to allow for additional theming. This builds on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10028, which built on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10019, which in turn built on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10015, which in turn built on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10014, which in turn built on https://github.com/mozilla-mobile/mozilla-vpn-client/pull/9983. All 5 have been merged into `main`.

In this PR (roughly in order of commits):
- Files are reworked
   - theme.js is renamed to sizing.js
   - the colors.js file is broken out into 3 files: `colors.js` (identical for all themes), `theme.js` (customized for each theme), and `derived-theme.js` (identical for all themes)
   - The 2 color files identical for all themes were moved up to the main theme directory
   - There is a hacky thing at the top of `colors.js` and the bottom of `derived-theme.js` to allow us to merge all 3 color files into one before trying to read it as a JS object. This is well-commented. (Without this, I was having trouble letting `theme.js` refer to colors named in `colors.js`.)
- `theme.cpp` is reworked to match the restructured files
- Added a sample of another theme. (It's very ugly, and I didn't touch all the colors - just the very basic colors to make a sample, very imperfect dark mode.)
- Added a python file to make sure our colors don't stray from the intention. It will run on every PR via the PR linting work. I've tested all of it's checks by manually changing color and QML files. This file:
   - Ensures that `theme.js` files and `derived-theme.js` do not contain any hex color codes.
   - Ensures that all `theme.js` files use the exact same list of theme-defined colors.
   - Ensures there are no explicit add hex color codes in any QML file.
   - Ensures that all `MZTheme.colors.x` colors in all QML files come from either `theme.js` or `theme-derived.js` (to make sure we don't use anything like `MZTheme.colors.red` in a QML file).

To see the secondary theme, use the developer debug menu's theme selection menu.

## Reference

VPN-6213

## Checklist
    

src/theme.cpp
strseb, line 1: I really love the cleanup that you did with this. 

WDYT: would it be possible to codify a theme now in c++? Like `QObject -> BaseColor -> MainColor` and have the Theme.cpp simply expose the current color QObject as property? 
Because with your nice cleanup i dont see anything the jsengine does the c++ preprocessor and a few consexpr could not do. And it's probably faster to expose a static QObject data vs concatting a string, evaling it in js and then exposing the static data.
Note not for now, this looks great but as followup task in our backlog. :)

strseb (approved):

scripts/ci/check_colors.py
```diff
@@ -0,0 +1,141 @@
...
+except FileNotFoundError:
+    exit(f"Path not found: {args.themeDirectory[0]}")
+except Exception as e:
+    exit(f"An error occurred when finding all themes: {e}")
+
+if len(all_themes) == 0:
+    exit(f"No themes found")
+
+###
+# 1. All theme's `theme.js` should have the same colors
```
oskirby, line 89: Minor nit - if we intend on enforcing this as a lint, then we should just have a single copy of `colors.js` that gets used by all themes (eg: by class inheritance or file linking or something).

oskirby (approved):
Looks great! This is awesome!

scripts/ci/check_colors.py
mcleinman: To make sure I understand, which information are you looking for a single copy of? Currently, there is only a single copy of `colors.js`.

There are 3 files assembled togethered for colors:
- `colors.js`- These are the base colors, including hex codes. There is only one copy of it, shared among all themes.
- `theme.js` - This is the custom work done for each theme. It can only refer to colors in `colors.js`, and all `theme.js` files need to expose the exact same named colors.
- `theme-derived.js` - some objects that are assembled from colors picked in `theme.js`. There is only one copy of it, and it is shared among all themes.

The colors in `theme.js` and `theme-derived.js` are the only ones that can be used in QML files.

oskirby: Oh, okay, so the idea here is that we are just checking that every `theme.js` defines the same set of colors but the values of those colors can be different for each theme...  and every theme can pull from a common `colors.js` for the hex codes.

Please disregard my nit then - this is good as is.

src/theme.cpp
mcleinman: As I got into this, I found myself unsure why we were using a JS object to start with. (Because it's only QML actually using it?) I agree it's slower, though luckily it's only loaded at the boot of the app.

If it's possible to keep this structure (base colors that cannot be used in the app, theme colors that are used in app but only refer to base colors, derived theme colors that can be used in the app and refer to theme and base colors), I'm happy to move this into C++.

PR #10057, opened 2024-11-20
Add extensionTelemetry

strseb:
## Description

This PR adds commands for the extension to send telemetry to the client. 
It's a bit bigger, therefore i've split this across 3 commits, for easyier reivew. 
1. In : https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10057/commits/d6e1d7782b1f1780a36d494f795f4af65a3b7786
I've only added the metrics, the data review happens in a google doc, i can link that if it's done. 
2. In https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10057/commits/a897f14912442bb1d84cc31e74fb0da6a81e362c
I add 3 commands:  `{t:telemetry, name: metric_name: args: any}` `{t:start_session}` `{t:stop_session}` - Those allow the extension to send any events or counts. For the bools we save them and attach them to the next ping, similar how we do that with our settings related bools. 
3. https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10057/commits/fa836d69596e6d222634581dbb09e3dc82ee378d
I tried the least refactoring solution to allow us to enable telemetry either for the extension or the client. All data is attached to a ping with a name. I've added the ability for glean to filter out on a per `ping_name` basis. I use that to filter the extensions pings if they are not enabled and the client pings if those are disabled.

### DIFF
--- a/src/glean/mzglean.cpp
+++ b/src/glean/mzglean.cpp
@@ -52,7 +52,7 @@ MZGlean::MZGlean(QObject* parent) : QObject(parent) {
   connect(AndroidVPNActivity::instance(),
           &AndroidVPNActivity::eventRequestGleanUploadEnabledState, this,
           [&]() {
-            broadcastUploadEnabledChange(
+            broadcastClientUploadEnabledChange(
                 SettingsHolder::instance()->gleanEnabled());
           });
 #endif
@@ -79,8 +79,14 @@ void MZGlean::initialize(const QString& channel) {
 
       connect(SettingsHolder::instance(), &SettingsHolder::gleanEnabledChanged,
               s_instance, []() {
-                s_instance->setUploadEnabled(
-                    SettingsHolder::instance()->gleanEnabled());
+                updatePingFilter();
+                upadateUploadEnabled();
+              });
+      connect(SettingsHolder::instance(),
+              &SettingsHolder::extensionTelemetryEnabledChanged, s_instance,
+              []() {
+                updatePingFilter();
+                upadateUploadEnabled();
               });
     }
 
@@ -115,10 +121,16 @@ void MZGlean::initialize(const QString& channel) {
 #  else
     SettingsHolder* settingsHolder = SettingsHolder::instance();
     Q_ASSERT(settingsHolder);
+    auto const clientTelemetryEnabled = settingsHolder->gleanEnabled();
+    auto const extensionTelemetryEnabled =
+        settingsHolder->extensionTelemetryEnabled();
 
-    glean_initialize(SettingsHolder::instance()->gleanEnabled(),
-                     gleanDirectory.absolutePath().toUtf8(), channel.toUtf8(),
-                     QLocale::system().name().toUtf8());
+    auto const shouldUpload =
+        extensionTelemetryEnabled | clientTelemetryEnabled;
+
+    updatePingFilter();
+    glean_initialize(shouldUpload, gleanDirectory.absolutePath().toUtf8(),
+                     channel.toUtf8(), QLocale::system().name().toUtf8());
 
     setLogPings(settingsHolder->gleanLogPings());
     if (settingsHolder->gleanDebugTagActive()) {
@@ -130,16 +142,21 @@ void MZGlean::initialize(const QString& channel) {
 }
 
 // static
-void MZGlean::setUploadEnabled(bool isTelemetryEnabled) {
-  logger.debug() << "Changing MZGlean upload status to" << isTelemetryEnabled;
+void MZGlean::upadateUploadEnabled() {
+  auto const settings = SettingsHolder::instance();
+  auto const clientTelemetryEnabled = settings->gleanEnabled();
+  auto const extensionTelemetryEnabled = settings->extensionTelemetryEnabled();
+
+  auto const shouldUpload = extensionTelemetryEnabled | clientTelemetryEnabled;
+  logger.debug() << "Changing MZGlean upload status to" << shouldUpload;
 
 #if not(defined(MZ_WASM))
-  glean_set_upload_enabled(isTelemetryEnabled);
+  glean_set_upload_enabled(shouldUpload);
 #endif
 
-  broadcastUploadEnabledChange(isTelemetryEnabled);
+  broadcastClientUploadEnabledChange(clientTelemetryEnabled);
 
-  if (isTelemetryEnabled) {
+  if (clientTelemetryEnabled) {
 #if defined(MZ_ANDROID) || defined(MZ_IOS)
     // need to reset installation ID, as it would have been cleared
     QString uuid = mozilla::glean::session::installation_id.generateAndSet();
@@ -162,7 +179,7 @@ void MZGlean::setUploadEnabled(bool isTelemetryEnabled) {
 }
 
 // static
-void MZGlean::broadcastUploadEnabledChange(bool isTelemetryEnabled) {
+void MZGlean::broadcastClientUploadEnabledChange(bool isTelemetryEnabled) {
 #if defined(MZ_ANDROID)
   logger.debug() << "Broadcasting MZGlean upload status to Android Daemon.";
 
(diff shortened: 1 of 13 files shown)


mcleinman:
This can have an updated title (dropping `WIP`), right?

strseb:
@mcleinman yes sorry :)

mcleinman (commented):
Thanks for this! Several questions.

It seems like we're doing a bunch of work to allow the user to have glean enabled for the extension but not the client (or the reverse). I'm not sure this is worth the code needed for it - was that a product directive?

src/glean/mzglean.cpp
```diff
@@ -130,16 +142,21 @@ void MZGlean::initialize(const QString& channel) {
 }
 
 // static
-void MZGlean::setUploadEnabled(bool isTelemetryEnabled) {
-  logger.debug() << "Changing MZGlean upload status to" << isTelemetryEnabled;
+void MZGlean::upadateUploadEnabled() {
```
mcleinman, line 145:
```suggestion
void MZGlean::updateUploadEnabled() {
```

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,223 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+# "impression" category events, are events where the user has just moved on to page X
+# Events under this category are recorded when a new screen is loaded
```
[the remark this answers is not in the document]
strseb: Hahaha copy pasted and missed it. will remove :)

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,223 @@
...
+        - https://mozilla-hub.atlassian.net/browse/VPN-6714
+      data_reviews:
+        - TBD
+      data_sensitivity:
+        - interaction
+      notification_emails:
+        - sstreich@mozilla.com
+        - vpn-telemetry@mozilla.com
+      expires: 2025-06-30
+    fx_protection_disenabled:
```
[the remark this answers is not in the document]
strseb: Yes! thanks

strseb:
> It seems like we're doing a bunch of work to allow the user to have glean enabled for the extension but not the client (or the reverse). I'm not sure this is worth the code needed for it - was that a product directive?

Yeeeep. Product ask (q_q )" - for the extension we have to offer an extra data opt-in flow just for data related to the extension.
And we the fear is otherwise people press "yes" to collect data but still have glean disabled so nothing gets out.

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,223 @@
...
+      expires: 2025-06-30
+      extra_keys:
+        message_state:
+          description: |
+            Full,PartialOn,PartialOff,Off,
+          type: string
+    used_feature_diable_firefox_protection:
+      type: boolean
+      description: >
+        True if the user has disabled the VPN for firefox
```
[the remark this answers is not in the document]
mcleinman: But unlike `using_system_language`, we're sending it in every start ping (in `recordAllFlags`). This seems more appropriate for a lifetime of `user` - a lifetime of `ping` where we record it from a app-saved setting doesn't feel like how we'd typically handle Glean.

src/settingslist.h
```diff
@@ -783,3 +785,23 @@ SETTING_BOOL(addonApiSetting,        // getter
              false                   // sensitive (do not log)
 )
 #endif
+
+SETTING_UINT(extensionTelemetryFlags,        // getter
```
mcleinman, line 789: This feels like an anti-pattern for Glean, and has some possible data retention concerns. The point of Glean is that it handles recording our metrics, and the lifetimes should be able to handle our use cases.

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,223 @@
...
+        - https://mozilla-hub.atlassian.net/browse/VPN-6714
+      data_reviews:
+        - TBD
+      data_sensitivity:
+        - interaction
+      notification_emails:
+        - sstreich@mozilla.com
+        - vpn-telemetry@mozilla.com
+      expires: 2025-06-30
+    error_screen:
```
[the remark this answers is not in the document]
strseb: Added!

strseb:
> Maybe this is a silly question to ask, but why do we need an API for the extension send telemetry to the client, only for the client to turn around and send to glean? Couldn't the extension send the telemetry to glean on its own? Surely Mozilla already has some solid mechanisms for intstrumenting extensions in this way?

That is a fair question! - Glean.js is not supported for extensions anymore, so noone knows what would break, there is `browser.telemetry` forwarding events to firefox glean but for that we need to be privileged, which is a no for a lot of reasons. 
Therefore we're here

src/settingslist.h
strseb: You are totally right, TIL about `lifetime =! ping` - ripped all of that code out and moved that into the metrics.yaml

src/telemetry/extension_metrics.yaml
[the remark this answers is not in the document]
strseb: Removed that part :)

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,223 @@
...
+      notification_emails:
+        - sstreich@mozilla.com
+        - vpn-telemetry@mozilla.com
+      expires: 2025-06-30
+      extra_keys:
+        message_state:
+          description: |
+            Full,PartialOn,PartialOff,Off,
+          type: string
+    used_feature_diable_firefox_protection:
```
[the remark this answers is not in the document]
strseb: done!

src/telemetry/pings.yaml
```diff
@@ -73,3 +73,23 @@ daemonsession:
       (Android only) Sent every three hours during an active VPN session.
     daemon_end: >
       Sent after as session ending metric is recorded.
+extensionsession:
```
mcleinman, line 76: Nit: There were two line breaks, I thought there should be just one (but wasn't clear, sorry). It was updated to be zero.

src/telemetry/extension_metrics.yaml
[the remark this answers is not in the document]
mcleinman: Thanks!

src/glean/mzglean.cpp
```diff
@@ -79,8 +79,14 @@ void MZGlean::initialize(const QString& channel) {
...
-                s_instance->setUploadEnabled(
-                    SettingsHolder::instance()->gleanEnabled());
+                updatePingFilter();
+                upadateUploadEnabled();
+              });
+      connect(SettingsHolder::instance(),
+              &SettingsHolder::extensionTelemetryEnabledChanged, s_instance,
+              []() {
+                updatePingFilter();
+                upadateUploadEnabled();
```
mcleinman, line 89: Nit: typo
```suggestion
                updateUploadEnabled();
```

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,236 @@
...
+      data_reviews:
+        - TBD
+      data_sensitivity:
+        - interaction
+      notification_emails:
+        - sstreich@mozilla.com
+        - vpn-telemetry@mozilla.com
+      expires: 2025-06-30
+
+    fx_protetion_disabled:
```
mcleinman, line 71: More than other typos, I think this typo should be fixed before merging as it's difficult to change metric names.

```suggestion
    fx_protection_disabled:
```

src/telemetry/extension_metrics.yaml
```diff
@@ -0,0 +1,236 @@
...
+        - vpn-telemetry@mozilla.com
+      extra_keys:
+        error:
+          description: |
+            The Type of Error: 
+              client_too_old, client_not_opened, unsupported_platform, install_needed
+              subscription_needed, signin_needed, split_tunneled
+          type: string
+      expires: 2025-06-30
+    fx_protection_enabled:
```
mcleinman, line 53:
```suggestion

    fx_protection_enabled:
```

mcleinman: Nit: Line break

mcleinman (approved):
Looks good. A few nits, and one metric typo that I think should be fixed before merging. Thank you!

PR #10061, opened 2024-11-21
WIP: Prod Testing Workflows

oskirby:
## Description
~~This is all broken - don't look.~~ Well, I'll be damned! It works.

Since we have a lot of builds happening in the VPN, across a lot of platforms, it's a bit of a problem that we don't really have a solid connection between the builds that we ship to users and the thing that gets tested by our CI. This PR attempts to resolve that discrepancy by downloading the build artifacts directly off taskcluster and running the functional test suite against them.

To get there, we did a bunch of things:
- Write some Github action magic to resolve the taskcluster jobs that are associated with a given workflow run.
- Once the build tasks are known, we then have to wait for the builds to finish and then download the artifacts off taskcluster.
- We do a bit of refactoring the functional tests into a reusable workflow and feed it with the artifacts from Taskcluster.
- Then we go completely wild and use a `matrix` strategy to run the tests on basically every kind of runner that github will give us.

As a downside, there are a bunch of tests that appear to be permabroken when run from the production artifacts.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
This was a fun experiment, but we don't realistically have the time and energy to get it polished enough to land.

PR #10064, opened 2024-11-22
Add Ctest support for functional testing

oskirby:
## Description
This is mostly just a minor annoyance that every time I have to run a functional test I have to remember and run a bunch of commands manually to get the functional testing envrionment setup. This makes it so that we can run a tests with something like the following:

 - `ctest -L functional`: Run all the functional tests
 - `ctest -L unit`: Run all the unit tests.
 - `ctest -L auth`: Run the authentication tests.
 - `ctest -L qml`: Run the QML tests.
 - `ctest -R <test name>`: Run a single test.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

PR #10076, opened 2024-11-26
Fix buggy globbing rules in addon build

oskirby:
## Description
I have noticed that the flatpak builds started failing after merging #10064 due to a failure to build the test addons, which I failed to notice because the flatpak builds were kinda flaky lately anyways. The problem seems to be that we have a buggy use of `GLOB_RECURSE` in addons.cmake that doesn't include any glob patterns and can fail to match any addons when `CMAKE_CURRENT_SOURCE_DIR` and `ADDON_SOURCE_DIR` differ.

While we're at it, we don't need to use inline python to parse the `ADDON_ID`. We have since upgraded to CMake > 2.20 which now includes native support for JSON parsing.

## Reference
Example build failure: https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/12019520242/job/33506273223

## Checklist
    
### DIFF
--- a/scripts/cmake/addons.cmake
+++ b/scripts/cmake/addons.cmake
@@ -34,7 +34,10 @@ function(add_addon_target NAME)
 
     # Scan for addon manifests if a source directory was provided.
     if(ADDON_SOURCE_DIR)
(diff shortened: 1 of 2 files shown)


mcleinman (approved):
Thanks 👍🏻

strseb (approved):

PR #10078, opened 2024-11-26
VPN-6742: update minimum iOS version to iOS 15

mcleinman:
## Description

Change minimum version, remove unneeded code.

## Reference

VPN-6742

## Checklist
    

mcleinman:
I'm having trouble getting a local iOS build going. However, if this builds on the server, we should be good to go. (I recently updated macOS, which required switching to a new version of Xcode. I've solved 3 different blockers to getting a build going, am still running into other novel issues.)

oskirby (approved):
That turned out to be a very messy diff for just deleting code, but I think it's good.

PR #10086, opened 2024-12-02
Fix addon cache keying

oskirby:
## Description
It looks like the attempted caching of addons for functional testing didn't work as expected. In particular I think that calling `hashFiles('addons/')` doesn't recursively scan the directory, which means that it doesn't react to changes in the addon content. Maybe using a `hashFiles('addons/**') with a globstar instead will do the trick.

While we're at it, let's throw the addons build tooling into the hash too.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

mcleinman (approved):
Thanks!

PR #10087, opened 2024-12-02
Fix matrix evaluation for wasm functional tests

oskirby:
## Description
I think this was introduced by #10053 which made an error in how we pass the generated test list to the functional test jobs. This should fix it and make the wasm job consistent with Linux and MacOS.

## Reference
Introduced by #10053

## Checklist
    

mcleinman (approved):
Thanks!

PR #10092, opened 2024-12-03
Assorted fixes for Xcode 16

oskirby:
## Description
I've been trying to get my iOS development environment working again after the latest round of updates from Apple. Here's the fixes I've had to apply in order to get going.

The simulator even works too, but switching from device to simulator (and vice versa) is broken and requires a complete rebuild. Looks to be something to do with how sentry/breakpad gets built that bakes in the platform config. That probably needs to be turned into a fat binary or framework or something.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -165,15 +165,11 @@ int CommandUI::run(QStringList& tokens) {
       qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
     }
 #ifdef MZ_IOS
-    else {
-      QString simDeviceName = QString(qgetenv("SIMULATOR_DEVICE_NAME"));
-      if (!simDeviceName.isEmpty()) {
-        // The network extension doesn't actually start when running in the
-        // simulator - use a mocked daemon instead.
-        logger.debug() << "Mocking daemon for:" << simDeviceName;
-        MockDaemon* daemon = new MockDaemon(qApp);
-        qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
-      }
+    else if (!qEnvironmentVariable("SIMULATOR_DEVICE_NAME").isEmpty()) {
+      // If we are running in the iOS device simulator, the network extension
+      // is not supported in this environment - use a mocked daemon instead.
+      MockDaemon* daemon = new MockDaemon(qApp);
+      qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
     }
 #endif
 
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -164,6 +164,18 @@ int CommandUI::run(QStringList& tokens) {
       MockDaemon* daemon = new MockDaemon(qApp);
       qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
     }
+#ifdef MZ_IOS
+    else {
+      QString simDeviceName = QString(qgetenv("SIMULATOR_DEVICE_NAME"));
+      if (!simDeviceName.isEmpty()) {
+        // The network extension doesn't actually start when running in the
+        // simulator - use a mocked daemon instead.
+        logger.debug() << "Mocking daemon for:" << simDeviceName;
+        MockDaemon* daemon = new MockDaemon(qApp);
+        qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
+      }
+    }
+#endif
 
     MozillaVPN vpn;
     logger.info() << "MozillaVPN" << Constants::versionString();
--- a/src/models/devicemodel.cpp
+++ b/src/models/devicemodel.cpp
@@ -71,19 +71,20 @@ bool DeviceModel::fromSettings(const Keys* keys) {
 
 namespace {
 
-bool sortCallback(const Device& a, const Device& b, const Keys* keys) {
-  if (a.isCurrentDevice(keys)) {
-    return true;
-  }
-
-  if (b.isCurrentDevice(keys)) {
-    return false;
-  }
-
+bool sortCallback(const Device& a, const Device& b) {
   return a.createdAt() > b.createdAt();
 }
 
 }  // anonymous namespace
+
+// Check if the current device exists, and if so move it to to the front.
+void  DeviceModel::moveCurrentDevice(const Keys* keys) {
+  for (qsizetype index = 0; index < m_devices.length(); index++) {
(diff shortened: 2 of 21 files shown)


src/commands/commandui.cpp
```diff
@@ -164,6 +164,18 @@ int CommandUI::run(QStringList& tokens) {
       MockDaemon* daemon = new MockDaemon(qApp);
       qputenv("MVPN_CONTROL_SOCKET", daemon->socketPath().toLocal8Bit());
     }
+#ifdef MZ_IOS
+    else {
+      QString simDeviceName = QString(qgetenv("SIMULATOR_DEVICE_NAME"));
+      if (!simDeviceName.isEmpty()) {
```
mcleinman, line 170:
```suggestion
      bool isSimDevice = !QString(qgetenv("SIMULATOR_DEVICE_NAME")).isEmpty;
      if (isSimDevice) {
```

Nit: I find it more readable when it looks something like this (here and in other spots we check for sim). I'm a tad concerned a far-future reader will not immediately understand that this is checking just for iOS simulator (and not generally checking for iOS), but it's up to you.

src/models/devicemodel.cpp
```diff
@@ -71,20 +71,21 @@ bool DeviceModel::fromSettings(const Keys* keys) {
...
 
-bool sortCallback(const Device& a, const Device& b, const Keys* keys) {
-  if (a.isCurrentDevice(keys)) {
-    return true;
-  }
-
-  if (b.isCurrentDevice(keys)) {
-    return false;
-  }
-
```
mcleinman, line 82: Why remove this, out of curiousity?

src/models/devicemodel.cpp
```diff
@@ -71,20 +71,21 @@ bool DeviceModel::fromSettings(const Keys* keys) {
...
+bool sortCallback(const Device& a, const Device& b) {
   return a.createdAt() > b.createdAt();
 }
 
 }  // anonymous namespace
 
+// Check if the current device exists, and if so move it to to the front.
+void DeviceModel::moveCurrentDevice(const Keys* keys) {
```
mcleinman, line 81: To confirm, this was required for simulator to run as well?

mcleinman (changes requested):
I'm getting this build failure when building for sim:
```
Building for 'iOS-simulator', but linking in dylib (/Users/mcleinman/code/mozilla-vpn-client/build-ios/qtglean/unified/debug/QtGleanBindings.framework/QtGleanBindings) built for 'iOS'
```

I'm happy to pair on this if you want - may be easiest.

src/models/devicemodel.cpp
oskirby: There is a static assert in `std::sort` that requires the sort comparator to satisfy strict-weak ordering. It is implemented by invoking, `compare(a, a)` and `compare(b, b)` both of which must always evaluate to false. When compiled for debug we hit this assert and the app crashes.

I am not sure if this is just a simulator thing, or an Xcode 16 thing, but I figured it was better to fix it.

oskirby: The fix was implemented by sorting on creation time, and then add a second step which looks for the current device and moves it to the head of the list.

oskirby:
> I'm getting this build failure when building for sim:
> 
> ```
> Building for 'iOS-simulator', but linking in dylib (/Users/mcleinman/code/mozilla-vpn-client/build-ios/qtglean/unified/debug/QtGleanBindings.framework/QtGleanBindings) built for 'iOS'
> ```
> 
> I'm happy to pair on this if you want - may be easiest.

I find that switching from device-to-simulator (or vice-versa) i have to delete the CMake build dir and start again. Once built for one, or the other, it becomes sticky and will only build that way.

I think the underlying cause is that static libraries we build ourselves (eg: the rust stuff, wireguard-go, and sentry) get the target platform baked into them and result in this kind of linker error. There are a handful of ways to tackle this:
- We go out of our way to force the simulator to target `x86_64`, build device as normal targeting `arm64` and then merge the simulator and device builds together as an Apple universal binary (this is the approach taken by [Qt](https://bugreports.qt.io/browse/QTBUG-101276), by the way).
- We build the library using native CMake (eg: `add_library(libname <sources>)` and rely on the Xcode generator to figure it out.
- We build the library as a framework, which I think has some way to distinguish device vs. simulator builds for the same target arch.

oskirby:
I dunno how much of a pain the simulator-switching experience really is but I figured the fix is too big of a thing to tackle in this PR, and at least it's less broken than before :)

mcleinman:
My error was starting from a blown-away build folder for iOS - so starting from scratch. I suppose it wasn't a fully cleaned repo, I can try again with a fully clean repo.

Agreed that switching can be saved for another PR.

src/models/devicemodel.cpp
oskirby: This provides a replacement for the code that was removed from `DeviceModel::sortCallback()` which was intended to ensure the current device is always first in the list. This was necessary to fix the static assert that I was running into and causing my simulator builds to crash. The assert may not have been fatal on device builds, but they were still occuring.

mcleinman:
I'm still getting errors when building for simulator, when building the iosglean target:

```
Ld /Users/mcleinman/code/mozilla-vpn-client/build-ios/src/Debug-iphonesimulator/IOSGlean.framework/IOSGlean normal (in target 'iosglean' from project 'Mozilla VPN')
...
ld: building for 'iOS-simulator', but linking in dylib (/Users/mcleinman/code/mozilla-vpn-client/build-ios/qtglean/unified/debug/QtGleanBindings.framework/QtGleanBindings) built for 'iOS'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

My steps: 
1. blow away the build-ios folder
2. comment out the xlink lines in `rustlang.cmake` (as that was causing an error)
3.  run `qt-cmake -S . -B build-ios -DCMAKE_OSX_SYSROOT=$(xcrun --sdk iphoneos --show-sdk-path)`
4. change the build settings to use the normal architectures
5. build for an iPhone 16 simulator

I'm going to send you a meeting invite so we can look at this syncronously.

mcleinman (changes requested):
I'm not sure how I feel about adding yet another state here. Let's all talk it through in our next engineering meeting? Or if you feel like this PR is getting stale, we can schedule a meeting next week if enough of us are around.

oskirby:
I started this workstream with the intention of getting the iOS simulator working and it kind of got subsumed by trying to get builds working in Xcode 16. This has kind of grown to the point where it's no longer clear what the point of this PR is. I think the better path forward is to split this PR into two distinct chunks of work so that they can be reviewed separately.

PR #10097, opened 2024-12-06
Window Polish on Windows

strseb:
The Designers aren't looking quick lets change some things. 

IMHO RN our Window on Windows has multiple issues: 

| Dark  | Light |
|-------|-------|
| old_dark | old_light 

❌ The Titlebar shows a very pixelated icon
❌ The Icon is not the one in our [figma windows "window"](https://www.figma.com/design/lHbmTJ5xaTFWhI8MMYyLVH/Key-screen---Windows?node-id=0-1&node-type=canvas&m=dev)
❌ The Titlebar is not responsive to neither theme / sys prefrence
❌ The Titlebar is not using the same background as the client app

Whenever this opens on windows it looks ... cheap, especially the title bar not beeing part of the client space, looks outdated. It's giving cross plattform electron app.  
So today i needed a fun break. 

| Dark  | Light |
|-------|-------|
|  new_dark  | new_light | 

So this pr adds a theme listener, that: 
✅  Updates the Titlebar to the Themes BG color 
✅  Render icon from the SVG, respecting the coloscheme
✅  Set that icon for the window only.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,8 @@ qt_add_executable(mozillavpn MANUAL_FINALIZATION)
 
 mz_target_handle_warnings(mozillavpn)
 
+find_package(Qt6 REQUIRED COMPONENTS Svg)
+
 target_link_libraries(mozillavpn PRIVATE
     Qt6::Quick
     Qt6::Test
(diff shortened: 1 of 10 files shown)


lesleyjanenorton (approved):
YES!

PR #10098, opened 2024-12-06
Remove unused includes

strseb:
## Description
clangd complains a lot about not used imports, so i went through some to clean that up or scope it to the platform where it is needed.


src/settingsholder.h
```diff
@@ -9,8 +9,9 @@
 #include <QObject>
 #include <QVariant>
 
-#include "constants.h"
-#include "feature/feature.h"
+// Used in the ettingslist.h
```
mcleinman, line 12:
```suggestion
// Used in the settingslist.h
```

mcleinman (approved):
Thank you!!!

PR #10099, opened 2024-12-06
qtglean bindings must match min version

mcleinman:
## Description

After https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10078 was merged in, Xcode Cloud reported failures because `ITMS-90208: Invalid Bundle - The bundle MozillaVPN.app/Frameworks/QtGleanBindings.framework does not support the minimum OS Version specified in the Info.plist.`

Updating the minimum version to match.

## Reference

VPN-6742 follow up

## Checklist
    
### DIFF
--- a/scripts/cmake/Info.plist.QtGleanBindings
+++ b/scripts/cmake/Info.plist.QtGleanBindings
@@ -17,6 +17,6 @@
     <key>CFBundleVersion</key>
     <string>1</string>
     <key>MinimumOSVersion</key>
-    <string>14.0</string>
+    <string>15.0</string>
 </dict>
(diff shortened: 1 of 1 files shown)


oskirby (approved):

oskirby:
For what its worth, this file should probable be generated using CMake's `configure_file()` command like we do for the rest of the `Info.plist` files in the project, in which case we can consistently set the versions and app identifiers from `CMAKE_OSX_DEPLOYMENT_TARGET` and `BUILD_IOS_APP_IDENTIFIER`

But this is fine for now.

PR #10141, opened 2024-12-20
VPN-6318: Add reauth to subscription management flow

mcleinman:
## Description

My understanding of this ticket is that once we switch from in app auth to web based auth, the call to get subscription details will fail due to insufficient scopes. This PR adds a flow when this call fails - if it fails due to authentication and the user is using web-based authentication, we show a popup prompting the user to reauthenticate on web.

If we ever get a bad response for getting subscription details for the account profile screen and it's a `AuthenticationRequiredError`, this screen will pop up. This is what we want in the transition to web, but if this error pops up other times... it isn't the worst to try to reauthenticate, so I think this is okay. 

## Testing 
Since you need a specific token to trigger this (if I understand the ticket correctly), I hacked together a test setup:
1. Add [this diff](https://github.com/user-attachments/files/18216359/sample.txt) into the codebase. (It will force a reauth when running getsubscriptiondetails - but only once [for each app launch] and only when using web-based auth.)
2. Turn InAppAuthentication on (it defaults to `off` in staging) via Developer->Features menu. 
3. Sign out.
4. Sign back in (via in app authentication)
5. Turn InAppAuthentication back off
6. Hard quit and relaunch the app
7. In Settings, tap the Account row. You should get the reauth experience!

Video of it in process: https://github.com/user-attachments/assets/f213739a-53fe-4190-a9d6-c3245f0fa6b2

## Reference

VPN-6318

## Checklist
    
### DIFF
--- a/src/profileflow.cpp
+++ b/src/profileflow.cpp
@@ -80,7 +80,7 @@ void ProfileFlow::start() {
   });
 
   connect(task, &TaskGetSubscriptionDetails::mustTransitionAuthToWeb, this,
-          [this]() { setState(StateNeedsWebReauthentication); });
+          [this]() { setState(StateAuthenticationNeeded); });
 
   TaskScheduler::scheduleTask(task);
   m_currentTask = task;
--- a/src/profileflow.cpp
+++ b/src/profileflow.cpp
@@ -97,11 +97,12 @@ void ProfileFlow::reauthenticateViaWeb() {
             setState(StateError);
             ProfileFlow::reset();
           });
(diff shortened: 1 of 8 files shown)


flodolo (approved):

oskirby:
So, what's the difference between `StateNeedsWebReauthentication` and `StateAuthenticationNeeded` when inAppAuthentication is disabled? Just a little unclear on why we need a new state for this.

oskirby (approved):
Looks good

mcleinman:
Merging because this test failure is unrelated to the PR, and we have another ticket to address the test failure.

PR #10150, opened 2024-12-23
Maybe fix flatpak build checkout failures

oskirby:
## Description
I have noticed a lot of Flatpak build failures lately, which seem to end with a failure to checkout the Mozilla VPN sources, but the build never seems to fail on the `main` branch. I suspect that this is due to an inability to find the merge commit SHA in whatever build worker is running the build. We might have better luck using `github.event.pull_request.head.sha` instead, which is the latest commit on the PR branch rather than the merge commit.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -30,10 +30,12 @@ jobs:
       - name: Update git reference
         shell: bash
         working-directory: linux/flatpak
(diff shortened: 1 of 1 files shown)


oskirby:
I think this might have actually been fixed by #10153

PR #10152, opened 2024-12-23
FXVPN-32 follow up: Add extension badge for addons

mcleinman:
## Description

New badge type for addons. (I missed this when initially implementing the spec for the message.) Since this addon message will only be shown to users on the latest version of the client (2.25.0), we can be assured that all users who see this message will have a client that can actually show this badge.

Screenshot 2

## Reference

FXVPN-32

## Checklist
    
### DIFF
--- a/src/ui/screens/messaging/ViewMessage.qml
+++ b/src/ui/screens/messaging/ViewMessage.qml
@@ -103,6 +103,8 @@ MZViewBase {
                             return badgeInfo.surveyBadge
                         case MZAddonMessage.Subscription:
                             return badgeInfo.subscriptionBadge
+                        case MZAddonMessage.Extension:
+                            return badgeInfo.subscriptionBadge
                         }
                     }
 
@@ -135,6 +137,11 @@ MZViewBase {
                             'badgeBackgroundColor': MZTheme.colors.normalLevelBackground
                         };
                         property var subscriptionBadge: {
+                            'badgeText': MZI18n.InAppMessagingSubscriptionBadge,
+                            'badgeTextColor': MZTheme.colors.normalLevelMain,
+                            'badgeBackgroundColor': MZTheme.colors.normalLevelBackground
+                        };
+                        property var extensionBadge: {
                             'badgeText': MZI18n.InAppMessagingSubscriptionBadge,
                             'badgeTextColor': MZTheme.colors.normalLevelMain,
                             'badgeBackgroundColor': MZTheme.colors.normalLevelBackground
--- a/addons/message_try_firefox_extension/manifest.json
+++ b/addons/message_try_firefox_extension/manifest.json
@@ -15,7 +15,7 @@
     "id": "message_try_firefox_extension.24",
     "title": "vpn.tryFirefoxExtension.title",
     "subtitle": "vpn.tryFirefoxExtension.subtitle",
-    "badge": "new_update",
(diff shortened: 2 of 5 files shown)


src/ui/screens/messaging/ViewMessage.qml
```diff
@@ -139,6 +141,11 @@ MZViewBase {
                             'badgeTextColor': MZTheme.colors.normalLevelMain,
                             'badgeBackgroundColor': MZTheme.colors.normalLevelBackground
                         };
+                        property var extensionBadge: {
+                            'badgeText': MZI18n.InAppMessagingSubscriptionBadge,
```
oskirby, line 145: Shouldn't this be something like `MZI18n.InAppMessagingExtensionBadge` ?

oskirby (changes requested):

src/ui/screens/messaging/ViewMessage.qml
mcleinman: 🤦🏻 I'm not sure what happened here; I fixed it in both the spots that needed it.

oskirby (approved):
LGTM

mcleinman:
Merging because this test failure is unrelated to the PR, and we need to wait for l10n team to be back from their break to fix the broken test.

flodolo:
> Merging because this test failure is unrelated to the PR, and we need to wait for l10n team to be back from their break to fix the broken test.

Can you clarify this? The request before Christmas was to revert the target-language change on `extras.xliff`, and that was done right away: https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/7da43c55c078456cd2b14df3d7ec7a52848bc5e9

If the test is still failing, it can't be fixed in the l10n repo, unless it's about something else.

mcleinman:
> > Merging because this test failure is unrelated to the PR, and we need to wait for l10n team to be back from their break to fix the broken test.
> 
> Can you clarify this? The request before Christmas was to revert the target-language change on `extras.xliff`, and that was done right away: [mozilla-l10n/mozilla-vpn-client-l10n@7da43c5](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/7da43c55c078456cd2b14df3d7ec7a52848bc5e9)
> 
> If the test is still failing, it can't be fixed in the l10n repo, unless it's about something else.

Thank you! I hadn't realized you snuck it in before you left. It seems like this may not have been the culprit - I will investigate this further today.

mcleinman:
> > > Merging because this test failure is unrelated to the PR, and we need to wait for l10n team to be back from their break to fix the broken test.
> > 
> > 
> > Can you clarify this? The request before Christmas was to revert the target-language change on `extras.xliff`, and that was done right away: [mozilla-l10n/mozilla-vpn-client-l10n@7da43c5](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/7da43c55c078456cd2b14df3d7ec7a52848bc5e9)
> > If the test is still failing, it can't be fixed in the l10n repo, unless it's about something else.
> 
> Thank you! I hadn't realized you snuck it in before you left. It seems like this may not have been the culprit - I will investigate this further today.
> 
> UPDATE: No, we haven't had a clean i18n merge since then. So it may be the culprit still - merging and testing now.

For when folks are back from their breaks - I've left additional research and thoughts in the conversation on Slack.

PR #10157, opened 2024-12-30
Downgrade version number

mcleinman:
We're going to re-cut a new 2.25.0 release branch (likely later this week). While we could wait until the release branch is cut and then do a similar PR against the release branch, doing it against `main` (this PR) means we're not blocked on PR approvals and our lengthy tests during the release process. (After the release branch is cut, I'll put up a PR to revert this to 2.26.0 on `main`.)


oskirby (approved):
I have a sneaking suspicion this might cause some funky breakage in CI, but it'll be transient so this should be okay.

mcleinman:
I'm going to wait and merge this right before I create the new branch - thus we won't terrible break CI tests for other PRs coming in this week.

PR #10162, opened 2025-01-03
Improve Windows firewall logging

oskirby:
## Description
While investigating VPN-6293, it seems that the logging around the windows firewall kind of sucks and it would be nice if we could get a log of the error codes rather than just reporting than an error occurred. This refactors a bit of how the firewall gets set up so that we can log the codes.

While we're at it, I also cleaned up some of the firewall setup by using FWPM transactions and scope guards to handle the unhappy path.

## Reference
JIRA issue [VPN-6293](https://mozilla-hub.atlassian.net/browse/VPN-6293)

## Checklist
    
[VPN-6293]: https://mozilla-hub.atlassian.net/browse/VPN-6293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/windows/daemon/windowsfirewall.cpp
+++ b/src/platforms/windows/daemon/windowsfirewall.cpp
@@ -159,44 +159,51 @@ bool WindowsFirewall::initSublayer() {
 }
 
 bool WindowsFirewall::enableInterface(int vpnAdapterIndex) {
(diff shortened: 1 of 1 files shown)


strseb (approved):
👍

PR #10168, opened 2025-01-06
Disable broken wasm test

strseb:
## Description

For reasons i don't know the language page causes a dlopen , which is not supported. Sadly emscripten does not want to tell WHAT is trying to load :( so for now let's disable the test on wasm, given we have coverage from the other plattforms


mcleinman (approved):
Let's put something on those tickets about re-enabling this test once we upgrade Qt.

PR #10169, opened 2025-01-06
WIP WASM: Publish debug info 

strseb:
## Description
so i'm trying to hunt a bug. But it seems we have no debug info. Let's see if this publishes a usable dwarf file, so i can properly set breakpoints. 

Update: IT WORKS !! 

image


mcleinman (approved):
👍🏻

PR #10172, opened 2025-01-08
VPN-6799: Add tests for timeToMainScreen metric

oskirby:
## Description
There has been some concern that we might have a regression in our `timeToMainScreen` performance, which is intended to measure the startup time of the application. To make sure that this isn't some bug in the metric accumulation itself, we should add a functional test to ensure that it's generating sensible data.

Unfortunately, this turned into a bit of a mess. There was a hack in the code that used QML/JS to serialize the metric value for testing, but this fails when handling a histogram metric, like `TimingDistributionMetric` so this required some refactoring to make it go. In particular:
- Abandon the QML/JS implementation so that we can write our own metric serialization code in C++.
- Add a `BaseMetric` type to try and simplify how metrics are serialized, and give them a unified interface.
- Add an extra test to make sure that the `CustomDistributionMetric` we use for RX/TX transfer stats still work.

And after all that is done, we can now check the `timeToMainScreen` metric as follows:
```
let timing = await vpn.gleanTestGetValue("performance", "timeToMainScreen", "");
```

## Reference
JIRA issue: [VPN-6799](https://mozilla-hub.atlassian.net/browse/VPN-6799)
JIRA issue: [VPN-6186](https://mozilla-hub.atlassian.net/browse/VPN-6186)

## Checklist
    
[VPN-6799]: https://mozilla-hub.atlassian.net/browse/VPN-6799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcleinman (approved):
Looks great, thanks!

PR #10173, opened 2025-01-08
VPN-6581: disable adjust

mcleinman:
## Description

We want to disable adjust, but in a way that we can re-enable it if we later decide to. Commenting out this section removes the `MZ_ADJUST` flag in the builds, which disables adjust.

You can verify that Adjust is not running by finding the log line `Debug: Adjust not included in build.` shortly after the app initializes, rather than `Debug: Adjust included in build.`. (Build with a placeholder Adjust token when testing this; without one, Adjust will never be included in the build whether that section is commented out or not.)

## Reference

VPN-6581

## Checklist
    

strseb (approved):
🔥🔥🔥🔥

PR #10177, opened 2025-01-10
Nit: add browser path to log

strseb:
## Description
Looking at your convo in the ticket, maybe the browser is not excluded. 
Let's also log the path, so we can assert that.


oskirby (approved):
LGTM

PR #10178, opened 2025-01-10
VPN-6681 part 2: iOS and Android launch screens, Debug menu to show all images

mcleinman:
## Description

Finishing touches for asset-related dark mode work:
- Android and iOS now have launch screens that respect the system's color scheme. Important: This follows the system's dark mode/light mode, not our app's theming. (We cannot follow app theming, as launch screen is setup happens before app is fully open. Additionally, following the system's color mode preference makes sense for the launch screen, which is seen in the transition from the system to the app.) One consequence of this: This goes live immediately, and isn't behind a feature flag.
- The new debug option for an image playground now works - it shows a chosen image 4 times - against both light and dark background, and both backgrounds have the light and dark version of the image.
- The animation playground has been updated so the user can select one of several colors to use as the background.

## Reference

VPN-6681

## Checklist
    

android/resources/all/drawable/splash_background.xml
```diff
@@ -1,4 +1,4 @@
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
-    <item android:drawable="@android:color/black" />
+    <item android:drawable="@android:color/white" />
```
lesleyjanenorton, line 2: FWIW I don't think we really need changes to the splash screens since they're only seen for a moment.

lesleyjanenorton (approved):
R+ with one comment

src/ui/screens/getHelp/developerMenu/ViewImagesPlayground.qml
```diff
@@ -23,99 +25,127 @@ Item {
...
+                  sourceSize.height: 150
+                  sourceSize.width: 150
+                  fillMode: Image.PreserveAspectFit
+              }
+          }
+          
+            RowLayout {
+              spacing: 25
+              Rectangle {
+                height: 150
```
lesleyjanenorton, line 83: Weird QML quirk/gotcha: QML components that are inside ("children") of Layout components (RowLayout, ColumnLayout, AnythingLayout) need to use a different sizing rules. Not doing so can cause truly bizarre, completely unexpected behavior, a lot of console complaining, and even crashes. In this case `height` should be `Layout.preferredHeight` and `width` should be `Layout.preferredWidth`. Same thing for the height/width rules on L96-97.

mcleinman: Updated, thanks

PR #10189, opened 2025-01-14
Fix dependency for translation tools

mcleinman:
## Description

Translation tools was broken by this PR: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10180

From searching, it was becasue react and react-dom had different versions. For some reason it wasn't working when both were on 19.0.0, so I used the latest one that seems to work.

## Reference

N/A

## Checklist
    

strseb (approved):

tools/translationsreport/app/package.json
strseb, line 1: Nit: `^` -> dependabot will provide an update for the next version. let's just pin the version if we need it :)

PR #10190, opened 2025-01-14
VPN-6318 follow up: Fix for testing

mcleinman:
## Description

https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10141/ had a testing bug. This set of tasks results in being hung on the new "reauth" popup:
        - Reset app
        - Launch app - in debug menu use staging server, flip InAppAuth on (so that it is on for staging)
        - Relaunch app, sign in (via in app auth)
        - In Debug menu turn InAppAuth pref off
        - In inspector, give `force_subscription_management_reauthentication` command
        - Tap into the "Account" screen in settings

This was because we re-run profile flow after the reauth, but weren't clearing out the flag to force reauth before the second run. (In testing before putting up the original PR, I was testing via a different path.)

## Reference

VPN-6318

## Checklist
    

oskirby (approved):
Looks good

PR #10196, opened 2025-01-16
basti/nativeMessaging

strseb:
## Description

This PR "fixes" "the "bridge_pong" - the extension and client only expect that we pass json-objects through this impl, the native messaging part however expected a raw string. - This now gives nm the ability to handle commands similar to the client. 

Then i add a start command that launches on windows the client. Feedback welcome if there is a "standard" way to that on linux os'es too.

### DIFF
--- a/extension/bridge/src/commands.rs
+++ b/extension/bridge/src/commands.rs
@@ -26,17 +26,17 @@
          "bridge_ping" =>{
              crate::io::write_output(std::io::stdout(),&json!({"status": "bridge_pong"}))
                  .expect("Unable to Write to STDOUT?");
-             return Ok(true);
+             Ok(true)
          }
          "start" =>{
             let out = launcher::start_vpn();
             crate::io::write_output(std::io::stdout(),&out)
                 .expect("Unable to Write to STDOUT?");
-            return Ok(true);
+            Ok(true)
         }
          _ =>{
              // We did not handle this.
-             return Ok(false);
+             Ok(false)
          }
      }
  }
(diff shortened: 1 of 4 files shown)


extension/bridge/src/commands.rs
```diff
@@ -0,0 +1,77 @@
...
+             // We did not handle this.
+             return Ok(false);
+         }
+     }
+ }
+ 
+
+#[cfg(target_os = "windows")]
+mod launcher {
+    const CLIENT_PATH: &str = "C:\\Program Files\\Mozilla\\Mozilla VPN\\Mozilla VPN.exe";
```
mcleinman, line 47: is there any way it's ever installed to a different directory?

mcleinman (approved):
Approving this, I'd ask that another reviewer also look at this. I neither know rust nor this portion of the code base, so this is a very tentative approval.

extension/bridge/src/commands.rs
```diff
@@ -0,0 +1,78 @@
...
+        match result {
+            Ok(_) => json!("{status:'requested_start'}"),
+            Err(_) => json!("{error:'start_failed'}"),
+        }
+    }
+
+
+}
+
+#[cfg(not(target_os = "windows"))]
```
oskirby, line 72: After doing a bit of Googling, I think the way to handle this on Linux would be to use a deep-link to try and launch the VPN client. This could then be invoked using the `xdg-open` tool. A similar approach could also be done on Windows too, but the invocation would be different.

This is not a request to change anything here, we can tackle that in a future PR.

oskirby (approved):
I think this looks good.

PR #10215, opened 2025-01-23
WIP: Build VPN for aarch64-pc-windows-msvc

strseb:
## Description

TODO:
- [x] Builds on my machine (tm)
- [ ] Check Updated Split-Tunnel Driver
- [ ] Produces a Working MSI for aarch64
- [ ] Generate an ARM-MSVC toolchain on tc
- [ ] Cross Compile QT-aarch64 on tc
- [ ] Cross Compile Client on tc. 
- [ ] Firgure out how to deploy VCRedist without MSM.
- [ ] Setup Signer on tc.


lottie/lib/lottiestatus.h
```diff
@@ -49,12 +49,11 @@ class LottieStatus final : public QObject {
...
     m_error = true;
     m_errorString = errorString;
-    emit changed(m_playing, m_currentTime, m_totalTime, m_error, m_errorString);
+    emit changed();
   }
 
  signals:
-  void changed(bool playing, double currentTime, int totalTime, bool error,
-               const QString& errorString);
+  void changed();
```
strseb, line 56: So while at it i'm building against qt6.9-beta2 and moc throws an error here, as we never read this.

PR #10216, opened 2025-01-23
Give the extension the ability to get the parent process info

strseb:
## Description
Based on top of: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10196

## Reference

Right now we only check if the Split-Tunnel List includes `Firefox*` - but if i.e you have nightly excluded and use the extension on ESR, we still block out. 
Given that Firefox will spawn the nativemessaging process, we can ask it to check it's parent executable. 
This allows the extension to check if the Firefox used is excluded or not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,9 +378,9 @@ dependencies = [
 
 [[package]]
 name = "core-foundation-sys"
-version = "0.8.4"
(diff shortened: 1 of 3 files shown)


oskirby (approved):
LGTM

PR #10217, opened 2025-01-23
Nit: Push disabled app list

strseb:
## Description

RE; https://github.com/mozilla-extensions/mozilla-vpn-extension/pull/195

let's push the updated list to extension when it changes, so it can re-calculate if the current firefox is affected or not.


mcleinman (approved):
👍🏻

PR #10225, opened 2025-01-27
VPN-6632: Fix split tunnel when IPv6 unreachable

oskirby:
## Description
We have a bit of a bug in our handling of split-tunnelling when there is no IPv6 connectivity. In particular, if there are no IPv6 routers on the LAN then the network interface will only have link-local addresses, which cannot be used to enable the split tunneling driver. To ensure that split tunneling works on such networks, we need to take care when selecting addresses to avoid link-local addresses.

This PR rewrites ` WindowsSplitTunnel::getAddress()` so that we rank IPv6 addresses to select the address which is most likely to route to the internet, while also ignoring explicitly unroutable addresses such as link-local and loopback.

While we are at it, we also identified an endian conversion bug in the interface selection logic that would also sometimes result in invalid split tunneling configuration, and removed an unused firewall rule.

## Reference
Heavily inspired by #10001 
JIRA Issue: [VPN-6632](https://mozilla-hub.atlassian.net/browse/VPN-6632)

## Checklist
    
[VPN-6632]: https://mozilla-hub.atlassian.net/browse/VPN-6632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/windows/daemon/windowsdaemon.cpp
+++ b/src/platforms/windows/daemon/windowsdaemon.cpp
@@ -51,7 +51,8 @@ WindowsDaemon::~WindowsDaemon() {
 
 void WindowsDaemon::prepareActivation(const InterfaceConfig& config) {
   // Before creating the interface we need to check which adapter
-  // routes to the server endpoint
+  // routes to the server endpoint. This will be selected as the outgoing
+  // interface for split-tunnelled traffic.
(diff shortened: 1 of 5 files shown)


strseb (approved):
LGTM!!

PR #10238, opened 2025-01-30
VPN-6706: fix timer after server unavailable

mcleinman:
## Description

If a connection attempt results in a "server unavailable" modal, the user is encouraged to select a new server. However, that results in the timer being stuck at 0:00:00:
1) Since the first server never fully connected, the timer is never started
2) Since the second server is technically a server switch, the timer isn't touched when switching servers

Since this could be fragile, I added a test to make sure this doesn't break in the future. While writing that test:
- `forceServerUnavailable` got some parameters added to it, as the CLI command requires parameters (previously, this testing command was unused)
- I changed the stub server response to have different "public keys". We gray out unavailable servers based on their public key - so when we marked a server unavailable in our test, *all* servers were being marked unavailable because they shared a public key.
- I created a new helper function to prevent code duplication.

## Reference

VPN-6706

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -526,6 +526,12 @@ void Controller::activateInternal(
   activateNext();
 }
 
+void Controller::startTimerIfInactive() {
+  if (m_connectedTimeInUTC == QDateTime()) {
+    m_connectedTimeInUTC = QDateTime::currentDateTimeUtc();
+  }
+}
+
 void Controller::clearConnectedTime() {
   if (!isSwitchingServer) {
     m_connectedTimeInUTC = QDateTime();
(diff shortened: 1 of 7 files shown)


src/controller.cpp
```diff
@@ -526,6 +526,12 @@ void Controller::activateInternal(
   activateNext();
 }
 
+void Controller::startTimerIfInactive() {
+  if (m_connectedTimeInUTC == QDateTime()) {
```
oskirby, line 530: It would be a little more clear here to do `m_connectedTimeInUTC.isNull()` or `!m_connectedTimeInUTC.isValid()` here, because it's not exactly clear what equality to the default constructor means.

oskirby (approved):
Looks pretty good.

src/controller.cpp
mcleinman: Good call - updated this to be clearer. Thanks.

Issue #9989, opened 2024-10-27
Qt 6.8. Client Crashes on "Select Location"


strseb:
Hi @strseb 

I am not sure if the purpose of this PR is fix all the problems with Qt6.8, I applied this patch (also this one 9961) to the stable version, and the UI is usable, but one thing is not working, when I click in the list of servers "Select location", the UI crash with the following traceback:

[trace.txt](https://github.com/user-attachments/files/17532325/trace.txt)

_Originally posted by @XavierCLL in https://github.com/mozilla-mobile/mozilla-vpn-client/issues/9984#issuecomment-2439773442_

XavierCLL:
I rebuilt the latest stable version with the QT68 patches and this problem is gone. I think it is due to some regressions in QT68 and the Archlinux devs have been applying some fixes to the QT libraries, so I assume that with QT 6.8.1 this will work fine.

herzenschein:
I was trying the Mozilla VPN app on openSUSE and I got that crash. Then I updated my system and this got fixed.

After checking the backtrace and the openSUSE package for qt6-declarative (since the crash was happening in QtQuick), I managed to find out that the revision that fixed it in openSUSE was [this](https://build.opensuse.org/request/show/1221250), which adds the fix to https://bugreports.qt.io/browse/QTBUG-129500. It's a [well known](https://wiki.qt.io/Qt_6.8_Known_Issues) Qt 6.8 issue, and it matches the backtrace listed here (QQuickItem::mapToItem). If your distro backports the fix to 6.8.0 you'll get it, otherwise I think you should get the fix in 6.8.1.

strseb:
I think we can close this as this is a Qt issue fixed in 6.8.1. Feel free to reopen if you disagree.

PR #10256, opened 2025-02-05
VPN-6671 Add a warning for qt 6.8.0

strseb:
## Description
As seen in those issues: 
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/9989
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/9966

We do know it's broken, it's not our fault. 
All we can do here is throw a warning for folks to use 6.8.1 instead :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,7 @@ find_package(Qt6 COMPONENTS
     Xml
     Svg
 )
+include("scripts/cmake/check_qt_breakage.cmake")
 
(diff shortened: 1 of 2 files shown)


oskirby:
I'm curious about just how broken it really is on Qt 6.8, because I have Qt 6.8.1 on my Fedora 41 machine and it seems to be working fine.

strseb:
6.8.0 is broken and crashes the moment we try to get a bounding rect i.e in the serverlist to calc the view height. 
6.8.1 is already and fine, so this only would apply to 6.8.**0**

oskirby (approved):

PR #10257, opened 2025-02-06
FXVPN-377 Fix DNS Inclusion

strseb:
**Fix the DNS handling in StateOnPartial**

If you put a custom DNS that is not in your LAN, it did not get added to the allowed-ip list in state-on-partial. This caused our firewall rule to catch this and block it.


mcleinman (commented):
This makes sense. Just one question out of curiosity - why a new class for this?

strseb:
For tests :) - We have in all the unit-tests a `moccontroller.cpp` - So any code in controller.cpp is not compiled into that unit_test binary - it seemed simpler to just add a new aux namespace that can be tested vs a refactoring controller.cpp so we can compile that into there.

mcleinman (approved):
👍🏻

PR #10269, opened 2025-02-12
remove now-unused android signing script

bhearsum:
As far as I know, this used to be used to sign through autograph edge, which we no longer do AFAIK.


strseb (approved):
thanks!!

mcleinman:
FYI this script is also used to sign a release build that you've built locally and need to test. I'll be adding it back in shortly.

mcleinman:
Except... that URL we hit may no longer be active. I'll dig into this next week.

mcleinman:
Per chat w/ autograph team, going to leave this alone for now unless it becomes very painful to our team. If it does, we'll learn how to set up a local autograph environment to sign this.

PR #10271, opened 2025-02-13
VPN-6828 - Remove uneeded Window destruction

strseb:
## Description
Hiding and Re-rendering the window can cause the flicker seen in the issue. The best thing, that was completly unneeded. 

output


mcleinman (approved):
👍🏻

PR #10281, opened 2025-02-17
FXVPN-383 (?) Proxy: Enable TCP Keepalive

strseb:
## Description

When Excluding youtube i noticed quite a bit of buffering, as in the background fetching just stalls. Yt is loading this via http 1.1 so it is going to be holding a bunch of tcp connections open to re-use. The devtools list a lot of fetches aborting with "NS_BINDING_ABORTED". 
So i tried enabling keepalive and this problem goes away :)

### DIFF
--- a/extension/socks5proxy/src/socks5.cpp
+++ b/extension/socks5proxy/src/socks5.cpp
@@ -32,6 +32,10 @@ void Socks5::newConnection(T* server) {
     if (!socket) {
       return;
     }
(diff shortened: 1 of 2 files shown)


oskirby (approved):
This doesn't seem to have affected the buffering issue in my test scenarios, but it hasn't done anything to hinder performance or make anything worse, so I am okay with this change.

PR #10282, opened 2025-02-17
FXVPN-381: Let the proxy use the Local DNS-Server

strseb:
## Description
RN we currently are only using the System DNS. 
Which means we can't attribute a query to the originating client. Therefore our "block all non vpn dns" rule will block all attempts of not using the vpn-dns. 
To circumvent, let's do dns in-process, using c-ares. 

output


oskirby:
My initial test of this PR results in pretty significant breakage when attempting to exclude sites. Seems like I am not able to resolve addresses at all. I'll try to dig into why and post some updates.

lesleyjanenorton:
@strseb about to give this another run. @oskirby would you mind checking this out again as well?

lesleyjanenorton:
@strseb 

I'm getting a lot of 'proxy server is refusing connections' running this with the latest XPI... I'll keep tinkering.

PR #10286, opened 2025-02-19
Upgrade NodeJS toolchain

strseb:
## Description
Let's upgrade it to the Latest LTS.


.github/workflows/gh_pages.yaml
```diff
@@ -107,9 +107,6 @@ jobs:
         uses: actions/checkout@v4
 
       - uses: actions/setup-node@v4
-        with:
-          node-version: "16"
```
strseb, line 111: Removing those, as it should respect the one in package.json

PR #10288, opened 2025-02-19
VPN-6872: update message for 2.25.1

mcleinman:
## Description

Adding an update message for 2.25.1. Since 2.25.1 is just for Windows, we need to maintain a 2.25 update message for all other platforms. Thus, I duplicated the files involved (and added a new `platforms` condition to the 2.25 message).

We also needed to skip a test that catches if there are 2 update messages - but in this case (because of the different versions for different platforms) we want one.

While the 2.25.1 message could omit some things like the `"javascript": "osCheck.js"` condition, I decided to keep them - they don't hurt (except the most minor of a one-time performance hit, as we're running a few lines of unnecessary code), and I'm concerned if we remove lots of this we'll forget to add it back in when creating the messages for 2.26. (Typically, we don't change much when updating the update message.)

## Reference

VPN-6872

## Checklist
    
### DIFF
--- a/addons/message_update_v2.25.1/enable.js
+++ b/addons/message_update_v2.25.1/enable.js
@@ -1,0 +1,56 @@
+(function(api) {
+// "Extra" fields are used only to get a localized string.
+api.addon.composer.remove('extra_1');
+
+if (('updateTime' in api.settings)) {
+  api.addon.date = (api.settings.updateTime.getTime() / 1000);
+}
+
+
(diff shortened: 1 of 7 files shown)


lesleyjanenorton (approved):
thank you!

strseb (approved):

mcleinman:
Flatpak test is an existing test failure unrelated to this PR; merging with the failed test.

PR #10291, opened 2025-02-20
Match Qt 6.6 NDK Version

strseb:
## Description
All i wanted to change is: 

```
const val ndkVersion = "23.1.7779620"
```
to 
```
const val ndkVersion = "25.1.8937393"
```
So that we match the NDK version our QT builds are using. 

tldr; Lots of build tech dept. 

------

So here is what i had to do: 
output

#### Upgrading the NDK bricks our build of wireguard-go. 
We currently invoke `externalNativeBuild` which calls cmake to compile this [cmake lists](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/android/daemon/tunnel/CMakeLists.txt) which then calls [make to run this ](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/android/daemon/tunnel/libwg-go/Makefile) which then downloads go1.16 and builds a binary with cgo. Something there breaks with an arcane error. Honestly i had no clue to fix this. 
So i ripped that all out and replaced it with [gomobile](https://go.dev/wiki/Mobile) which according to the gowiki is the current "good" way to build a go library and get FFI bindings for free. This produces an `.aar` archive we can "just" drop into our build system. (also this should now make the android client buildable on windows as that was the blocking thing)
Bonus we can rip out a lot of glue code. 

#### Using gomobiles aar inside an aar is forbidden in gradle 
We now have an .aar file that we want to include in `/daemon` which also is an android library. Gradle does not allow this. 
We splitted the android project into multiple aar's purely for the case we might want to android test the daemon. This has not manifested. Sooooo that needs to go. Therefore i have flattened all internal aar's into the root `app`. 
This was only cosmetic as in the end gradle put them all in the same classpath anyway. 

#### Adjusting Glean for the flattened dependencies
We currently call to generate the Glean bindings twice: one for the front-end aar and the daemon-aar. As they are now in the same package, i just unified that. 

#### Upgrading go bricks wireguard-go on mac
gomobile forces at least go23 however on mac we get 
```
[task 2025-03-02T20:11:33.331Z] # golang.zx2c4.com/wireguard
[task 2025-03-02T20:11:33.331Z] link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
```
This pr is already waay to big. So i did split .env into env-android and use that, removing all android things from the env.

#### Fixing ktlint. 

Now with the .aar's now having a strong dependency on  the qt generated code by `androiddeploy-qt` it's not trivial to run ktlint anymore. `cd android/ && gradlew ktlintformat` does not work. 
So i'm ripping that out and use ktlint from conda. While at it we're upgrading this. This means we need to format a few files. 
I will do that after this was reviewed, because otherwise this pr will become utterly unreadable 💀


hneiva (commented):
CI bits ✅

lesleyjanenorton:
> Something there breaks with an arcane error. Honestly i had no clue to fix this.

Reminds me of the time our c code stopped talking to our Balrog go code.

PR #10304, opened 2025-02-25
FXVPN-381: Naomi's extra socks proxy hacking

oskirby:
## Description
Working on some extra fixes and workarounds to try and improve the socks proxy's streaming performance.

I found that it's not sufficient to resolve the DNS server once on startup as the device can switch networks and change DNS servers, which meant that it's better if we move the DNS server selection into the notification hooks already built into `windowsbypass.cpp` and to make this work, I had to refactor the `DNSResolver` class a bit.

## Reference
Based on PR #10282
JIRA Issue: [FXVPN-381](https://mozilla-hub.atlassian.net/browse/FXVPN-381)

## Checklist
    

strseb:
> I found that it's not sufficient to resolve the DNS server once on startup as the device can switch networks and change DNS servers, which meant that it's better if we move the DNS server selection into the notification hooks already built into `windowsbypass.cpp` and to make this work, I had to refactor the `DNSResolver` class a bit.

Currently the PR is resolving the nameserver [once per Socks5Connection](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/basti/local_proxy_dns/extension/socks5proxy/src/socks5connection.cpp#L290), not one per start :) 
(which might be still excessive but does not seem to be *that* costly )

oskirby:
> Currently the PR is resolving the nameserver [once per Socks5Connection](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/basti/local_proxy_dns/extension/socks5proxy/src/socks5connection.cpp#L290), not one per start :) (which might be still excessive but does not seem to be _that_ costly )

Not quite, the `static QHostAddress localAddr` at the start of the `DNSServerLookup::getLocalDNSName()` means that it only gets resolved once upon first use, and then sticks around indefinitely.

strseb:
Oh right i forgot i added that, shame on me! SORRY.

lesleyjanenorton:
@oskirby Is this ready to come out of Draft?

oskirby:
Just changing the target from `basti/local_proxy_dns` to `main` and once the CI passes I think we are good to give this one a merge.

PR #10306, opened 2025-02-27
FXVPN-381: Fix proxy flow control bug

oskirby:
## Description
At last, I was able to chase down the cause of my poor streaming performance. It turns out there's a bug in the proxy flow-control logic that was able to get a socket stuck into a deadlocked state under heavy unidirectional inbound traffic loads, such as streaming.

The bug is caused by getting the arguments transposed in the call to `QTcpSocket::proxy()` in which case when the TCP sockets are full, and a `bytesWritten()` signal occurs, we incorrectly try to proxy data in the wrong direction. If this doesn't cause any traffic to flow, it can result in a situation where no other signal will be emitted to unblock the connection.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

strseb (approved):
amazing find !!

PR #10308, opened 2025-02-27
VPN-6842: invalid server fix

mcleinman:
## Description

Ultimately this is pretty simple - We had been assuming that switching the config only happened when the VPN was active. However, when the server/city wasn't available, the VPN was never activated - and thus switching the server was happening when no VPN was active, and getting us into this terrible state

Unfortunately, I wasn't able to figure out precisely how this is happening. I can confirm that when we send the update message from `iosController` to the `iostunnel`, it eventually sends an `update` message to `WireGuardAdapter` - which immediately returns an `.invalidState` back to `iostunnel` (because the adapter wasn't in a `started` state), which in turn doesn’t do anything for the `ioscontroller`. `Even if we run a `disconnectOnErrorCallback` in this situation it doesn’t fully fix the issue. I am not quite sure how either iOS or Wireguard's non-standard (for iOS) packet filtering is getting activated in this situation - which I think is the source of this issue, that one half (but not the other).  is getting activated.

This behavior - try to turn on VPN, essentially bork the iOS device's network capabilities until rebooting (or removing the VPN in system settings) - has been seen in many similar corner case situations. It's my hope that this fix stamps out this class of bugs.

Added some logging in `iostunnel.swift`. Also a small cleanup in `inspector.md`.

## Reference

VPN-6842

## Checklist
    
### DIFF
--- a/docs/Troubleshooting/inspector.md
+++ b/docs/Troubleshooting/inspector.md
@@ -31,7 +31,6 @@ On desktop, use `ws://localhost:8765`.
     3. Note the iOS device's IP address. (Settings app -> Wi-Fi -> ⓘ by network name. See "IP Address" in "IPv4 Address" section)
 3. On the desktop computer, go to the [inspector page](
 https://mozilla-mobile.github.io/mozilla-vpn-client/inspector/).
-On desktop, use `ws://localhost:8765`.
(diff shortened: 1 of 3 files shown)


oskirby (approved):
LGTM

PR #10309, opened 2025-02-27
VPN-6884: remove MZCard and streamline some colors

mcleinman:
## Description

This PR does a few things:
- Removes `MZCard`, as it doesn't seem like it's been used in a while. If this seems like a poor idea, please let me know - it's not a must-do (it was a side effect of looking at colors we could streamline, because it allows us to...)
- Removes the `color.card` color object, as it was only being used in `MZCard`.
- Moves `bgColorTransparent` to `theme-derived`, as we don’t need to explicitly pick that (it comes from `bgColor`).
- Renames `radioButtonDisabledHovered` to `disabledButtonHovered` and renames `radioButtonDisabledPressed` to `disabledButtonPressed` to reflect the fact they’re used in more than just radio buttons - they’re used in many disabled buttons.
- In one spot, updates `colors.inputHighlight` to `colors.divider`. They're currently the same actual color (`grey10`), and this spot is definitely a divider.
- Specifies a background color for `MZInputBackground` - without this, several things like drop down menus and search bars were always showing a white background.
- Removes two explicit colors which slipped through the color re-engineering. (In one case we use a named color, in the other case I removed the color because I couldn't find any difference in behavior without it.) It also adds a test for this sort of color, to make sure it doesn't happen again. 

## Reference

VPN-6884

## Checklist
    

nebula/ui/components/CMakeLists.txt
```diff
@@ -19,7 +19,6 @@ qt_add_qml_module(components
          MZButton.qml
          MZCallout.qml
          MZCancelButton.qml
-         MZCard.qml
```
lesleyjanenorton, line 22: :wave: :tada:

lesleyjanenorton (approved):
👍 love it. thank you!

PR #10317, opened 2025-03-04
[Proxy] Produce a sorted dnsserver list by metric

strseb:

PR #10332, opened 2025-03-11
socksproxy: Fix localhost DNS and use-after-free crash

oskirby:
## Description
In continuing my testing with the socksproxy and the DNS resolution woes, I came across a couple of bugs that can lead to breakage.

The first issue is that when selecting DNS servers, when the first choice is a link-local IPv6 address, we don't set a scope ID in the address, which can prevent `c-ares` from being able to connect to it. We can fix this just by calling `QHostAddress::setScopeId()` when building the list of DNS servers. While implementing this, I also performed a bit of cleanup by moving the DNS server selection into `WindowsBypass::updateNamserver()`

The second issue, which was a bit trickier to track down, is that we have a use-after-free bug in the socks proxy that can occur between the `c-ares` resolver thread and the Qt thread. This happens when the `Socks5Connection` is closed before DNS resolution finishes, in which case the `arg` provided to the c-ares callback points to freed memory. To resolve this, we need to verify that the `QObject` is valid before invoking the `onResolutionFinished()` method. We add a `QHash` map to store the objects that requested resolution, and use the `QObject::destroyed()` signal to track when they are garabge collected. 

## Reference
Based upon PR #10317 by @strseb
JIRA Issues:
- [VPN-6891](https://mozilla-hub.atlassian.net/browse/VPN-6891)
- [VPN-6890](https://mozilla-hub.atlassian.net/browse/VPN-6890)
- [VPN-6874](https://mozilla-hub.atlassian.net/browse/VPN-6874)

## Checklist
    
[VPN-6891]: https://mozilla-hub.atlassian.net/browse/VPN-6891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-6890]: https://mozilla-hub.atlassian.net/browse/VPN-6890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-6874]: https://mozilla-hub.atlassian.net/browse/VPN-6874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extension/socks5proxy/bin/windowsbypass.cpp
+++ b/extension/socks5proxy/bin/windowsbypass.cpp
@@ -17,8 +17,6 @@
 #include <QScopeGuard>
 #include <QSettings>
 #include <QUuid>
-#include <algorithm>
-#include <ranges>
 
 #include "socks5.h"
 #include "winutils.h"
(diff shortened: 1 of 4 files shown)


mcleinman (approved):
This is far from my area of expertise, but from what I can see it looks good. Thanks!

Issue #10344, opened 2025-03-19
Stop running Taskcluster tasks on github-release events


ahal:
We now release through Shipit, so we shouldn't be running tasks here anymore. This will save resources and avoid confusion.

PR #10345, opened 2025-03-19
ci: stop running tasks on 'github-release' events

ahal:
We currently schedule a graph upon publishing Github releases, but this is an artifact of the old days. These days releases are managed through Shipit. We should turn these graphs off to save resources and reduce confusion.

There were also release indexes that were pointing at the `github-release` graph. This was bad because builds at these indexes were not actually what was being shipped. I adjusted this so relpro tasks will populate this index instead.

## Reference

#10344

## Checklist
    

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
```diff
@@ -58,13 +61,17 @@ def add_index_routes(config, tasks):
...
-            branch = branch.replace("/",".") # We can't have "/" in the index -> so releases/2.9.0 needs to be releases.2.9.0
+                branch = branch[len(_GIT_REFS_HEADS_PREFIX) :]
+            branch = branch.replace(
+                "/", "."
+            )  # We can't have "/" in the index -> so releases/2.9.0 needs to be releases.2.9.0
             context["branch"] = branch
 
-        elif config.params["tasks_for"] == "github-release":
+        elif config.params["tasks_for"] == "action" and config.params.get(
+            "shipping_phase"
```
jcristau, line 71: Are we actually making new builds as part of relpro graphs?

.taskcluster.yml
```diff
@@ -96,7 +96,6 @@ tasks:
         tasks_for in ["action", "cron"]
         || (isPullRequest && pullRequestAction in ["opened", "reopened", "synchronize"])
         || (tasks_for == "github-push" && (head_branch == "refs/heads/main" || head_branch[:19] == "refs/heads/releases"))
-        || (tasks_for == "github-release" && releaseAction == "published")
```
jcristau, line 99: Consider removing taskcluster/test/params/github-release.yml as well.

taskcluster/mozillavpn_taskgraph/transforms/release_index.py
ahal: Oh hmm.. You mean to say that we aren't making new builds as part of relpro, but we were making them as part of `github-release`, so now builds will be missing from this index?

Yeah.. seems like that might be the case. I suppose the signed builds should be good enough?

jcristau: Do we know if anything's using that index?  (I think any consumers should be pointed at https://archive.mozilla.org/pub/vpn/releases/, and this `RELEASE_INDEX_ROUTES` removed)

ahal: @mcleinman any thoughts on ^ ?

mcleinman: Apologies, which index is this one in question?

jcristau: This is about things like https://firefox-ci-tc.services.mozilla.com/tasks/index/mozillavpn.v2.mozilla-vpn-client.release.v2.26.0.latest

mcleinman: I think that is fine. (I just checked - There are two things we use that for in the RC process, but we can get those files from the commit checks instead.)

strseb (approved):

ahal:
Sorry, I neglected to land this. It still applies clean, so landing now.

PR #10366, opened 2025-04-02
VPN-6178: messages for 2.27

mcleinman:
## Description

Create addon messages for 2.27, and update a line of a readme to make a needed step explicit.

It's possible that the dark mode feature may not hit the 2.27 release. If that happens, we'll change these messages to the default update messages (already translated), and hang onto this new string for use in the 2.28 release notes (without renaming the `227updateMessage` key, as we don't want to cause additional work for translators).

Screenshot 96Screenshot 95

## Reference

VPN-6178

## Checklist
    

lesleyjanenorton (approved):

addons/strings.yaml
```diff
@@ -81,6 +81,11 @@ commonStrings:
     value: On Android, added step-by-step instructions to enable VPN connection on startup.
     comment: Second bullet point with a specific update in 2.25
 
+227updateMessage:
+  bullet1:
+    value: Dark mode - enjoy a sleek new look that’s easy on the eyes, day or night.
```
flodolo, line 86:
```suggestion
    value: Dark mode — enjoy a sleek new look that’s easy on the eyes, day or night.
```

flodolo (approved):
r+ with they em dash fixed (which monospace doesn't show…)

PR #10380, opened 2025-04-08
Replace hot.js with a qt app

strseb:
- Add cpp based hotreloader
- Remove hot.js and deps
- run npm prune to update package-lock.json
- Remove Navigatorreloader class -> This was only used while we also had the remote-content reloader which was never on prod. The InspectorHotreloader always uses flags Reload all, so this had no effect. 
- Update docs

## Description

This pr re-implements hot.js only using qt, removing some dependencies while at it :)


mcleinman (approved):
🚀🚀🚀 

Thanks you!

PR #10382, opened 2025-04-08
Remove buildsteps for the inspector

strseb:
## Description

The only dependencies the inspector had were xterm + lit. The rest was written in raw esm, so no bundling needed really.
We can get lit using an import map. Tried with xterm but i just replaced it with just 2 textboxes lmao. 

Therefore we can remove all that build tooling.

Note: 
While testing it seems the logs / network recording is broken, but it seems to have been broken in release aswell :/


oskirby (commented):
Shouldn't there be a change to the top-level `package-lock.json` too in order to reflect the removed dependencies?

mcleinman (approved):
👏🏻👏🏻👏🏻 Thank you

PR #10393, opened 2025-04-11
VPN-6971: Implement theme detection fallback on Linux via XDG portal

oskirby:
## Description
It's been reported that Linux doesn't correctly detect the system theme on some Qt versions, and some have reported it broken altogether. So, let's throw together a quick workaround by pulling the theme information from the [XDG Settings portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html).

## Reference
JIRA issue: [VPN-6971](https://mozilla-hub.atlassian.net/browse/VPN-6971)

## Checklist
    
[VPN-6971]: https://mozilla-hub.atlassian.net/browse/VPN-6971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -562,7 +562,9 @@ SETTING_STRING(theme,          // getter
 // Below Qt 6.5, this option won't be available.
 // Without this, the Apperance screen will
 // initially show no radio button selected.
-#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+#if defined(MZ_LINUX)
+#  define USING_SYSTEM_THEME_DEFAULT_VALUE true
+#elif QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
 #  define USING_SYSTEM_THEME_DEFAULT_VALUE true
 #else
 #  define USING_SYSTEM_THEME_DEFAULT_VALUE false
--- a/src/settingslist.h
+++ b/src/settingslist.h
@@ -559,9 +559,9 @@ SETTING_STRING(theme,          // getter
                false           // sensitive (do not log)
 )
 
-// Below Qt 6.5, this option won't be available.
-// Without this, the Apperance screen will
-// initially show no radio button selected.
+// This option is only supported for Qt 6.5 or later, with a special fallback
+// for Linux users via the XDG settings portal.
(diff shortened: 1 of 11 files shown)


src/settingslist.h
```diff
@@ -562,7 +562,9 @@ SETTING_STRING(theme,          // getter
 // Below Qt 6.5, this option won't be available.
 // Without this, the Apperance screen will
```
mcleinman, line 563: Update this comment to mention Linux?

src/theme.cpp
```diff
@@ -164,28 +175,19 @@ void Theme::setCurrentTheme(const QString& themeName) {
...
-#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
   if (usingSystemTheme) {
     setToSystemTheme();
   }
-#endif
 }
 
-#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
 void Theme::setToSystemTheme() {
   if (!Feature::get(Feature::Feature_themeSelection)->isSupported()) {
```
mcleinman, line 184:
```suggestion
  if (!Feature::get(Feature::Feature_themeSelection)->isSupported() || !Feature::get(Feature::Feature_themeSelectionIncludesAutomatic)->isSupported()) {
```

Since we're removing the Qt version check, we need another safety check here so we don't try to update this. I know we have the safety check down below in `currentSystemTheme`, but without this early return this can still lead to weird corner cases - for instance, if you turn the theme selection feature on, set it to "Dark", then turn the feature back off - I believe on the next launch this would reset it to light mode.

mcleinman (changes requested):
A couple small things on the theme side. The comment is a nit, but the early return seems important.

The Linux code generally looks okay from what I can tell, but that is not my specialty and my review focused on (and is more confident about) the shared code.

src/settingslist.h
oskirby: I can update the comment, sure.

mcleinman (approved):
Thanks!

Issue #10375, opened 2025-04-07
Build failure with QT 6.9.0: LottieStatus::changed signal expects 5 arguments, 0 provided


XavierCLL:
While building MozillaVPN version 2.26.0 using the latest version of Qt and GCC from Arch Linux, the build fails during the compilation of the Lottie component.

```bash
/run/media/xavier/SSD-Data/ArchLinux/AUR-packages/mozillavpn/src/build/lottie/lottie_autogen/ZJOKMXDKB7/../../../../mozillavpn-2.26.0/lottie/lib/lottiestatus.h:56:8: note: candidate: ‘void LottieStatus::changed(bool, double, int, bool, const QString&)’
   56 |   void changed(bool playing, double currentTime, int totalTime, bool error,
      |        ^~~~~~~
/run/media/xavier/SSD-Data/ArchLinux/AUR-packages/mozillavpn/src/build/lottie/lottie_autogen/ZJOKMXDKB7/../../../../mozillavpn-2.26.0/lottie/lib/lottiestatus.h:56:8: note:   candidate expects 5 arguments, 0 provided
make[2]: *** [lottie/CMakeFiles/lottie.dir/build.make:101: lottie/CMakeFiles/lottie.dir/lottie_autogen/mocs_compilation.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:3993: lottie/CMakeFiles/lottie.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 23%] Building CXX object src/webextension/CMakeFiles/mz_webextension.dir/connection.cpp.o
[ 23%] Built target formsplugin_init
[ 23%] Building CXX object src/webextension/CMakeFiles/mz_webextension.dir/server.cpp.o
   Compiling signature v0.2.0 (/run/media/xavier/SSD-Data/ArchLinux/AUR-packages/mozillavpn/src/mozillavpn-2.26.0/signature)
    Finished `release` profile [optimized] target(s) in 43.47s
[ 23%] Built target signature_builder
[ 23%] Linking CXX static library libmz_webextension.a
[ 23%] Built target mz_webextension
make: *** [Makefile:146: all] Error 2
```

OS: Arch Linux (fully updated)
MozillaVPN version: 2.26.0
Compiler: GCC 14.2.1
QT: 6.9.0

hansdampfinger666:
[...]

PR #10420, opened 2025-04-16
Qt 6.9 compat: Provide default args for LottieStatus::changed so moc can call it

strseb:
Looking at: https://github.com/mozilla-mobile/mozilla-vpn-client/issues/10375

it seems moc generates a function call that tries to call it without arguments. Given all our callsites are providing one, we can just add defaults so moc can happily compile :)


mcleinman (approved):
👍🏻

PR #10422, opened 2025-04-16
Qt 6.9 compat:  Remove Qt.labls.qmlmodels usage

strseb:
## Description
While fixing https://github.com/mozilla-mobile/mozilla-vpn-client/issues/10375 - i noticed anothing thing beeing utterly broken by qt 6.9.0. 

qmllmodels moved from qt.labs to the qt namespace but i could not see a way to conditionally do an `import` if we're on 6.9.+ - 
Now given all that did was switch colors of the button i'm taking your tricks using enums and just remove that dependency.


mcleinman (approved):
Looks great, tests great. Thanks.

PR #10440, opened 2025-04-23
VPN-6945: Fix crash in MacOSController

oskirby:
## Description
In PR #10358 we introduced a new mechanism for registering the VPN daemon, but due to my inexperience with Obj-C and how it handles object lifecycles - I inadvertently left a use-after-free bug in the `MacOSController` class.

This crash would only manifest in release builds, and it basically comes down to the fact that the `SMAppService` object is automatically released at the end of the function scope, so attempting to retain it as a class member causes the use-after-free bug. The most straightforward fix is to simply recreate the object in each call to `checkServiceStatus()`

## Reference
Bug introduced by #10358 

## Checklist
    
### DIFF
--- a/src/platforms/macos/macoscontroller.h
+++ b/src/platforms/macos/macoscontroller.h
@@ -13,7 +13,6 @@ class MacOSController final : public LocalSocketController {
 
  public:
   MacOSController();
-  ~MacOSController();
 
   void initialize(const Device* device, const Keys* keys) override;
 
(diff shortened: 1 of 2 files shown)


oskirby:
Actually, upon further debugging. I think we need a bit more work here. There's a couple of edge cases I didn't consider.

mcleinman (commented):
Great catch, thanks.

nit: Why are we converting from an NSString to a QString right before returning from the function, only to immediately convert it back to an NSString at the callsite? Can we just keep it an NSString?

Edit: Just realized you closed this out. Sorry.

PR #10463, opened 2025-04-29
Add setuptools to requirements.in

suvayu:
## Description

The glean parser script uses `pkg_resources`, and running the glean parser script fails. 

According to the release notes for Python 3.12[^1], `setuptools` is not included in a virtual environment by default any more and should be explicitly included.

I don't know what would be the appropriate version, the latest version worked for me.  So I used that.

[^1]: https://docs.python.org/3/whatsnew/3.12.html

## Checklist
    
### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,6 +2,7 @@ aqtinstall==3.2.0
 glean-parser==13.0.0
 oathtool==2.3.2
 pyrsistent==0.20.0
+setuptools==80.0.0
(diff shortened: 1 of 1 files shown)


suvayu:
Hi @strseb, thanks for the comment, agree with your analysis.  I would be fine with it if you pass along this fix upstream.  I only compile the VPN client since it's not packaged for Fedora.

PR #10465, opened 2025-04-29
Allow the extension to trigger a minimized start

strseb:
## Description
I'm currently hacking on the extension's "let's start the vpn when entering PBM". 
Given we switch to on-partial mode, nothing visually happens in the client when we activate the extension. 

So it's a bit off-putting that if you then open a private browsing mode a vpn windows appears in the center of your screen, empty, deactivated. Like it's judging you to activate it. 

image

But we actually already did this the moment you see the screen, so let's just allow the extension to start the vpn in the background 🥳

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ __pycache__
 lottie.mjs
 qml_lint_result.txt
 .qmlls.ini
-release_notes.txts
+release_notes.txt
 
 # Translation pipeline
 addon_ts/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -32,6 +32,7 @@ linux/netfilter/vendor/
 __pycache__
 lottie.mjs
 qml_lint_result.txt
+.qmlls.ini
```
strseb, line 35: I've switched to qt.6.9.0 and the qml language server drops those everywhere we have qml files. let's ignore those too,

Issue #10442, opened 2025-04-23
VPN UI gets broken by screen-rotation on Pixel Tablet


dholbert:
**Describe the bug**
Mozilla VPN's UI seems to expect a portrait-oriented orientation, but if a user's device is configured to auto-rotate nonetheless, it ends up breaking the Mozilla VPN UI

**VPN version:**
  2.27.0

**Affected Platforms:**
Android 15 (Google Pixel Tablet)

**Prerequisites:**
Android "Auto-rotate" setting enabled

**Steps to reproduce:**
1. Hold your tablet in landscape orientation.
2. Open Mozilla VPN (freshly installed, i.e. not authenticated). Notice that it's "letterboxed" to render in a in a portrait-oriented area in the center of your screen.
3. Rotate your screen (i.e. manually turn your device) to portrait orientation.
4. Rotate your screen (i.e. manually turn your device) back to landscape orientation.

**Expected result:**
UI should re-render nicely after the change in orientation.

**Actual result:**
- When I change to portrait orientation (step 3), the "Already a Subscriber" link is nearly clipped off the bottom of the screen for some reason.  (It's lower down than it would be if I were to launch the app while already holding my device in portrait orientation)
- When I change back to landscape orientation (step 4), most of the app's UI is clipped; I just see a cropped view of the top-left corner of the app.

**Additional notes:**
I'll post screenshots in subsequent comments here.

dholbert:
I get the same problem after I've signed in, too.  In that case, **the purple footer-bar with "Home|Messages|Settings" icons becomes permanently inaccessible as soon as I've rotated my device**.

Screenshots (at the same phases as in my previous comment, except now after having logged in):

![Image](https://github.com/user-attachments/assets/95cfee32-08e7-4fb4-b7f1-65419820db8e)

![Image](https://github.com/user-attachments/assets/c0c2ce9c-fe32-4566-b11f-72c96c14fd4c)

![Image](https://github.com/user-attachments/assets/5928b646-6574-43e6-bb8a-b8cd6cd653fb)

PR #10467, opened 2025-04-30
Android: Remove orientationlock

strseb:
## Description

> On [Android 12 (API level 31)](https://developer.android.com/about/versions/12/12L/summary#dev-device-orientation-request) and higher, device manufacturers can configure individual device screens (such as the tablet-sized screen of a foldable) to ignore this suggestion, and instead force an activity to be letterboxed within the user's preferred orientation of the device.
## Checklist

If we look at #10442 we're clearly in that letterboxing case that seems to fuck stuff up. 
My current hypothesis is that while we're letterboxed, qt still assumes the full width+height of the landscape screen


strseb:
https://github.com/user-attachments/assets/38a1bcc0-dd95-4b2e-b67d-38b4f1d0ceae

Yea, qt is fucked up by the letterboxing. If we assert that we support orientation, stuff will work :)

mcleinman (approved):
👍🏻

PR #10481, opened 2025-05-06
VPN-6945: Use XPC Service for daemon communication

oskirby:
## Description
This work extends the new `MacOSController` class to use a new XPC service for communication with the daemon. This gives us the ability to apply codesigning enforcement of the clients that connect to the service and has tighter integration with `launchd`. This should allow us to grant greater control over access to the daemon without relying on UID and session tracking like the local socket server does.

Much of the pain in figuring this out had to do with the threading model of XPC. It seems like XPC listeners and connections run in their own thread, which of course lacks a Qt event loop. This meant that methods and signals all needed to go through some kind of thread synchronization step to get into and out of the Qt world. For the most part, this is pretty easy using Qt's signals and meta-methods, but it took a bit of banging my head against the wall to figure it out.

As written, this requires the connecting clients to be codesigned with either a `Developer ID` certificate or an apple developer account, which could make it a bit more difficult for development.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
After some rebasing and splitting refactoring out into separate PRs. I am going to close and re-open this as PR #10494

PR #10487, opened 2025-05-07
Nit: Support Apple native string types in Logger

oskirby:
## Description
This adds Apple's native string types (`NSString*` for objective-C and it's C equivalent `CFStringRef`) to the `Logger` class so that they can be logged without needing to cast and convert them around. And while we're at it, use the new logger methods to cleanup a bunch of iOS and macOS code.

I keep find myself cherry-picking this patch between PRs. So maybe its probably time to split this into its own PR.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -54,6 +54,13 @@ Logger::Log& Logger::Log::operator<<(QTextStreamFunction t) {
   return *this;
 }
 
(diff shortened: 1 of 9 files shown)


mcleinman (approved):
Thanks! (Remember to add a Jira ticket for this if you want.)

PR #10494, opened 2025-05-12
VPN-7041: Use XPC Service for daemon communication

oskirby:
## Description
This work extends the new MacOSController class to use a new XPC service for communication with the daemon. This gives us the ability to apply codesigning enforcement of the clients that connect to the service and has tighter integration with launchd. This should allow us to grant greater control over access to the daemon without relying on UID and session tracking like the local socket server does.

Much of the pain in figuring this out had to do with the threading model of XPC. It seems like XPC listeners and connections run in their own thread, which of course lacks a Qt event loop. This meant that methods and signals all needed to go through some kind of thread synchronization step to get into and out of the Qt world. For the most part, this is pretty easy using Qt's signals and meta-methods, but it took a bit of banging my head against the wall to figure it out.

As written, this requires the connecting clients to be codesigned with either a Developer ID certificate or an apple developer account, which could make it a bit more difficult for development.

## Reference
Repopened from #10481
JIRA Issue: [VPN-7041](https://mozilla-hub.atlassian.net/browse/VPN-7041)

## Checklist
    
[VPN-7041]: https://mozilla-hub.atlassian.net/browse/VPN-7041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/macos/daemon/xpcdaemonserver.mm
+++ b/src/platforms/macos/daemon/xpcdaemonserver.mm
@@ -11,6 +11,7 @@
 #include "logger.h"
 #include "platforms/macos/macosutils.h"
 #include "platforms/macos/xpcdaemonprotocol.h"
+#include "version.h"
 
 constexpr const int XPC_SESSION_MAX_BACKLOG = 32;
 
@@ -96,6 +97,10 @@ shouldAcceptNewConnection:(NSXPCConnection *) newConnection {
 
 - (void)deactivate {
   QMetaObject::invokeMethod(self.daemon, "deactivate");
+}
+
+- (void)getVersion: (void (^)(NSString *))reply {
+  reply([NSString stringWithUTF8String:APP_VERSION]);
 }
 
 - (void)getStatus:(void (^)(NSString *))reply {
--- a/src/platforms/macos/daemon/xpcdaemonserver.mm
+++ b/src/platforms/macos/daemon/xpcdaemonserver.mm
@@ -1,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "xpcdaemonserver.h"
+
+#include "leakdetector.h"
+#include "logger.h"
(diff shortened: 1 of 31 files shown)


src/platforms/macos/macoscontroller.mm
```diff
@@ -74,51 +117,203 @@
...
-      LocalSocketController::initialize(nullptr, nullptr);
+    if ([service status] != SMAppServiceStatusEnabled) {
+      // We continue waiting for the service to start. The user likely needs
+      // to grant the daemon permission to run.
+      return;
     }
   }
+
+  // Create an XPC connection to the daemon.
+  QString daemonId = MacOSUtils::appId() + ".daemon";
```
mcleinman, line 151: We use this in at least 2 places. Make a helper function for this?

src/platforms/macos/daemon/xpcdaemonserver.mm
```diff
@@ -0,0 +1,278 @@
...
+
+  // Connections to the daemon require codesigning
+  // TODO: It would be nice if we could turn this off for developers somehow.
+  if (@available(macOS 13, *)) {
+    constexpr const char* oidExtAppleCodesign =
+        "(certificate leaf[field.1.2.840.113635.100.6.1.13] or" \
+        " certificate leaf[field.1.2.840.113635.100.6.1.12])";
+    constexpr const char* oidExtAppleCaIntermediate =
+        "(certificate 1[field.1.2.840.113635.100.6.2.6] or" \
+        " certificate 1[field.1.2.840.113635.100.6.2.1])";
```
mcleinman, line 62: Where is this coming from?

mcleinman (approved):
Overall looks good. A few bikeshedding nits, including:

If this requires development changes, please update the readme to match current steps needed to run the client.

src/platforms/macos/daemon/xpcdaemonserver.mm
mcleinman: Thanks for the detailed explanation. Maybe drop a link to this documentation (and one for 6.1.12) in the comments so it's clear to future maintainers where these came from?

PR #10497, opened 2025-05-13
VPN-6941: Update OSX permission screen to match SUMO

oskirby:
## Description
I overstepped during the UX design of the OSX permissions screen and we should have taken the SUMO article and Figma as the absolute source of truth and now it's a problem that the UX in the client doesn't match our support articles. The client should be updated to match these designs **exactly**

## Reference
JIRA Issue: [VPN-6941](https://mozilla-hub.atlassian.net/browse/VPN-6941)
Figma design: [link](https://www.figma.com/design/aqZq6eYrgAthI6BU3Soasr/macOS-Permissions-Update?node-id=97-157)
SUMO article: [link](https://support.mozilla.org/kb/allow-vpn-run-background-macos)

## Checklist
    
### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -1053,15 +1053,15 @@ permissionMacos:
   title:
     value: Allow VPN to run in background
     comment: Title of the screen directing macOS users to allow the VPN to run in background
-  body:
-    value: To use Mozilla VPN, please allow it to run in the background.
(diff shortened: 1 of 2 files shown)


oskirby:
The updated screen should look like the following:

[
Screenshot 2025-05-13 at 8 54 05 AM
](url)

PR #10503, opened 2025-05-15
Submit Telemetry Extras 

strseb:
In case of an event we did not submit the extras object we might have been sending from the extension


mcleinman (approved):
👍🏻

PR #10508, opened 2025-05-16
VPN-6945: Add SMAppService registration retries

oskirby:
## Description
Another entry in the endless saga of the MacOS daemon upgrade. This time we put some debugging energy into macOS13 to see what happens.

It seems like there is a bit of an API behaviour difference on macOS 13 vs. later operating systems. In particular `registerAndReturnError` can fail when permission has been denied, so we actually need to retry `regsiterAndReturnError` until we get a successful return code and the status of the service is `SMAppServiceStatusEnabled`

## Reference

## Checklist
    
[VPN-6945]: https://mozilla-hub.atlassian.net/browse/VPN-6945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/macos/macoscontroller.h
+++ b/src/platforms/macos/macoscontroller.h
@@ -19,11 +19,12 @@ class MacOSController final : public LocalSocketController {
   void getBackendLogs(QIODevice* device) override;
 
  private slots:
-  void checkServiceEnabled();
+  void registerService();
 
  private:
   NSString* plist() const;
 
+  bool m_permissionRequired = false;
   QTimer m_regTimer;
 };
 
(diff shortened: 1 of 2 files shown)


mcleinman (approved):
👍🏻

oskirby:
I've given this a test on both macOS 13/Ventura and 14/Sonoma and it seems to be working great. I don't have a macOS 15/Sequoia machine available yet so I think I'm just going to merge this as-is.

PR #10510, opened 2025-05-16
Nit: Fix universal builds of mozillavpnnp under Xcode

oskirby:
## Description
This is mostly just developer pain, but when performing a universal build under Xcode we find that the `cargo` tool in the path may not necessarily be the one installed from Conda, in which case it can fail to locate the Rust standard libraries for the cross-compiled targets.

To fix this, we basically just need to use the full path to the Conda-installed `cargo` and `rustc` tools. Fortunately, we already did all this in the `rustlang.cmake` scripts and we just need to use the variables they already defined.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

strseb (approved):
Nice!

PR #10514, opened 2025-05-19
Bugfix: ICWYU - Include windows.h

strseb:
## Description

I just upgraded my build machine to Qt 6.9.x 
we use stuff from windows.h but we got that from a QT import


oskirby (approved):

mcleinman (approved):
👍🏻

PR #10520, opened 2025-05-20
VPN-7042: Fix shutdown/upgrade deadlock

oskirby:
## Description
There is something of a deadlock that can occur during shutdown of the VPN client if the backend service is also unavailable. This can manifest on a macOS machine during package upgrade (or when triggered manually with `sudo pkill -x "Mozilla VPN"`).

The sequence of events that lead to the deadlock goes something like this:
1. The VPN client receives a `SIGTERM` signal and requests the Controller class to quit. The controller, is already disconnected so this is a no-op and the controller class emits `readyToQuit()`
2. The `MozillaVPN` class receives `readyToQuit` and starts the shutdown process. During this time, it flushes the task queue and calls `MozillaVPN::disconnect(true)` which requests the controller to disconnect and blocks until complete. This is done using a `TaskControllerAction` and manually running the `QEventLoop` until disconnected.
3. Meanwhile, the daemon terminates and the client catches this event in the `LocalSocketController` - this causes a background service error and emits a `disconnected()` signal. 
4. The controller handles the `disconnected` signal, finds that the next steps is `Quit` and emits another `readyToQuit()` signal.
5. The `MozillaVPN` class receives `readyToQuit` and starts the shutdown process again. It flushes the task queue, destroying the `TaskControllerAction` from step 2, leading the `QEventLoop` in `MozillaVPN::disconnect()` to run indefinitely because it never detects the task completion. This deadlocks the client and prevents it from terminating.

Steps 3-5 can repeat multiple times as the `LocalSocketController` continuously retries the connection to the daemon.

To fix this issue, we add some additional guard rails to prevent the deadlock from occurring by:

- The `Controller` class only sets the `m_nextStep = Quit` when it actually needs to wait for a signal.
- We update `TaskControllerAction::run()`so that it can run synchronously to completion when no signals are needed, this minimizes the window for race conditions to sneak in.

Then, to try and minimize the chances of triggering the race condition in existing clients, we limit the use of `pkill` during installation to target just the GUI applications (the daemon's will be shutdown later anyways by `launchctl bootout` anyways).

## Reference
JIRA Issue: [VPN-7042](https://mozilla-hub.atlassian.net/browse/VPN-7042)

## Checklist
    
[VPN-7042]: https://mozilla-hub.atlassian.net/browse/VPN-7042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/macos/pkg/scripts/postinstall.in
+++ b/macos/pkg/scripts/postinstall.in
@@ -35,7 +35,7 @@ exec 2>&1 > $LOG_DIR/postinstall.log
 echo "Running postinstall at $(date)"
 echo "Installing Mozilla VPN to ${APP_DIR}"
 
-pkill -x "Mozilla VPN" || echo "Unable to kill GUI, not running?"
+pkill -U "${USER}" -x "Mozilla VPN" || echo "Unable to kill GUI, not running?"
 sleep 1
 
 ## Before MacOS 13, take extra care to prevent tampering with the app.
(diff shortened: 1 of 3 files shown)


mcleinman (approved):
Incredible find. This looks good to me.

PR #10551, opened 2025-06-02
VPN-7054 reuse adapters if not properly removed

strseb:
## Description
We create a new adapter, in as we assume the old one was properly closed. In case we have a leftover i.e when the removal was not complete, we will create a 2nd one, colliding in the name, causing "mozilla vpn #2" i.e to appear. 
Let's attempt to re-use that adapter.

### DIFF
--- a/src/platforms/windows/daemon/wireguardutilswindows.cpp
+++ b/src/platforms/windows/daemon/wireguardutilswindows.cpp
@@ -12,13 +12,11 @@
 #include <ws2ipdef.h>
 
 #include <QFileInfo>
-#include <iostream>
 
 #include "leakdetector.h"
 #include "logger.h"
 #include "platforms/windows/windowscommons.h"
 #include "platforms/windows/windowsutils.h"
-#include "windowsdaemon.h"
 #include "windowsfirewall.h"
 #include "windowsroutemonitor.h"
 #include "wireguard.h"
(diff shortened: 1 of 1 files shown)


src/platforms/windows/daemon/wireguardutilswindows.cpp
```diff
@@ -421,7 +430,7 @@ QList<WireguardUtils::PeerStatus> WireguardUtilsWindows::getPeerStatus() {
   if (!m_adapter) {
     return {};
   }
-  DWORD bufferSize = 1024;
+  DWORD bufferSize = 2048;
   auto buffer = std::array<uint8_t, 2048>{};
```
strseb, line 434: the size was not the size of the buffer, not causing bugs (given we announce less space), that should be the same.

src/platforms/windows/daemon/wireguardutilswindows.cpp
```diff
@@ -12,13 +12,11 @@
 #include <ws2ipdef.h>
 
 #include <QFileInfo>
-#include <iostream>
 
 #include "leakdetector.h"
 #include "logger.h"
 #include "platforms/windows/windowscommons.h"
 #include "platforms/windows/windowsutils.h"
-#include "windowsdaemon.h"
```
strseb, line 21: Those imports were unused, so nuking them while at it

PR #10552, opened 2025-06-02
Lock down where we can load dll's at runtime

strseb:
## Description

We should only need to call loadlibrary with things in our app path and system32, really.


oskirby (approved):
LGTM

PR #10553, opened 2025-06-02
WIP: Fix reset&quit on android

strseb:
## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    

src/controller.cpp
```diff
@@ -1115,6 +1116,15 @@ bool Controller::deactivate(ActivationPrincipal user) {
...
 }
 
+QFuture<void> Controller::deactivateFuture(ActivationPrincipal user) {
+  auto const future =
+      QtFuture::connect(m_impl.get(), &ControllerImpl::disconnected);
+  if (!deactivate(user)) {
+    return QtFuture::makeReadyVoidFuture();
+  }
+  return future;
+}
```
mcleinman: Nit: Don't make `future` if not needed
```suggestion
  if (!deactivate(user)) {
    return QtFuture::makeReadyVoidFuture();
  }
  auto const future =
      QtFuture::connect(m_impl.get(), &ControllerImpl::disconnected);
    return future;
}
```

strseb: That was intentional :) - I.e the impl might emit disconnected during the call to disconnected() - if the future does not exst then, we will not catch it and have a permanently dangling future.

PR #10571, opened 2025-06-11
VPN-7070: Split the macOS daemon into a separate binary

oskirby:
## Description
We have a startup bug that I _think_ was introduced by the XPC service support in PR #10494. In particular, it seems that when trying to launch the GUI application, the operating system now mistakes the XPC service as another instance of the application and attempts to bring it into the foreground rather than starting another instance of the GUI. This fails because:
1. the XPC service is a `root` process and the user lacks sufficient privileges, and
2. it's a background process and has no UI to bring into the foreground.

The correct solution is to split the daemon into a separate binary tool so that it can be better distinguished from the GUI process. 

## Reference
JIRA Issue: [VPN-7070](https://mozilla-hub.atlassian.net/browse/VPN-7070)
Bug introduced by: #10494

## Checklist
    
[VPN-7070]: https://mozilla-hub.atlassian.net/browse/VPN-7070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
Marking this as ready for review now - seems to fix the issue.

Only remaining quirk is that it changes the path to the logfile from `/var/log/mozillavpn/mozillavpndaemon.log` and I am not sure how I feel about that, or what the cleanest way to change that would be.

PR #10573, opened 2025-06-13
Mzutils move server models

oskirby:
## Description
Some additional cleanup and refactoring cleans up external dependencies and moves the `Server`, `ServerCity`, `ServerCountry` and `ServerCountryModel` models into the `mzutils` library. In order to achieve this it was necessary to move the connection scoring algorithm out of the `ServerCity` model and into `ServerLatency` instead... then to make the score viewable in QML again we make the `connectionScore` a settable property in `ServerCity`

## Reference

    i.e Jira or Github issue URL

## Checklist
    

src/serverlatency.cpp
```diff
@@ -315,6 +314,52 @@ void ServerLatency::setLatency(const QString& pubkey, qint64 msec) {
...
+    if (rtt > 0) {
+      avgLatencyMsec += rtt;
+      numLatencySamples++;
+    }
+  }
+  if (numLatencySamples > 0) {
+    avgLatencyMsec += (numLatencySamples - 1);
+    avgLatencyMsec /= numLatencySamples;
+  }
+  city.setLatency(avgLatencyMsec);
```
mcleinman, line 340: I may be being dense, but I don't get the math here. Would it be clearer to just have a `totalLatencyMSec` variable?

```suggestion
    if (rtt > 0) {
      totalLatencyMsec += rtt;
      numLatencySamples++;
    }
  }
  if (numLatencySamples > 0) {
    avgLatencyMsec = totalLatencyMsec/numLatencySamples;
  }
  city.setLatency(avgLatencyMsec);
```

oskirby: Nothing has changed with the math here, the method was just moved from `ServerCity` into `ServerLatency`

As for what the math is doing, the step of:
```
avgLatencyMsec += (numLatencySamples - 1);
avgLatencyMsec /= numLatencySamples;
```

Because this is all done with integer math rather than floating-point values, this causes the average latency to round up to the next integer. Otherwise, the naive approach of `avgLatencySamples /= numLatencySamples` would round down to the nearest integer. This has the advantage that this calculation can't produce an `avgLatencyMsec` value of zero unless there are no latency samples measured.

mcleinman: Gotcha. I think I missed the forest for the trees - latency ms is going to be much higher than 1 (typically), and so.... yeah, I get it now. Thanks.

mcleinman (approved):

PR #10588, opened 2025-06-16
Refactor: Remove Simple-Networkmanager

strseb:
## Description
We have 2 classes that can create the QNetworkAccessManager neeeded for networking to function, the QMLEngine or the SimpleNetworkManager. 
This PR changes this so that we only have networkmanager which creates and owns the QNetworkAccessManager, removing the need for overloads or creating a snm to run tests :) . 
QMLEngineHolder will simply inject that QNetworkAccessManager into the QmlEngine.


src/mozillavpn.cpp
```diff
@@ -2275,7 +2274,6 @@ int MozillaVPN::runCommandLineApp(std::function<int()>&& a_callback) {
   QCoreApplication app(CommandLineParser::argc(), CommandLineParser::argv());
 
   Localizer localizer;
-  SimpleNetworkManager snm;
```
oskirby, line 2278: So, my understanding is that the `SimpleNetworkManager` class is used to provide a `NetworkManager` instance to the CLI tools that don't instantiate a QML engine, so that they can still make network requests. Does that use case still work?

Or maybe the bigger question, is it really worth the effort to keep the CLI tools around?

strseb: NetworkManager gets created on the first call to `NetworkManager ::instance()`
A QNetworkAccessManager  gets created on the first call to to `NetworkManager::networkAccessManager()`

So for CLI apps, things should be the same. The big change is that we no longer let QMLEngine create a QNetworkAccessManager, we instead point it to `NetworkManager::networkAccessManager()`

PR #10589, opened 2025-06-16
VPN-7002 fix - always load location

mcleinman:
## Description

This does 2 things:
- The bundle-related addons recheck their conditions on location updates.
- Activating on boot waits for location check. (Note: The server prioritization still does not run in this case

This works in my testing, both when "on boot" is active and when it is not.

## Reference

VPN-7002

## Checklist
    

oskirby (approved):
I think this looks good.

PR #10594, opened 2025-06-17
VPN-7090: Update macOS signers to use hardned_signing

oskirby:
## Description
After splitting the macOS GUI and daemon into separate binaries, we encountered a signing problem with the old `mac_sign_and_pkg_vpn` behaviour. In particular, it was trying to embed the same entitlements into every signed binary, but without a matching provisioning profile this results in an un-runnable daemon. What we really want is to sign the daemon with no entitlements.

I'm sure this could be solved by modifying `scriptworker` to carve out yet another exception for the VPN project, but we could also just switch to using the Firefox `mac_sign_and_pkg_hardened` behavior, in which case the entitlements and provisioning profiles can be configured as parameters in the task payload. This is much cleaner and allows us to add stuff to the bundle without having to update the `scriptworker` project.

Along the way, we also split out the generation of the macOS installer package into a separate taskcluster job, `repackage-macpkg` and did a bit of refactoring to the `macos/pkg` sub-project in order to enable this.

## Reference
JIRA Issue: [VPN-7090](https://mozilla-hub.atlassian.net/browse/VPN-7090)
JIRA Issue: [VPN-7091](https://mozilla-hub.atlassian.net/browse/VPN-7091)

## Checklist
    
[VPN-7090]: https://mozilla-hub.atlassian.net/browse/VPN-7090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/macos/pkg/CMakeLists.txt
+++ b/macos/pkg/CMakeLists.txt
@@ -8,7 +8,7 @@ project(mozillavpn-macpkg VERSION 1.0.0
         DESCRIPTION "MacOS Package for the Mozilla VPN"
 )
 
-# Determine the root of the project checkout - we scripts and localization.
+# Determine the root of the project checkout - we need scripts and localization.
 if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     set(PROJECT_ROOT_DIR ${CMAKE_SOURCE_DIR})
 else()


taskcluster/kinds/repackage/kind.yml
```diff
@@ -47,23 +49,29 @@ tasks:
 
     macpkg:
         description: repackage mozillavpn macpkg
-        run-on-tasks-for: [github-push]
+        run-on-tasks-for: [github-pull-request, github-push]
         fetches:
             fetch:
                - macos-cmake
-            signing:
+            from-dependency: # Updated to match dependencies in repackage.py
```
bhearsum, line 56: I wouldn't suggest addressing it here, but if you wanted to, an alternative solution to this would be to make signing work for PRs. We do this in other places (notably Gecko), but providing a separate pool of production grade workers that signs with non-production certificates. Doing this would simplify this repackage logic, and improve what PR testing covers.

taskcluster/config.yml
```diff
@@ -142,6 +142,31 @@ release-promotion:
 scriptworker:
     scope-prefix: project:mozillavpn:releng
 
+mac-signing:
```
bhearsum: I know we put this in the graph config in Gecko, but in your case it might be better to put into the kind, seeing as you only have one usage of it. (And honestly, in Gecko's case it probably belongs in a transform rather than the graph-level config.)

Totally up to; keeping it as is is fine too.

bhearsum (approved):
The taskgraph parts look fine to me, and clearly work on a basic level. If you'd like to test it more fully, we can [add another L3 branch](https://github.com/mozilla-releng/fxci-config/blob/0a56f7627894a00899c9a01ff5d4b7d6510be5c8/projects.yml#L1079-L1082) that you can push this work to before this merges. (If you want to just test it on `main` after merging I have no objections.)

If possible, it would be good to have another VPN developer review the non-taskgraph parts, and the signing configuration you're passing.

taskcluster/kinds/repackage/kind.yml
oskirby: I would love that actually, especially if we could make the output of PR's installable by QA and authorized developers (eg: by signing with a WWDR/Developer certificate).

bhearsum: This sounds a lot like https://bugzilla.mozilla.org/show_bug.cgi?id=1853450, which we're not set up for at the moment, unfortunately :(. But maybe this could be the kick we need to work on it. (This signing happens with a self generated / self signed cert at the moment.)

taskcluster/config.yml
oskirby: Moving it into the kind configuration does sound cleaner. I'll give that a try and see what happens.

oskirby: Okay, that is a lot cleaner, thank you for the suggestion!

macos/pkg/CMakeLists.txt
```diff
@@ -2,47 +2,27 @@
...
-    endif()
-    set(XLIFFTOOL ${CMAKE_SOURCE_DIR}/scripts/utils/xlifftool.py ${CMAKE_SOURCE_DIR}/3rdparty/i18n/${LOCALE}/mozillavpn.xliff)
-    execute_process(
-        RESULT_VARIABLE XLIFF_CHECK_RESULT
-        COMMAND ${XLIFFTOOL} --locale=${LOCALE} --check
-    )
-    if(NOT XLIFF_CHECK_RESULT EQUAL 0)
-        continue()
-    endif()
+# Determine the root of the project checkout - we scripts and localization.
```
mcleinman: "we scripts"? Should that be "w/ scripts"?

mcleinman (approved):
r+ for non-Taskcluster areas (didn't look at taskcluster ones)

macos/pkg/CMakeLists.txt
oskirby: Oops, I meant to write "we need scripts and localization"

I will make a pass through this file and update comments.

PR #10610, opened 2025-06-26
WIP: Use conda for Windows unit test actions

oskirby:
## Description
We sure do seem to be having a lot of unit test failures on Windows and I don't really have a good explanation for why. Seems like the unit tests are failing to link with some mangled object paths. We haven't touched anything recently that would explain it.

Also, the unit tests are the only thing in our CI still using the MSVC tools and they are slow, so let's give Conda and Clang a try instead.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/.github/workflows/test_unit.yaml
+++ b/.github/workflows/test_unit.yaml
@@ -143,11 +143,13 @@ jobs:
 
       - name: Building tests
         run: |
+          conda info
+
           mkdir ./build-win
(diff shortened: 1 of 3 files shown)


oskirby:
Okay - conda kinda sucks for this, the resuling conda environment is *massive* (8GB) to the point where is kinda infeasible to generate in a CI environment and most of the stuff we are installing via conda is already present in the Github `windows-latest` runner image.

PR #10607, opened 2025-06-26
VPN-5276: Remove use of actool to generate macOS AppIcon

oskirby:
## Description
The `actool` command we use to generate an asset catalog for the AppIcons, it a macOS specific tool for which I can't find an open-source alternative. As another approach to setting the macOS app icons, we can use the `icnsutil` package which will generate the icns file for the icons. 

## Reference
Prerequisite for [VPN-5276](https://mozilla-hub.atlassian.net/browse/VPN-5276)
Split from: #10584

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-5276]: https://mozilla-hub.atlassian.net/browse/VPN-5276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10619, opened 2025-07-02
Fix icnsutil when building with Xcode

oskirby:
## Description
We recently added a new `icnsutil` tool to generate the macOS AppIcon, but unfortunately it doesn't quite work on Xcode because the tool doesn't exist in Xcode's paths. To make it work, we need to use a tool that was found in CMake using `find_program` which will ensure it is invoked by the full path.

## Reference
Introduced in PR #10607

## Checklist
    

mcleinman (approved):
👍🏻

PR #10606, opened 2025-06-26
VPN-5276: Split apple and android conda environments

oskirby:
## Description
As a prerequisite to supporting OSX cross compilation, we are going to need to add a bunch of tooling to the conda environment and rather than injecting a lot of cruft into the Android build environment, it would be best if we just split the Conda environments for Apple and Android targets.

For what it's worth, we have never really had a common environment anyways, since the `conda_trim.sh` script removes a lot of stuff from the Android environment.

## Reference
Prerequisite for [VPN-5276](https://mozilla-hub.atlassian.net/browse/VPN-5276)
Split from: #10584

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-5276]: https://mozilla-hub.atlassian.net/browse/VPN-5276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10620, opened 2025-07-02
Ignore codesign tool from conda

oskirby:
## Description
When we split the Apple and Android conda environments in PR #10606 we also merged the contents of `scripts/macos/conda_install_extras.sh` into `env-apple.yml` which meant that the `cctools` package is now installed by default when developing for Apple.

This brings in a copy of the `codesign` tool in the Conda environment that conflicts with the `codesign` binary provided by Apple and unfortunately the conda version is unsuitable for our needs (it only seems to support adhoc signing). Therefore, we should prefer the Apple version instead when searching for a suitable signing tool.

## Reference
Introduced by PR #10606

## Checklist
    
### DIFF
--- a/scripts/cmake/osxtools.cmake
+++ b/scripts/cmake/osxtools.cmake
@@ -13,7 +13,25 @@ else()
 endif()
 
 if(CODE_SIGN_IDENTITY)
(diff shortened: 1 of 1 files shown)


mcleinman (approved):
Thanks!

PR #10623, opened 2025-07-02
VPN-7135: update links for addons

mcleinman:
## Description

Needed new links so that we don't ever get the test variant that doesn't show the PPP pricing. While we're at it, we'll update the stage links.

## Reference

VPN-7135

## Checklist
    
### DIFF
--- a/addons/message_upgrade_to_bundle_1/openLink.js
+++ b/addons/message_upgrade_to_bundle_1/openLink.js
@@ -1,9 +1,9 @@
 ((api) => {
   if (api.env.inProduction) {
     return api.urlOpener.openUrl(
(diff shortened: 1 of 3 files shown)


oskirby (approved):
I'll give this one a tentative approval, though I just want to call attention to the fact that `geo=US` is hard-coded here.

I know the intention behind the product bundle experiment is that it's limited to US customers only, but it still feels kind of incorrect to hard-code it here. Instead, we should probably populate this with the real `geo` value from the `Location` model and then let guardian handle the `geo` tag accordingly.

But, if this is intentional, then I am fine with it.

mcleinman:
Going to merge this even with the 4 failing Android tests, as they are fixed in another PR and are unrelated to the work here: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10622/

PR #10624, opened 2025-07-02
VPN-7122: Fix view logs when macOS daemon not running

oskirby:
## Description
When we added the XPC service API, we rewrote the implementation of `MacOSController::getBackendLogs()` and we introduced a bug where we can fail to read the logs in a way that never invokes the error handler either. This results in a call to `getBackendLogs()` that never closes the logfile, and results in an incomplete log file.

This occurs when the daemon is not running, and has never successfully been initialized. In such a situation `m_connection` is a null pointer, which results in no XPC call ever being made. To fix this, we can create an ad-hoc XPC connection to the daemon for reading the logs. This should ensure that an attempt is always made to read the logs, regardless of the state of the daemon.

## Reference
JIRA issue: [VPN-7122](https://mozilla-hub.atlassian.net/browse/VPN-7122)
JIRA issue: [VPN-7106](https://mozilla-hub.atlassian.net/browse/VPN-7106)

## Checklist
    
[VPN-7122]: https://mozilla-hub.atlassian.net/browse/VPN-7122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-7106]: https://mozilla-hub.atlassian.net/browse/VPN-7106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/daemon/daemon.h
+++ b/src/daemon/daemon.h
@@ -37,10 +37,10 @@ class Daemon : public QObject {
 
   QJsonObject getStatus();
   QString logs();
-  void cleanLogs();
 
   Q_INVOKABLE bool activate(const QString& json);
   Q_INVOKABLE bool deactivate() { return deactivate(true); };
+  Q_INVOKABLE void cleanLogs();
 
  signals:
   void connected(const QString& pubkey);
--- a/src/platforms/macos/daemon/xpcdaemonserver.mm
+++ b/src/platforms/macos/daemon/xpcdaemonserver.mm
@@ -226,7 +226,7 @@ shouldAcceptNewConnection:(NSXPCConnection *) newConnection {
 }
 
 - (void)cleanupBackendLogs {
-  QMetaObject::invokeMethod(self.daemon, [&]{ self.daemon->cleanLogs(); });
+  QMetaObject::invokeMethod(self.daemon, "cleanLogs");
 }
 
 @end
--- a/src/platforms/macos/macoscontroller.mm
+++ b/src/platforms/macos/macoscontroller.mm
@@ -238,7 +238,13 @@ void MacOSController::checkStatus() {
 }
 
 void MacOSController::getBackendLogs(QIODevice* device) {
-  auto conn = static_cast<NSXPCConnection*>(m_connection);
+  NSXPCConnection* conn = [NSXPCConnection alloc];
+  [conn initWithMachServiceName:machServiceName()
+                        options:NSXPCConnectionPrivileged];
+  conn.remoteObjectInterface =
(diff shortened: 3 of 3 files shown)


mcleinman (changes requested):
It seems like daemon log cleanup is broken - the daemon logs start at the same spot every time. Let me know if you want me to take this back to finish up that part.

oskirby:
> It seems like daemon log cleanup is broken - the daemon logs start at the same spot every time. Let me know if you want me to take this back to finish up that part.

When the daemon isn't running that isn't something that's fixable and we are making a best-effort to fetch logs in this case. The problem is that the log file is owned by `root` so only the root user can modify. Even though the client can manually dump the file, it lacks permission to modify the logs.

The log cleanup should still work as expected when the XPC call is successful, in which case the daemon will perform the cleanup.

mcleinman:
> > It seems like daemon log cleanup is broken - the daemon logs start at the same spot every time. Let me know if you want me to take this back to finish up that part.
> 
> When the daemon isn't running that isn't something that's fixable and we are making a best-effort to fetch logs in this case. The problem is that the log file is owned by `root` so only the root user can modify. Even though the client can manually dump the file, it lacks permission to modify the logs.
> 
> The log cleanup should still work as expected when the XPC call is successful, in which case the daemon will perform the cleanup.

I was previously testing with a valid daemon. However I may not have wiped the build folder between runs, so it may have had an out of date daemon. :(

When testing this PR on my machine (having blown away the build folder - my only copy of the VPN - before) it works well when the VPN is not active. However, when the VPN is active, it does not clear out the backend logs...

However, now I'm getting even weirder behavior - when I tap "view logs" with an active VPN, the VPN is deactivated. (I also get a momentary "background service error" banner.) I'm also seeing this on `main` - and I wonder if that is related to why it's not clearing logs when the VPN is active.

mcleinman:
(Sorry, merged `main` into this PR - meant to do that on one of my PRs, and I copied the wrong branch name from GH.)

oskirby:
I think we have a fix, the change to make use of a transient XPC connection revealed a use-after free bug in the lambda-capture used to invoke the `Daemon::cleanLogs()` method.

mcleinman (approved):
Tests great. Thanks!

PR #10626, opened 2025-07-03
VPN-7135 touch up

mcleinman:
## Description

After the last VPN-7135 PR, we got confirmation that we could remove these params from the URL. (Had gone ahead with the last one while awaiting an answer because this is time-sensitive.)

## Reference

VPN-7135

## Checklist
    

oskirby (approved):
LGTM!

PR #10634, opened 2025-07-07
VPN-7057: start up on boot when signing back in

mcleinman:
## Description

See comment in code for explanation.

I don't loooove this fix, but it was the best way I could figure out to do this. There were some cleaner/neater ways, but they all ended up with a side effect of turning on the VPN at the end of the onboarding sequence, which we don't want. 

## Reference

VPN-7057

## Checklist
    
### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -475,6 +475,18 @@ void MozillaVPN::maybeStateMain() {
     settingsHolder->setAdjustActivatable(false);
   }
 #endif
+
+  // This next bit covers specific situation: When signing out and signing back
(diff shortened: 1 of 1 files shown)


mcleinman:
> > The location check was #10589. Without it, when "start on boot" is enabled we activate the VPN fairly quickly. This VPN activation causes us to delete all other running tasks, including TaskGetLocation, and thus we don't have the client location. In turn, this lack of location (among other things) prevents us from seeing if we should show the PPP addon.
> 
> I see thanks! Now i would really prefer if we move that into it's own bool :D

Done! I've updated the comments as well, and re-tested this code in the same ways described originally in the PR.

PR #10646, opened 2025-07-10
VPN-7141: improved LAN bypass on iOS

mcleinman:
## Description

- Lots of UI tweaks - feel free to ask for a new UI object, rather than cramming all that in the existing toggle row.
- Need to update the SUMO article link when it's finalized.
- While iOS bulds are currently broken (pending VPN-7153), I tested this w/ AirDrop on an older build with these changes applied.
- @flodolo we spent a while discussing these strings in English - we want them to make clear that there are some usability tradeoffs if you miss with this toggle. Not sure if there are better ways to get that across. (Also, Santiago wanted that character recommendation - please let us know if that is a bad practice.)
- `excludeCellularServices` defaults to `true`, and I've [left that default](https://developer.apple.com/documentation/networkextension/nevpnprotocol/excludecellularservices). This means some traffic to the cell provider may escape the VPN tunnel, but without this we likely will impact WiFi calling and a few other things - and should adjust the strings. This has always escaped the VPN tunnel, and it seemed unnecessary to change that here, though I don't feel strongly. A more "killswitch" mode should probably bring this into the tunnel, but I didn't think it was needed here. Open to discussion though.

VPN-7168 and VPN-7169 will finalize the work for this epic. This standalone work moves us in the right direction, and will improve the current situation for users dramatically.

## Reference

VPN-7141

## Checklist
    

src/translations/strings.yaml
```diff
@@ -372,10 +372,13 @@ settings:
...
 
-localNetworkAccess:
-  labelTitle: Exclude local network traffic
-  subLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers.
-  disabledSubLabel: Access Apple CarPlay, AirDrop, AirPlay and local network devices like printers. The VPN must be off to edit this setting.
+bypassVPN:
+  labelTitle: Bypass VPN for select traffic
+  subLabel:
+    value: Allow RCS/MMS messaging, Apple features like CarPlay and AirDrop, and connections to local devices. May reduce network security.
+    comment: This string should be clear that major features like some messages will break, but at potential the expense of some security.
```
flodolo, line 379:
```suggestion
    comment: Make sure it's clear that this will allow major features like messaging apps, but potentially at the expense of some security.
```

Isn't the comment backward, given it's now bypass vs exclude?

src/translations/strings.yaml
```diff
@@ -372,10 +372,13 @@ settings:
...
+  learnMoreSentence: Learn more.
```
flodolo, line 380: Double-checking, as other _learn more_ strings don't have a closing period.

src/translations/strings.yaml
mcleinman: As confusing as it may be, "bypass VPN" and "exclude local network traffic [from VPN]" do the same thing. We're changing how it works under the hood and changing the default, so we want to be clearer in verbage.

src/translations/strings.yaml
mcleinman: I'll confirm this - the only reason I didn't reuse an existing one was that this one had a period.

mcleinman: Looking at the context this is used in versus the others, I can confirm we want this to have the period.

oskirby (approved):
Looks good to me

src/translations/strings.yaml
flodolo: Too late, since this landed, but the second part is still a bit of broken English though?

mcleinman: Yes that is terrible English, sorry. (I wrote that.) If we update the comment, will it update it on pontoon but not require re-translations from all our volunteers? If so (and that's the preferred path), I'll happily put up a PR to touch that up.

flodolo: Updating the comment won't force re-translation, so it's fine to do.

mcleinman: Fix is coming: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10686

PR #10648, opened 2025-07-10
VPN-7153: Use install_name_tool to set rpath for iOS frameworks

oskirby:
## Description
In PR #10609 we tried to merge all the rustlang environment hacking into a common `config.toml` file, which mostly worked, but in that change it seems like we broke the setting of the library rpath for iOS Frameworks.

This attempts to fix the issue by setting the rpath after the build using the `install_name_tool` program, after which we can remove the linker arguments.

## Reference

JIRA Issue [VPN-7153](https://mozilla-hub.atlassian.net/browse/VPN-7153)
Bug introduced by PR https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10609

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-7153]: https://mozilla-hub.atlassian.net/browse/VPN-7153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10658, opened 2025-07-15
Set both rpath and id using install_name_tool

oskirby:
## Description
Seems like our attempt to fix the rpath loading issues for the `QtGleanBindings` framework failed and still results in an iOS build that fails to launch. Digging a little deeper into the `otool` output between version 2.29.0 and 2.30.0 it seems that we missed the setting of `LC_ID_DYLIB` which was `@rpath/QtGleanBindings.framework` in 2.29.0 and some ugly path into the repository on 2.30.0.

In fact, the 2.29.0 release, didn't set an `LC_RPATH` at all, so think the proper fix is to just set `LC_ID_DYLIB`. This chan be achieved using  the `-id <name>` argument to `install_name_tool`

## Reference
Last attempt: #10648
JIRA Issue: [VPN-7153](https://mozilla-hub.atlassian.net/browse/VPN-7153)

## Checklist
    

oskirby:
Oops - wrong branch. Closing this.

PR #10660, opened 2025-07-15
VPN-7160: Force deactivation on multi/single hop server switch

oskirby:
## Description
There is a bit of a bug in the handling of server switches between single and multihop. For the most part, when performing a server switch we mostly just invoke the platform controller's `m_impl->activate()` method with the new connections and expect the backend to figure it out, and to make this a little smoother, the `m_impl->deactivate()` methods contain a check for server switching and skip the deactiation in this case.

However! This fails when switching from single hop to multihop and vice-versa. The problem occurs because activation is handled per-hop, in which case there might be a different number of hops before and after the switch, leading to stale connections in the daemon.

There's a lot of tech debt in how we handle multihop, but this should be a minimal fix that tries not to disturb too much load-bearing code. We try to do the following:

- Remove the `reason` from the `m_impl->deactivate()` method. Deactivate, really means deactivate. No exceptions.
- Extend the `Controller::switchServers()` to check for switching between single and multihop cases. If such a case is found, we must deactivate and continue on the disconnected signal. Otherwise, we can just proceed directly to activation.
- We can also simplify `Contoller::silentSwitchServers()`. Since the server data doesn't change it can always proceed directly to activation.
- This approach also makes `m_nextServerSelectionPolicy` redundant, as it is always `RandomzeServerSelection`

I think this should fix the issue, and I have tested it out on Linux. But unfortunately this does touch the controller API so we need to make sure it doesn't break other platforms too.

Some cleanup that came along for the ride:
- There is a lot of cruft in the `Controller` class that just deals with managing the connection time. This is silly. The controller can just set a timestamp when the connection started and let QML handle the refreshing and redrawing.
- Get rid of the stateful `isSwitchingServers` - this is something that should be knowable from `m_state`.
- Put a bit of effort into documenting the connection states.
- Use a `NextStep` to handle the disconnect -> switch cases.
- Rename `m_connectingTimer` to `m_confirmingTimer` as it's only really used for the confirming state.

Some food for thought, it seems like the only use of `StateConfirming` now is for Android onboarding. This could be cleaned up if we made the onboarding process a first-class API in the controller...  we already kinda did that for macOS with the permission required stuff. Maybe it's time to bring some of that cleanup to mobile platforms too.

## Reference
JIRA Issue: [VPN-7160](https://mozilla-hub.atlassian.net/browse/VPN-7160)

## Checklist
    
## Tested on

- [X] Linux
- [x] macOS
- [x] Windows
- [ ] Android
- [ ] iOS

[VPN-7160]: https://mozilla-hub.atlassian.net/browse/VPN-7160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -692,7 +692,7 @@ void Controller::disconnected() {
       emit readyToQuit();
       setState(StateOff);
       return;
-    
+
     case Update:
       emit readyToUpdate();
       setState(StateOff);
@@ -701,9 +701,10 @@ void Controller::disconnected() {
     case Reconnect:
       m_serverData = m_nextServerData;
       emit currentServerChanged();
-      activateInternal(DoNotForceDNSPort, RandomizeServerSelection, m_initiator);
+      activateInternal(DoNotForceDNSPort, RandomizeServerSelection,
+                       m_initiator);
       return;
-    
+
     default:
       break;
   }


mcleinman (changes requested):
Thanks for making this bit of code far more rational.

I couldn't replicate that JS crash, and in general it handled most of the timer/telemetry tests I threw at it. That's great news!

However, when switching between single hop and multihop:
- We show disconnection for a moment (switch, background color, etc.), which is something that I believe we've worked on not doing unless the user switches the VPN to off.
- The timer resets (which like any other server switch, we don't want to do)
- Telemetry is produced (for both daemon_session and vpnsession, as tested on iOS)  as if this was a new session - end, flush, and start pings are produced when switching between single hop and multihop. We want none of these to be produced (though we do need to record the data transfer telemetry in the daemon).

Furthermore (as tested on desktop) - with this PR, we do not record data transfer telemetry when switching servers (and not changing between single hop and multihop). (Tested as "switching from a single hop location to a new single location" on macOS - don't know about mobile, or silent switch, or multihop-to-singlehop switches.) On `main`, if we turn on the VPN, load a webpage, switch locations, load another webpage, and turn off the VPN - we see 2 data transfer entries in the telemetry. However, in this PR we only see 1, so we're missing some data.

Given that I'm probably closest to the telemetry and timer work, let me know if you'd like to pair on this - I'd be happy to. And if I can figure out how, it might be worth my writing some automated tests for this.

mcleinman (changes requested):
Tested these scenarios on macOS:
    - single hop, no server switch
    - multihop, no server switch
    - single hop, server switch to another single hop location
    - multihop, server switch to another multihop location
    - single hop, server switch to a multihop location
    - multihop, server switch to single hop location
    - silent server switch from app

Everything worked great. 

I started the same test regimen on iOS, with the addition of "silent server switch from daemon". Unfortunately, switching from single hop to multihop (and potentially other scenarios) result in extra daemon telemetry - the 3 pings (with reasons end, flush, and start) are produced when the server is switched. 

If you'd like, I'm happy to pair on this.

oskirby:
Based on the symptoms, I think I have another tweak that should work around the unexpected telemetry on iOS, but it feels like a bit of a hack rather than getting to the root of the issue. As I understand it for iOS (and Android) the problem is that the daemon/networkextension doesn't really know why a disconnect occurs and so its only option is to generate telemetry on deactivation. To fix it properly we should probably bring the `reason` argument back so that the backend can act on it accordingly...  but this would also be a completely new code path in the backend for Android and iOS, so maybe the hacky solution is better anyways.

The workaround is that platforms which don't support native multihop (eg: `m_impl->multihopSupported()` returns `false`) don't need to perform a full deactivation when switching from single-hop to multihop since it's all just single-hop under the hood. In which case, the `if (singlehop->multihop) { deactivate(); }` path is never called for these platforms.

oskirby:
I am thoroughly shamed at my lack of ability to test on real iOS hardware, so I am gonna go spend some money and get my iOS workflow going again before asking you to test this on my behalf.

mcleinman (approved):
Looks great!

Tested these scenarios on macOS:
- single hop, no server switch
- multihop, no server switch
- single hop, server switch to another single hop location
- multihop, server switch to another multihop location
- single hop, server switch to a multihop location
- multihop, server switch to single hop location
- silent server switch from app

I started the same test regimen on iOS, with the addition of "silent server switch from daemon". 

Everything worked great. Thanks for all your work on this one.

PR #10662, opened 2025-07-17
VPN-7142: Directly install daemon for macOS<13

oskirby:
## Description
It seems that installing a wrapper script into `/Library/PrivilegedHelperTools`  is a bit brittle and can lead to some problems in the XPC communication channel. Despite the daemon launching successfully, XPC messages just seem to end with an "invalid message". Surprisingly, If we replace the script with the daemon itself (which is how `SMJobBless` does it), then it all seems to work fine, so I am guessing that there is some codesign or security check that is failing.

So, let's do it the `SMJobBless` way and make the daemon actually installable. The only reason we didn't attempt this in the first place is that can no longer use a relative path inside the bundle to find the `wireguard-go` binary, and furthermore, because the daemon and `wireguard-go` are no longer running in the same bundle, we can no longer trust the bundle code signatures to prevent tampering of the `wireguard-go` tool.

Most of this patch is just re-implementing the codesign checks within `WireguardUtilsMacos` to verify `wireguard-go`. The codesign requirements we try to setup are:
- `anchor apple generic`: The certificate chain is rooted in an Apple certificate.
- `certificate leaf[subject.OU] = <team identifier>`: The `wireguard-go` was signed by the same apple development team as the daemon.
- `certificate leaf[subject.CN] = <common name>`: The `wireguard-go` was signed by the same certificate name as the daemon.

The next logical step to this journey would be to do away with the `postinstall` script altogether and just use `SMJobBless` for macOS 11/12 users, but that's a job for a future PR.

## Reference
JIRA Issue: [VPN-7142](https://mozilla-hub.atlassian.net/browse/VPN-7142)

## Checklist
    
[VPN-7142]: https://mozilla-hub.atlassian.net/browse/VPN-7142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/macos/daemon/wireguardutilsmacos.cpp
+++ b/src/platforms/macos/daemon/wireguardutilsmacos.cpp
@@ -6,6 +6,9 @@
 
 #include <errno.h>
 #include <net/route.h>
+#include <Security/SecRequirement.h>
+#include <Security/SecStaticCode.h>
+#include <Security/SecTask.h>
 
 #include <QByteArray>
 #include <QDir>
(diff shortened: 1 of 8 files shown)


src/platforms/macos/daemon/wireguardutilsmacos.cpp
```diff
@@ -65,6 +71,144 @@ void WireguardUtilsMacos::tunnelFinished(int exitCode,
...
+    QDir appPath(QCoreApplication::applicationDirPath());
+    appPath.cdUp();
+    appPath.cdUp();
+    appPath.cd("Resources");
+    appPath.cd("utils");
+    return appPath.filePath("wireguard-go");
+  }
+
+  // For earlier versions of macOS - this must be a fixed path
+  return "/Applications/Mozilla VPN.app/Contents/Resources/utils/wireguard-go";
```
mcleinman, line 88: Do users get a choice as to whether they're installing for just themselves (and not all users)? If so, in that case the path would be `~/Applications/Mozilla VPN.app/Contents/Resources/utils/wireguard-go`, no? Do we need to handle both these situations?

src/platforms/macos/daemon/wireguardutilsmacos.cpp
```diff
@@ -65,6 +71,144 @@ void WireguardUtilsMacos::tunnelFinished(int exitCode,
...
+}
+
+// static
+QString WireguardUtilsMacos::wireguardGoRequirements() {
+  static QString requirements;
+  if (!requirements.isEmpty()) {
+    return requirements;
+  }
```
mcleinman, line 96: What is this block? Should it read `QString requirements = wireguardGoRequirements();` like the similar line below?

mcleinman (commented):
Couple questions

src/platforms/macos/daemon/wireguardutilsmacos.cpp
oskirby: We explicitly deny relocated installations by setting `BundleIsRelocatable` off when building the installer package in [macos/pkg/CMakeLists.txt](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/ee30c7096ebade444dc0170bc1b8043d0ea342d1/macos/pkg/CMakeLists.txt#L89).

The application can only be installed in the root location.

src/platforms/macos/daemon/wireguardutilsmacos.cpp
oskirby: This is inside `WireguardUtilsMacos::wireguardGoRequirements()` so this suggestion would become a recursive call and probably an infinite loop.

Note that because `requirements` is a static variable, it will retain its value. This check therefore makes it so that the body of the function is only executed once and the result is cached for subsequent uses.

mcleinman (approved):

PR #10665, opened 2025-07-18
Revert "VPN-7092: Move from Taskcluster to Xcode Cloud for iOS test builds (#10640)"

mcleinman:
## Description

This reverts commit 27874285aa2bd090c084922de757405778d1af3d.

Back to TaskCluster for iOS - looks like we moved it out prematurely

## Reference

N/A

## Checklist
    
### DIFF
--- a/taskcluster/kinds/build/ios.yml
+++ b/taskcluster/kinds/build/ios.yml
@@ -1,0 +1,34 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+task-defaults:
+    description: "IOS Debug Build"
+    treeherder:
+        symbol: BD
+        kind: build
+        tier: 1
+        platform: ios/arm64
+    worker-type: b-macos
+    fetches:
+        fetch:
+            - miniconda-osx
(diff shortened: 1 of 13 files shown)


taskcluster/kinds/build/ios.yml
```diff
@@ -0,0 +1,34 @@
...
+            - miniconda-osx
+    worker:
+        taskcluster-proxy: true
+        chain-of-trust: true
+        max-run-time: 3600
+    release-artifacts:
+        - MozillaVPN.tar.gz
+    run:
+        using: run-task
+        use-caches: []
```
ahal, line 24: I believe this will disable the checkout cache, is this intentional? If not it would be better to remove this key entirely so the checkout cache remains in place.

Not having the checkout cache means that the worker will do a fresh clone of the repo each time, even if running tasks back to back.

ahal (approved):
Looks good to me modulo my point about the checkout cache

taskcluster/kinds/build/ios.yml
mcleinman: I don't quite know the history of why this was developed the way it was. Since this is a revert of what was done a week ago, I'm somewhat inclined to put it back the way it was before - and then make further changes if we want from that restored version.

oskirby (approved):
LGTM

PR #10668, opened 2025-07-21
Bump Qt/next to version 6.9.1

oskirby:
## Description
Let's also take a look at upgrading the Qt/next version for desktop platforms in preparation for a switch to Qt 6.9.1.

## Reference
JIRA Issues: [VPN-6767](https://mozilla-hub.atlassian.net/browse/VPN-6767) - Windows
JIRA Issues: [VPN-6768](https://mozilla-hub.atlassian.net/browse/VPN-6768) - macOS
JIRA Issues: [VPN-6769](https://mozilla-hub.atlassian.net/browse/VPN-6769) - Linux

## Checklist
    

oskirby:
Looks like the culprit for the sentry error is [QTBUG-132735](https://bugreports.qt.io/browse/QTBUG-132735)

src/feature/featurelistcallback.h
```diff
@@ -37,7 +37,7 @@ struct FeatureSupportedPlatforms {
   bool wasm = false;
 };
 
-consteval auto byPlatform(FeatureSupportedPlatforms support) {
```
strseb, line 40: This seems like a downgrade :D Why that?

oskirby: As a workaround for [QTBUG-132735](https://bugreports.qt.io/browse/QTBUG-132735) we downgraded to C++17 when building `.deb` packages, and `consteval` is only available in C++20 and beyond.

CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ if(APPLE)
     if(IOS)
         set(CMAKE_OSX_DEPLOYMENT_TARGET 15.0)
     else()
-        set(CMAKE_OSX_DEPLOYMENT_TARGET 11.0)
+        set(CMAKE_OSX_DEPLOYMENT_TARGET 12.0)
```
mcleinman: This cuts off [100-200 daily users](https://mozilla.cloud.looker.com/x/ZORcdVYoJnawDbGS3mcvYN) from future updates - is it required for 6.9? If so, we should at least make this clear to the rest of the team.

taskcluster/kinds/build/macos.yml
```diff
@@ -55,7 +55,7 @@ macos/opt:
 
 
 macos/osxcross:
-    description: "Mac Build (osxcross)"
+    description: "macOS Build (osxcross)"
```
mcleinman, line 58: Thank you! 🙏🏻

mcleinman (approved):
One question on macOS versions, but looking good. Thanks.

CMakeLists.txt
oskirby: According to Qt's documentation - this is the new minimum supported OS for Qt 6.9: https://doc.qt.io/qt-6/macos.html

I bumped this version while trying to chase down some build fails. It's not entirely clear why Qt increased this. I could try reverting it to see what happens. Or at the very least it should probably be conditional on Qt version.

oskirby: I was able to chase down the version change to: https://github.com/qt/qtbase/commit/79363ed6c86f23a881aab5e4e2d6ea65e3ca79df and it seems like it was simply done becasue macOS 11 reached EOL status. So I think it should probably be fine if we drop this back down to macOS 11.

strseb (approved):

CMakeLists.txt
mcleinman: Thanks! I think it's also okay for us to consider dropping macOS 11; I just think it should be a team-wide inform/discuss.

PR #10671, opened 2025-07-23
Simplify the tunnel.so compilation on android

strseb:
## Description

This PR moves our go-code around. 

## Before:

RN: we have a jni-c api (android/daemon/tunnel/libwg-go/jni.c) that exposes the jvm<->c layer. it itself calls into the api-go (android/tunnel/src/go/api-android.go) 

This is compiled by a makefile, called by a cmake file. The makefile actually downloads go, which is the reason we limited the compilation for android to mac/linux and also why we're still on go1.14. 

When trying to upgrading the NDK, the C-layer failed in multiple ways to compile.
From a prototype of the fxvpn-android integration i had a better solution:

## After

We now have a single cmake - file which will compile only one cgo shared library. So kotlin can now directly call into go. 
For that i created a new .aar, just to contain the build stuff. 
Also go is now sourced from conda. 

Functionally nothing changed, the go code is the same, the VPNService functions just moved into a diffrent package.
The only thing diffrent is the removal of the c glue code layer.


android/tunnel/src/go/go.mod
```diff
@@ -0,0 +1,32 @@
...
+	golang.zx2c4.com/wireguard v0.0.0-20210222142647-219296a1e787
+)
+
+require (
+	dmitri.shuralyov.com/gpu/mtl v0.0.0-20221208032759-85de2813cf6b // indirect
+	github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802 // indirect
+	github.com/go-gl/glfw/v3.3/glfw v0.0.0-20200222043503-6f7a984d4dc4 // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
+	github.com/jezek/xgb v1.0.0 // indirect
+	github.com/yuin/goldmark v1.4.13 // indirect
```
mcleinman, line 18: Why does this now need to be explicitly listed in here?

strseb: That's a go thing, `go install` automatically lists subdependencies. I did not manually write this :D 
so dependabot will handle this too. Nothing we need to manually care about.

mcleinman: It seems like we're picking up way more subdependencies than we did before, from looking the partner deleted file. Looking at just one of them - https://dmitri.shuralyov.com/gpu/mtl - it seems like it provides Go access to Apple's graphics rendering API... which feels a little weird for a dependency on Android?

env-android.yml
```diff
@@ -4,16 +4,17 @@ channels:
...
+  - go=1.24.5
   - rust-std-armv7-linux-androideabi=1.75
   - rust-std-x86_64-linux-android=1.75
   - rust-std-i686-linux-android=1.75
   - rust-std-aarch64-linux-android=1.75
-  - cmake=3.26.3
+  - cmake=3.31.6
   - ninja=1.11.0
   - conda-forge::openjdk=17
+  - wget
```
oskirby, line 17: Why add `wget`? Generally I recommend `curl` if it's something we are scripting as it generally comes pre-installed.

oskirby (approved):

PR #10672, opened 2025-07-23
Fix GitHub tests after taskcluster artifact trim

oskirby:
## Description
In PR #10650 we changed a bunch of toolchain artifacts from `zip` compression to `xz` instead and this saved a tremendous amount of disk space and network traffic, but as a downside it also forced us to change the URLs to some taskcluster artifacts, most notably the Qt6 builds for macOS and Windows.

Now that that change has landed and the taskcluster index has updated, we need to update the consumers of those toolchains as well.

Windows turned out to be a bit of a dumpster fire. The Github `windows-latest` runner image doesn't support `tar.xz` compression and tar extraction is painfully slow. To mitigate the pain on Windows, we moved the Qt setup into a resuable action and add support for caching the Qt directory.

This also revealed that we were misusing the cache action in our functional test workflow, so we fixed that too.

## Reference
Introduced by: #10650

## Checklist
    
### DIFF
--- a/docs/Building/macos.md
+++ b/docs/Building/macos.md
@@ -56,7 +56,7 @@ Get a static build of Qt made built in our CI.
 
 https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozillavpn.cache.level-3.toolchains.v3.qt-macos-6.6.latest/artifacts/public/build/qt6_macos.tar.xz
 
-Unzip the folder and remember the location for the configure step.
+Extract the archive and remember the location for the configure step.
 
 # Build
 
--- a/docs/Building/macos.md
+++ b/docs/Building/macos.md
@@ -54,7 +54,7 @@ When you next want to start building the VPN, all you need to do is activate you
 
 Get a static build of Qt made built in our CI.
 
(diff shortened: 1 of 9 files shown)


docs/Building/macos.md
```diff
@@ -54,7 +54,7 @@ When you next want to start building the VPN, all you need to do is activate you
 
 Get a static build of Qt made built in our CI.
 
-https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozillavpn.cache.level-3.toolchains.v3.qt-macos-6.6.latest/artifacts/public%2Fbuild%2Fqt6_mac.zip
+https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/mozillavpn.v2.mozillavpn.cache.level-3.toolchains.v3.qt-macos-6.6.latest/artifacts/public/build/qt6_macos.tar.xz
 
 Unzip the folder and remember the location for the configure step.
```
lesleyjanenorton: Nit/Later thing: Would `*Extract the folder` be more accurate now ?

lesleyjanenorton (approved):
👍

docs/Building/macos.md
oskirby: That would be a little more accurate, yeah

PR #10673, opened 2025-07-24
VPN-7187: Add BuildInfo namespace and cleanup version strings

oskirby:
## Description
We have a bit of a quirk in the official Flatpak builds. In order to make builds reproducable they fiddle with the clock to force build timestamps and the like to produce a constant value. This means that the buildNumber for flatpaks always comes out as `2.201111111111` which really isn't that helpful for identifying what was built.

To fix this, we can scrape the `BUILD_ID` from the git commit logs, if present. And since we're messing with the generated version stuff, let's also clean things up by moving it into `mzutils` and de-duplicating some version-getters. Most notably:
- `BuildInfo::version`: Contains the value of `CMAKE_PROJECT_VERSION`
- `BuildInfo::timestamp`: Is set to the UNIX timestamp when the build ran (seconds since 1970).
- `BuildInfo::number`: Contains the value of `BUILD_ID`
- `BuildInfo::commit`: Contains the Git SHA from `git rev-parse HEAD` or an empty string if there is no git info.

Then in terms of deduplication:
- Replace `Constants::versionString()` with `QCoreApplication::applicationVersion()` as it does literally the same thing, and it's even settable for testing.
- Replace `Constants::buildNumber` with `BuildInfo::number`
- Remove the generated `version.h` file.

We also tweak the developer menu to include the git commit.

## Reference
JIRA Issue: [VPN-7187](https://mozilla-hub.atlassian.net/browse/VPN-7187)

## Checklist
    
[VPN-7187]: https://mozilla-hub.atlassian.net/browse/VPN-7187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcleinman (approved):
I have some philosophical questions on doing a 200 line refactor for a small bug fix, including adding additional cmake build files (even if the steps within them would have been needed in *some* cmake file). 

My preference would be towards a more targeted fix for the actual bug here. IMO this patch neither improves nor worsens our codebase's complexity, and thus I'd default to not changing our codebase unnecessarily.

All that said, this code is already written, and as I said, I don't think it makes our codebase worse. Thus, I'm approving.

oskirby:
Let me take another attempt that this that just does the bare minimum to address the fix. Sorry for going overboard on this one.

oskirby:
I'm going to yeet this PR into the sun until.

PR #10691, opened 2025-07-30
VPN-7217: Fix Auth unit test hang on TOTP error

oskirby:
## Description
The Auth tests for Windows have been failing for a while, but we don't really notice it because it only runs on push (not on PR). Let's fix them.

This more or less just applies all the recent unit test fixes into the `auth_tests.yaml` workflow.

## Reference
JIRA issue: [VPN-7217](https://mozilla-hub.atlassian.net/browse/VPN-7217)

## Checklist
    
[VPN-7217]: https://mozilla-hub.atlassian.net/browse/VPN-7217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/qt-setup-taskcluster/action.yml
+++ b/.github/actions/qt-setup-taskcluster/action.yml
@@ -66,11 +66,16 @@ runs:
         tar -xf ${FILENAME}
 
     - name: Setup cmake paths
-      if: ${{ inputs.cmake-env == 'true' }}
-      working-directory: ${{ inputs.dest }}
+      if: ${{ inputs.cmake-env == 'true' && runner.os != 'Windows' }}
+      working-directory: ${{ inputs.dest }}/${{ steps.resolve.outputs.toolchain-path }}
       shell: bash
-      run: |
-        echo CMAKE_PREFIX_PATH=$(cd ${{ steps.resolve.outputs.toolchain-path }}/lib/cmake && pwd) >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


oskirby:
Wait a second - none of the auth tests have been running on any of the platforms. Arghhh!

oskirby:
Well, I'll be damned. Looks like we have a genuine bug in the in-app-auth code that the tests caught. I suspect there was an upstream change to how FxA responds to invalid TOTP codes and they changed the error code from `107` (invalid parameter) with a validation error on the code to `152` (invalid token confirmation code). This meant that we ignored the error handling and the auth hangs as a result.

It also looks like the first two tests (successful signup both with and without TOTP), would collide with one another. By using the same e-mail account, the second test would always fail because the account already exists. We fix that by giving it a different e-mail account so that the two tests can run independently.

oskirby:
If I were to make a spicy comment here. I am somewhat horrified by the error handling code here. It feels incredibly brittle and prone to bugs as FxA behavior subtly changes over time. I know that our long-term plans are to abandon this code and switch to web auth flows so I am not really motivated to dive in and rewrite anything. But it gives me bad vibes.

mcleinman (approved):
Great catch, thank you. I agree fully with your analysis about fragility. 😬 

And don't forget to file a Jira ticket (if you want).

PR #10695, opened 2025-07-31
VPN-7198: Install sentry-cli from pip

oskirby:
## Description
We have some sentry breakage as of late. Both the Windows and Android builds are failing to upload symbols because they don't have `sentry-cli` installed. Well, it also turns out that [sentry-cli](https://pypi.org/project/sentry-cli/) is now available as a pip package, so let's just use that since it ought to work as an installation method for all platforms.

And as a bonus, we can get proper version pinning now that it's in a `requirements.txt` file.

## Reference
JIRA Issue: [VPN-7198](https://mozilla-hub.atlassian.net/browse/VPN-7198)

## Checklist
    
[VPN-7197]: https://mozilla-hub.atlassian.net/browse/VPN-7197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPN-7198]: https://mozilla-hub.atlassian.net/browse/VPN-7198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/taskcluster/scripts/build/macos_osxcross.sh
+++ b/taskcluster/scripts/build/macos_osxcross.sh
@@ -13,6 +13,12 @@ print Y "Installing conda"
 source ${MOZ_FETCHES_DIR}/conda/bin/activate
 conda-unpack
 
+# Provide symlinks for some cctools to their standard name.
+# This is needed to make dsymutil happy as it will attempt to use 'lipo' to
+# extract debug symbols from universal Mach-O binaries.
+CCTOOLS_ARCH_TRIPLE=arm64-apple-darwin20.0.0
+ln -s ${CCTOOLS_ARCH_TRIPLE}-lipo ${CONDA_PREFIX}/bin/lipo
+
 # Use vendored crates - if available.
 if [ -d ${MOZ_FETCHES_DIR}/cargo-vendor ]; then
 mkdir -p .cargo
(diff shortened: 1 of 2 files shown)


taskcluster/scripts/build/macos_osxcross.sh
```diff
@@ -13,6 +13,12 @@ print Y "Installing conda"
 source ${MOZ_FETCHES_DIR}/conda/bin/activate
 conda-unpack
 
+# Provide symlinks for some cctools to their standard name.
+# This is needed to make dsymutil happy as it will attempt to use 'lipo' to
+# extract debug symbols from universal Mach-O binaries.
+CCTOOLS_ARCH_TRIPLE=arm64-apple-darwin20.0.0
```
mcleinman, line 19: Does the work in this file relate to sentry?

mcleinman: Same with macos.yml

mcleinman (commented):
One question on this, otherwise looks good. Thanks

taskcluster/scripts/build/macos_osxcross.sh
oskirby: This is adding generation of sentry symbols for the cross-compiled macOS builds too.

It also tests out that `sentry-cli` works as expected in the `conda-base` docker image, otherwise there is no way to test it in a PR as the Android sentry code only runs on `main`

PR #10709, opened 2025-08-05
VPN-7191: force daemon codesign identifier

oskirby:
## Description
WIP

## Reference

    i.e Jira or Github issue URL

## Checklist
    

oskirby:
After having a bit of a think about this, I am going to approach this differently. Had we found this bug before shipping 2.30 we could have avoided it by forcing the signing identifier, but now that its out in the wild it's too little too late and the only viable remedy is to change the launchd daemon name.

We will just need to be more aware in the future that the designated requirements mean we cannot change signing identifier of the daemon once it has been released to users.

PR #10717, opened 2025-08-08
Experiment: Support HTTP connect and MASQUE/UDP proxies

oskirby:
## Description
This does a bit of refactoring to add support for multiple proxy protocols in the `socksproxy` tool. To get there we refactor a bunch of generic proxy handling code into a parent `ProxyConnection` class which now has three concrete subclasses:
- `Socks5Connection`: Implements version 5 of the SOCKS protocol from [RFC1928](https://datatracker.ietf.org/doc/html/rfc1928)
- `HttpConnection`: Implements the HTTP `CONNECT` method from [RFC2616](https://datatracker.ietf.org/doc/html/rfc2616#section-9.9)
- `MasqueConnection`: Implements HTTP datagram and `connect-udp` support from MASQUE/[RFC9298](https://datatracker.ietf.org/doc/html/rfc9298)

The proxy protocol being used is snooped by peeking at the start of the header data using the `isProxyType()` static method. This allows the `Socks5` class to create the correct subclass to handle the connection.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/extension/socks5proxy/src/masqueconnection.h
+++ b/extension/socks5proxy/src/masqueconnection.h
@@ -4,6 +4,7 @@
 
 #ifndef MASQUECONNECTION_H
 #define MASQUECONNECTION_H
+#ifdef PROXY_MASQUE_ENABLED
 
 #include "httpconnectionbase.h"
 
@@ -40,4 +41,5 @@ class MasqueConnection final : public HttpConnectionBase {
   bool m_capsuleDrop = false;
 };
 
+#endif  // PROXY_MASQUE_ENABLED
 #endif  // MASQUECONNECTION_H
--- a/extension/socks5proxy/src/masqueconnection.h
+++ b/extension/socks5proxy/src/masqueconnection.h
@@ -5,15 +5,15 @@
 #ifndef MASQUECONNECTION_H
 #define MASQUECONNECTION_H
 
-#include "httpconnection.h"
+#include "httpconnectionbase.h"
 
 class QIODevice;
 
-class MasqueConnection final : public HttpConnection {
+class MasqueConnection final : public HttpConnectionBase {
   Q_OBJECT
 
  public:
-  explicit MasqueConnection(QIODevice* socket) : HttpConnection(socket){};
+  explicit MasqueConnection(QIODevice* socket) : HttpConnectionBase(socket){};
   ~MasqueConnection() = default;
 
   // Peek at the socket and determine if this is a MASQUE connection.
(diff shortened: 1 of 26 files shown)


extension/socks5proxy/src/socks5.cpp
```diff
@@ -37,20 +40,68 @@ void Socks5::newConnection(T* server) {
...
+}
 
-    ++m_clientCount;
-    emit incomingConnection(con);
-    emit connectionsChanged();
+void Socks5::tryCreateProxy(QIODevice* socket) {
+  // Inspect the socket to see what kind of protocol its using.
+  HttpRequest req = HttpRequest::peek(socket);
+  ProxyConnection* con = nullptr;
+  if (HttpConnection::isProxyType(req)) {
```
strseb, line 60: Can we make this a compile flag and have it for debug for now? 

This is a very fun experiment, but i don't see the extension using this for now given ff support is not there yet.

extension/socks5proxy/src/masqueconnection.h
```diff
@@ -0,0 +1,43 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef MASQUECONNECTION_H
+#define MASQUECONNECTION_H
+
+#include "httpconnectionbase.h"
+
+class QIODevice;
+
+class MasqueConnection final : public HttpConnectionBase {
```
strseb, line 14: Maybe something pointing out this only implements connect-udp? Given we probably need classes for bind-udp and connect-tcp too at some point.

strseb (approved):
🔥🔥🔥🔥 This is amazing!

extension/socks5proxy/src/socks5.cpp
oskirby: I would probably leave the `HttpConnection` support enabled since the HTTP proxies are pretty standard (and probably even better supported than SOCKS, IMO).

But yeah, we can turn off the MASQUE stuff for release builds for the time being.

extension/socks5proxy/src/masqueconnection.h
oskirby: I would think that `bind-udp` and `connect-tcp` should be supportable within the same `MasqueConnection` class if/when we feel like implementing them.

`bind-udp` probably isn't as useful in this tool though, as it's almost certainly going to be stuck behind NAT, making the bind relatively useless.

PR #10724, opened 2025-08-11
VPN-6767: Add a Qt next build for Windows

oskirby:
## Description
This adds a new `build-windows/next` job to taskcluster that builds the Mozilla VPN client for Windows using Qt 6.9.2. This turned out to be a bit of a mess to figure out because we needed to upgrade the compilation environment to Visual Studio 2022 to get it working.

## Reference
JIRA issue: [VPN-6767](https://mozilla-hub.atlassian.net/browse/VPN-6767)
Qt issue: [QTBUG-135452](https://bugreports.qt.io/browse/QTBUG-135452)

## Checklist
    
[VPN-6767]: https://mozilla-hub.atlassian.net/browse/VPN-6767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
Arghhhhhh! I think there's another bug here. Seems like the latest and greatest CMake from Visual Studio is 3.31, but this causes `cmake_autogen` to silently turn into a no-op and the build falls apart from there. I suspect this is something to do with [CMAKE_AUTOGEN_BETTER_GRAPH_MULTI_CONFIG](https://cmake.org/cmake/help/latest/variable/CMAKE_AUTOGEN_BETTER_GRAPH_MULTI_CONFIG.html) that was added in CMake 3.29

oskirby:
A new discovery! Seems like there was a binary compatibility change in VS2022 version 17.10 which we are running into when using a Visual studio bundle zipped up like this. I suspect this is the reason why Qt is using VS2022 version 17.9 in their CI (which annoyingly is *not* an LTSC release).

We are running afoul of binary compatibility rules because we are building with VS 17.10, but then trying to run the Qt tools on a system that has an older VC runtime. Ideally we would fix this by upgrading the vcredist package, but I don't know if we have the permissions to do that.

See:
- https://developercommunity.visualstudio.com/t/Access-violation-in-_Thrd_yield-after-up/10664660
- https://github.com/actions/runner-images/issues/10004

mcleinman (approved):

PR #10738, opened 2025-08-19
VPN-6766 - Android Qt 6.9

mcleinman:
While this builds on my machine™️, I haven't done the work to make TaskCluster happy. More worrisome, there is a runtime crash on launch, occuring as soon as something C++ tries to grab the Android context. I'm going to put this down for a bit and come back to it, but for posterity:

Yet to do
- Get it working without runtime errors
- Set this up as qt-next on taskcluster, keep prior and subsequent Android build around

Thoughts
- Can I get rid of the qtbindings module and see what happens, since the actual Qt stuff should be coming in via other means?
- Drastic, but I could just reduce the number of Android modules. Would that help?

A partial list of things I've tried, but didn't fix the run time crash:
        - [ ] Tried a new class like AppContextHolder (different one for both daemon and main app), and getting that initialized first, before anything else. That didn’t help.
        - [ ] Tried expiclity adding Qt::Core to src/CmakeLists.txt with the other ones - didn’t help
        - [ ] I checked the combined manifest file. It was being combined just fine.
        - [ ] In VPNActivity.java, confirmed the onCreate activity was the same one being used in onResume via System.identityHashCode(this), and so we weren’t accidentally creating a new one each time or something
        - [ ] Changed android:launchmode in Manifest file from “singleTop” to “singleTask”
        - [ ] Added in PRE_LINK into the ndk_openssl_merged command in android_openssl.cmake - didn’t help
        - [ ] Commenting out `QNetworkInformation::loadDefaultBackend();` gets it a little bit further than otherwise, but only because it skips what the first crash would be.
        - [ ] Commented out `removeUnusedEntriesAfterDays = 30` (from android/settings.gradle). Don’t remember why - maybe there was a warning?
        - [ ] Removing `RESOURCE_PREFIX /qt/qml` from src/ui/CMakeLists.txt, as there is a comment about removing that once we only compile with it
        - [ ] Android/common/build.gradle has 2 entries under `dependencies` of `implementation project(path: ':qtBindings’)`. Removing one of them doesn’t change anything.
        - [ ] Added `implementation fileTree(dir: '../libs', include: ['*.jar', '*.aar’])` under `dependencies` in `android/common/build.gradle` (Why? Probably something I saw online. Don’t remember.)
        - [ ] Including `repositories {  flatDir { dirs($rootDir/libs) } }` in `android/common/build.gradle` and `android/vpnClient/build.gradle`
        - [ ] Using a namespace of 'org.mozilla.firefox.vpn' for vpnClient
        - [ ] Upping the compileSdkVersion and targetSdkVersion to 35 (and making a few other small adjustments so things compile), and seeing if that fixes everything (it doesn’t)
    - [ ] I believe we shouldn’t be using QtApplicationBase/QtActivityBase - should use the derived class (in my PR)
    - [ ] ClientCommon was never used and was throwing warnings, so I removed it

------

## Description

    Describe your changes

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -101,7 +101,6 @@
 #include <QJsonDocument>
 #include <QLocale>
 #include <QQmlApplicationEngine>
-#include <QQmlContext>
 #include <QScreen>
 #include <QTimer>
 #include <QUrl>
@@ -2307,6 +2306,7 @@ int MozillaVPN::runCommandLineApp(std::function<int()>&& a_callback) {
 
   logger.info() << "MozillaVPN" << QCoreApplication::applicationVersion();
   logger.info() << "User-Agent:" << NetworkManager::userAgent();
+  Localizer localizer;
   return callback();
 }
 
@@ -2325,16 +2325,9 @@ int MozillaVPN::runGuiApp(std::function<int()>&& a_callback) {
 #endif
 
   QApplication app(CommandLineParser::argc(), CommandLineParser::argv());
-  // This object _must_ live longer than MozillaVPN to avoid shutdown crashes.
-  QQmlApplicationEngine* engine = new QQmlApplicationEngine();
-  MozillaVPN vpn;
-  QmlEngineHolder engineHolder(engine);
 
-  // TODO pending #3398
-  QQmlContext* ctx = engine->rootContext();
-  ctx->setContextProperty("QT_QUICK_BACKEND", qgetenv("QT_QUICK_BACKEND"));
-
-  if (SettingsHolder::instance()->stagingServer()) {
+  SettingsHolder settingsHolder;
+  if (settingsHolder.stagingServer()) {
     Constants::setStaging();
     LogHandler::instance()->setStderr(true);
   }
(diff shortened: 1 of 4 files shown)


android/adjust/AndroidManifest.xml
oskirby, line 1: Random non-blocking thought: At what point can we just remove Adjust entirely rather than putting effort into maintaining this dead code?

src/mozillavpn.cpp
```diff
@@ -306,12 +306,17 @@ void MozillaVPN::initialize() {
 #endif
 
   m_private->m_captivePortalDetection.initialize();
+  logger.debug() << "Captive Portal Detection initialized";
```
oskirby, line 309: This is a lot of debug to leave behind in this file. I'd clean it up before merging.

oskirby (approved):

android/adjust/AndroidManifest.xml
mcleinman: I thought about that when working on this, but wanted to do minimal changes in the Qt upgrade PR. I know, I know, for some hilarious definition of "minimal changes" - but I didn't want unrelated stuff in there. I'd be happy with a quick follow up pulling this out completely, after we get a clean Android regression test.

PR #10745, opened 2025-08-26
VPN-7011 Add the concept of SystemApps to split tunnel model

strseb:
## Description

This adds a new Checkbox to the bottom of the App List. 
If the AppList model contains at least one app with the label "system app" it is shown. 
System Apps are only shown in the list if that checkbox is true. 

## Demo
https://github.com/user-attachments/assets/85eaa2ea-6d67-464c-9631-dcdac5e1fb91

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -153,7 +153,7 @@ splittunnel:
     comment: Displays inside the information card at the top of the Excluded apps screen when the VPN is on
   btnShowSystemApp: 
     value: Show system apps
-    comment: Label for a checkbox that if checked will show the hidden "system/internal" apps. i.e on Android "Android Auto" is a system app.   
+    comment: Label for a checkbox that, if checked, will show the hidden "system/internal" apps. e.g. "Android Auto" is a system app on Android.
 multiHopFeature:
   singleHopToggleCTA: Single-hop
   multiHopToggleCTA: Multi-hop


src/translations/strings.yaml
```diff
@@ -151,7 +151,9 @@ splittunnel:
   infoCardDescription2:
     value: VPN must be off to edit Excluded apps
     comment: Displays inside the information card at the top of the Excluded apps screen when the VPN is on
-
+  btnShowSystemApp: 
+    value: Show System apps
```
flodolo: Is there a reason for System being uppercase?

flodolo (approved):

src/applistprovider.h
```diff
@@ -18,25 +18,36 @@ class AppListProvider : public QObject {
 
   virtual ~AppListProvider() = default;
 
+  // A Stable Identifier for an Application - should not change between
+  // invokations.
+  typedef QString AppId;
+  struct AppListEntry {
```
oskirby: As a minor suggestion, we could reuse the `AppPermission::AppDescription` class rather than defining a new type here.

src/translations/strings.yaml
strseb: Nope! Adjusted that :)

src/applistprovider.h
strseb: That is a great suggestion, now all applistproviders are just emitting the List<AppDescription>directly :)

src/translations/strings.yaml
```diff
@@ -151,7 +151,9 @@ splittunnel:
   infoCardDescription2:
     value: VPN must be off to edit Excluded apps
     comment: Displays inside the information card at the top of the Excluded apps screen when the VPN is on
-
+  btnShowSystemApp: 
+    value: Show system apps
+    comment: Label for a checkbox that if checked will show the hidden "system/internal" apps. i.e on Android "Android Auto" is a system app.
```
```suggestion
    comment: Label for a checkbox that, if checked, will show the hidden "system/internal" apps. e.g. "Android Auto" is a system app on Android.
```

Issue #10728, opened 2025-08-13
Unable to stay logged in


Inky19:
**Describe the bug**
I installed Mozilla VPN on a Kubuntu computer for a friend, but for the past few days, it has been automatically disconnecting every time the computer is restarted.
The client also resets if I use the “Quit Mozilla VPN” option in the tray bar and wait a moment. However, if I restart the GUI just a few seconds after quitting and the tray icon has disappeared, it reconnects almost every time.

Mozilla VPN is installed according to the method described on the [Mozilla website](https://support.mozilla.org/en-US/kb/how-install-mozilla-vpn-linux-computer). I also tried the new Flatpak version and the same still problem occurs.
kwallet6 and keepassxc are installed so it seems to be a different problem than #9624 as suggested by [this comment](https://github.com/mozilla-mobile/mozilla-vpn-client/issues/9624#issuecomment-2941725788). I also tried installing gnome-keyring, but that didn't change anything.

I suspect an error when storing the token. I tried logging in from the command line with “mozillavpn login” and the browser window opens and validates the request, but the command in the terminal hangs with no output and never finishes.
“mozillavpn status” always displays “User status: not authenticated”, even when the GUI is running and connected.

I tried deleting everything related to Mozilla VPN in the home folder (`~/.local/share/Mozilla/Mozilla VPN/`, `~/.config/mozilla/`, and `~/.cache/Mozilla/Mozilla VPN`), but that didn't help.

I have no problem staying logged in with this account on my other devices (Android and Arch Linux).

**VPN version:**
2.30.0 (2.30.0~rc20250718211135-static1)

**Affected Platform details:**
OS : Kubuntu 25.04  
DE : KDE Plasma 6.3.4  
QT : 6.8.3  
Kernel : 6.14.0-27-generic (64 bit)
Graphics platform : Wayland

Inky19:
@ValentinaPC I've already tried using the flatpak version of Flathub (as mentioned briefly in my first post). However, the same behavior seems to occur and the VPN keeps disconnecting after a reboot, just like the deb version.

oskirby:
Would you be able to provide us with a log file to help see what's happening when the application tries to run? You can do this by clicking `View Logs` in the `Get Help` menu.

[...]

Issue #10720, opened 2025-08-10
Split-tunneling does not work in Linux Mint


mwadle:
**Describe the bug**
Split tunneling does not work on Linux Mint. Apps selected as 'Excluded apps' still have network traffic routed through the VPN.

**VPN version:**
2.30.0 (2.30.0-noble1)

**Affected Platforms:**
Linux Mint 22.1 Cinnamon

**Prerequisites:**

**Steps to reproduce:**
Turn VPN on with an excluded app selected
Begin capturing traffic through your primary network interface
Generate internet traffic from an excluded app
Note that all internet traffic is either local or sent through the VPN

Alternatively - Exclude your web browser from the VPN connection
Using the excluded web browser, navigate to https://whatismyipaddress.com/
Note that the displayed IP is the VPN's IP address, not yours

**Expected result:**
Internet traffic from excluded apps does not get sent to the VPN's IP address.

**Actual result:**
All non-local traffic is sent through the VPN.

**Additional notes:**

[mozillavpn-2025-8-10.log](https://github.com/user-attachments/files/21707797/mozillavpn-2025-8-10.log)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-7208)

oskirby:
Unfortunately, we cannot support split tunneling with the Cinnamon desktop environment at present. The underlying cause is that Cinnamon doesn't perform group processes by control groups when launching apps, which is necessary for traffic classification by application.

Support for this feature is blocked by linuxmint/cinnamon#12015

We expected this feature to be disabled when installed on Linux mint, which suggests that we have a logic error in [FeatureCallback_splitTunnel()](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/78a1b3fbfeac3b60eeae942ad1d5bfe13e5b5086/src/feature/featurelistcallback.h#L97) that allowed it to be visible.

mwadle:
Thanks for the thorough response. You saved me from some serious troubleshooting!

I plan to take a look and see why the feature was available to me on Mint, but it'll probably be a week or so before I get the opportunity

mwadle:
My XDG_CURRENT_DESKTOP string is 'X-Cinnamon' which is interesting because I am running Cinnamon, not xfce. 

[...]

PR #10746, opened 2025-08-26
VPN-7212: Support KDE cgroup parsing

oskirby:
## Description
It's been a while since we took a look at how application scoping is handled on KDE, and there seem to be a few bugs in there. It seems that modern versions of KDE are now launching applications as systemd `service` units rather than `scope` or `slice` as we typically see on Gnome. Failing to handle this causes the VPN client to unnecessarily create a new systemd unit at startup.

This cgroup parsing change also affects how split tunneling is handled in KDE, and rather than implementing this twice, we expose a new `XdgPortal::parseCgroupAppId()` method that the split-tunneling `AppTracker` class can make use of when tracking application launches.

It also seems that our KDE version detection broke during the transition to KDE 6. So let's update that one too.

## Reference
Systemd: [desktop environments](https://systemd.io/DESKTOP_ENVIRONMENTS/)

Found while investigating:
- [VPN-7212](https://mozilla-hub.atlassian.net/browse/VPN-7212) / #10728
- [VPN-7208](https://mozilla-hub.atlassian.net/browse/VPN-7208) / #10720

## Checklist
    
[VPN-7212]: https://mozilla-hub.atlassian.net/browse/VPN-7212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VPN-7208]: https://mozilla-hub.atlassian.net/browse/VPN-7208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/linuxutils.cpp
+++ b/src/platforms/linux/linuxutils.cpp
@@ -73,11 +73,13 @@ QVersionNumber LinuxUtils::gnomeShellVersion() {
   QDBusInterface iface("org.gnome.Shell", "/org/gnome/Shell",
                        "org.gnome.Shell");
   if (!iface.isValid()) {
+    logger.debug() << "Unable to read Gnome version";
     return QVersionNumber();
   }
 
   QVariant shellVersion = iface.property("ShellVersion");
   if (!shellVersion.isValid()) {
+    logger.debug() << "Invalid Gnome version string";
     return QVersionNumber();
   }
   return QVersionNumber::fromString(shellVersion.toString());
@@ -88,11 +90,13 @@ QVersionNumber LinuxUtils::kdePlasmaVersion() {
   QDBusInterface iface("org.kde.plasmashell", "/MainApplication",
                        "org.qtproject.Qt.QCoreApplication");
   if (!iface.isValid()) {
+    logger.debug() << "Unable to read KDE version";
     return QVersionNumber();
   }
 
   QVariant appVersion = iface.property("applicationVersion");
   if (!appVersion.isValid()) {
+    logger.debug() << "Invalid KDE version string";
     return QVersionNumber();
   }
   return QVersionNumber::fromString(appVersion.toString());
--- a/src/platforms/linux/linuxutils.cpp
+++ b/src/platforms/linux/linuxutils.cpp
@@ -84,22 +84,18 @@ QVersionNumber LinuxUtils::gnomeShellVersion() {
 }
 
 // static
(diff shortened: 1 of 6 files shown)


src/platforms/linux/xdgportal.cpp
```diff
@@ -210,16 +203,24 @@ QString XdgPortal::parseCgroupAppId(const QString& cgroup) {
...
     if (cgSplit.last() == "autostart") {
       cgSplit.removeLast();
     }
     appId = cgSplit.last().section('@', 0, 0);
   } else if ((suffix == "scope") || (suffix == "slice")) {
+    // Systemd scopes and slices can take the forms:
+    //   app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
+    //   app[-<launcher>]-<ApplicationID>-<RANDOM>.slice
+    if (cgSplit.length() < 3) {
+      return QString();
```
mcleinman, line 223: Should we log something in all these cases that we're returning a blank string? In the future, it would allow us to better diagnose what is happening from user logs.

src/platforms/linux/linuxutils.cpp
```diff
@@ -84,22 +84,18 @@ QVersionNumber LinuxUtils::gnomeShellVersion() {
...
 
-  QByteArray result = proc.readAllStandardOutput();
-  for (const QByteArray& line : result.split('\n')) {
-    if (line.startsWith("KDE Frameworks: ")) {
-      auto vstr = QString::fromUtf8(line.last(line.size() - 16));
-      return QVersionNumber::fromString(vstr);
-    }
+  QVariant appVersion = iface.property("applicationVersion");
+  if (!appVersion.isValid()) {
+    return QVersionNumber();
```
mcleinman, line 100: Same - should we log something here?

mcleinman (approved):
Looks good, thanks. 

May want to add a bit more logging for future debugging of unexpected situations, per comments.

src/platforms/linux/xdgportal.cpp
oskirby: I don't think every empty string in here needs to be logged, as this method can be called rather frequently (eg: every cgroup creation will trigger it, whether it was created for an application or not). But for scopes, slices and services starting with `app-` it would be safe to log anything that appears malformed.

PR #10755, opened 2025-09-04
update docs as 'sign.sh' no longer exists

mcleinman:
In trying to sign a locally produced Android build, I learned that we [no longer can use the script](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10269). (We won't be able to either - two different deprecations would make this tough to bring back.) However, the script was referenced in a couple places in the documentation.

I updated both spots, and then realized the `scripts/README.md` doc was very out of date. I updated some of it, and added TODOs in spots that we still need to work on.


docs/Building/android.md
```diff
@@ -79,9 +79,7 @@ We have a gradle task setup for that.
 
 If you need to work with subscriptions or other play store functionality in the emulator, you will need to sign the apk.
 
-Make sure have an environment variable `AUTOGRAPH_TOKEN` set (ask a fellow dev).
+There formerly was a `sign.sh` script that helped with this. The server it relied no longer exists.
```
strseb, line 82: Nit: i don't like historic things :D - given we have git history, we can go back in time.

docs/Building/android.md
```diff
@@ -79,9 +79,7 @@ We have a gradle task setup for that.
...
+TaskCluster builds should be signed, and so creating a draft PR should create a signed build that could be used.
```
strseb: Creating a draft pr does NOT create a real signed build. 
Otherwise any of us could just create a WIP PR with evilthings and publish that as an official mozilla vpn apk. 

Draft pr's are signed same as local builds using this key but that is just convinence
(as otherwise you'd need to uninstall the app when switchting from tc-pr builds to local builds) : 
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/android/debug.keystore

docs/Building/android.md
```diff
@@ -79,9 +79,7 @@ We have a gradle task setup for that.
...
 
-Then run
-```
-    ./scripts/android/sign.sh .tmp/src/android-build/build/outputs/apk/release
-```
+If locally created builds must be signed, talk to the Autograph team about setting up a local autograph instance on your machine (and update the documentation)!
```
strseb: Btw thinking about this, we could make this all in taskcluster and just add a promote-style task, for when people are needing this to debug IAP. But thats fine to keep :)

strseb (approved):
r+wc

docs/Building/android.md
mcleinman: You can only go find things that you know exist. In this case, the code may be helpful to a new member of the team who otherwise wouldn't know to go dig in the git history for it. 

What do you think?

docs/Building/android.md
mcleinman: 🤦🏻 of course, removed this line. thanks

docs/Building/android.md
mcleinman: Agreed it could be done on TC. For now, I don't think it's worth worrying about - as I just ran into the first situation in 6+ months where I wanted to sign something, and I didn't actually *need* it. Just leaving these breadcrumbs in case it is helpful in the future to a developer working on this.

docs/Building/android.md
strseb: fair point, lets keep it :)

PR #10776, opened 2025-09-16
VPN-7255 - Assert correct QLocalSocket client

strseb:
## Description
On non debug builds we should  get  `BY_HANDLE_FILE_INFORMATION` for the client connecting to the windows daemon. 
Given we know that both client&daemon share the same file, we can fetch the unique identifier of the proccess and assert it's the same. 
From BY_HANDLE_FILE_INFORMATION: 
> The identifier (low and high parts) and the volume serial number uniquely identify a file on a single computer. To determine whether two open handles represent the same file, combine the identifier and the volume serial number for each file and compare them.


oskirby (approved):
Looks good to me

PR #10779, opened 2025-09-17
VPN-5469: Fix unsecured WiFi detection on macOS 14 and later

oskirby:
## Description
Starting with macOS 14, reading the Wi-Fi SSID requires location services to be enabled for the application, but unfortunately this would lead to a bunch of popups and security notices for our application that doesn't make sense for the minor functionality added by knowing the SSID (IIUC, it is only used to fill in the notification text).

However, the current `MacOSNetworkWatcher` implementation uses the existence of an SSID as a proxy for whether the current network interface is a Wi-Fi device, which causes us to ignore the network changes because we treat them as wired networks.

This PR works around the issue by checking for the 802.11 PHY mode instead to determine if the interface is Wi-Fi or Wired and then uses a generic text string in place of the SSID-specific notification text.

Also, because the BSSID/SSID may not be known, we should refactor the network watcher to use a global notification cooldown rather than one specific to the network.

## Reference
JIRA Issue: [VPN-5469](https://mozilla-hub.atlassian.net/browse/VPN-5469)

## Checklist
    
[VPN-5469]: https://mozilla-hub.atlassian.net/browse/VPN-5469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -281,6 +281,7 @@ notifications:
   connectionFailedMessage: Sorry, something went wrong with your VPN connection.
   subscriptionNotFound: Your Mozilla VPN subscription cannot be found. Please open the app to resolve this subscription issue.
   unsecuredNetworkTitle: Unsecured Wi-Fi network detected
+  unsecuredNetworkGeneric: "Wi-Fi is not secure. Click here to turn on VPN and secure your device."
   unsecuredNetworkMessage:
     value: "%1 is not secure. Click here to turn on VPN and secure your device."
     comment: "%1 is the Wi-Fi network name"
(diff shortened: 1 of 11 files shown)


flodolo (approved):

src/translations/strings.yaml
```diff
@@ -281,6 +281,7 @@ notifications:
   connectionFailedMessage: Sorry, something went wrong with your VPN connection.
   subscriptionNotFound: Your Mozilla VPN subscription cannot be found. Please open the app to resolve this subscription issue.
   unsecuredNetworkTitle: Unsecured Wi-Fi network detected
+  unsecuredNetworkGeneric: "Wi-Fi is not secure. Click here to turn on VPN and secure your device."
```
mcleinman: nit: We usually refer to it as "Wi-Fi network" in strings, I think
```suggestion
  unsecuredNetworkGeneric: "Wi-Fi network is not secure. Click here to turn on VPN and secure your device."
```

mcleinman (approved):
One nit, otherwise looks great. Thanks.

PR #10822, opened 2025-10-16
report curl failure

mcleinman:
reporting curl failure


oskirby:
This is a little overly complicated, can't we just add the `-f` or `--fail` cURL option (see: [here](https://curl.se/docs/manpage.html#--fail)) and let the github action fail out in the usual way?

oskirby (approved):

PR #10829, opened 2025-10-20
Remove subscription management screen

artfwo:
## Description

Remove subscription management screen. This change doesn't touch subscription data retrieval code and models, as those are still used by addons.

## Reference

[    i.e Jira or Github issue URL](https://mozilla-hub.atlassian.net/browse/VPN-7311)

## Checklist
    
### DIFF
--- a/src/ui/screens/getHelp/ViewGetHelp.qml
+++ b/src/ui/screens/getHelp/ViewGetHelp.qml
@@ -73,6 +73,18 @@ MZViewBase {
             accessibleName: title
         }
 
+        MZExternalLinkListItem {
+            objectName: "deleteAccount"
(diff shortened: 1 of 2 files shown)


artfwo:
> Thanks for those changes. They all look great! I'm just now remembering that we talked about moving the delete account to the help screen. What would you think about moving it there, and making it an `MZExternalLinkListItem` instead of a button? Otherwise, this is looking great! Thank you.

Right, thanks for the reminder! Now moved in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10829/commits/f01111e2c6acd704f2645fa25e4c0982830073b5 .

Issue #10808, opened 2025-10-12
Mozilla VPN with Qt 6.10 has some UI glitches (and a patch for build it)


XavierCLL:
### 1. Patch for build correctly with Qt 6.10

Because Qt 6.10 have new rule for private component it should be explicit find in cmake, patch here: [cmake-qt610-private.patch](https://aur.archlinux.org/cgit/aur.git/tree/cmake-qt610-private.patch?h=mozillavpn&id=4a687e61aeb20208ac45b54a2cbe0d7c37b36b48)

### 2. UI glitches with Qt 6.10

Although is possible build it with (1) the UI present some glitches, for example the main display is empty:

<img width="616" height="1138" alt="Image" src="https://github.com/user-attachments/assets/97813072-0d31-4b1d-8acd-f59cbb9f32c4" />

## VPN version:
2.31.1

## Affected Platforms:
All (but only tested in Archlinux)

PR #10832, opened 2025-10-21
VPN-7309: Qt 6.10 QML loading fixes

oskirby:
## Description
Some Linux distributions have started to ship Qt 6.10, and it seems that the VPN client fails to build, and even if it does build we fail to load the main screen.

The build failure is relatively straightforward to fix, and comes from missing dependencies on the `GuiPrivate` and `QmlPrivate` libraries. The fix for this was provided by @XavierCLL

However, it seems that once compiled, some QML content fails to load. I was able to trace this down to something in how the screens are instantiated by the `Navigator` class. It seems that creating them in C++ and then moving them into QML causes some unspecified breakage (the component starts loading but never finishes). We can work around the issue by instantiating the screens in QML. This means that we only need to provide the desired URL and we can cleanup the `Navigator` class a bit too.

## Reference
JIRA Issue: [VPN-7309](https://mozilla-hub.atlassian.net/browse/VPN-7309)
Github issue: #10808

## Checklist
    
[VPN-7309]: https://mozilla-hub.atlassian.net/browse/VPN-7309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/nebula/ui/components/navigator/MZNavigatorLoader.qml
+++ b/nebula/ui/components/navigator/MZNavigatorLoader.qml
@@ -56,7 +56,7 @@ StackView {
       if (stackView.get(pos+1).sourceComponent === null ||
           (MZNavigator.loadingFlags === MZNavigator.ForceReload)) {
         stackView.get(pos+1).sourceComponent = null;
-        stackView.get(pos+1).sourceComponent = MZNavigator.component;
+        stackView.get(pos+1).source = MZNavigator.loadUrl;
       }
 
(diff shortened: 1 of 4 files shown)


XavierCLL:
Thanks @oskirby I personally tested this patch on Archlinux and now the UI is working normally

PR #10839, opened 2025-10-27
VPN-6600: Android page size

mcleinman:
## Description

The range of libraries we include in our bundle required various fixes to support 16kb page sizes. Notably, we needed to upgrade Qt to 6.9.3 to do so.

I used Android Studio's APK Analyzer in this work. Ultimately, there are still two 4kb mentions in it:
1) A metadata file shows "4kb" in the analyzer, though not with the full 4kb error initially shown for other libraries. Including it may not cause us to fail checks. However, if it does, we may be able to remove it with no side effects by adding this to `build.gradle` files:
```
        resources {
          excludes += [
            "META-INF/com.google.android.gms.play-services-tasks.properties",
            "META-INF/**/play-services-tasks.properties"
          ]
        }
```

2) While there are no longer any libraries that show the `APK does not support 16 KB devices` warning, the [overall enclosing folder still shows that warning](https://github.com/user-attachments/assets/4b82aada-1a79-480d-9e77-0f123432668e). I have no idea why, and searching hasn't led to results. I'm wondering if this is an error? After this is merged, I want to upload an APK to test and see if Google Play accepts it.

## Reference

VPN-6600

## Checklist
    
### DIFF
--- a/android/adjust/build.gradle
+++ b/android/adjust/build.gradle
@@ -12,11 +12,6 @@ android {
         targetSdkVersion Config.targetSdkVersion
 
         buildConfigField  "String", "ADJUST_SDK_TOKEN" ,  '"' + (project.properties["adjusttoken"] ?: "no_token_found") +  '"'
-        externalNativeBuild {
-          cmake {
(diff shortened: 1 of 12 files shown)


oskirby (approved):
LGTM

PR #10855, opened 2025-11-05
Re-enable wireguard key rotation

artfwo:
## Description

We have couple of issues with regular wireguard key rotation: the server response is not very fast, and the client can attempt to connect while the key is being regenerated. This PR addresses the latter problem.

We have three places where key is regenerated currently:

1. On application startup, using [maybeRegenerateDeviceKey](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/mozillavpn.cpp#L1225), which regenerates keys if key version is < 2.5 and no custom DNS is used.
2. When connecting, using the same maybeRegenerateDeviceKey function.
3. From [KeyRegenerator](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/keyregenerator.cpp) with a timer, which regenerates keys [weekly](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/constants.h#L253) if the feature flag is enabled.

Both `KeyRegenerator` and `maybeRegenerateDeviceKey` schedule a TaskAddDevice to replace current device keys, and the server reply can take a few seconds to arrive (i've got as much as 7-8 seconds in my tests sometimes). VPN activation clears the task queue, so we may end up in a situation when the keys are updated on the server and not updated in the client.

This PR is one way to solve it by removing cases 1 and 3 above and moving the timer logic from KeyRegenerator under 2, which becomes a single point where key is regenerated immediately before connecting if required (see update `maybeRegenerateDeviceKey()` below). Failure to regenerate the key will abort connection.

There's a new controller state - RegeneratingKey - that updates the main screen (otherwise it's possible to trigger multiple key updates by repeatedly pushing the toggle).

One alternative solution I can think of would be locking guardian access (at least for adddevice and connection tasks), but that would require implementing some sort of locking mechanism.

Raising this mainly for discussion of the approach now, not for detailed review. More testing is probably needed around state changes (and some functional tests as well). One problem that I've discovered while working on this is that keys aren't updated when CLI command `mozillavpn activate` is used, that also seems to affect current stable version (no maybeRegenerateDeviceKey check?). Baku has also hinted that experimenting with cancelable flag for the affected tasks could result in a simpler fix.

## Reference

[VPN-2693](https://mozilla-hub.atlassian.net/browse/VPN-2693)

## Checklist
    
[VPN-2693]: https://mozilla-hub.atlassian.net/browse/VPN-2693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/translations/strings.yaml
```diff
@@ -1142,3 +1142,8 @@ navBar:
   settingsTab:
     value: Settings
     comment: Label for the tab on the navigation bar that takes the user to the settings screen
+
+controller:
+  regeneratingKey:
+    value: Updating keys
+    comment: Connection status label for the main screen
```
flodolo: This comment is not very helpful for localizers. Which keys are being updated?

artfwo: @flodolo thanks, i've improved it in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10855/commits/07c75bc0acefd98b940ec352bcfacc91d203d1ab

this string could be removed or changed depending on how this PR goes further, i'll ping you for re-reviewal of these parts then!

src/ui/screens/home/controller/ControllerView.qml
```diff
@@ -141,6 +141,46 @@ Item {
...
+            when: (VPNController.state === VPNController.StateRegeneratingKey)
+
+            PropertyChanges {
+                target: boxBackground
+                color: MZTheme.colors.primary
+            }
+
+            PropertyChanges {
+                target: logoTitle
+                //% "Connecting…"
```
strseb: This comment tells the qml-translation pipeline what "real" string we want to have for that, we only need that type of comment once, so it's not needed here :)

src/ui/screens/home/controller/VPNToggle.qml
```diff
@@ -150,6 +150,34 @@ MZButtonBase {
...
+
+            PropertyChanges {
+                target: toggle
+                color: MZTheme.colors.connectingToggleBackground
+                border.color: MZTheme.colors.primary
+            }
+
+            PropertyChanges {
+                target: toggleButton
+                //% "Turn VPN off"
```
strseb: Same here

src/constants.h
```diff
@@ -250,7 +250,7 @@ constexpr const char* AUTOGRAPH_STAGE_FINGERPRINTS[] = {
     nullptr  // list termination
 };
 
-PRODBETAEXPR(qint64, keyRegeneratorTimeSec, 604800, 300);
+PRODBETAEXPR(qint64, keyRegenerationTimeSec, 604800, 300);
```
strseb: 300 is quite aggressive for us dev's given we know it's slow. Therefore it might annoy us :D 
`43200` - 12h? So once per workday? Or 1h?

src/mozillavpn.cpp
```diff
@@ -973,8 +969,17 @@ void MozillaVPN::activate() {
   // is the right time to do it.
   maybeRegenerateDeviceKey();
 
-  TaskScheduler::scheduleTask(
-      new TaskControllerAction(TaskControllerAction::eActivate));
+  TaskScheduler::scheduleTask(new TaskFunction([this]() {
```
strseb, line 977: suggestion: maybe early return and no else branch :) 
```
TaskScheduler::scheduleTask(new TaskFunction([this]() {
    if (modelsInitialized()) {
      TaskScheduler::scheduleTask(
          new TaskControllerAction(TaskControllerAction::eActivate));
      return;
    } 
    logger.error() << "Failed to complete the key regeneration";
    REPORTERROR(ErrorHandler::RemoteServiceError, "vpn");
    setState(StateInitialize);
});
```

src/mozillavpn.cpp
```diff
@@ -1226,35 +1231,44 @@ void MozillaVPN::maybeRegenerateDeviceKey() {
...
+  if (Feature::get(Feature::Feature_keyRegeneration)->isSupported()) {
+    qint64 timeSinceLastKeyRegeneration =
+        QDateTime::currentSecsSinceEpoch() -
+        settingsHolder->keyRegenerationTimeSec();
+    logger.debug() << "Time since last key regeneration: "
+                   << timeSinceLastKeyRegeneration;
+
+    if (timeSinceLastKeyRegeneration > Constants::keyRegenerationTimeSec()) {
+      logger.debug() << "Key regeneration needed for an expired device key";
+      mustRegenerateKey = mustRegenerateKey | true;
```
strseb, line 1247: honest question (non-blocking): why did you choose to `mustRegenerateKey = (false | true)` instead of just `mustRegenerateKey = true`? :)

src/mozillavpn.cpp
```diff
@@ -1226,35 +1231,44 @@ void MozillaVPN::maybeRegenerateDeviceKey() {
...
   }
 
-  // We need a new device key only if the user wants to use custom DNS servers.
-  if (settingsHolder->dnsProviderFlags() ==
-      SettingsHolder::DNSProviderFlags::Gateway) {
-    logger.debug() << "Removal needed but no custom DNS used.";
-    return;
+  // We need a new device key if the user wants to use custom DNS servers
+  // and previous key was saved with version 2.5
+  if (!mustRegenerateKey && settingsHolder->dnsProviderFlags() !=
```
strseb, line 1253: thought: I wonder if this block is needed overall. Given 2.5.0 is more then a few years old. 
Maybe that is a good case for a glean-telemetry-event?

tests/functional/testConnection.js
```diff
@@ -181,4 +182,22 @@ describe('Connectivity', function() {
...
+  it('Connect to VPN - key regeneration is triggered for expired keys', async () => {
+    await vpn.waitForQuery(queries.screenHome.CONTROLLER_TITLE.visible());
+
+    await vpn.setSetting('connectionChangeNotification', 'true');
+    await vpn.setSetting('keyRegenerationTimeSec', 0);
+    await vpn.clickOnQuery(queries.screenHome.CONTROLLER_TOGGLE.visible());
+
+    await vpn.waitForCondition(async () => {
+      return await vpn.getQueryProperty(
+        queries.screenHome.CONTROLLER_SUBTITLE, 'text') === 'Preparing you connection';
```
strseb: typo? :)

tests/functional/testConnection.js
```diff
@@ -6,6 +6,7 @@ const assert = require('assert');
 const vpn = require('./helper.js');
 const queries = require('./queries.js');
 const setup = require('./setupVpn.js');
+const guardianEndpoints = require('./servers/guardian_endpoints.js')
```
oskirby: I don't see where this is used. Can it be removed?

src/controller.h
```diff
@@ -42,6 +42,10 @@ class Controller : public QObject, public LogSerializer {
     // a disconnected() signal from the ControllerImpl.
     StateDisconnecting,
 
+    // Requesting key to be regenerated due to version migration or expiration
+    // timeout.
```
oskirby: Can you expand on this comment to explain what happens after the key is regnerated? For example, do we go back to `StateIdle` or continue towards achieving an active connection?

oskirby (approved):
Looks good, just a couple of minor comments.

src/mozillavpn.cpp
artfwo: so the line is always `x = x | true` (or `& false`) and it's just easier to find, but i don't mind adapting the style if you have any other suggestions.

tests/functional/testConnection.js
artfwo: Right, a leftover from earlier test draft. Thanks for spotting!

src/mozillavpn.cpp
artfwo: Agreed, probably not needed. Are you suggesting we create a new event here to count 2.5 key migrations?

src/constants.h
artfwo: It seems the second value becomes active when staging servers are enabled. We can as well set it to 2 weeks (that's what mullvad recommends), I don't have a strong opinion here. I'll change it to 24h for now.

PR #10856, opened 2025-11-05
VPN-7324: 2.33 addon messages

mcleinman:
## Description

Moving towards giving more descriptive update and what's new messages. It's possible some of these features won't make it, and we'll hold the bullet point for the following release.

## Reference

VPN-7324

## Checklist
    

addons/strings.yaml
```diff
@@ -92,6 +92,41 @@ commonStrings:
     value: An option, under “Preferences”, to enable support for RCS/MMS messaging, Apple features like CarPlay and AirDrop, as well as connections to local devices.
     comment: First bullet point with a specific update in 2.31 - shown only on iOS
 
+233updateMessage:
+  bullet1:
+    value: "[all] Finish transitioning subscription management to the Mozilla Accounts website."
```
```suggestion
    value: "[all] Finish transitioning subscription management to the Mozilla accounts website."
```

Did you mean "Finished"? "Finish" sounds like a plan, not a release note. Also, accounts always lowercase.

addons/strings.yaml
```diff
@@ -92,6 +92,41 @@ commonStrings:
...
+    comment: Bullet point with a specific update in 2.33. Should have tag that indicates it applies to all platforms.
+  bullet2:
+    value: "[all] Added regular key rotation, improving security in the event of extreme, unexpected situations."
```
flodolo: `extreme, unexpected situations` sounds unnecessarily ominous. What are those?

addons/strings.yaml
```diff
@@ -92,6 +92,41 @@ commonStrings:
...
+    value: "[iOS] If you are using that beautiful new iPhone Air, Mozilla VPN should now look even better on it."
+    comment: Bullet point with a specific update in 2.33. Should have tag that indicates it applies to iOS platform.
+  bullet4:
+    value: "[iOS] Improvements on when notification permissions are requested when the app launches for the first time."
+    comment: Bullet point with a specific update in 2.33. Should have tag that indicates it applies to iOS platform.
+  bullet5:
+    value: "[Windows, macOS, Linux] In a rare situation, the status icon wasn’t updating. We dug in deep, and it should again work for everyone."
+    comment: Bullet point with a specific update in 2.33. Should have tag that indicates it applies to the 3 desktop platforms.
+  bullet6:
+    value: "[Linux] If your distro ships with that latest Qt 6.10, builds from scratch may have had issues. They no longer should."
```
flodolo, line 112: This seems relevant for an extremely small number of users? Will they even see the update message if they build their own?

mcleinman:
@flodolo our team had talked for a while about moving to better release notes. The "bug fixes and improvements" generalities weren't things we were happy with.

The hope was to move to this style of release notes for each release moving forward. Is there an amount of content that would be helpful to limit this to? (We'd need a decent amount, unfortunately - releasing to 5 platforms, we want everyone to see at least a couple improvements for their platform.) Or is this style of release notes something that is too taxing for our translators? We're definitely open to a discussion on this.

To give specific answers - 2.33 will have a first RC cut next Wednesday (Nov 12), with a release going out around Nov 19. Definitely could do another RC with additional translations between the 12th and 19th if desired.

addons/strings.yaml
mcleinman: A user that builds on their own will see the messages relevant to the version they built as - so they may indeed see this message.

Part of the goal for this was showing the amount of work that goes into this. So even if this affects a relatively small portion of users, it felt worth including.

addons/strings.yaml
mcleinman: Good catch. We went back and forth on tense here, and didn't get it perfect on this one. Will put up a fix shortly.

addons/strings.yaml
mcleinman: Your device is compromised and someone gets the keys from your device. Open to suggestions on how to improve this - but in general mentioning "your device could be compromised" seemed even more ominous, in my opinion.

flodolo:
> To give specific answers - 2.33 will have a first RC cut next Wednesday (Nov 12), with a release going out around Nov 19. Definitely could do another RC with additional translations between the 12th and 19th if desired.

If this is the plan going forward, we should have a bit more time for localization in the schedule.

mcleinman:
Tests are failing due to a separate issue. Since we need to get this merged to maximize time for translators, I'm merging this even with broken tests.

PR #10861, opened 2025-11-10
Disable caches in qt-toolchain tasks

strseb:
## Description

https://firefox-ci-tc.services.mozilla.com/tasks/RwYGQPLmT-OOsv40dTF9uQ

This is the cursed task - according to the logs we seem to pull qt.6.6.3. 
But looking at the artifact, we have both qt 6.9.2 (qt-host-tools\) and 6.6.3 (qt-host-tools\gcc_64\)! 

## What i think that happened: 

- The taskcluster update has seemed to force rebuilt of our toolchain tasks. 
- The tools-6.9 task was scheudled before the 6.6 task.
- The cache of the 6.9 task was reused on the 6.6. which seems to be the pwd.
- We have hence all the 6.9 files already there, causing the wierd artifact. 
--- 

What this patch does: 
1. Disable caches on toolchain tasks, by putting as task-default:
```yaml
    run: 
        use-caches: []
```

2. Change the artifact name to be explicit, so it's clear what version of qt-tools and qt are used in a task i.e 
```yaml

macos/opt:
    description: "macOS Build (release)"
   (..)
    fetches:
        toolchain:
            - qt-macos-6.6
            - qt-tools-6.6.3
```

### DIFF
--- a/taskcluster/kinds/build/addons.yml
+++ b/taskcluster/kinds/build/addons.yml
@@ -11,7 +11,7 @@ addons-bundle:
         platform: addons/all
     fetches:
         toolchain: 
-            - qt-tools-6.6.3
+            - qt-tools-6.6
     attributes:
         build-type: addons/opt  # needed by the signing kind
     worker-type: b-linux
--- a/taskcluster/kinds/build/macos.yml
+++ b/taskcluster/kinds/build/macos.yml
@@ -49,7 +49,7 @@ macos/opt:
             - cargo-vendor
             - conda-osxcross
             - qt-macos-6.6
-            - qt-tools-6.6.3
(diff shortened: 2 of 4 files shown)


taskcluster/kinds/build/addons.yml
```diff
@@ -11,7 +11,7 @@ addons-bundle:
         platform: addons/all
     fetches:
         toolchain: 
-            - qt-tools
+            - qt-tools-6.6.3
```
artfwo: Does this have to be `6.6.3` or would `6.6` suffice (as it does with qt version)?

strseb: that would work, let's make it consistent

PR #10865, opened 2025-11-14
VPN-7310: Refactor web extension bridge

oskirby:
## Description
This is an attempt to refactor the native messaging/web extension bridge. In doing so we eliminate the standalone Rust bridge project and move its functionality into the main VPN executable. While we're at it we also replace the TCP socket with a UNIX domain socket (or named pipe for Windows).

## Reference
JIRA Issue: [VPN-7310](https://mozilla-hub.atlassian.net/browse/VPN-7310)

## Checklist
    
[VPN-7310]: https://mozilla-hub.atlassian.net/browse/VPN-7310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6,7 +6,6 @@
 
 #include "commandlineparser.h"
 #include "leakdetector.h"
-#include "platforms/windows/windowsutils.h"
 #include "stdio.h"
 #include "version.h"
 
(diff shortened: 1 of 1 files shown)


src/webextension/server_macos.cpp
```diff
@@ -0,0 +1,18 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <QDir>
+#include <QStandardPaths>
+
+#include "server.h"
+
+bool WebExtension::Server::isAllowedToConnect(qintptr sd) {
+  // TODO: Implement Me!
```
strseb, line 11: Can we open a ticket for that? :)

oskirby: To a certain degree, it's a bit unclear what checks would be needed here. By switching to a UNIX domain socket with `UserAccessOption` set, this socket is only reachable by the same user account. So we don't really need to be as elaborate here compared to the daemon to guard against privilege escalation.

Issue #10867, opened 2025-11-16
Github Action RPM does not install on Fedora 43 and QT 6.10


arnobl:
**Describe the bug**
On Fedora 43 and Qt 6.10 (the QT version of the KDE Fedora).
Trying to install a dev version of the VPN using the RPM built by the Github action  *Linux packages*.
Does not work as the RPM requires `libQt6Gui.so.6(Qt_6.9_PRIVATE_API)(64bit)`.

**VPN version:**
I used the RPM package built by the Github action (Linux Packages) from Nov 16 (branch `main`).

**Affected Platforms:**
Fedora 43

**Steps to reproduce:**
Remove a previously installed version of the VPN. Update the system to use QT 6.10. Try to install the RPM.

**Expected result:**
Should install

**Actual result:**
DNF reports installation conflicts: `libQt6Gui.so.6(Qt_6.9_PRIVATE_API)(64bit)`

PR #10879, opened 2025-11-25
Build RPM packages for currently supported Fedora releases

artfwo:
## Description

Try to build RPMs on currently supported Fedora releases 43 and 42 (Fedora 41 is EOL by the end of November 2025).

## Reference

[JIRA](https://mozilla-hub.atlassian.net/browse/VPN-7368)
https://github.com/mozilla-mobile/mozilla-vpn-client/issues/10867

## Checklist
    
### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -14,6 +14,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  supported-fedora-releases:
+    name: Fetch currently supported Fedora releases
+    runs-on: ubuntu-22.04
+    outputs:
+      supported-releases: ${{ steps.fetch-supported-releases.outputs.releases }}
+    steps:
+      - id: fetch-supported-releases
+        name: Fetch currently supported Fedora releases
+        run: |
+          API_RESPONSE=$(curl -s https://endoflife.date/api/v1/products/fedora)
+          SUPPORTED_RELEASES=$(echo $API_RESPONSE | jq -c '[ .result.releases[] | select(.isEol == false) | { "fedora-version": .name } ]')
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yaml
```diff
@@ -45,11 +45,16 @@ jobs:
...
   rpmbuild:
-    name: RPM Packages
+    name: RPM Packages (Fedora ${{ matrix.fedora-version }})
     needs: source-bundle
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # currently supported Fedora releases
+        fedora-version: [43, 42]
```
oskirby: Non-blocking comment: Is there an API or something that we can easily fetch to get the list of fedora versions dynamically? For example, the PPA workflows fetch the [distro-info-data](https://launchpad.net/ubuntu/+source/distro-info-data) to determine the currently supported releases.

If this is something that we have to maintain manually, it's also something that we'll inevitably forget to do.

artfwo: I've updated the workflow to fetch fedora release status from https://endoflife.date/fedora

Interestingly, it tries to build for FC41 which is [scheduled for EOL on November, 19th](https://fedorapeople.org/groups/schedule/f-41/f-41-key-tasks.html), but still not listed in the [end of life release list](https://docs.fedoraproject.org/en-US/releases/eol/), so it's likely still alive. I guess we should be more of less fine with that.

strseb (approved):

.github/workflows/linux.yaml
```diff
@@ -14,6 +14,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  supported-fedora-releases:
```
oskirby, line 17: Nicely done!

PR #10889, opened 2025-12-03
VPN-7399: Fix crash when toggling addons signature feature

mcleinman:
## Description

This was introduced in [this PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10863), when we reordered our initialization to support modern Android.

This change resulted in not setting `stage` until after the feature model was initialized, which then caused features using `FeatureCallback_inStaging` to not be setup correctly.

## Reference

VPN-7399

## Checklist
    

oskirby (approved):
LGTM

PR #10891, opened 2025-12-04
VPN-2199: Windows ARM Support

oskirby:
## Description
This should get most of the way towards producing an arm64 build for Windows. Much of the heavy lifting to get here was previously done by the work to get cross-compilation to Windows working from Linux.

This PR:
- Updates the Rust version and adds the `aarch64-pc-windows-msvc` target.
- Undo hardcoded assumptions that Windows implies x86-64.
- Split the windows cross toolchain files into `x86_64-toolchain.cmake` and `aarch64-toolchain.cmake`
- Install both `x64` and `arm64` targets of the windows SDK into the cross conda environment.
- Add a toolchain job to fetch the VS2022 SDK for arm64.
- Add a toolchain job to build static Qt for arm64 on Windows.
- Patch in a workaround for briansmith/ring#2216

I am currently struggling a bit to figure out how to package the resulting artifacts as an MSI installer, so I think I am going to leave that piece for a future PR.

## Reference
JIRA Issue: [VPN-2199](https://mozilla-hub.atlassian.net/browse/VPN-2199)
Depends on: briansmith/ring#2216
Depends on: Bugzilla [1944293](https://bugzilla.mozilla.org/show_bug.cgi?id=1944293)

## Checklist
    
[VPN-2199]: https://mozilla-hub.atlassian.net/browse/VPN-2199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


windows/split-tunnel/get.ps1
strseb, line 1: I don't seem to find a refrence to this getting called, should we remove this?

taskcluster/config.yml
```diff
@@ -52,6 +52,11 @@ workers:
             implementation: generic-worker
             os: windows
             worker-type: b-win2022
+        b-win11-2024h2-arm64:
```
strseb, line 55: RN we only have level-1 Workers, we should followup with releng, that we also get the level-3 queue up before we merge this, otherwise those tasks will be stuck :)

3rdparty/sentry
strseb, line 1: I think this breaks the tc-ios tasks :c

windows/split-tunnel/get.ps1
oskirby: That is fine by me.

taskcluster/config.yml
oskirby: Ah! Good to know

3rdparty/sentry
oskirby: Looks like the change that broke things was getsentry/sentry-native#1083 which got released in sentry-native 0.7.14. We could probably just revert that back to 0.7.13 to get unstuck, which still includes the fixes for Windows arm64.

PR #10894, opened 2025-12-04
VPN-6600 addendum: Update glean to fix library on x86_64 devices

mcleinman:
## Description

One of the Android architectures wasn't ready for 16 kb page sizes. This was due to a library that was packaged by Glean. Thus, we needed to upgrade Glean. That required an upgrade to python, rust, and several other major dependencies. Additionally, we needed to do a bunch of work to our custom glean implementation.

I believe this is a minimal diff. Feel free to add commits and emerge when I'm on PTO.

Transparency: A lot of the Linux build work was done w/ the help of Claude.

## Reference

VPN-6600, part 2

## Checklist
    
### DIFF
--- a/linux/debian/rules
+++ b/linux/debian/rules
@@ -36,12 +36,6 @@ override_dh_clean:
 	dh_clean --exclude=3rdparty
 
 override_dh_auto_configure:
-	@echo "DEBUG [debian/rules]: HOME=$(HOME)"
-	@echo "DEBUG [debian/rules]: PATH=$(PATH)"
-	@echo "DEBUG [debian/rules]: RUSTUP_HOME=$(RUSTUP_HOME)"
-	@echo "DEBUG [debian/rules]: CARGO_HOME=$(CARGO_HOME)"
-	@which rustc || echo "rustc not found in PATH"
-	@which cargo || echo "cargo not found in PATH"
 	dh_auto_configure -- -DWEBEXT_INSTALL_LIBDIR=/lib -DBUILD_ID=$(DEB_VERSION) -DCMAKE_CXX_STANDARD=17 -DBUILD_TESTS=OFF ${OPENSSL_BUILD_ARGS}
 
 override_dh_auto_test:
--- a/taskcluster/scripts/build/linux_build_dpkg.sh
+++ b/taskcluster/scripts/build/linux_build_dpkg.sh
@@ -6,28 +6,6 @@
 
 set -e
 
-# Source cargo environment if rustup is installed
-echo "DEBUG: HOME is $HOME"
-echo "DEBUG: Checking for rustup installation..."
-if [ -f "$HOME/.cargo/env" ]; then
-    echo "DEBUG: Found $HOME/.cargo/env, sourcing it..."
-    source "$HOME/.cargo/env"
-    echo "DEBUG: After sourcing, PATH is: $PATH"
-    echo "DEBUG: which rustup: $(which rustup 2>&1 || echo 'not found')"
-    echo "DEBUG: which cargo: $(which cargo 2>&1 || echo 'not found')"
-    echo "DEBUG: which rustc: $(which rustc 2>&1 || echo 'not found')"
-    if which rustup >/dev/null 2>&1; then
-        echo "DEBUG: rustup show:"
-        rustup show 2>&1 || echo "rustup show failed"
-        echo "DEBUG: rustup default:"
-        rustup default 2>&1 || echo "rustup default failed"
-    fi
-else
-    echo "DEBUG: $HOME/.cargo/env not found!"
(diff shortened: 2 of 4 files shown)


linux/debian/control
```diff
@@ -9,7 +9,7 @@ Build-Depends: debhelper (>= 13),
                gcc (>=4:8.0.0~),
                g++ (>=4:8.0.0~),
                golang (>=2:1.18~),
-               cargo (>=1.75),
+               cargo (>=1.82),
```
oskirby, line 12: This will cause a failure to build when submitted to the launchpad PPA build servers. The reason we set the minimum version to 1.75 here is because that's the version of `cargo` available from the Ubuntu security updates for Ubuntu/Jammy.

There should be workaround available somehow, and in theory we should be able to install an official 1.82 version (see: https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/rust/)

Sometimes a backport is available, in which case specifying the dependency as: `cargo (>= 1.82) | cargo-1.82` might do the trick. We have had to do similar workarounds in the past like this (eg: see #6855 for how `gcc-8` compatibility was patched into Ubuntu/Bionic).

oskirby: If this works, all the hacking to get `rustup` to work should become unnecessary, which would clean up this PR significantly.

qtglean/android.cmake
```diff
@@ -6,6 +6,19 @@ include(${CMAKE_SOURCE_DIR}/scripts/cmake/rustlang.cmake)
 
 set(GLEAN_VENDORED_PATH ${CMAKE_SOURCE_DIR}/3rdparty/glean)
 
+message(STATUS "Building Glean Rust library...")
```
oskirby, line 9: What exactly does this bundle build and how does it get consumed by the VPN project?

I have a bit of skepticism around this change, because I don't see how the target architecture would get passed to the resulting artifacts. I would expect this to produce libraries targeting the host architecture when doing a naked `cargo build --releas` like this.

qtglean/ios.cmake
```diff
@@ -79,6 +82,19 @@ target_sources(iosglean PUBLIC
     ${GLEAN_VENDORED_PATH}/glean-core/ios/Glean/Glean.h
 )
 
+message(STATUS "Building Glean Rust library...")
+execute_process(
```
oskirby, line 86: Same comment here as for `qtglean/android.make`

oskirby:
@artfwo to update the Cargo.lock file, I had to manually tweak the versions of some of the packages by specifying an explicit version.

So, I started by updating all the dependencies with `cargo update`, and then rolled back the problematic ones by doing `cargo update <problematic-package>@<latest-version-that-supports-1.82>` and then working through the dependencies until the project builds.

IIUC, it was really just the `icu_provider` and friends that were causing grief.

qtglean/android.cmake
artfwo: Without the bundle, `uniffi-bindgen generate` stage will fail with the following error:
```
% RUST_BACKTRACE=1 3rdparty/glean/target/debug/uniffi-bindgen generate --language swift --out-dir /Users/apopov/work/mozilla-vpn-client/build-ios/src/glean /Users/apopov/work/mozilla-vpn-client/3rdparty/glean/glean-core/src/glean.udl
thread 'main' panicked at tools/embedded-uniffi-bindgen/src/main.rs:48:5:
libglean_ffi could not be found in /Users/apopov/work/mozilla-vpn-client/3rdparty/glean/target
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: uniffi_bindgen::find_library_file
             at ./3rdparty/glean/tools/embedded-uniffi-bindgen/src/main.rs:48:5
   3: uniffi_bindgen::gen_bindings
             at ./3rdparty/glean/tools/embedded-uniffi-bindgen/src/main.rs:103:36
   4: uniffi_bindgen::main
             at ./3rdparty/glean/tools/embedded-uniffi-bindgen/src/main.rs:170:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

With this code, `libglean_ffi.dylib` (or `libglean_ffi.so`) is generated under `3rdparty/glean/target/release`, where `uniffi_bindgen::find_library_file` expects it.

PR #10895, opened 2025-12-05
Workaround for the crash on Pixel devices

artfwo:
## Description

This workaround seems to fix the crash we're experiencing on some Pixel devices, as [described](https://qt-project.atlassian.net/browse/QTBUG-140501?focusedCommentId=2228362) in the Qt bug tracker (we're also getting "W/Qt A11Y : Accessibility: populateNode for Invalid ID" messages, but not many of them). As the impact seems to be wider than just the Pixel device it's happening on, maybe we can use this until the underlying issue is properly fixed?

## Reference

[VPN-7377](https://mozilla-hub.atlassian.net/browse/VPN-7377)

## Checklist
    
[VPN-7377]: https://mozilla-hub.atlassian.net/browse/VPN-7377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -2390,6 +2390,12 @@ int MozillaVPN::runGuiApp(std::function<int()>&& a_callback) {
   callback();
 #endif
 
+#if defined(MZ_ANDROID) && (QT_VERSION > QT_VERSION_CHECK(6, 10, 1))
+#  error \
+      "Revisit the code below and replace QCoreApplication::exec() \
+      with app.exec() if the app no longer crashes on Pixel 2 XL, \
+      see https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10895"
+#endif
   return QCoreApplication::exec();
 }
 


mcleinman (commented):
Were you able to repro the issue? Did you confirm this fixes it? If so, I'm okay with this fix.

I think we should leave a comment here in the code as to why this isn't `app.exec()`, to prevent us from erroneously changing it back to match all the other `app.exec()` instances in the code.

src/mozillavpn.cpp
```diff
@@ -2390,7 +2390,7 @@ int MozillaVPN::runGuiApp(std::function<int()>&& a_callback) {
   callback();
 #endif
 
-  return app.exec();
+  return QCoreApplication::exec();
```
strseb, line 2399: Could you add a macro that generates an error when QT version is bigger then 6.10 && platform == android with a comment that we should re-visit this fix on the next QT upgrade.

PR #10913, opened 2025-12-15
VPN-7378 - change Window Decoration background on Android based on Sy…

strseb:
## Description

Given that our window now has insets, the easieest things to recover the top App Bar is to 
draw a contrast color below it, just like we did before :) 

Agreed, this is a bit hacky. The Proper way probably would be to use https://developer.android.com/reference/android/view/WindowInsetsController and tell android how the app is rendered and what app-bar text color we'd need - however i assume that most people will just use the system default, which just works. So i wont bother.

### DIFF
--- a/src/ui/main.qml
+++ b/src/ui/main.qml
@@ -84,7 +84,7 @@ ApplicationWindow {
     color: {
         // This is only relevant on android
         if(Qt.platform.os !== "android"){
-         return MZTheme.colors.bgColor;
+            return MZTheme.colors.bgColor;
         }
         if (MZTheme.currentSystemTheme === MZTheme.currentTheme) {
             return MZTheme.colors.bgColor;


src/ui/main.qml
```diff
@@ -81,7 +81,20 @@ ApplicationWindow {
...
+         return MZTheme.colors.bgColor;
+        }
+        if (MZTheme.currentSystemTheme == MZTheme.currentTheme) {
+            return MZTheme.colors.bgColor;
+        } 
+        if (MZTheme.currentSystemTheme === "main") {
+            return "white";
+        } else {
+            return "black";
+        }
```
artfwo, line 96: Inconsistent indentations and usage of `==` here. Would it work if the first condition is `!Qt.platform.os === "android" || MZTheme.currentSystemTheme === MZTheme.currentTheme`?

src/ui/main.qml
```diff
@@ -83,10 +83,10 @@ ApplicationWindow {
     title: MZI18n.ProductName
     color: {
         // This is only relevant on android
-        if(!Qt.platform.os === "android"){
+        if(Qt.platform.os !== "android"){
          return MZTheme.colors.bgColor;
```
artfwo: Still mixed indentation levels :)

```suggestion
            return MZTheme.colors.bgColor;
```

If you'd like to separate the android OS condition from `MZTheme.currentSystemTheme === MZTheme.currentTheme`, have you considered restructuring the binding as `color: Qt.platform.os !== "android" ? MZTheme.colors.bgColor : { ... }`

strseb: I've tried to use a block expression there but it seems qt-compiler really does not like it :c

PR #10919, opened 2025-12-16
VPN-7429: Update taskcluster Qt builds to 6.10

oskirby:
## Description
This should update our taskcluster builds to Qt version 6.10.1 for all the desktop platforms and WASM. Since this is the latest and greatest that Qt has to offer, there's no use in the Qt/next builds anymore so we have removed their build jobs too.

Some other details of note:
- We now have both x86_64 and aarch64 release builds for Windows. As such there was a bit of reorganizing of the task names:
  - `build-windows/opt` -> `build-windows-x86_64/opt`
  - `signing-windows/opt` -> `signing-windows-x86_64/opt`
  - Added `build-windows-aarch64/opt`
  - Added `signing-windows-aarch64/opt`
  - Added `repackage-msi-aarch64`
 - The `emsdk` conda package hasn't been maintained in forever and this was holding back the WASM upgrade. So we have removed the package and are installing EMSDK via a fetch instead.
 - Qt 6.10 added ICU/localization support using the windows SDK via the `-feature-winsdkicu` feature. Unfortunately, this resulted in linker issues when building for windows, on windows taskcluster workers, so I opted to disable this new feature. FWIW, this feature seemed to work fine on my local windows dev machine. 

There is a little more work to go in order to sign the aarch64 MSI installer, but I will save that for another PR.

## Reference
JIRA Issue: [VPN-7429](https://mozilla-hub.atlassian.net/browse/VPN-7429)

## Checklist
    
[VPN-6764]: https://mozilla-hub.atlassian.net/browse/VPN-6764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPN-7429]: https://mozilla-hub.atlassian.net/browse/VPN-7429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/taskcluster/kinds/build/windows.yml
+++ b/taskcluster/kinds/build/windows.yml
@@ -44,9 +44,7 @@ windows-x86_64/native:
     run:
         exec-with: powershell
         command: taskcluster/scripts/build/windows_clang_cl.ps1
-    # TODO: Set this before merging to minimize the need for Windows
-    # workers.
-    #requires-level: 3
+    requires-level: 3
 
 windows-x86_64/opt:
     description: "Windows Build (Release/x86_64)"


taskcluster/kinds/fetch/kind.yml
```diff
@@ -13,17 +13,6 @@ task-defaults:
     docker-image: {in-tree: base}
 
 tasks:
-    win-dev-env:
```
strseb, line 16: Finally i can nuke that bucket. Thanks!

env-apple.yml
```diff
@@ -27,4 +27,4 @@ dependencies:
       - -r requirements.txt
       - -r taskcluster/scripts/requirements.txt
 variables:
-  QT_VERSION: 6.6.3
+  QT_VERSION: 6.9.3
```
artfwo: If taskcluster builds are at 6.10 already, maybe we can as well bump Qt versions here and in env-windows.yml below?

oskirby: Good catch! I keep forgetting about these (in part because taskcluster doesn't use them)

PR #10930, opened 2025-12-23
VPN-7385: fix mac quitting issue

mcleinman:
## Description

macOS was having issues when quitting - when VPN was active it would hang, and when VPN was inactive it would close normally but macOS would pop a "do you want to report crash to developer?" prompt. (Or potentially the reverse - I forget.)

I believe this fixes the issue, and from my (macOS) testing doesn't mess anything up for the extension. The reason this is new: We [changed from QTcpServer to QLocalServer in a PR since](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10865) the 2.32 release which [according to the documentation](https://doc.qt.io/qt-6/qlocalserver.html) should only get a adapter passed to it if it is the parent. A slight change, but enough to affect this apparently.

## Reference

VPN-7385

## Checklist
    
### DIFF
--- a/src/commands/commandui.cpp
+++ b/src/commands/commandui.cpp
@@ -176,9 +176,6 @@ int CommandUI::run(QStringList& tokens) {
   }
 #endif
 
-#ifdef MVPN_WEBEXTENSION
-  std::unique_ptr<WebExtension::Server> extensionServer{nullptr};
-#endif
   std::unique_ptr<KeyRegenerator> keyRegenerator{nullptr};
 
   // Ensure that external styling hints are disabled.
@@ -382,11 +379,11 @@ int CommandUI::run(QStringList& tokens) {
 #endif
 
 #ifdef MVPN_WEBEXTENSION
-    QPointer newServer =
+    QPointer webExtensionServer =
         new WebExtension::Server{new WebExtensionAdapter(qApp)};
-    extensionServer.reset();
     QObject::connect(vpn->controller(), &Controller::readyToQuit,
-                     extensionServer.get(), &WebExtension::Server::close);
+                     webExtensionServer.get(), &WebExtension::Server::close);
+    webExtensionServer.clear();
 #endif
 
 #ifdef MZ_ANDROID


strseb (changes requested):
> [according to the documentation](https://doc.qt.io/qt-6/qlocalserver.html) should only get a adapter passed to it if it is the parent.

Failing to find the refence?
While the fix is correct, this kinda is counter intutive on how qt works. 

We should fix it here 
```
    extensionServer.reset(
        new WebExtension::Server{new WebExtensionAdapter(qApp)});
```


The problem is during creation i passed a parent but also put it in a unique_ptr. So we delete the pointer once on the destructor of qApp and when commandui scope gets destroyed. 

I'd prefrer we just change the unique_ptr to a qPointer. 

(Honestly i should write a clazy check for that... that is a footgun)

mcleinman:
> > [according to the documentation](https://doc.qt.io/qt-6/qlocalserver.html) should only get a adapter passed to it if it is the parent.
> 
> Failing to find the refence? While the fix is correct, this kinda is counter intutive on how qt works.
> 
> We should fix it here
> 
> ```
>     extensionServer.reset(
>         new WebExtension::Server{new WebExtensionAdapter(qApp)});
> ```
> 
> The problem is during creation i passed a parent but also put it in a unique_ptr. So we delete the pointer once on the destructor of qApp and when commandui scope gets destroyed.
> 
> I'd prefrer we just change the unique_ptr to a qPointer.
> 
> (Honestly i should write a clazy check for that... that is a footgun)

Thanks, updated.

src/commands/commandui.cpp
strseb: Remove that, it's not needed :)

src/commands/commandui.cpp
```diff
@@ -382,8 +382,9 @@ int CommandUI::run(QStringList& tokens) {
 #endif
 
 #ifdef MVPN_WEBEXTENSION
-    extensionServer.reset(
-        new WebExtension::Server{new WebExtensionAdapter(qApp)});
+    QPointer newServer =
```
strseb: maybe `QPointer webExtensionServer`?

src/commands/commandui.cpp
```diff
@@ -382,8 +382,9 @@ int CommandUI::run(QStringList& tokens) {
...
+        new WebExtension::Server{new WebExtensionAdapter(qApp)};
+    extensionServer.reset();
     QObject::connect(vpn->controller(), &Controller::readyToQuit,
                      extensionServer.get(), &WebExtension::Server::close);
```
strseb: This should derefrence the new pointer :)

mcleinman:
@strseb made some changes that I think make far more sense now. Would love one more quick review.

PR #10934, opened 2025-12-29
VPN-7417: fix android colors on older devices

mcleinman:
## Description

Some of the recent Android changes resulted in a nearly-invisible status bar and navigation buttons (triangle/circle/square) on older Android devices (you can barely see them):
Untitled

This fixes that (by manually setting a color, along with not setting some of the new flags), while not introducing regressions to more modern Android devices:
Screenshot 47

## Reference

VPN-7417

## Checklist
    
### DIFF
--- a/src/platforms/android/androidcommons.cpp
+++ b/src/platforms/android/androidcommons.cpp
@@ -159,7 +159,6 @@ void AndroidCommons::setStatusBarTextColor(bool isLight) {
       window.callMethod<void>("setStatusBarColor", "(I)V", 0xFF000000);
     }
   });
-
 }
 
 bool AndroidCommons::clearPendingJavaException(const char* where) {


src/platforms/android/androidcommons.cpp
```diff
@@ -149,6 +149,18 @@ void AndroidCommons::launchPlayStore() {
                                      appActivity.object());
 }
 
+void AndroidCommons::setStatusBarTextColor(QString statusBarTextColor) {
```
strseb: VERY nitpicky here, but is a string the best argument here? Maybe a bool? Or an Enum.

mcleinman: This is part of #10945 that got merged into here. From a comment in that PR: `I didn't include Theme [the eisting enum] in AndroidCommons because it ended up with a circular dependency terribleness because of the VPNAndroidCommons class. For expediency, I just turned it into a string.`

Will turn into a boolean, though.

PR #10944, opened 2026-01-05
VPN-6405: fix red dot on notifications bar button

mcleinman:
## Description

After an incredibly deep dive in learning about SettingGroups, this ended up as a 2 line fix. While we setup the setting groups properly, they apparently look for related settings that are *persisted to disk*. Since a default setting isn't persisted to disk, it missed when it changed (as from one perspective, the setting was actually being set for the first time), and thus didn't update the bar button.

## Reference

VPN-6405

## Checklist
    

oskirby (approved):
LGTM

PR #10960, opened 2026-01-12
Use the in-memory buffer for loading add-ons

bakulf:
## Description

It can happen that an add-on is modified between the computation of its hash and the loading of the resources. This patch fixes the issue loading the add-on from the memory buffer.

## Reference

https://bugzilla.mozilla.org/show_bug.cgi?id=2009454

## Checklist
    

oskirby:
Looks good to me!

PR #10965, opened 2026-01-13
Re-introduce polkit with additional auth checks on Linux

artfwo:
## Description

Re-introduce polkit with additional authorization checks on Linux. `status()` and `cleanupLogs()` check now authorizations via polkit on Linux.

Authentication for dbusservice methods can be skipped for admin users if an appropriate polkit rule is installed as described in `docs/building/Linux.md`

The rule is automatically installed by Debian and Fedora packaging scripts.

## Reference

[VPN-7391](https://mozilla-hub.atlassian.net/browse/VPN-7391)

## Checklist
    
[VPN-7391]: https://mozilla-hub.atlassian.net/browse/VPN-7391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cmake/linux.cmake
+++ b/src/cmake/linux.cmake
@@ -187,7 +187,7 @@ if(NOT BUILD_FLATPAK)
         install(FILES ${CMAKE_SOURCE_DIR}/linux/org.mozilla.vpn.rules-debian
             RENAME org.mozilla.vpn.rules
             DESTINATION ${CMAKE_INSTALL_DATADIR}/polkit-1/rules.d)
-    elif(EXISTS /etc/redhat-release)
+    elseif(EXISTS /etc/redhat-release)
         install(FILES ${CMAKE_SOURCE_DIR}/linux/org.mozilla.vpn.rules-others
             RENAME org.mozilla.vpn.rules
             DESTINATION ${CMAKE_INSTALL_DATADIR}/polkit-1/rules.d)


docs/Building/linux.md
```diff
@@ -44,6 +45,16 @@ Install
 
     sudo cmake --install build
 
+Copy polkit authorization rules
```
oskirby: This could be replaced by adding a cmake `install(FILE ...)` command. For example:

```cmake
if(EXISTS /etc/debian_version)
   install(FILES ${CMAKE_SOURCE_DIR}/linux/org.mozilla.vpn.rules-debian
        RENAME org.mozilla.vpn.rules
        DESTINATION ${CMAKE_INSTALL_DATADIR}/polkit-1/rules.d)
elif(EXISTS /etc/redhat-release)
   install(FILES ${CMAKE_SOURCE_DIR}/linux/org.mozilla.vpn.rules-others
        RENAME org.mozilla.vpn.rules
        DESTINATION ${CMAKE_INSTALL_DATADIR}/polkit-1/rules.d)
else()
   message(INFO "Unknown Linux distribution, please install polkit rules manually")
endif()
```

linux/debian/control
```diff
@@ -17,6 +17,7 @@ Build-Depends: debhelper (>= 13),
                libgcrypt20-dev,
                libgl-dev,
                libopengl-dev (>= 1.3.0~),
+               libpolkit-gobject-1-dev,
```
oskirby, line 20: Does this require a runtime dependency as well?

docs/Building/linux.md
artfwo: Good idea, fixed in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10965/commits/4d6571a1b1fec4552294ada5d276c5144be28744 and https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10965/commits/816d4d8f0af2cc5f4298ace170c595d70a85e8f2

linux/debian/control
artfwo: In this case it should be handled by dh_shlibdeps and dh_gencontrol, but we don't use ${shlibs:Depends} substitution variable in debian/control to include that, so I've just added it in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10965/commits/973cc61b397ba651a9e7c1ae51d5a94e04bdf334

artfwo: Ok, I see why we don't use subst variables now. Reverted and added the dependency in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10965/commits/ca4d72969bf2df42fb2d4d8d51c4241d75319336

Issue #10826, opened 2025-10-19
Blank screen (due to timer overflow?)


emilio:
**Describe the bug**

The GUI is blank. The console is totally flooded by:

> [19.10.2025 19:32:01.315] Warning: QTimer::start: negative intervals aren't allowed; the interval will be set to 1ms.

**VPN version:**

Mozilla VPN 2.31.1

**Affected Platforms:**

Linux

**Steps to reproduce:**

Run the mozillavpn binary.

**Expected result:**

A working GUI.

**Actual result:**

Blank screen.

**Additional notes:**

I think at least the console spam seems like an integer overflow?

```
(gdb) p now
$14 = {d = {d = 0x199fdf5e9828f, data = {status = 143, msecs = 1760902375810}}}
(gdb) p expire
$15 = {d = {d = 0x19a8dec90f94f, data = {status = 79, msecs = 1763317682425}}}
(gdb) p now.msecsTo(expire)
$16 = 2418906615
(gdb) bt
#0  checkInterval (caller=0x7ffff5086699 "QTimer::start", interval=...) at /usr/src/debug/qt6-base/qtbase/src/corelib/kernel/qtimer.cpp:252
#1  QTimer::start (this=0x555556f14988, interval=...) at /usr/src/debug/qt6-base/qtbase/src/corelib/kernel/qtimer.cpp:285
#2  QTimer::start (this=0x555556f14988, msec=-1876060681) at /usr/src/debug/qt6-base/qtbase/src/corelib/kernel/qtimer.cpp:244
#3  0x0000555555710e34 in AddonConditionWatcherTriggerTimeSecs::maybeStartTimer (this=0x555556f14970)
    at /usr/src/debug/mozillavpn/mozillavpn-2.31.1/src/addons/conditionwatchers/addonconditionwatchertriggertimesecs.cpp:56
#4  0x0000555555710f01 in operator() (__closure=0x555556ee6a40)
    at /usr/src/debug/mozillavpn/mozillavpn-2.31.1/src/addons/conditionwatchers/addonconditionwatchertriggertimesecs.cpp:30
#5  operator() (__closure=<optimized out>) at /usr/include/qt6/QtCore/qobjectdefs_impl.h:116
[...]

PR #10968, opened 2026-01-13
Fix timer rollover in AddonApi::setTimedCallback

oskirby:
## Description
After ignoring my macbook for a couple of weeks, I returned to it in order to discover that it was displaying the infinite timer loop bug reported in #10826 so I went and chased down the root cause.

The problem occurs because of an integer truncation bug when scheduling a timer via `AddonApi::setTimedCallback()` with a duration greater than 24 days. This can occur when the `message_upgrade_to_annual_plan` addon detects that the subscription is between 14 and 63 days old.

Changing the `AddonApi::setTimedCallback()` API to take a 64-bit integer for the timeout value prevents the truncation from occurring.

## Reference
Github issue: #10826

## Checklist
    
### DIFF
--- a/src/addons/addonapi.cpp
+++ b/src/addons/addonapi.cpp
@@ -94,7 +94,7 @@ void AddonApi::initialize() {
   m_timer.setSingleShot(true);
 }
 
-void AddonApi::setTimedCallback(int interval, const QJSValue& callback) {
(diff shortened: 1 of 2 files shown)


artfwo (approved):
Great find, thanks!

Issue #10806, opened 2025-10-10
CLI `wgconf` broken for multihop


Erik-S:
**Describe the bug**
The `wgconf` CLI command prints a conf file for the current connection. For multihop connections, it prints out a single-hop config due to a bug. 

See [Mullvad docs](https://mullvad.net/en/help/wireguard-and-mullvad-vpn#multihop-wireguard) on multihop for the correct format. 

Buggy code: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/3fcb4bcaffe740d0b59a00fa1c45045fcfcfe029/src/commands/commandwgconf.cpp#L73-L76

**VPN version:**
2.31.0 (2.202509161519)

**Affected Platforms:**
All platforms with CLI.

**Prerequisites:**
None (other than a multihop connection)

**Steps to reproduce:**
1. Set up a multihop VPN connection.
2. Run `mozillavpn wgconf` (`'\Mozilla VPN.exe' wgconf` on Windows)

**Expected result:**
The `Endpoint` line should be:

```
Endpoint = <entrypoint server IP>:<exit server multihop port>
```

**Actual result:**
We get both the wrong IP and the wrong port. Note that this is a working single-hop config, so it's hard to detect.

```
Endpoint = <exit server IP>:<entrypoint server multihop port>
```

**Additional notes:**

PR #10973, opened 2026-01-15
VPN-7308: Fix wgconf command for multi hop and CLI regressions

encrypt94:
## Description

-  Rework `SettingsHolder` initialization to rely exclusively on `instance()`, fixing CLI breakage caused by double initialization . Update all affected tests
-  Fix `servers` CLI command failing due to `ServerData` not being initialized before `TaskServers` is invoked
-  Fix `wgconf` CLI command not working correctly with multi-hop 
-  Update `clang-format-hooks` broken by LLVM 18

## Reference

-  Jira Issue: [VPN-7308](https://mozilla-hub.atlassian.net/browse/VPN-7308)
- Github Issue: #10806 
- clang-format-hooks [github issue](https://github.com/barisione/clang-format-hooks/issues/33)

## Checklist
    
[VPN-7308]: https://mozilla-hub.atlassian.net/browse/VPN-7308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commands/commandwgconf.cpp
+++ b/src/commands/commandwgconf.cpp
@@ -36,14 +36,14 @@ int CommandWgConf::run(QStringList& tokens) {
     QString appName = tokens[0];
 
     CommandLineParser::Option hOption = CommandLineParser::helpOption();
-    CommandLineParser::Option mullvaldMultihop(
-        "m", "mullvald-multihop", "Generate config for Mullvad multihop.");
+    CommandLineParser::Option mullvadMultihop(
+        "m", "mullvad-multihop", "Generate config for Mullvad multihop.");
     CommandLineParser::Option wireguardMultihop(
         "w", "wireguard-multihop", "Generate config for Wireguard multihop.");
 
     QList<CommandLineParser::Option*> options;
     options.append(&hOption);
-    options.append(&mullvaldMultihop);
+    options.append(&mullvadMultihop);
     options.append(&wireguardMultihop);
 
     CommandLineParser clp;
@@ -57,8 +57,8 @@ int CommandWgConf::run(QStringList& tokens) {
     }
 
     QTextStream stream(stdout);
-    if (mullvaldMultihop.m_set && wireguardMultihop.m_set) {
-      stream << "Cannot use both --mullvald-multihop and --wireguard-multihop"
+    if (mullvadMultihop.m_set && wireguardMultihop.m_set) {
+      stream << "Cannot use both --mullvad-multihop and --wireguard-multihop"
              << Qt::endl;
       return 1;
     }
@@ -76,7 +76,7 @@ int CommandWgConf::run(QStringList& tokens) {
 
     const bool useWireguardMultihop =
 #if defined(MZ_LINUX)
-        !mullvaldMultihop.m_set;
+        !mullvadMultihop.m_set;
 #else
         wireguardMultihop.m_set;
 #endif
@@ -134,7 +134,7 @@ int CommandWgConf::run(QStringList& tokens) {
               IPAddress(exitServer.ipv6AddrIn()));
         }
       } else {
-        // configure main peer for Mullvald multihop
+        // configure main peer for Mullvad multihop
         exitPeerComment = entryPeerComment + "\n" + exitPeerComment;
         exitConfig.m_serverIpv4AddrIn = entryServer.ipv4AddrIn();
         exitConfig.m_serverIpv6AddrIn = entryServer.ipv6AddrIn();
@@ -165,7 +165,7 @@ int CommandWgConf::run(QStringList& tokens) {
                 "generate a\n";
       stream << "# configuration that works across all supported systems by "
                 "using the\n";
-      stream << "# --mullvald-multihop flag.\n\n";
+      stream << "# --mullvad-multihop flag.\n\n";
       stream << entryConfig.toMultiHopWgConf(exitConfig,
                                              QMap<QString, QString>(),
                                              exitPeerComment, entryPeerComment)


src/commands/commandservers.cpp
```diff
@@ -54,6 +54,7 @@ int CommandServers::run(QStringList& tokens) {
     }
 
     MozillaVPN vpn;
+    vpn.initializeServerData();
```
strseb: QQ: why not `vpn.serverData()->initialize();`

src/utils/interfaceconfig.cpp
```diff
@@ -113,3 +113,27 @@ QString InterfaceConfig::toWgConf(const QMap<QString, QString>& extra) const {
 
   return content;
 }
+
+QString InterfaceConfig::toMultiHopWgConf(const InterfaceConfig& exitConfig, const QMap<QString, QString>& extra) const {
```
strseb: I don't think this produces quite the wireguard config it should? Have you tried this and looked at wireshark if that really is multi hopping?

Just ran it and it gave me this: 

```
[Interface]
PrivateKey =  c: 
Address = 10.128.0.90/32, fc00:bbbb:bbbb:bb01:d::5a/128

[Peer] # Entry Peer
PublicKey = 9NuXfdBjkHkVy5IdQN+8wMNS7CiFC3n+VRdFsPzgmVM=
Endpoint = 85.203.53.104:9059
AllowedIPs = 85.203.53.140/32, 2a03:1b20:5:3::140/128

[Peer] # Exit peer
PublicKey = ZCvlPfPzOf728BIQcSmWzGFuInKK0SdVTyTCZkdrvUk=
Endpoint = 85.203.53.140:3003
AllowedIPs = 0.0.0.0/0, ::/0, 10.64.0.1/32, fc00:bbbb:bbbb:bb01::1/128, 10.124.0.0/20
```

Which looks fine-ish, but when activated, wireguard will actually just directly send traffic to the exit peer, as it's reachable from my ethernet hop, here is my routing table. 
```
=======================================================================
Interface List
 39...........................WireGuard Tunnel
  8...........................Tailscale Tunnel
  9...b4 2e 99 f2 78 a6 ......Realtek Gaming GbE Family Controller
 16...0a 62 30 31 26 88 ......Bluetooth Device (Personal Area Network)
  1...........................Software Loopback Interface 1
 19...00 15 5d ef 64 a5 ......Hyper-V Virtual Ethernet Adapter
=======================================================================´

IPv4 Route Table
===========================================================================
Active Routes:
Network Destination        Netmask          Gateway       Interface  Metric
          0.0.0.0          0.0.0.0     192.168.86.1   192.168.86.140     25
          0.0.0.0          0.0.0.0         On-link       10.128.0.90      0
        10.64.0.1  255.255.255.255         On-link       10.128.0.90      0
       10.124.0.0    255.255.240.0         On-link       10.128.0.90      0
    10.124.15.255  255.255.255.255         On-link       10.128.0.90    256
      10.128.0.90  255.255.255.255         On-link       10.128.0.90    256
    85.203.53.140  255.255.255.255         On-link       10.128.0.90      0
        127.0.0.0        255.0.0.0         On-link         127.0.0.1    331
        127.0.0.1  255.255.255.255         On-link         127.0.0.1    331
  127.255.255.255  255.255.255.255         On-link         127.0.0.1    331
      169.254.0.0      255.255.0.0         On-link    169.254.83.107    261
   169.254.83.107  255.255.255.255         On-link    169.254.83.107    261
  169.254.255.255  255.255.255.255         On-link    169.254.83.107    261
     172.25.112.0    255.255.240.0         On-link      172.25.112.1   5256
     172.25.112.1  255.255.255.255         On-link      172.25.112.1   5256
   172.25.127.255  255.255.255.255         On-link      172.25.112.1   5256
     192.168.86.0    255.255.255.0         On-link    192.168.86.140    281
   192.168.86.140  255.255.255.255         On-link    192.168.86.140    281
   192.168.86.255  255.255.255.255         On-link    192.168.86.140    281
        224.0.0.0        240.0.0.0         On-link         127.0.0.1    331
        224.0.0.0        240.0.0.0         On-link    192.168.86.140    281
        224.0.0.0        240.0.0.0         On-link      172.25.112.1   5256
  255.255.255.255  255.255.255.255         On-link         127.0.0.1    331
  255.255.255.255  255.255.255.255         On-link    192.168.86.140    281
  255.255.255.255  255.255.255.255         On-link      172.25.112.1   5256
```


On desktop we add a few custom route table routes to make sure we actually only reach exit node through the wireguard interface, i'm not sure if the classic wireguard client does the same. 


FWIW maybe this should generate a config similar to how we do it on mobile, see: 
See https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/src/controller.cpp#L464-L485
or the mullvad docs https://mullvad.net/en/help/wireguard-and-mullvad-vpn#multihop-wireguard

strseb (changes requested):
Thank you! Just 2 open questions c: otherwise it looks great!

src/utils/interfaceconfig.cpp
oskirby: To add on to this, we have two versions of mutlihop in the client.

There is "real multihop" where the VPN client encrypts the packet twice and sends the resulting packet to the exit server. In order for this to work we need some fairly precise control over the routing behavior and as such this isn't available for every platform.

The fallback when this isn't available is Mullvad's multihop, which is where we simply set the wireguard port to a special number to indicate the exit server to be used, send it to the entry server and it will redirect it to their true destination. This is used by iOS, Android and the Linux Flatpak.

encrypt94: We did a four-eyes check with @strseb  to understand why this configuration works for me but, for basti, traffic goes straight to the exit node. It appears to be a Linux - Windows issue.

The expectation is that the WireGuard client infers and applies a route from the first peer block (entry), so that all traffic destined for the second peer (exit) is sent through the VPN interface.

On Linux (Ubuntu 24.04 in my case) this works with both NetworkManager and wg-quick. Below is what my routing table looks like after setting up the VPN using this configuration (103.124.165.191 is my exit):
```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface    MSS   Window irtt
0.0.0.0         192.168.0.1     0.0.0.0         UG    600    0        0 wlp69s   0     0      0
10.64.0.1       0.0.0.0         255.255.255.255 UH    0      0        0 mozvpn   0     0      0
10.124.0.0      0.0.0.0         255.255.240.0   U     0      0        0 mozvpn   0     0      0
103.124.165.191 0.0.0.0         255.255.255.255 UH    0      0        0 mozvpn   0     0      0
192.168.0.0     0.0.0.0         255.255.255.0   U     600    0        0 wlp69s   0     0      0
```
On Windows this doesn't seem to occurr, to address that we can generate two differents config:
-  This config as "real multihop" where this setup is known to work (Linux for now; macOS still to be verified).
-  Mullvald's multihop as fallback as described below by @oskirby 

Additionaly we should:
- Provide a CLI flag allowing users to switch between the two configuration outputs.
- Generate a heavily commented WireGuard configuration explaining what type of configuration it is, where it should be used, and how to switch between modes and warning user about possible DNS leaks
- If possible, allow users to choose which registered device to generate the configuration for via a CLI flag.

src/commands/commandservers.cpp
encrypt94: emh...good question! I totally missed this getter 🤦 Happy to revert and use this instead!

src/commands/commandwgconf.cpp
```diff
@@ -27,61 +30,163 @@ CommandWgConf::~CommandWgConf() { MZ_COUNT_DTOR(CommandWgConf); }
...
     Q_ASSERT(!tokens.isEmpty());
-    if (tokens.length() > 1) {
-      QList<CommandLineParser::Option*> options;
-      return CommandLineParser::unknownOption(this, tokens[1], tokens[0],
-                                              options, false);
+    QString appName = tokens[0];
+
+    CommandLineParser::Option hOption = CommandLineParser::helpOption();
+    CommandLineParser::Option mullvaldMultihop(
+        "m", "mullvald-multihop", "Generate config for Mullvad multihop.");
```
strseb: `mullvald` Typo i think :)

encrypt94: 🤦

PR #10987, opened 2026-01-20
VPN-7389: update toggle color on theme switch

mcleinman:
## Description

This code has been stable for a while, but the bug seemingly revealed itself as part of the Qt 6.10 update. This seems to fix the issue. This connection is required due to the logic in the block immediately above it.

## Reference

VPN-7389

## Checklist
    

artfwo (approved):
Some mixed indentation levels here (2 spaces, the file has 4). Looks good otherwise!

PR #10995, opened 2026-01-22
Fix VPN-6774: VPN cannot be turned OFF after the server unavailable modal is closed

encrypt94:
## Description

   Fix VPN-6774 by adding new `StateConnectionError` to `Controller` to properly handle connection failures caused by server unavailability (e.g. switching to or connecting to an offline server).
   Server unavailable events  now  trigger a transition to this new state, which displays a dedicated error view and updates the systray with a red status indicator.
   
Screenshot from 2026-01-29 17-47-08

Also, fixes a bug  where the connection timer remain stuck after a server-unavailable error on first connection with a subsequent successful connection.

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-6774)

## Checklist
    
### DIFF
--- a/src/ui/screens/home/controller/ConnectionStability.qml
+++ b/src/ui/screens/home/controller/ConnectionStability.qml
@@ -16,11 +16,6 @@ Item {
     property real gridFlow: grid.flow
     id: stability
 
-    //% "Check Connection"
-    //: Message displayed to the user when the connection is unstable or
-    //: missing, asking them to check their connection.
-    readonly property var textCheckConnection: qsTrId("vpn.connectionStability.checkConnection")
-
     function setColumns() {
         if (!visible) {
             return;
@@ -48,7 +43,10 @@ Item {
             // TODO maybe use a different string than ModalHeaderText here
             return MZI18n.ServerUnavailableModalHeaderText;
         }
-        return textCheckConnection;
+        //% "Check Connection"
+        //: Message displayed to the user when the connection is unstable or
+        //: missing, asking them to check their connection.
+        return qsTrId("vpn.connectionStability.checkConnection");
     }
 
     GridLayout {


src/controller.cpp
```diff
@@ -323,6 +325,17 @@ void Controller::serverUnavailable() {
   logger.info() << "Server Unavailable - Ping succeeded: " << m_pingReceived;
 
   emit readyToServerUnavailable(m_pingReceived);
+  // REPORTERROR(ErrorHandler::ConnectionFailureError, "controller");
+
+  if (m_state == StateConnecting) {
+    setError(ErrorNoServerAvailable);
+    // No tunnel established yet, herw we should force network down to avoid
```
encrypt94: This is likely a corner casa but may lead to traffic leaks as no tunnel is set up, consider if we want to kill network to prevent them.
Two ways we can land here:
1. Open the VPN app -> Set servers unvailable for the current city  using the inspector -> Try to connect -> Close the popup
2. Force offline all server for a city (i.e. Tirana) using iptables -> Open the VPN app -> setup a connection to a different city -> Switch to Tirana -> Close server unavailable popup -> Switch off the vpn -> Switch on the vpn without changing server

src/controller.cpp
```diff
@@ -323,6 +325,17 @@ void Controller::serverUnavailable() {
   logger.info() << "Server Unavailable - Ping succeeded: " << m_pingReceived;
 
   emit readyToServerUnavailable(m_pingReceived);
+  // REPORTERROR(ErrorHandler::ConnectionFailureError, "controller");
```
encrypt94: Should we report a ConnectionFailureError here or the pop up it's enough?

oskirby: I don't think we need to report an error, the popup should be sufficient to notify the user.

src/controller.cpp
oskirby: nit: typo on the comment `herw` should be `here`

oskirby: I think these corner cases are acceptable as they both require some user action around the situation to get there.

The case we want to avoid is where the VPN client can autonomously get to the `StateConnectionError` (eg: because a silent server switch fails) and the traffic is unprotected.

src/ui/screens/home/controller/ControllerView.qml
```diff
@@ -338,6 +338,43 @@ Item {
...
+
+            PropertyChanges {
+                target: boxBackground
+                color: MZTheme.colors.primary
+            }
+
+            PropertyChanges {
+                target: logoTitle
+                //% "VPN is on"
+                text: qsTrId("vpn.controller.activated")
```
artfwo, line 353: Should a new string be introduced here? Otherwise we still get "VPN is on" label in the error state.

encrypt94: In the connection error view on figma we have the "VPN is on" string and it kinda makes sense as the tunnel is on but i personally agree, this could be misleading

artfwo: I'm fine with that. But is ok okay that we're checking for "VPN is off" in the test below? https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10995/changes#diff-93e987b4fbebeb6efe9bd42f991e1872a8f5717df8dedbb0364b07ee35f92a8bR202

encrypt94: Test updated to test the new error screen

src/ui/screens/home/controller/ConnectionStability.qml
```diff
@@ -16,6 +16,11 @@ Item {
     property real gridFlow: grid.flow
     id: stability
 
+    //% "Check Connection"
+    //: Message displayed to the user when the connection is unstable or
+    //: missing, asking them to check their connection.
+    readonly property var textCheckConnection: qsTrId("vpn.connectionStability.checkConnection")
```
artfwo: Nit: does it have to be a property if it's only ever used once?

encrypt94: Thanks for pointing me in the right direction with the l10n linter issue!

oskirby (approved):
Looks good!

Issue #450, opened 2021-01-12
Release .rpm files for RPM based Linux distributions


lolrepeatlol:
First, congrats on the release! 🥳

A lovely Linux distro I find myself on from time to time is Fedora. Since it's not an Ubuntu-based system and rather an RPM-based distro, it requires a different package format called an .rpm. Do you mind providing such a file?

Other popular distros that use this kind of file include Red Hat Linux, OpenSUSE, and CentOS.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-433)

j-silver:
In the meantime, I've built the client on Fedora 33. 

When I launch the daemon I get this error, which is related to some polkit and/or dbus setting specific to Fedora, and the process terminates.

```
sudo ./build/bin/mozillavpn linuxdaemon
[sudo] password for user:

[13.01.2021 10:49:32.393] (main - LogHandler) Log file: /root/.local/share/mozillavpn.txt
[13.01.2021 10:49:32.393] (main - Command) MozillaVPN 2.0.2
[13.01.2021 10:49:32.394] (main - CryptoSettings) Read the settings file
[13.01.2021 10:49:32.394] (main - SettingsHolder) Creating SettingsHolder instance
[13.01.2021 10:49:32.394] (main - Localizer) Loading language: en-GB
[13.01.2021 10:49:32.394] (main - SettingsHolder) Setting languageCode
[13.01.2021 10:49:32.396] (linux - main) Connecting to DBus...
[13.01.2021 10:49:32.397] (linux - main) Connection failed - name: org.freedesktop.DBus.Error.AccessDenied message: Request to own name refused by policy
[13.01.2021 10:49:32.405] (main - CryptoSettings) Writing the settings file
[13.01.2021 10:49:32.405] (main - CryptoSettings) Write plaintext JSON file

```

Any idea of what needs to be done?
I can open another issue if necessary.

dannycolin:
> I can open another issue if necessary.

Yes please. It's always better to limit a bug report to only one specific issue ;).

pojntfx:
Managed to get it to build & work on Fedora:

```shell
git clone git@github.com:mozilla-mobile/mozilla-vpn-client.git
cd mozilla-vpn-client/
git submodule init
git submodule update
pip install glean_parser pyhumps pyyaml
sudo dnf install -y qt qt-devel polkit-devel resolvconf
./scripts/generate_glean.py
[...]

PR #11015, opened 2026-01-28
VPN-433: Try building a static RPM

oskirby:
## Description
Now that firefox has figured out RPM signing, and is shipping RPM-based builds, we should make an effort to maintain parity. This attempts to setup an RPM build using static Qt. We chose AlmaLinux 8 as the base image on which to build (the spiritual successor to CentOS 8).

To get this working, we make the following changes:
- Switch the base image for the Qt toolchain from `debian:bullseye` to `almalinux:8` to lower the glibc minimum version to 2.28
- Enable the OpenSSL runtime backend and update to OpenSSL 3.x
- The OpenSSL change bumps the minimum support Debian release to Bookworm.
- The update to Debian Bookworm allows us to eliminate the cmake/golang/rustlang hacks and workarounds.
-  Eliminate the `LinuxCryptoSettings` support for v1 encrypted files to drop a ton of shared library dependencies.

## Reference
Github Issue: #450 
JIRA Issue: [VPN-433](https://mozilla-hub.atlassian.net/browse/VPN-433)

## Checklist
    
[VPN-433]: https://mozilla-hub.atlassian.net/browse/VPN-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oskirby:
It looks like `libsecret` has a bunch of dependencies that don't have quite as stable of an ABI as we would like, so it might be time to just drop support for the legacy `LinuxCryptoSettings` class and just switch permanently to the XDG portal instead.

taskcluster/docker/linux-qt6-build/Dockerfile
```diff
@@ -69,27 +69,11 @@ RUN apt -y install libwayland-dev \
...
-RUN mkdir -p /tmp/cmake
-RUN curl -sSL https://github.com/Kitware/CMake/releases/download/v3.26.5/cmake-3.26.5-linux-x86_64.tar.gz -o /tmp/cmake/cmake-static.tar.gz
-RUN echo "130941ae3ffe4a9ee3395514787115a273a8d1ce15cb971494bb45f7e58bb3c3 /tmp/cmake/cmake-static.tar.gz" | sha256sum -c -
-RUN tar -C /usr/local -xzf /tmp/cmake/cmake-static.tar.gz --strip-components=1
-RUN rm -rf /tmp/cmake
+# Also pull in packages from build-depends as early as possible
+# %include linux/debian/control
+ADD topsrcdir/linux/debian/control /root/mozillavpn-dpkg-control
+RUN sed -rie '/\s+(qt6-|qml6-|libqt6|qmake)/d' /root/mozillavpn-dpkg-control && \
+	mk-build-deps -i -r  --tool="apt-get --no-install-recommends --yes" /root/mozillavpn-dpkg-control
```
hneiva: nit: 2 spaces

taskcluster/docker/linux-rpm-build/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-ARG DOCKER_BASE_IMAGE=fedora:36
+ARG DOCKER_BASE_IMAGE=almalinux:8
```
hneiva, line 5: nit/suggestion: Add a comment here about the `linux-build-fedora-fc40` builds so it's a bit more obvious this is used for more than one OS.

oskirby: we could probably just drop the `linux-build-fedora-*` builds (or switch them to a latest/rolling target) after merging this PR. We don't really use the fedora builds for anything and they are horribly out of date.

linux/debian/control
```diff
@@ -47,7 +47,7 @@ Depends: wireguard (>=1.0.20200319),
          libegl1,
          libgles2,
          libpolkit-gobject-1-0,
-         libssl3,
+         libssl3 | libssl3t64,
```
artfwo: I think libssl3t64 on debian should have a "provides: libssl3" control field, but more generally, could we just let debhelper generate the share library dependencies here via dh_shlibdeps(1)?

oskirby: Good point, it does seem that `libssl3t64` provides `libssl3` so that was an unecessary addition.

I doubt that `dh_shlibdeps` would find the SSL dependency - the runtime Qt TLS backend manually calls `dlopen()` to load the library rather than sorting it out through the linker. Likewise, the `ldd` tool doesn't reveal any OpenSSL dependencies either.

artfwo: makes sense. we can still use debhelper for a bunch of other dependencies i think, but that's another topic. the PR looks good!

PR #11016, opened 2026-01-28
VPN-7457: add app exclusions help center link to toast

mcleinman:
## Description

Adding link to sumo. This is on Windows only, but perhaps we should use it for all app exclusions errors?

I tried a version where the link just said "Help", but that didn't seem as good as this version.

There is currently a conversation in slack - we may change the link. If so, I'll make a follow-on PR.

Screenshot 67

## Reference

VPN-7457

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZSystemAlert.qml
+++ b/nebula/ui/components/MZSystemAlert.qml
@@ -84,6 +84,19 @@ MZAlert {
                 }
             },
             State {
+                name: MZErrorHandler.SplitTunnelErrorAlertWithLink
+                PropertyChanges {
+                    target: alertBox
+                    alertText: MZI18n.SplittunnelErrorActivating
+                    alertActionText: MZI18n.GetHelpHelpCenter
+                    visible: true
+
+                    onActionPressed: ()=>{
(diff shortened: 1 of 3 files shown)


nebula/ui/components/MZSystemAlert.qml
```diff
@@ -83,6 +83,19 @@ MZAlert {
...
+            State {
+                name: MZErrorHandler.SplitTunnelErrorAlertWithLink
+                PropertyChanges {
+                    target: alertBox
+                    alertText: MZI18n.SplittunnelErrorActivating
+                    alertActionText: MZI18n.GetHelpHelpCenter
+                    visible: true
+
+                    onActionPressed: ()=>{
+                        MZUrlOpener.openUrlLabel("sumoExcludedApps")
```
artfwo, line 95: `sumoExcludedApps` should be registered with `registerUrlLabel` I think?

artfwo (commented):
Thanks! I have one question below, looks good otherwise!

nebula/ui/components/MZSystemAlert.qml
mcleinman: This is a link already used in the client, and so it is already registered.

PR #11020, opened 2026-01-29
Uplift VPN-7457 to release branch

mcleinman:
## Description

Uplifting #11016 

## Reference

VPN-7457

## Checklist
    
### DIFF
--- a/src/errorhandler.cpp
+++ b/src/errorhandler.cpp
@@ -49,8 +49,13 @@ QList<ErrorTypeData> s_errorData{
 
     ErrorTypeData(ErrorHandler::ControllerError, true,
                   []() { return ErrorHandler::ControllerErrorAlert; }),
+#ifdef MZ_WINDOWS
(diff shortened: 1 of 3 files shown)


src/errorhandler.cpp
```diff
@@ -49,8 +49,13 @@ QList<ErrorTypeData> s_errorData{
 
     ErrorTypeData(ErrorHandler::ControllerError, true,
                   []() { return ErrorHandler::ControllerErrorAlert; }),
+#ifdef MZ_WINDOWS
```
lesleyjanenorton, line 52: Not sure we really need this check or Windows-specific toast given the SUMO page is platform agnostic. Is there any issue with adding the SUMO link to all split tunnel error toasts?

mcleinman: I made the same comment in the initial PR (`This is on Windows only, but perhaps we should use it for all app exclusions errors?`), but it was approved there without response to that. Let's keep this for now (so it matches main), and update there if it feels warranted?

PR #11021, opened 2026-01-30
Refactor glean uploader to use Qt network library

oskirby:
## Description
The Qt/glean uploader is a VPN special creation in `qtglean/src/uploader.rs` that uses the `reqwest` library to submit telemetry pings. Seems simple enough, but this brings in an entire other TLS stack (by way of OpenSSL) to the project that complicates the build - especially when statically linking. It would be far more advantageous if we could just re-use the Qt networking libs.

So, with that goal in mind, let's write a new uploader that uses Rust-to-C FFI in order to make use of Qt. This should let us drop the `reqwest` dependency and all of its OpenSSL build pain.

## Reference
Unblocks: #11015

## Checklist
    
### DIFF
--- a/scripts/cmake/rustlang.cmake
+++ b/scripts/cmake/rustlang.cmake
@@ -464,8 +464,10 @@ function(add_rust_library TARGET_NAME)
         )
     endif()
 
-    get_filename_component(LINKER_FILENAME "${CMAKE_LINKER}" NAME_WE)
-    if(LINKER_FILENAME STREQUAL "ld" OR LINKER_FILENAME MATCHES "-ld$")
+    # Some GNU toolchains throw linker warnings, so we may need to explicitly
+    # allow duplicate symbols during linking. This can occur if we are linking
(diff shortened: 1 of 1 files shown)


oskirby:
So, this clears out a whole bunch of Rust dependencies and reduces the potential for OpenSSL linking pain, but we can't quite switch to the full `-openssl-runtime` Qt option (eg: use version-agnostic dynamic linking) because of the transition between OpenSSL version 1.1.x and 3.0

The Qt/TLS backend can target either 1.1.x or >=3.0, but not both, so switching to `-openssl-runtime` would raise our minimum supported Linux distros to:
 - Debian Bookworm
 - Ubuntu Jammy
 - RHEL 9

Note that the official Qt builds appear to use runtime OpenSSL detection compiled against OpenSSL 3.0 or later (See notes [here](https://doc.qt.io/qt-6/ssl.html#enabling-and-disabling-ssl-support-when-building-qt-from-source))

PR #11041, opened 2026-02-09
VPN-7444: Fix CLI login command

encrypt94:
## Description

Restore CLI login command functionality with the following changes
-   Refactor the `Localizer` singleton to be accessed only via the `::instance()` method to avoid double initialization errors
-   Print the authentication link to stdout instead of using `QDesktopServices::openUrl` when running in CLI mode
-   Initialize `ServerData`, check for existing instances, and run the event listener to handle deep links
-   Fix the event listener and `MozillaVPN` to handle deep links in CLI mode without attempting to show any windows
-   Add a `--headless` flag that sets a `headless=1` parameter in the auth URL, which displays the guardian success page with a code for the user to copy and paste (see related guardian PR)

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7444)

## Checklist
    
### DIFF
--- a/src/authenticationlistener.cpp
+++ b/src/authenticationlistener.cpp
@@ -44,7 +44,13 @@ AuthenticationListener* AuthenticationListener::create(
     case AuthenticationInApp:
       return new AuthenticationInAppListener(parent);
     case AuthenticationInBrowserHeadless:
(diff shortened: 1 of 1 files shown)


oskirby (approved):
Looks good!

PR #11043, opened 2026-02-09
Show multihop info in CLI `status' command

encrypt94:
## Description

Display information for entry and exit servers in the CLI `status` command when using multihop, in the following format.

```
Entry Server country code: it
Entry Server country: Italy
Entry Server city: Palermo
Exit Server country code: ar
Exit Server country: Argentina
Exit Server city: Buenos Aires
```

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/commands/commandstatus.cpp
+++ b/src/commands/commandstatus.cpp
@@ -113,14 +113,18 @@ int CommandStatus::run(QStringList& tokens) {
       stream << "Entry Server country: "
              << model->countryName(sd->entryCountryCode()) << Qt::endl;
       stream << "Entry Server city: " << sd->entryCityName() << Qt::endl;
+      stream << "Exit Server country code: " << sd->exitCountryCode()
+             << Qt::endl;
+      stream << "Exit Server country: "
+             << model->countryName(sd->exitCountryCode()) << Qt::endl;
+      stream << "Exit Server city: " << sd->exitCityName() << Qt::endl;
+    } else {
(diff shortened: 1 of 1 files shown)


src/commands/commandstatus.cpp
```diff
@@ -107,11 +107,20 @@ int CommandStatus::run(QStringList& tokens) {
...
-    stream << "Server country code: " << sd->exitCountryCode() << Qt::endl;
-    stream << "Server country: " << model->countryName(sd->exitCountryCode())
+    if (sd->multihop()) {
+      stream << "Entry Server country code: " << sd->entryCountryCode()
+             << Qt::endl;
+      stream << "Entry Server country: "
+             << model->countryName(sd->entryCountryCode()) << Qt::endl;
+      stream << "Entry Server city: " << sd->entryCityName() << Qt::endl;
+    }
+    const QString multihopExit = sd->multihop() ? "Exit " : "";
```
artfwo: Nit: The `sd->multihop()` condition is checked just above this, this whole code could be probably simplified as follows:

```
if (sd->multihop()) {
    ...
    stream << "Exit server country code: " << sd->exitCountryCode()
           << Qt::endl;
} else {
    stream << "Server country code: " << sd->exitCountryCode()
           << Qt::endl;
}
```

(With the bonus that we don't break any strings that we might want to search in the code later for debugging, but in this case it doesn't matter as much)

PR #11052, opened 2026-02-10
Nit: Allow autodetection of OSX sysroot

oskirby:
## Description
I'm setting up a new macbook for development, and despite installing Xcode and its associated SDKs, cmake throws an error due to an unset `SDKROOT` variable. We can easily work around that just by running `xcrun` to find the default MacOSX SDK.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/docs/Building/macos.md
+++ b/docs/Building/macos.md
@@ -7,47 +7,13 @@ Install Xcode - https://developer.apple.com/xcode/
 Before you start this process, open Xcode, go to settings, accounts, and sign in with your
 Apple ID. If you are a Mozilla developer, this is an Apple ID associated with your LDAP account.
 
-## Activate conda
+## Setup conda
 
 ```
 conda env create -f env-apple.yml
 conda activate vpn
 ```
 
-See [here](./index.md#conda) for conda environment instructions.
-
-Your Xcode install comes with a copy of the MacOS-SDK.
-We need to tell the conda environment where to find it.
-
-Find the sdk path
-
-```
-xcrun --sdk macosx --show-sdk-path
-```
-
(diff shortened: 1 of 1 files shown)


src/cmake/macos.cmake
```diff
@@ -73,7 +73,8 @@ elseif(CMAKE_OSX_SYSROOT)
     execute_process(OUTPUT_VARIABLE OSX_SDK_PATH OUTPUT_STRIP_TRAILING_WHITESPACE
         COMMAND xcrun --sdk ${CMAKE_OSX_SYSROOT} --show-sdk-path)
 else()
-    message(FATAL_ERROR "One of CMAKE_OSX_SYSROOT or ENV{SDKROOT} must be defined")
+    execute_process(OUTPUT_VARIABLE OSX_SDK_PATH OUTPUT_STRIP_TRAILING_WHITESPACE
```
mcleinman, line 76:
```suggestion
    execute_process(OUTPUT_VARIABLE SDKROOT OUTPUT_STRIP_TRAILING_WHITESPACE
```

Should this be SDKROOT?

mcleinman (approved):
Thanks. One potential question, other wise looks great.

Also, could you update `macos.md` instructions to remove the SDKROOT part (and maybe put the backup locations in No Shame)?

src/cmake/macos.cmake
oskirby: This isn't an environment variable, it's a CMake variable that we use to pass into the `wireguard-go` build a couple of lines down. So I think the name is fine (and probably better to keep it distinct from an environment variable of the same name).

PR #11058, opened 2026-02-11
Fix some more macos setup pain

oskirby:
## Description
Some more findings from setting up a new development mac...

The first problem occurs in getting an Apple Development signing certificate is that typically it seems to magically appear through Xcode when you open the Mozilla VPN `.xcodeproj` folder and have signed into your Apple developer account. However, creating an Xcode project fails on `main` right now due to the use of conditionally included sources in the `socksproxy` project.

Then, we continue to experience jankiness in the automated provisioning profile detection. Most notably the algorithm currently just picks the most recent profile that matches the app being signed. But this doens't necessarily select a profile that will install on your machine. So let's further limit the search to profiles that set `ProvisionsAllDevices` to `true` in order to weed out specialized testing profiles.

I should probably update the build docs to talk about how the signing certificates spring into existence while I'm in here...

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/extension/socks5proxy/src/masqueconnection.cpp
+++ b/extension/socks5proxy/src/masqueconnection.cpp
@@ -6,13 +6,13 @@
 
 #ifdef PROXY_MASQUE_ENABLED
 
-#include <QDebug>
-#include <QNetworkDatagram>
-#include <QUdpSocket>
-#include <QUrl>
-#include <QtEndian>
+#  include <QDebug>
+#  include <QNetworkDatagram>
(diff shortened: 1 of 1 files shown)


mcleinman:
Is this ready for review, or should I wait for it to be off draft?

mcleinman (commented):
Thanks. Please update anything needed in `docs/Building/macos.md` as part of this - I don't think it requires any changes, but I only gave that file a quick glance.

PR #11059, opened 2026-02-11
VPN-7475: addon messages

mcleinman:
## Description

For the two 2.34 addons. (I regret not giving the `version233.bullet11` a more generic ID, as I expect to use it in most releases.

Screenshot 70Screenshot 69Screenshot 68

## Reference

VPN-7475

## Checklist
    
### DIFF
--- a/addons/strings.yaml
+++ b/addons/strings.yaml
@@ -155,9 +155,6 @@ commonStrings:
   bullet4:
     value: "[macOS, Linux] Better experience in the rare situation where credentials are not found upon app launch."
     comment: Bullet point with a specific update in 2.34. Should have tag that indicates it applies to macOS and Linux platforms.
-  bullet5:
-    value: "[Linux] Improved installation directories for polkit."
-    comment: Bullet point with a specific update in 2.34. Should have tag that indicates it applies to Linux platform.
 
 tryFirefoxExtension:
   title:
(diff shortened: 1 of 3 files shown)


oskirby (approved):
LGTM

addons/strings.yaml
```diff
@@ -142,6 +142,23 @@ commonStrings:
...
+    value: "[all] In the rare situation where all servers in a location are unavailable, it was difficult for people to turn off the VPN. This is now easier."
+    comment: Bullet point with a specific update in 2.34. Should have tag that indicates it applies to all platforms.
+  bullet3:
+    value: "[macOS, Windows, Linux] Small improvements in the system status bar icon when switching servers."
+    comment: Bullet point with a specific update in 2.34. Should have tag that indicates it applies to macOS, Windows, and Linux platforms.
+  bullet4:
+    value: "[macOS, Linux] Better experience in the rare situation where credentials are not found upon app launch."
+    comment: Bullet point with a specific update in 2.34. Should have tag that indicates it applies to macOS and Linux platforms.
+  bullet5:
+    value: "[Linux] Improved installation directories for polkit."
```
flodolo: This needs a comment to explain what polkit is, and if it should be translated.

mcleinman: Thanks. I considered this some more, and it feels "under the hood" enough that it's just covered by the general "performance improvements" line at the top, and probably doesn't deserve its own bulletpoint. I've removed this line.

PR #11076, opened 2026-02-16
VPN-7488: Install an AppArmor profile for mozillavpn

oskirby:
## Description
It seems that without an AppArmor profile, launching the VPN binary will inherit the security context of its calling process. This can result in the situation where the VPN client is running unconfined and a browser attempts to communicate with the VPN native messaging socket, but it winds up with a different security context.

This connection is then rejected by the authorization check in `WebExtension::Server::isAllowedToConnect()`. Creating an unconfined AppArmor policy should fix the issue by ensuring a consistent security context.

In order to play nicely with various distros that might do their own weird thing, we should also be put some more guardrails around this policy check:
- We should check which security module is in use and disable this check if we don't recognize it.
- For AppArmor, disable this check if the `mozillavpn` process has no policy to prevent breakage on non-Debian distributions that use AppArmor (eg: openSUSE).
- Try to maintain parity for SELinux by checking for an `unconfined_t` type.

For some future work - we could try writing an SELinux containment policy, which would eliminate the need for all of this since we could set an appropriate security label on the UNIX socket.

## Reference
Fixes: mozilla/multi-account-containers#2852
JIRA Issue: [VPN-7488](https://mozilla-hub.atlassian.net/browse/VPN-7488)

## Checklist
    
[VPN-7488]: https://mozilla-hub.atlassian.net/browse/VPN-7488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/webextension/server_linux.cpp
+++ b/src/webextension/server_linux.cpp
@@ -48,7 +48,7 @@ bool WebExtension::Server::isAllowedToConnect(qintptr sd) {
     return true;
   }
 
-  // Fetch the peer security context. 
+  // Fetch the peer security context.
   char peersec[NAME_MAX];
   socklen_t peerlen = sizeof(peersec);
   if (getsockopt(sd, SOL_SOCKET, SO_PEERSEC, peersec, &peerlen) < 0) {


src/cmake/linux.cmake
```diff
@@ -215,3 +223,11 @@ if(NOT BUILD_FLATPAK)
 
     install(SCRIPT ${CMAKE_SOURCE_DIR}/scripts/linux/postinst.cmake)
 endif()
+
+if(BUILD_APPARMOR)
+    configure_file(${CMAKE_SOURCE_DIR}/linux/apparmor.profile.in
+        ${CMAKE_CURRENT_BINARY_DIR}/apparmor.profile)
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/apparmor.profile
```
encrypt94, line 230: nit: when building from source, this file will be installed to `/usr/local/etc/` and the native manifest to `/usr/local/lib/mozilla/native-messaging-hosts/mozillavpn.json`, both of these paths are incorrect an the files will be ignored. We should update the developer documentation and add `-DCMAKE_INSTALL_PREFIX="/"`  in the configure command.

PR #11081, opened 2026-02-18
VPN-7467 Create non-playstore release

encrypt94:
## Description
Create a release variant that does not rely on Google Play Services (GPS), allowing the app to be published on GitHub and our website so users with devices that do not support GPS can install and use the VPN.

Currently, we rely on Google Play Services only for in-app purchases. This PR updates the build scripts to generate two separate app flavors:
- playstore with IAP enabled
- website with web purchase enabled

**Note: marked as a draft as we need `AndroidAuthenticationListener` from #11036 for this to work properly**

## Reference

  [Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7467)

## Checklist
    
### DIFF
--- a/src/networkmanager.cpp
+++ b/src/networkmanager.cpp
@@ -105,7 +105,9 @@ QByteArray NetworkManager::userAgent() {
   {
     QStringList flags;
     flags.append(QString("sys:") + NetworkManager::osVersion());
-    flags.append("iap:true");
+    if (!Feature::get(Feature::Feature_webPurchase)->isSupported()) {
+      flags.append("iap:true");
+    }
 
     QTextStream out(&userAgent);
     out << Constants::NETWORK_USERAGENT_PREFIX << "/"


src/cmake/android.cmake
```diff
@@ -78,3 +78,10 @@ if( ${Qt6_VERSION} VERSION_GREATER_EQUAL 6.4.0)
         ${OPENSSL_LIBS_DIR}/libssl_3.so
     APPEND)
 endif()
+
+option(MZ_ANDROID_WEBSITE_BUILD "Build apk without google play services" OFF)
```
strseb: Just some wording thoughts, maybe Website build is a bit opague? With that enabled, we kinda don't have any propriatary dependencies. 
Maybe foss, foss-android, vanilla-android, no-gps *-build? 

Totally optional tho.

src/feature/featurelist.h
```diff
@@ -238,9 +238,10 @@ FEATURE(webPurchase,           // Feature ID
         FeatureCallback_true,  // Can be flipped off
         QStringList(),         // feature dependencies
         byPlatform({
-            .windows = true,
-            .macos = true,
-            .gnu_linux = true,
+          .windows = true, .macos = true, .gnu_linux = true,
+#if defined(MZ_ANDROID) && defined(MZ_ANDROID_WEBSITE_BUILD)
```
strseb: Nit: you only need to check for MZ_ANDROID_WEBSITE_BUILD, right? Given that implies we're building for android,

src/cmake/android.cmake
encrypt94: I don't have a strong opinion about the name so i'm open to changing. I chose this one after looking at how other foss apps names their flavors, some of them use the distribution channel for naming, even if it’s not always consistent.
Some examples:
- Signal - website, play, github 
- Nextcloud - generic, gplay, huawei 
- Element - gplay, fdroid 
- Conversations - playstore, free 
- Thunderbird (K9) - full, foss 
- OsmAnd - androidFull, gplay(Free|Full), huawei

I went with a Signal-like naming convention because...taste
Let's go with foss

PR #11082, opened 2026-02-18
VPN-7464 Add flag to disable authentication in app in CLI login command

encrypt94:
## Description

The CLI now implements its own authentication flow, independent of Feature_inAppAuthentication. We cannot rely on this feature flag to enable or disable authentication in the CLI because, on desktop, we may want AIA enabled for the CLI while disabled for the GUI.

Additionally, feature flags are loaded asynchronously. In the CLI context, we cannot depend on them since we would need to wait (somehow) for `TaskGetFeatureList` to complete before proceeding.

To avoid this, the CLI AIA flow is now enabled or disabled at compile time by a dedicated flag.  

## Reference

 [Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7464)

## Checklist
    
### DIFF
--- a/src/commands/commandlogin.h
+++ b/src/commands/commandlogin.h
@@ -5,8 +5,6 @@
 #ifndef COMMANDLOGIN_H
 #define COMMANDLOGIN_H
 
-#define MZ_CLI_AUTHENTICATION_IN_APP
-
 #include "command.h"
 
 class CommandLogin final : public Command {


src/commands/commandlogin.cpp
```diff
@@ -4,6 +4,9 @@
 
 #include "commandlogin.h"
 
+#ifdef MZ_CLI_AUTHENTICATION_IN_APP
+#  include "authenticationinapp/authenticationinapp.h"
+#endif
```
artfwo, line 9: is MZ_CLI_AUTHENTICATION_IN_APP always going to be defined?

encrypt94: Is is now (in commandlogin.h). The intent of this PR is that once we decide to drop AIA, we'll remove that #define

Issue #8740, opened 2023-12-04
Write a functional test suite for the Mozilla VPN CLI


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-5981)
┆Reporter: Beatriz Rizental Machado

PR #11096, opened 2026-02-24
VPN-5981 Add functional tests for CLI

encrypt94:
## Description

Add functional tests for the CLI also add `-t` / `--testing` flag to all CLI commands, the flag testing organization name and guardian staging server.

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7190)
#8740 

## Checklist
    
### DIFF
--- a/src/commands/commandlogin.cpp
+++ b/src/commands/commandlogin.cpp
@@ -7,7 +7,6 @@
 #if defined(MZ_CLI_AUTHENTICATION_IN_APP)
 #  include "authenticationinapp/authenticationinapp.h"
 #endif
-#include "authenticationinapp/authenticationinapp.h"
 #include "authenticationlistener.h"
 #include "commandlineparser.h"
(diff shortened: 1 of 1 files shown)


src/commands/commandlogin.cpp
```diff
@@ -4,14 +4,15 @@
 
 #include "commandlogin.h"
 
-#ifdef MZ_CLI_AUTHENTICATION_IN_APP
+#if defined(MZ_CLI_AUTHENTICATION_IN_APP)
 #  include "authenticationinapp/authenticationinapp.h"
 #endif
 #include "authenticationinapp/authenticationinapp.h"
```
artfwo: this is included unconditionally after the conditional include, is one of these unnecessary?

encrypt94: the second include should not be there, thanks for spotting

PR #11114, opened 2026-03-05
VPN-7505 Implement security guidelines for Github actions / 2

encrypt94:
## Description

Implemented security guidelines for Github actions

- Remove deprecated action seanmiddleditch/gha-setup-ninja, ninja is included by default in github actions runner
- Setup dependabot to update github actions on a weekly basis with a cooldown period of 7 days
- Explicitly set permissions to {} when no special permission in required
- Enforce review from MozillaVPN team members of any changes made to .github/ using CODEOWNERS
- Replaced the use of GITHUB_ENV with GITHUB_OUT
- Mitigate oved broad permissions by moving permissions at job level for task gh_pages and i18n_update

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7505)

## Checklist
    
### DIFF
--- a/.github/actions/qt-setup-taskcluster/action.yml
+++ b/.github/actions/qt-setup-taskcluster/action.yml
@@ -1,10 +1,6 @@
 name: Setup Qt
 description: Setup Qt from Taskcluster artifacts
 inputs:
-  dest:
-    description: 'Destination path'
-    required: true
-    default: ${{ github.workspace }}/3rdparty
   toolchain-alias:
     description: 'Taskcluster toolchain alias'
     required: false
@@ -26,6 +22,11 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set destination path
+      id: set-dest
+      shell: bash
+      run: echo "path=${GITHUB_WORKSPACE}/3rdparty" >> $GITHUB_OUTPUT
+
     - name: Resolve artifact task
       id: resolve
       shell: bash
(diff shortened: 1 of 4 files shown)


mcleinman (approved):
Thanks!

.github/workflows/gh_pages.yaml
```diff
@@ -52,6 +48,7 @@ jobs:
   logviewer:
     name: Add Logviewer
     runs-on: ubuntu-latest
+    permissions: {}
```
fkiriakos07: I think you can set `permissions: {}` on the workflow level instead of each job level to achieve the same thing but I guess it also doesn't hurt to set it on the job level

.github/workflows/macos_tests.yaml
```diff
@@ -46,7 +47,9 @@ jobs:
           ref: main
           sparse-checkout: MacOSX15.5.sdk/
           persist-credentials: false
-      - run: echo "SDKROOT=$(find ${{ github.workspace }}/3rdparty/macos-sdk -name 'SDKSettings.plist' -printf '%h\n')" >> $GITHUB_ENV
+      - name: Set SDK root path
+        id: set-sdkroot
+        run: echo "SDKROOT=$(find ${{ github.workspace }}/3rdparty/macos-sdk -name 'SDKSettings.plist' -printf '%h\n')" >> $GITHUB_OUTPUT
```
fkiriakos07: better to replace github actions expressions with env variables, this is the [reference](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) for available variables in github actions

```suggestion
        run: echo "SDKROOT=$(find ${GITHUB_WORKSPACE}/3rdparty/macos-sdk -name 'SDKSettings.plist' -printf '%h\n')" >> $GITHUB_OUTPUT
```

CODEOWNERS
```diff
@@ -2,6 +2,9 @@
 /src/translations/strings.yaml @flodolo
 /src/translations/extras @flodolo
 
+# GitHub workflows and configuration - Repository team reviews for security
+/.github/ @oskirby @lesleyjanenorton @strseb @mcleinman @artfwo @encrypt94
```
fkiriakos07, line 6: would it be better if y'all have a team to setup here to better manage access, like the `@mozilla-mobile/releng` team below

.github/actions/qt-setup-taskcluster/action.yml
```diff
@@ -65,17 +70,12 @@ runs:
         curl -sSL "${QT6_TOOLCHAIN_URL}" -o ${FILENAME}
         tar -xf ${FILENAME}
 
-    - name: Setup cmake paths
-      if: ${{ inputs.cmake-env == 'true' && runner.os != 'Windows' }}
+    - name: Get CMake prefix path
+      if: ${{ inputs.cmake-env == 'true' }}
+      id: get-cmake-path
       working-directory: ${{ inputs.dest }}/${{ steps.resolve.outputs.toolchain-path }}
```
fkiriakos07: as we discussed on slack, it would be better not to use input parameters to set the working directory and use a fixed location if possible. Input parameters can be set externally and should be assumed not trusted, there is also a risk of path traversal in addition to code injection

PR #11120, opened 2026-03-11
VPN-7459 part 2: iOS VoiceOver improvements

mcleinman:
## Description

After https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11091/, QA kicked this back as it still had issues for Messages and Settings. I've made a more general solution, but also have made it iOS-only. Most of the things I'm changing were done by a former engineer who had done a deep dive into Accessibility, and were needed for Android. I'm hesitant to mess with things that could affect other platforms' accessibility - hence the conservative iOS-only for most of this.

In addition to fixing the "tap to select bottom bar buttons" fixes, I've included a few other things that jumped out at me:
- "Sign Out" button was read twice - once for the label, once for the button
- Settings rows were read twice
- VoiceOver didn't share which bottom bar button was currently selected

## Reference

VPN-7459

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZLinkButton.qml
+++ b/nebula/ui/components/MZLinkButton.qml
@@ -138,7 +138,8 @@ MZButtonBase {
             font.family: fontName
             wrapMode: Text.WordWrap
             opacity: loaderVisible ? 0 : 1
-            Accessible.ignored: !visible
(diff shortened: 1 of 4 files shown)


oskirby (approved):
LGTM

PR #11122, opened 2026-03-12
VPN-7512: Update Linux taskcluster builds

oskirby:
## Description
Updating the Linux builds on taskcluster:
- Remove Ubuntu/Plucy (EOL) and replace it with Ubuntu/Resolute (new LTS).
- Add a taskgraph transform to follow a moving docker image tag (eg: `latest`).
- Replace all the fedora builds with the `fedora:latest` docker image.
- Disable the use of `dh_dwz` as this seems to break with golang 1.25.

## Reference
JIRA Issue: [VPN-7512](https://mozilla-hub.atlassian.net/browse/VPN-7512)
DWZ/Golang issue: [#33204](https://sourceware.org/bugzilla/show_bug.cgi?id=33204)

## Checklist
    
[VPN-7512]: https://mozilla-hub.atlassian.net/browse/VPN-7512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/linux/debian/rules
+++ b/linux/debian/rules
@@ -39,3 +39,5 @@ override_dh_installinfo:
 
 override_dh_builddeb:
 	dh_builddeb -- -Zgzip
+
+override_dh_dwz:


taskcluster/mozillavpn_taskgraph/transforms/docker_image.py
```diff
@@ -0,0 +1,38 @@
...
+        # Docker Hub API to generate the image tag including the current digest
+        dockerhub_regex = re.compile('{dockerhub:[a-z0-9:]+}')
+        for key, value in task.get("args", {}).items():
+            match = dockerhub_regex.search(value)
+            if match:
+                dh_args = match.group(0)[1:-1].split(':')
+                digest = dockerhub_fetch_digest(dh_args[1], dh_args[2])
+                if digest:
+                    tag = f'{dh_args[1]}:{dh_args[2]}@{digest}'
+                    task["args"][key] = value[0:match.start()] + tag + value[match.end():]
```
bhearsum, line 36: If I understand correctly, digests will now be included in the docker image `FROM` statements, and the digest used will be calculated at decision task time. This will end up having the effect of rebuilding docker images much more often (basically whenever the upstream images rebuild). This might be annoying at times, but I presume the idea is to catch bustage much sooner?

A possible alternative here would be to actually put the digests in the Dockerfiles and have dependabot or renovate bump them for you, which would avoid rebuilds at unfortunate times, but moves away from using a rolling image altogether.

I have no strong feelings either way, I'm mainly writing this to make sure I understand what's going on, and be clear about the tradeoffs.

PR #11153, opened 2026-03-25
Remove Qt Compat Graphic effects

strseb:
## Description
My build with Qt 6.11 fails to create the QML window, due to an error in Qt5 GraphicEffects Compat module. 
it might just be a blip with my build, but i thought thats a good opportunity to replace the old compat module with Qt6 QML Effects.

Note given i'm using RectangularShadow - so we now have MINIMUM reqiurement of Qt 6.9. If we want to avoid that, i probably could also use the multi-effect element, just like the others.


mcleinman:
> This is awesome, I would love to be rid of the Qt5 compat module, but I don't think we can merge this until we drop support for our PPA builds as those are holding us back to Qt 6.2.4 (Jammy) and Qt 6.4.2 (Noble).

@strseb with this, do you want a review now, or should we hold this for a bit?

oskirby:
I think once 2.37.0 is released, we should remove the PPA builds from our CI jobs, which should allow us to move forward with this patch.

mcleinman:
> I think once 2.37.0 is released, we should remove the PPA builds from our CI jobs, which should allow us to move forward with this patch.

PPA builds are now removed from CI.

nebula/ui/compat/CMakeLists.txt
```diff
@@ -4,15 +4,9 @@
...
-    URI compat 
+    URI compat
     STATIC
     RESOURCE_PREFIX /nebula
     QML_FILES
-        qt6/MZColorOverlay.qml
-        qt6/MZLinearGradient.qml
-        qt6/MZRadialGradient.qml
-        qt6/MZRectangularGlow.qml
         qt6/MZAnimatedRingsShader.qml
```
oskirby: If the `MZAnimatedRingsShader` is the only thing left, it doesn't really make sense for this to be its own module anymore. IIUC, this isn't really here as a compatibility QML type anyways it just wound up here as a shader-related thing.

src/ui/qt6winhack.qml
```diff
@@ -4,8 +4,6 @@
 
 // See https://mozilla-hub.atlassian.net/browse/VPN-1894
 
-import Qt5Compat.GraphicalEffects
-import QtGraphicalEffects 1.0
 import Qt.labs.qmlmodels 1.0
```
oskirby: It's probably worth asking the question: Do we still need this file?

oskirby (approved):
Looks good - only some non-blocking comments.

oskirby:
There is also the qt5compat module in `linux/flatpak/org.mozilla.vpn.yml` that can be removed.

PR #11169, opened 2026-03-31
VPN-7203 - Add a warning when not running with "unrestricted" background usage permission

strseb:
This adds a warning if the User is running the App not in "unrestricted" background usage mode. - Clicking it will force a system prompt. We will show this warning only until interacted with it once, so users can dismis it. 

https://github.com/user-attachments/assets/93bda8aa-0549-47ba-ab12-c9c95be9b85b

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -360,8 +360,11 @@ settings:
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
   batteryOptimizationWarning:
-    value: For the best experience, tap here to ensure reliable app functionality by disabling battery optimization.
+    value: Allow Mozilla VPN to run in background to ensure VPN connection is maintained
     comment: Warning shown in the settings screen on Android when battery optimization is enabled for the app.
+  batteryOptimizationWarningButton:
+    value: Allow background permissions
+    comment: Button label for the call to action to allow background permissions on Android.
   appExclusionClearAllApps:
     value: Clear all
     comment: Button label to clear all apps from the exclusions list
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -359,6 +359,9 @@ settings:
   dnsSettingsWarning: These options may cause some websites to break or not display content correctly.
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
+  batteryOptimizationWarning:
+    value: For the best experience, tap here to ensure reliable app functionality by disabling battery optimization.
+    comment: Warning shown in the settings screen on Android when battery optimization is enabled for the app.
   appExclusionClearAllApps:
     value: Clear all
     comment: Button label to clear all apps from the exclusions list
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -360,7 +360,7 @@ settings:
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
   batteryOptimizationWarning:
-    value: Allow Mozilla VPN to run in background to ensure VPN connection is maintained
+    value: Allow Mozilla VPN to run in background to ensure the VPN connection is maintained
     comment: Warning shown in the settings screen on Android when battery optimization is enabled for the app.
   batteryOptimizationWarningButton:
     value: Allow background permissions
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -360,7 +360,7 @@ settings:
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
   batteryOptimizationWarning:
-    value: On some devices, Mozilla VPN may be disconnected in the background. If you've experienced this, allow it to run in the background.
+    value: On some devices, Mozilla VPN may be disconnected in the background. If you’ve experienced this, allow it to run in the background.
     comment: Warning shown in the settings screen on Android when battery optimization is enabled for the app.
   batteryOptimizationWarningButton:
     value: Allow background permissions
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -360,7 +360,7 @@ settings:
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
   batteryOptimizationWarning:
-    value: Allow Mozilla VPN to run in background to ensure the VPN connection is maintained
+    value: On some devices, Mozilla VPN may be disconnected in the background. If you've experienced this, allow it to run in the background.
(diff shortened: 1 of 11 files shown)


flodolo (approved):

mcleinman (changes requested):
I really don't like using the `MZInformationCard` with a mousearea for this, from a design perspective. Even with `tap here` I'm not sure it's discoverable, and I'm of the opinion we should keep the info cards read-only. If you want, I'm happy to build an optional button in the info card. Want me to work on that real quick?

If we went that route, I'd suggest the card say `Allow Mozilla VPN to run in background to ensure VPN connection is maintained.` combined with `Allow background permissions` for the button. Also, if this is something that will probably happen, I'd suggest we make this an error info card, or something even stronger.

mcleinman:
> > If you want, I'm happy to build an optional button in the info card. Want me to work on that real quick?
> 
> I will never refuse an offer to help me on UI stuff! Please!

@strseb PR is up: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11173

src/translations/strings.yaml
```diff
@@ -359,6 +359,12 @@ settings:
   dnsSettingsWarning: These options may cause some websites to break or not display content correctly.
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
+  batteryOptimizationWarning:
+    value: Allow Mozilla VPN to run in background to ensure VPN connection is maintained
```
```suggestion
    value: Allow Mozilla VPN to run in background to ensure the VPN connection is maintained
```

Sounds weird without an article?

flodolo (approved):

src/ui/screens/settings/ViewSettingsMenu.qml
```diff
@@ -24,6 +24,31 @@ MZViewBase {
...
+            _infoContent: MZTextBlock {
+                id: batteryWarningText
+                Layout.fillWidth: true
+                text: MZI18n.SettingsBatteryOptimizationWarning
+                //verticalAlignment: Text.AlignVCenter
+            }
+            _buttonAction: function() {
+                    MZSettings.hasDismissedBatteryOptimization = true
+                    BatteryOptimizer.triggerBatteryOptimizationIntent()
+            }
```
artfwo, line 46: once we set `hasDismissedBatteryOptimization` to true, the warning won't reappear, right? (e.g. if the user dismissed the system dialog)

artfwo (approved):

src/ui/screens/settings/ViewSettingsMenu.qml
strseb: That is the idea :) - The intention is to give a hint, not to force them to enable this.

mcleinman:
Why do we want to make this dismissable? If we're pretty confident this is going to hurt the VPN experience, why not leave it there? (This also covers the case where someone taps the button, goes to system page, gets a text or something and goes to deal with it, but then forgets to come back to giving the system permission.)

src/ui/screens/settings/ViewSettingsMenu.qml
```diff
@@ -24,6 +24,31 @@ MZViewBase {
...
+            Layout.preferredWidth: Math.min(window.width - MZTheme.theme.windowMargin * 2,
+                                            MZTheme.theme.navBarMaxWidth)
+            Layout.minimumHeight: batteryWarningText.height + MZTheme.theme.windowMargin * 2
+            Layout.alignment: Qt.AlignHCenter
+
+            _infoContent: MZTextBlock {
+                id: batteryWarningText
+                Layout.fillWidth: true
+                text: MZI18n.SettingsBatteryOptimizationWarning
+                //verticalAlignment: Text.AlignVCenter
```
mcleinman, line 42: remove?

mcleinman (approved):
Approved, but with the prior comment that we may not want to make this dismissable. That said, won't block on that.

strseb:
>Why do we want to make this dismissable? If we're pretty confident this is going to hurt the VPN experience, why not leave it there? (This also covers the case where someone taps the button, goes to system page, gets a text or something and goes to deal with it, but then forgets to come back to giving the system permission.)

My rationale would be that this actually can hurt your battery lifetime,(i.e keep the deamon alive longer when not connected) and depending on the manufaturer it might not even be needed. I.e my google pixel i never even had to do this and i have weeks of background connectivity.  
Even after dismissal people can go to the apps permissions and enable/disable this as needed, i dont think we need to build a 2nd permanent ui for that.  
So i thought of this as maybe more as a hint of "hey this exists"

mcleinman:
> > Why do we want to make this dismissable? If we're pretty confident this is going to hurt the VPN experience, why not leave it there? (This also covers the case where someone taps the button, goes to system page, gets a text or something and goes to deal with it, but then forgets to come back to giving the system permission.)
> 
> My rationale would be that this actually can hurt your battery lifetime,(i.e keep the deamon alive longer when not connected) and depending on the manufaturer it might not even be needed. I.e my google pixel i never even had to do this and i have weeks of background connectivity. Even after dismissal people can go to the apps permissions and enable/disable this as needed, i dont think we need to build a 2nd permanent ui for that. So i thought of this as maybe more as a hint of "hey this exists"

So I understand - the issue is that sometimes this causes the VPN to be closed in the background, but sometimes it does nothing harmful at all? And we're trying to find messaging that lets folks know that? If so, perhaps the message should be adjusted to "...in some situations, this causes the VPN to be disconnected. If this is something you've experienced, consider changing this setting..."? (There isn't a way we can track if the daemon has been closed by the OS, right?)

strseb:
>"...in some situations, this causes the VPN to be disconnected. If this is something you've experienced, consider changing this setting..."?

Wouldn't this be too long for the text box? :sweat_smile: 

>There isn't a way we can track if the daemon has been closed by the OS, right?

There is [ApplicationExitInfo](https://developer.android.com/reference/android/app/ApplicationExitInfo) - theoretically we could check whether the last exit was not clean and take that as a signal, but then again we'd need to be trusting the manufacturer works per-spec (and does not just simulate a swipe-kill)  - where if it would be this would all not be needed as we have the `android:foregroundServiceType="systemExempted"` flag on the daemon.

mcleinman:
> > "...in some situations, this causes the VPN to be disconnected. If this is something you've experienced, consider changing this setting..."?
> 
> Wouldn't this be too long for the text box? 😅

Doesn't look like it (that ellipses at the end was how I wrote it):
Screenshot 92

 
> > There isn't a way we can track if the daemon has been closed by the OS, right?
> 
> There is [ApplicationExitInfo](https://developer.android.com/reference/android/app/ApplicationExitInfo) - theoretically we could check whether the last exit was not clean and take that as a signal, but then again we'd need to be trusting the manufacturer works per-spec (and does not just simulate a swipe-kill) - where if it would be this would all not be needed as we have the `android:foregroundServiceType="systemExempted"` flag on the daemon.

That seems not worth it. But maybe a slightly longer explanation?

src/translations/strings.yaml
```diff
@@ -359,6 +359,12 @@ settings:
   dnsSettingsWarning: These options may cause some websites to break or not display content correctly.
   dnsSettingsInputPlaceholder: Enter custom DNS
   dnsSettingsDisconnectWarning: Changing these settings will cause the VPN to disconnect and reconnect automatically.
+  batteryOptimizationWarning:
+    value: On some devices, Mozilla VPN may be disconnected in the background. If you've experienced this, allow it to run in the background.
```
```suggestion
    value: On some devices, Mozilla VPN may be disconnected in the background. If you’ve experienced this, allow it to run in the background.
```

flodolo (approved):

mcleinman (approved):
Thanks!

PR #11188, opened 2026-04-08
Support IPv6-only networks on Windows

artfwo:
## Description

Support IPv6-only networks on Windows, basically mirroring the Linux implementation from #11171 .

One notable addition is a fix for latency measurement on IPv6-only network: hasIPv4Connectivity() would return true for APIPA addresses (169.254.x.x) on adapters with no IPv5 routing, causing ServerLatency to attempt IPv4 pings anyway.

## Reference

[VPN-4368](https://mozilla-hub.atlassian.net/browse/VPN-4368)

## Checklist
    
[VPN-4368]: https://mozilla-hub.atlassian.net/browse/VPN-4368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/windows/daemon/wireguardutilswindows.cpp
+++ b/src/platforms/windows/daemon/wireguardutilswindows.cpp
@@ -368,6 +368,21 @@ bool WireguardUtilsWindows::deleteInterface() {
   return true;
 }
 
+static bool hasRouteToIPv4(const QString& ipv4addr) {
+  SOCKADDR_INET dest = {};
+  dest.Ipv4.sin_family = AF_INET;
+
+  if (InetPtonA(AF_INET, qPrintable(ipv4addr), &dest.Ipv4.sin_addr) != 1) {
+    return true;
+  }
+
+  MIB_IPFORWARD_ROW2 bestRoute = {};
+  SOCKADDR_INET bestSource = {};
+  DWORD result =
+      GetBestRoute2(nullptr, 0, nullptr, &dest, 0, &bestRoute, &bestSource);
+  return result == NO_ERROR;
+}
+
 bool WireguardUtilsWindows::updatePeer(const InterfaceConfig& config) {
   // Enable the windows firewall for this peer.
   if (!m_firewall->enablePeerTraffic(config)) {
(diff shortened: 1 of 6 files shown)


src/platforms/windows/daemon/wireguardutilswindows.cpp
```diff
@@ -368,6 +368,21 @@ bool WireguardUtilsWindows::deleteInterface() {
   return true;
 }
 
+static bool hasRouteToIPv4(const QString& ipv4addr) {
```
oskirby: I suspect this test will cause problems during server switching, as we might create a an IPv4 route to this destination through the tunnel, and we are really trying to answer the question "is this destination reachable outside of the VPN"?

A quick and dirty check might be to look at `m_routeMonitor->m_interfaceMetricsIpv4.isEmpty()` to see if any non-VPN interfaces have a valid IPv4 interface metric. The most correct solution would be to fetch the routing table with `GetIpForwardTable2(AF_INET, ...);` and iterate through it while skipping the VPN interface... similar to how the `WindowsRouteMonitor` refreshes the captured routes.

oskirby: Possibly something like this:

```C++
// windowsroutemonitor.cpp
quint64 WindowsRouteMonitor::getExclusionRouteLuid(const IPAddress& prefix) {
    MIB_IPFORWARD_ROW2* row = m_exclusionRoutes.value(prefix, nullptr);
    return row ? row->InterfaceLuid.Value : 0;
}

// wireguardutilswindows.cpp
WireguardUtilsWindows::updatePeer() {
  bool hasIpv4 = true;

  // Exclude the server address, except for multihop exit servers.
  if (m_routeMonitor && (config.m_hopType != InterfaceConfig::MultiHopExit)) {
    m_routeMonitor->addExclusionRoute(IPAddress(config.m_serverIpv4AddrIn));
    if (!config.m_serverIpv6AddrIn.isEmpty()) {
      m_routeMonitor->addExclusionRoute(IPAddress(config.m_serverIpv6AddrIn));
    }
    hasIpv4 = m_routeMonitor->getExclusionRouteLuid(IPAddress(config.m_serverIpv4AddrIn)) != 0;
  }
}
```

The idea being that we add the exclusion route, then determine if we have a valid route by checking if the route monitor found an interface to bind it to.

oskirby (changes requested):
I think a better solution for `hasRouteToIPv4()` needs to be found to ensure that we don't mistake an IPv4 route into the tunnel as valid during silent server switching.

oskirby (approved):
Looks good!

Issue #5259, opened 2022-12-19
[Feature Request] Linux ARM Build


R8s6:
Hi,

I understand ARM build is currently [not supported](https://support.cdn.mozilla.net/sq/kb/mozilla-vpn-compatible-my-device), but since Android builds are available, there's apparently no/few intrinsic technical obstacles to build it on ARM Linux distribution.

Raspberry Pis and Chromebooks are popular ARM linux computers with millions of users, so it'd be awesome if there's an ARM build for GNU Linux.

Alternatively, since Mozilla VPN is running on Wireguard, if we can find/export the wireguard config after login to our user account page, it would be easier to set up on an ARM linux running Wireguard.

Yes, what one can do now is run Mozilla VPN on an x86 machine first and use wireguard's built-in commands to find out the configs, then use them on an ARM machine, but it takes a few extra steps.

Thank you very much!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-3520)

valmynd:
I'd also love to have a way to use Mozilla VPN on ARM Linux.

There is a port to Mac OS - would be great to be able to use it on Ubuntu Asahi as well.

joshgiesbrecht:
Same. Would very much like support for my portable RPi device!

PR #11196, opened 2026-04-09
VPN-3520: linux aarch64 static deb build

encrypt94:
## Description

Add deb packages with statically linked qt for aarch64.
Tested on my aarch64 headless machine running Ubuntu and on @artfwo aarch64 debian laptop.

Changes introduced in this PR:
-  annotated tools and python libs as `:native` in `linux/debian/control`
-  inject cross build params in  `linux/debian/rules`
-  `linux_build_dpkg.sh` add `--cross-arch` flag to setup cross compilation environment
-  `scripts/utils/qt6_compile.sh` add `--cross-arch`, `--host-path`, `--toolchain-file` flags to setup the cross compilation
-  `compile_qt_6_linux.sh` add `--cross-arch` flag to install aarch64 dependencies and cross compile the package
-  `mk-build-deps` added to the sudoers allowlist in the existing `linux-dpkg-build` and `linux-qt6-build` images

Taskcluster tasks:

- `qt-linux-aarch64` toolchain task -  runs `compile_qt_6_linux.sh --cross-arch aarch64` inside the new `linux-aarch64-deb-build image`  produces `public/build/qt6_linux_aarch64.tar.xz`
- `linux-aarch64/release-deb` build task - fetches both qt-linux-aarch64 and qt-tools-6.10, runs `linux_build_dpkg.sh --cross-arch aarch64`                                                      
- add `linux-aarch64-deb-build`  aarch64 enabled build image based on debian bookworm

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-3520)
#5259 

## Checklist
    

strseb (approved):
The taskcluster parts lgtm

ahal (approved):
I'm not really sure that releng needs to be codeowners here anymore, but task changes look reasonable to me.

The one suggestion I'll make as a potential follow-up, you could probably reduce a lot of duplication in your kind.ymls by using the "matrix" transforms:
https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/matrix.html

If you've ever used matrix with Github actions, it's very similar.

taskcluster/docker/linux-aarch64-deb-build/Dockerfile
```diff
@@ -0,0 +1,122 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Docker image for cross-compiling the Mozilla VPN DEB packages for aarch64.
+# Uses Debian Bookworm for best arm64 multiarch and dpkg cross-build support.
+
+ARG DOCKER_BASE_IMAGE=debian:bookworm
```
oskirby: To avoid duplication of our Dockerfiles, can we merge this with `taskcluster/docker/linux-dpkg-build/Dockerfile` and specify the build architecture through an argument?

For example:

```dockerfile
ARG DEB_BUILD_ARCH=amd64
ENV DEB_BUILD_ARCH=$DEB_BUILD_ARCH

# Convenience to set all of the DEB_BUILD_xxx variables
RUN export $(dpkg-architecture)

RUN apt-get -y install --no-install-recommends \
    crossbuild-essential-${DEB_BUILD_ARCH} \
    gcc-${DEB_BUILD_GNU_TYPE} \
    g++-${DEB_BUILD_GNU_TYPE} \
    binutils-${DEB_BUILD_GNU_TYPE} \
    ...

RUN apt install -y libdbus-1-dev:${DEB_BUILD_ARCH} ...
```

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -13,91 +13,140 @@ helpFunction() {
...
+  if [[ ! -d "${QT_HOST_PATH}" ]]; then
+    echo "qt-host-tools not found at ${QT_HOST_PATH}" >&2
+    exit 1
+  fi
+
+  # Put Qt host tools on PATH so cross-build CMake steps can find moc/rcc
+  export PATH="${QT_HOST_PATH}/bin:${PATH}"
+
+  # Rust cross-compilation
+  export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="${CROSS_ARCH}-linux-gnu-gcc"
```
oskirby: It would be better to do this by adding an appropriate call to `__rust_build_toolchain_config()` for this case in `scripts/rustlang.cmake`

Something like:
```cmake
# ... all of the other platform-specific cases...
elseif(CMAKE_CROSSCOMPILING AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
    __rust_build_toolchain_config(
        FILENAME ${CMAKE_BINARY_DIR}/cargo_home/config.toml
        ARCH ${CMAKE_SYSTEM_PROCESSOR}-linux-gnu-gcc
        RUSTFLAGS "-Clinker=${CMAKE_C_COMPILER}"
    )
else()
    __rust_build_toolchain_config(FILENAME ${CMAKE_BINARY_DIR}/cargo_home/config.toml)
endif()
```

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -13,91 +13,140 @@ helpFunction() {
...
+  export CARGO_BUILD_TARGET="${CROSS_ARCH}-unknown-linux-gnu"
+  if command -v rustup > /dev/null 2>&1; then
+    rustup target add "${CROSS_ARCH}-unknown-linux-gnu" 2>/dev/null || true
```
oskirby: `rustup target add` is potentially an expensive operation, that might fetch gigabytes of stuff from the network. This should go in the dockerfile so that it can be cached.

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -13,91 +13,140 @@ helpFunction() {
...
+  fi
+
+  # Go cross-compilation
+  export GOARCH="${CROSS_DEB_ARCH}"
```
oskirby: This would be better done by adding checks to `build_go_archive()` in `scripts/cmake/golang.cmake` to choose a better default for `GOARCH` when cross compiling. Otherwise, I think that function already sets every other variable correctly.

I think somehthing like the following should do the trick:
```
if (NOT GOBUILD_GOARCH)
    if (CMAKE_CROSSCOMPILING)
        set(GOBUILD_GOARCH ${CMAKE_SYSTEM_PROCESSOR})
    else()
        execute_process(OUTPUT_VARIABLE GOBUILD_GOARCH OUTPUT_STRIP_TRAILING_WHITESPACE COMMAND ${GOLANG_BUILD_TOOL} env GOARCH)
    endif()
endif()
```

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -13,91 +13,140 @@ helpFunction() {
...
-  echo "  -d, --dist DIST     Build packages for distribution DIST (defaut: ${VERSION_CODENAME})"
-  echo "  -s, --static        Build packages for statically linked Qt."
-  echo "  -h, --help          Display this message and exit"
+  echo "  -d, --dist DIST         Build packages for distribution DIST"
+  echo "  -s, --static            Build with statically linked Qt (x86_64)"
+  echo "  --cross-arch ARCH       Cross-compile for ARCH (supported: aarch64)"
+  echo "  -h, --help              Display this message and exit"
 }
 
+CROSS_ARCH=""
```
oskirby: I would recommend reusing the `DEB_BUILD_ARCH` and `DEB_BUILD_GNU_CPU` for these variables to match up with the standard cross-compilation variables we get from `dpkg-architecture`.

Then cross compilation can be as simple as:
```
   export DEB_BUILD_ARCH=arm64
   ./linux_build_dpkg.sh -s
```

linux/debian/rules
```diff
@@ -3,6 +3,7 @@
 export DH_VERBOSE=1
 
 DEB_HOST_MULTIARCH ?= $(shell dpkg-architecture -qDEB_HOST_MULTIARCH)
+DEB_HOST_ARCH ?= $(shell dpkg-architecture -qDEB_HOST_ARCH)
```
oskirby, line 6: ~~For correctness, these should be `DEB_BUILD_ARCH` and `DEB_BUILD_MULTIARCH`~~

~~In the language of Linux cross compilation, the terms are:~~
- ~~Host: The system on which the command is running.~~
- ~~Build: The system on which the binary is expected to run.~~
- ~~Target: The system for which the binary will generate code for (only really used for compilers).~~

oskirby (commented):
I have a bunch of comments that I think can simplify this a bit and align things with typical Debian cross-compilation environments (eg: rely on `DEB_BUILD_ARCH` more).

There is nothing particularly spicy in the taskcluster changes here, so I don't think we need to block on releng to merge this.

linux/debian/rules
encrypt94: Yeah, i agree! however in the somewhat unintuitive Debian/GNU cross-compilation terminology they are defined the opposite way:
- Host: The machine the package is built for.
- Build: The machine the package is built on. 

<https://manpages.debian.org/testing/dpkg-dev/dpkg-architecture.1.en.html#TERMS>
<https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html>

oskirby: Oh! You are correct and I am the one who got that backwards!
I even looked at that manpage and somehow still got it wrong.

taskcluster/docker/linux-qt6-build/Dockerfile
```diff
@@ -4,6 +4,10 @@
 
 FROM debian:bookworm
 
+ARG DEB_HOST_ARCH=amd64
+ARG DEB_HOST_GNU_CPU=x86_64
+ARG RUST_VERSION=1.85.0
```
oskirby: To be totally pedantic, `mk-build-deps` will install a rust toolchain from the available debian packages, of which the `linux/debian/control` file only specifies the minimum version (not an exact version). Rather than passing a variable here that can eventually go out of sync, it would be more correct to get the toolchain version using a shell command (eg: `rustc --version | awk '{print $2}'`)

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -13,91 +13,127 @@ helpFunction() {
...
 done
 
+# Validate cross-arch and resolve the Debian architecture name
+if [[ -n "$DEB_HOST_ARCH" ]]; then
+  case "$DEB_HOST_ARCH" in
+    arm64) DEB_HOST_GNU_CPU="aarch64" ;;
+    *) echo "ERROR: unsupported --deb-host-arch value: ${DEB_HOST_ARCH}" >&2; exit 1 ;;
+  esac
+  STATICQT=Y
+  DIST="${DIST:-static-${DEB_HOST_ARCH}}"
```
oskirby: I would undo the changes to the `DIST` variable as this seems to produce a bit of a mess for the filename. Looking at the build jobs that ran on this PR, the produced `.deb` files wind up with names like: `mozillavpn-dbgsym_2.38.0~pr11196-static-arm641_arm64.deb`

We don't need to inject the architecture here as `dpkg-buildpackage` already includes it in the canonical filename of `<package>_<version>_<arch>.deb`

taskcluster/scripts/toolchain/compile_qt_6_linux.sh
```diff
@@ -10,14 +10,43 @@ export XZ_DEFAULTS="-T0"
 
 QT_SOURCE_DIR=$(find $MOZ_FETCHES_DIR -maxdepth 1 -type d -name 'qt-everywhere-src-*' | head -1)
 QT_SOURCE_VERSION=$(echo $QT_SOURCE_DIR | awk -F"-" '{print $NF}')
+ARTIFACT_NAME="qt6_linux.tar.xz"
+CROSS_ARCH=""
+DEB_HOST_ARCH="x86_64"
```
oskirby: I think this is supposed to be `DEB_HOST_ARCH=amd64` as the default?

oskirby (approved):
I think this looks good enough to merge, just a couple of minor nits at this point.

taskcluster/scripts/build/linux_build_dpkg.sh
encrypt94: You're right! QQ is there a particular reason why we append a `1` to  DPKG_PACKAGE_DIST_VERSION [here](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/09e72eae7e577c39bb9116152380483909b5ea0e/taskcluster/scripts/build/linux_build_dpkg.sh#L74) or is it just a longstanding typo?

taskcluster/docker/linux-qt6-build/Dockerfile
encrypt94: I like this!

taskcluster/scripts/build/linux_build_dpkg.sh
oskirby: Debian packaging is very strict about version numbers being unique, and this can cause problems if we need to re-run a build job for one reason or another. So, for that reason we append a revision number to the end of the `${DIST}` chunk of the version.

For example, in the PPA automation workflow we set it to `${GITHUB_RUN_ATTEMPT}`.

Issue #843, opened 2021-04-13
Mac: Support Split tunnel


strseb:
This is possible in theory, apparently:
-> If we're a mac store app, we can provide a configuration profile ourselves
-> Otherwise we could need similar to windows run our own kext... like express :) 

A gpl3 ref implementation for such a kext is here : 
https://github.com/pia-foss/desktop-mac-network-kext

Issue #522, opened 2021-01-28
Offer split tunneling or app permissions feature on all Mozilla VPN platforms


chrismore:
Currently, the split tunneling feature only exists on the Android platform, which is called "app permissions" in the settings. This is a common feature for VPNs and it should be available across all platforms we offer today.

We do have issue #416 already fixed on Android, which defines an allow-list as Android has many "system apps" that users would likely never interact with. Some of those system apps are common Google apps like Gmail. The allow-list defines, which system apps we should display in the UI.

birdsarah:
Do we want this to be informed by the anti-profile linking discussions?

Cimbali:
Linux is really only partially supported, see #1869

Cimbali:
In fact, while the apps permission interface works on Linux, the actual split-tunnelling seems to not work at all on Ubuntu 20.04 and other Linux platforms, see #6394

PR #11199, opened 2026-04-10
VPN-7095: Split tunnelling support for macOS

oskirby:
## Description
Split tunneling for macOS is possible using the `NETransparentProxyProvider` network extension class. The gist of how it works is that we can intercept packet flows and make use of the packet flow metadata to determine the source application. Once a flow has been identified for split tunneling, it can then be bound to a specific interface for routing outside of the VPN.

To make this work, we need to ensure that the default route gets cloned with `RTF_IFSCOPE` set - this ensures that packets which get bound to the default route interface can still get handled correctly. As such, we wind up re-impementing the `MacOSRouteMonitor` class inside the network extension.

This work is hidden behind a feature flag for now, and it will take some followup work to make it viable for user consumption. Most notably:
- The extension requires an updated signing configuration with new entitlements and a provisioning profile.
- Installation of the extension throws a lot of permissions screens, which makes the user experience kind of awful.
- It would be a lot cleaner if we got rid of the daemon and used a `NEPacketTunnelProvider` extension - then we could kill two permission-screen birds with one stone.

It should also be called out that this removes the `splitTunnel` feature flag, and we replace it with a runtime determination of split tunneling support via a new `Controller::splitTunnelSupported()` API. This cleans up the platform logic to determine when split tunneling is available, but as a possible downside we lose the ability to disable split tunneling. I could put the feature back if we feel it's necessary, but it also seems like a bit of a potential footgun...  not sure on that.

## Reference
Github issues:
 - #843
 - #522

JIRA Issues:
  - [VPN-7108](https://mozilla-hub.atlassian.net/browse/VPN-7108): Implement app image and app list provider classes for macOS
  - [VPN-7100](https://mozilla-hub.atlassian.net/browse/VPN-7100): Implement config channel to exclude macOS apps by bundle/signing identifer
  - [VPN-7099](https://mozilla-hub.atlassian.net/browse/VPN-7099): Handle UDP packet flows in macOS split-tunnel extension
  - [VPN-7098](https://mozilla-hub.atlassian.net/browse/VPN-7098): Handle TCP packet flows in macOS split-tunnel extension
  - [VPN-7097](https://mozilla-hub.atlassian.net/browse/VPN-7097): Install and manage macOS split-tunnel extension

## Checklist
    
[VPN-7108]: https://mozilla-hub.atlassian.net/browse/VPN-7108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taskcluster/scripts/signing/entitlements.xml
```diff
@@ -1,27 +1,34 @@
...
-	<key>com.apple.security.network.client</key>
-	<true/>
-	<key>com.apple.security.network.server</key>
-	<true/>
+         </array>
+         <key>com.apple.security.network.client</key>
+         <true/>
+         <key>com.apple.security.network.server</key>
+         <true/>
+         </dict>
```
artfwo: this looks a bit out of place, a typo? there's a 2nd closing </dict> tag below, but only one opening <dict>

macos/networkextension/VPNSplitTunnelProvider.mm
```diff
@@ -0,0 +1,496 @@
...
+  }
+  self.settings.includedNetworkRules = @[includeAllRule];
+
+  auto excludeRules = [[NSMutableArray<NENetworkRule*> new] init];
+
+  // Exclude LAN traffic
+  [excludeRules addObject:[VPNSplitTunnelProvider matchRoute:@"10.0.0.0" andPrefix:8]];
+  [excludeRules addObject:[VPNSplitTunnelProvider matchRoute:@"172.16.0.0" andPrefix:12]];
+  [excludeRules addObject:[VPNSplitTunnelProvider matchRoute:@"192.168.0.0" andPrefix:16]];
+  [excludeRules addObject:[VPNSplitTunnelProvider matchRoute:@"224.0.0.0" andPrefix:4]];
```
artfwo, line 194: I think link-local traffic (169.254.x.x/16) should be excluded here too. What about ipv6 traffic?

src/platforms/macos/macoscontroller.mm
```diff
@@ -230,10 +258,60 @@ emit initialized(true, jsObj.value("connected").toBool(),
...
+  BOOL okay = [session startTunnelWithOptions:options andReturnError:&error];
+  if (error) {
+    logger.warning() << "proxy start error:" << error.localizedDescription;
+  } else if (!okay) {
+    logger.warning() << "proxy start failed";
+  } else {
+    // Save the session and retain it.
+    [session retain];
+    m_session = session;
+  }
```
artfwo, line 301: if `startTunnelWithOptions` fails, is the error merely logged, or displayed in the UI too?

src/platforms/macos/macosextensionloader.mm
```diff
@@ -0,0 +1,110 @@
...
+        logger.debug() << "proxy prefs setup failed:"
+                       << saveErr.localizedDescription;
+      }
+      [self.manager loadFromPreferencesWithCompletionHandler:^(NSError* loadErr){
+        if (loadErr != nil) {
+          logger.debug() << "proxy prefs load failed:"
+                         << loadErr.localizedDescription;
+        }
+        completionHandler(nil);
+      }];
```
artfwo, line 105: not sure i understand what's happening here. when saveToPreferencesWithCompletionHandler reports an error, the code logs it but calls completionHandler(nil) after `loadFromPreferencesWithCompletionHandler` completes?

if the preferences fail to save, is manager state going to change too?

macos/networkextension/VPNSplitTunnelProvider.mm
```diff
@@ -0,0 +1,496 @@
...
+
+  // Evaluate whether the source of this flow should be excluded from the VPN.
+  if (![self matchAppFlow:flow]) {
+    return NO;
+  }
+
+  // Perform flow bypassing.
+  if ([flow isKindOfClass:[NEAppProxyTCPFlow class]]) {
+    NEAppProxyTCPFlow* tcpFlow = (NEAppProxyTCPFlow*)flow;
+    BypassTcpFlow* handler = [BypassTcpFlow createBypass:tcpFlow withInterface:self.ipv4Interface];
```
artfwo: what about ipv6 connections from excluded apps?

macos/networkextension/VPNSplitTunnelProvider.mm
```diff
@@ -0,0 +1,496 @@
...
+      if (error) {
+        NSLog(@"flow closed with error: %@", error);
+      }
+    }];
+
+    std::atomic_fetch_add(&m_handledTcpFlows, 1);
+    return YES;
+  } else if ([flow isKindOfClass:[NEAppProxyUDPFlow class]]) {
+    NEAppProxyUDPFlow* udpFlow = (NEAppProxyUDPFlow*)flow;
+    BypassUdpFlow* handler = [BypassUdpFlow createBypass:udpFlow withInterface:self.ipv4Interface];
```
artfwo: ditto - what about ipv6 connections from excluded apps?

src/platforms/macos/daemon/macosdaemon.h
artfwo: this seems to be removed from macosdaemon.cpp. it looks like no usage of this function is left, so it's safe to remove here as well.

taskcluster/scripts/signing/entitlements.xml
oskirby: good catch, I missed that one

macos/networkextension/bypasstcpflow.mm
```diff
@@ -0,0 +1,171 @@
...
+
+- (void)handleOutbound:(void (^)(NSError *error)) completionHandler {
+  [self.flow readDataWithCompletionHandler:^(NSData *data, NSError *error) {
+    if (error) {
+      [self closeConnection:error completionHandler:completionHandler];
+      return;
+    }
+
+    // If there was no data, try again.
+    if (!data) {
```
encrypt94, line 79: it's possible for both error and data to be nil at the same time? I'm concerned that blindly retrying could lead to an infinite loop, but I haven't found any documentation covering this case

src/platforms/macos/macosappimageprovider.mm
```diff
@@ -0,0 +1,122 @@
...
+
+  // Find the application bundle.
+  NSBundle* bundle = [NSBundle bundleWithIdentifier: id.toNSString()];
+  if (bundle == nil) {
+    return QImage();
+  }
+
+  // Get the application's CFBundleIconName key.
+  NSString* iconName = [bundle objectForInfoDictionaryKey:@"CFBundleIconName"];
+  if (iconName == nil) {
```
encrypt94, line 100: Should we also use CFBundleIconFile as a fallback?

src/platforms/macos/macosapplistprovider.mm
```diff
@@ -0,0 +1,75 @@
...
+void MacOSAppListProvider::getApplicationList() {
+  QList<AppDescription> out;
+
+  QProcessEnvironment pe = QProcessEnvironment::systemEnvironment();
+  if (pe.contains("HOME")) {
+    fetchEntries(pe.value("HOME") + "/Applications", out);
+  }
+
+  fetchEntries("/Applications", out);
+  fetchEntries("/System/Applications", out);
```
encrypt94, line 73: `fetchEntries("/System/Applications/Utilities", out);` should be add here to make `Terminal.app` available in the list

macos/networkextension/routemanager.mm
```diff
@@ -0,0 +1,501 @@
...
+  const struct sockaddr* gateway = rtmLookupAddr(rtm, RTA_GATEWAY, addrlist);
+  const struct sockaddr* dst = rtmLookupAddr(rtm, RTA_DST, addrlist);
+  nw_interface_t iface = nw_interface_create_with_index(ifindex);
+  if (m_delegate) {
+    NSData* gwData = [NSData dataWithBytes:gateway
+                                    length:gateway->sa_len];
+    [m_delegate defaultRouteChanged:dst->sa_family
+                       viaInterface:iface
+                        withGateway:gwData];
+  }
```
encrypt94, line 347: leak on iface? `nw_release(iface)` was here but removed in `7bd332d`

encrypt94 (commented):
Impressive work!

macos/networkextension/routemanager.mm
oskirby: I think the use of an explicit `nw_release(iface)` was an error, and the `iface` will be automatically released when it goes out of scope at the end of this method (assuming I understand Objective-C ARC, which is doubtful).

src/platforms/macos/macoscontroller.mm
oskirby: This only creates a log message for now

oskirby: My longer-term goal with the macOS network extension is to migrate all of the daemon's functionality into it, at which point this error would more sensibly connect to the backend-failure error handling.

But for now, a failure to launch the split tunneling extension is not a fatal event

macos/networkextension/VPNSplitTunnelProvider.mm
oskirby: Unfortunately, we don't really know whether this is an IPv4 or IPv6 connection at this point and this causes some difficulty in determining which of `self.ipv4interface` or `self.ipv6interface` should be used here, so we just use `self.ipv4interface` and hope for the best. If both the ipv4 and ipv6 default routes go over the same network interface, then split tunneling happens to work just fine regardless of the connection is using IPv4 or IPv6.

For macOS 15.0 and later, we could possibly get a hint by doing:
```
if (nw_endpoint_get_type(tcpFlow.remoteFlowEndpoint) == nw_endpoint_type_address) {
  int family = nw_endpoint_get_address(tcpFlow.remoteFlowEndpoint)->sa_family;
  nw_interface_t boundInterface = (family == AF_INET6) ? self.ipv6interface : self.ipv4interface;
}
```

However, this hint only works if the flow was created using the socket API. Otherwise, the endpoint will only give us a domain name or URL. In which case we don't know the address family that will be used.

macos/networkextension/VPNSplitTunnelProvider.mm
oskirby: For UDP, the situation is a even less clear. At the time of the flow creation there is no guarantee we will get any indication of where the packet flow is headed as there is no `self.remoteFlowEndpoint` property on the `NEAppProxyUDPFLow` class and instead we are given an optional `self.localFlowEndpoint` (or `self.localEndpoint` for macOS < 15.0) and these endpoints are likely to be `nil` unless the flow gets bound to a port.

Probably the best way to resolve this would be to pass both the ipv4 and ipv6 interfaces into `createBypass` and then defer the socket creation until after we handle the first datagram in the flow.

It's messy, I agree.

src/platforms/macos/macosextensionloader.mm
oskirby: It's been a while since I wrote this, but IIRC this block of code is basically just to force `self.manager.enabled` to `true`. The [NEVPNManager](https://developer.apple.com/documentation/networkextension/nevpnmanager/isenabled?language=objc) documentation states that:

> Note that you will need to re-load the VPN configuration from the preferences in order to see the change in value.

So, what we are doing here is setting `self.manager.enabled` then saving the prefs, and immediately re-loading the prefs. After which we notify the `MacOSController` of that setup should be finished by calling the `completionHandler()` We don't explicitly handle errors here...  should anything fail we will wind up with `self.manager.enabled` as `false` and the split-tunneling feature will appear disabled (the check for this occurs in `MacOSController::splitTunnelSupported()`

artfwo (approved):
Looks good!

taskcluster/scripts/signing/entitlements.xml
```diff
@@ -1,27 +1,40 @@
...
-         </array>
-        <key>com.apple.developer.team-identifier</key>
-         <string>43AQ936H96</string>
+<dict>
+	<key>com.apple.application-identifier</key>
+	<string>43AQ936H96.org.mozilla.macos.FirefoxVPN</string>
+
+	<key>keychain-access-groups</key>
+	<array>
+		<string>43AQ936H96.mozillavpn</string>
```
hneiva, line 14: unrelated: I wonder how hard it would be to migrate the app id to mozillavpn instead of FirefoxVPN

hneiva (approved):
Taskcluster bits LGTM

PR #11200, opened 2026-04-10
VPN-5455: iOS app intents

mcleinman:
## Description

This brings App Intents to iOS. This allows users to activate/deactivate the VPN connection via Shortcuts, Spotlight search, Siri-suggested actions, build it into automation w/ Shortcuts app, etc. This also provides the building blocks for several other future iOS features: https://mozilla-hub.atlassian.net/browse/VPN-5453

This is fairly straightforward in a native iOS app that is using Apple's translations system. We *must* use Apple's translation system, as iOS handles much of this (identifying a user's intent when talking to Siri, etc.) without waking our app. Given our setup, you'll see a variety of build scripts to take our translation system and convert it into the `xcstrings` files that iOS requires. This conversion is done every time the iOS app is built, to provide the freshest strings possible. Some of this script was written by Claude, but I've carefully reviewed it (and reworked much of it). I've added a bunch of checks in the translation script, so we should catch problems when iOS builds fail.

There are some wonky translation things, like needing the literal sting `\(applicationName)` in any Siri commands, having a potentially different number of translations in different languages for the activation Siri commands (we'd like to offer any reasonable ones - Turn On, Activate, Connect, Swtich On, etc. - but the count may differ by language). I've put in my first attempt, but am happy to refine based on feedback from l10n team.

Several other things:
- To provide system feedback like "Connected to Seattle, WA", I did some work to include the server names in a config.
- There is a bunch of work to suppress our app's notifications if the activation/deactivation came via App Intent. App Intents (for instance, "Hey Siri turn on Mozilla VPN") give a confirmation message, making our app's notification redundant - the user was getting two notifications in a row (one from system, one from our app). We cannot suppress the system's notification, so we're suppressing our own.
- From testing, it seems like Apple will accept the closest match for a translation. For instance, someone whose system is set to use Latin American Spanish (which we don't support) should be able to turn on Siri using translations for other Spanishes that we do support.
- You can trigger the connection via Siri if the app is open, or if it's closed - it works well either way.
- If you're not signed in (but have the app downloaded), it gives an error message. The same is true if you're signed in, but have not given permissions for the network extension to be installed.

Some decisions:
- Where an icon is available, we're using a shield, not a lock. That seems more in line with what we do in the client.
- Surprising to me, the canonical terminology we use with users is "turn on" and "turn off", not "activate" and "deactivate". We obviously use activate/deactivate far more in the code. I've followed this practice by using "Turn On" as the canonical action - though Siri allows a range of commands in addition to "turn on", including "activate".

While this may be surprising based on the length of the PR, I deferred some future work to additional tickets/PR to try to keep this "small-ish", for some definition of "small-ish":
- VPN-7374 - improve timer countup in app when starting from intent
- Should there be an action for "toggle VPN", turning it on/off (based on current status)? Should their be an action for asking whether the VPN is currently active (which may be helpful for Shortcuts, etc.)?
- Bring this to macOS - it should be fairly trivial at this point
- If the VPN is off and you change the setup in the app (switch servers, change from multihop to singlehop, etc.), the setup is not persisted to the network extension. Thus, if you then activate the VPN via "Hey Siri", you get the prior server, not the one you changed to in the app.

## Images/videos

From searching:
https://github.com/user-attachments/assets/36b5453d-b5ca-4d0d-aeac-bbd752219277

Can build into shortcuts:
https://github.com/user-attachments/assets/7777e77e-130e-4ebc-b46e-96699dade7fc

Offered as suggestion (see "Turn On Mozilla VPN" line):
Screenshot 93

No videos of "Hey Siri, turn on VPN" - but that is part of this PR as well

## Reference

VPN-5455

## Checklist
    

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,228 @@
...
+        print(f"No translations found: {xliff_path}")
+        exit(1)
+    return translations
+
+SPECIAL_LOCALE_MAP = {
+    'zh_CN': 'zh-Hans',
+    'zh_TW': 'zh-Hant',
+}
+
+def normalize_locale(locale):
```
flodolo, line 57: Probably no need for a function: `SPECIAL_LOCALE_MAP.get(locale, locale).replace('_', '-')`.

src/translations/strings.yaml
```diff
@@ -1071,3 +1071,72 @@ controller:
...
+  turnOnAlreadyConnectedError:
+    value: VPN is already connected
+    comment: Error given by Siri when VPN already has an active connection
+  turnOffError:
+    value: No active VPN connection
+    comment: Error given by Siri when unable to turn off VPN due to a lack of existing connection
+
+iosAppIntentsActivate:
+  activate1:
+    value: Activate ${applicationName}
```
flodolo: We shouldn't introduce a new and unrecognized format for variable placeholders. We should stick to `%@`, and replace it later — if needed. That way we'll also have linters to check if a translation is missing it.

flodolo (changes requested):
How urgent is this work? I'll be out until April 20, and this might require a larger conversation (I definitely need time to review it).

I don't think the "5 variants" strings for a command works: what if a locale doesn't have 5, and the translations are the same? What if they need more? Do we need a 1:1 match with English?

mcleinman:
> How urgent is this work? I'll be out until April 20, and this might require a larger conversation (I definitely need time to review it).
> 
> I don't think the "5 variants" strings for a command works: what if a locale doesn't have 5, and the translations are the same? What if they need more? Do we need a 1:1 match with English?

Definitely okay waiting until you're back to have a discussion about this. Some of the decisions - like the activate words - I took a first stab at, but thought you may have suggestions on how to improve this to work with pontoon and other standards.

We'll talk when you're back - it might be easiest to set up some time to talk through some of this?

scripts/utils/generate_xcstrings.py
mcleinman: Given that the function is used twice in the code, I prefer keeping the function to minimize chances that a future maintainer does the `replace('_', '-')` while forgetting to do the check for Chinese.

src/translations/strings.yaml
mcleinman: This has been updated. Thanks.

scripts/utils/generate_xcstrings.py
flodolo: I would still use the one-liner

```python
return SPECIAL_LOCALE_MAP.get(locale, locale).replace('_', '-')
```

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,230 @@
...
+    if locale in SPECIAL_LOCALE_MAP:
+        return SPECIAL_LOCALE_MAP[locale]
+    return locale.replace('_', '-')
+
+def get_locales(i18n_dir):
+    """Returns list of locale codes that have a mozillavpn.xliff."""
+    locales = []
+    for entry in os.listdir(i18n_dir):
```
flodolo, line 65: Not a functional change, just more pythonic (replace for+if with list comprehension).

```python
      locales = [
          entry for entry in os.listdir(i18n_dir)
          if os.path.isfile(os.path.join(i18n_dir, entry, "mozillavpn.xliff"))
      ]
```

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def build_localizable_xcstrings(main_strings, locale_translations):
+    """Build Localizable.xcstrings dict for iosAppIntents.main.* strings."""
+    strings = {}
+
+    for data in main_strings.values():
+        english_value = data['value'][0]
+        string_id = data['string_id']
+        comment = data['comments'][0] if data['comments'] else None
```
flodolo, line 82: Should we care about comments since these files are generated and not used anywhere by the localization pipeline?

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,230 @@
...
+        if comment:
+            entry['comment'] = comment
+
+        localizations = {}
+        for locale, translations in locale_translations.items():
+            translated = translations.get(string_id)
+            if translated:
+                for placeholder in re.findall(r'%.{0,2}@', english_value):
+                    if placeholder not in translated:
+                        print(f"Error: placeholder '{placeholder}' from missing from {locale} translation of '{string_id}'")
```
flodolo, line 94:
```suggestion
                        print(f"Error: placeholder '{placeholder}' missing from {locale} translation of '{string_id}'")
```

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,230 @@
...
+
+def build_phrase_section(phrase_strings, locale_translations):
+    """Build the xcstrings entry for a set of Siri phrases (stringSet format)."""
+    en_values = [using_app_placeholder(v) for data in phrase_strings.values() for v in data['value']]
+
+    localizations = {
+        'en': {'stringSet': {'state': 'new', 'values': en_values}}
+    }
+
+    for locale, translations in locale_translations.items():
```
flodolo, line 118: You're going over `en` a second time in this cycle.

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,230 @@
...
+        english_value = data['value'][0]
+        string_id = data['string_id']
+        comment = data['comments'][0] if data['comments'] else None
+
+        entry = {}
+        if comment:
+            entry['comment'] = comment
+
+        localizations = {}
+        for locale, translations in locale_translations.items():
```
flodolo, line 89: This goes over `en` a second time.

flodolo (commented):
As mentioned, we need to figure out how to approach asymmetry between English and translations.

I don't understand a lot about iOS, but I assume that things are going to get confusing if locales repeat the same translations, because they don't have all the variants (e.g. `Turn on %@` and `Turn %@ on` will be identical in most locales). Maybe we need some sort of deduplication? Is iOS going to be OK with that?

```python
  if locale_values:
      seen = dict.fromkeys(locale_values)  # preserves order, removes dupes
      localizations[normalize_locale(locale)] = {
          'stringSet': {'state': 'translated', 'values': list(seen)}
      }
```

One possible solution is to change the approach:
- Use one string for all forms of a command, one variant per line.
- Split lines and generate the string catalog. Make sure to lint each line individually for the placeholder.

```yaml
iosAppIntentsActivate.phrases:
  value: |
    Turn on %@ VPN
    Activate %@ VPN
    Connect to %@ VPN
    Enable %@ VPN
    Start %@ VPN
    Turn %@ VPN on
    Use %@ VPN
  comments:
    - >
      Siri phrase variants for activating the VPN via App Intents.
      Each line is a separate phrase Siri will recognize.
      Rules:
        - Write natural, conversational phrases a user might say to Siri.
        - Do NOT translate literally — write phrases natural in your language.
        - Every line MUST contain %@ (the app name placeholder).
        - Add or remove lines freely; your language may need fewer or more variants.
        - No line should be a duplicate of another.
```

scripts/utils/generate_xcstrings.py
mcleinman: I suppose it's not strictly necessary, but it's nice to see them in the iOS translations view if reviewing the code.

scripts/utils/generate_xcstrings.py
mcleinman: I don't see it in this one, though. We pull out the English value initially for one purpose: to get the proper format for the placeholder (as sometimes it's `%@`, other times things like `%1$@`). We don't append the English translation initially - only when it is iterated through as a language. 

Perhaps I'm missing something - are you seeing something else here?

(That said, I need update this so English no longer is tagged as `translated`.)

scripts/utils/generate_xcstrings.py
mcleinman: Great catch, thanks. Fixed.

scripts/utils/generate_xcstrings.py
mcleinman: Thanks, updated

scripts/utils/generate_xcstrings.py
mcleinman: Updated, thanks. (I'm not a python pro - thanks for the tips. You've seen the sum total of all production Python code I've ever written; this pipeline code is the first time I've really done Python.)

mcleinman:
> As mentioned, we need to figure out how to approach asymmetry between English and translations.
> 
> I don't understand a lot about iOS, but I assume that things are going to get confusing if locales repeat the same translations, because they don't have all the variants (e.g. `Turn on %@` and `Turn %@ on` will be identical in most locales). Maybe we need some sort of deduplication? Is iOS going to be OK with that?
> 
> ```python
>   if locale_values:
>       seen = dict.fromkeys(locale_values)  # preserves order, removes dupes
>       localizations[normalize_locale(locale)] = {
>           'stringSet': {'state': 'translated', 'values': list(seen)}
>       }
> ```
> 
> One possible solution is to change the approach:
> 
>     * Use one string for all forms of a command, one variant per line.
> 
>     * Split lines and generate the string catalog. Make sure to lint each line individually for the placeholder.
> 
> 
> ```yaml
> iosAppIntentsActivate.phrases:
>   value: |
>     Turn on %@ VPN
>     Activate %@ VPN
>     Connect to %@ VPN
>     Enable %@ VPN
>     Start %@ VPN
>     Turn %@ VPN on
>     Use %@ VPN
>   comments:
>     - >
>       Siri phrase variants for activating the VPN via App Intents.
>       Each line is a separate phrase Siri will recognize.
>       Rules:
>         - Write natural, conversational phrases a user might say to Siri.
>         - Do NOT translate literally — write phrases natural in your language.
>         - Every line MUST contain %@ (the app name placeholder).
>         - Add or remove lines freely; your language may need fewer or more variants.
>         - No line should be a duplicate of another.
> ```

Duplicates shouldn't matter - we're basically making dictionaries for iOS for Siri phrases for a given intent, and I'm not sure a dup will matter. One issue with my approach is the maximum number of translations is whatever I put for English. I don't *love* that.

If one-per-line (and many lines per translation) is the best path for this, I'm happy to do it that way. Will get another commit up for this shortly.

src/translations/strings.yaml
```diff
@@ -1071,3 +1071,80 @@ controller:
...
+    comment: Description used to provide more context in Shortcuts and other places
+  turnOnConfirmation:
+    value: Mozilla VPN connected
+    comment: Confirmation string used by Siri after turning on VPN
+  turnOnConfirmationSingleHop:
+    value: Mozilla VPN connected through %@
+    comment: Confirmation string used by Siri after turning on VPN. `%@` must be included, and it will be replaced with the name of the exit city.
+  turnOnConfirmationMultiHop:
+    value: Mozilla VPN connected through %1$@ via %2$@
+    comment: Confirmation string used by Siri after turning on VPN. `%1$@` will be replaced with the exit city, and `%2$@` will be replaced with the entry city. They must be included.
```
```suggestion
    comment: Confirmation string used by Siri after turning on VPN. `%1$@` will be replaced with the exit city, and `%2$@` will be replaced with the entry city.
```

The last part is not really necessary, and we have linters to make sure that's the acse.

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,234 @@
...
+        exit(1)
+    return locales
+
+
+def build_localizable_xcstrings(main_strings, locale_translations):
+    """Build Localizable.xcstrings dict for iosAppIntents.main.* strings."""
+    strings = {}
+
+    for data in main_strings.values():
+        english_value = data['value'][0]
```
flodolo: I think this can return a string or a list, so maybe worth doing this?

```
value = data['value']
english_value = value if isinstance(value, str) else value[0]
```

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,234 @@
...
+          translations[trid] = unit.findtext(f'{XLIFF_NS}source', default='')
+        else:
+          translations[trid] = unit.findtext(f'{XLIFF_NS}target', default='')
+    if not translations:
+        print(f"No translations found: {xliff_path}")
+        exit(1)
+    return translations
+
+def using_app_placeholder(value):
+    return value.replace("%@",f"${{applicationName}}")
```
flodolo:
```suggestion
    return value.replace("%@", "${{applicationName}}")
```

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,234 @@
...
+
+    localizations = {}
+    for locale, translations in locale_translations.items():
+        locale_values = []
+        translation_block = translations.get(string_id)
+        if translation_block is None:
+          continue
+        for translation in translation_block.split('\n'):
+          if translation:
+              if not f"%@" in translation:
```
flodolo:
```suggestion
              if not "%@" in translation:
```

scripts/utils/generate_xcstrings.py
flodolo: No, you're correct, we need to loop over `en` here.

flodolo (commented):
The generated .ts file looks quite broken compared to previous runs?
https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/24855325028


```
    <message id="vpn.languageView.systemLanguageButtonTitle">
        <location filename="translations/generated/i18nstrings_p.cpp" line="73"/>
        <source>S
y
s
t
e
m
 
l
a
n
g
u
a
g
e</source>
        <extracomment>the title for the system language radio button</extracomment>
    </message>
```

flodolo:
Side note: you might want to run `ruff format` and `ruff check` to clean up Python scripts a bit.
https://pypi.org/project/ruff/

scripts/utils/generate_strings.py
```diff
@@ -231,7 +231,10 @@ def serialize(string):
             return ret.replace('"', '\\"')
 
         for key, data in strings.items():
-            output.write(f"    //% \"{serialize(data['value'])}\"\n")
+            # Without `replace("\n","\\\\n")`, the C++ file erroneously ends up with lines that should have been commented out.
+            raw = data['value'] if isinstance(data['value'], str) else '\n'.join(data['value'])
```
flodolo: I think `data['value']` is always a list? Try this (untested, as I don't have qt locally).

```python
  processed = [v.replace('\n', '\\n') for v in data['value']]
  output_value = serialize(processed)
  output.write(f"    //% \"{output_value}\"\n")
```

flodolo (changes requested):

src/translations/CMakeLists.txt
```diff
@@ -8,3 +8,39 @@ generate_translations_target(translations
     ${CMAKE_SOURCE_DIR}/src/translations
     ${CMAKE_SOURCE_DIR}/3rdparty/i18n
 )
+
+if(IOS)
```
strseb, line 12: The way this is built now, is that this is running at the configure step, everytime even when nothing changed. 
Can we change this to a custom_command so it runs in a build job ? :)

src/notificationhandler.cpp
```diff
@@ -103,6 +103,18 @@ void NotificationHandler::showNotification() {
...
+  // enabled (via SettingsManager), otherwise we'll introduce a corner case bug:
+  // If notifications are not active when an iOS AppIntent is run AND the
+  // notification is then toggled to be active in Settings AND the next
+  // notification comes from an action in the client (like turning on the VPN
+  // from the client), then that first single notification would be suppressed.
+  // But with this before the SettingsManager checks, we'll always clear the
+  // shouldSkipNextNotification flag appropriately.
+  if (vpn->controller()->shouldSuppressNextNotification()) {
+    return;
+  }
```
strseb, line 116: This is a case we want to avoid but IMHO for me that this comment is this big but really needed makes me think we should put this somewhere else :D 

I will not block on this ofc but WDYT: 
-> Have notifcationhandler::pause() unpause() - where we just return in notifyInternal(). 
-> Optional: a RAI element for notificationHandler::Supressor that calls those on constr/destr

-> The ioscontroller.swift activate sends a signal "isAppinent"
-> ioscontroller.mm does something along 
```
callback:^(BOOL a_connected, BOOL isAppIntent) {
        logger.debug() << "State changed: " << a_connected;
        auto unpause;
        if(isAppIntent){
           // User was already notified
          NotificationHandler::pause();
          unpause =  qScopeGuard({ 
             NotificationHandler::unpause();
           });
        }
        if (a_connected) {
          // Will now not emit a notification as suppressor is alive
          emit connected(m_serverPublicKey);
          return;
          // from here on notifications are possible as scopeguard re-enables them.
        }
        emit disconnected();
      }];
   
``` 
Then a simple comment of "siri already notifiied of the activation" is enough :)

oskirby: I prefer this approach as well - it does the same job and is much more targeted to the issue at hand rather than spreading the hack across multiple files and APIs.

scripts/utils/generate_xcstrings.py
```diff
@@ -0,0 +1,234 @@
+#!/usr/bin/env python3
```
oskirby, line 1: For (re)build dependency reasons, I am not a fan of scripts that "generate everything, everywhere, all at once" since they tend to either: A) kill re-build performance since they need to do way too much work, or B) don't get rebuilt properly and require blowing out your build-dir and starting again from scratch.

I would recommend refactoring this script so that it takes a single XLIFF/YAML file, and generates a single xcstrings file. Then they can be stitched together using a CMake build rule with proper dependency handling:

```
add_custom_command(
    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/example/Localizable.${LOCALE}.xcstrings
    DEPENDS
        ${CMAKE_CURRENT_SOURCE_DIR}/example/strings.yaml
        ${CMAKE_CURRENT_SOURCE_DIR}/example/${LOCALE}/mozillavpn.xliff
    COMMAND generate_xcstrings.py 
         --xliff ${CMAKE_CURRENT_SOURCE_DIR}/example/${LOCALE}/mozillavpn.xliff
         --input ${CMAKE_CURRENT_SOURCE_DIR}/example/strings.yaml
         --output$ {CMAKE_CURRENT_BINARY_DIR}/example/Localizable.${LOCALE}.xcstrings
)
```

oskirby: Also, we basically already have such a script: `xlifftool.py` and it supports generating files from templates. We already do this for localizing the macOS installer package...

Using a template, such as `macos/pkg/welcome.html.in` containing the contents:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
  </head>
  <body style="color: #1a1919; line-height: 1.5; font-family: -apple-system, BlinkMacSystemFont, 'Helvetica', sans-serif; font-size: 13px; ">
    <br />
    <p style="margin: 5px;">@MACOSINSTALLER_WELCOME_MESSAGE1@</p>
    <br />
    <p style="margin: 5px;">@MACOSINSTALLER_WELCOME_MESSAGE2@</p>
  </body>
</html>
```

It gets processed by a custom command in `macps/pkg/CMakeLists.txt` as follows:
```cmake
add_custom_command(
    OUTPUT
        ${CMAKE_CURRENT_BINARY_DIR}/Resources/${LOCALE}.lproj/Localizable.strings
        ${CMAKE_CURRENT_BINARY_DIR}/Resources/${LOCALE}.lproj/conclusion.html
        ${CMAKE_CURRENT_BINARY_DIR}/Resources/${LOCALE}.lproj/welcome.html
    DEPENDS
        ${PROJECT_ROOT_DIR}/3rdparty/i18n/${LOCALE}/mozillavpn.xliff
        ${CMAKE_CURRENT_SOURCE_DIR}/conclusion.html.in
        ${CMAKE_CURRENT_SOURCE_DIR}/welcome.html.in
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Resources
    COMMAND ${CMAKE_COMMAND} -E make_directory ${LOCALE}.lproj
    COMMAND ${XLIFFTOOL} -x ${CMAKE_CURRENT_SOURCE_DIR}/welcome.html.in -o ${LOCALE}.lproj/welcome.html
    COMMAND ${XLIFFTOOL} -x ${CMAKE_CURRENT_SOURCE_DIR}/conclusion.html.in -o ${LOCALE}.lproj/conclusion.html
    COMMAND ${XLIFFTOOL} -g macosinstaller --strip 1 -f macos -o ${LOCALE}.lproj/Localizable.strings
)
```

oskirby (changes requested):
In terms of comment severity:
- `shouldSuppressNextNotification`: irks me, but this is non-blocking, and the PR can merge with this as is.
- `generate_xcstrings.py`: Blocking. Please refactor this so that it can build individual files with dependency handling using `add_custom_command()`

As for the actual meat of the iOS intent handling, that all looks good as far as I can tell.

src/notificationhandler.cpp
mcleinman: How would you feel about doing a "skipNextNotification" from this C++ layer (that lives in the notification controller) instead of pause/unpause? I considered a bunch of ways of doing this. While I don't love my current solution and am very open to modifying it, I had some concern about pause/unpause-style ones - I was worried that in some conditions, they'd unpause too soon:

`Notification::showNotification` is fired when Controller emits `stateChanged`, which happens happens towards the end of `setState`. On the other side, this `ControllerImpl` callback emits `connected` and `disconnnected`, which in turn call `Controller::connected` and `Controller::disconnected`, which each call `setState` after doing a lot of other work.

I have some concern that at least on some slower systems, we'd end up unpausing before `showNotification` is called.

Further, if this code path gets modified in the future (and controller paths like this change a lot), it could break this functionality and nobody would notice. Without Swift testing infrastructure (and likely Swift infrastructure that interacts with other layers), we're not able to write tests for paths like "starting in Swift controller, and see if notification is fired in C++" .

oskirby: I think adding an `NotificationHandler::skipNext()` is also a good approach. That works for me.

Is there a case where we might call `NotificationHandler::skipNext()` and then the expected 'next' notificatinon never occurs? That is one situation where the `pause()`/`unpause()` approach might work better. For example, is there a failure case where the tunnel doesn't start after an App Intent and nothing happens?

mcleinman: Forgot this in my earlier note - There are many reasons it might fail when coming from `Controller::activate`. There are far fewer when turning on the VPN directly via network extension, which is why I went with that path. (There is a "combo" solution, where we "skipNext" (which ultimately sets a boolean) in C++, but reset that boolean in the error paths. However, there are a *lot* of error paths from the controller. While I could find them all currently, I'm fairly confident that at some point down the line - if we add a new failure path for instance - we would end up introducing a bug.)

I'm happy to implement this, but I think we're going to trade some types of complexity for others. Currently, this touches more files - but we're going to need to keep two booleans if we slim it down.

As is currently, written, we touch these surfaces:
- IOSController (Swift): Keeps `shouldSkipNextNotification` state, sets it to `true` in a couple places, and has a `shouldSuppressNextNotification` function.
- IOSController (C++): `shouldSuppressNextNotification` function that just relays the Swift result
- Controller (C++):`shouldSuppressNextNotification` function that relays the ControllerImpl result
- ControllerImpl.h: Has a default `false` for `shouldSuppressNextNotification`
- NotificationHandler (C++): Checks Controller's `shouldSuppressNextNotification`

Moving to being set in the controller callback, we touch these surfaces:
- IOSController (Swift): Unfortunately, this *still* needs to keep a boolean for `shouldSkipNextNotification` state - the callback comes when the iOS system calls `Notification.Name.NEVPNStatusDidChange` - and so without some state, we have no way to connect the callback with how the VPN was activated. (We could call something to set the boolean on NotificationHandler earlier, but then we get into the "lots of failure states to consider" issue.) Additionally, this state is passed (and then cleared) when the state change callback is called.
- IOSController (C++): Callback takes a new argument as to whether it came from an intent. It calls something in notification handler to set this as a boolean over there.
- NotificationHandler (C++): New boolean to suppress next notification. Also add a function that sets this to `true` (called by C++ IOSController), and check this boolean (and reset it to false) before sending a notification.

While this second one manages to skip modifications to the C++ Controller and the default ControllerImpl.h, it comes at the cost of adding a boolean for state to a second file. Personally, I think it's more straightforward and less error-prone for the future if we just have the Swift file hold the state, and not have to relay the state to a boolean elsewhere. That said, I do *not* feel strongly about this - and would happily be outvoted here. I looked at this a bunch of ways before putting up the PR, and I don't see a single way that is straightforward, minimizes code, and minimizes chance of future bugs.

mcleinman: To add another cherrybomb in here - I have *never* liked these activate/deactivate system notifications. I think they provide no value - a user just chose to do something, we don't need a confirmation notification that it happened. (Perhaps there should be a notification if the user is no longer in the client and there is a *failure*.) So my preferred solution would be "rip out activate/deactivate notifications", but that seems like a bigger can of worms and not something I wanted to open in this work stream.

scripts/utils/generate_xcstrings.py
mcleinman: I agree about the two foot guns that I'd like to avoid. The xcstrings format is essentially structured json, and Apple's newest format for translations. (And of course, it's required for app intents - can't use the older `strings` type.) I started working on rewriting this into two sections - one pythons script that takes one xliff file and writes snippets to disk, another that assembles those snippets. Ultimately, this would be a *lot* of overhead, as Localizable.xcstrings is essentially a bunch of entries that look like this:
```
    "vpn.iosAppIntentsMain.turnOnAction": {
      "comment": "Title used by system to offer turning on VPN",
      "localizations": {
        "fr": {
          "stringUnit": {
            "state": "translated",
            "value": "Urntay Onway Ozillamay PVNWAY"
          }
        },
        "en": {
          "stringUnit": {
            "state": "new",
            "value": "Turn On Mozilla VPN"
          }
        }
    }
}
```

While `AppShortcuts.xcstrings` could fairly easily combine discrete files, there isn't a cheap way to do that for Localizable.xcstrings. I'm clocking the python script at 380-500 ms currently, in total. I'm afraid as soon as we start writing a bunch of files to disk, this will take significantly more time - but with the advantage of not rebuilding it on every run.

Could we make every mozillavpn.xliff file a dependency in cmake? If they're all listed under `DEPENDS` for a custom command, it would not rebuild if nothing changed, right? (I'd also write some code to ensure that all languages were in that list - so we don't miss new ones.)

flodolo:
Most of the comments I left in the last pass are still unaddressed?

scripts/utils/generate_xcstrings.py
mcleinman: This is returning the piece from the `strings.yaml` file, and not for the Siri activation word part - just the rest of it. The rest should all be one liners (and I can't imagine a situation where we would use a list for anything other than the Siri activation part), so I believe this should suffice.

scripts/utils/generate_strings.py
mcleinman: Thanks - this fixes a bug I introduced in there. Putting up fix shortly.

mcleinman:
> The generated .ts file looks quite broken compared to previous runs? https://github.com/mozilla-mobile/mozilla-vpn-client/actions/runs/24855325028
> 
> ```
>     <message id="vpn.languageView.systemLanguageButtonTitle">
>         <location filename="translations/generated/i18nstrings_p.cpp" line="73"/>
>         <source>S
> y
> s
> t
> e
> m
>  
> l
> a
> n
> g
> u
> a
> g
> e</source>
>         <extracomment>the title for the system language radio button</extracomment>
>         <translation type="unfinished"></translation>
>     </message>
> ```

I'm pretty sure that last fix will take care of this as well - it was a shared function in a file used by this script.

mcleinman:
> Most of the comments I left in the last pass are still unaddressed?

Apologies - you're correct, I had not addressed your comments. I believe I have now, and I've ruff'd up the new file. Should be ready for another review. Thanks for all your help with this one.

scripts/utils/generate_xcstrings.py
oskirby: Yes, there's no issue with gathering all of the `mozillavpn.xliff` files under `DEPENDS` and that would be correct in this case where the output of this script depends on literally all of them.

In order to detect when new languages get added, that is also pretty easy in the CMake world. the `file(GLOB ... CONFIGURE_DEPENDS)` option does precisely that, and will trigger a re-configuration when the detected fileset changes.

So, you would do:
```cmake
file(GLOB_RECURSE XLIFF_FILE_LIST CONFIGURE_DEPENDS
    "${CMAKE_SOURCE_DIR}/3rdparty/i18n/*/mozillavpn.xliff"
)
add_custom_command(
    OUTPUT Localizable.xcstrings
    DEPENDS ${XLIFF_FILE_LIST}
    COMMAND ${CMAKE_SOURCE_DIR}/scripts/utils/generate_xcstrings.py ...
)
```

oskirby: And yes, I see now why we can't break this down into a separate localization file per locale, the format requires them all to get merged together.

src/notificationhandler.cpp
oskirby: I dislike the activate/deactivate notifications too, they kind of just devolve into noise. But if we were to get rid of them, that is work that should go into another PR.

oskirby: I think my distaste for the `shouldSuppressNextNotification()` approach is it creates a complicated dependency interplay between the various controller classes and the the `NotificationHandler` that seems like a recipe for creating future bugs as code evolves.

Consider:
  - `shouldSuppressNextNotification()` is not idempotent and it's value changes after being called. It's actually more like a semaphore or reference count in that regard.
  - A notification has to occur in order to clear it, so if something goes amiss with the activation then we can wind up suppressing some other notification at some future time.
 - The setting of this flag, and how it gets cleared are driven by two very different areas of the codebase.

flodolo (changes requested):
Almost there, but looks like we lose the comment to the multiline strings

```
    <message id="vpn.iosAppIntentsActivate.phrases">
        <location filename="translations/generated/i18nstrings_p.cpp" line="1557"/>
        <source>Activate %@
Turn on %@
Turn %@ on
Switch on %@
Switch %@ on
Connect %@
Connect to %@
</source>
    </message>
```

Also tested the full extraction here https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/565

mcleinman:
> Almost there, but looks like we lose the comment to the multiline strings
> 
> ```
>     <message id="vpn.iosAppIntentsActivate.phrases">
>         <location filename="translations/generated/i18nstrings_p.cpp" line="1557"/>
>         <source>Activate %@
> Turn on %@
> Turn %@ on
> Switch on %@
> Switch %@ on
> Connect %@
> Connect to %@
> </source>
>         <translation type="unfinished"></translation>
>     </message>
> ```
> 
> Also tested the full extraction here [mozilla-l10n/mozilla-vpn-client-l10n#565](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/565)

This has been fixed, I believe. 

Two issues:
1) There was a typo. While `comments` are used in some systems, our yaml file requires `comment`. 
2) Needed the same change for multiline comments that we did for multiline translation strings.

flodolo (approved):
Seems to work now 👍🏼

src/translations/CMakeLists.txt
```diff
@@ -8,3 +8,42 @@ generate_translations_target(translations
...
+    set(TRANSLATIONS_OUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
+    file(MAKE_DIRECTORY "${TRANSLATIONS_OUT_DIR}")
+
+    file(GLOB_RECURSE XLIFF_FILE_LIST CONFIGURE_DEPENDS
+        ${CMAKE_SOURCE_DIR}/3rdparty/i18n/*/mozillavpn.xliff
+    )
+
+    add_custom_command(
+        OUTPUT ${TRANSLATIONS_OUT_DIR}/Localizable.xcstrings ${TRANSLATIONS_OUT_DIR}/AppShortcuts.xcstrings
+        DEPENDS ${XLIFF_FILE_LIST}
```
oskirby: This should probably also depend on:
- `${GENERATE_XCSTRINGS_SCRIPT}`
- `${CMAKE_SOURCE_DIR}/src/translations/strings.yaml`

src/translations/CMakeLists.txt
```diff
@@ -8,3 +8,42 @@ generate_translations_target(translations
...
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    )
+
+    set_source_files_properties(
+        ${TRANSLATIONS_OUT_DIR}/Localizable.xcstrings
+        ${TRANSLATIONS_OUT_DIR}/AppShortcuts.xcstrings
+        PROPERTIES GENERATED TRUE
+    )
+
+    add_custom_target(generate_xcstrings
```
oskirby, line 44: Maybe add a comment that this is just a springboard to ensure `generate_xcstrings` runs to completion before starting the `mozillavpn` build? Also, after fixing the `ANY` script execution bug, did we check if this was even necessary?

oskirby (approved):
I still maintain my dislike of the `shouldSuppressNextNotification` approach, but that is not enough for me to hold up the pipeline of PRs. Otherwise, this all looks good to me.

Approved!

src/translations/CMakeLists.txt
mcleinman: I just spent longer with it than I should have probably, and it does indeed seem necessary. I also realized that the current format didn't actually allow the files to show up in Xcode, so I've fixed that.

mcleinman:
> I still maintain my dislike of the `shouldSuppressNextNotification` approach, but that is not enough for me to hold up the pipeline of PRs. Otherwise, this all looks good to me.
> 
> Approved!

Thanks! I'm happy to talk about a better way, even after merging this. Like my comment above said - I don't love this, but getting rid of `shouldSuppressNextNotification` would probably mean adding a second state boolean elsewhere -and I hate that even more.

src/translations/CMakeLists.txt
mcleinman: Good call, added.

PR #11202, opened 2026-04-13
VPN-7473 Bundle All builds & installs in one INSTALLING.md

strseb:
## Description

I kinda took inspiration of [jj-vcs](https://docs.jj-vcs.dev/latest/install-and-setup/) - to bundle all builds and modes of compilation we support. I wanted to have a single file, to just overall reduce the amount of files flying around. 

I've updated the wasm guide, as that was a bit out of date. 

Then i've linked all install options directly in the readme. As imho installing & building is probably the most interesting questions. 

The hooks and dev tool hints were moved to Contributing.md, just to trim the readme.


mcleinman:
What are the new scripts? And nudge on the linter issues.

strseb:
> What are the new scripts? And nudge on the linter issues.

Update for the wasm compilation steps, given we already have a conda env-wasm.yml that we already maintain we can use it. that script will setup the conda-env for you, just as we already do on CI.

mcleinman (approved):
Looks good. Would love another set of eyes on the wasm setup scripts before merging.

INSTALLING.md
```diff
@@ -0,0 +1,398 @@
...
+```bash
+cmake -S . -B build -GNinja
+cmake --build build
+```
+
+The built binary is at `build/src/mozillavpn`. Note that `wireguard-tools` is also required at runtime.
+You can run the app with `mozillavpn`
+Alternatively, you can use two terminals to run the daemon manually 
+`sudo mozillavpn linuxdaemon` and separately `mozillavpn `
+
```
encrypt94, line 145: Missing install instructions with `sudo cmake --install build`

INSTALLING.md
```diff
@@ -0,0 +1,398 @@
...
+Then install the Python build dependencies:
+
+```bash
+pip install -r requirements.txt
+```
+
+Then configure and build:
+
+```bash
+cmake -S . -B build -GNinja
```
encrypt94, line 137: We should add the flag `-DCMAKE_INSTALL_PREFIX="/"` otherwise native manifest json files (also apparmor files) ends up being installed in the wrong directory and the browser extension won't work

INSTALLING.md
```diff
@@ -0,0 +1,398 @@
...
+
+Configure and build:
+
+```bash
+qt-cmake -S . -B build-wasm -GNinja \
+    -DCMAKE_BUILD_TYPE=MinSizeRel \
+    -DBUILD_TESTS=OFF
+cmake --build build-wasm
+```
+You will find the index.html in `<build-folder>/wasm_build/index.html`
```
encrypt94, line 384: I think this can be changed to `./build-wasm/wasm-build/index.html` as it will always be the same path

encrypt94 (approved):
r+wc, i love the new format!
Gave the WASM build process a try on a clean env, worked perfectly

PR #11203, opened 2026-04-13
Support IPv6-only networks on macOS

artfwo:
## Description

Add IPv6-only network support for macOS, following the Linux support added in #11171

ICMPv6 ping support in MacOSPingSender, in particular `icmp6SocketReady()` is a bit simplified compared to IPv4 implementation, as  IPv6 never includes the IP header in received data so no header-skipping required.

## Reference

[VPN-4369](https://mozilla-hub.atlassian.net/browse/VPN-4369)

## Checklist
    
[VPN-4369]: https://mozilla-hub.atlassian.net/browse/VPN-4369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/macos/daemon/wireguardutilsmacos.cpp
+++ b/src/platforms/macos/daemon/wireguardutilsmacos.cpp
@@ -8,8 +8,6 @@
 #include <Security/SecRequirement.h>
 #include <Security/SecStaticCode.h>
 #include <Security/SecTask.h>
-#include <arpa/inet.h>
-#include <errno.h>
 #include <net/route.h>
 
 #include <QByteArray>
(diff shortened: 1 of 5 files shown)


mcleinman (approved):
Looks good to me, but would like Naomi's approval on here as well, as she is deeper in this area than I am.

src/platforms/macos/daemon/wireguardutilsmacos.cpp
```diff
@@ -28,6 +30,28 @@ constexpr const char* WG_RUNTIME_DIR = "/var/run/wireguard";
...
+static bool hasRouteToIPv4(const QString& ipv4addr) {
+  int sock = ::socket(AF_INET, SOCK_DGRAM, 0);
+
+  if (sock < 0) {
+    return true;
+  }
+
+  auto guard = qScopeGuard([sock] { ::close(sock); });
+
+  struct sockaddr_in dest {};
```
oskirby, line 43: Don't we need to set `dest.sin_len = sizeof(dest);` on OSX?

oskirby (approved):
LGTM. I think the gist of it looks good. Just the one comment about setting `sin_len`

src/platforms/macos/daemon/wireguardutilsmacos.cpp
artfwo: Good catch, thanks! fixed in 57765f02064c3971972b7271da69449476925848

artfwo:
@oskirby @mcleinman there was the same issue with server switching, as we previously had on Linux. By the time updatePeer runs again for a server switch, the wireguard tunnel is already up and has a catch‑all IPv4 default through utun4.

I've removed the hasRouteToIPv4() function and switched to a new getter in MacosRouteMonitor - hasIpv4DefaultRoute(). MacosRouteMonitor also tracks the underlying network's default gateways and ignores tunnel-interface routes, and the change (https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11203/commits/0f77a288d00967b1c56bebb04425057929643cc7) fixed both single-hop and multi-hop server switching. Can you have another look whenever time allows?

src/platforms/macos/macospingsender.h
```diff
@@ -17,16 +17,19 @@ class MacOSPingSender final : public PingSender {
...
+  bool isValid() override { return (m_socket >= 0) || (m_socket6 >= 0); };
 
   void sendPing(const QHostAddress& dest, quint16 sequence) override;
 
  private slots:
   void socketReady();
+  void icmp6SocketReady();
 
  private:
   QSocketNotifier* m_notifier = nullptr;
```
oskirby, line 29: as a non-blocking comment, it seems like we only ever use one of these at a time, so we could just re-use the one `QSocketNotifier` and socket handle for both the IPv4 and IPv6 case, we just need to store a flag somewhere to indicate the address family being used.

Eg:
```
QSocketNotifier* m_notifier = nullptr;
int m_socket = -1;
int m_family = AF_INET;
```

oskirby (commented):
LGTM

PR #11208, opened 2026-04-15
fix typos to allow Android RC builds

mcleinman:
In today's edition of "release automation fails in unexpected ways we have never seen before on VPN team", it seems like we had a typo in a recent update. It causes these two jobs to fail in the release process:
https://firefox-ci-tc.services.mozilla.com/tasks/JTUth8DySYuEBGT_xhmVpg
https://firefox-ci-tc.services.mozilla.com/tasks/Cf4PMDX6T8SQgJQjIgX1CQ

After this is merged, it will be updated to the 2.36.0 release branch.


oskirby (approved):
LGTM

Eijebong (approved):

PR #11222, opened 2026-04-20
Nit: Fix -Werror,-Wunused-const-variable

strseb:
## Description

Clang Seems unhappy that those are unused when compiling on linux.


encrypt94 (approved):
Clang deserves to be happy!

PR #11223, opened 2026-04-20
Nit: Const-i-fy FeatureList impl

strseb:
## Description

feature.h costs more than it should. Every feature allocates a Feature*that lives in both a hashmap and a QList. Every string field starts with UTF-8 char* and gets parsed into to a QString — one temp allocation per field, per feature, at startup.

FeatureList::MaybeInit ends up doing roughly as many temp allocations as the entire MozillaVPN constructor.

before29

Evaluation isn't free either. Feature::isEnabled walks through multiple std::functions, a string-based dependency lookup, and a SettingsHolder round-trip in case Guardian ever flipped the feature. Even a feature hard-wired to
 FeatureCallback_false goes through the full pipeline — the compiler can't
 prove "always false" because the list is built at runtime.

But most of our flags are hard-coded against compile defines.
We're paying for flexibility we never use, on every start and every check.

My proposal is to swap this into a "Pay What you need" model with 3 Tiers of FeatureFlags: 

-> ConstantFeature : Value is known at compile time (Platform::windows, Qt version, build defines).                                                    constexpr — zero allocation, fully foldable by the compiler. 

-> RuntimeFeature - Value needs a runtime check but never flips (e.g. "disable Windows split-tunnel when Mullvad is detected", Android SDK version).  One function-pointer call.  

-> Dynamic Feature: OverridableFeature,  Dev-menu and Guardian (/api/v1/features) can override.                                                                           Function-pointer call + settings lookup, only when you opt in. 

None of the feature structs are QObjects. For QML, a small FeatureProxy
 view-model wraps any AnyFeature reference and exposes the same API QML was
 already using — so MZFeatureList.get("foo").isSupported still works
 unchanged.

---- 

I've split this into 2 Commits to make it nicer to review, part 1 only concerns changes in `/feature` - 
part 2 is a bulk change for consumers of the featurelist.


src/feature/features.h
```diff
@@ -0,0 +1,354 @@
...
+#endif
+
+#ifdef MZ_MACOS
+inline constexpr bool macos = true;
+#else
+inline constexpr bool macos = false;
+#endif
+
+#ifdef MZ_LINUX
+inline constexpr bool linux_ = true;
```
mcleinman, line 29: Such a tiny nit, sorry - but this is `linux_`?

mcleinman (commented):
One tiny tiny nit. Thanks for this.

mcleinman (approved):
That was supposed to approval w/ that non-blocking nit. Sorry.

src/feature/features.h
strseb: just `linux` is a reserved keyword in gcc 😰 - i though about gnu_linux, but not sure if editors would 
 auto suggest that if you type plattform::linux.

PR #11225, opened 2026-04-21
Fix firewall inbound mark/clear mismatch on server switch

artfwo:
## Description

Multihop server switching on ipv6-only networks may break in some configurations, when `hasRouteToIPv4` in `deletePeer` returns a different result than it did in `updatePeer`, so `clearInbound` is called with an address that wasn't marked.

This is an attempt to fix it by tracking endpoint address actually passed to `markInbound`, so the addresses in `updatePeer` and `deletePeer` are the same.

A potentially cleaner fix would be to pass the endpoint address explicitly to `deletePeer`, rather than having `deletePeer` re-derive it, but that would require a larger refactor of WireguardUtils interface and all implementations.

## Reference

[VPN-4367](https://mozilla-hub.atlassian.net/browse/VPN-4367)

## Checklist
    
[VPN-4367]: https://mozilla-hub.atlassian.net/browse/VPN-4367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/daemon/wireguardutilslinux.cpp
+++ b/src/platforms/linux/daemon/wireguardutilslinux.cpp
@@ -284,6 +284,8 @@ bool WireguardUtilsLinux::updatePeer(const InterfaceConfig& config) {
     return false;
   }
 
+  m_peerEndpoints[config.m_serverPublicKey] = endpointAddr;
+
   // Set/update peer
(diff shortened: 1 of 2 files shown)


oskirby (approved):
Looks good, nice find

Issue #9511, opened 2024-05-08
Long app name is not wrapped in App Exclusions


data-sync-user:
**Important details:**

* The issue was not reproducible in 2.21.

**Version/build found in:**

* 2.22.0 (2.202405071948);

**Reproducible on:**

* All;

**Prerequisites:**

* Have Mozilla VPN installed and open;

* Have an account with an active subscription

* Be signed in;

**Steps to reproduce:**

1. On your device, rename any app that is part of Mozilla VPN's app exclusion list, to a longer name;
1. Go to Mozilla VPN and navigate to Settings - App exclusions and add the renamed app from step 1;
1. Observe the screen;

**Expected result:**

* There are no UI issues for the app with the long name, the app name is wrapped.

**Actual result:**

* There are UI issues -> the full name of the app is cut, not visible, instead of being wrapped;

**Notes:**

* See attached ss;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/VPN-6371)

PR #11228, opened 2026-04-23
VPN-6371 Fix text wrapping on excluded apps list

encrypt94:
## Description

Fix text wrapping on excluded app list

Screenshot from 2026-04-23 13-06-39

Also: restore text wrapping for city names. When fixing the right padding issue in #11219 i replaced wrapping with right truncation. This is incorrect, restore text wrapping as intended.

Screenshot from 2026-04-23 13-24-09

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-6371)

#9511 

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZRadioDelegate.qml
+++ b/nebula/ui/components/MZRadioDelegate.qml
@@ -149,8 +149,7 @@ RadioDelegate {
         text: radioButtonLabelText
         lineHeightMode: Text.FixedHeight
         lineHeight: MZTheme.theme.labelLineHeight
-        wrapMode: Text.NoWrap
-        elide: Text.ElideRight
+        wrapMode: Text.Wrap
     }
 
     background: Rectangle {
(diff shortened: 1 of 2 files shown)


nebula/ui/components/MZRadioDelegate.qml
```diff
@@ -149,8 +149,7 @@ RadioDelegate {
         text: radioButtonLabelText
         lineHeightMode: Text.FixedHeight
         lineHeight: MZTheme.theme.labelLineHeight
-        wrapMode: Text.NoWrap
-        elide: Text.ElideRight
```
artfwo, line 153: i wonder why eliding didn't work as expected

artfwo (approved):

nebula/ui/components/MZRadioDelegate.qml
encrypt94: oh it was actually working as expected, it prevented the overflow. However it wasn’t the intended behavior for this component, in general we should not truncate but display the full translation whenever possible. I stupidly switched from wrapping to eliding while fixing the padding bug.

PR #11238, opened 2026-04-29
VPN-7542: iOS control center shortcut

mcleinman:
## What it does

This gives us a toggle control that can be used in 3 places:

As a control center widget:
https://github.com/user-attachments/assets/c952f39b-fb42-42b8-b9ec-38d0fb8b998f

From the lockscreen for folks that like VPNs more than camera (but set that this can only be toggled when unlocked):
https://github.com/user-attachments/assets/7f2e61fb-3af2-423e-b7ba-0ae747312c19

And if you really love toggling your VPN, it should work as something [you can set the action button](https://support.apple.com/guide/iphone/use-and-customize-the-action-button-iphe89d61d66/ios) to. (I don't have any devices with an action button - this will be tested on other team members' phones after it is merged.)

## Discussion

While these 3 use cases are fairly minimal, this should allow us to make a proper widget with little additional work. That is going to be a future PR.

This PR adds a new extension to our client, and all the logging and such that that would entail. I also added some new extensions to share code between the main client and the widget extension. 

I could have reused a bit more code (feedback when turning on the VPN, for instance), but that would have involved restructuring even more code, adding more utility classes, etc. For a few small things (like this feedback), we have a few lines of identical code as I thought the overhead wasn't worth it. I think I struck the right balance between sharing code with this new extension and not re-architecting too much of our iOS client, but if you want me to go harder on shared code between targets, just say the word - I love nothing more than making this code Swift-ier.

## Reference

VPN-7542

## Checklist
    

flodolo (commented):
This one might be problematic, because of English…

What if a language doesn't have a perfect match for the concept of "toggle/switch"? As in a single, general-purpose verb that indicates flipping between two states without additional context.

How strongly do we feel about having this?

mcleinman:
@flodolo Flipping the order of these questions:

> How strongly do we feel about having this?

I feel fairly strongly that we need this. This work is also what handles the forthcoming widget work. In my mind, the widget is a big part of what we've been working towards with these PRs.

> What if a language doesn't have a perfect match for the concept of "toggle/switch"? As in a single, general-purpose verb that indicates flipping between two states without additional context.

Would something like "change Mozilla VPN status" work for all languages? I'm up for adding more options in English - "Flip %@" could also be added, though doesn't get away from the "one word" thing. Perhaps also "

Frustratingly, I'd imagine the actual words are hardly going to be seen for this one. While users are able to access it, I'd imagine it will rarely be selected (compared to "Siri turn on/off...", etc.) for use by a user. Instead, it powers the switches in several places in the OS - but in those cases, the actual language is never actually seen.

mcleinman:
I'm hoping that failing iOS build will be fixed by updating the Xcode version being used on TaskCluster to match what is on my machine, where it is building just fine: https://mozilla-hub.atlassian.net/browse/RELOPS-2350

flodolo:
> Would something like "change Mozilla VPN status" work for all languages? I'm up for adding more options in English - "Flip %@" could also be added, though doesn't get away from the "one word" thing. Perhaps also "

Counter question/argument: how discoverable are the sentences used to trigger a command?

In Italian — and several other Romance languages — "Toggle X" would be "Enable/disable X". Nobody in their right mind would say this for a command. 

We're going to create something that works for English, but will hardly ever work for other languages.

mcleinman:
> > Would something like "change Mozilla VPN status" work for all languages? I'm up for adding more options in English - "Flip %@" could also be added, though doesn't get away from the "one word" thing. Perhaps also "
> 
> Counter question/argument: how discoverable are the sentences used to trigger a command?
> 
> In Italian — and several other Romance languages — "Toggle X" would be "Enable/disable X". Nobody in their right mind would say this for a command.
> 
> We're going to create something that works for English, but will hardly ever work for other languages.

I'm not sure they're very discoverable - especially for this one.

As a user, I'm much more likely to say "Turn on Mozilla VPN" or "Deactivate Mozilla VPN" rather than "Toggle VPN". Given the other 3 intents we're offering, I don't see this one being used much, except behind the scenes in this widget control and the future widget.

And I just confirmed we don't *need* to provide Siri phrases for all intents. Thus, I've now removed them - and this intent will only be used by widgets. I think that solves our issue.

flodolo (approved):

oskirby (approved):
I don't see anything amiss in here, looks good.

I'm not particularly good at Swift though, so there's a good chance I've missed something.

mcleinman:
I've confirmed that the failing iOS build is TaskCluster-specific - as this branch built just fine on Xcode Cloud. I'm going to merge this even with the failing test.

PR #11239, opened 2026-04-29
Nit: Remove uneeded cli arg

strseb:
## Description
I don't see that this is used anywhere, so let's keep it for debug builds.


oskirby (approved):
LGTM

mcleinman (approved):

PR #11245, opened 2026-04-30
VPN-7532: Remove glean

mcleinman:
## Description

Removes Glean. A couple other things - namely Sentry and a DNT header - are dependent on the setting, so the actual setting is being kept. 

What do we think about the strings (see screenshots below)? This PR doesn't change them yet, but we probably could change it to just say "crash reports" (removing "diagnostics" and "product usage") to keep it clearer. Of course, if we updated the string and then changed things again in the future to add some more telemetry back in, we'd then need to get explicit buy-in from users for telemetry - and so it may be simpler to keep the language we currently are using. Thoughts?

IMG_0046
IMG_0045

## Reference

VPN-7532

## Checklist
    

strseb:
>  just say "crash reports" (removing "diagnostics" and "product usage") to keep it clearer. 

Let's keep them both for a simple reason on how we gate sentry. 

if there is a crash dump, there is an extra consent screen anyway given this is extra sensitive. 
Otherwise if enabled, sentry still sends a ping with the OS info to us, that implies product usage and probably diagnostics.

strseb:
The glean submodule needs to go too, no?

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreBinder.kt
```diff
@@ -17,7 +17,6 @@ open class CoreBinder : Binder() {
         const val setStartOnBoot = 15
         const val reactivate = 16
         const val clearStorage = 17
-        const val setGleanUploadEnabled = 18
```
strseb, line 20: Can we either move all below that -1 or make a comment // 18 == currently unused? 

same for requestGleanUploadEnabledState

linux/flatpak/org.mozilla.vpn.yml
```diff
@@ -21,7 +21,7 @@ cleanup:
   - /share/man
   - /share/doc
 modules:
-  # Dependency for jsonschema and glean-parser, needs to be installed as a
+  # Dependency for jsonschema, needs to be installed as a
   # binary wheel because it has a bunch of complex rust dependencies that are
   # difficult to replicate offline.
   - name: python3-rpds
```
oskirby: I believe we can drop `python3-rpds` too as that was only used to make the glean binding generation go.

oskirby: These particular dependencies were popped out of `flatpak-glean-parser.yaml` because of the pain in building hybrid python/rust packages inside a flatpak.

oskirby (approved):
Image

linux/flatpak/org.mozilla.vpn.yml
mcleinman: Good call, removed this. (And swapped in YAML, which we needed as it had been a dependency's dependency.)

mcleinman:
> The glean submodule needs to go too, no?

good catch. almost missed the big part. it's now gone.

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreBinder.kt
mcleinman: How would you feel if I did this in a follow-on PR? There are a *bunch* of  unused numbers from prior changes. I'd like to reorder them all (instead of just this one), but that feels like possibility for a minor error to slip in, and I'd love to break that work out from this PR. (Additionally, it seems like there were some that were removed from one file but not another - for instance, `ACTION_SET_NOTIFICATION_FALLBACK` is in `androidvpnactivity.h` as 9, but doesn't exist in `CoreBinder.kt`. I'd like to standardize these as part of the work.)

linux/flatpak/org.mozilla.vpn.yml
```diff
@@ -21,30 +21,28 @@ cleanup:
...
   # binary wheel because it has a bunch of complex rust dependencies that are
   # difficult to replicate offline.
-  - name: python3-rpds
+  - name: python3-pyyaml
     buildsystem: simple
     build-commands:
       - pip3 install --verbose --exists-action=i --no-index --find-links="file://${PWD}"
-        --prefix=${FLATPAK_DEST} "rpds-py" --no-build-isolation
+        --prefix=${FLATPAK_DEST} "PyYAML" --no-build-isolation
     sources:
```
oskirby, line 29: `PyYAML` should be able to build from source, so this could be:

```
  sources:
    - type: file
      url: https://files.pythonhosted.org/packages/54/ed/79a089b6be93607fa5cdaedf301d7dfb23af5f25c398d5ead2525b063e17/pyyaml-6.0.2.tar.gz
      sha256: d584d9ec91ad65861cc08d42e834324ef890a082e591037abe114850ff7bbc3e
```

And then we can remove the `only_arches` shenanigans.

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreBinder.kt
strseb: Totally fine! Can you open a task about it so we dont forget it? :)

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreApplication.kt
```diff
@@ -11,7 +11,7 @@ import org.qtproject.qt.android.QtApplicationBase
  * CoreApplication - extends QtApplication
  */
 class CoreApplication : org.qtproject.qt.android.QtApplicationBase(), Configuration.Provider {
```
strseb: You can remote the  `, Configuration.Provider` and and conf getter. 
Glean was the only reason we imported WorkerManager

(that should fix the android build c: )

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreApplication.kt
strseb: This can go, also i think we can yeet that dependency, there should be no other users.

strseb (approved):
Image

android/common/src/main/java/org/mozilla/firefox/qt/common/CoreBinder.kt
mcleinman: Done: https://mozilla-hub.atlassian.net/browse/VPN-7593

scripts/android/cmake.sh
```diff
@@ -10,13 +10,6 @@ if [ -f .env ]; then
...
 
-cleanup() {
-  if [ -f "$WORKSPACE_ROOT/3rdparty/glean/glean-core/uniffi.toml.backup" ]; then
-    rm -f $WORKSPACE_ROOT/3rdparty/glean/glean-core/uniffi.toml
-    mv $WORKSPACE_ROOT/3rdparty/glean/glean-core/uniffi.toml.backup $WORKSPACE_ROOT/3rdparty/glean/glean-core/uniffi.toml
-  fi
-}
-
 cleanup_and_die() {
   cleanup
```
strseb: All callers of cleanup need to be removed :)

PR #11248, opened 2026-05-04
Spring-Cleanup: Remove App Parent Class

strseb:
## Description

The App class originally was added as we wanted to have multiple apps with a shared base of logic  (i.e authentification, subscription) - that abstraction however was never removed when we shelved the idea. So we have only one Implementer. Sooo this abstraction is no longer needed.


src/ui/screens/initialize/ViewInitialize.qml
```diff
@@ -274,7 +274,7 @@ Item {
           delegate: ColumnLayout {
               visible: false
               width: parent.width
-              spacing: MZTheme.the5me.windowMargin / 2
+              spacing: MZTheme.theme.windowMargin / 2
```
mcleinman, line 277: I'm so so so embarrassed. Thanks

src/mozillavpn.h
```diff
@@ -192,6 +232,30 @@ class MozillaVPN final : public App {
...
+  void aboutNeeded();
+  void updatingChanged();
+  void accountDeleted();
+
+  void authenticationStarted();
+  void authenticationAborted();
+  void authenticationCompleted();
+
+  // For Glean
+  void initializeGlean();
```
mcleinman: I believe the Glean stuff should be removed at this point.

tests/unit/testmodels.cpp
```diff
@@ -448,7 +448,9 @@ void TestModels::deviceModelFromJson() {
 
       if (devices > 0) {
         QVERIFY(dm.deviceFromPublicKey(devicePublicKey.toString()) != nullptr);
-        QVERIFY(dm.deviceFromUniqueId() != nullptr);
+        if (!Device::uniqueDeviceId().isEmpty()) {
```
mcleinman, line 451: Why were the changes in this file needed, out of curiosity?

src/mozillavpn.cpp
```diff
@@ -130,12 +135,48 @@ MozillaVPN* MozillaVPN::instance() {
...
+MozillaVPN::MozillaVPN()
+    : QObject(nullptr), m_private(new MozillaVPNPrivate()) {
   MZ_COUNT_CTOR(MozillaVPN);
 
+#ifdef MZ_WASM
+  NetworkRequest::setRequestHandler(WasmNetworkRequest::deleteResource,
+                                    WasmNetworkRequest::getResource,
+                                    WasmNetworkRequest::postResource,
+                                    WasmNetworkRequest::postResourceIODevice);
+#endif
```
mcleinman, line 178: This section is new, right? Why was it added?

mcleinman (approved):
Few questions, but looks great. Thanks.

tests/unit/testmodels.cpp
strseb: tldr; local wierd linux quirks. 

Moved finally from windows 11 on my dev machine to linux. Somehow qt cant find uniqueDeviceId on that device, hence is empty. then in deviceFromUniqueId we get a nullptr (ofc, what would be the device of no id).

src/mozillavpn.h
strseb: Ah right!

src/mozillavpn.cpp
strseb: That actually comes from app.cpp. 

I think those essentially "mock" the network requests for wasm.

PR #11249, opened 2026-05-04
Nit: Update Linter Clang Version

strseb:
## Description

Looking at our env, dev's are getting clang-20. It makes different formatting decisions than clang-16 soo it's sometimes annoying to unify those assumptions.


encrypt94:
looks like the code needs a quick round of linting with clang-20 :)

mcleinman (approved):

PR #11252, opened 2026-05-05
VPN-7560 2.37 addons

mcleinman:
## Description

Update and what's new messages for 2.37 cycle.

Bullet points 1, 2, 4, and 6 may not be ready in time. If they are not, we'll pull them from the addon (but keep them in `strings.yaml`) and use them in the next release. (The question around bullet 6 is why it's not combined with bullet 5 - want to be able to remove just bullet 6 if needed.)

This will be the first time users will see a promo message on the main screen.
Screenshot 99Screenshot 100

Screenshot 101Screenshot 102

## Reference

VPN-7560

## Checklist
    

flodolo (approved):

addons/strings.yaml
```diff
@@ -199,6 +199,32 @@ commonStrings:
...
+    value: "[iOS] Add a switch for Mozilla VPN to Control Center or the lock screen. And on the latest iPhones, it’s even possible to choose Mozilla VPN to be the Action button function."
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to iOS platform.
+  bullet5:
+    value: "[macOS] IPv6-only networks are now supported."
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to macOS platform.
+  bullet6:
+    value: "[Windows] IPv6-only networks are now supported."
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to Windows platform.
+  bullet7:
+    value: "[Linux] Use GNOME+Wayland? Window decoration is back, so Mozilla VPN stands out from the background."
```
oskirby: I think there is a grammatical error here, it should read:
> Use GNOME+Wayland? Window decorations **are** back, so Mozilla VPN stands out from the background.

addons/strings.yaml
```diff
@@ -199,6 +199,32 @@ commonStrings:
     value: "[Linux] Ubuntu users may have had an issue when using Firefox Multi-Account Containers with custom VPN locations. This is no longer a problem."
     comment: Bullet point with a specific update. Should have tag that indicates it applies to Linux platform.
 
+237updateMessage:
+  bullet1:
+    value: "[macOS] App exclusions are now available for macOS. Want to let some programs bypass the VPN completely? You now can."
```
oskirby, line 204: This is very unlikely to be ready for users in 2.37, but it will ship behind a feature flag in this release.

oskirby (commented):
Looks good, just one minor grammar nit.

addons/strings.yaml
flodolo: If VPN is a single window, decoration makes sense?

addons/strings.yaml
mcleinman: Sounds good. Let's leave this in, as it's trivial to remove the bullet from this release and use the string next time.

addons/strings.yaml
mcleinman: "Window decoration are back" or "Window decorations are back"? I don't know what these are canonically called, so deferring to your Linux expertise on whether `decoration` can be plural here.

oskirby: I have always understood the terminology to be "window decorations", as in a singular window can have multiple decorations applied to it.

mcleinman: Sounds good, updated to "window decorations are back". Thanks.

mcleinman:
I've rerun this several times. One failure each time, but at least 3 different tests (in the same file) are the failure. I am pretty confident it is not my PR causing this failure. Given the translation deadline we're trying to hit, I'm going to merge this even with the failed test. I'll monitor further CI tests to see if this is systemic.

PR #11261, opened 2026-05-08
VPN-5290 - Add a 'never' option to reccomended serverlist

strseb:
## Description

Rn we do not handle the case if the QDateTime is invalid, like when we never updated the serverlist. 
 
image

### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -94,6 +94,12 @@ serversView:
     value: "Last updated: %1. To update this list please disconnect from the VPN."
     comment: The time (or in extremely rare circumstances, the date, day of the week, or "yesterday") which the recommended servers list was last updated and instructions to re-enable the refresh button. Argument is the time or date/day
   recommendedRefreshLastUpdatedLabelYesterday: yesterday
+  recommendedRefreshNoUpdateLabel:
+    value: Tap to update
+    comment: Label next to the refresh button shown when the recommended list has never been updated yet. The whole row is clickable and tapping triggers an update.
+  recommendedRefreshNoUpdateDisabledLabel:
+    value: Disconnect from the VPN to update this list.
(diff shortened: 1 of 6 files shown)


flodolo (approved):

src/translations/strings.yaml
```diff
@@ -94,6 +94,9 @@ serversView:
     value: "Last updated: %1. To update this list please disconnect from the VPN."
     comment: The time (or in extremely rare circumstances, the date, day of the week, or "yesterday") which the recommended servers list was last updated and instructions to re-enable the refresh button. Argument is the time or date/day
   recommendedRefreshLastUpdatedLabelYesterday: yesterday
+  recommendedRefreshLastUpdatedLabelNever:
+    value: never
+    cooment: In case the recommended list has not yet been updated.
```
encrypt94: little typo in the key

encrypt94 (approved):
lgtm!

src/translations/strings.yaml
flodolo: Good catch. We should probably have scripts yelling when an unknown attribute is added to a key?

strseb: Good catch! Added this to the parser and actually found another typo :D

mcleinman:
Could we change the visibility of this field instead?

strseb:
> Could we change the visibility of this field instead?

Then we would either have the "retry" button without context, or we'd hide the whole line removing the option to manually trigger the task.

mcleinman:
> > Could we change the visibility of this field instead?
> 
> Then we would either have the "retry" button without context, or we'd hide the whole line removing the option to manually trigger the task.

What about changing the whole string to one that just says "Update" or "Refresh" next to that button? That seems like a better experience for users.

strseb:
@mcleinman  what about this? 

image

flodolo (approved):

mcleinman (approved):
Looks great. Thanks!

encrypt94 (approved):

PR #11265, opened 2026-05-11
VPN-7603: fix translations

mcleinman:
## Description

Siri suggestions were showing keys. Shortcuts app was sometimes showing the wrong language.

I *believe* this fixes it, or at least gets us close. We're into some real dark arts of iOS at this point - how the app offers intents to the system. The details are not offered by Apple, and when building the iOS app it may parse those pieces out as part of the build process - hopefully before we run our translations script. The internet disagrees on some of these details, and AI doesn't know anything definitive either (though it will lead you down bad paths for as long as you'll let it). Further making things difficult, it caches this stuff hard - seems like rebooting the device is the only way to ensure you've cleared cache when switching languages. The behavior I've seen in testing has been inconsistent.

There may be more work to do on this in the future, but this at least gets us closer for now.

## Reference

VPN-7603

## Checklist
    
### DIFF
--- a/scripts/utils/generate_xcstrings.py
+++ b/scripts/utils/generate_xcstrings.py
@@ -76,9 +76,13 @@ def get_locales(i18n_dir):
 
 def build_localizable_xcstrings(main_strings, locale_translations):
     """Build Localizable.xcstrings dict for iosAppIntents.main.* strings."""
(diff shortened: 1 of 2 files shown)


oskirby (approved):
LGTM

PR #11271, opened 2026-05-12
VPN-7608: app intent translation fixes

mcleinman:
## Description

Addresses 3 things: 
1) Poorly escaped string (thanks for finding that in the logs)
2) Some translations messed up and put multiple phrases on one line. I'm working w/ l10n team on this, but also let's check for this when building.
3) I was surprised that we were missing several languages in the iOS app. From looking into it, it seems like they're just locales that haven't translated these strings yet. Let's add some logging around that.

Important: The second thing will cause iOS builds to fail once this is merged, as there are messed up translations currently. (I've already fixed the ones I could in pontoon, and asked l10n team about a couple others - so hopefully this will be fixed shortly.) Failing builds seems okay: We do not want to go to prod w/ the current state. TaskCluster is already failing (so no change), and it will cause Xcode Cloud to fail for a bit. For local builds, we'll just need to comment out a line for the time being. But it's important to catch these malformed translations, and this hammer seems like the best way to make sure we do in the future (IMO).

## Reference

VPN-7608

## Checklist
    
### DIFF
--- a/scripts/utils/generate_xcstrings.py
+++ b/scripts/utils/generate_xcstrings.py
@@ -142,6 +142,8 @@ def build_phrase_section(phrase_strings, locale_translations):
                     "values": locale_values,
                 }
             }
+        else:
+            print(f"No translations found for {string_id} in {locale}")
 
     return {"localizations": localizations}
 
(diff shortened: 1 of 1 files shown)


oskirby (approved):
LGTM

PR #11272, opened 2026-05-12
VPN-7604: wait for confirmation when activating from intent

mcleinman:
## Description

Just because I couldn't find a path that caused an issue when optimistically returning success, doesn't mean QA couldn't. This optimistic return fails in the case lack of network. (However, it did set the VPN to auto-connect the next time there was a network.) Now we give a 3 second timeout - return success when connected, and return error if the timeout ends. Three seconds seems like enough time to set up the connection without false postives on the timeout, but I'm open to discussion.

## Reference

VPN-7604

## Checklist
    

artfwo (approved):
I'd say 3 seconds is a bit tight on a slow network and would increase the timeout (maybe to 5?).

I also wonder if it's possible to avoid polling here? E.g. could we subscribe to controller's connected signal instead?

PR #11275, opened 2026-05-13
Add Obfuscation features support and udp over tcp obfuscator on Linux and Android

encrypt94:
## Description

This PR adds support for obfuscation features and udp over tcp obfuscators.
A new "Obfuscation" option in "Privacy Features" , allows users to configure the following obfuscation features when available on the platform (the menu item will be hidden if a feature is not available.)
- Always connect using port 53 - not an obfuscation method, simply exposes an existing feature and allows the user to always use port 53 (compatible with LWO)
- UDP over TCP - the only obfuscator implemented in  this pr, uses udp over tcp crate to tunnel WireGuard traffic over TCP
- LWO - lighweight WireGuard scrambler, planned as the next obfuscator
- LWO over port 53 - lighweight WireGuard scrambler, planned as the next obfuscator
- MASQUE - tunnel WireGuard traffic over `CONNECT-UDP`
- Shadowsocks - tunnel WireGuard traffic over shadowsocks

This menu lets users select the `ObfuscationPolicy` to use. An obfuscation policy can specify a port selection, an obfuscation method, or a combination of both.

`SettingsHolder::ObfuscationPolicy` is converted to `Server::ObfuscationMethod` in `Controller` and `ServerData`. This allows the server selection logic to be obfuscation method aware.

When obfuscation is turned on or changed, servers that were previously unreachable may become available again so this action triggers a reset on all server in cooldown.

In the unlikely event that no server in a city matches the selected obfuscation configuration, a "Server Unavailable" dialog is displayed.

When a server is not available the modal will propose user to enable obfuscation or try a different obfuscation method.  

## Obfuscators
Obfuscators are implemented in a Rust crate located under the `obfuscators/` directory.
An obfuscator acts as a local UDP proxy running on a local UDP port, the outbound sockets must be protected to avoid traffic loops.

On Linux (and Windows in the future except for the marking part):
- The crate produces a standalone binary (`mozilla-obfuscator`) installed alongside the daemon.
- When the VPN is turned on or a server switch occurs, the obfuscator is launched.
- The obfuscator requires `CAP_NET_ADMIN` privileges to set `SO_MARK`.
- The obfuscator listens on a free port and writes the port to stdout.
- The MozillaVPN daemon parses the port from stdout and rewrites the server/port in WireGuard's peer config (alternatively, we can probe a free port in the daemon and send it to the obfuscator but this approach may be prone to race contitions)

On Android (and likely iOS in a similar way):
- The crate is accessed via FFI using JNA.
- `VPNActivity` starts the obfuscator.
- `VPNActivity` calls `protect` on the obfuscator's sockets.
- `VPNActivity` rewrites the local port in the configuration.

This architecture was chosen because most of the libraries required to implement obfuscation protocols are available in the Rust ecosystem: UDP-over-TCP, Shadowsocks, and QUIC stacks.

Impacts on the rust build bits:
- added `add_rust_binary` in `rustlang.cmake` and refactored `add_rust_library`  to split out shared parts
- updated rust to 1.91 because:
  - `udp-over-tcp` requires rust >= 1.87 
  - only rust 1.89 and 1.91 are available in Ubuntu repost
  - i hit a weird bug with cargo 1.8x installed from ubuntu repos being unable to install crates from git (solved in 1.91)
- rust 1.91 is not available in `debian:bookworm` so is now installed using rustup in `linux-qt6-build` 

## UI
When Obfuscation features are enabled, a “Obfuscation is on” label appears below the timer.
Screenshot from 2026-07-16 16-51-32
A new Obfuscation menu is available in Settings > Privacy Features.
menu

When Anti-Censorship features are available on the platform, a radio will be displayed in the menu with name and brief description
Screenshot from 2026-07-16 16-59-10
Screenshot from 2026-07-16 16-59-19

Disabled
Screenshot from 2026-07-16 17-15-37
Screenshot from 2026-07-16 17-15-41

Help text
Screenshot from 2026-07-16 17-17-48

Server unavailable dialog and notification show when obfuscation is enabled
sun2
sun2n

Server unavailable dialog and notification show when obfuscation is not enabled
sun1
sun1n

## Reference

[VPN-7588 - Create generic obfuscation interface for anti censorship features](https://mozilla-hub.atlassian.net/browse/VPN-7588)
[VPN-7566 - Create anti censorship features menu](https://mozilla-hub.atlassian.net/browse/VPN-7566)
[VPN-7582 - Implement UDP over TCP on Linux](https://mozilla-hub.atlassian.net/browse/VPN-7582)
[VPN-7583 - Implement UDP over TCP on Android](https://mozilla-hub.atlassian.net/browse/VPN-7583)

## Checklist
    

oskirby:
I've had kind of a love/hate relationship with building rust binaries over the years, but this might be a good case for splitting the obfuscator into a standalone binary that can be invoked by the daemon as needed rather than baking it into the client.

The obfuscator should only ever operate on unprivileged information, should require no root permissions, and it probably has a pretty heavy workload given that it needs to live in the packet processing loop.

```C++
QProcess * obfuscator = new QProcess(this);
QStringList args;
args << "--method" << "lwo";
args << "--pubkey" << config.m_serverPublicKey;
args << "--server" << config.m_serverIpv4AddrIn;
obfuscator->start("/usr/bin/mozillavpn-obfuscator", args);

connect(obfuscator, &QProcess::started, this, &Example::handleObfuscatorStarted);
connect(obfuscator, &QProcess::finished, this, &Example::handleObfuscatorFinished);
```

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
 
 settings:
   privacySettings: Privacy features
+  antiCensorshipSettings: Anti-Censorship features
```
flodolo: Is this coming from the Content Team? Is this terminology approved by Legal?

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
+  antiCensorshipSettingsWarning: These options may cause connection stability issues.
+  antiCensorshipPort53Title: Connect using port 53
+  antiCensorshipPort53Body: This may help on networks that block ports or UDP traffic.
+  antiCensorshipMasqueTitle: MASQUE
```
flodolo: Why is this uppercase? Should it be translated? Same question for SHADOWSOCKS later.

flodolo (changes requested):

src/translations/strings.yaml
mcleinman: Similar question - I'm curious if you'd be open to working on wordsmithing a bunch of these. "What can xxxx do for me?", for instance, feels like a different voice than most of our product copy.

src/ui/screens/settings/antiCensorship/AntiCensorshipFeaturesList.qml
```diff
@@ -0,0 +1,78 @@
...
+            dividerTopMargin: dividerSpacing
+            visible: MZFeatureList.get(modelData.objectName).isSupported
+            showDivider: !isOnboarding || index + 1 !== repeater.model.length
+
+            onClicked: {
+                if(MZSettings.obfuscationMethod == modelData.settingValue) {
+                    MZSettings.obfuscationMethod = MZSettings.NoObfuscation;
+                } else {
+                    MZSettings.obfuscationMethod = modelData.settingValue;
+                    // Maybe automatically disable alwaysPort53 when enabling an obfuscation method
```
mcleinman: If these options are mutually exclusive, we should not use switches here - we should use radio buttons, like in the DNS settings screen.

mcleinman:
This is amazing, thank you! I'm focusing on UI right now, will let others handle the underlying code who are better positioned to review that piece.

src/ui/screens/settings/antiCensorship/AntiCensorshipFeaturesList.qml
encrypt94: Good point. Obfuscators are mutually exclusive, and "Always use port 53" only works with LWO. What do you think about using a toggle switch for "Always use port 53" and radio buttons for the obfuscators, and then disabling the switch when anything other than LWO is selected?

mcleinman: What about splitting it into two radio buttons? Something like "LWO via typical ports" and "LWO using port 53"? We don't need to touch the settings structure in code - under the hood, we just set those settings as appropriate.

src/translations/strings.yaml
encrypt94: > Is this coming from the Content Team? Is this terminology approved by Legal?
Thanks for raising this. It’s not coming from the Content Team and hasn’t been approved by Legal yet. Legal is currently reviewing the language.

> Similar question - I'm curious if you'd be open to working on wordsmithing a bunch of these. "What can xxxx do for me?", for instance, feels like a different voice than most of our product copy.

Sure i'm totally open to change the descriptions! i'll try to come up with something that feels more aligned with the rest of our product copy :) suggestions are super appreciated!

src/translations/strings.yaml
encrypt94: Shadowsocks casing is an error. MASQUE is an acronym, so it should be kept uppercase. Both of them are proper nouns and should not be translated

src/translations/strings.yaml
mcleinman: I think this needs to be done in conjunction with legal, but I'm wondering if "anti-censorship features" is immediately understandable by the average person.

Maybe something like "Advanced networking features" or "restricted network settings" or "alternative connection methods" or "hostile network defense" or "censorship resistance" or "network workaround" or... maybe this is something we can talk out in a meeting next week?

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
 
 settings:
   privacySettings: Privacy features
+  antiCensorshipSettings: Anti-Censorship features
+  antiCensorshipSettingsWarning: These options may cause connection stability issues.
```
mcleinman:
```suggestion
  antiCensorshipSettingsWarning: These options should be used when a network operator is blocking connection to VPN servers. They may cause stability issues.
```

mcleinman: Should we also include something about how this will not help with a website blocking VPN connections? This seems important for setting user expectations.

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
+  antiCensorshipPort53Title: Connect using port 53
+  antiCensorshipPort53Body: This may help on networks that block ports or UDP traffic.
```
mcleinman:
```suggestion
  antiCensorshipPort53Body: This may help on networks that block specific ports or UDP traffic.
```

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
+  antiCensorshipMasqueTitle: MASQUE
+  antiCensorshipMasqueBody: This may help on networks that block VPN protocols by disguising VPN traffic as regular HTTPS traffic.
```
mcleinman:
```suggestion
  antiCensorshipMasqueBody: This disguises VPN traffic as regular HTTPS traffic, and may help on networks that block VPN protocols by .
```

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
 
 settings:
   privacySettings: Privacy features
+  antiCensorshipSettings: Anti-Censorship features
+  antiCensorshipSettingsWarning: These options may cause connection stability issues.
+  antiCensorshipPort53Title: Connect using port 53
```
mcleinman:
```suggestion
  antiCensorshipPort53Title: Always connect using port 53
```

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
+  antiCensorshipPort53Body: This may help on networks that block ports or UDP traffic.
+  antiCensorshipMasqueTitle: MASQUE
+  antiCensorshipMasqueBody: This may help on networks that block VPN protocols by disguising VPN traffic as regular HTTPS traffic.
+  antiCensorshipLwoTitle: Lightweight obfuscation
+  antiCensorshipLwoBody: The fastest obfuscation method. May help on networks that detect or block WireGuard traffic without adding significant overhead.
+  antiCensorshipShadowsocksTitle: SHADOWSOCKS
+  antiCensorshipShadowsocksBody: A reliable censorship circumvention protocol that may help bypass restrictive or heavily filtered networks.
+  antiCensorshipUdpOverTcpTitle: UDP over TCP
+  antiCensorshipUdpOverTcpBody: This may help on networks that block UDP traffic by tunneling it over TCP.
```
mcleinman:
```suggestion
  antiCensorshipUdpOverTcpBody: This tunnels UDP traffic over TCP, and may help on networks that block UDP traffic.
```

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
 settings:
   privacySettings: Privacy features
+  antiCensorshipSettings: Anti-Censorship features
+  antiCensorshipSettingsWarning: These options may cause connection stability issues.
+  antiCensorshipPort53Title: Connect using port 53
+  antiCensorshipPort53Body: This may help on networks that block ports or UDP traffic.
+  antiCensorshipMasqueTitle: MASQUE
+  antiCensorshipMasqueBody: This may help on networks that block VPN protocols by disguising VPN traffic as regular HTTPS traffic.
+  antiCensorshipLwoTitle: Lightweight obfuscation
+  antiCensorshipLwoBody: The fastest obfuscation method. May help on networks that detect or block WireGuard traffic without adding significant overhead.
```
mcleinman: Can we give a little bit more here? We give a minor technical explanation for the others, and this once we just call "lightweight" with no detail.

src/translations/strings.yaml
```diff
@@ -329,6 +329,18 @@ systray:
...
+  antiCensorshipShadowsocksTitle: SHADOWSOCKS
+  antiCensorshipShadowsocksBody: A reliable censorship circumvention protocol that may help bypass restrictive or heavily filtered networks.
```
mcleinman:
```suggestion
  antiCensorshipShadowsocksBody: [Insert a phrase of technical detail.] This may help bypass restrictive or heavily filtered networks, and sometimes works when all other ones do not.
```

Are there tradeoffs to this, like in speed? Is this the first one folks should try, or the last?

src/translations/strings.yaml
```diff
@@ -880,6 +892,18 @@ devices:
...
+    value: Anti-Censorship features
+    comment: Title label for the Anti-Censorship features help sheet
+  antiCensorshipHeader:
+    value: What can Anti-Censorship features do for me?
+    comment: Header label for the Anti-Censorship features help sheet
+  antiCensorshipBody1:
+    value: Anti-Censorship features can help when your provider or network blocks VPN or UDP traffic.
+    comment: Body label for the Anti-Censorship features help sheet
+  antiCensorshipBody2:
+    value: These features may slow down your connection, so turn them off when you no longer need them.
```
mcleinman:
```suggestion
    value: These features may slow down the connection, so turn them off when they are not needed.
```

src/translations/strings.yaml
```diff
@@ -880,6 +892,18 @@ devices:
     comment: Title for the menu bar when viewing the devices or device limit pages
 
 helpSheets:
+  antiCensorshipTitle:
+    value: Anti-Censorship features
+    comment: Title label for the Anti-Censorship features help sheet
+  antiCensorshipHeader:
+    value: What can Anti-Censorship features do for me?
```
mcleinman: Maybe remove this header? I'm not sure it adds anything, given that we have the title as well.

src/translations/strings.yaml
mcleinman: And of course, if we rename this we'll need to update many of these strings.

src/platforms/linux/daemon/dbusservice.h
```diff
@@ -43,6 +43,7 @@ class DBusService final : public Daemon, protected QDBusContext {
   WireguardUtils* wgutils() const override { return m_wgutils; }
   IPUtils* iputils() override;
   DnsUtils* dnsutils() override;
+  void markObfuscatorSockets(int v4, int v6);
```
artfwo: Can't find where is this method called, am I missing anything?

src/daemon/daemon.cpp
```diff
@@ -409,8 +469,25 @@ bool Daemon::switchServer(const InterfaceConfig& config) {
...
+  if (config.m_obfuscationMethod != Server::ObfuscationMethod::NoObfuscation &&
+      config.m_hopType != InterfaceConfig::MultiHopExit) {
+    obfuscator = createObfuscator(config);
+    if (!obfuscator->start()) {
+      logger.error() << "Failed to start obfuscator on switch"
+                     << config.m_obfuscationMethod;
+      return false;
+    }
+    peerConfig.m_serverIpv4AddrIn = "127.0.0.1";
+    peerConfig.m_serverIpv6AddrIn = QString();
```
artfwo, line 491: Why is this empty string? Should this be `::1` instead as in `activate()` above?

src/daemon/obfuscator/qprocessobfuscator.cpp
```diff
@@ -0,0 +1,130 @@
...
+#endif
+  // The obfuscator needs to bind to the same interface as the WireGuard peer
+  return args;
+}
+
+QString QProcessObfuscator::binaryName() const {
+#if defined(MZ_WINDOWS)
+  return QStringLiteral("mozillavpn-obfuscator.exe");
+#elif defined(MZ_LINUX)
+  return QStringLiteral("mozillavpn-obfuscator");
```
artfwo, line 123: The called process may be resolved to anything when PATH is modified. Looks like a mild concern given that the daemon drops the privileges now, but the called binary can still modify network configuration.

src/platforms/linux/daemon/dbusservice.h
encrypt94: It was dead code thanks for spotting!

src/daemon/daemon.cpp
encrypt94: good catch! it's the opposite :)  m_serverIpv6AddrIn should always be empty as the obfuscator only listens on 127.0.0.1

src/translations/strings.yaml
flodolo: This information should be covered in comments for each string.

Also, unless I missed a reply, my general question about `Anti-censorship` and involving the Content Team or Product still stands. To me, that feels like a slippery slope term.

encrypt94: Comments added! The strings and language were reviewed by Legal and Policy and received a thumbs-up. Sorry, i forgot to post an update in reply to your previous comment.

src/translations/strings.yaml
encrypt94: Legal and Policy gave it a thumbs-up, aaand I just realized my first reply ended up under the quote :facepalm:

src/translations/strings.yaml
```diff
@@ -328,6 +328,32 @@ systray:
...
   privacySettings: Privacy features
+  antiCensorshipSettings: Anti-censorship features
+  antiCensorshipSettingsWarning1: These features will not help when a specific website or app blocks content because of the VPN.
+  antiCensorshipSettingsWarning2: They may cause degraded network speed.
+  antiCensorshipSettingsEnable: Enable Anti-censorship features
+  antiCensorshipPort53Title: Always connect using port 53
+  antiCensorshipPort53Body: Routes traffic through the port commonly used for DNS. May help on networks that block UDP traffic.
+  antiCensorshipMasqueTitle:
+    value: MASQUE
+    comment: MASQUE is an acronym it should NOT be translated and should be written in all caps.
```
```suggestion
    comment: MASQUE is an acronym, it should NOT be translated.
```

The last part is implicit. Same for similar comments.

src/translations/strings.yaml
```diff
@@ -328,6 +328,32 @@ systray:
...
+  antiCensorshipMasqueBody: Disguises VPN traffic as regular HTTPS traffic. May help on networks that block VPN protocols.
+  antiCensorshipLwoTitle:
+    value: LWO
+    comment: LWO is an acronym it should NOT be translated and should be written in all caps.
+  antiCensorshipLwoBody: Fastest obfuscation method. May help on networks that detect or block WireGuard traffic without adding significant overhead.
+  antiCensorshipShadowsocksTitle:
+    value: Shadowsockss
```
```suggestion
    value: Shadowsocks
```

src/translations/strings.yaml
```diff
@@ -328,6 +328,32 @@ systray:
...
+    comment: Shadowsocks is a protocol name it should NOT be translated.
+  antiCensorshipShadowsocksBody: May help bypass restrictive or heavily filtered networks, and sometimes works when all others do not.
```
```suggestion
  antiCensorshipShadowsocksBody: May help bypass restrictive or heavily filtered networks, and could work when all other approaches do not.
```

I think we need a noun here ("others what?").

src/translations/strings.yaml
```diff
@@ -1086,6 +1118,9 @@ controller:
   regeneratingKey:
     value: Preparing your connection
     comment: Connection subtitle label for the main screen when the app is regenerating WireGuard keys
+  antiCensorshipOn:
```
flodolo: Do we need an off version?

flodolo (approved):

src/translations/strings.yaml
encrypt94: We don't. We only need the "On" string as a reminder that the feature is enabled. In the help text, we tell users to turn it off when it's not needed because it may reduce connection speed.

src/utils/models/server.h
```diff
@@ -6,26 +6,42 @@
...
 
 #include <QList>
+#include <QObject>
 #include <QPair>
 #include <QString>
 
 class QJsonObject;
 
-class Server final {
+class Server final : public QObject {
```
artfwo: I wonder if inheriting QObject here would have any unwanted side-effects, given that Server is stored and copied by value elsewhere in the code.

src/utils/interfaceconfig.h
```diff
@@ -24,6 +25,7 @@ class InterfaceConfig {
 
   HopType m_hopType;
   QString m_privateKey;
+  QString m_publicKey;
```
artfwo, line 28: This doesn't seem to be used in obfuscator's buildArgs, is it supposed to be used elsewhere?

src/daemon/obfuscator/qprocessobfuscator.h
```diff
@@ -0,0 +1,32 @@
...
+class QProcessObfuscator final : public Obfuscator {
+ public:
+  explicit QProcessObfuscator(const InterfaceConfig& config);
+  ~QProcessObfuscator() override;
+
+  bool start() override;
+  quint16 localPort() const override { return m_localPort; }
+
+ private:
+  bool isRunning() const { return m_process.state() != QProcess::NotRunning; }
```
artfwo: Can't find where this is called. If it's not used, then it probably can be removed?

src/utils/interfaceconfig.h
encrypt94: this is needed for LWO obfuscation, which I'm currently working on. It'll be included in a follow-up PR.

src/utils/models/server.h
encrypt94: good point. Q_GADGET would be a better fit here

src/cmake/linux.cmake
```diff
@@ -104,6 +104,19 @@ if(NOT BUILD_FLATPAK)
     include(${CMAKE_SOURCE_DIR}/scripts/cmake/golang.cmake)
     add_go_library(netfilter ${CMAKE_SOURCE_DIR}/linux/netfilter/netfilter.go)
     target_link_libraries(mozillavpn PRIVATE netfilter)
+
+    # Add the obfuscator binary
```
oskirby, line 108: Is there a reason why we can't use the obfuscator on Flatpaks? Or is that work for a future PR?

src/daemon/obfuscator/qprocessobfuscator.cpp
oskirby: We could make that a little tighter by generating an absolute path based on the application's binary, along the lines of `QCoreApplication::applicationDirPath() + "/mozillavpn-obfuscator"`

taskcluster/docker/base/Dockerfile
```diff
@@ -39,14 +39,16 @@ RUN apt-get update -qq \
...
-                          cargo-1.85 \
+                          cargo-1.91 \
                           curl \
                           rpm \
                           libglib2.0-0 \
                           ccache \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+RUN ln -s /usr/bin/cargo-1.91 /usr/bin/cargo
```
oskirby: I would recommend adding `/usr/lib/rust-1.91/bin` to the `PATH` variable instead instead of linking stuff into `/usr/bin`

In fact, we already have a Debian package build workaround that kind of does this in [linux/debian/rules](https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/linux/debian/rules#L9-L13) and we might just need to update it for version 1.91

src/translations/strings.yaml
```diff
@@ -328,6 +328,33 @@ systray:
...
+    comment: MASQUE is an acronym, it should NOT be translated.
+  obfuscationMasqueBody: Disguises VPN traffic as regular HTTPS traffic. May help on networks that block VPN protocols.
+  obfuscationLwoTitle:
+    value: LWO
+    comment: LWO is an acronym, it should NOT be translated.
+  obfuscationLwoBody: Fastest obfuscation method. May help on networks that detect or block WireGuard traffic without adding significant overhead.
+  obfuscationShadowsocksTitle:
+    value: Shadowsocks
+    comment: Shadowsocks is a protocol name, it should NOT be translated.
+  obfuscationShadowsocksBody: May help bypass restrictive or heavily filtered networks, and could work when all other approaches do not.
```
oskirby: Compared to the other obfuscation methods, this text does little to explain what Shadowsocks does, it reads more as a generic recommendation to use it rather than any kind of technical explanation.

nebula/ui/components/MZInterLabel.qml
```diff
@@ -16,6 +16,7 @@ Text {
     lineHeight: MZTheme.theme.labelLineHeight
     wrapMode: Text.Wrap
     color: MZTheme.colors.fontColor
+    opacity: enabled ? 1 : 0.5
```
mcleinman: I'd rather change the color than add opacity - we've run into issues before, as it looks different depending on the background color.

We could create a `fontColorMuted`, similar to how we use `fontColorInvertedMuted`.

nebula/ui/components/MZTextBlock.qml
```diff
@@ -10,6 +10,7 @@ import Mozilla.Shared 1.0
 // MZTextBlock
 Text {
     color: MZTheme.colors.fontColor
+    opacity: enabled ? 1 : 0.5
```
mcleinman: Same as above - would prefer a color rather than opacity

src/ui/screens/settings/privacy/obfuscation/ViewObfuscation.qml
```diff
@@ -0,0 +1,103 @@
...
+
+        MZInformationCard {
+            objectName: "obfuscationSettingsViewInformationCard"
+            cardType: MZInformationCard.CardType.Warning
+            Layout.preferredWidth: Math.min(window.width - MZTheme.theme.windowMargin * 2, MZTheme.theme.navBarMaxWidth)
+            Layout.alignment: Qt.AlignHCenter
+
+            _infoContent: ColumnLayout {
+                id: textBlocks
+                spacing: 0
```
mcleinman: Maybe 5 or 10 points of spacing here? Right now it looks pretty tight.

src/ui/popups/ServerUnavailablePopup.qml
```diff
@@ -31,5 +37,15 @@ MZSimplePopup {
                 window.showServerList(true);
                 root.close();
             }
+        },
+        MZButton {
+            text: MZI18n.ServerUnavailableModalConfigureObfuscationButtonLabel
```
mcleinman, line 46: Should we only have this button visible if there is an obfuscation setting currently active?

encrypt94: It should always be visible, only the modal text should change.
If obfuscation is not enabled, the user can click "Configure obfuscation" to enable it. Otherwise, the same button can be used to change the obfuscation method.

src/cmake/linux.cmake
encrypt94: planned for a future PR!

src/daemon/obfuscator/qprocessobfuscator.cpp
encrypt94: I forgot to answer the original comment sorry, this is done [here](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11275/changes#diff-fa376f4c1cdde8060c68d132ad76c9ca81873a98c4ef7bb2ba28c6034bbfa1d1R40)

src/translations/strings.yaml
```diff
@@ -568,11 +595,28 @@ crashreporter:
 
 serverUnavailableModal:
   headerText: Server unavailable
-  bodyText: This server location is temporarily unavailable. Choose a new location.
+  bodyText1:
+    value: This server location is temporarily unavailable.
+    comment: First part of the server unavailable modal body. Followed by bodyText2NoObfuscation when obfuscation is disabled, or bodyText2Obfuscation when obfuscation is enabled.
```
flodolo: Please don't use this approach. You end up with 3 strings and poorer localizability, instead of 2 full strings.

* `This server location is temporarily unavailable. Choose a new location or try a different obfuscation method.`
* `This server location is temporarily unavailable. Choose a new location or try to enable obfuscation.`

src/translations/strings.yaml
```diff
@@ -568,11 +595,28 @@ crashreporter:
...
+    value: Choose a new location or try to enable obfuscation.
+    comment: Second part of the server unavailable modal body, shown when no obfuscation policy is selected. Follows bodyText1.
   buttonLabel: Choose a new location
+  configureObfuscationButtonLabel: Configure obfuscation
 
 serverUnavailableNotification:
-  bodyText: This server location is temporarily unavailable. Please open the app to choose a new location.
+  bodyText1:
+    value: This server location is temporarily unavailable.
+    comment: First part of the server unavailable notification. Followed by bodyText2NoObfuscation when obfuscation is disabled, or bodyText2Obfuscation when obfuscation is enabled.
```
flodolo: Same here, use 2 full sentences.

src/translations/strings.yaml
```diff
@@ -885,6 +929,15 @@ devices:
     comment: Title for the menu bar when viewing the devices or device limit pages
 
 helpSheets:
+  obfuscationTitle:
+    value: Obfuscation
+    comment: Title label for the Obfuscation features help sheet
+  obfuscationBody1:
+    value: These options should be used when a network is blocking connections to VPN servers. They will not help when a specific website or app blocks content because of the VPN.
```
flodolo, line 937:
```suggestion
    value: These options should be used when a network is blocking connections to VPN servers. They will not help when a specific website or app block content because of the VPN.
```

Don't think `blocks` is correct here.

flodolo (changes requested):

src/daemon/daemon.cpp
```diff
@@ -409,8 +475,32 @@ bool Daemon::switchServer(const InterfaceConfig& config) {
...
+    // The obfuscator only binds 127.0.0.1, so clear the IPv6 endpoint to keep
+    // WireGuard from selecting an [::1]
+    peerConfig.m_serverIpv6AddrIn = QString();
+    peerConfig.m_serverPort = obfuscator->localPort();
+  }
+
+  // The exit hop never needs an obfuscator, assign the empty obfuscator to
+  // allow changing of the obfuscation method on server switch.
+  if (config.m_hopType != InterfaceConfig::MultiHopExit) {
+    m_obfuscator = std::move(obfuscator);
```
artfwo: When `m_obfuscator` is assigned and `updatePeer(peerConfig)` is called later, what happens if updatePeer fails, e.g. because of network error? would it be safer to call updatePeer first and move obfuscator if that succeeds?

artfwo (approved):

src/translations/strings.yaml
encrypt94: i think `blocks` is correct because the subject "website or app" is singular

flodolo: I stand corrected, it's singular 🤯

flodolo (approved):

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -132,6 +133,17 @@ fi
 sudo mk-build-deps --tool 'apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends' \
    --install --remove $MK_BUILD_DEPS_ARGS "$(pwd)/mozillavpn-source/debian/control"
 
+# /usr/bin/cargo and /usr/bin/rustc may be installed as cargo-<version> and rustc-<version>
```
oskirby: Is this still necessary after the update to `debian/rules` for Rust 1.91?

oskirby (approved):

src/ui/popups/ServerUnavailablePopup.qml
mcleinman: Ah, gotcha - When off, it's a nudge towards using obfuscation features.

src/translations/strings.yaml
encrypt94: what do you think about this rewording? "Wraps VPN traffic in an encrypted Shadowsocks connection to make it less recognizable. Could work when all other approaches do not.

PR #11276, opened 2026-05-13
Drop libcap dependency

artfwo:
Drop libcap dependency on Linux.

Systemd unit files for socksproxy and the daemon now have [CapabilityBoundingSet](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#CapabilityBoundingSet=) and [NoNewPrivileges](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#NoNewPrivileges=) options so neither binary now runs with full root caps.


oskirby (approved):
Looks good

PR #11277, opened 2026-05-13
VPN-7597: fix login bug, and save the on demand rule

mcleinman:
## Description

We need to skip passing the config to iOS if we're logging out or logging in. This has been added to `controller.cpp`, and again allows logging in and logging out.

I snuck another couple fixes in here as well, in `ioscontroller.swift`:
1) The auto reconnect was being saved. 🤦🏻 
2) I found that if the config was changed while the VPN was disconnected, there was an error when starting the VPN from intent. (This error cleared itself for future runs after starting the VPN from client.) Turns out the session had changed.

## Reference

VPN-7597

## Checklist
    

oskirby (approved):
LGTM

PR #11283, opened 2026-05-14
fix(l10n): correct typo 'commnent' to 'comment' in strings.yaml

Tarunswamy-Muralidharan:
## Description

The l10n string extraction was failing because of an unknown field 'commnent' in the strings.yaml file. This corrects the typo so that localization extraction can proceed successfully.

## Changes

- Fixed typo: 'commnent' → 'comment' in  line 466

## Verification

- The field now reads 'comment' which is an allowed field in the l10n extraction
- This aligns with the main branch where this typo was already fixed

Fixes #11282

## Related

- The same typo exists in the releases/2.35.0 branch and will be fixed in a separate PR.

Issue #11282, opened 2026-05-14
l10n extraction failing for field typo (`commnent`)


flodolo:
String extraction in the l10n report is current failing
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/actions/runs/25848758103/job/75949955740

```
HEAD is now at 223c19f79 [Bot] Update i18n (#11139)
The key vpn.connectionInfo.unitPing contains unknown field(s) ['commnent']. Allowed fields: ['comment', 'value']
Generating strings... Failed
```

We need to fix the typo across all release branches
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/releases/2.35.0/src/translations/strings.yaml#L466
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/releases/2.36.0/src/translations/strings.yaml#L466

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/VPN-7616)

flodolo:
Running string extraction again 🤞🏼

flodolo:
OK, it worked https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/569

PR #11284, opened 2026-05-14
fix(l10n): correct typo 'commnent' to 'comment' in strings.yaml (#11283)

flodolo:
## Description

This is porting the same fix we already have on main and releases/2.36.0 to unblock #11282 

## Reference

Issue #11282


oskirby (approved):
LGTM

PR #11310, opened 2026-05-21
Remove superseeded linux build jobs

strseb:
## Description

In order to unblock https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11153 - I'm removing the builders that build ppa-ish packages. As we now have the static-qt linked packages on the mozilla ppa, rpm's on archive and flatpacks available.


strseb:
Question for @oskirby should `ppa-automation.yaml` also need to be removed?

encrypt94:
I think `taskcluster/docker/linux-dpkg-build/Dockerfile` can be removed as well

artfwo:
are we getting rid of RPMs as well?

strseb:
> are we getting rid of RPMs as well?

No :)  we still have the `linux64/release-rpm` task, which pushes the static-x86 rpm to archive.mozilla.org. 
Which we maybe also should push to the gcp-package thing.

oskirby:
> I think `taskcluster/docker/linux-dpkg-build/Dockerfile` can be removed as well

I would keep it for the time being, I would like to keep one Ubuntu build going of some flavor but maybe turn it into a rolling target.

PR #11323, opened 2026-05-28
VPN-5454: ios widget

mcleinman:
## Description

Adds widgets for iOS. This includes a bunch of iOS-specific code around colors and such to make this happen, as this stuff all needs to be SwiftUI. These colors/names are identical to what they are in our app.

- This changes the dark mode toggle to green (both in app and widget), something done in consult w/ Lesley
- added `var turnOnConfirmation: LocalizedStringResource` so we can stop reusing code all over the place.
- tried adding the globe (from our main screen) to the larger widget, but it didn't look good
- Known issue - we localize the widget text based on iOS system locale. We localize the actual city name based on the app's locale. Thus, for users that have a different app locale and system locale, they may show a weird translation for city name. (We could fix this by mapping all cities translations to the iOS-specific pipeline, but that doesn't seem worthwhile to me. Open to other opinions.)
- Known issue - we get limited control on updating widgets and controls. All widgets and all controls should be updated to the appropriate state no matter whether the VPN is toggled on/off by intent/Siri, the client, a control, or the widget. The one exception: The lock screen control (when used) may not update the widget. The request to update the widget seemingly disappears - iOS may not run it. We do not have full visibility into how Apple decides when to run this, and thus for now I recommend we accept this corner case.

Video: https://github.com/user-attachments/assets/664bd9bc-c49d-45af-bc84-89734e00b3f0
You'll note in the video that toggling via control doesn't update the widget. We get a very limited-but-secret amount of widget refreshes from iOS, and then it stops listening to requests for them. The limit doesn't apply when the request comes from the app, but apparently does when it comes from a control. I've been testing this a lot over the past hour, and I've apparently hit the limit. So don't worry - it usually is toggling the widget when the control center button turns off/on.

## Reference

VPN-5454

## Checklist
    

flodolo:
I assume the number of commits is not expected?

flodolo (approved):

mcleinman:
> I assume the number of commits is not expected?

Yes - artifact of basing this off un-merged PRs, that were then changed before merging. Have re-merged (not rebased 😬) those predecessor PRs and `main` several times, leading to this.

The code is correct, even though commits are ugly. My opinion is doing a rebase at this point increases risk of introducing a bug due to fixing merge conflicts - and given that we flatten the commits before merging to `main`, this is okay for now. I'm open to disagreements on that though.

ios/widgetextension/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ set_target_properties(widgetextension PROPERTIES
     XCODE_GENERATE_SCHEME TRUE
     XCODE_ATTRIBUTE_PRODUCT_BUNDLE_IDENTIFIER "${BUILD_IOS_APP_IDENTIFIER}.widget-extension"
     XCODE_ATTRIBUTE_SWIFT_VERSION "5.0"
+    XCODE_ATTRIBUTE_SWIFT_OPTIMIZATION_LEVEL "-Onone"
```
oskirby: Why do we need to disable optimization here? Is this a leftover debug thing?

ios/widgetextension/ToggleWidget.swift
```diff
@@ -0,0 +1,204 @@
...
+
+struct SmallToggleWidget: View {
+  @Environment(\.colorScheme) var colorScheme
+  let entry: VPNStatusEntry
+
+  var body: some View {
+  VStack(alignment: .leading, spacing: 0) {
+    Image("logo")
+      .resizable()
+      .frame(width: 30, height: 30)
```
oskirby, line 64: Just for my own personal curiosity, where do these dimensions come from? Are they specified by Apple somewhere for standard widget sizes or something?

scripts/ci/check_colors.py
```diff
@@ -141,3 +142,34 @@ def getQmlFiles(directory):
     for color in colors_used:
         if color not in color_list:
             exit(f"Unexpected color {color} found in {qml_file_path}")
+
```
oskirby, line 145: Minor nit - if we are going through the effort of writing a script to parse the nebula `colors.js` file and validate that the corresponding swift file matches, it's not that much more complicated to parse `colors.js` and generate the swift file instead. That would preserve `colors.js` as the source of truth for theming.

src/controller.cpp
```diff
@@ -393,7 +405,8 @@ auto Controller::setupConfigs(DNSPortPolicy dnsPort,
   exitConfig.m_serverPort = exitServer.choosePort();
   exitConfig.m_allowedIPAddressRanges = allowedIPList;
   exitConfig.m_dnsServer = DNSHelper::getDNS(exitServer.ipv4Gateway());
-  exitConfig.m_exitCity = exitServer.cityName();
+  exitConfig.m_exitCity =
```
oskirby, line 408: I guess I missed something here in the last review of app intents stuff. But there is some ambiguity here as to what the `entryCity` and `exitCity` are used here for:
- Are these values localized, or do they correlate to the values that would be used for settings?
- Do the settings become invalid if the language changes?
- For `multihopSupported() == true` platforms, what does it mean when we have two `InterfaceConfigs`?

I guess it doesn't really matter much, since these are just used for iOS to pass the city name around.

oskirby (approved):
Looks good. None of my comments are sufficient to hold this up.

Approved!

ios/widgetextension/ToggleWidget.swift
mcleinman: The overall widget sizes are Apple: https://developer.apple.com/design/human-interface-guidelines/widgets#System-family-widgets

The 30x30 icon for the logo is just what looked decent to me in this sized widget. Open to tweaks.

ios/widgetextension/CMakeLists.txt
mcleinman: It seems to run fine without it. This may have been needed when we were handling translations differently? I'm not sure, but I've removed it.

scripts/ci/check_colors.py
mcleinman: While 16 lines are copying colors, the majority of this file is widget-specific code.

The color check was one of the last things I did for this PR. As much as I've been adding scripting/checks, I'd love to not automate this piece. Without automating this, we can be more intentional about which colors we're using in the widget with the smaller list (and simplify picking colors for future engineers), and I'd imagine the "create part of the color file" code would be more brittle than these checks.

That said, I don't feel strongly on this - if you'd really like this, please let me know.

src/controller.cpp
mcleinman: * Exit city and entry city were added as part of the app intents work - this is how we're able to show the city name in response.  However, there was an issue - they weren't localized. This PR updates them to be localized. (This leads to the edge case in the fourth bullet point in the PR description, however.)
* Yes, [when the language is updated we send a new config](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11323/changes#diff-53fc727c3acacc0c0b7c531be394c8074b1211826f68cff4e8217b65c536ab8fR91-R101).
* As you point out, 2 config systems don't use this yet. That said, I've packed both entry and exit city names into the exit config - I figured if we use this on other OSes that support proper multihop, we can just pull these translated names from the exitConfig. Would that work okay?

I've added a comment in `interfaceconfig.h` to explain why we have these. Can also rename the variable if that seems helpful.

scripts/ci/check_colors.py
oskirby: I don't feel strongly about it either. Let's leave this as it currently stands.

mcleinman:
@oskirby I'm merging this to get this in for string freeze, but please let me know if you want me to do follow ups for any of these - I'm happy to.

PR #11329, opened 2026-06-01
Nit: Remove shared lib support from add_rust_library

oskirby:
## Description
The iOS framework support in `add_rust_library()` has always kind of bothered me as it created a bunch of headaches every time I needed to touch the Rust helpers due to being a complete separate path for iOS, and it frequently resulted in broken iOS builds.

However! Now that glean has been removed from the codebase, this is no longer used by anything, so for my own piece of mind, let's remove this tangle of CMake pain.

If we ever need to add Rust frameworks back, I freely accept that assignment in the future if and when we need it.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/scripts/cmake/rustlang.cmake
+++ b/scripts/cmake/rustlang.cmake
@@ -158,16 +158,10 @@ endif()
 ### Helper function to get the rust library filename with extension.
 #
 # Sets the variable "RUST_LIBRARY_FILENAME" with the value.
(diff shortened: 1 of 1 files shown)


encrypt94:
hey i'm sneaking in to kill all the cleaning fun! :wave: can we keep it?
I'm actually using it in #11275 to build a shared library that we load via JNA on Android

PR #11331, opened 2026-06-02
Set minimum required Qt version to 6.9

oskirby:
## Description
Now that we are deprecating the PPA, let's also start enforcing a minimum Qt version against which the app will build. 

This is mostly a straightforward removal of backwards compatibility hacks - except for the fact that the newest Github runner images are on Ubuntu Noble, and therefore only provide Qt 6.4.2. Thus we encountered a bunch of pain by switching our Linux tests over to the static Qt builds from Taskcluster.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

src/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ mz_target_handle_warnings(mozillavpn)
 
 find_package(Qt6 REQUIRED COMPONENTS Svg)
 if(Qt6_VERSION VERSION_GREATER_EQUAL 6.10.0)
+    set(QT_NO_PRIVATE_MODULE_WARNING ON)
```
mcleinman, line 14: out of curiosity, why this change?

src/cmake/linux.cmake
```diff
@@ -43,25 +43,23 @@ target_sources(mozillavpn PRIVATE
...
 
     if (QT_FEATURE_static)
         find_package(Qt6 REQUIRED COMPONENTS WaylandClientPrivate)
         qt_import_plugins(mozillavpn INCLUDE Qt6::QWaylandIntegrationPlugin)
         qt_import_plugins(mozillavpn INCLUDE Qt6::QWaylandAdwaitaDecorationPlugin)
         target_link_libraries(mozillavpn PRIVATE Qt6::WaylandClientPrivate)
-    endif()
 
-    target_link_libraries(mozillavpn PRIVATE PkgConfig::polkit)
+        qt_import_plugins(mozillavpn INCLUDE Qt6::QOffscreenIntegrationPlugin)
```
mcleinman, line 61: this is new, right?

src/cmake/android.cmake
```diff
@@ -67,12 +67,10 @@ set_property(TARGET mozillavpn PROPERTY QT_ANDROID_EXTRA_LIBS
     APPEND)
 
 
-if( ${Qt6_VERSION} VERSION_GREATER_EQUAL 6.4.0)
-    set_property(TARGET mozillavpn PROPERTY QT_ANDROID_EXTRA_LIBS
+set_property(TARGET mozillavpn APPEND PROPERTY
+    QT_ANDROID_EXTRA_LIBS
         ${OPENSSL_LIBS_DIR}/libcrypto_3.so
-        ${OPENSSL_LIBS_DIR}/libssl_3.so
-    APPEND)
```
mcleinman, line 74: Don't we still need the `APPEND` here?

mcleinman (commented):
Looks good, but would love another review on this as well before merging. Thanks!

src/cmake/linux.cmake
oskirby: Yes, this was necessary to get the Linux functional tests working with a static Qt, since the tests use the offscreen rendering plugin in lieu of setting up an virtual X environment.

src/cmake/android.cmake
oskirby: It's still there, just reorganized a little

src/CMakeLists.txt
oskirby: Qt/CMake generates a bunch of warnings when trying to link against the private modules, this just suppresses the warning message since it's an intentional linkage (for sentry symbolification)

src/cmake/android.cmake
mcleinman: yes, yes it is. Sorry, fully missed that up top.

strseb (approved):
Nice!!

PR #11332, opened 2026-06-02
VPN-7644: fix some CI tests

mcleinman:
## Description

Most of the common CI test failures - 
- IP Info: having something hit wrong (should be fixed here)
- Addons: Has date off by a minute (should be fixed here)
- Addons: Unexpected shows "Yesterday". I'm not 100% sure on this one yet, it is not fixed here.
- MessagesAddonPopover: Sometimes times out partway through the run. I have a hopeful fix here, but I'm not confident it will do it.

## Reference

VPN-7644

## Checklist
    
### DIFF
--- a/tests/functional/testMessageAddonPopover.js
+++ b/tests/functional/testMessageAddonPopover.js
@@ -17,6 +17,10 @@ async function loadAddonsAndNavigateHome(testAddonName) {
       'Mozilla.Shared', 'MZAddonManager', 'loadCompleted', 'true');
 
   await vpn.waitForQueryAndClick(queries.navBar.HOME.visible());
+  await vpn.waitForQueryAndClick(
+      queries.navBar.HOME.visible());  // shouldn't be needed - but next line is
+                                       // timing out on some Linux runs, so
+                                       // let's be extra sure we get to home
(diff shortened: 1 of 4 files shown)


mcleinman:
Well, functional tests all passed on first try, so that's an encouraging sign.

tests/functional/testMessageAddonPopover.js
```diff
@@ -17,6 +17,10 @@ async function loadAddonsAndNavigateHome(testAddonName) {
       'Mozilla.Shared', 'MZAddonManager', 'loadCompleted', 'true');
 
   await vpn.waitForQueryAndClick(queries.navBar.HOME.visible());
+  await vpn.waitForQueryAndClick(
```
oskirby: A common edge case that trips things up in functional testing is expecting a thing to be clickable once `visible()` is `true` but this often doesn't work whenever there are loading delays or property animations. The usual fix when I've run into these kind of test issues is to add a `ready` property that becomes `true` only after all the transitions and animations are finished.

mcleinman: Thanks, updated.

oskirby (approved):
LGTM

PR #11334, opened 2026-06-03
Update cargo in base taskcluster dockerfile to 1.85

oskirby:
## Description
It seems that dependabot is having troubles updating the `Cargo.lock` file due to a version incompatibility with the `cargo` version on the taskcluster Docker containers. The problem occurs because the base build image uses Cargo 1.75. So lets update that to 1.85.

There is an argument to be made that we should go even newer (eg: to match #11275)

## Reference
Example dependabot failure: #11320

## Checklist
    

jcristau (approved):
Would it be worth looking at bumping the base image from 22.04 to something newer, or would that affect the build's installability on older distros?

oskirby:
There shouldn't be anything preventing us from updating the base image to something newer. I can give that a try while we're at it.

PR #11335, opened 2026-06-03
VPN-7638 2.38 addons

mcleinman:
## Description

Planning on pulling whichever bullets don't hit 2.38.

Screenshot 110Screenshot 111
Screenshot 112Screenshot 113

## Reference

VPN-7638

## Checklist
    

addons/strings.yaml
```diff
@@ -225,6 +225,17 @@ commonStrings:
     value: "[Linux] Support for IPv6-only networks was introduced in the last release, though switching servers while using multihop on IPv6-only networks sometimes had issues. It no longer should."
     comment: Bullet point with a specific update. Should have tag that indicates it applies to Linux platform.
 
+238updateMessage:
+  bullet1:
+    value: "[all] When contacting support, including logs is now optional. While Mozilla VPN logs have never included personal data, we now only default to including logs when it would help diagnose your issue."
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to all platforms.
+  bullet2:
+    value: "[iOS] When on Airplane mode, trying to start the VPN gives an error. However if starting via Siri or Shortcuts, the VPN would be turned on after Airplane mode was deactivated. This no longer happens."
```
```suggestion
    value: "[iOS] When on Airplane mode, trying to start the VPN gives an error. However, if starting via Siri or Shortcuts, the VPN would be turned on after Airplane Mode was deactivated. This no longer happens."
```

Looks like that's what it's called?
https://support.apple.com/en-us/108785

addons/strings.yaml
```diff
@@ -225,6 +225,17 @@ commonStrings:
...
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to iOS platform.
+  bullet3:
+    value: "[all] The `Check for Updates` text is, once again, centered in its button."
```
flodolo: This feels like oversharing 😅

flodolo (approved):

addons/strings.yaml
mcleinman: Fair enough - removed this one.

addons/strings.yaml
mcleinman: Thanks, good catch. Fixed in 2 spots.

addons/strings.yaml
```diff
@@ -225,6 +225,17 @@ commonStrings:
     value: "[Linux] Support for IPv6-only networks was introduced in the last release, though switching servers while using multihop on IPv6-only networks sometimes had issues. It no longer should."
     comment: Bullet point with a specific update. Should have tag that indicates it applies to Linux platform.
 
+238updateMessage:
```
oskirby, line 228: As a general thought, I am noticing that as we make our way through these releases the naming convention behind the string IDs is kind of falling apart. These strings cross across multiple releases and addons and it's kind of turning into a mess.

With the power of hindsight, it would be cleaner if we just put all of these "what's new strings" into one giant bucket and pick them into addons as we need. For example:

```yaml
whatNewMessage:
  vpn7576ShareLogsOption:
     value: "[all] When contacting support, ..."
     comment: some stuff
```

Note: This is just an observation on how things are getting messy, I am not requesting any changes.

oskirby (approved):
LGTM

addons/strings.yaml
oskirby: As a bonus - this approach gives the feature owner a place to add their what's new string when they are authoring their feature PR rather than dumping that responsibility onto the release manager who might be less familliar with the feature being added.

flodolo (approved):

addons/strings.yaml
mcleinman: I love this, especially the "key is the Jira ticket". I wanted some subcategories so that it didn't just end up as a 200 string dictionary after a couple years, but I agree the current way of doing it isn't hitting it right. What would you think about a "2026q2" category, etc.? Some stuff would still cross the quarter boundaries, but it prevents the misnamed key issue we currently have.

I'm not sure we'd want "add your own string" in - to keep a coherent voice in the product notes, I would want to be able to do some wordsmithing (by me or some other person), and that's hard to do after strings are merged. Maybe "add a comment with your suggested update note"?

Either way, I'm going to merge this right now to get this in for now. But next release, I'd like to change how this works.

oskirby: > What would you think about a "2026q2" category, etc.? 

That sounds fine. I agree that eventually we might wind up with an enormous list of strings and having some way to organize them would be a good idea. That's also why I put the ticket number in the key so that we can at least sort and search them in a meaningful way.

> I'm not sure we'd want "add your own string" in

I agree that keeping a common voice is a good thing, but I would like there to be some way to offload the duty of authoring the whats new message to the feature owner when they are building the feature rather than leaving at as an afterthought for some someone else to deal with. There is no support for such a thing right now, but maybe add a boolean to the YAML to indicate that a string isn't finalized yet and to hold off translating it?

For example:
```yaml
whatNewMessage:
  vpn7576ShareLogsOption:
     value: "[all] When contacting support, ..."
     placeholder: true
     comment: some stuff
```

Issue #5636, opened 2023-01-15
[Android] Contact Support – mandatory fields need to have asterisk sign


mMARUF:
**Steps to Reproduce**
1. Open Mozilla VPN in Android 
2. Navigate to the 'Contact Support'
3. Observe the mandatory fields. 

**Expected result:**
Mandatory fields need to have asterisk sign

**Actual result:**
Mandatory fields do not have asterisk sign

**Additional notes:**

  **VPN Version:** 2.12.0
  **Device:** Sony Xperia 5 ii
  **Android 11**

**Attachment:**
![Screenshot_20230113-130727](https://user-images.githubusercontent.com/45368054/212533008-cfe2a0df-854c-4635-8d2f-016287a458f6.png)

data-sync-user:
➤ Matt Lichtenstein commented:

Santiago Andrigo this is an enhancement suggestion from a user, please advise

slsanthoshraj19-dotcom:
Hi! I'd like to work on this issue ,So is it still available to work on

PR #11339, opened 2026-06-04
Add missing asterick for important fields in contact form

Sukumarsawant:
## Description
 Added red asterisk indicators (*) to all mandatory fields in the Contact Support form to clearly communicate to users which fields are required before submission.

Fields updated:

Email address & Confirm email

## Reference

https://github.com/mozilla-mobile/mozilla-vpn-client/issues/5636

## Checklist
- [x] My code follows the style guidelines for this project.
- [x] I have not added any packages that contain high-risk or unknown licenses (GPL, LGPL, MPL, etc.; consult DevOps if in question).
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added thorough tests where needed.

### DIFF
--- a/src/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
+++ b/src/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
@@ -112,7 +112,7 @@ MZViewBase {
                 MZBoldLabel {
                     property string enterEmailAddress: MZI18n.InAppSupportWorkflowSupportFieldHeader
 
-                    text: enterEmailAddress
+                    text: enterEmailAddress + " *"
                     lineHeight: MZTheme.theme.labelLineHeight
                     lineHeightMode: Text.FixedHeight
                     wrapMode: Text.WordWrap
--- a/src/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
+++ b/src/ui/screens/getHelp/contactUs/ViewContactUsForm.qml
@@ -72,7 +72,7 @@ MZViewBase {
                 MZBoldLabel {
                     property string enterEmailAddress: MZI18n.InAppSupportWorkflowSupportEmailFieldLabel
 
-                    text: enterEmailAddress
+                    text: enterEmailAddress + " *"
                     lineHeight: MZTheme.theme.labelLineHeight
                     lineHeightMode: Text.FixedHeight
                     wrapMode: Text.WordWrap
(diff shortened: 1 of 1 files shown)


Sukumarsawant:
@strseb @mcleinman  do have a look once and let me know if you there's anything more need in it . The fix looked simple to me yet stale for years now somehow?

Ekram-B2 (commented):
Just a person taking a look. Simple change but effective towards being a great UI enhancement!

strseb (changes requested):
Hey! First off, thank you very much for the PR!

Just a small change, then this is good to go. Visually this is fine but the current change will also change the text that the screen reader reads to something like depending on os+screenreader combo
```
“Enter your email star”
or
“Enter your email asterisk”
```
Which may be more confusing then helpful. Please add 
```
Accessible.name: enterEmailAddress
```
so that the text for screen readers remains text-only :)

Sukumarsawant:
> Hey! First off, thank you very much for the PR!
> 
> Just a small change, then this is good to go. Visually this is fine but the current change will also change the text that the screen reader reads to something like depending on os+screenreader combo
> 
> ```
> “Enter your email star”
> or
> “Enter your email asterisk”
> ```
> 
> Which may be more confusing then helpful. Please add
> 
> ```
> Accessible.name: enterEmailAddress
> ```
> 
> so that the text for screen readers remains text-only :)

hey sorry for the delay. GitHub somehow didn't ping me for this one. I am adding this change
Also, I never considered (missed) that the screen reader would read it that way; thanks !!

Sukumarsawant:
Have pushed the change @strseb, do tag me if there's anything wrong in this or anything I missed
Also, is there any way I could test the CI (any commands or so )?
Or would it be required to run manually?

mcleinman:
Could you re-merge main into your branch? That will allow this failing test to pass, and then we'll merge this. (Something changed in a dependency which causes this, and it's been fixed in `main`.) Thanks!

Sukumarsawant:
> Could you re-merge main into your branch? That will allow this failing test to pass, and then we'll merge this. (Something changed in a dependency which causes this, and it's been fixed in `main`.) Thanks!

rebased 👍   though would need approval to check the CI i.e if it still fails

strseb (approved):
Sorry for the wait! Looks great now :) Thank you!

PR #11347, opened 2026-06-09
VPN-7645: Implement log truncation without an explicit write/reopen

oskirby:
## Description
So, we have logic in the app to try and truncate the logs files, but this logic only executes whenever we open or write a log file. However, many users may open the VPN client and leave it running for weeks of not months at a time, this can potentially let the log file grow without bounds. What we really need to do is evaluate whether the logs need truncation whenever we are adding a log message.

This required some refactoring to add a check in `LogHandler::addLog()` to evaluate the log file size, and re-open it whenever it grows too large.

## Reference
JIRA Issue: [VPN-7645](https://mozilla-hub.atlassian.net/browse/VPN-7645)

## Checklist
    
[VPN-7645]: https://mozilla-hub.atlassian.net/browse/VPN-7645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcleinman:
>  What we really need to do is evaluate whether the logs need truncation whenever we are adding a log message.

Do we know how cheap this is? Would it be better run as a chron-ish job, maybe as part of the TaskGroup handled with the `m_periodicOperationsTimer`?

oskirby:
> Do we know how cheap this is? Would it be better run as a chron-ish job, maybe as part of the TaskGroup handled with the `m_periodicOperationsTimer`?

The check is extremely cheap, it just checks the current file position via `QFileDevice::pos()` and compares it to a constant. The truncation operation is a bit more expensive as we wind up copying a significant portion of the log file, but forunately that should be an infrequent event.

src/utils/loghandler.cpp
```diff
@@ -317,44 +275,72 @@ bool LogHandler::makeLogDir(const QDir& dir) {
...
-void LogHandler::openLogFile(const QMutexLocker<QMutex>& proofOfLock) {
+void LogHandler::setLogDevice(QFileDevice* device,
+                              const QMutexLocker<QMutex>& proofOfLock) {
   Q_UNUSED(proofOfLock);
-  Q_ASSERT(!m_logFile);
-  Q_ASSERT(!m_output);
+  Q_ASSERT(device);
+
+  m_output.reset(device);
+  connect(this, &LogHandler::logEntryAdded, device,
```
encrypt94: I see a possible race condition here, routing the file write through `logEntryAdded` uses a `QueuedConnection` when a line is logged off the main thread then the write may run unlocked and race with `writeLogs()` or `cleanupLogFile()` under the lock. Force a `DirectConnection` should keep the write operation under mutex lock.

oskirby: Good catch! Maybe I was getting a bit ambitious doing everything through a signal here.

encrypt94 (approved):
LGTM

PR #11349, opened 2026-06-09
VPN-7632 Fix title drifting right with wider shadows in Adwaita decoration

encrypt94:
## Description

Fix title drifting right with wider shadows in Adwaita decoration by  subtracting both side margins from the rect width.

The window title rect was built using the full `surfaceRect.width()` while its origin was already offset by `margins().left()`. Since the title was centered within this `rect (dx = (top.width() - size.width()) / 2)`, the left margin was counted twice, pushing the title right by `margins().left()` pixels.

This was always slightly off-center in the Adwaita decoration plugin, but became clearly visible after
`ceShadowsWidth` grew from 10 to 30 (except for me because i was looking at the shadows)

From the top:

- Title with unpatched qtwayland plugin
- Title after shadows patch 
- Fixed title

Screenshot from 2026-06-09 19-20-53

## Checklist
    

oskirby (approved):
LGTM

PR #11350, opened 2026-06-09
VPN-7646: Accessory widgets

mcleinman:
## Description

Adding several iOS accessory widgets - 2 round, 2 inline (where we get just text and an SF symbol), and one rectangular.

Also snuck in some improvements to existing widgets:
- It doesn't seem like a good candidate for Standby (nightstand mode on iOS) nor Carplay, so removed them from these spots. (Dark mode didn't work for Standby, and I realized it wasn't worth fixing it.) Inexplicably deprioritizing CarPlay is only available on iOS 26+, hence the split logic there.
- Restructured code to allow for more reuse

## Images
Single hop (both round, sole rectangular, and the city name inline one - inline one is next to date on top)
IMG_0069 IMG_0070

Multi-hop (both round, sole rectangular, and the simple inline one - inline one is next to date on top)
IMG_0073IMG_0074

## Reference

VPN-7646

## Checklist
    
### DIFF
--- a/ios/widgetextension/CityAccessoryWidget.swift
+++ b/ios/widgetextension/CityAccessoryWidget.swift
@@ -7,6 +7,58 @@ import NetworkExtension
 import SwiftUI
 import WidgetKit
 
+struct CityInlineAccessoryWidget: View {
+  @Environment(\.colorScheme) var colorScheme
+  let entry: VPNStatusEntry
+
+  var body: some View {
+    ViewThatFits(in: .horizontal) {
+      HStack(spacing: 4) {
+        Image(systemName: entry.isConnected ? "shield.lefthalf.filled" : "shield.lefthalf.filled.slash")
+          .font(.system(size: 26))
+        if let exitCity = entry.exitCity, !exitCity.isEmpty {
+          if let entryCity = entry.entryCity, !entryCity.isEmpty {
+            Text(LocalizedStringResource("vpn.multiHopFeature.multiHopToggleCTA", defaultValue: "Multi-hop"))
+          } else {
+            Text(exitCity)
+          }
+        }
+      }
+    }
+  }
+}
+
+struct CityRectangularAccessoryWidget: View {
+  @Environment(\.colorScheme) var colorScheme
+  let entry: VPNStatusEntry
+
+  var body: some View {
+    ZStack {
+      if entry.isConnected {
+        AccessoryWidgetBackground()
+      }
+      HStack {
+        VStack {
+          Image("logo")
+            .resizable()
+            .frame(width: 25, height: 25)
+          Spacer()
+          Image(systemName: entry.isConnected ? "shield.lefthalf.filled" : "shield.lefthalf.filled.slash")
+            .font(.system(size: 25))
+        }
+        Spacer()
+        CityTextComponentView(entry: entry)
+          .allowsTightening(true)
+          .minimumScaleFactor(0.5)
+        }
+        .padding(8)
+    }
+    .containerBackground(for: .widget) { // need this for AccessoryWidgetBackground to work, it seems
+      WidgetColors.backgroundColor(colorScheme, isConnected: entry.isConnected)
+    }
+  }
+}
+
 struct CityCircularAccessoryWidget: View {
   @Environment(\.colorScheme) var colorScheme
   let entry: VPNStatusEntry
@@ -54,6 +106,10 @@ struct CityAccessoryWidgetView: View {
     switch family {
     case .accessoryCircular:
       CityCircularAccessoryWidget(entry: entry)
+    case .accessoryInline:
+      CityInlineAccessoryWidget(entry: entry)
+    case .accessoryRectangular:
+      CityRectangularAccessoryWidget(entry: entry)
     default:
       // This should never be seen
       Text("Error: Unsupported widget size")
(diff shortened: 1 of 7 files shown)


src/translations/strings.yaml
```diff
@@ -1146,6 +1146,22 @@ toggleWidget:
     value: Turn Mozilla VPN on and off, and see the current location.
     comment: More information about what the Mozilla VPN widget contains.
 
+logoAccessoryWidget:
+  description:
+    value: See current Mozilla VPN status.
+    comment: More information about what the logo widget contains - which is simply current VPN status.
+  vpnOn:
+    value: Mozilla VPN on
+    comment: Very short string shown when VPN is active
```
flodolo: Is this displayed anywhere? If not, what's the reason to keep it short?

flodolo (approved):

src/translations/strings.yaml
mcleinman: It is displayed - it's the string above the clock in the second set of images. iOS will truncate it if it's too long, so if a language's word for "on" or "off" is lengthy (are there languages where that is true?), we'd probably want this string to simply say "VPN on". Is that something I should put in the comment?

flodolo: I think we should try to figure out how many characters are available (roughly), and suggest to just use "VPN on/off" if longer than X.

ios/widgetextension/CityAccessoryWidget.swift
```diff
@@ -0,0 +1,135 @@
...
+
+struct CityAccessoryWidget: Widget {
+  static let kind = "org.mozilla.ios.FirefoxVPN.CityAccessoryWidget"
+
+  var body: some WidgetConfiguration {
+      return StaticConfiguration(kind: Self.kind, provider: VPNStatusProvider()) { entry in
+            CityAccessoryWidgetView(entry: entry)
+          }
+          .configurationDisplayName(LocalizedStringResource("vpn.mobileOnboarding.panelOneTitle", defaultValue: "Mozilla VPN"))
+           .description(LocalizedStringResource("vpn.cityAccessoryWidget.description", defaultValue: "See current Mozilla VPN status, along with the current server."))
```
oskirby: there is a bit of weird spacing/indentation on this line

ios/widgetextension/CityAccessoryWidget.swift
```diff
@@ -0,0 +1,135 @@
...
+  let entry: VPNStatusEntry
+
+  var body: some View {
+    ViewThatFits(in: .horizontal) {
+      HStack(spacing: 4) {
+        Image(systemName: entry.sysImageName)
+          .font(.system(size: 26))
+        if let exitCity = entry.exitCity, !exitCity.isEmpty {
+          if let entryCity = entry.entryCity, !entryCity.isEmpty {
+            Text(LocalizedStringResource("vpn.multiHopFeature.multiHopToggleCTA", defaultValue: "Multi-hop"))
```
oskirby, line 21: I think there is an argument to be made that if we are extremely limited on space for text, it would be more useful to show the exit city rather than "multi-hop". The apparent location of the user is probably more of a meaningful context hint than the details of how they are connected to that location.

This is purely a comment, and you can ignore it if you feel that multi-hop is more interesting to the user.

ios/widgetextension/ToggleWidget.swift
```diff
@@ -12,11 +12,15 @@ struct VPNStatusEntry: TimelineEntry {
...
+
+  var sysImageName: String {
+    return isConnected ? "shield.lefthalf.filled" : "shield.lefthalf.filled.slash"
+  }
 }
 
 struct VPNStatusProvider: TimelineProvider {
   func placeholder(in context: Context) -> VPNStatusEntry {
-    VPNStatusEntry(date: Date(), isConnected: true, entryCity: "-------", exitCity: nil)
+    VPNStatusEntry(date: Date(), isConnected: true, entryCity: "Portland", exitCity: nil)
```
oskirby, line 23: Was this a debug change that got committed accidentally?

mcleinman: No, I wanted to give a better placeholder than simply "--------". I realize it wouldn't be translated, but I also don't believe this would ever be seen in typical use. If you think it should go back to the dashes, let me know.

ios/widgetextension/CityAccessoryWidget.swift
mcleinman: I fully agree with that, in part. In order of importance, what I think is most important to know in this situation:
1) You're on multi-hop (this is deliberately set, and we should be clear you're on single hop)
2) What the multi-hop exit city is
3) What the multi-hop entry city is

I don't think we'd hit all 3 of these in the space we have, but I'd love it if we could find a way to get both of the first two pieces of information in there. We don't have a common glyph/iconography for multi-hop, other than the arrow between the two cities. I don't think `-> Seattle, WA` would do a good enough job getting this across. Do you have thoughts?

ios/widgetextension/CityAccessoryWidget.swift
mcleinman: Thanks, fixed.

src/translations/strings.yaml
mcleinman: Roughly 16-18 characters, depending on the device. Should I add something to the comment that asks to translate it as simply "VPN on" if it's going to be more than 16 characters otherwise?

flodolo: Yes, please do that.

mcleinman: Done, thanks

oskirby (approved):
Looks good

PR #11353, opened 2026-06-10
VPN-7649 - Add more diagnostics to AndroidConnectionHealth

strseb:
## Description

It bugs me that we really cannot help that person. What is interesting is that connection health on android seems to find this issue. So let's use it's final path where we're out of options to log all network info that is not sensitive. Maybe its an mtu issue, or ipv6? Did we even get a handshake in wireguard? 
We don't know and cannot ask the user here. So let's add more data

### DIFF
--- a/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
+++ b/android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
@@ -30,6 +30,7 @@ class ConnectionHealth(service: VPNService) {
     private var mAltEndpoint: String = ""
     private var mResetUsed = false
     private var mPanicStateReached = false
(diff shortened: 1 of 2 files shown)


android/daemon/src/main/java/org/mozilla/firefox/vpn/daemon/ConnectionHealth.kt
```diff
@@ -16,10 +16,6 @@ import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
 
 class ConnectionHealth(service: VPNService) {
-    enum class ConnectionStability {
```
mcleinman, line 19: Good catch, thanks for cleaning this up. (I think I was the one who left it.)

mcleinman (approved):
Thanks for this - should help for future cases like this

PR #11359, opened 2026-06-10
VPN-7668: fix CI tests

mcleinman:
Currently windows Unit tests are failing, after merge of [recent PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11331/): `tar: D\:\a\\mozilla-vpn-client\\mozilla-vpn-client/3rdparty: Cannot open: No such file or directory`

(GitHub actions run the action version on `main`, so the CI tests passed on the PR that introduced this.)

VPN-7668

### DIFF
--- a/.github/actions/qt-setup-taskcluster/action.yml
+++ b/.github/actions/qt-setup-taskcluster/action.yml
@@ -42,7 +42,10 @@ runs:
         # even though the archive is truncated.
         curl -sSL -r0-4096 "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/${TASKID}/artifacts/${NAME}" -o qt-toolchain-head.tar.xz
         TARBALL_PATH=$((xzcat qt-toolchain-head.tar.xz | tar t | head -1 | grep -o '^[^/]*') 2>/dev/null || true)
-        echo "toolchain-path=${GITHUB_WORKSPACE}/3rdparty/${TARBALL_PATH}" >> $GITHUB_OUTPUT
+        # On Windows, $GITHUB_WORKSPACE uses backslashes; normalize to forward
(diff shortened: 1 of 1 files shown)


lesleyjanenorton (approved):
👍

.github/actions/qt-setup-taskcluster/action.yml
oskirby, line 69: The problem actually occurs here when we try to extract the archive. The reason I failed to catch this in the initial PR is because this step only runs on a cache miss, and the caching is only enabled on Windows (because archive extraction is quite slow only on Windows).

oskirby: There is some nuance to how Github caching works that makes it somewhat tricky to know when the cache hits or misses. Which can obscure when a line like this actually runs. I'm sorry that I failed to check it thoroughly.

.github/actions/qt-setup-taskcluster/action.yml
```diff
@@ -42,7 +42,10 @@ runs:
         # even though the archive is truncated.
         curl -sSL -r0-4096 "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/${TASKID}/artifacts/${NAME}" -o qt-toolchain-head.tar.xz
         TARBALL_PATH=$((xzcat qt-toolchain-head.tar.xz | tar t | head -1 | grep -o '^[^/]*') 2>/dev/null || true)
-        echo "toolchain-path=${GITHUB_WORKSPACE}/3rdparty/${TARBALL_PATH}" >> $GITHUB_OUTPUT
```
oskirby, line 45: I don't think this is going to fix the problem that you have described. Note that the `toolchain-path` output isn't actually used in this action when we run the `tar` extraction.

PR #11363, opened 2026-06-11
VPN-7671 - fix to again show all languages on stage

mcleinman:
## Description

It's me, hi, I'm the problem, it's me - I'm the one who introduced the bug. In https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11323/ I added this check, but this causes the Localizer to be initialized earlier than before, and that means we haven't set staging yet - and thus we were only loading the languages that passed the translation threshold. Let's move it to where it always should have been.

Two other things while we're in here:
- Remove an unneeded `#include "localizer.h"`. This was maybe missed [when related parts were removed](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11041/).
- I didn't know that some locales were only shown on stage. Let's make this explicit when we show locales that would otherwise fail the threshold test.

*Stage / prod*
Screenshot 118Screenshot 119

Separately, Swedish is mistakenly not included. I'll fix that in another PR.

## Reference

VPN-7671

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -88,17 +88,6 @@ Controller::Controller() {
   connect(&m_handshakeTimer, &QTimer::timeout, this,
           &Controller::handshakeTimeout);
 
-  // if the locale changed, send the new translation for the city name to iOS
-  // widget
-  connect(Localizer::instance(), &Localizer::localeChanged, this, [this]() {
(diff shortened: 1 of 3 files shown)


encrypt94 (approved):
LGTM! Thank you!

PR #11364, opened 2026-06-11
VPN-7672: catch when target/source doesn't match directory

mcleinman:
## Description

We've been missing Swedish in the app for a while because we were calculating the percent translated as 0. This is because of a [change in the l10n repo](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/commit/31c27065ad127a917ecb10dbe958a493fbf0d28f#diff-20faa5838b85995e04a3db7fc60ead56a8c25c7dbb8a10051767db828fd47bce) that resulted in the directory not matching the target or source languages. If this happens in the future, let's fail builds to make sure it doesn't sneak in.

(Separately, I'm working w/ l10n to get Swedish sorted out. I will not merge this PR until we fix it, otherwise all builds would start failing.)

Two other things in here:
- Better logging of translation percent
- Fixing typo in a comment

----

After merging Swedish, we were still getting failures. Turns out our demo test infrastructure wasn't formatted correctly, and potentially we hadn't been translations for all tests. There are now a few edits of test files to fix this - they now match their prod counterparts.

## Reference

VPN-7672

## Checklist
    

encrypt94 (approved):
LGTM

PR #11367, opened 2026-06-13
VPN-7665: always show an update addon

mcleinman:
## Description

Addon messages are only shown in a given locale if they are 100% translated. (This is adjustable, but defaults to 100% and we rarely adjust it.) This means that an unfortunate downside of our descriptive release notes is that a handful of locales aren't translating the bullet points, and thus are not showing an "update now" message to those users. 

This PR creates a backup addon of our default update text. When building addons, it takes some info (like the version number) from the most recent "update now" addon, and inserts it into this default one. The default addon utilizes the "locale" condition, and is only shown in the locales that were calculated to be less than 100% translated for that latest update addon. Thus, a given locale should only ever see one update addon - if they're seeing the default one, it's because they're not fully translated and thus will not see the typical one.

As part of this work, I restructured the translation work in addons into an additional python file, to more easily reuse some of it.

I considered another ways of approaching this: Keep just one update addon, but swap in new strings if it's not fully translated. Essentially, throw stuff into the enable javascript (run when an addon is turned on) to swap in the default strings if the bullet points aren't fully translated. 
This has lots of issues: 
- There is no way at the JS level to know what language we're currently in, nor how translated it is.
- We'd need to check that the common strings are fully translated, otherwise we'd need to turn the addon off.
- I don't believe we re-run `enable` when changing languages in the app.
- Making this work on older clients would be tough, as we'd need client updates as part of it.

Additionally, [one commit](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11367/changes/a4c88694f500e4107b81d9e2214d00b3904113f8) fixes two long standing issues which were revealed in testing:
1) The locales condition for addons forces all locale codes to lower case. This was causing it not to match locales like `pt_PT`, as we were testing something like `pt_PT == pt_pt`. This has been in the codebase for years. My guess is it's never been noticed because we've done limited locale conditions over the years; the only current one is just for English (which has a lowercase locale code already, so the bug doesn't reveal itself).
2) If everything is translated for an addon *except the promo text* (that new feature that shows a quick blurb on the main screen), we don't show the addon. That's a bummer, as we shouldn't block on that. (For instance, during testing the default update addon for pt_PT was missing *just* the promo text.) I've added code that skips the promo text when calculating translation completion for addons. (Part of our existing code for filtering on whether to show a promo: `obj.addon.promoText && obj.addon.promoText.length > 0`, so we're already handling the situation well in the client, and don't need changes there.)

## Reference

VPN-7665

## Checklist
    

scripts/addon/adjust_default_update_addon.py
```diff
@@ -0,0 +1,65 @@
...
+    dateLine = None
+    with open(args.latest_update_manifest, "r", encoding="utf-8") as f:
+        for line in f:
+            if '"shortVersion"' in line:
+                shortVersionLine = line
+            if '"date"' in line:
+                dateLine = line
+            if '"max_client_version"' in line:
+                maxClientLine = line
+
```
encrypt94, line 42:
```suggestion
localesLine = None
```

src/addons/addon.cpp
```diff
@@ -241,7 +241,7 @@ QList<ConditionCallback> s_conditionCallbacks{
        QStringList locales;
        QJsonArray localeArray = value.toArray();
        for (const QJsonValue& v : localeArray) {
-         locales.append(v.toString().toLower());
+         locales.append(v.toString());
```
encrypt94, line 244: maybe `addons/conditionwatchers/addonconditionwatcherlocales.cpp:57` should be changed the same way

encrypt94 (commented):
r+wc thank you!

encrypt94 (approved):

src/addons/addon.cpp
mcleinman: Done. I don't think this matters - that major language part is always lower case for us - but nice to keep this consistent.

mcleinman: That said - I believe this code has a bug. It's using "contains" - in this case matching the entire element (`es` or `en` for example) against the entire list, to see if the list contains this element. It's *not* seeing if any elements within the list contain that string as part of its substring - which is what it wants to be doing. (This PR wouldn't reveal the bug, as we're manually checking fallback languages in the new python script.) 

I just put up a commit fixing this - I'd appreciate a quick set of eyes on it.

encrypt94 (approved):

PR #11377, opened 2026-06-17
VPN-7680: IP address rotation button (behind feature flag)

mcleinman:
## Description

If this wasn't going to live behind a feature flag, I'd want a slightly thinner icon - but given it's not revealed to normal users, I think this is okay.

Screenshot 121 Screenshot 122

- `ControllerView.qml` work - keeps the window open when it is refreshing. This was added for VPN-3613, and I believe my change keeps the original intent.
- `src/ipaddresslookup.cpp` work - changes the label back to "Loading" after tapping the button
- `nebula/ui/utils/MZAssetLookup.js` - we need it to be white in both dark and light mode. This adds a dictionary entry, but we're reusing existing assets

## Reference

VPN-7680

## Checklist
    
### DIFF
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -541,7 +541,7 @@ connectionInfo:
     comment: Accessible label for the button to restart the speed test upon its completion.
   rotateIPAddress:
     value: Attempt to change IP address
-    comment: Accessible label for the button to try to get a new server in the same locale
+    comment: Accessible label for the button to try to get a new server in the same location
 
 crashreporter:
   mainTitle:
--- a/src/translations/strings.yaml
+++ b/src/translations/strings.yaml
@@ -539,6 +539,9 @@ connectionInfo:
   restartSpeedTest:
     value: Restart speed test
     comment: Accessible label for the button to restart the speed test upon its completion.
+  rotateIPAddress:
+    value: Attempt to change IP address
+    comment: Accessible label for the button to try to get a new server in the same locale
 
 crashreporter:
   mainTitle:
--- a/nebula/ui/utils/MZAssetLookup.js
+++ b/nebula/ui/utils/MZAssetLookup.js
@@ -27,6 +27,10 @@ var imageLookup = {
     filenameLight: 'qrc:/nebula/resources/refresh.svg',
     filenameDark: 'qrc:/nebula/resources/refresh-dark.svg'
   },
+  'RefreshArrowsIPInfo': {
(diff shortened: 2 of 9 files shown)


src/translations/strings.yaml
```diff
@@ -539,6 +539,9 @@ connectionInfo:
   restartSpeedTest:
     value: Restart speed test
     comment: Accessible label for the button to restart the speed test upon its completion.
+  rotateIPAddress:
+    value: Attempt to change IP address
+    comment: Accessible label for the button to try to get a new server in the same locale
```
```suggestion
    comment: Accessible label for the button to try to get a new server in the same location
```

_location_, _region_? Quite certain it shouldn't be _locale_

flodolo (approved):

strseb (approved):
Ooof that screen feels so empty q_q".

In my head somehow i would have imagined that the button ends up on the right side to the ip. 
So we could build a table of 
```
ExitIP | 1.2.3.4.5 | [ Rotate ]

(and for the futute)
Ping | 12ms 
Bytes Sent | 10 PB
Byte recieved 
...

```
But anyway i'm going off topic,  r+ given this is not ready yet for public users. 

Thank you! c:

mcleinman:
> Ooof that screen feels so empty q_q".
> 
> In my head somehow i would have imagined that the button ends up on the right side to the ip. So we could build a table of
> 
> ```
> ExitIP | 1.2.3.4.5 | [ Rotate ]
> 
> (and for the futute)
> Ping | 12ms 
> Bytes Sent | 10 PB
> Byte recieved 
> ...
> ```
> 
> But anyway i'm going off topic, r+ given this is not ready yet for public users.
> 
> Thank you! c:

Oooh, I like that better. 

> Ooof that screen feels so empty q_q".
> 
> In my head somehow i would have imagined that the button ends up on the right side to the ip. So we could build a table of
> 
> ```
> ExitIP | 1.2.3.4.5 | [ Rotate ]
> 
> (and for the futute)
> Ping | 12ms 
> Bytes Sent | 10 PB
> Byte recieved 
> ...
> ```
> 
> But anyway i'm going off topic, r+ given this is not ready yet for public users.
> 
> Thank you! c:

Good call! Done! The only thing is if there are multiple lines - say IPv4 and IPv6, both will show the icon, but pressing either icon will refresh both of them. For something that isn't shown to all users, I think this is okay.

I added some code to IpAddressLookup so that the refresh icon goes away when it's working - it's only shown once it's done.

Screenshot 125

PR #11236, opened 2026-04-28
VPN-7592: Implement wireguard tunnel in the network extension

oskirby:
## Description
In the next iteration of the split tunneling work, we merge the daemon and network-extension together. This allows some significant improvements in how the tunneling operates. Most notably:
  - There is now only one privileged helper thing to install rather than two.
  - The network extension error handling can be correctly plumbed into the permission required state.
  - Split tunneling provider can be better coupled to interface up/down actions.
  - Routing logic is much simpler:
    - The wireguard interface sets `RTF_IFSCOPE` on its default route.
    - Excluded apps become a no-op (let the system do its thing).
    - Included apps get bound to the VPN interface.

Much of this work came from PR #10460 and was just updated to move it into the network extension.

## Quirks and TODO items
Because the split tunneling provider only catches TCP and UDP packet flows. There might be some traffic that can sneak passed the VPN when its active now (ICMP is the obvious case here). This could be a some future research...  possibly engaging with the macOS firewall to block any such traffic. Theoretically Apple might also have some secret sauce that can create traffic flows that skip the `NETransparentProxyProvider` so we are kind of trusting Apple a lot more with this approach.

DNS. There seems to be some contention when it comes to the DNS configuration when `NETransparentProxyProvider` is used (see [here](https://developer.apple.com/forums/thread/726368)). ~So as-of now, it doesn't seem like we have a good solution to setting the DNS config.~ To get this working, we more or less use the same approach as the old daemon and fork off a helper tool to overwrite DNS configuration through the `SystemConfiguration` framework.

No support for native multihop. This might be doable with some future work, but it complicates the API significantly. As a first cut, I decided to disable the feature for now.

## Reference
JIRA Issue: [VPN-7592](https://mozilla-hub.atlassian.net/browse/VPN-7592)
Experimental boringtun PR: #10460

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-7592]: https://mozilla-hub.atlassian.net/browse/VPN-7592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11381, opened 2026-06-20
Refactor Controller::getStatus() API to use signals

oskirby:
## Description
Breaking the `Controller::getStatus()` API refactoring work out of PR #11236 and into a standalone PR.

The gist of this is that `Controller::getStatus()` is kind of a mess. It is both brittle and difficult to change by passing a huge number of arguments, and the need to keep a list of callback methods is awkward and prone dangling pointer issues. We can simplify this by:

1. Gathering all of the status bits together into a class.
2. Report their changes using a Qt signal emission.
3. Replace the callback hook thing with a `Qt::SingleShotConnection`.

## Reference
Split from: #11236

## Checklist
    
### DIFF
--- a/src/controller.h
+++ b/src/controller.h
@@ -36,6 +36,8 @@ class ControllerStatus {
   ControllerStatus() {}
   ControllerStatus(const QJsonObject& obj);
 
+  void clear();
+
   bool m_connected;
   QDateTime m_timestamp;
   QHostAddress m_ipv4Address;
--- a/src/controller.h
+++ b/src/controller.h
@@ -38,14 +38,14 @@ class ControllerStatus {
 
   void clear();
 
-  bool m_connected;
+  bool m_connected = false;
   QDateTime m_timestamp;
   QHostAddress m_ipv4Address;
   QHostAddress m_ipv6Address;
   QHostAddress m_ipv4Gateway;
   QHostAddress m_ipv6Gateway;
-  quint64 m_rxBytes;
-  quint64 m_txBytes;
+  quint64 m_rxBytes = 0;
+  quint64 m_txBytes = 0;
 };
 
 class Controller : public QObject, public LogSerializer {
--- a/src/controller.h
+++ b/src/controller.h
@@ -20,6 +20,32 @@
 class Controller;
 class ControllerImpl;
 
+class ControllerStatus {
+  Q_GADGET
+
+  Q_PROPERTY(bool connected MEMBER m_connected);
+  Q_PROPERTY(QDateTime timestamp MEMBER m_timestamp);
+  Q_PROPERTY(QHostAddress ipv4Address MEMBER m_ipv4Address);
+  Q_PROPERTY(QHostAddress ipv6Address MEMBER m_ipv6Address);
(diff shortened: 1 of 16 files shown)


src/platforms/android/androidcontroller.cpp
```diff
@@ -76,10 +76,13 @@ AndroidController::AndroidController() {
...
-        emit statusUpdated(doc.object()["endpoint"].toString(),
-                           doc.object()["deviceIpv4"].toString(),
-                           doc.object()["tx_bytes"].toInt(),
-                           doc.object()["rx_bytes"].toInt());
+        ControllerStatus st;
+        st.m_connected = true;
+        st.m_ipv4Gateway = QHostAddress(doc.object()["endpoint"].toString());
+        st.m_ipv4Address = QHostAddress(doc.object()["deviceIpv4"].toString());
+        st.m_rxBytes = doc.object()["tx_bytes"].toInteger();
+        st.m_txBytes = doc.object()["rx_bytes"].toInteger();
```
encrypt94: rx/tx bytes are swapped

src/controller.h
```diff
@@ -20,6 +20,34 @@
...
+  void clear();
+
+  bool m_connected;
+  QDateTime m_timestamp;
+  QHostAddress m_ipv4Address;
+  QHostAddress m_ipv6Address;
+  QHostAddress m_ipv4Gateway;
+  QHostAddress m_ipv6Gateway;
+  quint64 m_rxBytes;
+  quint64 m_txBytes;
```
encrypt94: NIT: i think it's better to have a safe default and set 
```c++
bool m_connected = false;
quint64 m_rxBytes = 0;
quint64 m_txBytes = 0;
````
in the remote case `m_status` it's accessed before the first `statusUpdated` signal emission.

encrypt94 (changes requested):

src/platforms/android/androidcontroller.cpp
oskirby: Good catch

encrypt94 (approved):
LGTM! Thank you! :)

mcleinman:
I know this is merged, but are there any places that rxbytes or txbytes are used anymore? Are they candidates for removal?

oskirby:
> I know this is merged, but are there any places that rxbytes or txbytes are used anymore? Are they candidates for removal?

To my knowledge, nothing uses the rx/tx counters anymore.

Once upon a time we used to graph them in the connection information screen. Then the graphs got removed as a part of the speedtesting feature and nothing was ever put in its place when the speedtest feature was deprecated.

mcleinman:
> > I know this is merged, but are there any places that rxbytes or txbytes are used anymore? Are they candidates for removal?
> 
> To my knowledge, nothing uses the rx/tx counters anymore.
> 
> Once upon a time we used to graph them in the connection information screen. Then the graphs got removed as a part of the speedtesting feature and nothing was ever put in its place when the speedtest feature was deprecated.

We also used it for a minute for telemetry, putting them in histogram buckets. You good if I put up a PR to remove them?

oskirby:
> We also used it for a minute for telemetry, putting them in histogram buckets. You good if I put up a PR to remove them?

Go for it!

PR #11382, opened 2026-06-20
Nit: Make NSError loggable

oskirby:
## Description
The `NSError` type is one of the macOS Foundation error reporting types, and it can be toll-free-bridged into the already-supported CFErrorRef type. So let's add the appropriate logging wrapper to support it in the `Logger` class.

While we're at it, we are also adding the trailing space for the other macOS logging types to keep them consistent with the rest of the logger class.

This should fix a couple of error logs where we print out a raw pointer rather than the error message.

## Reference
Split from: #11236 

## Checklist
    

mcleinman (approved):
Thanks

PR #11385, opened 2026-06-23
VPN-7684: fix onboarding icon color

mcleinman:
## Description

This is from [the removal of Qt compat effects](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11153). Turns out colorization just bases it off the icon's color, unless we give it a brightness. This was only happening on 2 of the 4 onboarding icons as the other 2 were white, and thus applied the color as expected. 

The icon in Server is a light enough color that it didn't make a difference, but I added it there for good measure as well.  These are the only two spots in our code that have `colorization`.

Fixed:
Screenshot 131
Screenshot 130
Screenshot 129

## Reference

VPN-7684

## Checklist
    
### DIFF
--- a/nebula/ui/components/MZStepProgressBarDelegate.qml
+++ b/nebula/ui/components/MZStepProgressBarDelegate.qml
@@ -99,7 +99,7 @@ Column {
 
             MultiEffect {
                 anchors.fill: parent
-
+                brightness: 1.0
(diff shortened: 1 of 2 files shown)


strseb (approved):
Ohhhhh sorry i didnt catch that diffrence.

PR #11388, opened 2026-06-24
VPN-7678: add text to widgets when user isn't signed in

mcleinman:
## Description

Adds a default text to widgets when the user is signed out, or doesn't have a network extension.

IMG_0077 IMG_0076

If a user removes the network extension via iOS system settings, the widgets will not immediately update. In other cases - for instance, if they sign out - it will.

## Reference

VPN-7678

## Checklist
    

src/translations/strings.yaml
```diff
@@ -1147,6 +1147,9 @@ toggleWidget:
   description:
     value: Turn Mozilla VPN on and off, and see the current location.
     comment: More information about what the Mozilla VPN widget contains.
+  unauthenticatedMessage:
+    value: Before using widget, first turn on Mozilla VPN from app.
```
flodolo: I know this is temporary, but pointing out that it sounds broken. "before using **this** widget"?

src/translations/strings.yaml
```diff
@@ -1158,6 +1161,9 @@ logoAccessoryWidget:
   vpnOff:
     value: Mozilla VPN off
     comment: Shown as part of inline widget when VPN is inactive.  If translation will be more than 16 characters, instead translate `VPN on` to keep this within available space.
+  unauthenticatedMessage:
+    value: Turn on via app
+    comment: Shown in small widgets when user is not signed in, or has rmoeved the network extension. For the inline widget, this must be kept to 16 characters or less. If translation will be more than 16 characters, instead translate `Open app` to keep this within available space.
```
```suggestion
    comment: Shown in small widgets when user is not signed in, or has removed the network extension. For the inline widget, this must be kept to 16 characters or less. If translation will be more than 16 characters, instead translate as `Open app` to keep this within available space.
```

mcleinman: Thanks

src/translations/strings.yaml
mcleinman: Good call, updated.

mcleinman:
Confirmed strings w/ Lesley

flodolo (approved):

artfwo (approved):

PR #11389, opened 2026-06-24
VPN-7659: Initialize localizer before using translations

mcleinman:
## Description

While this was recently fixed in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/10939, it was reintroduced in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11041/ when we improved localizer behavior.

I think this is a hacky fix, and I'm open to better solutions. 

Before / after
IMG_0078 IMG_0079 

## Reference

VPN-7659

## Checklist
    
### DIFF
--- a/src/ipaddresslookup.cpp
+++ b/src/ipaddresslookup.cpp
@@ -10,6 +10,7 @@
 
 #include "controller.h"
 #include "leakdetector.h"
+#include "localizer.h"
 #include "logger.h"
 #include "mozillavpn.h"
 #include "tasks/ipfinder/taskipfinder.h"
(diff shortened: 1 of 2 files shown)


src/ipaddresslookup.cpp
```diff
@@ -43,6 +44,13 @@ void IpAddressLookup::reset() {
   logger.debug() << "Resetting the data";
 
   if (m_state != StateWaiting) {
+    // On iOS (and possibly Android), we need to ensure the Localizer
+    // is initialized before setting it to the loading string ID in
+    // a few lines. This is a hacky way to do it, but it works.
+    // See VPN-7659 for more info.
+    bool unusedBooleanUsedForInitialization =
+        Localizer::instance()->isRightToLeft();
```
encrypt94: Thank you for spotting this regression!
is the call to `isRightToLeft()` necessary? Can't we just call `Localizer::instance()` to initialize the Localizer?

mcleinman: It is not - I didn't realize we could just call the instance. Just updated the code and tested it. Thanks.

encrypt94 (approved):
LGTM! thanks :)

PR #11400, opened 2026-06-26
VPN-7688: donate app intents on iOS

mcleinman:
## Description

Apple recommends we "donate" app intents - tell the system when the user performs these actions manually. iOS then associates when/where/etc., and is more likely to surface our app intents as Siri suggestions. Of course, none of this is guaranteed.

Example: Someone opens the app and turns on the VPN every day when they arrive at work. Siri - sharing none of this with us - realizes that this happens at their work location, and may start surfacing the “Turn On VPN” suggestion when they are at work.

We should not donate app intents when the action came from the actual intent. I believe both these routes meet this requirement.

I did not create app intent donations for toggle intent (which is used under the hood for the control widgets, so makes little sense here) nor the "get current status" intent (which is mostly there to be used in Shortcuts, if desired).

## Reference

VPN-7688

## Checklist
    

oskirby (approved):
This seems simple enough.

oskirby (approved):

PR #11406, opened 2026-06-30
Remove major language flag in AddonConditionWatcherLocales

mcleinman:
## Description

`CheckMajorLanguageCode` isn’t really used. Because up until VPN-7665, we’ve hardly ever used the “locales” condition for addons. So we’ve never really used `AddonConditionWatcherLocales` or cared about which of the two major language code variants were used in deciding whether there was a match. Let's remove them.

In https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11367, I was confused about the intended behavior for this. I changed it, and then changed it back. (The crux: If the major part of the current app locale matches a locale that the addon should be activated for based on addon conditions, then activate it. That is, the addon must list plain “es” as an addon locale to match when a user’s locale is “es_MX”. If the addon lists “es_AR” as a locale it should be activated it, it will be turned off if the user selects “es_MX”. Though some might expect that matching major locale would mean "es_AR" as an app language would activate an addon with a locale condition including "es_MX".)

Going back further, this flag has caused [other bugs](https://mozilla-hub.atlassian.net/browse/VPN-5569). And frustratingly, we've essentially never used this feature.

I think we should rip out this variant:
- We have essentially never used this functionality.
- We don’t have too many locales with country variants. If we remove this, for an example in Spanish, we’d change from listing “es” in the locales list to “es_MX”, “es_AR”, “es_ES”, and “es_CL”. IMO: This is clearer, and more straightforward. Especially since we don’t have a plain “es” language that we translate to, which is what we'd include if we were using the "major language" variant flag to match Spanish variants.
- How it is designed to work is confusing, at least for me - see above, and I missed the goal on my first read.
- We have fallback locales for addons (and main app), which seems like it covers this situation pretty well.

Currently, there is just one addon in prod that uses the locales condition. It’s `message_fpn_trial_subscription_expiring`, which is only shown to folks on an Firefox Private Network trial. There should be zero people who see this addon, as FPN has been dead for years. I spotchecked what addons were around in 2.20, which went out 2.5 years ago - no other ones in that time used “locales” as a condition. I think this is one of those things where code was written, but we almost never utilized it outside of testing. And so the “check major locale code” / “don’t check major locale code” difference has never really mattered.

## Reference

N/A

## Checklist
    
### DIFF
--- a/src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
+++ b/src/addons/conditionwatchers/addonconditionwatcherlocales.cpp
@@ -10,22 +10,17 @@
 
 // static
 AddonConditionWatcher* AddonConditionWatcherLocales::maybeCreate(
-    QObject* parent, const QStringList& locales,
-    MajorLanguageCodePolicy majorLanguageCodePolicy) {
+    QObject* parent, const QStringList& locales) {
   if (locales.isEmpty()) {
     return nullptr;
   }
 
-  return new AddonConditionWatcherLocales(parent, locales,
-                                          majorLanguageCodePolicy);
+  return new AddonConditionWatcherLocales(parent, locales);
 }
 
 AddonConditionWatcherLocales::AddonConditionWatcherLocales(
-    QObject* parent, const QStringList& locales,
-    MajorLanguageCodePolicy majorLanguageCodePolicy)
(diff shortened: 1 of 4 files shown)


encrypt94 (approved):
LGTM

PR #11408, opened 2026-07-01
VPN-7690 part 2: Move to modern translation type

mcleinman:
## Description

QA mentioned they still don't see it. And I realized we're not seeing any stage-only languages. 

This is identical to https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11363 - which fixed this less than 3 weeks ago. And the regression was caused by me in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11389, as I initialized the localizer (which causes settings to come online) earlier.

Since this is a near-immediate regression (within 3 weeks of my fix), let's do a stronger fix. Let's move this to the modern translation type. From my testing, this does both the things:
- "Loading" is shown, instead of the string ID.
- We set staging in an appropriate spot when opening the app - before the localizer turns on. Which means  we're appropriately-showing stage-only languages.

## Reference

VPN-7690

## Checklist
    
### DIFF
--- a/src/mozillavpn_p.h
+++ b/src/mozillavpn_p.h
@@ -22,7 +22,6 @@
 #include "networkwatcher.h"
 #include "releasemonitor.h"
 #include "serverlatency.h"
-#include "settingsholder.h"
 #include "statusicon.h"
 
 struct MozillaVPNPrivate {
--- a/src/ipaddresslookup.cpp
+++ b/src/ipaddresslookup.cpp
@@ -44,14 +44,8 @@ void IpAddressLookup::reset() {
   logger.debug() << "Resetting the data";
 
   if (m_state != StateWaiting) {
-    // VPN-7659: On iOS and Android, we need to ensure the Localizer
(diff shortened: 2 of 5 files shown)


mcleinman:
@flodolo We're moving from historic to modern translation type. Will this require a re-translation? Or is adding it to the yaml file enough?

src/mozillavpn_p.h
```diff
@@ -22,7 +22,6 @@
 #include "networkwatcher.h"
 #include "releasemonitor.h"
 #include "serverlatency.h"
-#include "settingsholder.h"
```
mcleinman, line 25: I noticed this was no longer used here, so I removed it.

flodolo:
> @flodolo We're moving from historic to modern translation type. Will this require a re-translation? Or is adding it to the yaml file enough?

It requires retranslation, but that's OK.

flodolo (approved):

artfwo (approved):

PR #11420, opened 2026-07-06
VPN-7681: 2.39 update and whats new messages

mcleinman:
## Description

Messages for 2.39 cycle:

Screenshot 138Screenshot 137

## Reference

VPN-7681

## Checklist
    

addons/strings.yaml
```diff
@@ -236,10 +236,19 @@ commonStrings:
...
+
+2026q3:
   # vpn7655:
     # value: This is an example. Draft a release note for your feature. It's ID should be the Jira issue. Please leave it as comment to allow for edits.
     # comment: Bullet point with a specific update. Should have tag that indicates it applies to [all/iOS/Linux/whichever] platform.
+  vpn7646: # Used in 2.39
+    value: "[iOS] Even more widgets are now available! Lock screen widgets are included in this release."
+    comment: Bullet point with a specific update. Should have tag that indicates it applies to iOS platform.
+  pr11382: # No Jira ticket, so using PR for identifier. Used in 2.39
+    value: "[macOS] Improved logging of some errors, allowing easier diagnoses of problems."
```
```suggestion
    value: "[macOS] Improved logging of some errors, allowing easier diagnosis of problems."
```

Don't think we want plural here?

addons/strings.yaml
```diff
@@ -236,10 +236,19 @@ commonStrings:
...
-#2026q3:
+# Has not been used yet (from above)
+# 237updateMessage.bullet1
+
+2026q3:
   # vpn7655:
     # value: This is an example. Draft a release note for your feature. It's ID should be the Jira issue. Please leave it as comment to allow for edits.
     # comment: Bullet point with a specific update. Should have tag that indicates it applies to [all/iOS/Linux/whichever] platform.
+  vpn7646: # Used in 2.39
+    value: "[iOS] Even more widgets are now available! Lock screen widgets are included in this release."
```
flodolo: Not an iOS user: are `Lock screen widgets` widgets only available on the lock screen, or now all widgets are available also on the lock screen? I'm afraid we won't get good translations for these with the current text.

lesleyjanenorton: it may be that this should actually be 'Lock Screen' per https://developer.apple.com/design/human-interface-guidelines/widgets/

lesleyjanenorton:
```suggestion
    value: "[iOS] Even more widgets are now available! Lock Screen widgets are included in this release."
```

lesleyjanenorton: Updated

lesleyjanenorton: @flodolo All widgets are not available on the Lock Screen.

flodolo (approved):

PR #11422, opened 2026-07-08
Fix shutdown semaphore signaling in network extension

oskirby:
## Description
Just noticing a little quirk in the network extension. When restarting the extension for any reason, the extension hangs for about 5 seconds during the shutdown process. This seems to be caused by a missing semaphore signal when the worker thread exits.

This PR should address that and ensure the shutdown happens promptly.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

encrypt94 (approved):
LGTM!

PR #11423, opened 2026-07-08
VPN-7101: Update system settings links for network extension

oskirby:
## Description
When installing the macOS network extension, there are some steps that the user must navigate in order to grant permissions for the network extension to be installed. Until those steps are complete, the application is blocked in the `Permission Required` state and a screen is displayed to the user. This screen includes a clickable link to open the system settings screen.

This PR attempts to update the clickable links to get the user closer to the appropriate settings screen. There are three cases we need to address here:
- Network Extension feature disabled: In this case the daemon is installed via the `SMAppService` API, which includes a helper method to direct the user to the correct screen.
- macOS 15 and later: The user needs to be directed to the `Extensions` section of the `Login Items` settings panel. On this panel they will need to find the `Mozilla VPN Network Extension` and enable it. The exact look and feel of this panel varies by OS version.
- macOS 13-14: The system extension is blocked and a message is displayed in the `Security` settings panel. The user needs to explicitly click `Allow` to unblock the installation process.

## Reference
JIRA Issue: [VPN-7101](https://mozilla-hub.atlassian.net/browse/VPN-7101)

## Checklist
    
[VPN-7101]: https://mozilla-hub.atlassian.net/browse/VPN-7101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/macos/macosutils.mm
+++ b/src/platforms/macos/macosutils.mm
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "macosutils.h"
+
+#include "feature/features.h"
 #include "logger.h"
 #include "qmlengineholder.h"
 
@@ -12,6 +14,7 @@
 #include <objc/objc.h>
 
 #import <Cocoa/Cocoa.h>
+#import <Foundation/Foundation.h>
 #import <ServiceManagement/ServiceManagement.h>
 
 namespace {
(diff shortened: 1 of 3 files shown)


src/platforms/macos/macosutils.mm
```diff
@@ -33,8 +36,26 @@
...
 }
 
-void MacOSUtils::openSystemSettingsLoginItems() {
-  [SMAppService openSystemSettingsLoginItems];
+void MacOSUtils::openSystemSettingsLink() {
+  if (!Feature::isEnabled(Feature::networkExtension)) {
+    // When using a daemon installed via the SMAppService API, there is a
+    // helper method to go directly to the appropriate settings screen.
+    [SMAppService openSystemSettingsLoginItems];
+  } else if (@available(iOS 15.0, *)) {
```
oskirby: Whoops! This should be testing `macOS` not `iOS`...  give me a sec to fix that.

src/platforms/macos/macosutils.mm
```diff
@@ -33,8 +36,26 @@
...
+    // Users on macOS 15 and later: System extensions are managed via a special
+    // section in the Login Items panel. We can navigate directly to the
+    // extensions section and, from there, the user must find the Mozilla VPN
+    // network extension and enable it.
+    NSString* url = @"x-apple.systempreferences:com.apple.LoginItems-Settings.extension?ExtensionItems";
+    [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:url]];
+  } else {
+    // Users on macOS 13/14: When the system extension is blocked, and a message
+    // along the lines of "System software from application 'Mozilla VPN' was
+    // blocked from loading" will be displated in the Security panel. The users
```
encrypt94: typo :)

src/platforms/macos/macosutils.mm
```diff
@@ -33,8 +36,26 @@
...
+    // When using a daemon installed via the SMAppService API, there is a
+    // helper method to go directly to the appropriate settings screen.
+    [SMAppService openSystemSettingsLoginItems];
+  } else if (@available(macOS 15.0, *)) {
+    // Users on macOS 15 and later: System extensions are managed via a special
+    // section in the Login Items panel. We can navigate directly to the
+    // extensions section and, from there, the user must find the Mozilla VPN
+    // network extension and enable it.
+    NSString* url = @"x-apple.systempreferences:com.apple.LoginItems-Settings.extension?ExtensionItems";
+    [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:url]];
```
encrypt94: can we log if openURL fails to open? it will probably never happen but if the link changes in a future version of macOS and someone installs an older version of the client, we'd at least have some evidence in the logs

encrypt94 (approved):
r+wc

src/platforms/macos/macosutils.mm
oskirby: I have added some logs, but it is worth noting that any URL with the scheme `x-apple.systempreferences` will open the system settings. So, if this URL changes in the future it will still open something successfully and will probably just navigate to the wrong place, and thus the log is unlikely to ever get hit.

artfwo (approved):

PR #11424, opened 2026-07-08
Obfuscation features part 2 - Add LWO obfuscator

encrypt94:
## Description

Add LWO v1 and v2 obfuscators.

Protocol version selection is opaque and it's done in the `Controller`:
- select a server that supports LWO (v1 or v2).
- check whether the `Controller` supports v2
- if v2 is available, set lwoVersion to v2 in `InterfaceConfig`
- use v1 otherwise

LWO v2 is currently supported only on a single staging server.

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7567)

## Checklist
    
### DIFF
--- a/obfuscators/src/lwo/v1.rs
+++ b/obfuscators/src/lwo/v1.rs
@@ -1,0 +1,199 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! LWO v1 packet transformation.
+
+use std::sync::Arc;
+
+use rand::rngs::OsRng;
+use rand::RngCore;
+use tokio::net::UdpSocket;
+
+use super::{
+    WgAddr, COOKIE_REPLY, COOKIE_REPLY_SZ, DATA, DATA_OVERHEAD_SZ, HANDSHAKE_INIT,
+    HANDSHAKE_INIT_SZ, HANDSHAKE_RESP, HANDSHAKE_RESP_SZ, MAX_PACKET,
+};
+
+/// Bit to set in the second byte of the WG header to enable LWO
+const OBFUSCATION_BIT: u8 = 0b10000000;
+
+pub async fn outbound(local: Arc<UdpSocket>, remote: Arc<UdpSocket>, wg_addr: WgAddr, key: [u8; 32]) {
+    let mut buf = vec![0u8; MAX_PACKET];
+    let mut rng = OsRng;
+    loop {
+        let Ok((n, src)) = local.recv_from(&mut buf).await else {
+            break;
+        };
+        *wg_addr.lock().unwrap() = Some(src);
+        obfuscate(&mut rng, &mut buf[..n], &key);
+        let _ = remote.send(&buf[..n]).await;
+    }
+}
+
+pub async fn inbound(local: Arc<UdpSocket>, remote: Arc<UdpSocket>, wg_addr: WgAddr, key: [u8; 32]) {
+    let mut buf = vec![0u8; MAX_PACKET];
+    loop {
+        let Ok(n) = remote.recv(&mut buf).await else {
+            break;
+        };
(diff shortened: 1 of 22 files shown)


src/controller.cpp
```diff
@@ -525,9 +535,20 @@ auto Controller::setupConfigs(
...
       entryConfig.m_serverPort = 53;
     }
-    // If UDP over TCP is enabled choose a dedicated port
-    if (obfuscationMethod == Server::ObfuscationMethod::UdpOverTcp) {
-      entryConfig.m_serverPort = entryServer.chooseTcpPort();
+
+    switch (obfuscationMethod) {
+      case Server::ObfuscationMethod::UdpOverTcp:
+        // If UDP over TCP is enabled choose a dedicated port
+        exitConfig.m_serverPort = exitServer.chooseTcpPort();
```
artfwo: Should this be `entryServer.chooseTcpPort()` instead, both here and below?

obfuscators/src/lwo/v1.rs
```diff
@@ -0,0 +1,214 @@
...
+        let Some(addr) = *wg_addr.lock().unwrap() else {
+            continue;
+        };
+        deobfuscate(&mut buf[..n], &key);
+        let _ = local.send_to(&buf[..n], addr).await;
+    }
+}
+
+pub fn obfuscate(rng: &mut impl RngCore, packet: &mut [u8], key: &[u8; 32]) {
+    log::error!("Obfuscating packet of size {}", packet.len());
```
artfwo: Per-packet logging at error level might bloat the logs very quickly, `log::info!` in `deobfuscate` also looks a bit excessive, are these necessary?

artfwo (changes requested):

obfuscators/src/lwo/v1.rs
encrypt94: not necessary, this is a leftover debug log! thanks for spotting it!

src/daemon/daemon.cpp
```diff
@@ -385,6 +385,17 @@ bool Daemon::parseConfig(const QJsonObject& obj, InterfaceConfig& config) {
...
+  if (!obj.contains("lwoVersion")) {
+    config.m_lwoVersion = 1;
+  } else {
+    QJsonValue value = obj.value("lwoVersion");
+    if (!value.isDouble()) {
+      logger.error() << "lwoVersion is not a number";
+      return false;
+    }
+    config.m_lwoVersion = value.toInt();
+  }
```
artfwo, line 402: Given that every other field is validated in `parseConfig`, should we probably reject anything other than 1 and 2?

src/daemon/obfuscator/qprocessobfuscator.cpp
```diff
@@ -112,7 +105,22 @@ QStringList QProcessObfuscator::buildArgs(const InterfaceConfig& config) {
...
+    case Server::ObfuscationMethod::LWO:
+      args << QStringLiteral("lwo");
+      args << QStringLiteral("--lwo-version")
+           << QStringLiteral("v%1").arg(config.m_lwoVersion);
+      args << QStringLiteral("--public-key") << config.m_publicKey;
+      args << QStringLiteral("--server-public-key") << config.m_serverPublicKey;
+      break;
+    default:
+      return {};
+  }
```
artfwo, line 122: Nit: when keys are passed as command-line arguments, device and server WireGuard public keys will appear in `/proc/*/cmdline` for every local user. They aren't secrets, but I wonder if we should pass them through environment or stdin instead?

artfwo (approved):

src/daemon/obfuscator/qprocessobfuscator.cpp
encrypt94: Thanks for pointing this out! I don't think this is introduces a confidentiality or security problem.
Both are public keys, and the server key is also publicly available using the server IP, which is itself exposed as a command-line argument. So i'd prefer to keep them as command-line arguments for simplicity and consistency.
That said, happy to change this if there's a broader concern with exposing pubkeys as arguments.

PR #11427, opened 2026-07-09
VPN-7670 Add aarch64 deb packages to archive.mozilla.org and packages.mozilla.org

encrypt94:
## Description

Publish aarch64 packages to archive.mozilla.org and packages.mozilla.org

## Reference

[VPN-7670](https://mozilla-hub.atlassian.net/browse/VPN-7670)

## Checklist
    
[VPN-7670]: https://mozilla-hub.atlassian.net/browse/VPN-7670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/taskcluster/kinds/beetmover-apt/kind.yml
+++ b/taskcluster/kinds/beetmover-apt/kind.yml
@@ -16,9 +16,6 @@ task-defaults:
     description: Import .deb package into Google Artifact Registry
     run-on-tasks-for: [action]
     from-deps:
-        with-attributes:
-            build-type:
-                - linux64/release-deb
         set-name: null
     worker:
         product: vpn
@@ -27,10 +24,36 @@ task-defaults:
 tasks:
     promote-linux64-deb:
         worker-type: beetmover-apt-stage
+        from-deps:
+            with-attributes:
+                build-type:
+                    - linux64/release-deb
+        attributes:
+            shipping-phase: promote-client
+
+    promote-linux-aarch64-deb:
+        worker-type: beetmover-apt-stage
+        from-deps:
+            with-attributes:
+                build-type:
+                    - linux-aarch64/release-deb
         attributes:
             shipping-phase: promote-client
 
     ship-linux64-deb:
         worker-type: beetmover-apt
+        from-deps:
+            with-attributes:
+                build-type:
+                    - linux64/release-deb
+        attributes:
+            shipping-phase: ship-client
+
+    ship-linux-aarch64-deb:
+        worker-type: beetmover-apt
+        from-deps:
+            with-attributes:
+                build-type:
+                    - linux-aarch64/release-deb
         attributes:
             shipping-phase: ship-client
--- a/taskcluster/kinds/beetmover-apt/kind.yml
+++ b/taskcluster/kinds/beetmover-apt/kind.yml
@@ -5,7 +5,6 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
-    - taskgraph.transforms.from_deps
     - mozillavpn_taskgraph.transforms.beetmover_apt
     - taskgraph.transforms.task:transforms
 
@@ -15,8 +14,6 @@ kind-dependencies:
 task-defaults:
     description: Import .deb package into Google Artifact Registry
     run-on-tasks-for: [action]
-    from-deps:
-        set-name: null
     worker:
         product: vpn
         max-run-time: 1800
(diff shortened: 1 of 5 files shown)


taskcluster/kinds/beetmover-apt/kind.yml
```diff
@@ -27,10 +24,36 @@ task-defaults:
...
+                    - linux-aarch64/release-deb
         attributes:
             shipping-phase: promote-client
 
     ship-linux64-deb:
         worker-type: beetmover-apt
+        from-deps:
+            with-attributes:
+                build-type:
+                    - linux64/release-deb
```
oskirby: IIUC, the `from-deps` transform takes the task and duplicates it for each dependency. This is how the `build-signing` transorm works. So I think you only need a single `ship-linux64-deb` and then you would specify multiple `build-type` values for each thing that needs to be shipped this way.

For example:
```
ship-linux64-deb:
   worker-type: beetmover-apt
   from-deps:
      with-attributes:
         build-type:
            - linux64/release-deb
            - linux-aarch64/release-deb
```

encrypt94: tried that it apparently fails with `Exception: duplicate tasks with label beetmover-apt-ship-linux64-deb`

taskcluster/kinds/beetmover-promote/kind.yml
```diff
@@ -97,6 +97,13 @@ tasks:
             build: build-linux64/release-deb
         attributes:
             build-type: linux64/release-deb
+    linux-aarch64-deb:
+        run-on-tasks-for: [action]
+        release-artifacts: [mozillavpn_arm64.deb]
+        dependencies:
+            build: build-linux-aarch64/release-deb
+        attributes:
+            build-type: linux-aarch64/release-deb
```
ahal, line 106: This is just a suggestion, would recommend for a follow-up. But there seems to be a lot of duplication in this file. You could use the [matrix transforms](https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/matrix.html) to help reduce some of it if you wanted. Works very similarly to the Github Actions matrix concept.

taskcluster/kinds/beetmover-promote/kind.yml
```diff
@@ -97,6 +97,13 @@ tasks:
             build: build-linux64/release-deb
         attributes:
             build-type: linux64/release-deb
+    linux-aarch64-deb:
+        run-on-tasks-for: [action]
```
ahal, line 101: Just as an fyi, this is a no-op. Fine to leave it though, it acts like a comment of sorts.

taskcluster/kinds/beetmover-apt/kind.yml
ahal: I believe the duplicate task exception is because you have `set-name: null` in the `task-defaults` section. Normally `from-deps` attempts to create a new label automatically, but that key tells it not to (under the assumption that you'll handle that yourself in some later transform). So try removing that key?

Otherwise if you're just using `from-deps` to add a single dependency, it's a bit overkill. You could just set the `dependencies` key manually instead and avoid `from-deps` entirely.

taskcluster/kinds/beetmover-apt/kind.yml
```diff
@@ -27,10 +24,36 @@ task-defaults:
 tasks:
     promote-linux64-deb:
```
ahal, line 22: This is a pre-existing issue, so this is just a drive-by comment and not something to solve here.. but having separate tasks for promote / ship seems wrong. Normally there's a single task that has a `shipping-phase` attribute that gets set to either "promote" or "ship" depending on the context. Then the behaviour of the task is adjusted accordingly.

ahal (changes requested):
This looks good. Requesting changes for the `from-deps` comment. The other comments are all drive-by suggestions of pre-existing issues.

taskcluster/kinds/beetmover-apt/kind.yml
encrypt94: I think part of the issue was that `taskcluster/mozillavpn_taskgraph/transforms/beetmover_apt.py` was only getting the primary dependency. I changed it to use get_dependencies

taskcluster/scripts/build/linux_build_dpkg.sh
```diff
@@ -148,5 +148,11 @@ for FILENAME in ${BUILD_ARTIFACTS}; do
   if echo "${PACKAGE_NAME}" | grep -q -e "-dbgsym$"; then
     PACKAGE_EXT="ddeb"
   fi
-  cp -v "${FILENAME}" "/builds/worker/artifacts/${PACKAGE_NAME}.${PACKAGE_EXT}"
```
oskirby, line 151: Non-blocking comment/gripe: I would prefer that we insert the architecture into the filename for all platforms and not just the cross-compiled ones. Just keep it consistent between x86_64 and aarch64.

But, that might cause an whole lot of pain downstream if we have to rename things for the x86_64 release, so feel free to ignore this comment if that happens to be difficult.

oskirby (approved):
LGTM, but I would defer to @ahal for the final word on this.

taskcluster/scripts/build/linux_build_dpkg.sh
encrypt94: same, but i'd like to investigate the impact a little more before making that change. I'm a bit worried about the potential side effects, so i'd prefer to keep this as-is for now :)

ahal (approved):
Lgtm!

PR #11428, opened 2026-07-10
VPN-7101: Update permission required title image and button

oskirby:
## Description
To better guide macOS users through the installation of the Mozilla VPN network extension, we will need to apply some updates to the permission required screen that match up with the installation flow. This PR attempts to tweak the title image and system settings link on this screen accordingly.

Since the number of title image variations are getting out of hand, I decided now was a good time to refactor this a little. Rather than localizing the image for each possible case, I remove the title text from the SVG file, and instead use a text overlay that can be handled through our regular localization pipeline.

When the network extension feature is enabled on macOS 15, the resulting experience looks like this:
C46D28A5-F0AF-43E4-B2A6-A018BF0CA472

## Reference
JIRA Issue: [VPN-7101](https://mozilla-hub.atlassian.net/browse/VPN-7101)

## Checklist
    
[VPN-7101]: https://mozilla-hub.atlassian.net/browse/VPN-7101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ui/sharedViews/ViewPermissionRequiredOSX.qml
```diff
@@ -28,8 +28,38 @@ MZFlickable {
...
+            source: MZAssetLookup.getImageSource("MacosPermissionGeneric")
             fillMode: Image.PreserveAspectFit
+
+            Text {
+                id: imageTitle
+
+                text: getImageTitle()
+                horizontalAlignment: Text.AlignLeft
+
+                color: MZTheme.colors.useDarkAssets ? "#DFDEDE" : "#1E1E1E"
```
oskirby: To explain, these color codes were pulled out of the SVG, and there doesn't seem to be a match in the `colors.js`, so I opted to use raw hex codes here to keep it consistent with the "Mozilla VPN" text embedded in the SVG. This could be cleaned up by either:
1. Adding them to `colors.js`
2. Replacing the `Mozilla VPN` text with an overlay too, and then just using the closest theme colors.
3. Ignoring it for this one specific case.

oskirby: I went with option 2.

src/translations/strings.yaml
```diff
@@ -988,6 +988,15 @@ permissionMacos:
...
     comment: Paragraph requesting the user to allow the VPN to run in the background.
+  imageTitleBackground:
+    value: Allow in the Background
+    comment: Title in the permission required image, showing the Allow in the Background section of Login Items screen.
+  imageTitleBackgroundMacOS26:
+    value: App Background Activity
+    comment: Title in the permission required image, showing the App Background Acitivty section of the Login Items screen for macOS 26 and later.
+  imageTitleNetworkExtension:
+    value: Network Extensions
+    comment: Title in the permission required image for network extension installation.
```
mcleinman: We want these to be exactly what it would be in macOS system settings, right? If so, we need to be really explicit for translators.

Even if we are really specific, some translators don't have macs. And even those that do have macs, won't have all the versions required to get all 3.

Is there any canonical apple source (their help files, localized?) that you've found that we can use here? If not, is it worth restructuring this to remove the header in the image, and say something like "depending on which version of macOS you're on, this is under App Backgorund Activity, Network Extensions, or Allow in the Background"?

src/translations/strings.yaml
```diff
@@ -1003,6 +1012,9 @@ permissionMacos:
   openSettingsButtonLabelMacOS15:
     value: Open Login Items & Extensions
     comment: Label for the button that opens the login items settings page on macOS 15 and above
+  openSettingsButtonLabelGeneric:
+    value: Open System Preferences
```
mcleinman: should this be `Open System Settings`?

src/ui/sharedViews/ViewPermissionRequiredOSX.qml
mcleinman: My preference would be option 1, for future reference. Whatever you have at this point is fine, though. As you've found, option 3 causes a linter failure.

src/ui/sharedViews/ViewPermissionRequiredOSX.qml
```diff
@@ -28,8 +28,58 @@ MZFlickable {
...
+            source: MZAssetLookup.getImageSource("MacosPermissionGeneric")
             fillMode: Image.PreserveAspectFit
+
+            Text {
+                id: imageTitle
+
+                text: getImageTitle()
+                horizontalAlignment: Text.AlignLeft
+
+                color: MZTheme.colors.useDarkAssets ? MZTheme.colors.grey15 : MZTheme.colors.grey55
```
mcleinman: Linter is failing, as you can't call colors from code directly. These need to be added in `dark-mode.js` and `main.js`, calling them the same new variable name (if there isn't a variable that already uses these colors - I wouldn't expect there to be). Then use that variable name in this spot.

(This is to future-proof colors to more easily allow future additional themes, without the full audits/changes we had to do to add dark mode.)

mcleinman (changes requested):

src/ui/sharedViews/ViewPermissionRequiredOSX.qml
oskirby: I am noticing that, yeah. The closest existing theme color would be `MZTheme.colors.fontColorDark`, which corresponds to `colors.grey50` and `colors.white`. But I think the best solution here is to just add a new color in the theme for this.

mcleinman: Totally - The design was to add new colors and variables when needed, just never do it in QML.

This also made realize that while I added automated checks to force this behavior, it isn't documented in a readme. PR incoming to rectify this.

src/translations/strings.yaml
oskirby: You are correct, yes!

src/translations/strings.yaml
oskirby: Yeah, in the original images we grabbed the text directly from the macOS system screen and the intention was that the text should match. I'll search around to see if I can find some apple sources with the desired translations.

oskirby: It looks like Apple used to publish their translation files for macOS under [Glossaries](https://developer.apple.com/download/all/?q=Glossaries) but they stopped doing that in macOS 11 and later, and I am struggling to find something more modern.

oskirby: Aha, looks like there is a page in the Mac User Guide that does contain translations we can use: [Change Login Items & Extensions settings on Mac](https://support.apple.com/en-ca/guide/mac-help/mtusr003/mac)

I failed to find it on the first iteration, because it's not in the user guide's table of contents... but we already linked it in the strings.yaml and I just didn't read it closely enough.

mcleinman (approved):
Looks great, thanks.

src/translations/strings.yaml
```diff
@@ -988,6 +988,15 @@ permissionMacos:
   body:
     value: To use Mozilla VPN, please allow it to run in the background.
     comment: Paragraph requesting the user to allow the VPN to run in the background.
+  imageTitleBackground:
+    value: Allow in the Background
+    comment: Title in the permission required image, showing the Allow in the Background section of Login Items panel in the system preferences. See https://support.apple.com/en-ca/guide/mac-help/mtusr003/15.0/mac/15.0 for reference translations of these features provided by macOS.
```
flodolo, line 993: All these comment should not have `en-CA` in the link.

flodolo (commented):
Was there a specific urgency to merge these strings without review?

oskirby:
> Was there a specific urgency to merge these strings without review?

Sorry, that was my mistake. Should I revert this PR?

flodolo:
I defer to @mcleinman, since we're completing localization of what's new content for RC, and I don't think these should be included.

oskirby:
My appologies, I thought it would have been fine after the RC branch had been created, I will revert this ASAP

oskirby:
@flod the PR to revert that change is now up: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11438

PR #11435, opened 2026-07-14
Add XDG_CURRENT_DESKTOP to log artifacts

oskirby:
## Description
A common debugging question that we need to ask of Linux users is "what desktop shell are you using?" and this is something that we could just embed into the logs. So, let's do it to save ourself a roundtrip with the user during debug.

## Reference

    i.e Jira or Github issue URL

## Checklist
    

artfwo (approved):
Good one, thanks!

PR #11442, opened 2026-07-16
VPN-7709: add SBOM

mcleinman:
## Description

Per our discussion, there are some questions about whether this will include all our dependencies (especially Qt). However, we need to use this tool, so let's add it in and see what the output looks like.

From [instructions](https://mozilla-hub.atlassian.net/wiki/spaces/MozPolicies/pages/2770665546/Software+Bill+of+Materials+SBOM+Standard):

> Engineers must only use the Mozilla-approved reusable workflow for all SBOM generation. Use of third-party or custom SBOM tooling outside of the officially sanctioned workflow is not permitted, in order to ensure consistency, accuracy, and compliance across all products.

Additionally, it will produce one SBOM per release, even though not all dependencies are used by all platforms. I'll check in with the tool's creators to make sure this is okay.

## Reference

Email to engineers, VPN-7709

## Checklist
    
### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -1,0 +1,16 @@
+name: ssdlc-sbom
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+permissions:
+  actions: read
+  contents: write
+
+jobs:
+  sbom:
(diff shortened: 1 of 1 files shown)


oskirby (approved):
LGTM

encrypt94 (approved):
Qt produces its own SBOM during the build. An against the instructions way we can solve the missing Qt issue is by extracting those SBOMs from the build task and merging the SPDX files with the output of ssldc

PR #11450, opened 2026-07-21
Do not abort login when the location lookup fails

artfwo:
## Description

This is up for discussion, see linked issue. Possible fix for a bug where a transient network hiccup during login logs the user back out with a misleading "No internet connection, try again", even though authentication actually succeeded.

A Windows 11 user could not sign in over wired Ethernet (Wi-Fi worked fine) with a "No internet connection, try again." error message. Login isn't failing according to the logs: the OAuth flow completed, and the device was registered. The follow-up request to `/api/v1/vpn/ipinfo` (from `TaskGetLocation`) failed though, which reset with `RemoteHostClosedError`.

I'd say we could schedule the location task with `DoNotPropagateError`. The task still emits `completed()`, so the login flow proceeds to the main state; if the lookup fails, auto server selection will be skipped and the client can use the default server.

TaskAuthenticate, TaskAddDevice, TaskRemoveDevice, TaskAccount, TaskServers are unchanged, they still have PropagateError (or report errors unconditionally), so a genuine failure there will still propagate.

## Reference

[VPN-7712](https://mozilla-hub.atlassian.net/browse/VPN-7712)

## Checklist
    
[VPN-7712]: https://mozilla-hub.atlassian.net/browse/VPN-7712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mozillavpn.cpp
+++ b/src/mozillavpn.cpp
@@ -1335,8 +1335,9 @@ void MozillaVPN::scheduleRefreshDataTasks() {
   // https://mozilla-hub.atlassian.net/browse/VPN-3726 for more information.
   if (!m_private->m_location.initialized() &&
       !m_private->m_controller.isActive()) {
+    // Location is best-effort: failure should not abort a successful login.
     TaskGetLocation* locationTask =
-        new TaskGetLocation(ErrorHandler::PropagateError);
+        new TaskGetLocation(ErrorHandler::DoNotPropagateError);
     connect(locationTask, &TaskGetLocation::completed, [this]() {
       if (!m_locationInitialized) {
         logger.debug() << "Location initialized";


mcleinman (approved):
This seems like a reasonable solution. Thanks.

I'm curious how this leaves the recommended server screen. The message on that screen may be non-optimal, but we have the "refresh" button and I think that's okay given how rare of a case this is. I don't think it's worth more time to try to improve that "recommended server" experience as part of this fix.

artfwo:
> This seems like a reasonable solution. Thanks.
> 
> I'm curious how this leaves the recommended server screen. The message on that screen may be non-optimal, but we have the "refresh" button and I think that's okay given how rare of a case this is. I don't think it's worth more time to try to improve that "recommended server" experience as part of this fix.

When the error in TaskGetLocation is simulated, I see no messages on main or recommended server screen. When recommended server list is refreshed, it will be updated  (from latency/servers) and remain correct.

RecommendedLocationModel currently doesn't listen to Location::changed, so a location update won't refresh the model by itself.

oskirby (approved):

PR #11458, opened 2026-07-24
VPN-7720: ECDSA Signature validation failure on Balrog prod.

oskirby:
## Description
We have had some reports of the Balrog update flow failing on Windows/ARM recently. And after digging into things, it seems that there is an edge case in the ASN.1 integer conversion that we got wrong, leading to a false negative in the signature validation.

The problem occurs when one of the ECDSA signature pairs contains an unnecessary leading zero that results in an incorrect conversion into an ASN.1 signature struct, and the signature validation fails somewhere downstream from there. To fix this, we need to check for such padding and explicitly strip it out before passing it to [asn1_rs::Integer::new()](https://docs.rs/asn1-rs/latest/asn1_rs/struct.Integer.html#method.new)

## Reference
JIRA Issue: [VPN-7720](https://mozilla-hub.atlassian.net/browse/VPN-7720)

## Checklist
    
[VPN-7720]: https://mozilla-hub.atlassian.net/browse/VPN-7720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


signature/src/balrog.rs
```diff
@@ -210,6 +210,27 @@ impl<'a> Balrog<'_> {
         Err(BalrogError::HostnameMismatch)
     }
 
+    /* ASN.1 integers are signed, but ECDSA signature points are always unsigned.
+     * If the leading bit is zero, we must append an extra leading zero byte to
+     * ensure the sign is positive.
+     */
+    fn ecdsa_uint_to_asn1(value: &[u8]) -> Vec<u8> {
+        // Add padding to ensure the sign is positive.
+        if value[0] & 0x80 != 0 {
```
bhearsum, line 219: I realize this is not a new magic number, but what is 0x80 here? I'm guessing it's https://www.compart.com/en/unicode/U+0080?

bhearsum (approved):

signature/src/balrog.rs
oskirby: To explain this bit...

The signature we get from Balrog is just a raw pair of unsigned integers, and we need to convert them into ASN.1 integers. However, ASN.1 integers are variable-length signed values, with the leading bit in the leading byte indicating the sign. The test `value[0] & 0x80` is checking if the most-significant bit in the unsigned integer would conflict with the sign bit.

lesleyjanenorton:
Ceremonial R+

Thank you Naomi!

PR #11474, opened 2026-08-03
VPN-7726: Change to silent server switch

mcleinman:
## Description

Changing from updating config to a silent server switch - which among other things, suppresses the notification. Given that silent server switches eventually change the server data, I think this the proper route - though I may be missing something here, as I'm not as familiar with this new feature.

Additionally, I'd kindly ask for pointers on testing any obfuscation gotchas - it seems like this works well, but I may be missing some finer points.

This has been updated to randomize server selection when needed. (When picking a new random server, it only considers those servers in a given location that support the current obfuscation method.) 

For posterity in reviewing, this is the flow:
1. `SettingsWatcher::maybeServerSwitch` creates `TaskControllerAction(TaskControllerAction::eSilentSwitch, Controller::eServerCoolDownNotNeeded)`
2. When that task is run it calls `controller->silentSwitchServers(m_serverCoolDownPolicy)`
3. `silentSwitchServers` creates a `selectionPolicy` variable. It defaults to `DoNotRandomizeServerSelection`, and updates that to `RandomizeServerSelection` only if the cooldown policy passed in is `eServerCoolDownNeeded` - which is not what was passed in. This current PR adds something to change this to `RandomizeServerSelection` if the current server does not support the current obfuscation method.
4. `silentSwitchServers` calls `activateInternal` with that `selectionPolicy` variable.
5. `activateInternal` calls `setupConfigs` with that `selectionPolicy` variable.

## Reference

VPN-7726

## Checklist
    

src/settingswatcher.cpp
```diff
@@ -64,6 +64,10 @@ SettingsWatcher::SettingsWatcher(QObject* parent) : QObject(parent) {
           m_operationRunning = false;
         }
       }));
+
+  m_connections.append(connect(settingsHolder,
+                               &SettingsHolder::obfuscationPolicyChanged, this,
+                               [this]() { maybeServerSwitch(); }));
```
encrypt94: NIT: can't we just use `&SettingsWatcher::maybeServerSwitch` instead of the lambda?

encrypt94 (approved):
r+wc thank you!

src/settingswatcher.cpp
mcleinman: 🤦🏻 Yes, definitely. I've updated this. Thanks!

PR #11477, opened 2026-08-04
VPN-7666: Check write permissions for settings file, and add backup settings file location

mcleinman:
## Description

Occasionally users report having to re-authenticate every time the client is opened. In VPN-7666, we realized one cause was the `~/.config` directory (where we save the settings file) being owned by root. Since the client is launched by a non-root user account, it was unable to save the settings file to disk, causing the re-authentication on each launch.

We do two things in this PR:
- Check for write-ability of settings file, and log something when that isn't possible - previously, logs weren't alerting us that anything was amiss
- Use a backup location for the settings file if the main location cannot be reached. (Of course, if the main settings file location becomes available in the future, it will default to using that one.)

If the backup location is also blocked, the user will still have the "settings don't save" issue. However, we're now logging something for this - and given how rare we expect this case to be, I do not believe it's worthwhile to spend the time adding UX for this corner case.

## Reference

VPN-7666

## Checklist
    
### DIFF
--- a/src/settings/settingsmanager.cpp
+++ b/src/settings/settingsmanager.cpp
@@ -41,7 +41,48 @@ SettingsManager* SettingsManager::instance() {
 }
 
 // static
-QString SettingsManager::getOrganizationName() {
+QString SettingsManager::getOrganizationNameAndCheckPath() {
+  QString mainOrgName = formatOrganizationName();
+
+  // Occasionally users report having to re-authenticate every time the client
+  // is opened. In VPN-7666, we realized one cause was the ~/.config directory
+  // (where we save the settings file) being owned by root. Since the client is
+  // launched by a non-root user account, it was unable to save the settings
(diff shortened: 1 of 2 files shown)


oskirby (approved):
This seems reasonable. Looks good.

oskirby:
@mcleinman I think something has gone awry with this PR, and it has caused a mangled settings file name on Windows. The daemon logs are now getting stored as `C:\Windows\system32\config\systemprofile\AppData\Local\.log`

mcleinman:
> @mcleinman I think something has gone awry with this PR, and it has caused a mangled settings file name on Windows. The daemon logs are now getting stored as `C:\Windows\system32\config\systemprofile\AppData\Local\.log`

Looking at this code, I'm not quite sure how these changes would have affected that. But I want to fix it. Can you DM me logs? And given that I'm out for much of the next few days, if we need to just back this out for now, we can.

PR #11481, opened 2026-08-06
Add detailed connection info to IPInfoPanel

encrypt94:
## Description

> [!NOTE]
> This feature is intended for QA use and disabled by default

To run QA tests on the obfuscation features we use a set of Wireshark filters that require the entry server IP address.

We do not expose this information anywhere in the application, making the entry server IP difficult to obtain. This is especially true when obfuscation features are enabled (because the entry server endpoint is forced to 127.0.0.1 and not visible in the WireGuard configuration) or when testing on mobile platforms.

This PR introduces a new `showDetailedConnectionInfo` feature, disabled by default, that allows QA to access this information from the Connection info screen.

The current server information is stored in the `MozillaVPN`'s `ServerData` singleton and is updated when the controller transitions to `StateOn`.

Bonus: Fixes a minor issue in the feature list where the toggles could be pushed off-screen by long labels.

Screenshot from 2026-08-06 22-56-33
Screenshot from 2026-08-06 22-56-56
Screenshot from 2026-08-06 22-57-18

## Reference

[Jira Issue](https://mozilla-hub.atlassian.net/browse/VPN-7732)

## Checklist
    
### DIFF
--- a/nebula/ui/utils/MZAssetLookup.js
+++ b/nebula/ui/utils/MZAssetLookup.js
@@ -255,6 +255,10 @@ var imageLookup = {
     filenameLight: 'qrc:/nebula/resources/copy.svg',
     filenameDark: 'qrc:/nebula/resources/copy-dark.svg'
   },
+  'CopyIPInfo': {
+    filenameLight: 'qrc:/nebula/resources/copy-dark.svg',
+    filenameDark: 'qrc:/nebula/resources/copy-dark.svg'
+  },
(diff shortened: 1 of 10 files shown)


flodolo:
> This feature is intended for QA use and disabled by default

So, who can access the feature? If it's just for developers, we shouldn't probably expose strings for localization.

artfwo (approved):

mcleinman (approved):

PR #11484, opened 2026-08-07
Fix Linux crasher when NetworkManager is not running.

markstos:
On Linux, NetworkManager is popular but not univerally installed.

On Arch Linux for example, users can build up whatever networking stack they like. The Flatpak build assumes that NetworkManager is running and crashes when it isn't.

This patch prevents the crash, providing a generic error message instead.

Personally, I'd be interested in more extensive change communicates to the user the specific problem that NetworkManager isn't running, but that could be done in a separate scope of there is interest.

## Checklist
    
- [x ] If bug/feature is at all notable, I've added a draft release note as a comment in `addons/strings.yaml`. (Not notable-- rarely triggered)
- [x] I have performed a self review of my own code.  (Manually tested new build to confirm fix.)
- [x] I have commented my code PARTICULARLY in hard to understand areas (N/A)
- [x] I have added thorough tests where needed (N/A)

### DIFF
--- a/src/platforms/linux/netmgrcontroller.cpp
+++ b/src/platforms/linux/netmgrcontroller.cpp
@@ -39,6 +39,10 @@ NetmgrController::NetmgrController() {
   m_client =
       new QDBusInterface(DBUS_NM_SERVICE, DBUS_NM_PATH, DBUS_NM_INTERFACE,
                          QDBusConnection::systemBus(), this);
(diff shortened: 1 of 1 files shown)


encrypt94:
Hi @markstos sorry for the late review! thanks a lot for your PR!
I've opened a follow-up ticket to provide a proper error screen when NetworkManager is not detected.

PR #11489, opened 2026-08-10
Move setExitServerPublicKey and setEntryServerPublicKey into Controller::setupConfigs

encrypt94:
## Description

This is a follow-up to #11481.

We set the server's public keys in `Controller::activateInternal` or `Controller::maybeSendUpdatedConfigs`, based on what is returned by `Controller::setupConfigs`. When using Mullvad's multi-hop, we compact the exit and entry into a single `InterfaceConfig`, losing the information about the entry server and exposing only the exit.

With this PR, we move the `set(Exit|Entry)ServerPublicKey` calls to the end of `Controller::setupConfigs` so that we can correctly keep track of the two logical hops.

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -461,6 +461,11 @@ auto Controller::setupConfigs(
     exitConfig.m_vpnDisabledApps = settingsHolder->vpnDisabledApps();
   }
 
+  // Track the logical entry server independently of how many WireGuard peers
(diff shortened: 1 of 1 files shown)


strseb (approved):
lgtm c:

PR #11495, opened 2026-08-13
Obfuscation features part 4 - Flatpak

encrypt94:
## Description

 Enable obfuscation features in Flatpak builds:
- add stop / restart capability to `Obfuscator`
- defer connection inizialization to activate
- build and install  `mozillavpn-obfuscator`  for Flatpak
- spin the `mozillavpn-obfuscator` inside `NetmgrController`  and rewrite the peer to `127.0.0.1`
- add a **throw** route for the VPN server into the WireGuard policy table so the relay's traffic to the server escapes the tunnel 
- the throw route requires NetWork Manager [1.38](https://networkmanager.dev/docs/api/1.38/nm-settings-nmcli.html), emit a `backendFailure` and log a warning if network manager is too old (jammy is still on [1.36](https://networkmanager.dev/docs/api/1.36/nm-settings-nmcli.html))

## Reference

[VPN-7737](https://mozilla-hub.atlassian.net/browse/VPN-7737)

## Checklist
    
[VPN-7737]: https://mozilla-hub.atlassian.net/browse/VPN-7737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/platforms/linux/netmgrcontroller.cpp
+++ b/src/platforms/linux/netmgrcontroller.cpp
@@ -393,7 +393,11 @@ void NetmgrController::activateCompleted(const QDBusObjectPath& path) {
 }
 
 void NetmgrController::deactivate() {
-  m_obfuscator.reset();
+  // Stop the obfuscator relay but keep the instance, so it can be restarted on
+  // the same local port if the tunnel is reactivated from NetworkManager.
+  if (m_obfuscator) {
+    m_obfuscator->stop();
+  }
 
   if (!m_device || m_device->uuid() != m_uuid) {
     logger.warning() << "Client already disconnected";
--- a/src/platforms/linux/netmgrcontroller.cpp
+++ b/src/platforms/linux/netmgrcontroller.cpp
@@ -53,6 +53,18 @@ NetmgrController::NetmgrController() {
 NetmgrController::~NetmgrController() {
   MZ_COUNT_DTOR(NetmgrController);
   logger.debug() << "Destroying NetmgrController";
+
+  // Stop the obfuscator relay first so its process doesn't outlive us.
+  m_obfuscator.reset();
+
+  // Remove the connection from NetworkManager 
+  if (m_remote) {
+    QDBusReply<void> reply = m_remote->call("Delete");
+    if (!reply.isValid()) {
+      logger.warning() << "Failed to delete connection on shutdown:"
+                       << reply.error().message();
+    }
+  }
 }
 
 // static
--- a/src/platforms/linux/netmgrcontroller.cpp
+++ b/src/platforms/linux/netmgrcontroller.cpp
@@ -57,7 +57,7 @@ NetmgrController::~NetmgrController() {
   // Stop the obfuscator relay first so its process doesn't outlive us.
   m_obfuscator.reset();
 
-  // Remove the connection from NetworkManager 
+  // Remove the connection from NetworkManager
   if (m_remote) {
     QDBusReply<void> reply = m_remote->call("Delete");
     if (!reply.isValid()) {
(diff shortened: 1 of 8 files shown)


oskirby:
So, one of the activation paths available for Flatpak builds is where the user opens the Network control center, and turns on the VPN via the interface toggle in their OS. Normally, this takes the most recent VPN configuration and starts the interface.

The VPN client can detect this via the `deviceStateChanged()` signal and will update the VPN UI to match the network manager's interface state. However, during this event, there is nothing that would start the obfuscator, so it seems unlikely that the connection will activate correctly if an obfuscator was configured.

encrypt94:
> So, one of the activation paths available for Flatpak builds is where the user opens the Network control center, and turns on the VPN via the interface toggle in their OS. Normally, this takes the most recent VPN configuration and starts the interface.
> 

This turned out into a pile of painful corner cases. :/

Changes made to address the issue:
- destroy the connection when the client exits so it cannot be activated when we have no way to activate the obfuscator daemon
- delete any stale connection when initializing `NetmgrController`
- defer `AddConnection2`  to `activate` so that we don't have any connection on cold start. The connection will be created only after a user action or if auto connect is enabled. This is done with a blocking call with a 25s default timeout. Using a blocking call should not be a problem, this operation is nearly instant and we already use blocking calls
- add stop and restart capabilities to the obfuscator
- react to `deviceStateChanged()` by restarting the obfuscator if needed.
- stop the obfuscator on deactivate. This stops the QProcess but keeps the instance so it can be restarted if the VPNis deactivated from the client and then reactivated from the Network Control Center.

src/platforms/linux/netmgrcontroller.cpp
```diff
@@ -271,7 +248,60 @@ void NetmgrController::activate(const InterfaceConfig& config,
...
 
-  NetmgrDataList peers(wgPeer(config));
+  // The traffic to the real server must bypass the tunnel.
+  // In the Flatpak sandbox, we cannot use SO_MARK so instead we set
+  // "throw" route for each server endpoint into the WireGuard policy table
+  // "throw" needs NetworkManager >= 1.38.
+  if (obfuscate) {
+    static const QVersionNumber kThrowRouteMinVersion(1, 38);
+    if (m_version < kThrowRouteMinVersion) {
+      logger.warning() << "NetworkManager" << m_version.toString()
```
oskirby, line 258: What is the user experience if this condition is not met? Are the obfuscation features disabled, or does this render their VPN non-functional if they try to enable obfuscation?

src/platforms/linux/netmgrcontroller.cpp
```diff
@@ -294,6 +324,28 @@ void NetmgrController::activate(const InterfaceConfig& config,
...
+    logger.info() << "Adding connection:" << m_uuid;
+    QString path = QStringLiteral(DBUS_NM_PATH) + "/Settings";
+    QDBusInterface iface(DBUS_NM_SERVICE, path, nmInterface("Settings"),
+                         m_client->connection());
+    QVariantList addArgs;
+    addArgs << serializeConfig();
+    addArgs << (uint)IN_MEMORY;
+    addArgs << QVariantMap();
+    // Use a blocking AddConnection2 to keep the flow simple
+    QDBusReply<QDBusObjectPath> added =
```
oskirby, line 337: This is probably fine, but it will hang the UI thread until such time as we get the response over D-Bus. Are we sure that this call will always finish quickly (eg: no polkit authorization prompts to get in the way)?

PR #11496, opened 2026-08-13
Nit: add padding between obfuscation label and timer

encrypt94:
## Description

Add some padding between obfuscation label and timer. Shift everything a little up when obfuscation is enabled to avoid overlapping the toggle button.

Screenshot from 2026-08-13 17-26-21
Screenshot from 2026-08-13 17-27-15

## Reference

    i.e Jira or Github issue URL

## Checklist
    
### DIFF
--- a/src/ui/screens/home/controller/ControllerView.qml
+++ b/src/ui/screens/home/controller/ControllerView.qml
@@ -535,11 +535,15 @@ Item {
     Column {
         id: col
 
+        property int baseTopMargin: 24
+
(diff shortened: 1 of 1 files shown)


artfwo (approved):
Is the extra shifting even necessary (for the non-obfuscated case, relative to obfuscated)? Can the first line just stay at the same Y position, while the 2nd gets hidden when obfuscation is disabled?

mcleinman (approved):

PR #11502, opened 2026-08-17
adjust margins

mcleinman:
## Description

This matches other rows a bit better. I'd love to also give some vertical padding, but I played around with it for a bit and couldn't find a good way to add some.

And while we're in there, let's change to using the variables.

Before
Screenshot 161

After
Screenshot 162

## Reference

N/A

## Checklist
    

encrypt94 (approved):
Thank you!

PR #11508, opened 2026-08-19
Add "force update" info to documentation

mcleinman:
So we have some documentation next time this is needed.


oskirby (approved):
Looks good!

PR #11511, opened 2026-08-20
VPN-7730: Android piece

mcleinman:
I think this code should probably go somewhere else, since we use similar snippets in a few places at this point. 

But I don't want to stuff it into `Server::choosePort`, because I don't want to make `Server` dependent on settings. We could stuff it into the Controller, but Controller is already a beast. Open to other ideas, or just merging this one.

### DIFF
--- a/src/platforms/android/androidcontroller.cpp
+++ b/src/platforms/android/androidcontroller.cpp
@@ -188,6 +188,24 @@ void AndroidController::activate(const InterfaceConfig& config,
   QJsonArray jServers;
 
   jServers.append(jMainServer);
+
+  // Fallback/backup servers must pick their port the same way the main server
+  // did in Controller::setupConfigs(), otherwise a switch to a fallback server
+  // will silently drop the forced port.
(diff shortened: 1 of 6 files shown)


encrypt94 (approved):

src/platforms/android/androidcontroller.cpp
```diff
@@ -188,14 +189,32 @@ void AndroidController::activate(const InterfaceConfig& config,
   QJsonArray jServers;
 
   jServers.append(jMainServer);
+
+  // Fallback/backup servers must pick their port the same way the main server
+  // did in Controller::setupConfigs(), otherwise a switch to a fallback server
+  // will silently drop the forced port.
+  SettingsHolder::ObfuscationPolicy obfuscationPolicy =
+      static_cast<SettingsHolder::ObfuscationPolicy>(
+          settingsHolder->obfuscationPolicy());
```
encrypt94: s/settingsHolder/SettingsHolder::instance()/

PR #11521, opened 2026-08-25
VPN-7745 - adjust addons for force update warnings

mcleinman:
## Description

This reverses changes from https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11507, and updates the timing for the addons.

- Turns out the final deadline is going to remain the same, so no changes needed to the string. Adding it back in.
- We are going to show the first message a little bit later. In adjusting this, I realized we have an addon condition that handles timing, so and that seems more straightforward than manually re-implementing that in the javascript file. So I adjusted how we're handling timing.

I tested this by manually adjusting the system clock, and it seems to work.

## Reference

VPN-7745

## Checklist
    
### DIFF
--- a/addons/message_mandatory_update_to_web_auth_1/manifest.json
+++ b/addons/message_mandatory_update_to_web_auth_1/manifest.json
@@ -5,18 +5,20 @@
   "type": "message",
   "conditions": {
     "max_client_version": "2.34.9",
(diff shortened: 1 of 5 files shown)


flodolo (approved):

oskirby (approved):
LGTM

PR #11522, opened 2026-08-26
Update DNS resolvers in place during a server switch

artfwo:
## Description

Remove redundant reverting and restarting resolvers on a live interface under Windows and Linux, as `updateResolvers()` replaces resolver configuration on these platforms.

## Reference

[VPN-7733](https://mozilla-hub.atlassian.net/browse/VPN-7733)

## Checklist
    
[VPN-7733]: https://mozilla-hub.atlassian.net/browse/VPN-7733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/daemon/daemon.cpp
+++ b/src/daemon/daemon.cpp
@@ -85,10 +85,17 @@ bool Daemon::activate(const InterfaceConfig& config) {
         return false;
       }
 
+#if defined(MZ_MACOS)
+      // macOS restores the resolvers by terminating the DNS manager process,
+      // so they must be reverted before a new configuration can be applied
+      // (see https://mozilla-hub.atlassian.net/browse/VPN-7752)
       if (!dnsutils()->restoreResolvers()) {
         return false;
       }
+#endif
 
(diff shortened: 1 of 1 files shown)


encrypt94 (approved):
LGTM

src/daemon/daemon.cpp
```diff
@@ -85,10 +85,16 @@ bool Daemon::activate(const InterfaceConfig& config) {
         return false;
       }
 
+#if defined(MZ_MACOS)
```
oskirby, line 88: We should at least leave a breadcrumb here to note that we should find a solution to support in-place DNS updates for macOS too either here or in some future work.

oskirby (approved):

src/daemon/daemon.cpp
artfwo: Right, thanks for reminder, I've added a JIRA link.

PR #11552, opened 2026-09-15
Force update addon: show date as Oct 1/8 always, do not show on older iOS version

mcleinman:
In discussions with QA, we realized we don't want this date to be the "last client update" timestamp - as it would show both messages as the same timestamp. Removing this lets the natural date (Oct 1 for the first one, Oct 8 for the second one) show.

Additionally, we want to adjust which iOS versions we show this in.


oskirby (approved):
LGTM

PR #11562, opened 2026-09-17
VPN-7788: use Switch instead of SilentSwitch if we are in StateConnectionError

encrypt94:
## Description

We introduced this issue when we switched from a regular switch to a silent server switch on obfuscationPolicyChanged. After entering Server Unavailable mode, the controller state is StateConnectionError. SettingsWatcher::maybeServerSwitch requires the controller to be in StateOn with no operations in progress; otherwise, it is a no-op.

This is a workaround for 2.40. The proper solution is to use the silent switch.

## Reference

Introduced by #11474 
<https://mozilla-hub.atlassian.net/browse/VPN-7788>

## Checklist
    

src/settingswatcher.cpp
```diff
@@ -104,4 +104,27 @@ void SettingsWatcher::maybeServerSwitch() {
...
+
+  Controller* controller = MozillaVPN::instance()->controller();
+
+  // While connected, apply the new obfuscation policy with a silent server
+  // switch, so there is no user-visible disruption (VPN-7726).
+  if (controller->state() == Controller::StateOn) {
+    maybeServerSwitch();
+    return;
+  }
+
```
mcleinman: We're not considering other states like `StateConfirming` and `StateConnecting`. What if we did the check for connection error and such (lines 121/122) first, and if the conditions aren't met, we always run `maybeServerSwtich()`?

mcleinman (changes requested):

src/settingswatcher.cpp
encrypt94: good catch! thank you i changed the code to handle `StateConfirming` and `StateConnecting` `maybeServerSwitch()` will only run for StateOn, just for readability. The other states are no-ops, as they would be if we ran `maybeServerSwitch()`

mcleinman: I'd prefer to let `maybeServerSwitch` handle the no-op, rather than replicate it here. We sometimes run into bugs down the line when we switch this type of logic for a state in one place, but forget about another. Is there any harm in calling `maybeServerSwitch` in all other cases?

encrypt94: No harm at all,it was just for readability, so it’s clear from the function source code which states are handled and which are not. I changed it to call maybeServerSwitch in the default: case and moved the note about which states we will handle into a comment

mcleinman (approved):
Thanks!

PR #11566, opened 2026-09-18
VPN-7773: Fix AppTracker cgroup leaks

oskirby:
## Description
There seems to be a couple of cases where the `AppTracker` class that drives application detection and classification for split tunneling can lose track of control groups and D-Bus connections, possibly leading to resource exhaustion in the daemon or a breakdown of the split tunneling feature.

I have found the following possible edge cases:
 - Root cgroup removal is not handled in `cgroupsChanged()` leading to a leak of tracked cgroups.
 - `m_systemdInterface` gets overwritten upon each `userCreated()` signal, which 
   a. Leaks D-Bus connections, and
   b. Breaks the KDE fallback in a multi-user environment.

To fix this, we refactor the `AppTracker` to be strongly associated with a single systemd User object, and create a new `AppTracker` upon each `userCreated()` signal, and destruct it in the associated `userRemoved()` signal. While we're at it, we also try to eliminate some of the blocking D-Bus calls during the setup procedure.

## Reference
JIRA Issue: [VPN-7773](https://mozilla-hub.atlassian.net/browse/VPN-7773)

## Checklist
    
- [ ] My code follows the style guidelines for this project
- [ ] If bug/feature is at all notable, I've added a draft release note as a comment in `addons/strings.yaml`.
- [ ] I have not added any packages that contain high risk or unknown licenses (GPL,  LGPL, MPL, etc. consult with DevOps if in question)
- [ ] I have performed a self review of my own code
- [ ] I have commented my code PARTICULARLY in hard to understand areas
- [ ] I have added thorough tests where needed

[VPN-7773]: https://mozilla-hub.atlassian.net/browse/VPN-7773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11569, opened 2026-09-21
netfilter: Fix crash in traffic marking if firewall not initialized

oskirby:
## Description
I stumbled across this while testing #11566 and it should probably just be it's own standalone PR. There is a crash that can occur in the daemon during shutdown if the VPN has never been activated, in which case the netfilter chains/tables are all `nil` because they never got setup.

To fix it, make the netfilter traffic marking functions fail safely if the tables are uninitialized.
 
## Reference

    i.e Jira or Github issue URL

## Checklist
    

encrypt94 (approved):
LGTM thank you!

PR #11573, opened 2026-09-22
VPN-7750: Show a dedicated screen when NetworkManager is unavailable

artfwo:
## Description

Add `ViewNetworkManagerUnavailable` mirroring `ViewPermissionsRequiredOSX` and a corresponding signal (also semantically identical to `permissionsRequired` on macOS).

Show the new screen when NetworkManager is unavailable on Flatpak builds.

It behaves similarly to the macOS screen - shows only after the user is signed in.

Tested with `flatpak override --user --system-no-talk-name=org.freedesktop.NetworkManager org.mozilla.vpn`

## Reference

[VPN-7750](https://mozilla-hub.atlassian.net/browse/VPN-7750)

## Checklist
    
[VPN-7750]: https://mozilla-hub.atlassian.net/browse/VPN-7750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flodolo (approved):

src/ui/screens/ScreenPermissionRequired.qml
```diff
@@ -21,6 +21,8 @@ MZScreenBase {
             getStack().push("qrc:/qt/qml/Mozilla/VPN/sharedViews/ViewPermissionRequiredOSX.qml")
             return;
         }
-        // TODO: What to display otherwise?
```
oskirby, line 24: Just thinking ahead to other platforms that might need a `PermissionRequired` state, how can we make this conditional exclusive on Flatpaks? Something like `if (Qt.platform.os == "linux")` and possibly adding a `qgetenv("container") == "flatpak"`?

The reason being that some day we might want to refactor the mobile onboarding stuff to make use of this screen too.

src/platforms/linux/netmgrcontroller.cpp
```diff
@@ -62,7 +62,7 @@ QString NetmgrController::nmInterface(const QString& name) {
 void NetmgrController::initialize(const Device* device, const Keys* keys) {
   if (!m_client || !m_client->isValid()) {
     logger.warning() << "NetworkManager client is not available";
-    emit initialized(false, false, QDateTime());
+    emit networkManagerUnavailable();
```
oskirby: IMO, we should just `emit permissionRequired()` here and leave the `ControllerImpl` interface unchanged.

oskirby (approved):
My comments are pretty minor and if you feel strongly about either, feel free to ignore them.

R+

src/controller.cpp
```diff
@@ -212,6 +214,11 @@ void Controller::implPermRequired() {
   setState(StatePermissionRequired);
 }
 
+void Controller::implNetworkManagerUnavailable() {
+  logger.warning() << "Initialization blocked: NetworkManager unavailable";
+  setState(StatePermissionRequired);
```
encrypt94: I'm not sure `StatePermissionRequired` is the right fit here. We are not asking for permission, we're asking users to install a missing dependency.
I think a new state `StateInitializationError` will fit better then we can set a reason using  `setError` and show a screen based on the reaso, so this can be reused by  [this check in the obfuscators PR](https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11495/changes#diff-ca48af8468da3e7f519e5d9e919af6aa50b1ad9d916f0551c3c5b7bd2c40aa70R256)

src/platforms/linux/netmgrcontroller.cpp
encrypt94: agree about leaving the `ControllerImpl` interface unchanged.

src/controller.cpp
artfwo: If we leave ControllerImpl interface unchanged, and use permissionRequired signal to put the user on the "network manager unavailable" screen, then `StatePermissionRequired` is logically closer.

And the version check could be probably done in linux/netmgrcontroller.cpp then, wdyt?

oskirby: If we are searching for a more general name, then maybe `StateInitializationBlocked` and we can pass a reason why the initialization is blocked in the signal, eg: `emit initializationBlocked("permission required");`

But I think this is a refactor for another time.

artfwo: > But I think this is a refactor for another time.

+++

encrypt94 (approved):

src/controller.cpp
encrypt94: Works for me. Let’s merge it and refactor it later. Maybe we can open a Jira ticket as a reminder to refactor it.

artfwo: Yep, created one here https://mozilla-hub.atlassian.net/browse/VPN-7795

encrypt94: Thank you!

